PR #2, opened 2023-12-06
Small changes to README.md

gideonite:
* Cosmetic changes
* Adding a few small details

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ python -m fme.fcn_training.inference.inference examples/config-inference.yaml
 ## Configuration options
 See the `InferenceConfig` class in [this file](fme/fme/fcn_training/inference/inference.py) for
 description of configuration options. The [example config](examples/config-inference.yaml)
-shows some useful defaults for performing a 400-step (100-day) simulation (e.g. using a 6-hr time-step)
+shows some useful defaults for performing a 400-step simulation (100 days, with the 6-hour time step).
 
 ## Performance
 While inference can be performed without a GPU, it may be very slow in that case. In addition,


README.md
```diff
@@ -1,22 +1,24 @@
...
-This is rapidly changing research software. No guarantees are made of maintaining
-backwards compatibility.
+This is rapidly changing research software. We make no guarantees of maintaining backwards compatibility.
 
 ## Quickstart
 
-1. Clone this repository. Then assuming conda is available, run
+### 1. Clone this repository and install dependencies
+
+Assuming [conda](https://docs.anaconda.com/free/anaconda/install/index.html) is available, run
```
oliverwm1:
```suggestion
Assuming [conda](https://docs.conda.io/en/latest/) is available, run
```

oliverwm1: We don't want to direct people to the full Anaconda distribution which will contain many packages that are not needed.

README.md
```diff
@@ -37,15 +39,15 @@ python -m fme.fcn_training.inference.inference examples/config-inference.yaml
...
+shows some useful defaults for performing a 400-step (100-day) simulation (e.g. using a 6-hr time-step)
 
 ## Performance
 While inference can be performed without a GPU, it may be very slow in that case. In addition,
 I/O performance is critical for fast inference due to loading of forcing data and target data
 during inference.
 
 ## Analyzing output
-Various metrics are computed online by the inference code. These can be viewed via
+Various training and climate performance metrics are computed online by the inference code. These can be viewed via
```
oliverwm1:
```suggestion
Various climate performance metrics are computed online by the inference code. These can be viewed via
```
We don't have training code here, so don't think we should mention training metrics.

README.md
```diff
@@ -37,15 +39,15 @@ python -m fme.fcn_training.inference.inference examples/config-inference.yaml
 ## Configuration options
 See the `InferenceConfig` class in [this file](fme/fme/fcn_training/inference/inference.py) for
 description of configuration options. The [example config](examples/config-inference.yaml)
-shows some useful defaults for performing a 400-step (100-day) simulation.
+shows some useful defaults for performing a 400-step (100-day) simulation (e.g. using a 6-hr time-step)
```
oliverwm1:
```suggestion
shows some useful defaults for performing a 400-step simulation (100 days, with the 6-hour time step).
```
Otherwise I find this makes it sound like the 6-hour step is configurable and so increases confusion.

README.md
gideonite: Aha of course

gideonite:
Thanks for the feedback, great points

PR #3, opened 2024-03-07
Update README to link to zenodo repo with checkpoint

oliverwm1:
Add a link to a zenodo data repository which contains model checkpoint and a 1-year subsample of the validation data.

### DIFF
--- a/examples/config-inference.yaml
+++ b/examples/config-inference.yaml
@@ -11,7 +11,7 @@ logging:
   entity: your_wandb_entity
 validation_data:
   batch_size: 1
-  data_path: /path/to/validation/ic_0011
+  data_path: /path/to/data/directory
   data_type: xarray
   num_data_workers: 4
   n_samples: 1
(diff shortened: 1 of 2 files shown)


examples/config-inference.yaml
```diff
@@ -11,7 +11,7 @@ logging:
   entity: your_wandb_entity
 validation_data:
   batch_size: 1
-  data_path: /path/to/validation/ic_0011
+  data_path: /path/to/data/directory
```
gideonite: question: Do we want to document what we expect to be present in this directory? e.g. netcdf files named in a particular way.

gideonite (commented):
Looks like a big improvement. One question about the data_path

examples/config-inference.yaml
oliverwm1: I can add a short comment, but mostly I expect users to infer this from the example data we provide.

gideonite (approved):
LGTM!

PR #5, opened 2024-07-25
New public release

oliverwm1:
Update public code to latest.


mcgibbon:
When we do this in the future, I would suggest having the first commit be a replacement rather than deletion of the code, so it's possible to see what files have changed. I don't think this is super important to review but I was curious to check it for any potential issues.

fme/fme/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.1.0"
+__version__ = "0.2.0"
```
mcgibbon, line 1: We should update this in full-model also.

README.md
```diff
@@ -1,8 +1,8 @@
 # ACE: AI2 Climate Emulator
 This repo contains the inference code accompanying "ACE: A fast, skillful learned global atmospheric model for climate prediction" ([arxiv:2310.02074](https://arxiv.org/abs/2310.02074)).
 
-## DISCLAIMER
-This is rapidly changing research software. We make no guarantees of maintaining backwards compatibility.
+## Documentation
```
mcgibbon: Did you mean to remove the DISCLAIMER?

I would suggest not adding a header for a documentation link until we actually have said link. But we should probably include instructions for building the docs locally.

README.md
```diff
@@ -35,27 +35,12 @@ enough data to span the desired prediction period. The checkpoint is also availa
 `gs://ai2cm-public-requester-pays/2023-11-29-ai2-climate-emulator-v1/checkpoints/ace_ckpt.tar`.
 
 ### 3. Update configuration and run
-Update the paths in the [example config](examples/config-inference.yaml). Then in the
+Update the paths in the [example config](fme/docs/inference-config.yaml). Then in the
 `fme` conda environment, run inference with:
 ```
-python -m fme.fcn_training.inference.inference examples/config-inference.yaml
+python -m fme.ace.inference fme/docs/inference-config.yaml
```
mcgibbon, line 38: Do you want the README to show how to run the inference entrypoint or the evaluator entrypoint? Or do you want to delete this quickstart section entirely and reference the docs instead (I think we should do this when we have RTD docs if not now)?

README.md
```diff
@@ -35,27 +35,12 @@ enough data to span the desired prediction period. The checkpoint is also availa
...
-## Performance
-While inference can be performed without a GPU, it may be very slow in that case. In addition,
-I/O performance is critical for fast inference due to loading of forcing data and target data
-during inference.
-
-## Analyzing output
-Various climate performance metrics are computed online by the inference code. These can be viewed via
-[wandb](https://wandb.ai) by setting `logging.log_to_wandb` to true and updating `logging.entity`
-to your wandb entity. Additionally, raw output data is saved to netCDF by the inference code.
-
```
mcgibbon, line 58: The docs don't have any kind of wandb instructions right now, do we want to include some? Could make an issue in full-model if we do, or just decide we don't.

mcgibbon (commented):
LGTM. Some minor comments.

README.md
oliverwm1: It was intentional—we have a note in the docs about this being research software which I think is sufficient.

Sounds good, I'll remove the documentation section. I don't think we need to bother with local build instructions since we'll have the RTD page up soon.

README.md
oliverwm1: I wanted it to show the inference entrypoint, since this has the least requirements and I expect it is how most people will use the model. For now, I think we should keep the quickstart, but I agree once the RTD is hosted we could just have a link instead.

README.md
oliverwm1: There is a short section on Wandb integration in the docs quickstart. We could add more details there but for now I think it's okay.

README.md
```diff
@@ -1,9 +1,6 @@
 # ACE: AI2 Climate Emulator
 This repo contains the inference code accompanying "ACE: A fast, skillful learned global atmospheric model for climate prediction" ([arxiv:2310.02074](https://arxiv.org/abs/2310.02074)).
 
-## DISCLAIMER
-This is rapidly changing research software. We make no guarantees of maintaining backwards compatibility.
```
mcgibbon, line 5: Do we mean to remove this disclaimer?

mcgibbon (approved):
LGTM

README.md
oliverwm1: Yes, this was intentional. There is a similar warning in the docs.

Makefile
```diff
@@ -1,11 +1,16 @@
 VERSION ?= $(shell git rev-parse --short HEAD)
 IMAGE ?= fme
+REGISTRY ?= registry.nersc.gov/m4492/ai2cm
```
frodre: Could probably remove the NERSC specific items for REGISTRY and `push_shifter_image`

frodre (commented):
Thanks, @oliverwm1!   

I was thinking about most of the configuration/data processing files in the `scripts` folder.  Generally, I feel like it adds a lot of noise since there are a few important files for generating stats / ENSO metrics among them, but they are separated by folder so people could also probably just look for what they need.  

Personally, since I think this will be one of our more widely seen and used repositories, I think it feels better just to have only relevant files distributed in a package rather than our internal processing.  I would vote for cleaning up `scripts`.   BUT, I realize this also adds an inconvenient divergence when merging into this repository as those scripts might be updated in `full-model` and create conflicts if they are deleted here.  

Also, what's the plan for the documentation? Add that in after this merge?  We could also consider linking the ACE colab notebook with a note at the beginning of the quickstart.

Makefile
oliverwm1: Yup thanks for reminder—thought I had done this.

oliverwm1:
>Personally, since I think this will be one of our more widely seen and used repositories, I think it feels better just to have only relevant files distributed in a package rather than our internal processing. I would vote for cleaning up scripts. BUT, I realize this also adds an inconvenient divergence when merging into this repository as those scripts might be updated in full-model and create conflicts if they are deleted here.

I'm not too worried about conflicts. I did already delete some of the things in the `scripts` directory but agreed there is a lot of stuff in `data_process` that is not widely applicable. I'll make a commit deleting everything other than `get_stats.py` and see what others think.

> Also, what's the plan for the documentation? Add that in after this merge? We could also consider linking the ACE colab notebook with a note at the beginning of the quickstart.

Yes, building the docs / hosting on RTD will be a follow on PR. Agreed, once this is merged to main and the colab notebook can reflect that, we can add a link.

frodre (approved):
LGTM, thanks for the cleanup in the scripts folder

PR #6, opened 2024-07-31
Add readthedocs config

mcgibbon:
This PR adds configuration needed for readthedocs. A test RTD project is created [here](https://ace-testing.readthedocs.io/en/latest/index.html). This will be removed and re-created under the main branch and a long-term name after this PR is merged.

Changes:
- Added `.readthedocs.yaml`
- Added pinned requirements in `docs/requirements-pinned.txt` to be used by RTD

- [ ] Tests added


.readthedocs.yaml
```diff
@@ -0,0 +1,32 @@
...
+# formats:
+#    - pdf
+#    - epub
+
+# Optional but recommended, declare the Python requirements required
+# to build your documentation
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+   install:
+   - requirements: fme/docs/requirements-pinned.txt
```
oliverwm1: What's the reason to not just give the `fme/docs/requirements.txt` and `fme/requirements.txt` files? We will have to remember to recompile the `requirements-pinned.txt` when those change.

mcgibbon: Two reasons:
- I wasn't sure that would work (whether you could specify two files), and testing changes requires waiting for sphinx to recompile on RTD which takes a little while
- RTD strongly recommends pinning the package versions (see comment from the template directly above this), which makes sense for something as public-facing as docs.

mcgibbon:
Also FYI I am having a hard time getting automatic builds to work for this, even though the github web hook is set up and says it's working. Currently I'm launching builds manually.

.readthedocs.yaml
oliverwm1: Okay. Yeah I do see the rationale for pinning, but given the relatively infrequent updates to this repo, probably won't make much of a practical difference either way.

oliverwm1:
> Also FYI I am having a hard time getting automatic builds to work for this, even though the github web hook is set up and says it's working. Currently I'm launching builds manually.

Huh weird. Okay well yeah let's just build manually for now, and can sort out automating later on.

oliverwm1 (approved):
LGTM

PR #9, opened 2024-08-01
Add link to zenodo archive with checkpoint

oliverwm1:
Make zenodo archive with the ACE checkpoint findable without user having to click through to docs.


brianhenn (approved):
LGTM

Issue #10, opened 2024-08-26
Availability of model trained on E3SM?


ShihengDuan:
Hi ACE team,

Thanks for publishing the code and pertained model of ACE! I wonder if the zenodo repository is for the ACE model trained on GFDL model. Is there a public version of the model trained on E3SM EAM?

jpdunc23:
Hi @ShihengDuan, the weights for ACE-EAMv2 are available to download using this link (`ema_ckpt.tar`): https://portal.nersc.gov/archive/home/projects/e3sm/www/e3smv2-fme-dataset

There's an inference-ready 10-year valid dataset available at that link as well, but it's around 165 GB. 

Let us know if you run into any issues.

oliverwm1:
Also, to confirm, yes the zenodo repository (https://doi.org/10.5281/zenodo.10791087) is for the ACE model trained on FV3GFS as described in https://arxiv.org/abs/2310.02074.

PR #12, opened 2024-09-03
Add link to E3SMv2-trained paper and checkpoint

oliverwm1:
Users are interested in checkpoint trained on E3SMv2 (#10). PR adds a link to paper and checkpoint/data to the README.


frodre (approved):
LGTM.  Both links accessible, and the checkpoint data download started when I clicked it!

PR #18, opened 2024-10-02
Update for PyPI release

### LINKED ISSUE #14: Make `fme` package available on PyPI
This will make it easier to install for users.

frodre:
Updates for releases to PyPI.  Includes move to `pyproject.toml` the current suggested format for package configuration over `setup.py`.  Includes some deployment makefile targets for test/production.  

https://test.pypi.org/project/fme

Resolves #14


frodre:
Try an installation with ```uv pip install -i https://test.pypi.org/simple/ fme --extra-index-url https://pypi.org/simple```

pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "fme"
+version = "2024.10.3"
+description = "Train and evaluate weather/climate model emulators"
+readme = "README.md"
+requires-python = ">=3.9"
+license = {file = "LICENSE"}
```
frodre: I had to move the `pyproject.toml` to the top level in order to dynamically link the `README` and `LICENSE` files.  I did try to leave it under `fme`, but it's not permitted to access parent directories for dynamic builds.

@mcgibbon doesn't recall if there was a requirement that `setup.py` and other files exist in `fme` and not at the top level, but did mention that it would make things more difficult for having multiple packages.  Could be something worth discussing.

pyproject.toml
```diff
@@ -0,0 +1,38 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "fme"
+version = "2024.10.3"
```
frodre: This version number was just for testing things out on testpypi.  I'm thinking we should use the `2023.12.0` version that's tagged in the repo for actual pypi, but I wanted to check before actually committing a version.

Makefile
```diff
@@ -12,7 +13,20 @@ enter_docker_image: build_docker_image
...
 	pytest --durations 20 .
+
+# For maintainer use only
+# requires fme[deploy] to be installed
+
+build_pypi:
+	rm -rf dist
+	python -m build
+
+deploy_pypi: build_pypi
```
frodre, line 28: User needs an API key to push, so it's safe to include in make target (i.e., can't accidentally push without access, and pushing same version fails)

oliverwm1:
Although installing doesn't fail, and I can import `fme`, when I try to call a function that should be available, I get an AttributeError:
```
$ conda create -n test-pip python=3.10 pip
$ conda activate test-pip
$ pip install -i https://test.pypi.org/simple/ fme --extra-index-url https://pypi.org/simple
$ python
>>> import fme
>>> fme.get_device()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'fme' has no attribute 'get_device'
```

pyproject.toml
mcgibbon: This does fundamentally change our repo so that it is now the repo of a single python package, as opposed to being a repo containing a single python package (but which could contain more). Probably worth discussing in our weekly fme sync.

oliverwm1: I would lean to maintaining the current repo structure. Maybe we could duplicate the LICENSE within `fme` and make an `fme`-specific README within that folder? The benefit of current structure is we can have the scripts folder with data processing etc scripts that are distinct from the fme package.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "fme"
+version = "2024.10.5"
```
oliverwm1: Make sure `fme.core.version.py` is updated accordingly.

fme/requirements.txt
```diff
@@ -6,10 +6,10 @@ numpy<2
 wandb
 tensorly
 tensorly-torch
-xarray
+xarray>=2023.2.0
```
oliverwm1: Mostly out of curiosity.. was there a particular feature that required this version?

frodre: With the original setup.py `python>=3.8` the tests were failing without xarray above that version (which also forced the bump to `python>=3.9`).  I removed since the python requirement automatically solves the xarray version correctly.

oliverwm1: Thanks, that is helpful explanation. I'm +1 on requiring python 3.9 given EOL for 3.8.

frodre:
@oliverwm1 Okay, I've moved the `pyproject.toml` into the `fme` directory along with a new `README.md` and `LICENSE` to be accessed by that, updated version to rely on a single top-level `fme.__version__`, and changed that version to `2024.9.0` to reflect the last PR updates.

An example of what the package will appear as can be seen here: https://test.pypi.org/project/fme/2024.9.0/

oliverwm1 (approved):
LGTM! I tested the install from test-pypi and it worked for me.

PR #25, opened 2024-12-10
Updated public release (squash merge)

brianhenn:
Updates the codebase to a recent commit on full-model (`2428726`). This PR should be squashed-merged into main to avoid an incompatible git history change.

This PR was created by the following steps:

- Using the [script](https://github.com/ai2cm/full-model/pull/1411) in`full-model` to specify which files should be copied over, and filter the git history to only include these files
- Pushing this filtered repo to a branch on `ace` called `dev`
- Creating the `squash-merge-release` branch on ace, and then merge `dev` into this branch using `git merge dev --allow-unrelated-histories --strategy-option theirs` (necessary because GH will not allow opening a PR with incompatible histories, but `git merge --allow-unrelated-histories` results in a branch that is "compatible"; also nice for making releases from `dev` to `main`.
- Re-apply squashed change of all commits made in `ace` since the last public release, making some additional modifications as needed (eg tweaking GH actions test workflow and Makefile rules)
- Re-apply deletion of files that had been deleted or moved since last public release, as these were not caught by the merge strategy.


oliverwm1 (commented):
There are some instances of `full-model` that should be changed to `ace` in `fme/docs/installation.rst` and `fme/docs/conf.py`. Otherwise this looks good to me.

oliverwm1 (approved):
Approving, because we can also just address the URLs in a subsequent PR.

oliverwm1 (approved):

PR #33, opened 2025-07-14
Change default env behavior

elynnwu:
Skip installing healpix since it may not work for all platforms. This is currently not a publicly supported option, so we should not install by default.


brianhenn:
For context, the reason for this is that `earth2grid` installation is finicky and appears to fail on Perlmutter and possibly other platforms, so we don't want the public repo's conda/docker builds to try to install it. That means that HealPIX training/inference presumably won't work in the public code, but that seems fine for now.

PR #708, opened 2026-01-05
Enable inference with a non-local `experiment_dir`

spencerkclark:
Reviving a PR started by @oliverwm1 in full-model (ai2cm/full-model#2227).

Various small changes to ensure that most ACE inference outputs can be written to object storage. We add unit tests for the `cloud` functions added, as well as an evaluator integration test, which writes the experiment directory to an `fsspec` in-memory filesystem and verifies the expected outputs exist.  We also test that the expected errors and warnings are raised when attempting to write netCDFs from the `RawDataWriter` or `MonthlyDataWriter` or logging to a file in a non-local experiment directory.

All but raw or time-coarsened netCDFs can be written to non-local storage.  An error will be raised if such netCDFs are requested.  The zarr writer can be used to store raw or uniformly-time-coarsened outputs to object storage, though time selection and monthly coarsening are still not supported there.

Changes:
- Adds several functions to a new `fme.core.cloud` module to facilitate adding this functionality.
- Adds a warning that logging will not be saved to a file if the experiment directory is non-local.
- Adds error messages to note the circumstances when a non-local `experiment_dir` is not supported, e.g. during training or writing raw or time-coarsened netCDFs.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/data_writer/raw.py
+++ b/fme/ace/inference/data_writer/raw.py
@@ -122,7 +122,11 @@ class RawDataWriter:
             dataset_metadata: Metadata for the dataset.
         """
         if not is_local(path):
-            raise ValueError("RawDataWriter only supports local file systems.")
+            raise ValueError(
+                "The RawDataWriter only supports local file systems. Consider "
+                "using the ZarrWriter instead, which supports writing to a "
+                "non-local filesystem."
+            )
         filename = str(Path(path) / f"{label}.nc")
         self._save_names = save_names
         self.variable_metadata = variable_metadata


.github/workflows/tests.yaml
spencerkclark, line 1: While the changes here work for this trial build, if we were to adopt something like this, I think we might be better off switching to manage the environments in CI with conda to ensure robustness to caching. This would also likely be faster than the trial approach, since we would no longer need to build `h5py` and `netcdf4` from source.

spencerkclark: Since this would increase the complexity of setting up an environment (i.e. it would require more than just `pip`) we opted to switch to an approach to writing remote files that does not require the ability to write xarray datasets to bytes.

fme/core/cloud.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def inter_filesystem_copy(source: str | Path, destination: str | Path):
+    """Copy between any two 'filesystems'. Do not use for large files.
+
+    Args:
+        source: Path to source file/object.
+        destination: Path to destination.
+    """
+    with fsspec.open(source) as f_source:
```
brianhenn, line 17: Is there a case when `source` will be remote?

fme/core/logging_utils.py
```diff
@@ -64,6 +66,14 @@ def configure_logging(self, experiment_dir: str, log_filename: str):
             logging.basicConfig(level=logging.ERROR)
         logger = logging.getLogger()
         if self.log_to_file and self._dist.is_root():
+            if not is_local(experiment_dir):
+                warnings.warn(
+                    f"Logging to a file is only supported if the experiment "
+                    f"directory is on a local file system. Got experiment "
+                    f"directory={experiment_dir!r}, so no logs will be saved "
```
brianhenn: (nit) Is there a reason for the `!r` since `experiment_dir` is just a string?

fme/core/cli.py
brianhenn, line 77: This method is also used by the ACE training entrypoint. What's the intended behavior if `TrainConfig.experiment_dir` is a remote path -- presumable not allowed since eg writing checkpoints to remote is not supported, but I'm not sure what's blocking that at the moment.

fme/ace/inference/data_writer/raw.py
```diff
@@ -120,6 +121,8 @@ def __init__(
             coords: Coordinate data to be written to the file.
             dataset_metadata: Metadata for the dataset.
         """
+        if not is_local(path):
+            raise ValueError("RawDataWriter only supports local file systems.")
```
brianhenn: So is the thought that 1) writing netCDFs to remote using the tools in this PR is OK only for small file sizes 2) the raw and monthly data writers have large file sizes so they are disallowed (in favor of the Zarr writer) and 3) the reduced diagnostics from training/evaluation are netCDFs but they are probably small so they're OK to write to remote? That seems reasonable, but it's a fairly complex set of rules.

(suggestion) Maybe the docstrings for `experiment_dir` on `InferenceEvaluatorConfig` and `InferenceConfig` could a) reference the fact that writing to remote is now supported and b) note some of the limitations above.

fme/core/cloud.py
spencerkclark: I do not think so, but I also do not think it hurts to use `fsspec` for the `source` and `destination` files for symmetry.

fme/core/logging_utils.py
spencerkclark: Using `!r` will format `experiment_dir` in quotes, which in my opinion makes it clearer what is part of the error message and what was the provided value.

fme/ace/inference/data_writer/raw.py
spencerkclark: Yes, that is correct.  Indeed updating those docstrings would be helpful—let me know if the updates I made in 5ff2ea6 look good.

fme/core/cli.py
spencerkclark: Correct, this PR does not address using a non-local `experiment_dir` in training.  I went ahead and added an informative error message and updated the docstring in dc0ffb6 to note that this is not supported for now.

oliverwm1:
fyi @janheuser we were talking about streaming ACE output directly to cloud object storage. This PR should enable that when using the zarr format data writer. Let me know if this meets your needs.

PR #709, opened 2026-01-05
Add IC ensemble ability to evaluator and update inference aggregators for ensembles

### LINKED ISSUE #1045: Evaluator entrypoint does not support `n_ensemble_per_ic`
This is a convenient feature for generating ensembles from the same initial conditions using a stochastic model, but it is currently only available through the inference entrypoint.  It would be nice if the evaluator entrypoint supported it as well—it would just mean that all ensemble members from the same initial condition would be compared to the same target.

Arcomano1234:
Adds support for the same initial condition (IC) expanded into multiple stochastic ensembles (`n_ensemble_per_ic` > 1) during evaluation and inline inference, and utilizes the existing 1-step ensemble aggregators for step-20 metrics during inference. NetCDF / paired writer sample axes are aligned with initial_condition_times one row per (IC × member). 

Note due to some small overhead in the evaluator tests from the added ensembling, a few of the evaluator tests were switched to be --very-fast skipped as they were getting 3s timeoutted.

Example on [wandb](https://wandb.ai/ai2cm/ace/runs/bckqa119). 

Changes:
- `n_ensemble_per_ic` property to `InlineInferenceConfig`
- Add support for single step ensemble metrics in `fme/ace/aggregator/inference/main.py` (mainly used for step 20 ensemble stuff)
- Add inline inference and evaluator support for `n_ensemble_per_ic`. These needed to touch a few places mostly gridded_data, the getters, and inference dataloader.
- Decouple as much of the broadcasting of the ensembles from the data_loading processes. E.g., it is done in `evaluator.py` and `train_config.py`

- [x] Tests added

Resolves #1045


fme/ace/aggregator/inference/main.py
```diff
@@ -180,9 +182,9 @@ def build(
             time_mean_reference_data=time_mean,
             record_step_20=record_step_20,
             channel_mean_names=channel_mean_names,
-            log_nino34_index=self.log_nino34_index,
```
mcgibbon, line 224: Was this removed intentionally?

fme/ace/aggregator/inference/main.py
```diff
@@ -415,6 +442,15 @@ def record_batch(
                 gen_data_norm=gen_data_norm,
                 i_time_start=self._n_timesteps_seen,
             )
+        if self.n_ensemble_per_ic > 1:
```
mcgibbon, line 497: Check out `fme/ace/aggregator/one_step/main.py` and see if you like the structure there for how it separates ensemble and deterministic aggregators. I think what's here is fine too, maybe preferable.

fme/ace/aggregator/inference/test_main.py
```diff
@@ -45,26 +47,47 @@ def test_inference_evaluator_aggregator_channel_mean_names(
...
         log_nino34_index=False,
         save_diagnostics=False,
+        n_ensemble_per_ic=n_ensemble,
     )
 
     target_data = {
-        "a": torch.ones([batch_size, n_timesteps, nx, ny], device=get_device()),
-        "b": torch.ones([batch_size, n_timesteps, nx, ny], device=get_device()) * 3,
-        "c": torch.ones([batch_size, n_timesteps, nx, ny], device=get_device()) * 4,
+        "a": torch.ones(
```
mcgibbon: Suggestion (optional): Like you do in `test_inference_evaluator_aggregator_ensemble`, make two BatchData, call `.broadcast_ensemble` on them, and then use `PairedData.from_batch_data` instead of constructing an ensemble PairedData at a low level. That would avoid coupling this test to the low-level implementation of the internals of BatchData/PairedData, making it easier to change the way ensembles are handled later if we need to.

fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -145,10 +169,14 @@ def _get_variable_metrics(self, gen_data: TensorMapping):
             self._variable_metrics = {
                 "crps": {},
                 "ssr_bias": {},
+                "ensemble_mean_rmse": {},
             }
             for key in gen_data:
                 self._variable_metrics["crps"][key] = CRPSMetric()
                 self._variable_metrics["ssr_bias"][key] = SSRBiasMetric()
+                self._variable_metrics["ensemble_mean_rmse"][key] = (
```
mcgibbon, line 177: Could separate "add ensemble_mean_rmse metric" into its own PR for clearer review.

fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -212,6 +240,8 @@ def get_logs(self, label: str):
         Args:
             label: Label to prepend to all log keys.
         """
+        temp_keys = [f"{label}/{key}" for key, _ in sorted(self._get_data().items())]
+        print("Logging ensemble aggregator keys:", temp_keys)
```
mcgibbon: Leftover print.

fme/ace/data_loading/batch_data.py
```diff
@@ -168,6 +168,7 @@ def to_device(self) -> "BatchData":
             horizontal_dims=self.horizontal_dims,
             epoch=self.epoch,
             labels=self.labels.to(device) if self.labels is not None else None,
+            n_ensemble=self.n_ensemble,
```
mcgibbon, line 172: Could separate these changes into their own PR, it looks like they're fixing an existing bug. Would make it easier to tell if you updated the tests for this (if you did great, if you didn't please do).

Maybe for some of these methods we should have a more general "test all attributes that don't start with underscore are identical, except for `.device.`" type test that will always catch these issues, even if we add new attributes.

fme/ace/data_loading/batch_data.py
```diff
@@ -430,11 +435,10 @@ def broadcast_ensemble(self: SelfType, n_ensemble: int) -> SelfType:
                 self.labels.names,
             )
         return self.__class__(
-            data={k: v.to(get_device()) for k, v in data.items()},
-            time=time,
-            horizontal_dims=self.horizontal_dims,
+            data=repeat_interleave_batch_dim(self.data, n_ensemble),
+            time=xr.concat([self.time] * n_ensemble, dim="sample"),
             labels=labels,
-            epoch=self.epoch,
```
mcgibbon, line 463: It looks like `epoch=self.epoch` got deleted. Can you remove the changes to this method, which seem to be stylistic? I normally try to construct things before the final `return`-setting-many-attributes, in the case of larger inits like this one.

fme/ace/data_loading/batch_data.py
```diff
@@ -468,6 +473,20 @@ def forcing(self) -> TensorMapping:
     def target(self) -> TensorMapping:
         return {k: v for k, v in self.reference.items() if k in self.prediction}
 
+    def broadcast_ensemble(self) -> tuple[EnsembleTensorDict, EnsembleTensorDict]:
```
mcgibbon: Issue: I would expect `.broadcast_ensemble` to return the same type, like it does for BatchData. I'd also expect it to be a light wrapper that broadcasts the internal BatchData. Is it possible to do that here? If you do, the test that got refactored to low-level-constrict PairedData could have its changes limited to calling the `.broadcast_ensemble` method on the one it was already making.

fme/ace/aggregator/inference/main.py
Arcomano1234: No that must have been some things moving around and I missed it in the merge

fme/ace/aggregator/inference/main.py
Arcomano1234: Either works but I did have a slight preference for the structure I proposed vs in `fme/ace/aggregator/one_step/main.py` to better allow more complex operations with ensemble aggs

fme/ace/aggregator/inference/main.py
```diff
@@ -76,12 +77,13 @@ def get_logs(self, label: str): ...
     def get_dataset(self) -> xr.Dataset: ...
 
 
-class _TimeDependentAggregator(Protocol):
```
Arcomano1234, line 82: Diff looks weird but I removed `_TimeDependentAggregator` as it is unused currently

fme/ace/aggregator/inference/main.py
mcgibbon: `fme/ace/aggregator/one_step/main.py` looks very clean.

fme/ace/data_loading/getters.py
```diff
@@ -176,20 +177,24 @@ def get_inference_data(
...
+    # Use n_ensemble=1 in the dataset so broadcast never runs in DataLoader workers.
+    # Ensemble expansion is done in the stepper (and for IC in get_initial_condition).
     dataset = InferenceDataset(
         config=config,
         total_forward_steps=total_forward_steps,
         requirements=window_requirements,
         surface_temperature_name=surface_temperature_name,
         ocean_fraction_name=ocean_fraction_name,
         label_override=label_override,
+        n_ensemble=1,
```
mcgibbon: Question: Should InferenceDataset take n_ensemble as an argument, or do we want to ensure Dataset never broadcasts? I think we don't allow n_ensemble on the normal training/validation dataset for example, and that it's better to do that here, but correct me if I'm wrong.

fme/ace/data_loading/gridded_data.py
```diff
@@ -140,9 +140,13 @@ def alternate_shuffle(self):
 def _get_initial_condition(
     loader: DataLoader[BatchData],
     requirements: PrognosticStateDataRequirements,
+    n_ensemble_per_ic: int = 1,
 ) -> PrognosticState:
     for batch in loader:
-        return batch.to_device().get_start(
+        batch = batch.to_device()
+        if n_ensemble_per_ic > 1:
+            batch = batch.broadcast_ensemble(n_ensemble_per_ic)
```
mcgibbon: Currently `.broadcast_ensemble` is never called in `fme/ace/data_loading`, instead it gets called in the stepper code. Is there a reason to call it during data loading time for inference, or should we broadcast in the inference level similar to how we treat it in training?

fme/ace/stepper/single_module.py
```diff
@@ -1214,6 +1217,10 @@ def predict_paired(
             final state, which can be used as a new initial condition.
         """
         forcing = self.forcing_deriver(forcing)
+        if forcing.n_ensemble == 1 and initial_condition.as_batch_data().n_ensemble > 1:
```
mcgibbon, line 1220: This line means that you don't really need the forcing data to be broadcast to an ensemble during data loading at all - the only thing that's needed is for the initial condition to get broadcast during the inference setup code.

Arcomano1234: Yes, that’s the intention. I think this keeps us mostly consistent with how we are handing the expansion of ICs and not having them expanded by the dataloaders. There might have been a slightly better place to do this but it also felt fine here.

mcgibbon: To be clear I think that's a good thing, what I mean is we don't need broadcasting in the data loading code (but this PR has broadcasting in the data loading code, and that unnecessary code has a lot of unnecessary tests).

fme/ace/data_loading/getters.py
```diff
@@ -176,6 +177,8 @@ def get_inference_data(
             set to None if no ocean temperature prescribing is being used.
         ocean_fraction_name: Name of the ocean fraction variable. Can be set to None
             if no ocean temperature prescribing is being used.
+        n_ensemble: Number of ensemble members per initial condition (passed to
+            ``InferenceGriddedData`` only; ``InferenceDataset`` does not broadcast).
```
mcgibbon: Does `InferenceGriddedData` need to see the number of initial conditions or broadcast the data?

The config lives on `n_ensemble_per_ic` on `InlineInferenceConfig`. That means the ideal case is to isolate this responsibility to the code that corresponds to InlineInferenceConfig (the top level of inline inference). At that level, you can get the IC from the InferenceGriddedData (without a broadcast), and perform the broadcast there. This avoids passing n_ensemble_per_ic to the data loading, and avoids the associated tests - you only need a test at the InlineInference level where the config is added (the integration test should be sufficient) along with the existing unit tests that BatchData has a broadcasting method.

If for some reason we can't do that, this way is fine, but it requires significantly more coupling.

mcgibbon: The alternative is if `n_ensemble_per_ic` is only needed in the data loader, it could be owned by the data loader config, and used in the data loader to broadcast the IC. This doesn't work well if we use `n_ensemble_per_ic` for other things like initializing the aggregators (I imagine we do? but maybe not).

We could maybe push all this off to another PR, to get this in?

fme/ace/aggregator/inference/main.py
```diff
@@ -437,10 +455,21 @@ def __init__(
...
+        if self.n_ensemble_per_ic > 1:
+            summary_aggregators_list = (
+                list(self._aggregators.items())
+                + list(self._time_dependent_aggregators.items())
+                + list(self._ensemble_aggregators.items())
+            )
+        else:
+            summary_aggregators_list = list(self._aggregators.items()) + list(
+                self._time_dependent_aggregators.items()
+            )
```
mcgibbon: nit: a bit clearer
```suggestion
        summary_aggregators_list = list(self._aggregators.items()) + list(
            self._time_dependent_aggregators.items()
        )
        if self.n_ensemble_per_ic > 1:
            summary_aggregators_list.extend(
                self._ensemble_aggregators.items()
            )
```

mcgibbon (approved):
LGTM, one nit remaining.

PR #710, opened 2026-01-05
Add training configs for ACE2S used in HiRO - ACE manuscript

Arcomano1234:
Add the configurations for training ACE2S on X-SHiELD. Includes the pre-training on 1-step predictions of ERA5 and the multi-step fine-tuning on X-SHiELD.


configs/experiments/2025-11-15-ace2s-x-shield/pretrain-1-step-era5.yaml
```diff
@@ -0,0 +1,259 @@
...
+  entity: ai2cm
+train_loader:
+  batch_size: 8
+  num_data_workers: 8
+  prefetch_factor: 4
+  dataset:
+    strict: false
+    concat:
+      # Cut around the dates of ERA5 when parallel streams were merged
+      # For details see Section 3 of H Hersbach, 2020 (https://rmets.onlinelibrary.wiley.com/doi/10.1002/qj.3803)
```
mcgibbon, line 45: Suggestion (optional): May want to reference `https://github.com/ai2cm/explore2/blob/main/jeremym/era5/plot_skipped_years.py` even though it's private. Or we could pull it into this repo (doesn't have to be in this PR).

Arcomano1234: Yeah I was thinking about pulling that into this repo, but maybe under `scripts/` eventually and then reference that. If you have a better sense of where `plot_skipped_years.py` should live in `ace` I can add it in another PR.

PR #713, opened 2026-01-05
Add `StaticInputs` class

AnnaKwa:
This PR adds a container class `StaticInputs` which holds a list of `Topography` objects and has similar methods as `Topography` that will perform the same operations but on all of its fields (e.g. subsetting in lat/lon). This has no effect on the rest of downscaling code. Later changes will update the references to `Topography` where it is used in the downscaling code to iterate through the `StaticInputs.fields` and concat them along the input channel dimension.

The idea is to rename `Topography` -> `StaticInput` in a later change since this class will be used for more than just topography data. I left the name unchanged in this PR to minimize diffs in other files.
 
- [x] Tests added

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -8,21 +8,6 @@ on:
     branches:
       - main
 
-  # Add a button to manually trigger the workflow
-  workflow_dispatch:
-
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
-# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
-# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
-concurrency:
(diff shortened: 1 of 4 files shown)


fme/downscaling/data/topography.py
```diff
@@ -137,3 +138,61 @@ def get_topography_downscale_factor(
...
+                    f"All StaticInput fields must have the same coordinates. "
+                    f"Fields {i} and 0 do not match coordinates."
+                )
+
+    def __getitem__(self, index: int):
+        return self.fields[index]
+
+    @property
+    def input_tensors(self) -> list[torch.Tensor]:
+        if len(self.fields) > 0:
```
frodre, line 160: Based on the follow-on integration for this code, is it better to allow for empty `StaticInputs` with internal checks or just not initialize a static input object when none are specified?

AnnaKwa: I think keeping internal checks would allow for cleanup in a few places when there are repeated checks for whether the topography is None.

PR #720, opened 2026-01-06
Registry docs addition

brianhenn:
Adds a test that ensures that registered config classes for `ace` are imported into the namespace for docs purposes. Adds some missing symbols covered by the test.


fme/ace/test_symbols.py
```diff
@@ -69,3 +74,29 @@ def test_inference_config_nested_dataclass_symbols_are_imported():
...
+    "PreBuiltBuilder",
+    "MockModuleBuilder",
+    "MockStepConfig",
+]
+
+
+@pytest.mark.parametrize(
+    "config_registry_class",
+    [
+        CorrectorSelector,
```
brianhenn, line 89: My thought is that these registries are tightly coupled to `ace` and therefore it's appropriate to test them here even if they actually reside in `core`.

fme/ace/test_symbols.py
```diff
@@ -69,3 +74,29 @@ def test_inference_config_nested_dataclass_symbols_are_imported():
     finally:
         CHECKED_SYMBOLS.clear()
         MISSING_SYMBOLS.clear()
+
+
+WHITELISTED_SYMBOLS = [  # used only for testing, not in __all__
```
brianhenn, line 79: Alternatively, could avoid permanently registering these classes and thus having to whitelist by using context managers in the test modules.

fme/ace/test_symbols.py
```diff
@@ -69,3 +74,29 @@ def test_inference_config_nested_dataclass_symbols_are_imported():
...
+        CorrectorSelector,
+        ModuleSelector,
+        StepSelector,
+        PerturbationSelector,
+    ],
+)
+def test_registry_symbols_are_imported(config_registry_class):
+    registered_config_names = config_registry_class.get_available_types()
+    for registered_config_name in registered_config_names:
+        registered_config = config_registry_class.registry._types[
```
brianhenn, line 98: Relying on private `registry._types` here; could make public since we already make their name strings public.

PR #721, opened 2026-01-06
Add hiro ckpt train config

AnnaKwa:


frodre (approved):
LGTM other than the configuration test that's not passing.

Issue #717, opened 2026-01-05
Beaker CI test using out of date code


brianhenn:
The Beaker CI test (ACE checkpoint evaluator test) [checks out the prior repo](https://github.com/ai2cm/ace/blob/4c73d5b33ed56deb1e23a4198e74a058d93d0a57/scripts/ci_beaker/entrypoint.sh#L17-L24) instead of running with code in this repo. It seems like this test should be switched to using gantry for submission if we want to maintain it.

PR #723, opened 2026-01-06
Beaker CI test via gantry

brianhenn:
Updates the Beaker ACE evaluator CI workflow to use gantry for submission. This simplifies the test considerably since the code no longer needs to be checked out manually. 

See example of the passing test (manually triggered) here: https://github.com/ai2cm/ace/actions/runs/20773428895/job/59653944997

Resolves https://github.com/ai2cm/ace/issues/717.

### DIFF
--- a/fme/ace/data_loading/batch_data.py
+++ b/fme/ace/data_loading/batch_data.py
@@ -1,4 +1,5 @@
 import dataclasses
+import warnings
 from collections.abc import Callable, Collection, Iterable, Sequence
 from typing import Any, TypeVar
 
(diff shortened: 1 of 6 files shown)


.github/workflows/beaker.yaml
```diff
@@ -12,74 +12,69 @@ on:
...
+        run: |
+          gantry run \
+            --show-logs \
+            --name ${{ env.JOB_NAME }} \
+            --task-name ${{ env.JOB_NAME }} \
+            --description 'ACE evaluator test on CI' \
+            --beaker-image "$(cat latest_deps_only_image.txt)" \
+            --workspace ${{ env.BEAKER_WORKSPACE }} \
+            --priority low \
+            --preemptible \
```
elynnwu, line 54: Should we remove the preemptible flag? Since this is a separate workspace with 2 non-preemt slots. 
I think it's fine to keep it as it is since it usually runs very fast and is unlikely to get preempted, we could also just wait for beaker's minimum runtime feature.

oliverwm1: The preemptible flag does mean it is more likely to get scheduled (can open door to more clusters) so I think we should keep it, especially since chance of preemption is small given how quickly the job runs.

PR #724, opened 2026-01-07
Prevent backpropagation anomalies in energy corrector

spencerkclark:
### Issue

Our current energy corrector code is vulnerable to introduce anomalies in backpropagation through computing the `log` of the pressure at model layer interfaces in the `compute_layer_thickness` function.  Here by "anomalies" we mean places where undefined values (e.g. `-inf`, `inf`, `NaN`) come up within the chain of computation.  We have observed these anomalies to arise in two circumstances:

- Early in training with all datasets, the surface pressure at some grid cells can be less than zero.  `log` of a value less than zero in PyTorch results in `NaN`.  Previously to work around this we would skip computing and applying the energy for an entire training batch if any such grid cells existed (i.e. prior to any `NaN` appearing in computation).  In https://github.com/ai2cm/full-model/pull/1807, however, this was changed to permit computing the energy correction for each sample.  Any sample that contained a grid cell with negative surface pressure would result in a `NaN` temperature correction, which would then be replaced with zero.
- The ERA5 dataset has a top layer interface pressure of `0.0`.  This is present in all grid columns in all samples.  `log` of `0.0` results in `-inf`.  We immediately clamp this to `0.0`, effectively treating the model top has having a pressure of `1.0` Pa in this computation.

While the handling of the invalid values is appropriate in inference for both of these cases, it is not appropriate during training.  A `NaN` or `-inf` value appearing in the chain of computation leads to divergence when computing gradients in backpropagation, and ultimately `NaN` weights, crashing the training process.  This can be detected by running training in the `torch.autograd.detect_anomaly()` context, which provides a traceback for where invalid values first appear during backpropagation.  

### Solution

A solution that addresses both of these issues is to switch the order of the `clamp` and `log` within the `compute_layer_thickness` function.  Rather than compute the `log` and then `clamp`, we can first `clamp` the pressure at layer interfaces to a minimum value of `1.0` and then compute the `log`.  This prevents `-inf` or `NaN` values from ever appearing in the computation, therefore enabling safe backpropagation.

One minor consequence is that it now means that rather than skip energy correction for the entire batch (as we did originally) or crash training (as is the case currently), we now permit applying energy corrections—as physically dubious as they may be—in circumstances where some grid cells have a surface pressure of less than zero.  In those grid columns, for the purposes of computing the geopotential during the energy calculation, their layer interface pressure is clamped to a minimum value of `1.0`.  My expectation is that this should not materially affect results, since it only pertains to severely undertrained models (surface pressures less than zero are far outside the range we expect).

### Tests

I have added a specific test that we can backpropagate through `compute_layer_thickness` when the pressure at interface includes a value of zero.  I also modified the `test__force_conserve_total_energy` test to take into account that we now apply corrections even with negative surface pressures, and added a check that we can backpropagate through the result.  Prior to the one-line change in `compute_layer_thickness` these tests would fail with the following backpropagation-anomaly-related errors:

<details>
<p>

```
$ pytest -k "test__force_conserve_total_energy[True] or test_backpropagate_through_compute_layer_thickness"
================================================================================================================== test session starts ==================================================================================================================
platform darwin -- Python 3.11.14, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/spencerc/software/ace
configfile: pyproject.toml
plugins: anyio-4.12.1, zarr-3.1.5, cov-7.0.0
collected 772 items / 770 deselected / 2 selected

corrector/test_atmosphere.py F                                                                                                                                                                                                                    [ 50%]
test_atmosphere_data.py F                                                                                                                                                                                                                         [100%]

======================================================================================================================= FAILURES ========================================================================================================================
________________________________________________________________________________________________________ test__force_conserve_total_energy[True] ________________________________________________________________________________________________________

negative_pressure = True

    @pytest.mark.parametrize("negative_pressure", [True, False])
    def test__force_conserve_total_energy(negative_pressure: bool):
        tensor_shape = (5, 5)

        ops = LatLonOperations(
            0.5 + torch.rand(size=(tensor_shape[-2], 1)).broadcast_to(size=tensor_shape)
        )
        timestep = datetime.timedelta(seconds=3600)
        test_input = _get_corrector_test_input(tensor_shape, requires_grad=True)
        input_data, gen_data, forcing_data, vertical_coord = test_input
        extra_heating = 10.0

        if negative_pressure:
            input_data["PRESsfc"] = -1 * input_data["PRESsfc"]

        correction_expected = not negative_pressure

        corrected_gen_data = _force_conserve_total_energy(
            input_data=input_data,
            gen_data=gen_data,
            forcing_data=forcing_data,
            area_weighted_mean=ops.area_weighted_mean,
            vertical_coordinate=vertical_coord,
            timestep_seconds=timestep.total_seconds(),
            unaccounted_heating=extra_heating,
        )

        # ensure only temperature is modified
        for name in gen_data:
            if "air_temperature" in name:
                if correction_expected:
                    assert not torch.allclose(
                        corrected_gen_data[name], gen_data[name], rtol=1e-6
                    )
                else:
                    assert torch.allclose(
                        corrected_gen_data[name], gen_data[name], rtol=1e-6
                    )
            else:
                torch.testing.assert_close(corrected_gen_data[name], gen_data[name])

        # ensure forcing variables are not in the corrected data
        for name in forcing_data:
            assert name not in corrected_gen_data

        # ensure the corrected global mean MSE path is what we expect
        input = AtmosphereData(input_data, vertical_coord)
        corrected_gen = AtmosphereData(corrected_gen_data | forcing_data, vertical_coord)
        input_gm_mse = ops.area_weighted_mean(input.total_energy_ace2_path)
        corrected_gen_gm_mse = ops.area_weighted_mean(corrected_gen.total_energy_ace2_path)
        predicted_mse_tendency = (
            ops.area_weighted_mean(corrected_gen.net_energy_flux_into_atmosphere)
            + extra_heating
        )

        if correction_expected:
            expected_gm_mse = (
                input_gm_mse + predicted_mse_tendency * timestep.total_seconds()
            )
            torch.testing.assert_close(corrected_gen_gm_mse, expected_gm_mse)

        # ensure the temperature correction is constant
        corrected_gen_temperature = corrected_gen_data["air_temperature_1"]
        initial_gen_temperature = gen_data["air_temperature_1"]
        temperature_1_correction = corrected_gen_temperature - initial_gen_temperature
        assert torch.all(torch.eq(temperature_1_correction, temperature_1_correction[0, 0]))
        temperature_correction_0 = (
            corrected_gen_data["air_temperature_0"] - gen_data["air_temperature_0"]
        )
        torch.testing.assert_close(temperature_correction_0, temperature_1_correction)

        # ensure we can backpropagate through the result without anomalies
        with torch.autograd.detect_anomaly():
            # Take the sum to reduce corrected_gen_data to a scalar prior to
            # backpropagation to avoid the need to provide a gradient argument
            # to backward.
>           sum(tensor.sum() for tensor in corrected_gen_data.values()).backward()

corrector/test_atmosphere.py:411:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/_tensor.py:648: in backward
    torch.autograd.backward(
../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/autograd/__init__.py:353: in backward
    _engine_run_backward(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

t_outputs = (tensor(2616313., grad_fn=<AddBackward0>),), args = ((tensor(1.),), False, False, ()), kwargs = {'accumulate_grad': True, 'allow_unreachable': True}, attach_logging_hooks = False

    def _engine_run_backward(
        t_outputs: Sequence[Union[torch.Tensor, GradientEdge]],
        *args: Any,
        **kwargs: Any,
    ) -> tuple[torch.Tensor, ...]:
        attach_logging_hooks = log.getEffectiveLevel() <= logging.DEBUG
        if attach_logging_hooks:
            unregister_hooks = _register_logging_hooks_on_whole_graph(t_outputs)
        try:
>           return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
                t_outputs, *args, **kwargs
            )  # Calls into the C++ engine to run the backward pass
E           RuntimeError: Function 'DivBackward0' returned nan values in its 1th output.

../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/autograd/graph.py:824: RuntimeError
__________________________________________________________________________________________________ test_backpropagate_through_compute_layer_thickness ___________________________________________________________________________________________________

    def test_backpropagate_through_compute_layer_thickness():
        # Regression test to ensure we can properly handle a TOA interface pressure
        # of 0.0, as is present in the ERA5 dataset. The key for this test to be
        # meaningful is the presence of zeros in the pressure_at_interface tensor.
        pressure_at_interface = torch.tensor(
            [
                [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]],
                [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]],
            ],
            requires_grad=True,
        )
        air_temperature = torch.tensor(
            [
                [[300.0, 310.0], [300.0, 310.0]],
                [[300.0, 310.0], [300.0, 310.0]],
            ],
            requires_grad=True,
        )
        specific_total_water = torch.full((2, 2, 2), 0.1, requires_grad=True)
        dz = compute_layer_thickness(
            pressure_at_interface, air_temperature, specific_total_water
        )
        with torch.autograd.detect_anomaly():
            # Take the sum to reduce dz to a scalar prior to backpropagation to
            # avoid the need to provide a gradient argument to backward.
>           dz.sum().backward()

test_atmosphere_data.py:58:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/_tensor.py:648: in backward
    torch.autograd.backward(
../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/autograd/__init__.py:353: in backward
    _engine_run_backward(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

t_outputs = (tensor(22998.9180, grad_fn=<SumBackward0>),), args = ((tensor(1.),), False, False, ()), kwargs = {'accumulate_grad': True, 'allow_unreachable': True}, attach_logging_hooks = False

    def _engine_run_backward(
        t_outputs: Sequence[Union[torch.Tensor, GradientEdge]],
        *args: Any,
        **kwargs: Any,
    ) -> tuple[torch.Tensor, ...]:
        attach_logging_hooks = log.getEffectiveLevel() <= logging.DEBUG
        if attach_logging_hooks:
            unregister_hooks = _register_logging_hooks_on_whole_graph(t_outputs)
        try:
>           return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
                t_outputs, *args, **kwargs
            )  # Calls into the C++ engine to run the backward pass
E           RuntimeError: Function 'LogBackward0' returned nan values in its 0th output.

../../../../miniconda3/envs/2026-01-07-ace/lib/python3.11/site-packages/torch/autograd/graph.py:824: RuntimeError
```

</p>
</details>

### Changes

- Switches the order of the `clamp` and `log` calls on the interface pressure in `compute_layer_thickness` to ensure well-posed backpropagation even in the presence of interface pressures less than or equal to `0.0`.  
- Removes the `nan_to_num` call when applying the temperature correction in `_force_conserve_total_energy`, since it is no longer necessary.
- Applies an energy correction even when the surface pressure is less than zero.  For reasons described above I do not think this is a material behavior change.

- [x] Tests added


spencerkclark:
I think I agree with @mcgibbon's comment that the approach suggested by ChatGPT for computing the mass-weighted integral of the geopotential is superior to what is used currently in ACE for reasons he notes.  It also happens to be what is used in hydrostatic FV3 when computing the column-integrated total energy ([code](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/1ab589c5d0d49c7ea245abea8893bd13c937617c/model/fv_thermodynamics.F90#L90-L225)).

Switching to this approach would require rederiving the formula for computing the uniform temperature correction, which depends in part on how we are computing the mass-weighted integral of the geopotential ([Overleaf](https://www.overleaf.com/project/671c26c7be5005f658c74da7)). I think this would just be a matter of some algebra. I can look into that if we want.

No matter which approach we use, however, I think we would still need the `compute_layer_thickness` function for computing the geopotential height at the layer interfaces, and within this function we would need to guard against backpropagation anomalies due to unusual pressures.  For that reason we could consider merging this and then look at improving the overall approach in a later PR.

Issue #678, opened 2025-10-02
Pressure-level ML-derived variables


brianhenn:
Create pressure-level diagnostics via an ML step (rather than [interpolation](https://github.com/ai2cm/ace/issues/671) from ML prognostic variables). The step likely should be separate from generating the existing diagnostic variables to avoid needing to fully retrain SFNO and diluting its loss function. e.g., add the pressure-level diagnostics as additional decoder outputs.

Ongoing progress tracked via wandb [training](https://wandb.ai/ai2cm/ace?nw=gymtvcqqit)/[evaluation](https://wandb.ai/ai2cm/ace?nw=shuekl8q7gc) and [slides](https://docs.google.com/presentation/d/1luMEOky94LEY8Rgzty3w3xcmh6JkiKw7ybFRsgXZp8w/edit?slide=id.g38e92b73599_0_0#slide=id.g38e92b73599_0_0).

PR #726, opened 2026-01-07
Add separated Conv1x1 decoder

mcgibbon:
Adds the option of a configurable secondary/separated Conv1x1/MLP decoder to the `SingleModuleStep`, i.e., that predicts additional variables from the primary outputs but without influencing the prediction of the primary outputs.  

The name contradiction of having two modules in this step may be resolved in the future.

See production test job [here](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KFF0TS0Q94XD0Q22V35KBXSJ?taskId=01KFF0TS8EYZGVP6QFTWBFW2NB&jobId=01KFF0TSC2M29Y6JQEZ5D1ZMP3); also wandb view [here](https://wandb.ai/ai2cm/ace/panel/bodkvcnr0?nw=09zvcwnwe5bu) comparing with prior runs on this branch

Changes:
- Adds SingleModuleStepConfig.secondary_decoder: SecondaryDecoderConfig | None = None, where SecondaryDecoderConfig takes in a list of names of secondary diagnostic variables and a module configuration (currently "MLP" only). 

- [X] Tests added

Resolves https://github.com/ai2cm/ace/issues/678

### DIFF
--- a/fme/core/step/secondary_decoder.py
+++ b/fme/core/step/secondary_decoder.py
@@ -23,7 +23,8 @@ class SecondaryDecoderConfig:
 
     Parameters:
         secondary_diagnostic_names: Names of additional diagnostic variables, to be
-            diagnosed directly from outputs without access to latent variables.
+            diagnosed directly from outputs without access to latent variables (i.e.,
+            column-locally).
         network: Configuration for the decoder network.
     """
 


fme/core/step/secondary_decoder.py
```diff
@@ -0,0 +1,169 @@
...
+    layers: list[nn.Module] = []
+    if depth == 1:
+        layers.append(nn.Conv2d(in_dim, out_dim, kernel_size=1))
+    else:
+        layers.append(nn.Conv2d(in_dim, n_hidden, kernel_size=1))
+        layers.append(nn.GELU())
+        for _ in range(depth - 2):
+            layers.append(nn.Conv2d(n_hidden, n_hidden, kernel_size=1))
+            layers.append(nn.GELU())
+        layers.append(nn.Conv2d(n_hidden, out_dim, kernel_size=1))
```
mcgibbon: Double-check: this output layer has a bias vector, is that the same as what we did before? IMO it's probably better this way but we should avoid changing behavior from what you completed.

fme/core/step/secondary_decoder.py
```diff
@@ -0,0 +1,169 @@
...
+@dataclasses.dataclass
+class SecondaryDecoderConfig:
+    """
+    Configuration for the secondary decoder that computes additional diagnostic
+    variables from the main module's outputs.
+
+    Parameters:
+        secondary_diagnostic_names: Names of additional diagnostic variables, to be
+            diagnosed directly from outputs without access to latent variables.
+        network: Configuration for the decoder network (uses ModuleSelector registry).
```
mcgibbon:
```suggestion
        network: Configuration for the decoder network.
```
nit: No need to say what type is used in the description, types are automatically picked up by sphinx. The type hint won't go out-of-date if we ever change this, unlike a parameter docstring.

fme/core/step/secondary_decoder.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward pass through the decoder."""
+        return self._module(x)
+
+    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+        """Make the decoder callable."""
+        return self.forward(x)
+
+    def unpack(self, tensor: torch.Tensor, axis: int = 0) -> TensorDict:
```
mcgibbon: Issue: The responsibility for naming channels is split between two code levels, which is harder to maintain.

Suggestion: If the decoder will own the packer, we should probably have `__call__` return a TensorDict so the packer (and responsibilities around it) can remain private. If not, the packer should probably be owned by the class doing the unpacking, so that the secondary decoder can avoid knowing anything about output names.

It does seem like the only thing this class is responsible for is unpacking, so if we keep this class we should probably have it do the unpacking within `__call__`. Otherwise I would suggest to keep a `secondary_decoder_packer` in the StepABC implementation, and do away with this class entirely.

fme/core/step/single_module.py
```diff
@@ -222,15 +242,32 @@ def __init__(
...
+
+        if config.secondary_decoder is not None:
+            secondary_diagnostic_names = (
+                config.secondary_decoder.secondary_diagnostic_names
+            )
+            self.secondary_decoder: SecondaryDecoder | None = SecondaryDecoder(
+                in_dim=n_out_channels,
+                out_names=secondary_diagnostic_names,
+                network=config.secondary_decoder.network,
+            ).to(get_device())
```
mcgibbon: Suggestion: If this class remains, use a SecondaryDecoderConfig.build(dist.wrap_module) function to contain this. It could even return a NoSecondaryDecoder implementation that always returns an empty dict, instead of None, if you want to simplify the code below.

fme/core/step/single_module.py
```diff
@@ -300,7 +344,18 @@ def network_call(input_norm: TensorDict) -> TensorDict:
                 input_tensor,
                 labels=args.labels,
             )
-            return self.out_packer.unpack(output_tensor, axis=self.CHANNEL_DIM)
+            if self.secondary_decoder is not None:
+                secondary_output_tensor = self.secondary_decoder.wrap_module(wrapper)(
+                    output_tensor.detach()
```
mcgibbon: nit: A comment is worth including here because the `.detach()` might be unexpected for some readers - this would make it clear it's intentional.

```suggestion
                    output_tensor.detach()  # detach avoids changing base outputs
```

fme/core/step/single_module.py
```diff
@@ -338,6 +395,13 @@ def load_state(self, state: dict[str, Any]) -> None:
             # for backwards compatibility with old checkpoints
             del module["module.device_buffer"]
         self.module.load_state(module)
+        if self.secondary_decoder is not None:
```
mcgibbon: Nuanced Issue: This will currently break if we try to take an existing model without a secondary decoder, and "fine-tune" a new model to have one (which as I understand is one of the main ways we plan to use this feature).

However, this actually does _not_ break, because we don't use `load_state` for this, we use ParameterInitializer.apply_weights. load_state is only used to load a compatibly-configured state. For example, we already can't load_state between two modules with different weight shapes or different numbers or names of weights, as we do for fine-tuning.

In other words, the warning for `elif "secondary_decoder" in state` is a bit of an issue - it suggests we expect to run into this situation (which we don't) and that if it does happen it's OK to ignore it (which it probably isn't). In the unlikely situation where this gets triggered, there's probably a bug. I would suggest if you think this might happen and want to protect against it, you should upgrade it to an exception, and if you don't think it will happen, drop the warning so the code is clearer.

fme/core/step/test_secondary_decoder.py
```diff
@@ -0,0 +1,181 @@
...
+    MLPConfig,
+    SecondaryDecoder,
+    SecondaryDecoderConfig,
+    _build_mlp,
+)
+
+
+class TestBuildMLP:
+    def test_depth_1_creates_single_layer(self):
+        mlp = _build_mlp(in_dim=4, out_dim=2, n_hidden=8, depth=1)
```
mcgibbon: ~~nit: we should test the public methods (i.e. MLPConfig.build) instead of the private helper (`_build_mlp`), even though in this case they're identical.~~

Ah I guess what I mean here is, TestBuildMLP is not necessary because you have TestMLPConfig. I would suggest nit: deleting it to reduce our maintenance burden, and to move the missing cases onto TestMLPConfig.

fme/core/step/secondary_decoder.py
```diff
@@ -0,0 +1,169 @@
...
+from fme.core.dataset_info import DatasetInfo
+from fme.core.packer import Packer
+from fme.core.registry import ModuleSelector
+from fme.core.registry.module import Module, ModuleConfig
+from fme.core.typing_ import TensorDict
+
+
+@ModuleSelector.register("MLP")
+@dataclasses.dataclass
+class MLPConfig(ModuleConfig):
```
mcgibbon: nit: Put this in `fme/core/models/` where the conditional SFNO code also lives, for easier discoverability and organization of our available ModuleSelector options.

fme/core/step/secondary_decoder.py
brianhenn: I think this is the same. But I will also launch a job with the updates to to confirm that the behavior looks unchanged.

fme/core/step/test_secondary_decoder.py
brianhenn: Sounds good

fme/core/step/secondary_decoder.py
brianhenn: Opted to fold the packing into the `__call__`.

fme/core/step/single_module.py
```diff
@@ -276,7 +311,10 @@ def modules(self) -> nn.ModuleList:
         Returns:
             A list of modules being trained.
         """
-        return nn.ModuleList([self.module.torch_module])
+        modules = [self.module.torch_module]
+        if self.secondary_decoder.torch_module is not None:
+            modules.append(self.secondary_decoder.torch_module)
```
mcgibbon: Suggestion (optional): If you instead exposed a `list[nn.Module]` from `secondary_decoder.modules`, you could use `modules.extend` without the none-check.

fme/core/step/single_module.py
```diff
@@ -324,6 +367,7 @@ def get_state(self):
         state = {
             "module": self.module.get_state(),
         }
+        state["secondary_decoder"] = self.secondary_decoder.get_module_state()
```
mcgibbon: nit: set "secondary_decoder" the same way you set "module"

PR #727, opened 2026-01-07
Serialize static inputs with downscaling model

AnnaKwa:
- `DiffusionModel` can be initialized with optional static_inputs: StaticInputs arg, which will save this data with the model checkpoints. DiffusionModelConfig.build can also take this as optional arg when building the model object. CheckpointModelConfig also will build the model with this data if it is in the checkpoint.
- Training will save the static inputs with the checkpoints

- [x] Tests added

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -98,8 +98,6 @@ class DiffusionModelConfig:
         churn: The amount of stochasticity during generation.
         num_diffusion_generation_steps: Number of diffusion generation steps
         use_fine_topography: Whether to use fine topography in the model.
-        static_inputs: Optional mapping of variable names to dataset paths
-            for static input variables.
     """
 
     module: DiffusionModuleRegistrySelector
@@ -115,7 +113,6 @@ class DiffusionModelConfig:
     num_diffusion_generation_steps: int
     predict_residual: bool
     use_fine_topography: bool = False
-    static_inputs: dict[str, str] = dataclasses.field(default_factory=dict)
 
     def __post_init__(self):
         self._interpolate_input = self.module.expects_interpolated_input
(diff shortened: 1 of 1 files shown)


fme/downscaling/models.py
```diff
@@ -97,6 +98,8 @@ class DiffusionModelConfig:
         churn: The amount of stochasticity during generation.
         num_diffusion_generation_steps: Number of diffusion generation steps
         use_fine_topography: Whether to use fine topography in the model.
+        static_inputs: Optional mapping of variable names to dataset paths
```
frodre: Add a note that these input fields will be stored with model checkpoint.

fme/downscaling/models.py
```diff
@@ -273,6 +281,8 @@ def __init__(
                 coarse to fine.
             sigma_data: The standard deviation of the data, used for diffusion
                 model preconditioning.
+            static_inputs: Optional static inputs to the model that may be loaded
```
frodre, line 281: Add a note that for now if the model requires these fields and they are not present within the model, then they are expected to be loaded from the dataset configuration

fme/downscaling/train.py
```diff
@@ -444,9 +447,19 @@ def build(self) -> Trainer:
             )
         else:
             model_coarse_shape = train_data.coarse_shape
+
+        # load full spatial range of topography to save with model
+        # TODO: this will be replaced in the future with a more general call
+        # to get normalized static inputs from a model config field
+        full_topography = get_normalized_topography(
```
frodre, line 454: Currently not using the `DiffusionModelConfig` pathway to preserve backwards compatibility of configurations we used for the paper, right?

fme/downscaling/models.py
frodre: We also should have some information about when to use this config option vs. the build entrypoint.  Should have an error if both are specified at least.

AnnaKwa: Oops, this was not meant to be in the config attributes, for this PR it should just be a build arg.

fme/downscaling/train.py
AnnaKwa: Yes, this line will probably change in the future when the more general ability to use >1 static input is added, which will require changes in the allowed config structures.

PR #728, opened 2026-01-07
Pass model static inputs to dataset build calls at generation

AnnaKwa:
- data configs' build methods can take `static_inputs_from_checkpoint` as an optional arg when building the GriddedData/PairedGriddedData objects. (If using the full static input fields from the model, they have to be provided at data config build time since they must be subset to the same region as the datasets.) This implementation results in the smallest number of changes to existing code, but at the expense of adding some entanglement between the model and data classes. In the future we could disentangle this dependency between the data loader and model by enabling the built GriddedData objects to take in full static input fields and subset them to the same coordinate range as data.

- downscaling generation entrypoints (predict/inference/evaluator run scripts) will pass the model's static input (if present) when building the GriddedData objects


frodre (commented):
Thanks.  Overall implementation LGTM.  A couple things just for coverage:
* if you can add a test for a model that includes static inputs that would be useful coverage-wise, potentially into one of the integrations if that's easier. 
* Adding a TODO comment of why things are structured as they are with static inputs along with a link to the PRs would be useful for anyone perusing this intermediate code since it's currently a sticky web of interactions

PR #731, opened 2026-01-13
Ensure topography is on device in downscaling inference

AnnaKwa:
Fixes a bug where the static inputs loaded from serialized checkpoint data were not on device in the inference entrypoint.

Added serialized static input data to the model in the downscaling inference integration test. The fixture creating the data config is modified to allow the `topography` config field to be None, to ensure that there is no fine data being loaded in the integration test and the generation must use the model's serialized static inputs.

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -1176,14 +1176,15 @@ class Stepper(
             def checkpoint(module):
                 return optimizer.checkpoint(module, step=step)
 
-            state = self.step(
-                StepArgs(
-                    input=input_data,
-                    next_step_input_data=next_step_input_dict,
-                    labels=labels,
-                ),
-                wrapper=checkpoint,
-            )
+            with optimizer.autocast():
+                state = self.step(
+                    StepArgs(
+                        input=input_data,
+                        next_step_input_data=next_step_input_dict,
+                        labels=labels,
+                    ),
+                    wrapper=checkpoint,
+                )
             yield state
             state = optimizer.detach_if_using_gradient_accumulation(state)
 


fme/downscaling/test_train.py
```diff
@@ -205,7 +205,10 @@ def test_train_main_logs(default_trainer_config, tmp_path, very_fast_only: bool)
         assert len(keys) > 5 or keys == set(["train/batch_loss"])
 
 
-def test_restore_checkpoint(default_trainer_config, tmp_path):
+def test_restore_checkpoint(default_trainer_config, tmp_path, very_fast_only: bool):
```
AnnaKwa, line 208: I don't think this was related to the changes in this PR, but this test failed the 'very fast only' tests.

fme/downscaling/inference/test_output.py
```diff
@@ -69,31 +68,6 @@ def test_region_config_requires_time_range():
 # Integration test fixtures and helpers
 
 
-@pytest.fixture
```
AnnaKwa, line 72: This is moved to conftest.py

fme/downscaling/test_train.py
Arcomano1234: I assume its because there is more IO needed to read in the topography right?

AnnaKwa: Yeah, for some reason it did not fail in the previous PRs that changed the training to save the topography in the checkpoint.

fme/downscaling/inference/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+from fme.downscaling.test_utils import data_paths_helper
+
+
+@pytest.fixture
+def loader_config(tmp_path, request):
+    """Create DataLoaderConfig with test data."""
+    path = tmp_path / "test_data"
+    path.mkdir()
+    add_topography_path = request.param
+    # TODO: should probably consolidate cross imported
```
Arcomano1234, line 14: Optional: I know Jeremy favors removing TODOs and instead making issues. But I also see there other TODOs in the downscaling so it seems thats the system used for the downscaling code.

Arcomano1234 (approved):

Arcomano1234:
LGTM!

PR #734, opened 2026-01-13
Coupled stepper config removes deprecated `crps_training` key

### LINKED ISSUE #732: Old checkpoint not running after deprecated `crps_training`
Getting this [error](https://beaker.allen.ai/orgs/ai2/workspaces/ace/work/01KETZK244DPY95NR5T5CPT10G/logs?jobId=01KETZK2FK7MBDS8M91BFSKN39) in coupled evaluation after `crps_training` is removed when using an older checkpoint:
```
dacite.exceptions.UnexpectedDataError: can not match "crps_training" to any data class field
```

Probably need to update coupled stepper.

elynnwu:
In #2378, we remove deprecated `crps_training` config value, but that breaks coupled evaluation/inference because it has its own checkpoint loading. This PR fixes the bug using a type hook for `StepperConfig`.

Resolves #732

### DIFF
--- a/fme/downscaling/inference/conftest.py
+++ b/fme/downscaling/inference/conftest.py
@@ -1,0 +1,34 @@
+import pytest
+
+from fme.core.dataset.xarray import XarrayDataConfig
+from fme.downscaling.data import DataLoaderConfig
+from fme.downscaling.test_utils import data_paths_helper
+
+
+@pytest.fixture
+def loader_config(tmp_path, request):
+    """Create DataLoaderConfig with test data."""
+    path = tmp_path / "test_data"
+    path.mkdir()
+    add_topography_path = request.param
+    # TODO: should probably consolidate cross imported
+    # .      data path helpers to a single file instead
+    # .      of importing from test_train in each location
+    test_data_path = data_paths_helper(path)
+    topography_path = (
+        f"{test_data_path.fine}/data.nc" if add_topography_path is True else None
+    )
+
+    return DataLoaderConfig(
+        coarse=[
+            XarrayDataConfig(
+                data_path=str(test_data_path.coarse),
+                file_pattern="*.nc",
+                engine="netcdf4",
+            )
+        ],
+        batch_size=2,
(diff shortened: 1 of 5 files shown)


fme/coupled/stepper.py
```diff
@@ -630,7 +630,11 @@ def get_state(self):
     def from_state(cls, state) -> "CoupledStepperConfig":
         state = cls.remove_deprecated_keys(state)
         return dacite.from_dict(
-            data_class=cls, data=state, config=dacite.Config(strict=True)
+            data_class=cls,
+            data=state,
+            config=dacite.Config(
+                strict=True, type_hooks={StepperConfig: StepperConfig.from_dict}
```
mcgibbon: That is a really neat feature! I wish I knew about this when handling other deprecated keys.

jpdunc23:
The CoupledStepper model artifact was created using https://github.com/ai2cm/full-model/commit/e55355381

fme/ace/stepper/single_module.py
```diff
@@ -736,6 +736,13 @@ def remove_deprecated_keys(cls, state: dict[str, Any]) -> dict[str, Any]:
             del state_copy["crps_training"]  # training value, ok to break
         return state_copy
 
+    @classmethod
+    def from_dict(cls, state) -> "StepperConfig":
+        state = cls.remove_deprecated_keys(state)
+        return dacite.from_dict(
+            data_class=cls, data=state, config=dacite.Config(strict=True)
+        )
+
```
jpdunc23: It's a bit confusing to have both `StepperConfig.from_state` and `StepperConfig.from_dict` which are nearly identical. Maybe we can just add the line `state = cls.remove_deprecated_keys(state)` to `StepperConfig.from_state` and use it?

elynnwu: Good catch - I didn't notice `from_state` below. Updated

jpdunc23 (approved):
LGTM!

PR #735, opened 2026-01-13
Samudra bugfix: Use circular padding for longitude axis

elynnwu:
In Samudra, there's a bug in how padding is done, causing longitude=0 to have strange artifacts. This PR fixes this by separating the latitude/longitude circular padding.

The current behavior (ignore the actual bias, this is just from a random checkpoint eval to illustrate the artifacts):
image (3)
After padding fix:
image (4)


elynnwu:
Attaching experiments done by James to verify no obvious performance change before/after this bug fix.

> WandB workspace comparing validation / inline inference skill of Samudra with this bugfix to an older baseline: https://wandb.ai/ai2cm/ace-samudra-cm4?nw=wdn5g7js91

PR #737, opened 2026-01-13
Add additional diagnostics of the OHC budget

jpdunc23:
Adds three new ocean derived variables: `net_energy_flux_into_ocean_column`, `ocean_heat_content_tendency`, and `implied_tendency_of_ocean_heat_content_due_to_advection`. 

- [x] Tests added

### DIFF
--- a/fme/ace/stepper/test_single_module.py
+++ b/fme/ace/stepper/test_single_module.py
@@ -1892,8 +1892,8 @@ def _get_ocean_data_for_predict_paired(
         )
 
     # Set forcing variables to known values
-    if "hfds_prognostic" in data_dict:
-        data_dict["hfds_prognostic"] = torch.full(
+    if "hfds_forcing" in data_dict:
+        data_dict["hfds_forcing"] = torch.full(
             (n_samples, total_timesteps, *img_shape), 99.0, device=DEVICE
         ) + torch.arange(total_timesteps, device=DEVICE).view(
             total_timesteps, 1, 1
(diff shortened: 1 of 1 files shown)


jpdunc23:
Changes initially reviewed by @spencerkclark at https://github.com/ai2cm/full-model/pull/2376

fme/ace/stepper/test_single_module.py
```diff
@@ -1765,3 +1767,282 @@ def test_replace_derived_forcings_error():
...
+        )
+
+    if hfds_role == "next_step_forcing":
+        # hfds is a forcing, so net_energy_flux_into_ocean_column in prediction
+        # and reference are the same
+        torch.testing.assert_close(pred_flux, ref_flux)
+    elif hfds_role == "diagnostic":
+        # hfds is diagnostic, so the generated net_energy_flux_into_ocean
+        # differs from the reference
+        assert not torch.allclose(pred_flux, ref_flux)
```
jpdunc23, line 2048: @spencerkclark hopefully this is a bit less obtuse now. The actual value of `pred_flux` doesn't really matter so I refactored to just check that when `hfds` is a diagnostic the predicted `hfds` is used and the net flux into the ocean differs from the reference.

spencerkclark: Thanks, sorry my comment was a bit vague.  I think the main confusing aspect to me was that `"hfds_prognostic"` is listed as an input, but not an output—a comment clarifying its role (or maybe renaming the variable to better reflect its purpose) could possibly still be helpful.

jpdunc23: Agreed that's not a good name. I switched to `"hfds_forcing"`, hopefully that will be more clear now.

PR #738, opened 2026-01-14
Add LoRA support to conditional SFNO

mcgibbon:

Changes:
- Added options `lora_rank`, `lora_alpha`, `spectral_lora_rank`, and `spectral_lora_alpha` controlling the rank and strength of LoRA associated with Conv1x1 operations and with the spectral filter.

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/s2convolutions.py
+++ b/fme/core/models/conditional_sfno/s2convolutions.py
@@ -211,7 +211,7 @@ class SpectralConvS2(nn.Module):
         B, C, H, W = x.shape
 
         with torch.amp.autocast("cuda", enabled=False):
-            x = self.forward_transform(x)
+            x = self.forward_transform(x.float())
             if self._round_trip_residual:
                 x = x.contiguous()
                 residual = self.inverse_transform(x)
--- a/fme/ace/inference/data_writer/main.py
+++ b/fme/ace/inference/data_writer/main.py
@@ -1,8 +1,8 @@
 import dataclasses
 import datetime
+import os
 import warnings
 from collections.abc import Mapping, Sequence
-from pathlib import Path
 from typing import TypeAlias
 
 import numpy as np
@@ -10,6 +10,7 @@ import torch
 import xarray as xr
 
 from fme.ace.data_loading.batch_data import BatchData, PairedData, PrognosticState
+from fme.core.cloud import to_netcdf_via_inter_filesystem_copy
 from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.generics.writer import WriterABC
 
@@ -303,7 +304,7 @@ def _write(
     data_arrays["time"] = time_array
     ds = xr.Dataset(data_arrays, coords=coords)
     ds.attrs.update(dataset_metadata.as_flat_str_dict())
-    ds.to_netcdf(str(Path(path) / filename))
+    to_netcdf_via_inter_filesystem_copy(ds, os.path.join(path, filename))
 
 
 class DataWriter(WriterABC[PrognosticState, PairedData]):
--- a/fme/ace/inference/data_writer/monthly.py
+++ b/fme/ace/inference/data_writer/monthly.py
@@ -15,6 +15,7 @@ from fme.ace.inference.data_writer.utils import (
     DIM_INFO_LATLON,
     get_all_names,
 )
+from fme.core.cloud import is_local
 from fme.core.dataset.data_typing import VariableMetadata
 
 LEAD_TIME_DIM = "time"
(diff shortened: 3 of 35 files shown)


fme/core/models/conditional_sfno/lora.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import annotations
+
+import math
+
+import torch
+import torch.nn as nn
+
+
+class LoRAConv2d(nn.Conv2d):
```
mcgibbon, line 9: I don't like that this uses inheritance, but without this it's very difficult to have previously saved checkpoints load properly, or I'd need to copy-paste some lora code every single time we call a Conv2D (which I was doing at first, and it's a lot).

mcgibbon:
This code change does not affect performance of the non-LoRA training.

image

Machine noise is significant, but my baseline runs show if anything better performance on the experimental branch using LoRA. Because I've done nothing to improve performance on that branch, I expect this is noise.

Baseline run forked from main: https://wandb.ai/ai2cm/ace/runs/act5d5bu/overview?nw=ohyfh3fa3lf
Baseline run on experimental branch with LoRA code: https://wandb.ai/ai2cm/ace/runs/udrkj93s/overview?nw=ohyfh3fa3lf

mcgibbon:
Note the LoRA training runs (a few configurations added here) are also within this performance noise envelope:
image

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -257,23 +280,32 @@ def __init__(
...
         if use_mlp == True:
-            MLPH = MLP
             mlp_hidden_dim = int(embed_dim * mlp_ratio)
-            self.mlp = MLPH(
+            self.mlp = MLP(
                 in_features=embed_dim,
                 hidden_features=mlp_hidden_dim,
                 act_layer=act_layer,
                 drop_rate=drop_rate,
-                checkpointing=checkpointing,
```
elynnwu, line 267: MLP still accepts `checkpointing`, is it removed because we don't use it?  Otherwise we should restore it.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -67,9 +78,16 @@ def __init__(
         bias=False,
         use_tensorly=True,
         filter_residual: bool = False,
+        lora_rank: int = 0,
+        lora_alpha: float | None = None,
     ):  # pragma: no cover
         super(SpectralConvS2, self).__init__()
 
+        if in_channels != out_channels:
```
elynnwu, line 86: Should this be checked only when LoRA is enabled?

mcgibbon: This is a hidden limitation of the existing code, not specific to LoRA. Basically it was never used without this config, and a bug was introduced (before us) that broke this case.

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Good catch, a previous iteration of this PR required this to be removed but it doesn't anymore. Added it back.

We also don't use it, I'd perhaps suggest we remove it. But let's not look into that right now.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Specifically it's the part in `forward` where `xp = torch.zeros_like(x)` xp is the output tensor, and is initialized to look like the input tensor. This only works if the in and out channels are the same.

I don't think we need that part, it's for some kind of "padding" feature that was already removed/commented out code when we copied this over. But we also don't need in_channels != out_channels, so I disabled that instead.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -195,6 +203,10 @@ def __init__(
         filter_residual=False,
         affine_norms=False,
         filter_num_groups: int = 1,
+        lora_rank: int = 0,
```
Arcomano1234, line 206: There should be some documentation somewhere, even if its just so I remember why `lora_rank` vs `spectral_lora_rank` is important (e.g., `spectral_lora_rank` probably should be greater than `lora_rank`)

mcgibbon: The documentation of this and all other arguments is on the Builder class, since that's what gets exposed publicly. I'd rather not keep two copies up to date.

```
        lora_rank: Rank of the LoRA adaptations outside of spectral convolutions.
            0 (default) disables LoRA.
        lora_alpha: Strength of the LoRA adaptations outside of spectral convolutions.
            Defaults to lora_rank.
        spectral_lora_rank: Rank of the LoRA adaptations for spectral convolutions.
            0 (default) disables LoRA.
        spectral_lora_alpha: Strength of the LoRA adaptations for spectral convolutions.
            Defaults to spectral_lora_rank.
```

PR #739, opened 2026-01-15
List frozen parameters with include or exclude list

mcgibbon:
This PR modifies the way frozen parameters are specified, to use either an include or an exclude list, but not both. This relaxes a previous requirement that every parameter in the network be matched by either an include or exclude rule.

Changes:
- Frozen parameters are now specified with an include or exclude list, but not both. If a rule in the list is not used, an exception will be raised. If both are specified, a warning will be emitted when configuration is initialized and an exception will be raised if apply is called.
- Frozen parameters are now logged.

- [x] Tests added


mcgibbon:
This manner of configuring is working well for me in the LoRA fine-tuning, for example [this config](https://github.com/ai2cm/ace/blob/experiment/2026-01-15-stochastic-lora/configs/experiments/2025-09-23-stochastic/train-x-shield-n384-e1c9-1step-ft-20step-v2-lora-c256s128-v1-rs0.yaml#L112).

fme/ace/stepper/parameter_init.py
brianhenn: For whatever reason it's always been tricky for me to remember what a "parameter name" is, i.e., that this is using a single `nn.Module`'s state dict keys to identify what to freeze, and that a `FrozenParameterConfig` must be specified separately for each module it the step's `.modules` (more relevant now that we're adding multi-module steps). 

If that's correct, maybe helpful to state here what "parameter names" is referring to.

brianhenn: > All parameters must be specified
in either the include or exclude list, or
    an exception will be raised.

Is this still true?

brianhenn, line 89: (chore): docstring is out of date for `ParameterClassification`.

(nit): I've also been tripped up by the `exclude` argument appearing both on `ParameterClassification` and on the nearby `ParameterClassification.frozen`. Not sure it's worth another breaking rename but could try to give these user-configurable arguments more expressive names in the future (eg `ParameterClassification.excluded_weights`, especially if we end up supporting fine-tuning as a public feature.

fme/ace/stepper/parameter_init.py
```diff
@@ -30,32 +32,37 @@ class FrozenParameterConfig:
     a rule in one of the lists is not matched by any parameters in the model
     (including if it is already matched by an earlier rule).
 
+    By default no parameters are frozen.
```
brianhenn, line 33: > By default no parameters are frozen.

(suggestion) How is this default established? By having `include` and and `exclude` both be empty lists by default? Or by having `exclude` be `["*"]` by default (which is currently specified not on this object but on the `ParameterClassification.frozen` default value one level up)?

It's unintuitive to me what this PR's default behavior of `exclude = []` is supposed to mean when it's described as "if given then all other parameters are frozen" -- to me that would indicate that the default behavior is to freeze everything. 

I would suggest moving the default of `exclude=["*"]` to this object instead.

fme/core/wildcard.py
```diff
@@ -99,3 +100,80 @@ def apply_by_wildcard(
             f"include rules: {remaining_includes}, exclude rules: {remaining_excludes}"
         )
     return model
+
+
+def apply_by_include(
```
brianhenn, line 41: I think separating the `include` and `exclude` applications is an improvement over `apply_by_wildcard` in terms of readability. 

(suggestion, non-blocking) Could apply this same approach to `CopyWeightsConfig.apply`, such that  `apply_by_wildcard` can be deleted and we can avoid creating two different behaviors in terms of what `exclude` and `include` mean for fine-tuning weight freezing.

fme/ace/stepper/parameter_init.py
brianhenn, line 21: Also, cursor agent PR review suggests the following, which I think I agree with:

```
Documentation update: The docstring for FrozenParameterConfig was updated but could be clearer about the backwards compatibility behavior.

Consider documenting that both can be specified for backwards compatibility during inference, but applying freezing will raise an error.
```

mcgibbon: I agree, @jpdunc23 was suggesting we might use a nn.ModuleDict instead of nn.ModuleList so we can name the modules.

I updated ParameterInitializationConfig to make it clearer that one ParameterClassification is given for each nn.Module.

mcgibbon: Updated the config format and the docstring.

mcgibbon: Updated.

fme/ace/stepper/parameter_init.py
mcgibbon: Updated the config format to make this clearer. The default is now on this object as well.

fme/ace/stepper/parameter_init.py
mcgibbon: We should refactor this with breaking changes once we split the stepper config into train and inference parts, and ignore the training ones during inference loading. That will make it so we can do this change without breaking inference backwards compatibility.

mcgibbon: Reverted the config format change because it breaks inference. I did update the docstring though.

mcgibbon: Updated to say these are the state_dict names.

fme/ace/stepper/parameter_init.py
mcgibbon: I reverted the config change.

I thought about this for a while, and while I agree it's not obvious what `include=[], exclude=[]` should do, I think it's important so that when this is used, we can configure e.g. `include=[a, b, c]` without having to remember to manually override `exclude=[]`. To solve both of these, I updated `exclude` to default to None, so the default is more clearly `include=[]`. I'll also remove the exclude=["*"] default like you suggested.

fme/core/wildcard.py
mcgibbon: Good idea, I'll look at doing this in a follow on PR.

fme/ace/stepper/parameter_init.py
brianhenn: Great, I think this is much clearer.

PR #740, opened 2026-01-15
Enforce lat bounds (-88 deg, 88 deg) 

AnnaKwa:
Though the bin edges of the coarse and 32x fine grids in our HiRO X-SHiELD datasets align over the entire globe, the fine grid midpoints aren't evenly split on either side of the coarse midpoints (i.e. there are < 32 fine grid midpoints between the coarse grid midpoints) in the last couple coarse grid latitude bins at the poles. The [code](https://github.com/ai2cm/full-model/blob/027e316b7e273ab343e9c7817bb7431ee8125657/fme/downscaling/data/utils.py#L99) that calculates the subset of fine grid latitudes to select assumes there are 16 fine grid midpoints on either side of the coarse lat range midpoints, so if the model is configured to train on or generate regions at the poles (|lat midpoints|>88.24) this will cause a crash.

This PR updates the default `lat_extent` to be (-66, 70) deg if not specified, which corresponds with the training range. The config raises an error if explicitly set to extend outside of those bounds.

 
- [x] Tests added


fme/downscaling/inference/output.py
```diff
@@ -417,7 +422,7 @@ class TimeRangeConfig(DownscalingOutputConfig):
         default_factory=lambda: Slice(-1, 1)
     )
     lat_extent: ClosedInterval = field(
-        default_factory=lambda: ClosedInterval(-90.0, 90.0)
+        default_factory=lambda: ClosedInterval(-88, 88.0)
```
Arcomano1234: I think this was mentioned by @oliverwm1 but did we want to make the default -65 to 65 because for the foreseeable future we will only have model weights trained on this range?

PR #741, opened 2026-01-15
Allow diagnostic `hfds` and unaccounted heating in OHC correction

jpdunc23:
Allow the ocean heat content correction to use diagnostic `hfds` (the net downward heat flux into the ocean) and a configured constant global mean unaccounted heating term.

Changes:
- Deprecates boolean configuration of `ocean_heat_content_correction` in favor of `OceanHeatContentBudgetConfig`, keeping backward compatibility.
- Makes the OHC correction a bit more flexible by with optional geothermal heat flux and using `1 - land_fraction` when `sea_surface_fraction` is not provided.
- Clean up: Remove the deprecated `masking` attribute from `OceanCorrectorConfig`.

- [x] Tests added

### DIFF
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -57,7 +57,7 @@ class OceanHeatContentBudgetConfig:
     Parameters:
         method: Method to use for OHC budget correction. The available option is
             "scaled_temperature", which enforces conservation of heat content
-            by scaling the predicted potential temperature by a vertically and 
+            by scaling the predicted potential temperature by a vertically and
             horizontally uniform correction factor.
         constant_unaccounted_heating: Area-weighted global mean
             column-integrated heating in W/m**2 to be added to the energy flux


fme/core/corrector/ocean.py
```diff
@@ -116,17 +133,17 @@ def __call__(
...
+                self._config.ocean_heat_content_correction.constant_unaccounted_heating,
             )
-        if self._masking is not None:
-            # NOTE: masking should be applied last to avoid overwriting
-            gen_data = self._masking(gen_data)
         return dict(gen_data)
 
 
-class AreaWeightedSum(Protocol):
+class AreaWeightedMean(Protocol):
```
jpdunc23, line 147: This was changed to the `AreaWeightedMean` so that `constant_unaccounted_heating` can be set based on `inference/mean/weighted_mean_gen/implied_tendency_of_ocean_heat_content_due_to_advection`.

fme/core/corrector/ocean.py
```diff
@@ -135,41 +152,61 @@ def __call__(
...
-    ) / (global_gen_ocean_heat_content)
-
+    ) / global_gen_ocean_heat_content
     # apply same temperature correction to all vertical layers
     n_levels = gen.sea_water_potential_temperature.shape[-1]
     for k in range(n_levels):
         name = f"thetao_{k}"
-        gen.data[name] = gen.data[name] * torch.nan_to_num(
-            heat_content_correction_ratio, nan=1.0
-        )
```
jpdunc23, line 173: Unlike in the (older version of the) atmosphere total energy correction, `heat_content_correction_ratio` shouldn't be NaN during normal operation; if it is then there is some other issue at play (e.g., faulty data) so I think it's preferable to allow the NaN through rather than mask it and risk letting the issue go unnoticed.

spencerkclark: Agreed—as I found in #724, NaNs are problematic for backpropagation to begin with, even if they are later filtered out.

fme/core/corrector/ocean.py
```diff
@@ -50,30 +49,57 @@ def __call__(
         return out
 
 
+@dataclasses.dataclass
+class OHCBudgetConfig:
```
spencerkclark: nit: I slightly prefer spelling this out, since it is not extraordinarily long
```suggestion
class OceanHeatContentBudgetConfig:
```

fme/core/corrector/ocean.py
```diff
@@ -135,41 +152,61 @@ def __call__(
...
+    forcing_data: TensorMapping,
+    area_weighted_mean: AreaWeightedMean,
     vertical_coordinate: HasOceanDepthIntegral,
     timestep_seconds: float,
+    method: Literal["constant_temperature"] = "constant_temperature",
+    unaccounted_heating: float = 0.0,
 ) -> TensorDict:
+    if method != "constant_temperature":
+        raise NotImplementedError(
+            f"Method {method} not implemented for ocean heat content conservation"
```
spencerkclark: nit: minor suggestion to put the `method` in quotes in the error message:
```suggestion
            f"Method {method!r} not implemented for ocean heat content conservation"
```

fme/core/corrector/ocean.py
```diff
@@ -135,41 +152,61 @@ def __call__(
...
-    expected_change_ocean_heat_content = area_weighted_sum(
-        input.net_energy_flux_into_ocean * timestep_seconds,
+    try:
+        net_energy_flux_into_ocean = (
+            gen.net_downward_surface_heat_flux + forcing.geothermal_heat_flux
+        ) * forcing.sea_surface_fraction
+    except KeyError:
+        net_energy_flux_into_ocean = (
+            input.net_downward_surface_heat_flux + forcing.geothermal_heat_flux
+        ) * forcing.sea_surface_fraction
```
spencerkclark, line 196: nit: I guess there should never be a circumstance where `net_downward_surface_heat_flux` would be both a forcing and a diagnostic output (conceptually it would not make sense).  Nevertheless it might be safer to raise an error if it is ambiguous where to pull it from rather than default to pulling the generated version.

fme/core/corrector/ocean.py
```diff
@@ -135,41 +152,61 @@ def __call__(
 def _force_conserve_ocean_heat_content(
     input_data: TensorMapping,
     gen_data: TensorMapping,
-    area_weighted_sum: AreaWeightedSum,
+    forcing_data: TensorMapping,
+    area_weighted_mean: AreaWeightedMean,
     vertical_coordinate: HasOceanDepthIntegral,
     timestep_seconds: float,
+    method: Literal["constant_temperature"] = "constant_temperature",
```
spencerkclark: Strictly speaking, this is not a constant temperature correction like the atmosphere energy corrector uses, right?  Would a better name be `"scaled_temperature"` or something like that?

fme/core/corrector/ocean.py
```diff
@@ -135,41 +152,61 @@ def __call__(
...
     for k in range(n_levels):
         name = f"thetao_{k}"
-        gen.data[name] = gen.data[name] * torch.nan_to_num(
-            heat_content_correction_ratio, nan=1.0
-        )
-
+        gen.data[name] = gen.data[name] * heat_content_correction_ratio
+    if "sst" in gen.data:
+        gen.data["sst"] = (  # assuming sst in Kelvin
+            gen.data["sst"] - 273.15
```
spencerkclark: It could be useful to store `273.15` as `FREEZING_TEMPERATURE_KELVIN` in `constants.py`.

fme/core/corrector/ocean.py
jpdunc23: Ah, good catch. I was thinking of "constant" in the same sense as `constant_unaccounted_heating`.

fme/core/corrector/ocean.py
```diff
@@ -50,30 +50,57 @@ def __call__(
...
 
+@dataclasses.dataclass
+class OceanHeatContentBudgetConfig:
+    """Configuration for ocean heat content budget correction.
+
+    Parameters:
+        method: Method to use for OHC budget correction. The available option is
+            "scaled_temperature", which enforces conservation of heat content
+            by imposing a vertically and horizontally uniform potential
+            temperature correction.
```
spencerkclark: Minor suggested revision to be more precise about how this `method` operates:
```suggestion
        method: Method to use for OHC budget correction. The available option is
            "scaled_temperature", which enforces conservation of heat content
            by scaling the predicted potential temperature by a vertically and 
            horizontally uniform correction factor.
```

fme/core/corrector/ocean.py
spencerkclark: Optional if you still want to address this.

jpdunc23: This is already addressed with: https://github.com/ai2cm/ace/blob/8a76773033c51efa1ae437bec304fd6f91a7b1a3/fme/core/corrector/ocean.py#L167-L171

I'm cheating a bit with knowledge that `net_downward_surface_heat_flux` and `"hfds"` are the same, but I felt it would be a bit awkward to check the `OceanData` objects since currently the only way to do so AFAIK is to query each one in a try-except block. 

I figured this cheating is OK since our correctors are already cheating in this way (e.g., by knowing that `sea_water_potential_temperature` and `"thetao"` are the same variable) and we're moving away from supporting multiple multiple naming conventions.

spencerkclark: Ah I see, sorry for missing that in your latest commit.  Yes, I agree this approach is reasonable.  Thanks for clarifying!

PR #742, opened 2026-01-15
Fix bug causing step sampler to be ignored

mcgibbon:
A bug in our code was causing `train_n_forward_steps` on the StepperConfig to be ignored. This PR fixes that bug.

Changes:
- Fixed a bug causing StepperConfig.train_n_forward_steps to be ignored

- [x] Tests added

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -3,7 +3,6 @@ import dataclasses
 import datetime
 import logging
 import pathlib
-import warnings
 from collections.abc import Callable, Generator, Mapping
 from typing import Any, Literal, cast
 
@@ -515,11 +514,11 @@ class StepperConfig:
     parameter_init: ParameterInitializationConfig = dataclasses.field(
         default_factory=lambda: ParameterInitializationConfig()
     )
-    input_masking: StaticMaskingConfig | None = None
     train_n_forward_steps: TimeLength | TimeLengthSchedule | None = None
+    input_masking: StaticMaskingConfig | None = None  # inference
     derived_forcings: DerivedForcingsConfig = dataclasses.field(
         default_factory=lambda: DerivedForcingsConfig()
-    )
+    )  # inference
 
     @property
     def train_n_forward_steps_schedule(self) -> TimeLengthSchedule | None:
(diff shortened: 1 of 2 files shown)


fme/ace/stepper/test_single_module.py
```diff
@@ -144,6 +144,127 @@ def get_scalar_data(names, value):
     return {n: float(value) for n in names}
 
 
+def test_stepper_no_train_step_specified():
```
jpdunc23, line 149: Consider adding a test of the case where `train_n_forward_steps` is an `int`.

fme/ace/stepper/single_module.py
```diff
@@ -912,11 +908,12 @@ def get_loss_obj() -> StepLoss:
...
-        if epoch is None and self._train_n_forward_steps_schedule is not None:
-            raise EpochNotProvidedError(
-                "current configuration requires epoch to be provided "
-                "on BatchData during training"
+        if epoch is None:
+            warnings.warn(
+                "epoch is not provided in BatchData during call to train_on_batch;"
+                " will assume epoch 0, and some features may not work as expected."
             )
+            epoch = 0
```
jpdunc23: Could you just do:

```python
    if (
        epoch is None 
        and self._train_n_forward_steps_schedule is not None
        and len(self._train_n_forward_steps_schedule.milestones) > 0  # Just add this
    ):
        raise EpochNotProvidedError(...)
```

If so, then we can keep `test_train_on_batch_requires_epoch`. Otherwise, maybe we should also remove the `EpochNotProvidedError` since it is no longer used.

jpdunc23 (approved):
Looks good, but I left a couple of comments. Approved so feel free to merge when ready, but happy to take another pass if you decide to make changes.

fme/ace/stepper/single_module.py
mcgibbon: Updated.

fme/ace/stepper/test_single_module.py
mcgibbon: Added.

mcgibbon (commented):
Implemented your suggestions, looks like it works! Will merge once tests pass.

PR #743, opened 2026-01-15
Add docs page for downscaling inference

AnnaKwa:
Adds a docs page for downscaing inference entrypoint.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,48 @@
+
+# Contributing
+
+Thank you for your interest in contributing to this project! This is a research codebase
+for the [Ai2 Climate Modeling](https://allenai.org/climate-modeling) group. While we welcome
+external users, please note that we do not guarantee backwards compatibility or supporting
+features that don't align with our research priorities.
+
+## Reporting Issues
+
+We welcome bug reports and feature requests. Please open an issue to:
+- Report bugs you encounter
+- Suggest new features or improvements
+
+## Submitting Changes
+
+### Bug Fixes
+
+We welcome pull requests that fix bugs. Please include a clear description of the
+issue and how your fix addresses it. Ideally, your pull request should include a
+unit test that would fail without your bug fix.
+
+### New Features
+
+If you're an external user and want to implement a new feature, **please open an
+issue first to discuss it with us**. This helps ensure the feature aligns with
+the project's direction and prevents wasted effort.
+
+## Code Review
+
+Automated CI tests must pass before your pull request will be reviewed. All
+pull requests will be reviewed for:
+- Correctness and functionality
+- Code style, type-hinting and consistency
+- Alignment with project goals
+
+Please be responsive to feedback during the review process.
+
+## Developing
+
+You can install a developer environment using the `conda` package manager with
+`make create_environment`.
+
+Install pre-commit hooks to ensure linting and type-hinting requirements are met before
+committing code: `pre-commit install`.
+
+You can run local tests with `pytest`. As a shortcut for running only faster unit
+tests, use `make test_very_fast` from the root of the repository.
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -514,11 +514,11 @@ class StepperConfig:
     parameter_init: ParameterInitializationConfig = dataclasses.field(
         default_factory=lambda: ParameterInitializationConfig()
     )
-    input_masking: StaticMaskingConfig | None = None
     train_n_forward_steps: TimeLength | TimeLengthSchedule | None = None
+    input_masking: StaticMaskingConfig | None = None  # inference
     derived_forcings: DerivedForcingsConfig = dataclasses.field(
         default_factory=lambda: DerivedForcingsConfig()
-    )
+    )  # inference
 
     @property
     def train_n_forward_steps_schedule(self) -> TimeLengthSchedule | None:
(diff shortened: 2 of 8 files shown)


fme/downscaling/inference/inference.py
```diff
@@ -154,7 +154,7 @@ class InferenceConfig:
     sizes, and output variables. Outputs are processed sequentially, with generation
     parallelized across GPUs using distributed data loading.
 
-    Attributes:
+    Parameters:
```
AnnaKwa, line 157: These changes were made so that the autodocs classes render cleanly in the docs. Otherwise the docs has names with no descriptions in a parameters section, and then list the same names again with the descriptions in an attributes section.

fme/downscaling/inference/inference.py
```diff
@@ -166,8 +166,8 @@ class InferenceConfig:
         logging: Logging configuration.
         patch: Default patch prediction configuration.
 
-    Example YAML configuration:
-    ```yaml
+    Example YAML configuration::
+
```
AnnaKwa, line 171: Changed so the docstring renders in the docs

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+
+
+Example YAML Configuration
+--------------------------
+
+The following example shows a configuration which generates two outputs: one using ``EventConfig`` for a single time snapshot, and one using ``TimeRangeConfig`` for a time range.
+
+.. code-block:: yaml
+
+   experiment_dir: /results
```
frodre: This would be better stored as an actual yaml file that we can then test against similar to the `evaluator_config.rst`

frodre: We should match this to how it will look in the model release folder as well.

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+
+:class:`fme.downscaling.inference.output.EventConfig` is used for generating a single time snapshot over a spatial region. This is useful for capturing specific events like hurricane landfall, extreme weather events, or any single-timestep high-resolution snapshot of a region.
+
+.. autoclass:: fme.downscaling.inference.output.EventConfig
+   :show-inheritance:
+   :noindex:
+
+Example EventConfig:
+
+.. code-block:: yaml
```
frodre, line 92: Again, should have a testable config example so we know if this goes out of date.

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+
+   - name: "hurricane_landfall_2023"
```
frodre: While it's correct for the top-level config, I don't think these should be under a list item.

frodre: Same for the following config sections

frodre: Also, there might be a way to specify the doctest in line without needing the separated yaml file for these smaller config segment examples.

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+   - name: "first_year_indices"
+     n_ens: 4
+     max_samples_per_gpu: 4
+     time_range:
+         start: 0
+         stop: 36
+
+Example TimeRangeConfig with RepeatedInterval:
+
+.. code-block:: yaml
```
frodre, line 155: Does this link over to the RepeatedInterval docs?  This one is more complicated so a link would be nice so people could click and see the explanations for what this is actually doing.

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+Use the ``save_vars`` field to specify which variables to save to the output zarr file. If ``save_vars`` is ``null`` or not specified, all variables from the model output will be saved:
+
+.. code-block:: yaml
+
+   save_vars: ["PRATEsfc"]
+
+Patch Prediction for Large Domains
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For domains larger than the model's patch size, subdivision of the full domain into patches for prediction can be enabled. Configure this in the top-level ``patch`` section or override per-output:
```
frodre: While we could easily make it possible, we don't currently allow per-output overrides for patch.  It get's specified by the top-level inference config and there's a pathway in the output build, but it's currently hard coded to use the top-level patch all the time

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,281 @@
...
+
+.. code-block:: yaml
+
+   patch:
+       divide_generation: true
+       composite_prediction: true
+       coarse_horizontal_overlap: 0
+
+- ``divide_generation``: Enable patch-based generation for large domains.
```
frodre: Could we use an `autoclass` here to spit out the docstrings for the patch config?  Or is it worth duplicating the parameter descriptions here.

frodre (changes requested):
Thanks @AnnaKwa.  See my suggestions below.  The main one is I think that needs to be addressed is to tie the larger example config blocks to something that's also tested against the config build process.

frodre:
Also, we used to have a preview build/deploy option for the documentation in github actions.  It seems like that one may have been lost in the repo switch process.

docs/downscaling_inference.rst
AnnaKwa: Added links to APIs

docs/conf.py
```diff
@@ -231,3 +231,24 @@ def check_output(self, want, got, optionflags):
 doctest_global_setup = """
 import fme
 """
+
+
+# -- Custom docstring processing -----------------------------------------
+
+
+def remove_excluded_lines_from_docstring(app, what, name, obj, options, lines):
```
AnnaKwa, line 239: I added this feature so that long examples in docstrings, like the example yaml in the InferenceConfig, don't have to appear in the docs.

docs/downscaling_inference.rst
```diff
@@ -0,0 +1,279 @@
...
+Use the ``save_vars`` field to specify which variables to save to the output zarr file. If ``save_vars`` is ``null`` or not specified, all variables from the model output will be saved:
+
+.. code-block:: yaml
+
+   save_vars: ["PRATEsfc"]
+
+Patch Prediction for Large Domains
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For domains larger than the model's patch size, subdivision of the full domain into patches for prediction can be enabled. Configure this in the top-level ``patch`` section:
```
frodre: Currently, inference has some more restrictive behavior on output domain size given the patch training size.

 https://github.com/ai2cm/ace/blob/37617c4d409b16926d8281ac851f7ed64d4ede60/fme/downscaling/inference/inference.py#L72

I would just describe that we don't currently support generation on regions smaller than the training patch size, and that regions larger than the training patch size must use patch prediction.

frodre (approved):
Thanks, LGTM!

PR #748, opened 2026-01-20
Vendorize Apache 2.0 Nvidia Downscaling Code

frodre:
The code vendorized for diffusion/downscaling from the [EDM repo](https://github.com/NVlabs/edm) is under a CC BY-NC license, which is not compatible with Apache 2.0 (as flagged by legal).  This PR vendorizes code from `physicsnemo` (https://github.com/NVIDIA/physicsnemo/tree/327d9928abc17983ad7aa3df94da9566c197c468), which is Apache 2.0.

Changes:
- `fme.downscaling.modules.unets/preconditioners/layers`, `fme.downscaling.samplers`, and `fme.diffusion.stepper` vendorize code from `physicsnemo`
- Added regression tests against previous code for validation
- Removed unused `DhariwalUNet` module

- [x] Tests added


fme/downscaling/data/topography.py
```diff
@@ -116,7 +110,10 @@ def get_normalized_topography(path: str, topography_name: str = "HGTsfc"):
 
     topography_normalized = (topography - topography.mean()) / topography.std()
 
-    return Topography(data=torch.tensor(topography_normalized.values), coords=coords)
+    return Topography(
+        data=torch.tensor(topography_normalized.values, dtype=torch.float32),
```
frodre, line 120: I was getting a dtype mismatch in `test_inference.py`.  Potentially because normalization converts to float64 and `torch.tensor` preserves?

fme/downscaling/modules/preconditioners.py
```diff
@@ -1,58 +1,28 @@
...
+from physicsnemo.models.diffusion.preconditioning import EDMPrecond as _EDMPrecond
 
 
-class EDMPrecond(torch.nn.Module):
-    def __init__(self,
-        model,                              # The underlying neural network model. Add by <gideond@allenai.org>
-        label_dim       = 0,                # Number of class labels, 0 = unconditional.
-        use_fp16        = False,            # Execute the underlying model at FP16 precision?
-        sigma_data      = 0.5,              # Expected standard deviation of the training data.
+class EDMPrecond(_EDMPrecond):
```
frodre: Adapter to preserve our previous usage where we only provide an instantiated model object.  Seems like physicsnemo carries the model init info to then initialize the model.

fme/downscaling/modules/registry.py
```diff
@@ -143,43 +143,6 @@ def build(
         )
 
 
-@dataclasses.dataclass
-class UNetRegressionDhariwal:
```
frodre, line 147: Removed unused UNet style so we don't increase the vendorized dependencies for now.

fme/downscaling/modules/unets.py
```diff
@@ -256,262 +53,25 @@ def check_level_compatibility(
...
-            emb = emb + self.map_label(tmp)
-        emb = silu(emb)
-
-        # Encoder.
-        skips = []
-        for block in self.enc.values():
-            x = block(x, emb) if isinstance(block, UNetBlock) else block(x)
-            skips.append(x)
+    def forward(self, x, *args, **kwargs):
+        validate_shape(
```
frodre, line 381: Didn't have to do much here except recreate our validation which sources some more useful errors for the net.

fme/downscaling/samplers.py
```diff
@@ -1,80 +1,164 @@
 """
-This file is vendorized from edm/generate.py which you can find here:
-https://github.com/NVlabs/edm/blob/008a4e5316c8e3bfe61a62f874bddba254295afb/generate.py
+This file is vendorized from physicsnemo/physicsnemo/utis/generative/stochastic_sampler.py which you can find here:
```
frodre, line 2: Vendorized because this was a more extensive difference on Nvidia's side from what we previously used.   I mostly removed the unused conditioning/label/patching functionality.

fme/downscaling/samplers.py
```diff
@@ -1,80 +1,164 @@
...
-    x_next = latents.to(torch.float32) * t_steps[0]
-    latent_steps = [x_next]
-    for i, (t_cur, t_next) in enumerate(zip(t_steps[:-1], t_steps[1:])): # 0, ..., N-1
+    x_next = latents.to(torch.float64) * t_steps[0]
+    latent_steps = [x_next.to(latents.dtype)]
+    for i, (t_cur, t_next) in enumerate(zip(t_steps[:-1], t_steps[1:])):  # 0, ..., N-1
         x_cur = x_next
-
         # Increase noise temporarily.
-        gamma = min(S_churn / num_steps, np.sqrt(2) - 1) if S_min <= t_cur <= S_max else 0
```
frodre, line 42: We haven't been using `S_churn` > 0, but not sure of the effects of the removal of this minimum cap on the Nvidia side.

fme/downscaling/samplers.py
```diff
@@ -1,80 +1,164 @@
...
-        latent_steps.append(x_next)
-
-    return x_next, latent_steps
-
-
-#----------------------------------------------------------------------------
-# Wrapper for torch.Generator that allows specifying a different random seed
-# for each sample in a minibatch.
-
-class StackedRandomGenerator:
```
frodre, line 66: Unused by us

fme/downscaling/samplers.py
mcgibbon: This change likely is the cause of the regression test failing when S_churn is enabled.

fme/downscaling/samplers.py
```diff
@@ -1,80 +1,164 @@
...
-"Elucidating the Design Space of Diffusion-Based Generative Models"."""
 
-from typing import Tuple
+from typing import Callable, Optional
 
-import numpy as np
 import torch
+from torch import Tensor
+
+from physicsnemo.utils.patching import GridPatching2D
```
mcgibbon: I think this import is unused.

fme/diffusion/stepper.py
```diff
@@ -8,7 +8,9 @@
 
 import dacite
 import torch
+from physicsnemo.models.diffusion import Linear, PositionalEmbedding
```
mcgibbon: It looks like this module doesn't exist in the latest main of physicsnemo: https://github.com/NVIDIA/physicsnemo/tree/main/physicsnemo/models

diffusion_unets exists, but it doesn't have these symbols: https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/models/diffusion_unets/__init__.py

Here is the version you use, for reference: https://github.com/NVIDIA/physicsnemo/tree/327d9928abc17983ad7aa3df94da9566c197c468/physicsnemo/models/diffusion

mcgibbon: Linear and PositionalEmbedding are both basic torch modules without further dependencies, they will be easy to vendorize. The song u-net will be more difficult.

fme/downscaling/modules/preconditioners.py
mcgibbon: I am pretty sure we can remove the ModelMetaData import if we want to vendorize EDMPrecond - we don't use its features.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 cartopy>=0.22.0
 dacite
+einops>=0.8.1
```
frodre, line 3: MIT License

fme/downscaling/modules/preconditioners.py
frodre: Removed

frodre:
@mcgibbon, given the issue loading old checkpoints and the heavy dependency of `physicsnemo`, I went ahead and vendorized the UNet and EDMPrecond in addition to the sampler.  Only update to deps was `einops` for some of the vendorized layers.

AnnaKwa (approved):
Thanks Andre and Jeremy, the updates and regression tests look good to me.

PR #749, opened 2026-01-21
Spatial Distributed class. 

odiazib:
The SpatialTorchDistributed class is copied from the TorchDistributed class. I have not added all the methods and modifications needed for spatial parallelism. So far, I have done the following

- The comm object (distributed object from Makani) is initialized in __init__.
- The comm.py file was copied and saved in fme.core.distributed. This file requires nvidia-physicsnemo. Thus, nvidia-physicsnemo was added to the requirements.txt file.
 - The method local_batch_size was modified.

Changes:

- [ ] Tests added


fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,244 @@
...
+          self.local_rank = comm.get_local_rank()
+          self._device_id = self.local_rank
+          # distributed = True
+          torch.cuda.set_device(comm.get_local_rank())
+          torch.backends.cudnn.benchmark = True
+          torch.backends.cuda.matmul.allow_tf32 = True
+          torch.backends.cudnn.allow_tf32 = True
+        else:
+            raise ValueError(
+                "Spatial Distributed backend initialized without h_parallel_size or w_parallel_size set to 1"
```
elynnwu: I think you meant `h_parallel_size and w_parallel_size are both <=1`?

fme/core/distributed/comm.py
```diff
@@ -0,0 +1,201 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import time
+import logging
+import math
+from typing import Union
+import numpy as np
```
elynnwu: Make sure you run pre-commit, some of these imports are not used

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,244 @@
...
+        logger.debug(f"Barrier on rank {self.rank}")
+        torch.distributed.barrier(device_ids=self._device_ids)
+
+    def shutdown(self):
+        self.barrier()
+        logger.debug(f"Shutting down rank {self.rank}")
+        torch.distributed.destroy_process_group()
+
+
+def _gather_irregular(
```
elynnwu: Since these are shared functions from `torch_distributed`, move them to a separate file and import for both.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,244 @@
...
+class SpatialTorchDistributed(DistributedBackend):
+    """A Spatial distributed backend implementation."""
+
+    def __init__(self):
+        h_parallel_size = int(os.environ.get("H_PARALLEL_SIZE", 1))
+        w_parallel_size = int(os.environ.get("W_PARALLEL_SIZE", 1))
+        logger.debug(f" Spatial parallelism dimension in h {h_parallel_size}")
+        logger.debug(f" Spatial parallelism dimension in w {w_parallel_size}")
+        fin_parallel_size=1#args.fin_parallel_size
+        fout_parallel_size=1#args.fout_parallel_size
```
elynnwu: Are these always fixed to 1? Delete unused comments

fme/core/distributed/comm.py
```diff
@@ -0,0 +1,201 @@
...
+def cleanup():
+    global _DM
+    if _DM is not None:
+        _DM.cleanup()
+        _DM = None
+    return
+
+
+# initialization routine
+def init(model_parallel_sizes=[1, 1, 1, 1], model_parallel_names=["h", "w", "fin", "fout"], data_parallel_sizes=[1, -1], data_parallel_names=["ensemble", "batch"], verbose=False):
```
elynnwu: Will we ever configure `model_parallel_sizes` and `data_parallel_sizes` to something else?

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,244 @@
...
+        self.spatial_parallelism=False
+        if (h_parallel_size>1) or (w_parallel_size >1):
+          self.spatial_parallelism=True
+          logger.debug(" Spatial parallelism dimension in enable")
```
elynnwu: Spatial parallelism enabled

requirements.txt
```diff
@@ -20,3 +20,4 @@ torch>=2
 wandb[media]>=0.19.0
 xarray
 zarr>=3
+nvidia-physicsnemo
```
elynnwu: Let's make this a separate requirement file, make a `requirements-spatial-parallelism.txt`

fme/core/distributed/comm.py
odiazib: Yes, here we will configure comm with values other than 1:
https://github.com/ai2cm/ace/blob/ee2d73aa7d8a4e3dce119a8675924d7ddbf5d636/fme/core/distributed/spatial_torch_distributed.py#L43

fme/core/distributed/spatial_torch_distributed.py
odiazib: For now, these two parameters will be set to 1. I guess that in future work, the group could explore additional ways of parallelism in Makani.

fme/core/distributed/spatial_torch_distributed.py
odiazib: I did this: 
`from .torch_distributed import _gather_irregular, _pad_tensor_at_end, _unpad_tensor_at_end`

odiazib:
> I think it is worthwhile considering naming this something other than "Spatial" because while we are only doing the H and W decomposition here, we may do other types of decomposition in the future. I think a good alternative is simply "Makani" and I would prefix "makani_" in front of all files copied from the nvidia/makani repo, e.g., makani_co

Yes, we can adapt this naming convention. @elynnwu, @mcgibbon, @oliverwm1, do you have any recommendations on how to name this class and the Makani files? I like @mahf708’s idea of prefixing the filenames with “makani” and using “MakaniTorchDistributed” as the class name.

fme/core/distributed/comm.py
```diff
@@ -0,0 +1,201 @@
...
+from typing import Union
+import numpy as np
+
+# we are using the distributed manager from physicsnemo
+from physicsnemo.distributed.manager import DistributedManager
+from physicsnemo.distributed.config import ProcessGroupNode, ProcessGroupConfig
+
+# we need this
+_DM = None
+_COMM_ROOTS = {}
```
mcgibbon:
```suggestion
_COMM_ROOTS: dict[str, int] = {}
```

fme/core/distributed/comm.py
```diff
@@ -0,0 +1,201 @@
...
+    if _DM is not None:
+        return _DM.group(name)
+    else:
+        return None
+
+
+def get_root(name: str) -> int:
+    global _DM
+    global _COMM_ROOTS
+    if (name in _COMM_ROOTS) and (_DM.world_size > 1):
```
mcgibbon:
```suggestion
    if (name in _COMM_ROOTS) and (_DM is not None) and (_DM.world_size > 1):
```

fme/core/distributed/comm.py
mcgibbon: In particular, "h" and "w" parallelism (the main focus of this code) are done using model_parallel_sizes.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def __init__(self):
+        h_parallel_size = int(os.environ.get("H_PARALLEL_SIZE", 1))
+        w_parallel_size = int(os.environ.get("W_PARALLEL_SIZE", 1))
+        logger.debug(f" Spatial parallelism dimension in h {h_parallel_size}")
+        logger.debug(f" Spatial parallelism dimension in w {w_parallel_size}")
+        fin_parallel_size=1 #  in_parallel_size
+        fout_parallel_size=1 # fout_parallel_size
+        self.spatial_parallelism=False
+        if (h_parallel_size>1) or (w_parallel_size >1):
```
mcgibbon: Issue: This block uses two spaces for indentation instead of 4.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+          logger.debug(" Spatial parallelism enable.")
+          params: dict[str, Any] = {}
+          params["fin_parallel_size"] = fin_parallel_size
+          params["fout_parallel_size"] = fout_parallel_size
+          params["h_parallel_size"] = h_parallel_size
+          params["w_parallel_size"] = w_parallel_size
+          params["model_parallel_sizes"] = [h_parallel_size, w_parallel_size, fin_parallel_size, fout_parallel_size]
+          params["model_parallel_names"] = ["h", "w", "fin", "fout"]
+
+          comm.init(model_parallel_sizes=params["model_parallel_sizes"], model_parallel_names=params["model_parallel_names"], verbose=False)
```
mcgibbon: Issue: `params` is not used.

Suggestion:
```suggestion
          model_parallel_sizes= [h_parallel_size, w_parallel_size, fin_parallel_size, fout_parallel_size]
          model_parallel_names = ["h", "w", "fin", "fout"]

          comm.init(model_parallel_sizes=model_parallel_sizes, model_parallel_names=model_parallel_names, verbose=False)
```

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+class SpatialTorchDistributed(DistributedBackend):
+    """A Spatial distributed backend implementation."""
+
+    def __init__(self):
+        h_parallel_size = int(os.environ.get("H_PARALLEL_SIZE", 1))
+        w_parallel_size = int(os.environ.get("W_PARALLEL_SIZE", 1))
+        logger.debug(f" Spatial parallelism dimension in h {h_parallel_size}")
+        logger.debug(f" Spatial parallelism dimension in w {w_parallel_size}")
+        fin_parallel_size=1 #  in_parallel_size
+        fout_parallel_size=1 # fout_parallel_size
```
mcgibbon: Issue: fin and fout parallel size are unused (except by other code which is also unused).

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+
+          self.world_size = comm.get_world_size()
+          self._rank = comm.get_world_rank()
+          self.local_rank = comm.get_local_rank()
+          self._device_id = self.local_rank
+          # distributed = True
+          torch.cuda.set_device(comm.get_local_rank())
+          torch.backends.cudnn.benchmark = True
+          torch.backends.cuda.matmul.allow_tf32 = True
+          torch.backends.cudnn.allow_tf32 = True
```
mcgibbon: Potential Issue: these torch backend flags don't appear to be related to spatial parallelism.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+          params["w_parallel_size"] = w_parallel_size
+          params["model_parallel_sizes"] = [h_parallel_size, w_parallel_size, fin_parallel_size, fout_parallel_size]
+          params["model_parallel_names"] = ["h", "w", "fin", "fout"]
+
+          comm.init(model_parallel_sizes=params["model_parallel_sizes"], model_parallel_names=params["model_parallel_names"], verbose=False)
+
+          self.world_size = comm.get_world_size()
+          self._rank = comm.get_world_rank()
+          self.local_rank = comm.get_local_rank()
+          self._device_id = self.local_rank
```
mcgibbon: Issue: `self.local_rank` is unused except to set `self._device_id`, keep it as an unbound variable

```suggestion
          self._device_id =  comm.get_local_rank()
```

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+          # distributed = True
+          torch.cuda.set_device(comm.get_local_rank())
```
mcgibbon:
```suggestion
          torch.cuda.set_device(self._device_id)
```
Clearer intention.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @classmethod
+    def is_available(cls) -> bool:
+        """Check if torch distributed is available."""
+        h_parallel_size = int(os.environ.get("H_PARALLEL_SIZE", 1))
+        w_parallel_size = int(os.environ.get("W_PARALLEL_SIZE", 1))
+        spatial_parallelism=False
+        if (h_parallel_size>1) or (w_parallel_size >1):
+          spatial_parallelism=True
+        return torch.distributed.is_available() and spatial_parallelism
```
mcgibbon: Suggestion: Split off `self._spatial_parallelism_enabled()` logic into a separate helper function to make this function a bit clearer. Specifically, `torch.distributed.is_available()` is a very high level operation, while the spatial parallelism check here is very low-level. Splitting it to a helper keeps everything in `is_available` at the same level of abstraction.

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+    def total_ranks(self) -> int:
+        """Total number of processes."""
+        return self.world_size
+
+    def local_batch_size(self, batch_size: int) -> int:
+        return batch_size // comm.get_size("data")
+
+    def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor | None:
+        torch.distributed.all_reduce(tensor)
+        return tensor / self.total_ranks
```
mcgibbon: Is this correct? I would expect you need to handle spatial parallelism in these functions. Specifically, I expect you need to pass a `group` argument to `all_reduce` which tells it which spot on the globe it's in, so each process only reduces the mean in its patch. As written, I expect every rank will get the mean across all "tiles" and the spatial areas will be mixed into one.

This is done for example in physicsnemo where it uses `group=DistributedManager().group(process_group)` (not sure what process_group is) to set a group arg for the reduction function override for param.grad ([link](https://github.com/NVIDIA/physicsnemo/blob/main/physicsnemo/distributed/utils.py#L724))

Unfortunately I can't really find any other uses of distributed reduction in the physicsnemo code, not sure if they have inline spatial-map metrics the way we have.

requirements.txt
mcgibbon: This requirement should get removed from `requirements.txt`

mcgibbon:
I would discourage using "makani" as the name since the code is primarily from physicsnemo (not makani). Maybe we can use "Model", which is the terminology physicsnemo uses, instead of "Spatial".

fme/core/distributed/spatial_torch_distributed.py
```diff
@@ -0,0 +1,142 @@
...
+    def wrap_module(self, module: torch.nn.Module) -> torch.nn.Module:
+        if any(p.requires_grad for p in module.parameters()):
+            if using_gpu():
+                output_device = [self._device_id]
+            else:
+                output_device = None
+            return DistributedDataParallel(
+                SyncBatchNorm.convert_sync_batchnorm(module),
+                device_ids=self._device_ids,
+                output_device=output_device,
```
mcgibbon, line 124: `output_device` uses a duplicate of the code in `self._device_ids` - we should make this clearer by using it twice.

```suggestion
                device_ids=self._device_ids,
                output_device=self._device_ids,
```

odiazib:
> I would discourage using "makani" as the name since the code is primarily from physicsnemo (not makani). Maybe we can use "Model", which is the terminology physicsnemo uses, instead of "Spatial".

The file comm.py was a direct copy from Makani. It uses routines from PhysicsNemo. @mahf708, what do you think about using 'Model' instead of 'Makani' for the class name (`ModelTorchDistributed`)? I guess we will not use 'Makani' in the name of the files. Is this correct, @mcgibbon?

mahf708:
Yes, any name is good for me as long as it's clear. I defer to Jeremy and Elynn, so anything they decide is good for me

fme/core/distributed/spatial_torch_distributed.py
odiazib: I will remove these flags.

fme/core/distributed/spatial_torch_distributed.py
odiazib: I will remove these variables.

fme/core/distributed/spatial_torch_distributed.py
odiazib: I did not modify this method in the original branch. I added a couple of unit tests and the implementation works at least for the training part. I believe the only reductions involved are those that compute the loss and the metrics. This implementation should be fine in cases where spatial dimensions are not involved. If this approach is acceptable, I would like to address the remaining issue in a follow-up PR.

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+    lat = torch.linspace(-90, 90, nlat)
+
+    # Create the longitude tensor
+    lon = torch.linspace(0, 360, nlon)
+
+    input_tensor = torch.rand(batch_size, nlat, nlon)
+    return lat, lon, nlat, nlon, batch_size, input_tensor
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 4, reason="requires multi-GPU machine")
```
elynnwu: Can you relax this criteria to 2 GPUs? We can change the test to use 1x2 instead. I believe the GPU tests will run once you do that.

odiazib:
Hi @elynnwu and @mcgibbon 

Thanks for all your help reviewing this PR.  I believe I’ve addressed most of your comments, except for the remark about `reduce_mean` in the `ModelDistributedClass`. For that item, I added a couple of unit tests (based on the original branch), and in those tests we don’t need to modify the `reduce_mean` method since we are computing global means.
 I understand Jeremy’s comment (https://github.com/ai2cm/ace/pull/749#discussion_r2717770631), but I’d prefer to tackle that in a separate PR.
 
 A couple of other points:
- I noticed that Elynn added code so some of the unit tests from the large PR can run on the CPU ([see here](https://github.com/ai2cm/ace/blob/056e6be64397fb4c25672073bf0704be89b83f66/fme/ace/test_spatial_parallelism/test_coordinates_sp.py#L56)). Because this feature only works with multiple GPUs, I think we should only run those tests on machines with at least two GPUs.
- In this PR, I moved two tests:
 – https://github.com/ai2cm/ace/pull/749/changes#diff-dc411cc5de16d954da7529631e22cadcaf3b60f9a1a98e609b4bc51c452bdc45
 – https://github.com/ai2cm/ace/pull/749/changes#diff-97e504650e3ade1a2a43784a2a2508903fa7196d8cb0f1343333e1450f913a17
 In both, I added
` @pytest.mark.skipif(torch.cuda.device_count() < 4, reason="requires multi-GPU machine")` so they won’t run in CI unless there are at least four GPUs available.

Finally, I added the new unit tests under `fme/core/distributed/test_model_torch_distribute`d. Should I move them to a different directory?

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+            local_offset_w = crop_offset[1] + sum(shapes_w[: comm.get_rank("w")])
+        return (
+            slice(local_offset_h, local_offset_h + local_shape_h),
+            slice(local_offset_w, local_offset_w + local_shape_w),
+        )
+
+    def local_batch_size(self, batch_size: int) -> int:
+        return batch_size // comm.get_size("data")
+
+    def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor | None:
```
mcgibbon: This code is currently incorrect. Take for example a basic case where n_lat=1, n_lon=2, and W_PARALLEL_SIZE=2, running on 4 ranks. Let's say you have a batch_size=2 array with shape [n_batch, n_lat, n_lon], where the first batch member has data [[1, -1]], and the second batch member has data [[2, -2]].

If we were _not_ using spatial parallelism, we would have two ranks with the above data, the first rank having [[1, -1]] and the second rank having [[2, -2]], and the mean across ranks would be [[1.5, -1.5]]. This is the result we want to get regardless of whether or not we're using spatial parallelism.

If you're using spatial parallelism with 4 ranks, the ranks each have data [[1]], [[-1]], [[2]], and [[-2]]. The current code will reduce this to [[0]] on all ranks, but what we actually want is that the ranks in the "west" group reduce to [[1.5]], and that the ranks in the "east" group reduce to [[-1.5]], so that if we stitched it back to the global data we'd get [[1.5, -1.5]]. The current result of [[0, 0]] is a bug.

To fix it, we have to pass the reduction operation a keyword argument telling it what "group" this process is in. There is a helper method in physicsnemo for this, that I referenced earlier.

mcgibbon: I think it's very important to fix this before merging those methods, so that we aren't confused by incorrect answers on main. If we just aren't using the methods yet and want to implement them later, we should have the methods raise NotImplementedError() instead of giving the answers they currently do.

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+import torch
+
+from fme.core.coordinates import LatLonCoordinates
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+from fme.core.mask_provider import MaskProvider
+
+
+def int():
```
mcgibbon: Issue: This method name doesn't describe what it does, and overrides the `int` built-in. Please name it to something self-descriptive, like `setup()`.

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+    # Create the longitude tensor
+    lon = torch.linspace(0, 360, nlon)
+
+    input_tensor = torch.rand(batch_size, nlat, nlon)
+    return lat, lon, nlat, nlon, batch_size, input_tensor
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="requires multi-GPU machine")
+def test_lat_lon_ops_from_coords_w_sp():
+    lat_host, lon_host, nlat, nlon, batch_size, input_ = int()
```
mcgibbon: Suggestion (optional): This set-up function is only called once, so you could just put the set-up code in the test itself.

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+from fme.core.coordinates import LatLonCoordinates
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+from fme.core.mask_provider import MaskProvider
+
+
+def int():
+    # Define the sizes
+    torch.manual_seed(42)
+    torch.set_printoptions(precision=12, sci_mode=False)  # Adjust precision as needed
```
mcgibbon: Setting this would change the display settings for all tests than run after it. If we'd like to change the torch options to be more readable, we should put a line like this in our `conftest.py` so it runs on set-up.

```suggestion
```

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
mcgibbon: nit: remove trailing `_` from input or rename to `input_tensor`

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+    os.environ["H_PARALLEL_SIZE"] = "2"
+    os.environ["W_PARALLEL_SIZE"] = "1"
+    dist = Distributed.get_instance()
+    device = get_device()
+    lat = lat_host.to(device)
+    lon = lon_host.to(device)
+    coords = LatLonCoordinates(lat=lat, lon=lon)
+    gridded_ops = coords.get_gridded_operations(mask_provider=MaskProvider())
+    inp_local_host = (input_[:, *dist.get_local_slices((nlat, nlon))]).detach().clone()
+    inp_local = inp_local_host.to(device)
```
mcgibbon: nit: use full word for "inp"

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,51 @@
...
+
+    os.environ["H_PARALLEL_SIZE"] = "2"
+    os.environ["W_PARALLEL_SIZE"] = "1"
+    dist = Distributed.get_instance()
+    device = get_device()
+    lat = lat_host.to(device)
+    lon = lon_host.to(device)
+    coords = LatLonCoordinates(lat=lat, lon=lon)
+    gridded_ops = coords.get_gridded_operations(mask_provider=MaskProvider())
+    inp_local_host = (input_[:, *dist.get_local_slices((nlat, nlon))]).detach().clone()
```
mcgibbon: Suggestion: assert the shape of `inp_local_host` is what you expect. Otherwise this test would pass if spatial parallelism was getting disabled due to a bug.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @property
+    def total_ranks(self) -> int:
+        """Total number of processes."""
+        return self.world_size
+
+    def get_local_rank(self) -> int:
+        return self._device_id
+
+    def get_local_slices(self, crop_shape):
```
mcgibbon: Suggestion: the code here is pretty complex and this function will be widely used - add a unit test for get_local_slices. You may need to move the logic to some kind of helper function and test just the helper, since this object is difficult to construct in a test.

mcgibbon: Question: What is the meaning of "crop" shape? I thought this takes the full domain shape as its input argument.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def reduce_min(self, tensor: torch.Tensor) -> torch.Tensor | None:
+        torch.distributed.all_reduce(tensor, op=torch.distributed.ReduceOp.MIN)
+        return tensor
+
+    def reduce_max(self, tensor: torch.Tensor) -> torch.Tensor | None:
+        torch.distributed.all_reduce(tensor, op=torch.distributed.ReduceOp.MAX)
+        return tensor
+
+    def gather(self, tensor: torch.Tensor) -> list[torch.Tensor] | None:
```
mcgibbon, line 102: Suggestion: Raise NotImplementedError() instead or add a unit test for this and the other spatially-involved distributed methods (like gather_irregular).

fme/core/distributed/torch_distributed.py
```diff
@@ -74,6 +74,12 @@ def total_ranks(self) -> int:
         """Total number of processes."""
         return self.world_size
 
+    def get_local_rank(self) -> int:
+        return self._device_id
```
mcgibbon, line 78: Suggestion (optional): rename `self._device_id` to `self._local_rank`, since it is a little less confusing to see/read "my device id is my local rank" than reading "my local rank is my device id". Up to you, if you think the current way is clearer.

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
mcgibbon, line 1: Issue: It's out of style for our repo to have a directory containing just tests - generally we put tests in the same directory as the files they're testing.

If you want additional organization, you could put all of `model_torch_distributed` under a sub-directory, so you can use multiple modules and testing files. This better separates `comm.py` from the other backends as well.

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
```diff
@@ -0,0 +1,111 @@
...
+import torch
+
+from fme.ace.aggregator.one_step.reduced import MeanAggregator
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+from fme.core.gridded_ops import LatLonOperations
+from fme.core.loss import LossConfig
+
+
+def loss_builds_and_runs_wo_sp(global_mean_type):
```
mcgibbon:
```suggestion
def loss_builds_and_runs_without_spatial_parallelism(global_mean_type):
```

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
```diff
@@ -0,0 +1,111 @@
...
+    torch.save(data_tensor, os.path.join(tmp_path, "example_data.pt"))
+    torch.save(x, os.path.join(tmp_path, "x.pt"))
+    torch.save(y, os.path.join(tmp_path, "y.pt"))
+    print("loss", logs["metrics/loss"])
+    torch.save(logs["metrics/loss"], os.path.join(tmp_path, "loss.pt"))
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="requires multi-GPU machine")
+@pytest.mark.parametrize("global_mean_type", [None])
+def test_loss_builds_and_runs_with_sp(global_mean_type):
```
mcgibbon:
```suggestion
def test_loss_builds_and_runs_with_spatial_parallelism(global_mean_type):
```

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
```diff
@@ -0,0 +1,111 @@
...
+    with Distributed.force_non_distributed():
+        loss_builds_and_runs_wo_sp(global_mean_type)
+
+    os.environ["H_PARALLEL_SIZE"] = "2"
+    os.environ["W_PARALLEL_SIZE"] = "1"
+    nx = 8
+    ny = 8
+    tmp_path = "testdata-loss"
+    tensor_data_host = torch.load(os.path.join(tmp_path, "example_data.pt"))
```
mcgibbon: Question: where does this data come from?

fme/core/distributed/test_model_torch_distributed/test_coordinates_sp.py
odiazib: I will include the ‘int’ function in the test.

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
odiazib: I moved this block of code to the spatial-parallelism part of the test. When I wrote this code, ACE did not yet include the force_non_distributed context manager.

fme/core/distributed/test_model_torch_distributed/test_loss_sp.py
odiazib: This data came from the serial test, so I removed that part. Now that we have the force_non_distributed context manager, both tests can run simultaneously.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
odiazib: The unit test for get_local_slices is located at [here](https://github.com/ai2cm/ace/blob/oscar/sp-distributed-class/fme/core/distributed/model_torch_distributed/test_get_local_slices.py).

In this test, a tensor is created and then scattered across processes; its local slices are gathered into a new tensor. The test passes if the original tensor and the gathered tensor are equal.

I renamed `crop_shape` to `tensor_shape`.

fme/core/distributed/torch_distributed.py
odiazib: I did not see _local_rank defined in the base class. Should I add it?

fme/core/distributed/distributed.py
```diff
@@ -27,8 +28,10 @@ class Distributed:
     """
 
     def __init__(self, force_non_distributed: bool = False):
-        if TorchDistributed.is_available() and not force_non_distributed:
-            self._distributed: DistributedBackend = TorchDistributed()
+        if ModelTorchDistributed.is_available() and not force_non_distributed:
+            self._distributed: DistributedBackend = ModelTorchDistributed()
+        elif TorchDistributed.is_available() and not force_non_distributed:
+            self._distributed = TorchDistributed()
```
mahf708, line 34: do you wanna perserve the type like before, i.e.

```suggestion
            self._distributed: DistributedBackend = TorchDistributed()
```

fme/core/distributed/distributed.py
```diff
@@ -118,6 +124,9 @@ def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor:
         """
         return self._distributed.reduce_mean(tensor)
 
+    def get_local_slices(self, crop_shape):
+        return self._distributed.get_local_slices(crop_shape)
```
mahf708: I think in the other calls, you changed this to tensor_shape, right? it'd be good to keep things consistent

```suggestion
    def get_local_slices(self, tensor_shape):
        return self._distributed.get_local_slices(tensor_shape)
```

fme/core/distributed/non_distributed.py
```diff
@@ -32,6 +32,12 @@ def total_ranks(self) -> int:
         """Total number of processes."""
         return 1
 
+    def get_local_slices(self, crop_shape):
+        return tuple(slice(None, None) for _ in crop_shape)
```
mahf708: same, crop_shape->tensor_shape?

```suggestion
    def get_local_slices(self, tensor_shape):
        return tuple(slice(None, None) for _ in tensor_shape)
```

fme/core/testing/distributed.py
```diff
@@ -14,6 +14,9 @@ def __init__(self, fill_value: float, world_size: int):
     def local_batch_size(self, batch_size: int) -> int:
         return batch_size
 
+    def get_local_slices(self, crop_shape):
+        return tuple(slice(None, None) for _ in crop_shape)
```
mahf708: same crop_shape->tensor_shape

```suggestion
    def get_local_slices(self, tensor_shape):
        return tuple(slice(None, None) for _ in tensor_shape)
```

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def reduce_sum(self, tensor: torch.Tensor) -> torch.Tensor | None:
+        torch.distributed.all_reduce(tensor)
+        return tensor
+
+    def reduce_min(self, tensor: torch.Tensor) -> torch.Tensor | None:
+        torch.distributed.all_reduce(tensor, op=torch.distributed.ReduceOp.MIN)
+        return tensor
+
+    def reduce_max(self, tensor: torch.Tensor) -> torch.Tensor | None:
```
mahf708, line 89: not important, but for all reductions, you could potentially do a generic all-purpose reduce op that can reroute to the correct one or potentially do an einops-style one, e.g., the signature
```py
def reduce(tensor, op):
  # if op ...
  #   ...
  return tensor
```

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,153 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ModelTorchDistributed(DistributedBackend):
+    """A Spatial distributed backend implementation."""
+
+    def __init__(self):
+        h_parallel_size = int(os.environ.get("H_PARALLEL_SIZE", 1))
+        w_parallel_size = int(os.environ.get("W_PARALLEL_SIZE", 1))
```
mahf708, line 27: I am tiny bit uncomfortable with changing runtime behavior with env variables, but maybe that's how it's done? I defer to Jeremy and Elynn --- I've seen stuff like this in the og distributed class, so maybe that's the preferred way here

fme/core/distributed/model_torch_distributed/test_model_torch_distributed.py
```diff
@@ -0,0 +1,59 @@
+"""
+Note: AI was used to generate these unit tests."
+"""
+
```
mahf708: eh, after it is reviewed by humans, it's all good :) no need to keep this
```suggestion
```

fme/core/distributed/model_torch_distributed/test_coordinates_sp.py
```diff
@@ -0,0 +1,50 @@
...
+    gridded_ops = coords.get_gridded_operations(mask_provider=MaskProvider())
+    input_local_host = (
+        (input_tensor[:, *dist.get_local_slices((nlat, nlon))]).detach().clone()
+    )
+    input_local = input_local_host.to(device)
+    result_local = gridded_ops.area_weighted_mean(input_local, name="T_0")
+    result = dist.reduce_mean(result_local)
+    torch.testing.assert_close(result.to("cpu"), result_reference)
+    # Set H_PARALLEL_SIZE back to 1.
+    os.environ["H_PARALLEL_SIZE"] = "1"
```
mahf708: why I feel a little uneasy about changing runtime behavior with env variable is needing this type of clean-up 👀

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
odiazib: @mcgibbon, I reviewed the Makani implementation and found a couple of places where the “group” parameter was passed to the reduce operation. I ended up using “data” as the group in the `torch.distributed.all_reduce` routine. I also added a couple of unit tests for this method:

[Test 1](https://github.com/ai2cm/ace/blob/oscar/sp-distributed-class/fme/core/distributed/model_torch_distributed/test_reduce_meanv2.py) uses your simple example.
 
[Test 2](https://github.com/ai2cm/ace/blob/oscar/sp-distributed-class/fme/core/distributed/model_torch_distributed/test_reduce_mean.py) uses random values for the input tensor.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
odiazib: Using environment variables was an easy way to get things started with spatial parallelism. At this point, however, I believe we can consider an alternative.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
odiazib: @mcgibbon, any suggestions on this? I just copied and pasted from the torch.distributed class.

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
mahf708: it's ok to keep them (if that's the easiest thing)

fme/core/distributed/distributed.py
odiazib: I tried to add the type DistributedBackend to this line. However, when running pre-commit I get:

```
fme/core/distributed/distributed.py:34: error: Attribute "_distributed" already defined on line 32  [no-redef]
Found 1 error in 1 file (checked 1 source file)
```
I ran a single test, and it ran fine.

For this reason, I removed the type.

@mcgibbon, any suggestions?

fme/core/distributed/model_torch_distributed/model_torch_distributed.py
mahf708: Oh, if that follows a pattern already established, then please disregard my comment (sorry I should've checked)

fme/core/distributed/distributed.py
mahf708: ah interesting linting error. @odiazib feel free to disregard this comment and all my other comments; they were very minor

fme/core/testing/distributed.py
```diff
@@ -14,6 +14,9 @@ def __init__(self, fill_value: float, world_size: int):
     def local_batch_size(self, batch_size: int) -> int:
         return batch_size
 
+    def get_local_slices(self, tensor_shape):
+        return tuple(slice(None, None) for _ in tensor_shape)
+
     def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor:
```
elynnwu, line 20: Possibly a bug from previous changes, missing `group=None` here - looks like all `reduce_mean` calls have group as input

fme/core/distributed/model_torch_distributed/test_model_torch_distributed.py
```diff
@@ -0,0 +1,55 @@
+import os
+from unittest.mock import patch
+
+import pytest
+from model_torch_distributed import ModelTorchDistributed
```
elynnwu, line 5: Changed to `from fme.core.distributed.model_torch_distributed.model_torch_distributed import ModelTorchDistributed` so that this import always works and to be consistent with your other test imports.

fme/core/distributed/model_torch_distributed/test_reduce_meanv2.py
```diff
@@ -0,0 +1,86 @@
+import logging
```
elynnwu, line 1: Can you remove `test_reduce_mean.py`? Why are there two versions of this?

PR #750, opened 2026-01-21
Allow stochastic `CoupledStepper` training

jpdunc23:
Adds `n_ensemble` to `CoupledTrainStepperConfig` and handling for the ensemble dimension in `CoupledTrainStepper`.

Changes:
- `CoupledTrainStepper` init now has `config: CoupledTrainStepperConfig` as an arg rather than `loss: CoupledStepperTrainLoss`.

- [x] Tests added

### DIFF
--- a/fme/__init__.py
+++ b/fme/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "2026.1.1"
+__version__ = "2026.4.0"
 
 
 from . import ace, coupled


fme/coupled/test_train.py
```diff
@@ -242,10 +259,16 @@ def _write_test_yaml_files(
 
 
 @pytest.mark.parametrize(
-    "loss_atmos_n_steps",
-    [3, 0],
+    "loss_atmos_n_steps, crps_training",
+    [
+        # (3, False),
+        # (0, False),
```
elynnwu: I think you meant to keep these tests?

fme/coupled/stepper.py
```diff
@@ -1564,34 +1616,68 @@ def train_on_batch(
...
+            atmos_gen_data = unfold_ensemble_dim(
+                dict(gen_data.atmosphere_data.data), n_ensemble=1
+            )
         ocean_stepped = TrainOutput(
             metrics=metrics.get_ocean_metrics(),
-            gen_data=add_ensemble_dim(dict(gen_data.ocean_data.data)),
+            gen_data=ocean_gen_data,
             target_data=add_ensemble_dim(dict(ocean_forward_data.data)),
-            time=gen_data.ocean_data.time,
+            time=ocean_forward_data.time,
```
elynnwu, line 1708: Is this equivalent to `gen_data.ocean_data.time` when no ensemble is used?

fme/coupled/stepper.py
```diff
@@ -1350,16 +1378,22 @@ class ComponentTrainingConfig:
...
+            batch member. Default is 2 if ocean or atmopshere loss type is
+            EnsembleLoss, otherwise the default is 1. Must be 2 for EnsembleLoss
+            to be valid.
+        parameter_init: The coupled parameter initialization configuration for
+            fine-tuning a previously-trained coupled stepper.
     """
 
     ocean: ComponentTrainingConfig
     atmosphere: ComponentTrainingConfig
+    n_ensemble: int = -1
```
elynnwu: Copy the same comment from single module? `# sentinel value to avoid None typing of attribute`

fme/coupled/stepper.py
```diff
@@ -1218,6 +1222,30 @@ def _process_prediction_generator_list(
...
+        """Process generator list for ensemble atmosphere training.
+
+        Returns:
+            A tuple of (ocean_gen_data, atmos_gen_data) with explicit
+            ensemble dimension.
+
+        """
+        atmos_gen_data = process_ensemble_prediction_generator_list(
+            [
+                EnsembleTensorDict(x.data)  # FIXME: fix output_list typing
```
elynnwu: Maybe just cast `x.data` to `EnsembleTensorDict`?

fme/coupled/stepper.py
```diff
@@ -1564,34 +1616,68 @@ def train_on_batch(
...
             normalize=self.ocean.normalizer.normalize,
             derive_func=self.ocean.derive_func,
         )
         atmos_stepped = TrainOutput(
             metrics=metrics.get_atmosphere_metrics(),
-            gen_data=add_ensemble_dim(dict(gen_data.atmosphere_data.data)),
+            gen_data=atmos_gen_data,
             target_data=add_ensemble_dim(dict(atmos_forward_data.data)),
-            time=gen_data.atmosphere_data.time,
+            time=atmos_forward_data.time,  # Use original time (not broadcasted)
```
elynnwu: So is this the reason for the switch to forward data instead of gen for ocean too?

fme/coupled/stepper.py
```diff
@@ -1446,15 +1488,16 @@ class CoupledTrainStepper(
...
         """
         Args:
             stepper: The underlying coupled stepper for inference operations.
-            loss: The coupled loss object for computing per-component losses.
+            config: The train stepper config.
         """
         self._stepper = stepper
-        self._loss = loss
+        self._config = config
+        self._loss = self._config._build_loss(stepper)
```
elynnwu, line 1555: [nit] This would make `CoupledTrainStepper` harder to build in tests since you now need the full config, you can still just build the loss in `get_train_stepper` right?

fme/coupled/stepper.py
```diff
@@ -1522,18 +1565,27 @@ def train_on_batch(
                 prediction and target atmosphere data.
 
         """
-        # get initial condition prognostic variables
+        # Ensemble support: broadcast atmosphere data for ensemble training
+        n_ensemble = self._config.n_ensemble
+        atmos_data_ensemble = data.atmosphere_data.broadcast_ensemble(n_ensemble)
+        ocean_data_ensemble = data.ocean_data.broadcast_ensemble(n_ensemble)
```
elynnwu: My understanding is that you broadcast ocean data as well even though we currently only support training stochastic atmosphere, and the stochastic losses are propagated to ocean via the surface forcing variables. Can you add a short description in `n_ensemble`?

fme/coupled/stepper.py
```diff
@@ -1564,34 +1616,68 @@ def train_on_batch(
...
+        else:
+            gen_data = self._stepper._process_prediction_generator_list(
+                output_list, data_ensemble
+            )
+            ocean_gen_data = unfold_ensemble_dim(
+                dict(gen_data.ocean_data.data), n_ensemble=1
+            )
+            atmos_gen_data = unfold_ensemble_dim(
+                dict(gen_data.atmosphere_data.data), n_ensemble=1
+            )
```
elynnwu: Would be better if `_process_ensemble_prediction_generator_list` also accepts `n_ensemble=1` so you don't need to have if/else here.

fme/coupled/stepper.py
jpdunc23: Ya, the target data is `ocean_forward_data` so this is just a different but equivalent source for the time dim.

fme/coupled/stepper.py
jpdunc23: Refactored things a bit to get the typing in line with `fme/ace/stepper/single_module.py`. `CoupledTrainStepper` now deals with `ComponentEnsembleStepPrediction`, while `CoupledStepper` uses `ComponentStepPrediction`.

fme/coupled/stepper.py
jpdunc23: The main reason is that there is no more `gen_data` and `ocean_gen_data` and `atmos_gen_data` here are no longer `BatchData` objects, and getting the `time` attribute from the `ocean_forward_data` or `atmos_forward_data` is equivalent to what we had before (note that this is happening before `stepped = stepped.prepend_initial_condition(ic)`.

fme/coupled/stepper.py
jpdunc23: The changes to the tests are luckily minor since we were already building the `CoupledTrainStepperConfig` and using its `get_train_stepper()` method to build the `CoupledTrainStepper`. I can see some benefit to letting the loss be built outside of the `CoupledTrainStepper` init, but on the other hand this makes the init args identical to the `fme.ace.TrainStepper` init and avoids having to add an `n_ensemble: int` arg,  both of which are nice. Don't feel super strongly one way or another.

fme/coupled/stepper.py
jpdunc23: Was able to remove these `if n_ensemble > 1:` blocks.

fme/coupled/stepper.py
jpdunc23: I added some info about stochastic training assumptions to the `CoupledTrainStepper` docstring. Lmk if this is more clear now.

fme/coupled/stepper.py
```diff
@@ -755,46 +758,6 @@ def data(self) -> TensorDict:
     def step(self) -> int:
         return self._step
 
-    def detach(self, optimizer: OptimizationABC) -> "ComponentStepPrediction":
-        """Detach the data tensor map from the computation graph."""
-        return ComponentStepPrediction(
-            realm=self.realm,
-            data=optimizer.detach_if_using_gradient_accumulation(self.data),
-            step=self.step,
-        )
```
jpdunc23, line 764: `ComponentStepPrediction` is not returned by the `CoupledTrainStepper` methods now, so it doesn't need this `detach()` method.

fme/coupled/stepper.py
```diff
@@ -755,46 +758,6 @@ def data(self) -> TensorDict:
...
-
-    def __call__(
-        self,
-        prediction: ComponentStepPrediction,
-        target_data: TensorMapping,
-    ) -> torch.Tensor | None:
-        loss_obj = self._loss_objs[prediction.realm]
-        if loss_obj.step_is_optimized(prediction.step):
-            return loss_obj(prediction, target_data)
-        return None
```
jpdunc23, line 796: Moved below for a bit better organization. The only change is to the typing of the `prediction` arg to `__call__`, which is now `prediction: ComponentEnsembleStepPrediction`.

fme/coupled/stepper.py
```diff
@@ -1337,6 +1301,110 @@ def from_state(cls, state) -> "CoupledStepper":
...
+
+    @property
+    def data(self) -> EnsembleTensorDict:
+        return self._data
+
+    @property
+    def step(self) -> int:
+        return self._step
+
+    def detach_if_using_gradient_accumulation(
```
jpdunc23, line 1328: This replaces `ComponentStepPrediction.detach()`

fme/coupled/stepper.py
elynnwu: Okay, let's keep as it is to be consistent with `fme.ace.TrainStepper`

fme/coupled/stepper.py
elynnwu: Thanks, the docs are good.

PR #752, opened 2026-01-21
Add context-based positional embedding

mcgibbon:
This PR adds a configurable positional embedding based on context.

Changes:
- Added `context_pos_embed_dim: int` argument to NoiseConditionedSFNOBuilder
- Added positional embedding to Context, used in ConditionalLayerNorm

- [x] Tests added


mcgibbon:
Using this feature improves training, validation, and 20-step metrics for ERA5 1-step pretraining: https://wandb.ai/ai2cm/ace?nw=ayzklkr0ir8
image

fme/ace/registry/stochastic_sfno.py
```diff
@@ -46,41 +46,84 @@ class NoiseConditionedSFNO(torch.nn.Module):
...
                 dtype=x.dtype,
             )
         else:
             raise ValueError(f"Invalid noise type: {self.noise_type}")
 
+        if self.pos_embed is not None:
+            embedding_pos = self.pos_embed.repeat(noise.shape[0], 1, 1, 1)
+            if self.label_pos_embed is not None and labels is not None:
+                label_embedding_pos = torch.einsum(
+                    "blp, lpxy -> bpxy", labels, self.label_pos_embed
```
Arcomano1234:
```suggestion
                    "bl, lpxy -> bpxy", labels, self.label_pos_embed
```

Arcomano1234: `labels` is a 2d tensor with shape (B, L) right?

fme/core/models/conditional_sfno/layers.py
```diff
@@ -55,6 +56,7 @@ class Context:
     """
 
     embedding_scalar: torch.Tensor | None
+    embedding_pos: torch.Tensor | None
```
Arcomano1234, line 61: Add `embedding_pos` documentation to this class. E.g.,


    Parameters:
        embedding_scalar: The scalar embedding to condition on. The
            last dimension is the channel dimension.
        embedding_pos: The context based positional embedding, shape (batch_size, height, width)?
        labels: The labels to condition on, of shape (batch_size, n_labels).
        noise: The 2D noise embedding to condition on. The last
            three dimensions are (channels, height, width).

fme/core/models/conditional_sfno/test_layers.py
```diff
@@ -8,9 +8,10 @@
 
 @pytest.mark.parametrize("global_layer_norm", [True, False])
 @pytest.mark.parametrize("n_channels", [32])
-@pytest.mark.parametrize("embed_dim_scalar", [9, 0])
-@pytest.mark.parametrize("embed_dim_noise", [10, 0])
+@pytest.mark.parametrize("embed_dim_scalar", [20, 0])
```
Arcomano1234: Any reason you needed to change these values from old tests?

fme/core/step/test_step.py
```diff
@@ -159,8 +159,9 @@ def get_single_module_noise_conditioned_selector(
                             embed_dim=4,
                             noise_embed_dim=4,
                             noise_type="isotropic",
-                            filter_type="makani-linear",
+                            filter_type="linear",
```
Arcomano1234, line 162: Was this intentional? If so, should we also update the `filter_type` in  `fme/core/models/conditional_sfno/test_sfnonet.py` to "linear"?

Arcomano1234 (commented):
Left a few comments and had a few questions but overall looks good. My last suggestion which is optional is to have some type of warning if both label_pos_embed > 0 and pos_embed is true.

fme/core/models/conditional_sfno/test_layers.py
mcgibbon: Unclear, reverted and they still pass. Might be from a process of implementing it twice and re-merging main.

fme/core/step/test_step.py
mcgibbon: `test_sfnonet.py` intentionally has one test with "makani-linear", but uses "linear" for the "regression target hasn't changed" test. That way we know if either breaks, with more importance on the one we're actively using.

Arcomano1234: Thanks for the clarification that makes sense!

PR #753, opened 2026-01-23
Split stepper config into train and inference parts

mcgibbon:

Changes:

- [ ] Tests added

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -50,7 +50,7 @@ from fme.ace.stepper.parameter_init import (
     ParameterClassification,
     ParameterInitializationConfig,
 )
-from fme.ace.stepper.single_module import Stepper, StepperConfig, StepSelector
+from fme.ace.stepper.single_module import Stepper, StepSelector, TrainStepperConfig
 from fme.ace.stepper.time_length_probabilities import (
     TimeLengthMilestone,
     TimeLengthProbabilities,
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -24,7 +24,7 @@ from fme.ace.stepper import (
     load_stepper,
     load_stepper_config,
 )
-from fme.ace.stepper.single_module import StepperConfig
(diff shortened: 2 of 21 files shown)


mcgibbon:
> However, I don't think they accomplish the goal of "improving separation" in a way that can be practically leveraged elsewhere (e.g., in fme.coupled).

I believe it can be leveraged in fme.coupled. For example, here's a proof of concept which pretends to split ace's stepper into train and inference steppers, and then refactors the coupled stepper to do the same: https://github.com/ai2cm/ace/pull/754

This greatly simplifies the coupled training stepper, keeping only the attributes related to training and putting all of the "shuffle data back and forth between two sub-objects" logic into the inference stepper.

mcgibbon:
> I'd prefer to avoid additional config nesting.

Is there a motivation for this preference? We usually prefer config nesting over flattening, to better separate concerns in the config and have it match the object hierarchy. If the configs are getting too deeply nested in the coupled mode, we could expose a flatter user config in the coupled code and internally convert it to the ace nested versions.

Or maybe we can set it up so that the InferenceStepperConfig lives at the same level as TrainStepperConfig, and have `TrainStepperConfig.build` take in an `InferenceStepper` as argument.

mcgibbon:
> As I understand it, the motivation for this change is to improve the separation between two aspects of Stepper configuration

That's more a side-effect of the design of the planned change than the motivation. The configuration is not the primary change, though it's what I started with because of backwards compatibility concerns.

The motivation is that we have a lot of settings, methods and data currently attached to the Stepper which we don't need once we're done training, but the inference code is still exposed to them. This makes it harder for example to keep backwards compatibility for inference, but it also makes it harder to understand the Stepper code because a lot is going on. The idea is that the Stepper currently has two large and separate concerns, so they should be split into two objects.

The configuration change is just a side-effect of the fact that we want our configuration hierarchy to reflect our objects. While it's not great having a bunch of train and inference stuff at the same level of config, that's not the main issue. This new config hierarchy reflects the objects we'd like to refactor the current Stepper into because of these other concerns.

PR #755, opened 2026-01-23
Add `TrainStepper`

jpdunc23:
Add `TrainStepper` which implements `train_on_batch` for training `Stepper` modules. This intermediate PR makes no changes to training configuration. In future PRs, we plan to remove all training-specific stepper config attributes from `StepperConfig` and use `TrainStepperConfig` at the top level of `TrainConfig`.

Changes:
- Added `StepperConfig.get_train_stepper_config`, a temporary helper to create a `TrainStepperConfig` from the training-specific config attributes on `StepperConfig`.

- [x] Tests updated

### DIFF
--- a/latest_deps_only_image.txt
+++ b/latest_deps_only_image.txt
@@ -1,1 +1,1 @@
-andrep/fme-deps-only-0ec683fa
+annak/fme-deps-only-08db2881


fme/ace/stepper/single_module.py
```diff
@@ -492,74 +486,24 @@ class StepperConfig:
 
     Parameters:
         step: The step configuration.
-        loss: The loss configuration.
-        optimize_last_step_only: Whether to optimize only the last step.
-        n_ensemble: The number of ensemble members evaluated for each training
-            batch member. Default is 2 if the loss type is EnsembleLoss, otherwise
-            the default is 1. Must be 2 for EnsembleLoss to be valid.
         parameter_init: The parameter initialization configuration.
```
mcgibbon, line 500: `parameter_init` is something we only do at the start of training, we should be able to move it.

fme/ace/train/train_config.py
```diff
@@ -177,8 +180,17 @@ class TrainConfig:
         weather_evaluation: Configuration for weather evaluation.
             If None, no weather evaluation is run. Weather evaluation is not
             used to select checkpoints, but is used to provide metrics.
-        n_forward_steps: Number of forward steps during training. Cannot be given
-            at the same time as train_n_forward_steps in StepperConfig.
+        n_forward_steps: Number of forward steps during training.
```
mcgibbon: Please nest these and the respective method into a new config, something that uses `train_stepper_config.build(inference_stepper)` (or whatever these objects get named).

fme/core/generics/stepper.py
```diff
@@ -0,0 +1,105 @@
...
+
+from torch import nn
+
+PS = TypeVar("PS")  # prognostic state
+BD = TypeVar("BD")  # batch data
+FD = TypeVar("FD")  # forcing data
+SD = TypeVar("SD")  # stepped data
+
+
+class StepperABC(abc.ABC, Generic[PS, BD, FD, SD]):
```
mcgibbon: This isn't necessary until we use it in inference - we should add it in that PR. That way we know what methods it needs. Currently `StepperABC.method()` is not called anywhere for any method - its specific subclass implementations are called - meaning they're all unused (in the context of the ABC). By contrast, StepABC.step gets used by the stepper code.

fme/ace/stepper/single_module.py
```diff
@@ -906,40 +816,6 @@ def get_loss_obj() -> StepLoss:
...
-        else:
-            self._train_n_forward_steps_sampler = None
-        self._epoch = epoch
-
-    @property
-    def _loaded_loss_normalizer(self) -> StandardNormalizer:
-        if self._loss_normalizer is None:
-            loss_normalizer = self._step_obj.get_loss_normalizer()
-            self._loss_normalizer = loss_normalizer
-        return self._loss_normalizer
```
jpdunc23, line 935: The `_loaded_loss_normalizer` property was unused so this has been removed.

fme/ace/stepper/single_module.py
```diff
@@ -906,40 +816,6 @@ def get_loss_obj() -> StepLoss:
...
-        if self._epoch == epoch:
-            return
-        if self._train_n_forward_steps_schedule is not None:
-            assert epoch is not None  # already checked, but needed for mypy
-            self._train_n_forward_steps_sampler = probabilities_from_time_length(
-                self._train_n_forward_steps_schedule.get_value(epoch)
-            )
-        else:
-            self._train_n_forward_steps_sampler = None
-        self._epoch = epoch
```
jpdunc23, line 928: Moved to `TrainStepper`

fme/ace/stepper/single_module.py
```diff
@@ -999,15 +963,6 @@ def _append_training_history_from(
...
             self._training_history.extend(base_training_history)
 
-    @property
-    def effective_loss_scaling(self) -> TensorDict:
-        """
-        Effective loss scalings used to normalize outputs before computing loss.
-        y_loss_normalized_i = (y_i - y_mean_i) / loss_scaling_i
-        where loss_scaling_i = loss_normalizer_std_i / weight_i.
-        """
-        return self.loss_obj.effective_loss_scaling
```
jpdunc23, line 1009: Moved to `TrainStepper`

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1290,267 @@ def get_forward_data(
...
+            dataset_info = DatasetInfo.from_state(state["dataset_info"])
+        training_history = TrainingHistory.from_state(state.get("training_history", []))
+        stepper = config.get_stepper(
+            dataset_info=dataset_info,
+            training_history=training_history,
+            # don't need to initialize weights, we're about to load_state
+            apply_parameter_init=False,
+        )
+        stepper.load_state(state)
+        return stepper
```
jpdunc23, line 1369: These `Stepper` methods (`update_training_history`, `load_state`, and `from_state`) haven't actually changed.

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1290,267 @@ def get_forward_data(
...
+            normalizer=loss_normalizer,
+        )
+        return TrainStepper(
+            stepper=stepper,
+            config=self,
+            loss_obj=loss_obj,
+        )
+
+
+class TrainStepper(
```
jpdunc23, line 1434: I originally had `TrainStepperConfig` and `TrainStepper` in a separate file, but decided to put here to reduce the diff somewhat.

fme/ace/stepper/single_module.py
```diff
@@ -1364,10 +1580,12 @@ def train_on_batch(
         self._init_for_epoch(data.epoch)
         metrics: dict[str, float] = {}
         input_data = data.get_start(self.prognostic_names, self.n_ic_timesteps)
-        target_data = self.get_forward_data(data, compute_derived_variables=False)
-        data = self._forcing_deriver(data)
+        target_data = self._stepper.get_forward_data(
+            data, compute_derived_variables=False
+        )
+        data = self._stepper._forcing_deriver(data)
```
jpdunc23: One of three places where `TrainStepper` accesses private state on `Stepper`: `_forcing_deriver`, `_get_regularizer_loss`, and `_predict_generator`.

* `_predict_generator`: We can maybe refactor `_accumulate_loss` to use `get_prediction_generator` instead of `_prediction_generator`. 
* `_forcing_deriver`: This feels like something that should be able to live exclusively on `Stepper` so that `TrainStepper` doesn't need to worry about it. For example, maybe this can be called once in `Stepper.get_prediction_generator`?
* `_get_regularizer_loss`: This should (eventually) be owned by `TrainStepper`. It needs `ParameterInitializer` which for now we're not moving onto `TrainStepperConfig`. It also calls the `get_regularizer_loss` method of `Stepper._step_obj`; I couldn't find an example where this doesn't just return `torch.tensor(0.0)`, though I may have missed one. If not, maybe it is time to revisit whether or not we need the `_step_obj.get_regularizer_loss()` term.

fme/ace/stepper/single_module.py
```diff
@@ -789,6 +789,14 @@ def from_state(cls, state) -> "StepperConfig":
             data_class=cls, data=state, config=dacite.Config(strict=True)
         )
 
+    def get_train_stepper_config(self) -> "TrainStepperConfig":
```
jpdunc23, line 763: Temporary helper until `TrainStepperConfig` is added to `TrainConfig`

fme/ace/stepper/single_module.py
```diff
@@ -789,6 +789,14 @@ def from_state(cls, state) -> "StepperConfig":
...
+        return TrainStepperConfig(
+            loss=self.loss,
+            optimize_last_step_only=self.optimize_last_step_only,
+            n_ensemble=self.n_ensemble,
+            train_n_forward_steps=self.train_n_forward_steps,
```
mcgibbon, line 768: Are we agreed about `parameter_init` going in `TrainStepperConfig`, and it's happening in a later PR?

jpdunc23: Yes, agreed.

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1307,257 @@ def get_forward_data(
     def _get_regularizer_loss(self) -> torch.Tensor:
         return self._l2_sp_tuning_regularizer() + self._step_obj.get_regularizer_loss()
 
+    def train_on_batch(
```
mcgibbon: Question: Why do we need this NotImplemented method?

fme/ace/stepper/single_module.py
```diff
@@ -1474,102 +1699,112 @@ def update_training_history(self, training_job: TrainingJob) -> None:
...
-            state: The state to load.
-        """
-        self._step_obj.load_state(state["step"])
+        self._stepper.load_state(state)
 
     @classmethod
-    def from_state(cls, state) -> "Stepper":
-        """
-        Load the state of the stepper.
+    def from_state(cls, state: dict[str, Any]) -> "TrainStepper":
```
mcgibbon: Question: Why do we need this `NotImplemented` method?

fme/ace/stepper/single_module.py
```diff
@@ -1474,102 +1699,112 @@ def update_training_history(self, training_job: TrainingJob) -> None:
...
+
+    @property
+    def derive_func(self):
+        return self._stepper.derive_func
+
+    @property
+    def training_history(self) -> TrainingHistory:
+        return self._stepper.training_history
+
+    def predict(
```
mcgibbon: Suggestion: This method is unused, delete it.

fme/ace/stepper/single_module.py
```diff
@@ -1474,102 +1699,112 @@ def update_training_history(self, training_job: TrainingJob) -> None:
...
+        compute_derived_forcings: bool = True,
+    ) -> tuple[BatchData, PrognosticState]:
+        return self._stepper.predict(
+            initial_condition,
+            forcing,
+            compute_derived_variables,
+            compute_derived_forcings,
+        )
+
+    def predict_paired(
```
mcgibbon, line 1646: Suggestion: This method is unused, delete it.

fme/ace/stepper/test_single_module.py
```diff
@@ -171,10 +172,12 @@ def test_stepper_no_train_step_specified():
...
             ),
         ),
-        loss=StepLossConfig(type="MSE"),
     )
     dataset_info = get_dataset_info()
-    stepper = config.get_stepper(dataset_info)
+    train_stepper_config = TrainStepperConfig(
+        loss=StepLossConfig(type="MSE"),
+    )
+    stepper = train_stepper_config.get_train_stepper(config.get_stepper(dataset_info))
```
mcgibbon: Suggestion (optional): Now that these are decoupled, you could pass a much simpler mock/testing version of Stepper to `get_train_stepper`. Might be more trouble than it's worth for updating in this PR, but something to consider, especially if you can make a helper in `fme.ace.testing` that simplifies many tests.

fme/ace/train/train_config.py
```diff
@@ -177,8 +177,7 @@ class TrainConfig:
         weather_evaluation: Configuration for weather evaluation.
             If None, no weather evaluation is run. Weather evaluation is not
             used to select checkpoints, but is used to provide metrics.
-        n_forward_steps: Number of forward steps during training. Cannot be given
-            at the same time as train_n_forward_steps in StepperConfig.
```
mcgibbon, line 183: Is this doc change intentional, or leftover from previous refactors?

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1307,257 @@ def get_forward_data(
...
+
+    @property
+    def train_n_forward_steps_schedule(self) -> TimeLengthSchedule | None:
+        if self.train_n_forward_steps is None:
+            return None
+        if isinstance(self.train_n_forward_steps, TimeLengthSchedule):
+            return self.train_n_forward_steps
+        return TimeLengthSchedule.from_constant(self.train_n_forward_steps)
+
+    def get_train_window_data_requirements(
```
mcgibbon: Issue (optional): This method's API is a little awkward, in part because it takes in a large object (StepperConfig) it only uses a little bit of.

Suggestion (optional): Look at your current `StepperConfig.get_train_window_data_requirements` (which should probably get deleted). If you can get `n_forward_steps` from the TrainConfig, you can pass just this to `StepperConfig.get_evaluation_window_data_requirements`, minimizing the data transfer/coupling between these configs and minimizing duplicated code.

fme/ace/stepper/single_module.py
mcgibbon: Could move them in a follow-on PR if you like.

fme/ace/stepper/single_module.py
jpdunc23: I tried, but this method is used by the generic `Trainer`.

fme/ace/stepper/single_module.py
```diff
@@ -891,13 +889,7 @@ def get_loss_obj() -> StepLoss:
             base_training_history=self._parameter_initializer.training_history
         )
 
-        _1: PredictFunction[  # for type checking
-            PrognosticState,
-            BatchData,
-            BatchData,
-        ] = self.predict
```
jpdunc23, line 898: Not needed since `Stepper.predict` is unused externally.

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1299,249 @@ def get_forward_data(
...
+        stepper.load_state(state)
+        return stepper
+
+    def set_eval(self) -> None:
+        for module in self.modules:
+            module.eval()
+
+    def set_train(self) -> None:
+        for module in self.modules:
+            module.train()
```
jpdunc23, line 1377: These were previously implemented on `TrainStepperABC`. They are required by `Stepper` because `CoupledStepper` uses them.

fme/ace/stepper/single_module.py
jpdunc23: No longer needed.

fme/ace/stepper/single_module.py
jpdunc23: No longer needed.

fme/ace/stepper/test_single_module.py
jpdunc23: I was able to use `unittest.mock.Mock` in some of the tests, and simplified two more by reusing the existing `_get_stepper` helper.

In a future PR I agree it might be worth moving `_get_stepper` to be a public method in `fme.ace.testing` and reuse as much as possible where tests create `Stepper` instances throughout `fme`

fme/ace/train/train_config.py
jpdunc23: Good catch, it was leftover. Reverted the change.

fme/ace/stepper/single_module.py
jpdunc23: Good points, updated.

fme/ace/stepper/single_module.py
jpdunc23: Agreed and plan to handle in a future PR. I think this will require some significant refactors in `fme.coupled`.

jpdunc23:
> Could you please do a once-over of the refactored objects, and make sure any methods or attributes that aren't used are deleted or made private?

I've done so. As mentioned in https://github.com/ai2cm/ace/issues/767, `Stepper.predict` could be made private but we should discuss at the next ACE technical sync.

fme/ace/stepper/single_module.py
mcgibbon: I don't mind this being in another PR, but doesn't it "just work" in fme.coupled? We're keeping the APIs the coupled code is exposed to (the StepperConfig for example) stable in this PR.

If we do this refactor after this PR, let's make sure we do it before we change the config yaml API.

fme/core/generics/stepper.py
```diff
@@ -23,7 +23,14 @@ def get_metrics(self) -> TensorDict:
 SD = TypeVar("SD")  # stepped data
 
 
-class TrainStepperABC(abc.ABC, Generic[PS, BD, FD, SD, TO]):
+class StepperABC(abc.ABC, Generic[PS, FD, SD]):
+    @property
+    @abc.abstractmethod
+    def predict_paired(self) -> PredictFunction[PS, FD, SD]:
```
mcgibbon: Does StepperABC need to exist / is it used (i.e. is there ever a variable that has a StepperABC type, and not the type of a concrete implementation)? Inference only depends on the PredictFunction type, which is how we've gotten around the awkwardness of `predict` and `predict_paired` having two different names. It seems to me that this ABC is a more complicated way to pass a PredictFunction.

Otherwise, I don't know if there's a way around it, but it feels very strange to me that this object is generic to the type of data it's processing, but we have the `_paired` descriptor in the name that's meant to indicate we're using PairedData.

fme/ace/stepper/single_module.py
```diff
@@ -906,34 +877,6 @@ def get_loss_obj() -> StepLoss:
...
-
-    @property
-    def _loaded_loss_normalizer(self) -> StandardNormalizer:
-        if self._loss_normalizer is None:
-            loss_normalizer = self._step_obj.get_loss_normalizer()
-            self._loss_normalizer = loss_normalizer
-        return self._loss_normalizer
-
     @property
     def loss_obj(self) -> StepLoss:
```
mcgibbon, line 867: Issue: `loss_obj` is training-specific and should be on the training stepper - we don't need it for inference. The coupled code uses TrainStepper, so it should be agnostic to this choice.

Same with `effective_loss_scaling`.

If for some reason we really did want to keep this information with the inference stepper, it would go on the training history, since multiple losses can get used.

fme/ace/stepper/single_module.py
mcgibbon: nit: if you removed these from the Stepper, CoupledStepper and TrainStepper would still work, because their existing implementations of set_train and set_eval (from the generic class) access the `.modules` attribute on Stepper (which is what's really needed) through their own implementations of `.modules` (which is what the TrainStepperABC code was accessing).

However that version of the code is also a little confusing (which for example led to the impression this wouldn't work, or is giving me the false impression it would work, if I'm just wrong). I'm not sure which is better (since this one does expose more API, which might be confusing for different reasons).

fme/ace/stepper/single_module.py
```diff
@@ -1335,6 +1278,242 @@ def get_forward_data(
...
+
+    def __post_init__(self):
+        if self.n_ensemble == -1:
+            if self.loss.type == "EnsembleLoss":
+                self.n_ensemble = 2
+            else:
+                self.n_ensemble = 1
+
+    @property
+    def train_n_forward_steps_schedule(self) -> TimeLengthSchedule | None:
```
mcgibbon, line 1411: Question: Possibly a feature of the existing code, but why do both `train_n_forward_steps_schedule` and `get_n_forward_steps` exist? They do pretty similar things.

fme/ace/stepper/single_module.py
mcgibbon: Please make any of these private functions still being called public before merging this PR.

`_predict_generator`: I think this should just be a public method for now, that gets ingested by TrainStepper. There's the worry someone external will come to depend on it and then we will break compatibility, but I'd rather not make the dependencies in our code harder for us to track for that hypothetical (if we do, they can update their APIs - we give no guarantees).

`_forcing_deriver`: yes, it only does not because time isn't passed to the "bare" predict methods and those are needed to derive forcings. This is concerning to me though - it means we can't train a model using the deriver. Derived insolation is most "valid" to use with a model if it was trained with derived insolation, otherwise small differences between the deriver and the training data may matter.

In any case yes, we should move it there. For now please make it public if it's getting coupled like this.

`_get_regularizer_loss`: I'd be OK with removing the features that involve loss regularization, this is used by a feature that lets us put a loss on the weights becoming too far away from the initialization values (L2-SP regularization). We didn't use it because it didn't solve any problems, and LoRA serves the same purpose and is usually preferred nowadays.

We can only really remove it after these refactors are through though, since changes to the Stepper config will currently break inference. For now either move parameter_init to the TrainStepper (which I'd prefer if it's possible, or do it in the next PR), or make this public.

fme/ace/stepper/single_module.py
jpdunc23: This is planned for a future PR. The coupled code uses `Stepper`, not `TrainStepper`. This is intentional, since `CoupledStepper` doesn't use `TrainStepper.train_on_batch`. For the time being I'm leaving it on `Stepper` to avoid breaking the current `CoupledStepper` implementation.

Tasks I have in mind for future PR(s) are:

1. Add `CoupledTrainStepper`, analogous to `TrainStepper` (started by you in https://github.com/ai2cm/ace/pull/754)
2. Update the coupled `TrainConfig` to support direct configuration of each component's `loss` and `parameter_init`, rather than relying on these attributes from `StepperConfig`, as `CoupledStepper` currently does.
3. With this done, I plan to simultaneously remove `loss_obj` from `Stepper` and `loss` from `StepperConfig`.

There is a way to do 3 without first doing 1 and 2 by building each component's `StepLoss` directly from the `loss: StepLossConfig` attribute on `StepperConfig`, but I'd prefer to avoid this temporary workaround since it mostly won't survive in later stages of refactoring. 

That said, task 2 will also involve building the component loss objects in the coupled stepper code, so if you feel strongly then I'm willing to give it a shot and hopefully some of the effort will be worthwhile later.

fme/ace/stepper/single_module.py
jpdunc23: Good point, I hadn't looked carefully at `train_n_forward_steps_schedule`. The reason they both existed was because we haven't yet removed support for `TrainConfig.n_forward_steps`, so `get_n_forward_steps` just handled the fallback to this "default" int value for constructing the data requirements. I moved that "default" handling to a private method on `TrainBuilders` which in the future we can remove at the same time as removing `TrainConfig.n_forward_steps`.

Regarding timing of removing `TrainConfig.n_forward_steps`, I think we should do it at the same time as we add a `train_stepper` attribute to `TrainConfig` and remove the corresponding training config attributes from `StepperConfig`. This way folks will only have to update their train configs once (assuming they haven't already migrated to using `StepperConfig.train_n_forward_steps`.

fme/ace/stepper/single_module.py
mcgibbon: > This is planned for a future PR. The coupled code uses Stepper, not TrainStepper.

I think this is the correct final state, but it’s not necessary for this PR and is getting in the way of making the changes needed in ace. The coupled code currently depends on the stepper in ace that supports training. I don’t think that should be changed in this PR. We should make that change after more of 1-3 is complete.

fme/core/generics/stepper.py
jpdunc23: I removed `StepperABC`. The guarantees we need for generic inference are on the generic API of `run_inference`, so you're correct that it isn't required for type checking. 

The one way `StepperABC` may be useful is as a reminder / way to enforce what the minimal `Stepper` API should actually be. Maybe in a later PR we can think about how to best document this.

fme/ace/stepper/single_module.py
jpdunc23: Done.

fme/ace/stepper/single_module.py
jpdunc23: I agree this should work, but I personally prefer this more verbose version. Lmk if you prefer I bring back the TrainStepperABC method, in which case I will make it `@final` (previously it was overridden by CoupledStepper).

fme/core/generics/stepper.py
mcgibbon: In my experience it leads to confusion and difficult-to-modify code when there's an ABC that is never directly instantiated in the abstract sense, because invariably it gets additional methods that don't actually get used. But if what you mean is "the minimal Stepper API for use in the coupled code", you could make a `StepperForCoupledABC` within `fme.ace.stepper`, have the Stepper inherit it, and instantiate `self.stepper: StepperForCoupledABC = Stepper()` within the coupled code. If you hint it this way, mypy errors if you access a non-abc-documented method.

The one issue with this is, having multiple implementations is the only automated method encouraging the ABC to stay minimal. So we would need to take care when editing/reviewing edits to the ABC, and only approve additional methods when they get used in Coupled.

fme/ace/stepper/single_module.py
jpdunc23: I'm trying to scope this out but meeting a lot of resistance. I don't see a simple way to have CoupledStepper use TrainStepper without significant refactors in `fme.coupled` that I think are better left to the next PR.

- While it's true that CoupledStepper was using a "training" Stepper before, TrainStepper is a very different object from what Stepper was. On the other hand, aside from no longer having a `train_on_batch` method, Stepper is basically unchanged so I think CoupledStepper should still use it for the time being.
- To do what you want then CoupledStepperConfig will have to already use TrainStepperConfig in this PR. This is planned but involves significant refactoring in `fme.coupled`. I think we should wait until a later PR to do this refactoring.

mcgibbon: After looking at the code a bit I'm realizing this boils down to the choice to update the load_stepper function to return the more minimal inference-stepping class within this PR. I don't think at this point we should revert that change, so we can move forward with the way you had planned to do it.

You would have had more freedom to define the separation between the two classes in this PR if the load_stepper update weren't combined with it in the same PR (i.e. if load_stepper returned an object with the same API as it does in main), but we can move forward with the path you're currently on in this PR.

fme/core/generics/stepper.py
jpdunc23: I'm OK with having no ABC for Stepper.

PR #757, opened 2026-01-26
Simplify CopyWeightsConfig like frozen config

mcgibbon:
Changes:
- CopyWeightsConfig now takes either include or exclude arguments, and no longer requires every parameter be matched by a rule. This brings it in line with FrozenWeightsConfig.

- [ ] Tests added

### DIFF
--- a/fme/core/weight_ops.py
+++ b/fme/core/weight_ops.py
@@ -24,16 +24,16 @@ class CopyWeightsConfig:
     This is less efficient than true parameter freezing, but layer
     freezing is all-or-nothing for each parameter.
 
-    All parameters must be covered by either the include or exclude list,
-    but not both.
-
     Parameters:
-        include: list of wildcard patterns to overwrite
-        exclude: list of wildcard patterns to exclude from overwriting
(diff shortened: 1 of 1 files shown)


fme/core/weight_ops.py
```diff
@@ -69,12 +62,18 @@ def apply(self, weights: list[Mapping[str, Any]], modules: list[nn.Module]):
             raise NotImplementedError("only one module currently supported")
         if len(modules) != len(weights):
             raise ValueError("number of modules and weights must match")
-        for module, weight in zip(modules, weights):
-
-            def func(module, name):
-                overwrite_weight_initial_slice(module, name, weight[name])
-
-            apply_by_wildcard(module, func, self.include, self.exclude)
+        module = modules[0]
```
brianhenn, line 65: Thanks, this is more consistent with the note above.

fme/core/weight_ops.py
brianhenn: (chore): Update docstring to reflect that all parameters don't need to be covered.

mcgibbon: Updated

PR #758, opened 2026-01-27
Add `use_amp_bf16` option to `UNetDiffusionModule`

AnnaKwa:
This PR adds an optional config field `use_amp_bf16` to the downscaling diffusion UNet module to use automatic mixed precision casting to bfloat16 in the module's forward pass. By default this is set to False, which maintains previous behavior.

- [x] Tests added

### DIFF
--- a/fme/downscaling/modules/test_diffusion_registry.py
+++ b/fme/downscaling/modules/test_diffusion_registry.py
@@ -163,10 +163,10 @@ def test_UNetDiffusionModule_use_amp_precision(use_amp_bf16):
             f"dtypes {captured_dtypes}."
         )
     else:
-        assert all(
-            dtype == torch.float32 for dtype in captured_dtypes
-        ), ("Expected all dtypes to be float32 when use_amp_bf16=False, "
-        f"but got captured dtypes {captured_dtypes}.")
+        assert all(dtype == torch.float32 for dtype in captured_dtypes), (
+            "Expected all dtypes to be float32 when use_amp_bf16=False, "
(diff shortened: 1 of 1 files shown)


fme/downscaling/modules/preconditioners.py
```diff
@@ -106,7 +107,7 @@ def forward(
             class_labels=class_labels,
         )
 
-        if (F_x.dtype != dtype) and not torch.is_autocast_enabled():
```
AnnaKwa, line 109: Old API only checks if enabled on cuda, so CPU tests fail here.

fme/downscaling/modules/test_diffusion_registry.py
```diff
@@ -103,3 +103,71 @@ def test_UNetDiffusionModule_forward_pass():
...
+    if use_amp_bf16:
+        assert torch.bfloat16 in captured_dtypes, (
+            "Autocast to bfloat16 was set but did not find bfloat16 in captured "
+            f"dtypes {captured_dtypes}."
+        )
+    else:
+        assert all(
+            dtype == torch.float32 for dtype in captured_dtypes
+        ), "Expected all dtypes to be float32 when use_amp_bf16=False, "
+        f"but got captured dtypes {captured_dtypes}."
```
elynnwu: The second line message is not concatenated to the first one, something like this should work:
```suggestion
        ), ("Expected all dtypes to be float32 when use_amp_bf16=False, "
        f"but got captured dtypes {captured_dtypes}.")
```

PR #759, opened 2026-01-27
Energy correction supports other temperature prefix

elynnwu:
We previously hard code energy correction to `air_temperature_`. This PR checks air temperature prefixes in `ATMOSPHERE_FIELD_NAME_PREFIXES` and uses the one that matches in the dataset.

- [x] Tests added


elynnwu:
Good idea, updated to use stacker's `get_all_level_names`

PR #768, opened 2026-01-29
Simplify s2 contraction code by removing unused features

mcgibbon:
The spectral convolution code currently has a bunch of features we don't use, which use indirection and make it harder to understand and modify the filter. This PR removes the unused features and performs the contraction more directly.

Changes:
- `operator_type` other than dhconv may no longer be used in NoiseConditionedSFNO
- `separable=True` may no longer be used in NoiseConditionedSFNO (previously it would lead to an exception, as it had a bug)
- `factorization` other than `None` may no longer be used in NoiseConditionedSFNO
- `filter_type` value of `non-linear` is no longer supported, and was never used by us due to the artifacting it causes. Default is updated to `linear` which we have been using for all models.

- [ ] Tests added

### DIFF
--- a/scripts/data_process/configs/shield-amip-ensemble-c96-1deg-8layer.yaml
+++ b/scripts/data_process/configs/shield-amip-ensemble-c96-1deg-8layer.yaml
@@ -1,10 +1,10 @@
 runs:
   ic_0001: gs://vcm-ml-raw-flexible-retention/2024-06-29-C96-SHiELD-AMIP/regridded-zarrs/gaussian_grid_180_by_360/ic_0001
   ic_0002: gs://vcm-ml-raw-flexible-retention/2024-06-29-C96-SHiELD-AMIP/regridded-zarrs/gaussian_grid_180_by_360/ic_0002
-data_output_directory: gs://vcm-ml-intermediate/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset
+data_output_directory: gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset
 stats:
-  output_directory: gs://vcm-ml-intermediate/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset-stats
-  beaker_dataset: 2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset-stats
+  output_directory: gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset-stats
+  beaker_dataset: 2026-01-28-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset-stats
   start_date: "1940-01-01"
   end_date: "2021-12-31"
   data_type: FV3GFS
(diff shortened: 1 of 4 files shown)


fme/core/models/conditional_sfno/factorizations.py
```diff
@@ -1,260 +0,0 @@
...
-            x = _contract_sep_diagonal(x, weight)
-        elif operator_type == "dhconv":
-            x = _contract_sep_dhconv(x, weight)
-        else:
-            raise ValueError(f"Unkonw operator type {operator_type}")
-    else:
-        if operator_type == "diagonal":
-            x = _contract_diagonal(x, weight)
-        elif operator_type == "dhconv":
-            x = _contract_dhconv(x, weight)
```
mcgibbon, line 211: Note this line is the pathway we use in practice for all runs the past few years.

mcgibbon:
Note the code changes in this PR are tested by the fact that the PR does not update the "output is not changed" regression test in fme/core/models/conditional_sfno/test_sfnonet.py

fme/ace/registry/stochastic_sfno.py
```diff
@@ -145,8 +145,8 @@ class NoiseConditionedSFNOBuilder(ModuleConfig):
     """
 
     spectral_transform: Literal["sht"] = "sht"
-    filter_type: str = "non-linear"
-    operator_type: str = "diagonal"
+    filter_type: Literal["linear", "makani-linear"] = "linear"
+    operator_type: Literal["dhconv"] = "dhconv"
```
AnnaKwa, line 149: Docstring update to note this is the only option

fme/core/models/conditional_sfno/factorizations.py
AnnaKwa: 👍 could you add this note to the PR description for posterity?

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -50,11 +43,36 @@ def _contract_lora(
...
+        x: Complex input tensor of shape
+            (batch_size, in_channels, nlat, nlon)
+    """
     lora_A = torch.view_as_complex(lora_A)
     lora_B = torch.view_as_complex(lora_B)
     return torch.einsum("irx,rox,bixy->boxy", lora_A, lora_B, x)
 
 
+@torch.jit.script
+def _contract_dhconv(
```
AnnaKwa, line 61: This is somewhat different from the `_contract_dhconv` function that the old pathway was using: https://github.com/ai2cm/ace/blob/15ba27b8cd8b4c7f1421dbfb5ced776af102ca7d/fme/core/models/conditional_sfno/contractions.py#L184

Is the regression test passing enough of an assurance that training should be unchanged?

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -134,40 +168,14 @@ def __init__(
...
+        if self.operator_type == "dhconv":
+            self.weight.is_shared_mp = ["matmul", "w"]
         else:
-            assert factorization == "ComplexDense"
-            self.weight = nn.Parameter(scale * torch.randn(*weight_shape, 2))
-            if self.operator_type == "dhconv":
-                self.weight.is_shared_mp = ["matmul", "w"]
-            else:
-                self.weight.is_shared_mp = ["matmul"]
+            self.weight.is_shared_mp = ["matmul"]
```
AnnaKwa: Can remove if/else since there's no other option for `operator_type`

AnnaKwa (commented):
Mostly LGTM, I had a question about the updated contraction function.

fme/ace/registry/stochastic_sfno.py
mcgibbon: Updated.

fme/core/models/conditional_sfno/factorizations.py
mcgibbon: I'm not sure what you mean by "this note" - how should I link to a line of code that's being deleted?

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: view_as_complex/view_as_real is a no-op, it doesn't modify the data. All it does is translate between a `complex` dtype and one that has a `(2,)` final dimension of `float` dtypes.

The regression test IMO is assurance that training shouldn't be unchanged (though note training is not deterministic, already).

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Done.

fme/core/models/conditional_sfno/factorizations.py
AnnaKwa: Not a link, just a mention in the description like how you noted for `filter_type` that the new default had been used all along.

AnnaKwa (approved):

PR #769, opened 2026-01-29
Restore previous attribute propagation behavior in `compute_dataset.py`

### LINKED ISSUE #766: Differences in attribute propagation in data processing with xarray `2025.11.0`
As of xarray version `2025.11.0`, which we happened to upgrade to when bumping the data processing image in https://github.com/ai2cm/full-model/pull/2352, our vertical coarsening method now assigns units of the weights (e.g. `"Pa"`) instead of the units of the variable we are coarsening: https://github.com/ai2cm/ace/blob/e2e45d999506d5464aa3d104a0152b7b53c86308/scripts/data_process/compute_dataset.py#L327-L329  Identified through a `git bisect`, the specific PR in xarray that modified this was https://github.com/pydata/xarray/pull/10726.

The breaking change here was:

> Additionally, when attributes are preserved in binary operations, they now combine attributes from both operands using drop_conflicts (keeping matching attributes, dropping conflicts), instead of keeping only the left operand’s attributes.

This issue affects any dataset that we processed with this workflow (and any downstream workflow, including stats) after updating the image on 2025-12-01:
- `gs://vcm-ml-intermediate/2025-12-02-CM4-like-AM4-random-CO2`
- `gs://vcm-ml-intermediate/2025-12-17-vertically-resolved-c96-1deg-shield-amip-constant-co2-dataset`
- `gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-1deg-c96-shield-som-abrupt-co2-increase-fme-dataset`
- `gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-1deg-c96-shield-som-ensemble-fme-dataset`
- `gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset`

This is straightforward to fix, though it perhaps speaks to having more robust test coverage of our intermediate data processing script.  We should check the attribute behavior on other binary operations within the script as well.

spencerkclark:
This PR forces the use of the pre-xarray-`2025.11.0` attrs propagation behavior in all affected binary operations within the `compute_dataset.py` intermediate data processing script.

To understand the changes, note that prior to xarray `2025.11.0`, xarray would retain the attributes from the left-most object in a sequence of binary operations:

```
>>> xr.set_options(keep_attrs=True)
>>> a = xr.DataArray(0, attrs={"units": "a"})
>>> b = xr.DataArray(1, attrs={"units": "b"})
>>> c = xr.DataArray(2, attrs={"units": "c"})
>>> a + b + c
<xarray.DataArray ()> Size: 8B
array(3)
Attributes:
    units:    a
```

With `2025.11.0`, xarray now uses a `"drop_conflicts"` merge approach in binary operations. This can lead to sometimes counterintuitive behavior in sequences of more than two operations, for example:

```
>>> xr.set_options(keep_attrs=True)
>>> a = xr.DataArray(0, attrs={"units": "a"})
>>> b = xr.DataArray(1, attrs={"units": "b"})
>>> c = xr.DataArray(2, attrs={"units": "c"})
>>> a + b + c
<xarray.DataArray ()> Size: 8B
array(3)
Attributes:
    units:    c
```

In this case, the `"units"` attribute conflicted during the addition of `a` and `b`, which led to it being dropped before being added to `c`.  Since no conflict existed for the `"units"` attribute of `c`, the result of `a + b + c` retains the `"units"` attribute of `c`.

See also the xarray release notes regarding this change [here](https://docs.xarray.dev/en/stable/whats-new.html#id24).

Changes:
- Strategically drop the attributes of xarray objects involved in sequences of binary operations to ensure retention behavior matches that of xarray prior to `2025.11.0`. As far as I can tell there is no option we can set to restore the old behavior automatically—[it was hard coded in xarray](https://github.com/max-sixty/xarray/blob/8e245ccc3e87115603796435da1ba81723b90645/xarray/core/dataset.py#L7685-L7691).
- Adds unit tests, with checks of attribute propagation behavior, to the affected functions.

- [x] Tests added

Resolves #766

### DIFF
--- a/fme/ace/stepper/__init__.py
+++ b/fme/ace/stepper/__init__.py
@@ -4,6 +4,8 @@ from .single_module import (
     StepperConfig,
     StepperOverrideConfig,
     TrainOutput,
+    TrainStepper,
+    TrainStepperConfig,
     load_stepper,
     load_stepper_config,
     process_prediction_generator_list,
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -520,14 +520,6 @@ class StepperConfig:
         default_factory=lambda: DerivedForcingsConfig()
     )  # inference
 
-    @property
-    def train_n_forward_steps_schedule(self) -> TimeLengthSchedule | None:
-        if self.train_n_forward_steps is None:
-            return None
-        if isinstance(self.train_n_forward_steps, TimeLengthSchedule):
-            return self.train_n_forward_steps
-        return TimeLengthSchedule.from_constant(self.train_n_forward_steps)
-
     def __post_init__(self):
         if self.n_ensemble == -1:
             if self.loss.type == "EnsembleLoss":
@@ -539,29 +531,8 @@ class StepperConfig:
     def n_ic_timesteps(self) -> int:
         return self.step.n_ic_timesteps
 
-    def get_train_window_data_requirements(
-        self,
-        default_n_forward_steps: int | None,
-    ) -> DataRequirements:
-        if self.train_n_forward_steps is None:
-            if default_n_forward_steps is None:
-                raise ValueError(
-                    "default_n_forward_steps is required if "
-                    "train_n_forward_steps is not provided"
-                )
-            n_forward_steps: int | IntSchedule = default_n_forward_steps
-        elif isinstance(self.train_n_forward_steps, int):
-            n_forward_steps = self.train_n_forward_steps
-        else:
-            n_forward_steps = self.train_n_forward_steps.max_n_forward_steps
-        requirements = DataRequirements(
-            names=self.all_names,
-            n_timesteps=self._window_steps_required(n_forward_steps),
-        )
-        return self.derived_forcings.update_requirements(requirements)
-
     def get_evaluation_window_data_requirements(
-        self, n_forward_steps: int
+        self, n_forward_steps: int | IntSchedule
     ) -> DataRequirements:
         requirements = DataRequirements(
             names=self.all_names,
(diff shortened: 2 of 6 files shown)


scripts/data_process/compute_dataset.py
```diff
@@ -457,12 +466,20 @@ def compute_pressure_thickness(
...
     output_name: str,
     z_dim: str = "xaxis_1",
+    vertical_coordinate_file_is_local: bool = False,
 ):
     if output_name in ds.data_vars:
         return ds
 
-    with fsspec.open(vertical_coordinate_file) as f:
-        vertical_coordinate = xr.open_dataset(f, decode_timedelta=False).load()
+    if vertical_coordinate_file_is_local:
```
spencerkclark, line 474: This extra option was needed to avoid the test failure, e.g. [in this build](https://github.com/ai2cm/ace/actions/runs/21527773645/job/62035765138), related to h5netcdf / netCDF4 wheel incompatibility issues. Without modifying the way we build our environment, there is no other reliable way to fix this.

scripts/data_process/compute_dataset.py
```diff
@@ -397,12 +401,17 @@ def compute_ocean_fraction(
     land_fraction_name: str,
     sea_ice_fraction_name: str,
 ) -> xr.Dataset:
-    """Compute latent heat flux, if needed."""
+    """Compute ocean fraction, if needed."""
```
spencerkclark, line 404: This was an existing docstring typo, which I take the liberty to fix here.

elynnwu (approved):
Overall looks good to me. As a followup, we should make sure we don't have any of these attribute issues for ocean, sea ice, coupled, or E3SM data processing script. I did a quick scan at `compute_ocean_datset` and nothing obvious jump out. I definitely have some cleanup to do on E3SM side, but let's leave that to a separate PR.

spencerkclark:
Thanks @elynnwu—indeed we should look carefully at the other scripts too.  Hopefully we caught this early enough that datasets were not processed by those scripts with xarray `2025.11.0` yet.

PR #770, opened 2026-01-29
Add apex group norm to deps-only image

AnnaKwa:
This adds the Apex GroupNorm extension to the deps-only image. This increases the image size by 100 Mb, which is pretty small relative to the total ~10Gb size. The apex GroupNorm can be used in downscaling to speed up the forward pass (GroupNorm allows data to be passed in channels_last format which is more performant). 

It also edits the entrypoints for the dep-only image in the dockerfile as the entrypoint script was removed in PR #723.

After approval I will also build the beaker image and edit the tag in `latest_deps_only_image.txt`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,6 @@ build_nsight_image:
 	DOCKER_BUILDKIT=1 docker build --platform=linux/amd64 -f docker/Dockerfile -t $(IMAGE)-nsight:$(VERSION) --target nsight .
 	beaker image create $(IMAGE)-nsight:$(VERSION) --name $(IMAGE)-nsight-$(VERSION) --workspace ai2/ace
 
-build_nsight_image:
 # recommended to deactivate current conda environment before running this
 create_environment:
 	conda create -n $(ENVIRONMENT_NAME) python=3.11 $(CONDA_PACKAGES)


docker/Dockerfile
```diff
@@ -51,9 +51,7 @@ RUN apt-get update && apt-get install -y \
 RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq &&\
 chmod +x /usr/bin/yq
 
-COPY scripts/ci_beaker/entrypoint.sh /entrypoint.sh
```
AnnaKwa, line 54: Script no longer exists

docker/Dockerfile
```diff
@@ -73,3 +71,11 @@ RUN wget https://developer.nvidia.com/downloads/assets/tools/secure/nsight-syste
 
 ENTRYPOINT ["/bin/bash"]
 CMD ["-l"]
+
+
+FROM deps-only AS deps-only-apex
+
+COPY scripts/downscaling/install_apex_groupnorm.sh /tmp/install_apex_groupnorm.sh
```
oliverwm1: I'd put the script in the `docker` directory—can imagine using apex for some optimization unrelated to downscaling.

docker/Dockerfile
```diff
@@ -73,3 +71,11 @@ RUN wget https://developer.nvidia.com/downloads/assets/tools/secure/nsight-syste
 
 ENTRYPOINT ["/bin/bash"]
 CMD ["-l"]
+
+
+FROM deps-only AS deps-only-apex
```
oliverwm1: What's the downside to having apex in the regular `deps-only` image? That would be my preference so that all runs are using a consistent environment—and hopefully with the pinned commit this won't break too often in the future.

AnnaKwa: Not much given the image size increase is small. The one possible downside is that I had some inconsistency in the success of the apex installation in different environments/hardware. The first time I tried to install the build failed because something in the v2 group norm was unhappy being built with a non-B200 gpu. However in the dockerfile image build on my mac it works fine. I think that should mean that apex should build ok as long as this dockerfile is used.

oliverwm1 (approved):
LGTM

PR #771, opened 2026-01-29
Catch cancelled job with exec

elynnwu:
Per PeteW's [comment](https://allenai.slack.com/archives/C6MN19S05/p1769721826162159), use `exec gantry run` instead of `gantry run` so that the workload is cancelled if the GitHub Action job is cancelled.


elynnwu:
Manually triggered [here](https://github.com/ai2cm/ace/actions/runs/21498383560/job/61938870763)

PR #774, opened 2026-01-30
Vectorized generalized version of CRPS

Arcomano1234:
Currently CRPS is hardcoded to only allow 2 ensemble members. This is fine for training where we expect only 2 ensemble members, but in preparation for ensemble metrics we would like a version of CRPS that can take an arbitrary number of ensemble members. 

Changes:
- add almost fair CRPS that accepts n number of ensembles

- [x] Tests added

### DIFF
--- a/fme/ace/data_loading/batch_data.py
+++ b/fme/ace/data_loading/batch_data.py
@@ -198,6 +198,9 @@ class BatchData:
         n_ensemble: int = 1,
     ) -> "BatchData":
         _check_device(data, torch.device("cpu"))
+        if labels is not None:
+            if labels.tensor.device != torch.device("cpu"):
+                raise ValueError(f"labels must be on cpu, got {labels.tensor.device}")
         kwargs = cls._get_kwargs(horizontal_dims)
         if isinstance(labels, list):
             warnings.warn(
@@ -289,7 +292,7 @@ class BatchData:
                 raise ValueError("label_encoding must be provided if labels are used.")
             labels = None
         else:
-            labels = label_encoding.encode(list(sample_labels))
+            labels = label_encoding.encode(list(sample_labels), device="cpu")
         return BatchData.new_on_cpu(
             data=batch_data,
             time=batch_time,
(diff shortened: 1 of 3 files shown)


fme/core/ensemble.py
```diff
@@ -21,19 +21,28 @@ def get_crps(
...
+
+    if n_ens == 1:
+        internal_term = torch.zeros_like(target_term)
+    else:
+        # Pairwise |X_i - X_j|
+        # Shape: [batch, n_ens, n_ens, ...]
+        diffs = torch.abs(gen.unsqueeze(2) - gen.unsqueeze(1))
+
+        # Keep only unique pairs i < j
+        idx = torch.triu_indices(n_ens, n_ens, offset=1, device=gen.device)
```
mcgibbon, line 34: Smart way to do this!

fme/core/ensemble.py
```diff
@@ -21,19 +21,28 @@ def get_crps(
...
+
+    # Term 1: E|X - y|
+    target_term = torch.mean(torch.abs(gen - target), dim=1)
+
+    if n_ens == 1:
+        internal_term = torch.zeros_like(target_term)
+    else:
+        # Pairwise |X_i - X_j|
+        # Shape: [batch, n_ens, n_ens, ...]
+        diffs = torch.abs(gen.unsqueeze(2) - gen.unsqueeze(1))
```
mcgibbon: Suggestion (optional): This will compute the difference between every pair twice. Is it possible to use `idx` below directly with `gen` to compute only the values that go into `pairwise` below?

Another way to say this is, half the values are computed but unread by `pairwise = ` below.

It doesn't matter too much I suppose, but here's the GPT suggestion which looks like it should work: https://chatgpt.com/share/69835fa7-ad40-8011-a073-d8ab327bcd07

Arcomano1234: Yeah I agree. For some reason when I was formulating this I didn't think there was anyway to get around computing them twice but I believe this works.

PR #777, opened 2026-02-02
Reduce duplicated logging code, standardize logs

### LINKED ISSUE #536: LoggingConfig use requires low level logic
These are the specific lines which show up in our training codes that seem like a lot of boilerplate / could maybe have some features pulled into a helper:

```
    config.logging.configure_logging(config.experiment_dir, log_filename="out.log")
    env_vars = logging_utils.retrieve_env_vars()
    logging_utils.log_versions()
    beaker_url = logging_utils.log_beaker_url()
    config_as_dict = to_flat_dict(dataclasses.asdict(config))
    config.logging.configure_wandb(
        config=config_as_dict, env_vars=env_vars, resume=True, notes=beaker_url
    )
```

Some of this could be reduced by pulling more into the shared LoggingConfig class. There's a lot of logic piping outputs of LoggingConfig methods into inputs to LoggingConfig methods.

mcgibbon:
This PR moves low-level logging logic into the configure_logging helper, which now configures all types of logging including wandb.

All entrypoints now log all of the information we log at the start of runs.

Changes:
- `LoggingConfig.configure_logging` now takes in `config` and `resumable` arguments for configuring WandB.
- Reduced duplicated logging code in train and inference entrypoints.

- [ ] Tests added

Resolves #536

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,3 +46,18 @@ committing code: `pre-commit install`.
 
 You can run local tests with `pytest`. As a shortcut for running only faster unit
 tests, use `make test_very_fast` from the root of the repository.
+
+## Internal Development
+
+If you are making changes directly to ai2cm/ace, please follow these internal
+development guidelines.
+
+When making new branches, use the naming convention:
+`<type>/<short-description>`, where `<type>` is one of:
+- feature: Any new functionality in the repository including workflows and scripts but not including configurations. Should be the "default" type if it's unclear which to use.
+- refactor: No changes to features, just code restructuring or simplification.
+- fix: Bug fixes.
+- exp: Branch used for running experiments that is likely not intended to merge with mainline code. Functionality changes in these branches should first be PR'd using a feature branch.
+- config: Changes to baseline and experimental configurations under config/.
+- scripts: Changes isolated to a single script under scripts/ and subject to less rigorous review than changes to core code.
+- docs: Documentation changes only.
--- a/scripts/data_process/configs/CM4-1pctCO2-140yr-coupled.yaml
+++ b/scripts/data_process/configs/CM4-1pctCO2-140yr-coupled.yaml
@@ -8,7 +8,10 @@ coupled_datasets:
   coupled_sea_surface:
     precomputed_sea_ice_mask:
       zarr_path: /climate-default/2025-10-20-cm4-piControl-200yr-no-smoothing-coupled-ocean.zarr
-    ocean_extra_masked_names: ["HI", "sea_ice_volume"]
+    ocean_extra_masked_names: ["HI", "sea_ice_volume", "UI", "VI", "sfdsi"]
+    ocean_extra_fill_values:
+      UI: 0.0
+      VI: 0.0
     surface_flux_window_avg:
       window_timedelta: 120h
       first_timestamp: "0211-01-03T12:00:00"
--- a/scripts/data_process/configs/CM4-piControl-200yr-coupled.yaml
+++ b/scripts/data_process/configs/CM4-piControl-200yr-coupled.yaml
@@ -7,7 +7,7 @@ coupled_datasets:
     how: interpolate_sst
   coupled_sea_surface:
     sst_threshold: 291.75
-    ocean_extra_masked_names: ["HI", "sea_ice_volume"]
+    ocean_extra_masked_names: ["HI", "sea_ice_volume", "UI", "VI", "sfdsi"]
     surface_flux_window_avg:
       window_timedelta: 120h
       first_timestamp: "0151-01-03T12:00:00"
--- a/scripts/data_process/coupled_dataset_utils.py
+++ b/scripts/data_process/coupled_dataset_utils.py
@@ -240,6 +240,9 @@ class CoupledSeaSurfaceConfig:
             copies of the mask will be added as "mask_{extra_name}". NOTE: The
             sea-ice fraction and "ocean" sea-ice fraction masks don't need to be
             specified here.
+        ocean_extra_fill_values: Dictionary where keys are a subset of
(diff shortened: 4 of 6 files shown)


fme/downscaling/inference/inference.py
```diff
@@ -226,13 +226,9 @@ class InferenceConfig:
...
-        self.logging.configure_logging(self.experiment_dir, log_filename)
-
-    def configure_wandb(self, resumable: bool = False, **kwargs):
-        config = to_flat_dict(asdict(self))
-        env_vars = logging_utils.retrieve_env_vars()
-        self.logging.configure_wandb(
-            config=config, env_vars=env_vars, resumable=resumable, **kwargs
+        config = to_flat_dict(dataclasses.asdict(self))
+        self.logging.configure_logging(
+            self.experiment_dir, log_filename, config=config, resumable=True
```
jpdunc23:
```suggestion
            self.experiment_dir, log_filename, config=config, resumable=False
```

fme/downscaling/evaluator.py
```diff
@@ -196,13 +195,9 @@ class EvaluatorConfig:
...
     def configure_logging(self, log_filename: str):
-        self.logging.configure_logging(self.experiment_dir, log_filename)
-
-    def configure_wandb(self, resumable: bool = False, **kwargs):
         config = to_flat_dict(dataclasses.asdict(self))
-        env_vars = logging_utils.retrieve_env_vars()
-        self.logging.configure_wandb(
-            config=config, env_vars=env_vars, resumable=resumable, **kwargs
+        self.logging.configure_logging(
+            self.experiment_dir, log_filename, config=config, resumable=True
```
jpdunc23:
```suggestion
            self.experiment_dir, log_filename, config=config, resumable=False
```

fme/core/logging_utils.py
```diff
@@ -54,7 +55,34 @@ class LoggingConfig:
     def __post_init__(self):
         self._dist = Distributed.get_instance()
 
-    def configure_logging(self, experiment_dir: str, log_filename: str):
+    def configure_logging(
+        self,
+        experiment_dir: str,
+        log_filename: str,
+        config: Mapping[str, Any],
+        resumable: bool = False,
```
jpdunc23: The default value for `resumable` was changed from `True` to `False`. Please double check that `resumable=True` is explicitly passed everywhere `configure_wandb` was previously called without the `resumable` argument.

fme/ace/inference/evaluator.py
```diff
@@ -162,14 +161,9 @@ def __post_init__(self):
                     )
 
     def configure_logging(self, log_filename: str):
-        self.logging.configure_logging(self.experiment_dir, log_filename)
-
-    def configure_wandb(
-        self, env_vars: dict | None = None, resumable: bool = False, **kwargs
-    ):
         config = to_flat_dict(dataclasses.asdict(self))
```
jpdunc23: Should remove `to_flat_dict` here (and other inference entrypoints) as was done in the training scripts?

fme/core/logging_utils.py
```diff
@@ -123,10 +151,24 @@ def configure_wandb(
             experiment_dir=experiment_dir,
             resumable=resumable,
             dir=wandb_dir,
-            **kwargs,
```
jpdunc23, line 126: `**kwargs` is not longer used. Do we need it?

jpdunc23 (approved):
Thanks for the nice cleanup. I noticed a couple of minor inconsistencies and an unused arg. Otherwise, LGTM.

fme/ace/inference/evaluator.py
mcgibbon: This was actually a bug - I hadn't added to_flat_dict to the wandb helper. But it's a good idea, I added it and removed these calls.

fme/core/logging_utils.py
mcgibbon: Reverted the change back to default True.

fme/core/logging_utils.py
mcgibbon: Removed.

fme/downscaling/inference/inference.py
mcgibbon: While the default to this helper was `False`, there was a hard-code `resumable=True` where it's called on `generation_config.configure_wandb` below, so I think `resumable=True` is no change in behavior.

Can you check and make sure you agree?

fme/downscaling/evaluator.py
mcgibbon: Similarly, `resumable=True` is hard-coded where this is called in practice. Not sure if this was intended, strange to have it for an inference-like entrypoint.

jpdunc23: Yes, I see from the diff that in `main()` we had `evaluator_config.configure_wandb(resumable=True, notes=beaker_url)`. So I understand you're preserving that behavior, but why not add a `resumable: bool = False` default arg to `EvaluatorConfig.configure_logging` and pass `resumable=True` when calling `evaluator_config.configure_logging` below in `main()`? This would preserve the existing behavior of `EvaluatorConfig`.

fme/downscaling/inference/inference.py
jpdunc23: Same comment as https://github.com/ai2cm/ace/pull/777/changes#r2835488269

fme/downscaling/evaluator.py
mcgibbon: When I passed through this file the existing code was quite confusing - why is a default set that never gets used? I expected we run this with resumable=False based on the config method, but then saw it's actually resumable=True from the code below. I updated it to the style used in the training scripts as a cleanup.

I could split this out into a separate PR? I'll do that on Monday.

mcgibbon: Reverted.

PR #779, opened 2026-02-02
Vendorize song unet v2

AnnaKwa:
This adds a newer version of the SongUNet from NVIDIA's physicsnemo repo which incorporates various speedups since the original version we vendorized. Notably it can use the apex GroupNorm which can work with channels_last memory format, which is more performant. 

I'll break this into multiple PRs as the diff has gotten big.

- [x] moves existing vendorized physicsnemo unet and layer code into `fme/downscaling/modules/physicnemo_unets_v1` to keep things organized. 
- [x] add vendorized code for new unet into `physicnemo_unets_v2/unets.py, group_norm.py, layers.py`.  There are a ton of mypy complaints so mypy is instructed to ignore this dir. 
- [x] add `channels_last` init arg to `UNetDiffusionModule`. If This defaults to False to preserve the current behavior, and is only passed when `SongUNetv2` is built with config set to `use_apex_gn`. When True, the input and conditioning are converted to channels_last format in the forward pass.
- [x] minor update to dtype cast in `samplers.py` to allow tests to run on macs (mps does not work with float64)
- 

- [x] Tests added. However, the CI unit tests use pip install to set up the dependencies, while the apex installation needed to test the channels_last format is more involved (the dockerfile runs `docker/install_apex_groupnorm.sh` to install and it is kind of slow). Currently this test is skipped if apex is not found. I manually ran the test on a GPU machine with apex installed to check that it passed.

### DIFF
--- a/fme/downscaling/modules/unet_diffusion.py
+++ b/fme/downscaling/modules/unet_diffusion.py
@@ -20,17 +20,20 @@ class UNetDiffusionModule(torch.nn.Module):
         unet: The U-Net model.
         use_amp_bf16: use automatic mixed precision casting to bfloat16
             during forward pass
+        channels_last: whether to use channels last memory format. 
+            Default is False.
     """
 
     def __init__(
         self,
         unet: torch.nn.Module,
         use_amp_bf16: bool = False,
+        channels_last: bool = False,
     ):
         super().__init__()
         self.unet = unet.to(device=get_device())
         self.use_amp_bf16 = use_amp_bf16
-
+        self.channels_last = channels_last
         if self.use_amp_bf16:
             if get_device().type == "mps":
                 raise ValueError("MPS does not support bfloat16 autocast.")
@@ -55,10 +58,11 @@ class UNetDiffusionModule(torch.nn.Module):
             noise_level: The noise level of each example in the batch.
         """
         device = get_device()
-        latent = latent.to(device)
-        conditioning = conditioning.to(device)
+        memory_format = torch.channels_last if self.channels_last else None
+        latent = latent.to(device, memory_format=memory_format)
+        conditioning = conditioning.to(device, memory_format=memory_format)
         noise_level = noise_level.to(device)
-
+        
         with self._amp_context:
             result = self.unet(
                 latent,
@@ -66,5 +70,4 @@ class UNetDiffusionModule(torch.nn.Module):
                 sigma=noise_level,
                 class_labels=None,
             )
-
         return result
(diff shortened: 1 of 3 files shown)


fme/downscaling/modules/vendorized/unets_v2.py
```diff
@@ -0,0 +1,1571 @@
...
+                init_attn=init_attn,
+                init=init,
+                use_apex_gn=use_apex_gn,
+                amp_mode=amp_mode,
+                fused_conv_bias=fused_conv_bias,
+            )
+        else:
+            self.attn = None
+        # A hook to migrate legacy attention module
+        # Commented out after vendorizing because we do not attempt to load
```
AnnaKwa, line 882: This was causing an AttributeError due to `register_load_state_dict_pre_hook` not found (torch version difference maybe?) We do not attempt to load legacy checkpoints with this new module so I just commented it out.

fme/downscaling/modules/vendorized/unets_v2.py
```diff
@@ -0,0 +1,1571 @@
...
+                        f"Checkpoint contains both legacy and new keys for {old_key}"
+                    )
+
+
+# ------------------------------------------------------------------------------
+# Backbone architectures
+# ------------------------------------------------------------------------------
+
+
+class SongUNetv2(torch.nn.Module):
```
AnnaKwa, line 972: Changed to inherit from `torch.nn.Module` instead of the physicsnemo `Module`. This is the same thing that was done with the existing `SongUNet`.

fme/downscaling/modules/unet_diffusion.py
```diff
@@ -54,10 +58,16 @@ def forward(
             noise_level: The noise level of each example in the batch.
         """
         device = get_device()
+        memory_format = torch.channels_last if self.channels_last else None
+        latent = latent.to(device, memory_format=memory_format)
```
AnnaKwa, line 62: The `SongUNetv2` does have a check to convert x to channels_last in its forward pass, but the conversion is required at this level otherwise operations in the preconditioner covert the unet output back to contiguous format.

fme/downscaling/samplers.py
```diff
@@ -105,7 +105,8 @@ def stochastic_sampler(
         )
 
     # Time step discretization.
-    step_indices = torch.arange(num_steps, dtype=torch.float64, device=latents.device)
+    compute_dtype = torch.float32 if latents.device.type == "mps" else torch.float64
```
AnnaKwa, line 108: Changes in this module were added to allow tests to run on macbooks.

PR #783, opened 2026-02-02
Initialize label tensor on CPU during dataloading

Arcomano1234:
Small PR to address the issue where previously during dataloading the label tensor was initialized on device. This was problematic because torch does not allow you to initialize tensors on GPUs from a forked subprocess dataloader worker leading to an error
`RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`. Here is an example of this failing ([beaker log](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KGDERZKZZV1NQGYR21WYK040/logs?jobId=01KGDES02CT4SX6W28ZDA56F7F))

Changes:
- add device argument to `BatchLabels` `encode` property to ensure we initialize on CPU during dataloading calls.

- [x] Tests added


fme/ace/data_loading/batch_data.py
```diff
@@ -289,7 +289,7 @@ def from_sample_tuples(
                 raise ValueError("label_encoding must be provided if labels are used.")
             labels = None
         else:
-            labels = label_encoding.encode(list(sample_labels))
+            labels = label_encoding.encode(list(sample_labels), device="cpu")
```
mcgibbon, line 295: Suggestion (optional): Also add a check-and-raise inside `BatchData.new_on_cpu` that ensures labels are on the CPU.

Arcomano1234: In retrospect this bug would have been obvious if we did have a check for labels in `BatchData.new_on_cpu` so I added check.

PR #784, opened 2026-02-03
Use exec gantry run in GH action

oliverwm1:
@epwalsh recommended this so that the beaker workload gets cancelled if the GH action job is canceled.


oliverwm1:
closing since duplicate of #771

PR #786, opened 2026-02-03
Initialize conditional sfno weights to reasonably large value

mcgibbon:
Currently the conditional SFNO initializes spectral filter weights to stdev of `1/(n_in * n_out)` instead of the expected `1/sqrt(n_in)` to maintain variance. This leads to behaviors similar to zero-initializing these weights. The Conv1x1 layers also use a hard-coded 0.02 stdev initialization, instead of `1/sqrt(n_in)` which is the pytorch default.

This PR rectifies these behaviors, using initialization more similar to what's done in FCN3 and common in ML literature.

Changes:
- NoiseConditionedSFNO now initializes weights to maintain unit variance, leading to larger initial weight variance.

- [ ] Tests added


fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -168,6 +166,11 @@ def __init__(
             self.lpad = 0
             self.mpad = 0
 
+        if scale == "auto":
+            scale = math.sqrt(1 / (in_channels)) * torch.ones(self.modes_lat_local, 2)
```
mcgibbon, line 170: This is meant to replicate the behavior at https://github.com/NVIDIA/makani/blob/e59c2f4344764d1378c6b00615bf56b664c653da/makani/models/common/spectral_convolution.py#L93

Note in this PR grouped convolutions are not implemented, so num_groups is not used.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -826,7 +826,6 @@ def __init__(
         if self.pos_embed:
             self.pos_embed = get_pos_embed()
 
-        self.apply(self._init_weights)
```
mcgibbon, line 831: Note the lack of a similar call in https://github.com/NVIDIA/makani/blob/e59c2f4344764d1378c6b00615bf56b664c653da/makani/models/networks/fourcastnet3.py#L386

`reset_parameters` is already called automatically in `nn.Module.__init__`, so all this removal should do is avoid setting Conv2d weights to a hard-coded 0.02.

fme/core/models/conditional_sfno/testdata/test_sfnonet_output_is_unchanged.pt
mcgibbon, line 1: We updated the initialization, so we expect this regression target to change.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -202,7 +205,7 @@ def __init__(
             self.lora_scaling = 0.0
 
         if bias:
-            self.bias = nn.Parameter(scale * torch.zeros(1, out_channels, 1, 1))
+            self.bias = nn.Parameter(torch.zeros(1, out_channels, 1, 1))
```
mcgibbon, line 208: Scale was doing nothing here (these are zeros), but now that it's a non-scalar it causes an exception if we keep it.

PR #788, opened 2026-02-03
Enable grouped spectral convolution

mcgibbon:
This PR enables the use of grouped spectral convolutions for the "linear" SFNO filter type.

Changes:
- Make use of `num_groups` when `filter_type="linear"` is used in conditional SFNO

- [ ] Tests added


mcgibbon:
Note no configuration changes in this PR as the "makani-linear" filter already supports groups, so the configuration option already exists (it's just currently ignored for the "linear" filter we currently use).

fme/core/models/conditional_sfno/test_s2convolutions.py
```diff
@@ -1,8 +1,93 @@
...
+
+
+@pytest.mark.skipif(
+    get_device().type != "cuda",
+    reason=(
+        "This test is only relevant for CUDA since "
+        "it's testing speed of DHConv groups on GPU."
+    ),
+)  # noqa: E501
+def test_contract_dhconv_groups_are_faster():
```
mcgibbon, line 56: Note memory usage is not actually less. The weights are much smaller, but it turns out the state vector x is much larger than the weights, for basically every configuration I could come up with at 1-degree resolution (due mainly to the 360-dim not included in the weights)

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -48,14 +47,15 @@ def _contract_lora(
     Performs LoRA update contraction.
 
     Args:
+
         lora_A: LoRA A matrix of shape (in_channels, rank, nlat, 2)
```
Arcomano1234: The sizes in the documentation should change right? E.g., `lora_A: LoRA A matrix of shape (num_group, in_channels, rank, nlat, 2)`

mcgibbon: Good catch, updated.

Arcomano1234: Did you push this commit? I don't see any changes

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -171,31 +181,36 @@ def __init__(
...
         self.weight.is_shared_mp = ["matmul", "w"]
 
         if lora_rank > 0:
-            if self.weight.shape != (
-                in_channels,
-                out_channels,
-                self.modes_lat_local,
-                2,
-            ):
-                raise NotImplementedError(
```
Arcomano1234, line 187: Isn't this still true? Unless I don't understand what unpadded weights mean in the context of dhconv

fme/core/models/conditional_sfno/test_s2convolutions.py
Arcomano1234: Yeah this makes sense. I suspect this might change if we move to 0.25-degree

fme/core/models/conditional_sfno/test_s2convolutions.py
```diff
@@ -1,8 +1,93 @@
...
+    w_grouped = torch.randn(
+        G, C // G, C // G, H, 2, dtype=torch.float32, device=get_device()
+    )
+
+    def contract_grouped():
+        return _contract_dhconv(x_grouped, w_grouped)
+
+    grouped_result = benchmark(contract_grouped)
+
+    assert grouped_result.ms_per < 2 / G * ungrouped_result.ms_per, (
```
Arcomano1234, line 80: I am not familiar with the kernals that support `_contract_dhconv` nor do I have any experience with timings of this function, but I do worry that this timing test might end up being device specific or subject to change in the future when we change torch-harmonic or pytorch versions.

requirements.txt
```diff
@@ -11,6 +11,7 @@ netcdf4
 numpy
 omegaconf>=2.1.0
 pandas
+pkg_resources
```
Arcomano1234: Is this a previous bug getting fixed, because I see no reason this PR needed this new requirement?

mcgibbon: Yes, it is.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: padded weights were never implemented in our version of the code, there were some commented-out blocks of code for it with only partial implementations. I don't know why padded weights existed, I think as a way to try to performance-optimize the FFT?

fme/core/models/conditional_sfno/test_s2convolutions.py
mcgibbon: At 0.25-degree the disparity will only be greater, because the state vector is a factor of n_lon larger than the weights.

fme/core/models/conditional_sfno/test_s2convolutions.py
mcgibbon: The theoretical improvement is something like 1/G, 2/G seems to be a fairly generous bound. But yes we should be open to retuning this in the future.

fme/core/models/conditional_sfno/test_s2convolutions.py
Arcomano1234: Oh yeah thats very true

fme/core/models/conditional_sfno/s2convolutions.py
Arcomano1234: Okay just making sure I understood why we had this before. Sounds like removing that is fine, but I didn't know if it was related to things from this PR

PR #789, opened 2026-02-03
SHiELD AMIP p2K, 4K data process

brianhenn:
Adds configuration for processing the 1- and 4-deg SHiELD-AMIP perturbed SST (+2K, +4K) runs.


scripts/data_process/Makefile
```diff
@@ -61,7 +61,7 @@ SHIELD_RES ?= c96
 .PHONY: shield_AMIP_dataset
 shield_AMIP_dataset:
 	./compute_dataset.sh \
-	--config configs/shield-amip-ensemble-$(SHIELD_RES)-$(RESOLUTION)-8layer.yaml
+	--config configs/shield-amip-$(SHIELD_RES)-$(RESOLUTION)-8layer.yaml
```
spencerkclark, line 64: Is the idea that `SHiELD_RES` may now encode more information than just the spatial resolution?

brianhenn: Yeah, this was a bit of hack to make it possible to call the SHiELD-AMIP configs we already had plus the new ones.

PR #791, opened 2026-02-03
E3SMv3 AMIP preprocess as ERA5 naming standard

elynnwu:
This PR updates E3SMv3 AMIP preprocessing to follow ERA5 standard, which includes variable renaming, adding CO2 concentration, and updating topography.


scripts/data_process/compute_dataset_e3smv2.py
```diff
@@ -178,6 +179,7 @@ def open_dataset(
                     datasets[varname] = ds
         del var_paths["time_invariant"]
     for varname, paths in var_paths.items():
+        print(varname)
```
Arcomano1234: Is this print statement intentional? Probably don't need it

elynnwu:
Good point - it was a ramped run from Finn before, I updated the configs.

PR #792, opened 2026-02-03
Handle diagnostic pre-weighted `hfds` in the OHC corrector

jpdunc23:
Adds handling in the ocean corrector for models that generate `hfds_total_area`, a version of `hfds` that is weighted by `sea_surface_fraction` (see https://github.com/ai2cm/ace/pull/798).

Changes:
- Adds the `net_downward_surface_heat_flux_total_area` property to `OceanData`.
- `hfds` is be derived from `hfds_total_area` when the latter is present, similar to the way in which `sea_ice_fraction` is derived from `ocean_sea_ice_fraction`.

- [x] Tests added


fme/core/ocean_derived_variables.py
```diff
@@ -176,3 +176,12 @@ def sea_ice_fraction(
...
     """Compute the sea ice fraction."""
     return data.sea_ice_fraction
+
+
+@register(VariableMetadata("W/m**2", "Surface ocean heat flux"), exists_ok=True)
+def hfds(
+    data: OceanData,
+    timestep: datetime.timedelta,
+) -> torch.Tensor:
+    """Compute the sea ice fraction."""
```
elynnwu: Update doc here

fme/core/ocean_data.py
```diff
@@ -159,7 +160,21 @@ def sea_surface_fraction(self) -> torch.Tensor:
...
+                self.net_downward_surface_heat_flux_total_area
+                / self.sea_surface_fraction
+            )
+
+    @property
+    def net_downward_surface_heat_flux_total_area(self) -> torch.Tensor:
+        """Net heat flux downward across the ocean surface (below the sea-ice),
+        normalized by total grid cell area.
+        """
+        return self._get("net_downward_surface_heat_flux_by_total_area")
```
elynnwu: [nit] inconsistent naming, maybe just drop `by`?

jpdunc23: Good catch, thanks!

PR #793, opened 2026-02-03
Add EnsembleMean Aggregator 

Arcomano1234:
To break up #709 into manageable PRs, this PR adds the ensemble mean aggregator to `one_step/ensemble.py`. 
Changes:
- Add `EnsembleMeanRMSEMetric` class `one_step/ensemble.py` 

- [x] Tests added


fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -73,6 +73,30 @@ def get(self) -> torch.Tensor:
...
+        if self._total_rmse is None:
+            self._total_rmse = rmse
+        else:
+            self._total_rmse += rmse
+        self._n_batches += 1
+
+    def get(self) -> torch.Tensor:
+        if self._total_rmse is None:
+            raise ValueError("No batches have been recorded.")
+        return self._total_rmse / self._n_batches
```
mcgibbon, line 97: Issue: I don't see a `dist.reduce_mean` call anywhere, you'll need one for the result to use data from all ranks. This may affect the other metrics - you can fix it in another PR.

Arcomano1234: I do not believe this is a bug because we call `dist.reduce_mean` outside the individual ensemble metric classes. E.g., in `_EnsembleAggregator` we call 
`metric_value = self._dist.reduce_mean(
                    self._variable_metrics[metric][key].get()
                )`

mcgibbon: Agreed, thanks!

PR #794, opened 2026-02-04
Script for time coarsening existing training datasets

mcgibbon:
This PR adds `scripts/time_coarsen` for time-coarsening existing training datasets. This will allow us to test the ability of our model to run with 1-day and possibly 5-day timesteps.

Note stats will also need to be re-generated following time-coarsening. This can be run separately or later incorporated into this workflow.

Changes:
- Added `scripts/time_coarsen` for time-coarsening training datasets

- [x] Tests added


scripts/time_coarsen/run.sh
```diff
@@ -0,0 +1,45 @@
+
+#!/bin/bash
+
+set -e
+
+SCRIPT_PATH=$(git rev-parse --show-prefix)  # relative to the root of the repository
```
oliverwm1, line 6: I think these dataset processing scripts should run on GCP, e.g. using argo https://github.com/ai2cm/ace/blob/da99cfc8435b82c41877cf03de5887f695f7161b/scripts/data_process/compute_dataset.sh#L31

scripts/time_coarsen/time_coarsen.py
```diff
@@ -0,0 +1,91 @@
...
+    ds_snapshot = ds[config.snapshot_names].isel(
+        time=slice(config.coarsen_factor - 1, None, config.coarsen_factor)
+    )
+    ds_window = (
+        ds[config.window_names]
+        .coarsen(time=config.coarsen_factor, boundary="trim")
+        .mean()
+        .drop("time")
+    )  # use time of snapshots
+    # ds_coarsened = ds_snapshot#xr.merge([ds_snapshot, ds_window, ds_coords])
```
spencerkclark:
```suggestion
```

scripts/time_coarsen/time_coarsen.py
```diff
@@ -0,0 +1,91 @@
...
+    ds_coarsened = xr.merge([ds_snapshot, ds_window, ds_constants])
+    ds_coarsened.to_zarr(pair.output)
```
spencerkclark: Could optionally include a `history` attribute which gives some provenance information about the script used to produce the dataset, as well as the configuration.

scripts/time_coarsen/test_time_coarsen.py
```diff
@@ -0,0 +1,72 @@
...
+    config = Config(
+        paths=[PathPair(input=input_path, output=output_path)],
+        coarsen_factor=2,
+        snapshot_names=["temp"],
+        window_names=["temp_tendency"],
+    )
+    main(config)
+
+    # Read back the coarsened dataset
+    ds_coarsened = xr.open_zarr(output_path)
```
spencerkclark, line 92: Based on looking at the resulting zarr stores on WEKA, it appears that it happened to work, but it would be good to also test that the `"chunks"` and `"shards"` encoding for each variable matched that of the input zarr.

spencerkclark (commented):
The core mechanics of this script look good to me.  I'm happy to review again if/when we decide to move to running this on GCP instead of Beaker.

scripts/time_coarsen/run.sh
mcgibbon: Is this a comment that the script should also work when using GCP, or is it a request to refactor the workflow to use GCP? If it's a request, is it blocking? And would that mean these scripts should be moved into `data_process` since they would then use the same infrastructure?

I had decided to go this route because we already have the datasets on weka, I found this more intuitive to launch and debug (no need to rebuild an image to test new package code), and it avoids an extra copy of the dataset in the cloud. If the idea is we would rather not use beaker CPU resources for dataset computation, or that we want a copy of these datasets to exist in the cloud, then I think it would make sense to refactor it.

oliverwm1: My preference would be to refactor workflow to use GCP. This would improve consistency with other data processing workflows and avoid using GPU cluster for data processing. Also, having a copy of the daily dataset in the cloud is a good thing (space on weka is more limited, so we regularly have to delete datasets there). But not a blocking request.

scripts/time_coarsen/time_coarsen.py
mcgibbon: Be harder on me :P I should definitely include a history attribute! You put in a lot of work making sure we have it in other scripts, we should be consistent.

scripts/time_coarsen/time_coarsen.py
mcgibbon: Done.

scripts/time_coarsen/test_time_coarsen.py
mcgibbon: Done.

scripts/time_coarsen/test_time_coarsen.py
```diff
@@ -0,0 +1,113 @@
...
+
+    encoding = {}
+    for name, da in ds.variables.items():  # includes coords + data_vars
+        if da.ndim == 0:
+            continue
+        chunks = tuple(
+            2 if dim == "time" else 2 if dim == "lat" else 4 if dim == "lon" else 1
+            for dim in da.dims
+        )
+        encoding[name] = {"chunks": chunks, "compressor": compressor}
```
spencerkclark: Probably if `"chunks"` is propagated, `"shards"` will be too, but since propagating both will be the primary use-case of this script, I think it would be good to write the test with both:
```suggestion
        chunks = tuple(
            2 if dim == "time" else 2 if dim == "lat" else 4 if dim == "lon" else 1
            for dim in da.dims
        )
        shards = tuple(
            4 if dim == "time" else 2 if dim == "lat" else 4 if dim == "lon" else 1
            for dim in da.dims
        )
        encoding[name] = {"chunks": chunks, "compressor": compressor, "shards": shards}
```

scripts/time_coarsen/run.sh
```diff
@@ -0,0 +1,45 @@
...
+
+  # Extract additional args from config header
+  local extra_args=()
+  while IFS= read -r line; do
+    [[ "$line" =~ ^#\ arg:\ (.*) ]] && extra_args+=(${BASH_REMATCH[1]})
+  done < "$CONFIG_PATH"
+
+  gantry run \
+    --name "$job_name" \
+    --description 'Run ACE training' \
```
spencerkclark:
```suggestion
    --description 'Run ACE dataset time coarsening' \
```

PR #795, opened 2026-02-04
Move existing vendorized physicsnemo unets code into subdir

AnnaKwa:
This PR just moves the existing vendorized SongUNet and associated layer modules into `fme/downscaling/modules/physicsnemo_unets_v1`. This will make a future PR adding more vendorized code easier to review and keep the modules dir tidy. There are no actual code changes.

### DIFF
--- a/fme/downscaling/samplers.py
+++ b/fme/downscaling/samplers.py
@@ -104,8 +104,10 @@ def stochastic_sampler(
             f"{img_lr.shape[0]} vs {latents.shape[0]}."
         )
 
+    compute_dtype = torch.float32 if latents.device.type == "mps" else torch.float64
+
     # Time step discretization.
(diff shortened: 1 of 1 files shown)


AnnaKwa:
> With that in mind, unless there's a purpose of the `vendorized/` directory that I'm missing, I'd recommend naming this directory something that reflects where the data came from, like `downscaling/modules/physicsnemo` or `downscaling/modules/physicsnemo_diffusion`.

That sounds good. In the bigger PR I was having issues with a large number of mypy errors on the newer vendorized code so I moved it into a subdir that I told mypy to ignore. Then I figured that for consistency the rest of the older vendorized modules should go in there as well. I agree though that it would be preferable to separate the modules that come from different sources.

AnnaKwa:
Updated to move just the unet and related modules into a subdir, so they are clearly differentiated from the later v2 SongUNet and new layers that will be added in an upcoming PR.

PR #796, opened 2026-02-04
Fix: downscaling sampler dtype on MPS

AnnaKwa:
The updated vendorized sampler code ([Vendorize Apache 2.0 Nvidia Downscaling Code](https://github.com/ai2cm/ace/pull/748/changes#top)) casts to float64, which is not allowed when running on MPS and thus breaks a lot of tests when running locally on macs. This PR updates the cast to use float32 on MPS.

### DIFF
--- a/fme/downscaling/samplers.py
+++ b/fme/downscaling/samplers.py
@@ -104,8 +104,10 @@ def stochastic_sampler(
             f"{img_lr.shape[0]} vs {latents.shape[0]}."
         )
 
+    compute_dtype = torch.float32 if latents.device.type == "mps" else torch.float64
+
     # Time step discretization.
-    step_indices = torch.arange(num_steps, dtype=torch.float64, device=latents.device)
+    step_indices = torch.arange(num_steps, dtype=compute_dtype, device=latents.device)
     t_steps = (
         sigma_max ** (1 / rho)
         + step_indices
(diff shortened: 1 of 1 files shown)


fme/downscaling/samplers.py
```diff
@@ -104,8 +104,10 @@ def stochastic_sampler(
             f"{img_lr.shape[0]} vs {latents.shape[0]}."
         )
 
+    compute_dtype = torch.float32 if latents.device.type == "mps" else torch.float64
```
mcgibbon, line 107: Question (optional): Do we need torch.float64 at all here? It could be worth using torch.float32 in all cases, especially if we've been doing it until now. I'm in particular confused how using float64 precision can ever help if the net is predicting its output in float32, it's not like `x_next` and `denoised` have wildly differing orders of magnitude where we'd need that kind of precision. Maybe something to ask Nvidia.

AnnaKwa: I was also curious about that; I agree it would be something to ask Nvidia.

AnnaKwa: Cursor thinks it does make a difference in the numerical precision for the lower end of the noise. I wonder if this might be why Troy saw different behavior for nonzero churn values when he tested the newer vendorized code.

AnnaKwa: Looks like (yet another) version of the physicsnemo sampler in the `fme/diffusion` code notes this is done for numerical integration stability

PR #797, opened 2026-02-04
Add vendorized physicsnemo `SongUnetv2`

AnnaKwa:
This PR adds the vendorized modules needed for the newer physicsnemo SongUNet into `fme/downscaling/modules/physicsnemo_unets_v2`. This version allows the use of the apex group norm layer which can speed up the unet by enabling the use of channels_last format throughout. The few places where I edited or added to the vendorized code are marked with comments.

Some helper functions that validate unet and input shapes are moved out of `fme/downscaling/modules/physicsnemo_unets_v1/unets.py` into ``fme/downscaling/modules/utils.py` so the new unet can also perform these checks.

There were ~90 mypy errors in the vendorized modules so I marked that subdir for mypy to ignore.

- [x] Tests added


fme/downscaling/modules/physicsnemo_unets_v2/layers.py
```diff
@@ -0,0 +1,920 @@
...
+                use_apex_gn=use_apex_gn,
+                amp_mode=amp_mode,
+                fused_conv_bias=fused_conv_bias,
+            )
+        else:
+            self.attn = None
+        # A hook to migrate legacy attention module
+        # Commented out after vendorizing because we do not attempt to load
+        # legacy checkpoints, this is only used in SongUNetv2
+        # self.register_load_state_dict_pre_hook(self._migrate_attention_module)
```
AnnaKwa, line 840: This was causing an AttributeError  `register_load_state_dict_pre_hook not found `(torch version difference maybe?) We do not attempt to load legacy checkpoints with this new module so I just commented it out.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -0,0 +1,695 @@
...
+
+    return property(_getter, _setter, doc=doc)
+
+
+# ------------------------------------------------------------------------------
+# Backbone architectures
+# ------------------------------------------------------------------------------
+
+
+class SongUNetv2(torch.nn.Module):
```
AnnaKwa, line 86: Changed to inherit from torch.nn.Module instead of the physicsnemo Module. This is the same thing that was done with the existing SongUNet in `physicsnemo_unets_v1`
.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -0,0 +1,695 @@
...
+                    out_channels=out_channels,
+                    kernel=3,
+                    fused_conv_bias=True,
+                    **init_zero,
+                )
+
+        # Set properties recursively on submodules
+        self.profile_mode = profile_mode
+        self.amp_mode = amp_mode
+        if self.use_apex_gn:
```
AnnaKwa, line 539: Added this to ensure weights are also in channels_last format if using apex group norm. If the weights and the input don't match memory formats there will be extra conversions which slow down the pass.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -0,0 +1,695 @@
...
+        "profile_mode", bool, "Should be set to ``True`` to enable profiling."
+    )
+    amp_mode = _recursive_property(
+        "amp_mode",
+        bool,
+        "Should be set to ``True`` to enable automatic mixed precision.",
+    )
+
+    def forward(self, x, noise_labels, class_labels, augment_labels=None):
+        validate_shape(
```
AnnaKwa: Added the same shape validation as done in the existing unet.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -0,0 +1,695 @@
...
+        valid_decoder_types = ["standard", "skip"]
+        if decoder_type not in valid_decoder_types:
+            raise ValueError(
+                f"Invalid decoder_type: {decoder_type}. "
+                f"Must be one of {valid_decoder_types}."
+            )
+        check_img_resolution = (
+            min(img_resolution) if isinstance(img_resolution, list) else img_resolution
+        )
+        check_level_compatibility(check_img_resolution, channel_mult, attn_resolutions)
```
AnnaKwa, line 340: Added same check as done in existing unet.

fme/downscaling/modules/utils.py
```diff
@@ -0,0 +1,48 @@
+class NonDivisibleShapeError(ValueError):
```
AnnaKwa, line 1: This module is cut and pasted out of `physicsnemo_unets_v1.unets.py`, there are no changes to the code.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -0,0 +1,695 @@
...
+            if not (
+                (is_power_of_2 and d < self._input_shape_mult)
+                or (d % self._input_shape_mult == 0)
+            ):
+                raise ValueError(
+                    f"Input spatial dimensions ({x.shape[-2:]}) must be "
+                    f"either powers of 2 or multiples of 2**(N-1) where "
+                    f"N (={self._num_levels}) is the number of levels "
+                    f"in the U-Net."
+                )
```
frodre, line 575: It looks like they have a validation check on the input shape itself, which is a bit more restrictive than we've previously put, but works for our default patch size (16 -> 512).  I think we can probably just use this over our custom validation.

frodre (approved):
LGTM

PR #806, opened 2026-02-09
Add `SongUNetv2` to diffusion registry

AnnaKwa:
- Adds a `channels_last` init arg to the `UNetDiffusionModule`. If True, input tensors will be converted to channels_last memory format, which is more performant but should only be done if the underlying module's weights are also in this format. This defaults to False which maintains previous behavior for previous existing unets.
- Adds `SongUNetv2` as an option in the diffusion registry. If apex_gn is configured to True (utilizes group norm layer that preserves channels_last format), it will build the `UNetDiffusionModule` with `channels_last=True`.

- [x] Tests added. One issue is that the CI unit tests build the environment directly with pip instead of using an existing image, and I didn't want to add apex to the installation because it is pretty slow to build. This means the test that checks if `channels_last` is correctly used when apex is enabled  is skipped in the unit test suite. I did run this test manually on a GPU machine with apex to check that it passed.

### DIFF
--- a/fme/downscaling/modules/unet_diffusion.py
+++ b/fme/downscaling/modules/unet_diffusion.py
@@ -20,8 +20,9 @@ class UNetDiffusionModule(torch.nn.Module):
         unet: The U-Net model.
         use_amp_bf16: use automatic mixed precision casting to bfloat16 in forward pass
         channels_last: Convert input tensors to channels last format.
-            Conversion should only be used for UNet modules compatible with Apex GroupNorm, e.g., `SongUNetv2`.
-            Defaults to False for backwards compatibility.
+            Conversion should only be used for UNet modules compatible with
+            Apex GroupNorm, e.g., `SongUNetv2`. Defaults to False for backwards
+            compatibility.
     """
 
     def __init__(


fme/downscaling/modules/unet_diffusion.py
```diff
@@ -56,8 +61,8 @@ def forward(
         device = get_device()
         with self._amp_context:
             return self.unet(
-                latent.to(device),
-                conditioning.to(device),
+                latent.to(device, memory_format=self._memory_format),
```
AnnaKwa, line 65: The vendorized UNet code does this conversion in its forward pass. However, it is necessary to convert here at the top level forward pass otherwise operations in the preconditioner convert it back to contiguous format.

fme/downscaling/modules/unet_diffusion.py
```diff
@@ -19,16 +19,21 @@ class UNetDiffusionModule(torch.nn.Module):
     Args:
         unet: The U-Net model.
         use_amp_bf16: use automatic mixed precision casting to bfloat16 in forward pass
+        channels_last: Convert input tensors to channels last format.
+            Conversion should only used for SongUNetv2 unets using Apex GroupNorm.
```
frodre:
```suggestion
            Conversion should only be used for UNet modules compatible with Apex GroupNorm, e.g., `SongUNetv2`.
```

fme/downscaling/modules/unet_diffusion.py
```diff
@@ -19,16 +19,21 @@ class UNetDiffusionModule(torch.nn.Module):
...
+            Defaults to False.
```
frodre:
```suggestion
            Defaults to False for backwards compatibility.
```

PR #809, opened 2026-02-09
Add `CoupledTrainStepper`

jpdunc23:
Adds `train_stepper: CoupledTrainStepperConfig` to the coupled training config, which configures and builds a  `CoupledTrainStepper` implementing `TrainStepperABC`.

**WARNING**: This is a breaking change for existing coupled training configs.

Changes:
- Component stepper `loss: StepLossConfig` and `loss_contributions: LossContributionsConfig` are now configured via the `ocean: ComponentTrainingConfig` and `atmosphere: ComponentTrainingConfig` attributes of  `CoupledTrainStepperConfig`.
- `CoupledStepper` no longer implements `TrainStepperABC`.
- Removed public `loss_obj` and `effective_loss_scaling` properties from `fme.ace.stepper.Stepper` and added a new public method `build_loss`.

- [x] Tests added

### DIFF
--- a/fme/core/benchmark/test_timer.py
+++ b/fme/core/benchmark/test_timer.py
@@ -1,0 +1,115 @@
+from typing import Literal
+from unittest.mock import patch
+
+import pytest
+import torch
+
+from fme.core.benchmark.timer import CUDATimer, TimerResult
+
+
+@pytest.mark.parametrize("is_available", [True, False])
+def test_new_if_available(is_available: bool):
+    from fme.core.benchmark.timer import CUDATimer, NullTimer
+
+    with patch("torch.cuda.is_available", return_value=is_available):
+        timer = CUDATimer.new_if_available()
+    if is_available:
+        assert isinstance(timer, CUDATimer)
+    else:
+        assert isinstance(timer, NullTimer)
+
+
+@pytest.mark.skipif(
+    not torch.cuda.is_available(),
+    reason="CUDA is not available, skipping CUDATimer tests.",
+)
+def test_timer_with_child():
+    timer = CUDATimer()
+    with timer:
+        # get cuda to wait
+        torch.cuda._sleep(100_000)
+        with timer.child("child"):
+            torch.cuda._sleep(100_000)
+    result = timer.result
+    assert "child" in result.children
+    # parent time should include the child time, so it should be
+    # at least 2x the child time (since we sleep for the same amount of time in both)
+    assert result.avg_time >= 2.0 * result.children["child"].avg_time
+
+
+def _create_parent_result(avg_time: float) -> TimerResult:
+    return TimerResult(count=2, avg_time=avg_time, children={})
+
+
+def _create_child_result(avg_time: float) -> TimerResult:
+    return TimerResult(
+        count=2,
+        avg_time=1.0,
+        children={"child": TimerResult(count=2, avg_time=avg_time, children={})},
+    )
+
+
+@pytest.mark.parametrize(
+    "v1, v2, rtol, expect_raise",
+    [
+        (100, 101, 0.02, False),  # within 2%
+        (100, 103, 0.02, True),  # outside 2%
+        (100, 102, 0.02, False),  # exactly 2% is considered inside
+        (10000, 10201, 0.02, True),  # more than 2% is considered outside
+        (100, 102, 0.03, False),  # exactly 2% is within 3%
+    ],
+)
+@pytest.mark.parametrize("kind", ["parent", "child"])
+def test_assert_close(
+    v1: int, v2: int, rtol: float, kind: Literal["parent", "child"], expect_raise: bool
+):
+    if kind == "child":
+        result1 = _create_child_result(avg_time=v1)
+        result2 = _create_child_result(avg_time=v2)
+    else:
+        result1 = _create_parent_result(avg_time=v1)
+        result2 = _create_parent_result(avg_time=v2)
+    if expect_raise:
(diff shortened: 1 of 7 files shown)


fme/coupled/loss.py
```diff
@@ -57,11 +61,11 @@ class LossContributionsConfig:
 
     def build(
         self,
-        loss_obj: Callable[[TensorMapping, TensorMapping, int], torch.Tensor],
+        loss_obj: StepLoss,
         time_dim: int,
     ) -> StepLossABC:
         if self.n_steps == 0 or self.weight == 0.0:
-            return NullLossContributions()
+            return NullLossContributions(loss_obj)
```
jpdunc23, line 68: This preserves the existing behavior where we used a component stepper's `effective_loss_scaling` to compute `mse_fractional_components` metrics even if the stepper had no loss contribution in coupled training.

fme/coupled/test_loss.py
```diff
@@ -60,6 +61,10 @@ def __init__(
         self._loss_obj = loss_obj
         self._time_dim = time_dim
 
+    @property
+    def effective_loss_scaling(self):
+        raise NotImplementedError
```
mcgibbon:
```suggestion
        raise NotImplementedError()
```

fme/coupled/test_loss.py
```diff
@@ -153,13 +159,12 @@ def test_null_loss_contributions(steps_thru_atmos_7, ocean_config_kwargs):
     # test LossContributionsConfig with n_steps = 0
     atmos_loss_config = LossContributionsConfig()
     atmosphere_loss = atmos_loss_config.build(
-        loss_obj=lambda *_, **__: torch.tensor(5.25),
+        loss_obj=Mock(spec=StepLoss, side_effect=lambda *_, **__: torch.tensor(5.25)),
```
mcgibbon: nit: The three lines changed use three different ways to specify the loss side-effect - via mae_loss, via a lambda function returning a constant, and via a return_value instead of a side_effect. You could consider using return_value for this one to reduce that down to 2 ways, at least.

fme/coupled/test_stepper.py
```diff
@@ -1501,7 +1503,12 @@ def test_train_on_batch_with_derived_variables():
         n_forward_times_atmosphere=2,
         n_samples=3,
     )
-    output = coupler.train_on_batch(
+    train_stepper_config = CoupledTrainStepperConfig(
```
mcgibbon, line 1506: nit: Avoid the 3x copy-paste of this process by making a `get_train_stepper_and_batch` helper that does it and calls `get_stepper_and_batch` internally.

fme/coupled/stepper.py
```diff
@@ -1319,12 +1290,169 @@ def predict(
...
+@dataclasses.dataclass
+class ComponentTrainingConfig:
+    loss: StepLossConfig
+    loss_contributions: LossContributionsConfig = dataclasses.field(
+        default_factory=lambda: LossContributionsConfig()
+    )
+
+
+@dataclasses.dataclass
+class CoupledTrainStepperConfig:
```
mcgibbon, line 1333: Do you have an example of the updated training config committed somewhere I could check out? It would be nice to have a baseline config for coupled training, if so I could see the changes to the baseline in this PR.

Update: Ah I see test_train.py mostly fits this purpose, good. Still, could be nice to have a baseline in the future.

fme/coupled/test_train.py
```diff
@@ -109,8 +107,6 @@
           out_names: {ocean_out_names}
   atmosphere:
     timedelta: 1D
-    loss_contributions:
-      n_steps: {loss_atmos_n_steps}
     stepper:
       loss:
```
mcgibbon: Question: Why is `loss: type: MSE` in both the `atmosphere: stepper:` and in the `train_stepper: atmosphere:`? I am guessing because we haven't updated the ACE configs yet and it's required in the config, in which case that's fine, but I though I should ask to be sure.

fme/coupled/train/train.py
```diff
@@ -58,8 +58,8 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> Trainer:
         n_timesteps_atmosphere=n_timesteps_atmosphere,
         ocean_normalize=stepper.ocean.normalizer.normalize,
         atmosphere_normalize=stepper.atmosphere.normalizer.normalize,
-        ocean_loss_scaling=stepper.ocean.effective_loss_scaling,
-        atmosphere_loss_scaling=stepper.atmosphere.effective_loss_scaling,
+        ocean_loss_scaling=stepper.effective_loss_scaling.ocean,
```
mcgibbon: nit: pass `loss_scaling: stepper.loss_scaling` instead of two arguments containing the parts

fme/coupled/test_stepper.py
jpdunc23: Good idea, but I'll defer this cleanup to https://github.com/ai2cm/ace/pull/814 since the way in which the train stepper is built is going to change.

fme/coupled/stepper.py
jpdunc23: Agreed, I will work on a new PR to add the baseline.

fme/coupled/test_train.py
jpdunc23: That's right, although including it in the yaml here isn't strictly necessary since there is a default value on `StepperConfig`. I'll remove it so it's a bit clearer here.

PR #810, opened 2026-02-09
Add ability to override of prognostic variables with the "truth" during evaluation 

Arcomano1234:
During AMIP experiments and E3SM ACE there were several inference experiments that wanted to be conducted where a prognostic variable was to be overwritten with the "truth" at each timestep of the free-run. There have been previous hacky ways to do this, but this PR gives a configurable way to overwritten prognostic variables during inference or evaluation by using `step_with_adjustment` to overwritten the one-step prediction of the desired prognostic variables. 

Note `prescribed_prognostic_names` is added to `SingleModuleStepperConfig` , however, this will most likely be unused for most applications during and instead we use the `stepper_override` option to allow `prescribed_prognostic_names`  to be set during inference/eval. 

Example configuration can be found [here](https://github.com/ai2cm/ace/commit/fea1ac1de9a567bd11367f56552c6bc8b431da0f). With the intended use during inference/eval being:
```
stepper_override:
  prescribed_prognostic_names:
    - example_prognostic_var_1
    - example_prognostic_var_2
```

Changes:
- adds `prescribed_prognostic_names` to `SingleModuleStepperConfig` 
- adds `replace_prescribed_prognostic_names` method to override the serialized `prescribed_prognostic_names` (which will probably be set to None for trained ckpts) to allow `prescribed_prognostic_names` be set during inference/eval. 

- [x] Tests added


fme/core/step/single_module.py
```diff
@@ -426,4 +453,7 @@ def step_with_adjustments(
         output = corrector(input, output, next_step_input_data)
     if ocean is not None:
         output = ocean(input, output, next_step_input_data)
+    for name in prescribed_prognostic_names:
+        if name in next_step_input_data:
+            output = {**output, name: next_step_input_data[name]}
```
elynnwu, line 458: [nit] consider adding a warning for name in `prescribed_prognostic_names` but not in `next_step_input_data` in case of misconfiguration

fme/ace/step/fcn3.py
```diff
@@ -265,9 +272,11 @@ def output_names(self) -> list[str]:
     def next_step_input_names(self) -> list[str]:
         """Names of variables provided in next_step_input_data."""
         input_only_names = set(self.input_names).difference(self.output_names)
-        if self.ocean is None:
-            return list(input_only_names)
-        return list(input_only_names.union(self.ocean.forcing_names))
+        result = set(input_only_names)
```
elynnwu: I think `input_only_names` is already a set so you can just do:
`result = set(self.input_names).difference(self.output_names)`

fme/core/step/single_module.py
Arcomano1234: I believe all the checks before it gets here prevent that from happening but I will add a raise here just in case

fme/ace/step/fcn3.py
Arcomano1234: Good catch

elynnwu (approved):

PR #811, opened 2026-02-10
Add CUDATimer and NullTimer

mcgibbon:
This PR adds CUDATimer for timing CUDA operations, and NullTimer to provide a null API matching it.

This will be used for benchmarking in later PRs.

- [x] Tests added

### DIFF
--- a/fme/core/benchmark/test_timer.py
+++ b/fme/core/benchmark/test_timer.py
@@ -111,4 +111,5 @@ def test_assert_close_children_rtol_raises():
         avg_time=110.0,
         children={"child": TimerResult(count=2, avg_time=103.0, children={})},
     )
-    result2.assert_close(result1, rtol=0.5, children_rtol=0.2)
+    with pytest.raises(AssertionError):
+        result2.assert_close(result1, rtol=0.05, children_rtol=0.2)


fme/core/benchmark/timer.py
```diff
@@ -0,0 +1,171 @@
...
+    def child(self, name: str) -> "Self":
+        return self
+
+    def __enter__(self) -> "Self":
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> Literal[False]:
+        return False
+
+    def report(self) -> TimerResult:
```
elynnwu: Should this be `result` to be consistent with `CUDATimer`?

fme/core/benchmark/timer.py
```diff
@@ -0,0 +1,171 @@
...
+        return TimerResult(total_runs=0, avg_time=0.0, children={})
+
+
+_: Timer = NullTimer()
+del _
```
elynnwu, line 56: If I understand correctly, is this to enforce type-checking?

fme/core/benchmark/timer.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class Timer(Protocol):
+    def child(self, name: str) -> Self: ...
+    def __enter__(self) -> Self: ...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> Literal[False]: ...
+
+
+class NullTimer:
+    def context(self, name: str) -> contextlib.nullcontext:
```
elynnwu: Should `CUDATimer` also have `context`?

fme/core/benchmark/timer.py
```diff
@@ -0,0 +1,171 @@
...
+        self.start.record(self._stream)
+        self._start_recorded = True
+
+    def record_end(self):
+        if not self._start_recorded:
+            raise RuntimeError("record_start must be called before record_end")
+        if self._end_recorded:
+            raise RuntimeError("record_end has already been called on this EventPair.")
+        if self._stream is None:
+            raise RuntimeError("record_start must be called before record_end")
```
elynnwu, line 82: [nit] I think this is redundant because stream is always set in `record_start`

fme/core/benchmark/timer.py
```diff
@@ -0,0 +1,171 @@
...
+import contextlib
+import dataclasses
+from typing import Literal, Protocol, Self
+
+import torch
+
+
+@dataclasses.dataclass
+class TimerResult:
+    total_runs: int
```
elynnwu: [nit] I think we use runs in other contexts and here it's really the number of times a section is timed. Maybe change it to `n_timings` or simply `count`?

mcgibbon: I like `count`, updated.

fme/core/benchmark/timer.py
mcgibbon: Removed from NullTimer, this is left-over from before the timer itself was a context manager.

fme/core/benchmark/timer.py
mcgibbon: Good point. Since it's clearly not used, I deleted it.

fme/core/benchmark/test_timer.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    result2 = TimerResult(
+        total_runs=2,
+        avg_time=110.0,
+        children={"child": TimerResult(total_runs=2, avg_time=103.0, children={})},
+    )
+    result2.assert_close(result1, rtol=0.2, children_rtol=0.05)
+
+
+def test_assert_close_children_rtol_raises():
```
elynnwu, line 102: Do you want this to actually raise exception? It currently does not

fme/core/benchmark/timer.py
mcgibbon: Yes, that's correct.

fme/core/benchmark/timer.py
mcgibbon: I'd err on the side of keeping it - It makes it so `self._stream` type-hints as non-None after this point, and makes it clearer that this set-up is getting done (even though I agree we won't ever hit this block).

fme/core/benchmark/test_timer.py
mcgibbon: Ah true. Fixed.

elynnwu (approved):

PR #812, opened 2026-02-10
Add benchmark_memory context manager

mcgibbon:
This PR adds a benchmark_memory context manager to record CUDA memory allocation.

This will be used in later PRs to benchmark memory usage.

- [x] Tests added

### DIFF
--- a/fme/core/benchmark/test_memory.py
+++ b/fme/core/benchmark/test_memory.py
@@ -20,6 +20,7 @@ def test_cannot_get_result_before_end():
             bm.result
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")
 def test_ensure_cannot_restart():
     assert (
         not memory._benchmark_memory_started


fme/core/benchmark/memory.py
```diff
@@ -0,0 +1,86 @@
...
+        self._started = False
+        self._ended = False
+
+    def __enter__(self) -> "MemoryBenchmark":
+        global _benchmark_memory_started
+        if _benchmark_memory_started:
+            raise RuntimeError(
+                "benchmark_memory cannot be nested due to its use of globals"
+            )
+        _benchmark_memory_started = True
```
jpdunc23: Agent review flagged an edge case bug here. I don't have a env with CUDA set up to test quickly, but looks legit to me.

> In `MemoryBenchmark.__enter__`, the `global _benchmark_memory_started` is set to True before the `self._ended` check. If a user tries to reuse a finished instance, the `self._ended` guard raises a RuntimeError, but the global is left as True permanently -- blocking all future `MemoryBenchmark` usage in the process:

```python
# Reproducer:
bm = MemoryBenchmark()
with bm:
    pass  # _ended is now True

try:
    with bm:  # global set to True, then self._ended raises
        pass
except RuntimeError:
    pass

# Global is now stuck True -- this will fail:
with benchmark_memory():  # RuntimeError: cannot be nested
    pass
```

fme/core/benchmark/test_memory.py
```diff
@@ -0,0 +1,56 @@
...
+    assert bm2.result.max_alloc > bm1.result.max_alloc
+
+
+@pytest.mark.parametrize(
+    "v1, v2, rtol, expect_raise",
+    [
+        (100, 101, 0.02, False),  # within 2%
+        (100, 103, 0.02, True),  # outside 2%
+        (100, 102, 0.02, False),  # exactly 2% is considered inside
+        (10000, 10201, 0.02, True),  # more than 2% is considered outside
```
jpdunc23, line 53: It might be worth adding a case that shows the directionality, e.g.:


```suggestion
        (10000, 10201, 0.02, True),  # more than 2% is considered outside
        (10201, 10000, 0.02, False), # passes
```

fme/core/benchmark/memory.py
mcgibbon: Added a failing test and fixed the bug.

fme/core/benchmark/test_memory.py
mcgibbon: Updated

PR #803, opened 2026-02-09
Feature/profile sfno no changes

mcgibbon:
Short description of why the PR is needed and how it satisfies those requirements, in sentence form.

Changes:
- symbol (e.g. `fme.core.my_function`) or script and concise description of changes or added feature
- Can group multiple related symbols on a single bullet

- [ ] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

Resolves #<github issues> (delete if none)

PR #813, opened 2026-02-10
Pass timer contexts in conditional SFNO

mcgibbon:
This PR updates the conditional sfno FourierNeuralOperatorBlock and its children to pass Timer contexts.

These will be used in a later PR to perform CUDA profiling. For now, they are added without breaking changes to the code.

Changes:
- Updated conditional SFNO FourierNeuralOperatorBlock and its dependents to pass `timer: Timer` arguments during forward pass and use them to record times during code segments.

- [ ] Tests added


mcgibbon:
These contexts are used more fully in the benchmarking branch https://github.com/ai2cm/ace/pull/803, getting merged in parts. See that PR for an example plot of the timing split.

fme/core/models/conditional_sfno/sht.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        return f'nlat={self.nlat}, nlon={self.nlon},\n lmax={self.lmax}, mmax={self.mmax},\n grid={self.grid}, csphase={self.csphase}'
+
+    def forward(self, x: torch.Tensor, timer: Timer = NullTimer()):
+
+        assert(x.shape[-2] == self.lmax)
+        assert(x.shape[-1] == self.mmax)
+
+        with torch.autocast("cuda", enabled=False):
+            with timer.child("contraction"):
```
Arcomano1234, line 210: Did you intend for both the SHT and ISHT to have the same `contraction` timer?

Arcomano1234 (approved):
Left 1 comment but it LGTM

fme/core/models/conditional_sfno/sht.py
mcgibbon: Because of how child timers are nested, these are distinct timers, stored and plotted separately. But yes, I did intend the same name.

Arcomano1234: I see that makes sense I didn't look too closely at the timers PR and just wanted to make sure

PR #815, opened 2026-02-10
Add benchmarks with regression testing

mcgibbon:
This PR adds a system for registering benchmarks to be run in a new `fme.core.benchmark.run` entrypoint.

All benchmarks are regression tested for backwards compatibility.

Changes:
- Added BenchmarkABC and register_benchmark to record new benchmarks
- Added `fme.core.benchmark.run` entrypoint
- conftest.py updated to use deterministic algorithms during tests
- Added `fme.core.device.force_cpu` context to force cpu usage during regression tests

- [x] Tests added

### DIFF
--- a/fme/core/benchmark/test_run.py
+++ b/fme/core/benchmark/test_run.py
@@ -1,9 +1,13 @@
 import pathlib
 import tempfile
 
+import pytest
+import torch
+
 from fme.core.benchmark.run import main
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")
 def test_run():
     # Just test that the main function runs without error on a simple benchmark
     # We don't care about the output here, just that it completes successfully


mcgibbon:
An example plot from `fme.core.benchmark.run`:

csfno_block_tesla_t4_1c756074

fme/core/models/conditional_sfno/benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+                H=H,
+                L=L,
+                G=G,
+                conditional_embed_dim_noise=conditional_embed_dim_noise,
+                conditional_embed_dim_labels=conditional_embed_dim_labels,
+                conditional_embed_dim_pos=conditional_embed_dim_pos,
+            )
+
+        @classmethod
+        def _new_with_params(
```
AnnaKwa, line 49: What's the purpose of having this separate method rather than just having this in `new`?

fme/core/benchmark/run.py
```diff
@@ -0,0 +1,145 @@
...
+import os
+import pathlib
+import subprocess
+import sys
+
+import torch
+
+from fme.core.benchmark.benchmark import get_benchmarks
+
+RESULTS_PATH = pathlib.Path(os.path.abspath(os.path.dirname(__file__))) / "results"
```
AnnaKwa, line 14: It'd be nice to print at the end of `main` where your png is saved.

AnnaKwa, line 14: Could this be an optional CLI arg? The use case I'm thinking of is if I wanted to run and save the results to `/results` before a job starts on a cluster with different hardware since it's otherwise a pain to start an interactive job just to get that information.

fme/core/models/conditional_sfno/benchmark.py
mcgibbon: So that it can be re-used by the two different `new` implementations.

fme/core/benchmark/run.py
mcgibbon: Added as a CLI arg.

mcgibbon:
Added a basic regression test just to make sure it still runs.

PR #816, opened 2026-02-10
Cache venv for gpu tests

mcgibbon:
The python caching we use in github actions misses our GPU python environment, because it is installed differently, using uv. This PR fixes the caching for gpu tests.

- [ ] Tests added

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -823,20 +823,6 @@ class Stepper:
         self._no_optimization = NullOptimization()
         self._parameter_initializer = parameter_initializer
 
-        def get_loss_obj() -> StepLoss:
-            loss_normalizer = step.get_loss_normalizer()
-            if config.loss is None:
-                raise ValueError("Loss is not configured")
-            return config.loss.build(
-                dataset_info.gridded_operations,
-                out_names=config.loss_names,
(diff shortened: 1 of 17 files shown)


mcgibbon:
With caching completed in 12 minutes: https://github.com/ai2cm/ace/actions/runs/21885139515/job/63177967919
First run completed in 22 minutes: https://github.com/ai2cm/ace/actions/runs/21882622591/job/63169049706

.github/workflows/tests.yaml
```diff
@@ -71,13 +71,17 @@ jobs:
         with:
           path: ${{ env.pythonLocation }}
           key: ${{ env.CURRENT_WEEK }}-${{ env.pythonLocation }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('requirements-healpix.txt') }}-${{ hashFiles('constraints.txt') }}
+      - uses: actions/cache@v4
```
oliverwm1, line 74: can the existing cache action (which uses `env.pythonLocation`) be deleted now?

mcgibbon: My guess would be no, it caches the python environment used to install uv which is used to install the test environment. Like, maybe we can, but the runtime will probably go up a little bit.

PR #819, opened 2026-02-11
Fix a bug in `EMATracker` resumption during training

### LINKED ISSUE #483: Sudden dips in val / inline inference metrics after resuming from a checkpoint
Recent ocean emulator runs show weird dips in validation / inference metrics after restarting from a checkpoint. In the screenshots below, the dips coincide with the epoch where the job resumed.

Possible explanations I can think of:
- There's a bug in the [branch](https://github.com/ai2cm/full-model/pull/1195/commits/62df788bcfefac839b56b96649f0c9bec8e76fe2) used for this experiment.
- Model buffers have different states before and after the restart.
- The optimizer state isn't saved / loaded correctly. 

val_mean_loss_dip

inference_time_mean_rmse_dip

### LINKED ISSUE #668: Peculiar periodic spikes in inline inference error
In looking at the results of a group of recent training jobs I noticed oddly regular spikes in inline inference error ([link to WandB](https://wandb.ai/ai2cm/ace-som/workspace?nw=j831pxqv6a&panelDisplayName=inference%2Ftime_mean_norm%2Frmse%2Fchannel_mean&panelSectionName=inference%2Ftime_mean_norm%2Frmse)).  

<p align="center">
Image
</p>

Each of these training jobs was configured to run inline inference every four epochs.  Inline inference statistics are reported after the first epoch (1096 steps) and every four epochs (4384 steps) thereafter.  In almost all circumstances, every fifth inline inference RMSE reported is noticeably larger than those surrounding it.  This feels like a bug, but it is perplexing what it could be.  I will try to reproduce it with other inline inference frequencies, and possibly see if I can reduce it to a smaller example.

Using the same code, but different datasets and inline inference run after every epoch, behavior seems a bit more normal ([link to WandB](https://wandb.ai/ai2cm/ace-som/workspace?nw=nwuserspencerc_ai2&panelDisplayName=inference%2Ftime_mean_norm%2Frmse%2Fchannel_mean&panelSectionName=inference%2Ftime_mean_norm%2Frmse)); naturally spikes exist, but not on regular intervals that are consistent across seeds.

### LINKED ISSUE #682: Possible `ema_ckpt` resumption bug
WandB workspace view: https://wandb.ai/ai2cm/ace?nw=jbd4o7fvbye

We observed a jump in the inference error and validation loss when resuming a job with `ResumeResultsConfig`. ~~May be related to https://github.com/ai2cm/ace-internal-test/issues/626.~~ This seems to be non-specific to `ResumeResultsConfig` but instead potentially affects any resuming job (though possibly only those using `validate_using_ema=True`).

Image

There is no apparent jump in the batch loss, pointing to a potential issue with EMA resumption.

Image

jpdunc23:
Fixes a bug in training resumption where the resumed `EMATracker` state is overwritten by a fresh build of the `EMATracker` with `num_updates` equal to 0. Together with the default `faster_decay_at_start=True`, this pushed the EMA params much closer to the current step's module params than desired.

This bug was caused by setting `self._ema = build_ema(stepper.modules)` after calling `self.restore_checkpoint` in `Trainer` init. `restore_checkpoint` creates an `EMATracker` instance from the saved checkpoint state and assigns it to `Trainer._ema` and this was later overwritten by the new instance from `build_ema`.

- [x] Tests added

Resolves #668, resolves #682, resolves #483

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -71,13 +71,17 @@ jobs:
         with:
           path: ${{ env.pythonLocation }}
           key: ${{ env.CURRENT_WEEK }}-${{ env.pythonLocation }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('requirements-healpix.txt') }}-${{ hashFiles('constraints.txt') }}
+      - uses: actions/cache@v4
+        with:
+          path: .venv
+          key: ${{ env.CURRENT_WEEK }}-.venv-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('requirements-healpix.txt') }}-${{ hashFiles('constraints.txt') }}
       - name: Install dependencies
         run: |
           sudo apt-get update
           sudo apt-get install -y make git build-essential
           python -m pip install uv
           # can't use --system here because it doesn't work with --no-build-isolation and a healpix dependency
-          uv venv --python 3.11
+          if [ ! -d ".venv" ]; then uv venv --python 3.11; fi
           uv pip install -c constraints.txt -e .[dev]
           uv pip install --no-build-isolation -c constraints.txt -e .[dev,healpix,graphcast]
       - name: Run CUDA check and pytest


fme/core/ema.py
```diff
@@ -87,9 +89,9 @@ def __init__(
             raise ValueError("Decay must be between 0 and 1")
 
         self._module_name_to_ema_name = {}
-        self.decay = torch.tensor(decay, dtype=torch.float32)
+        self.decay = torch.tensor(decay, dtype=torch.float32).to(get_device())
         self._faster_decay_at_start = faster_decay_at_start
-        self.num_updates = torch.tensor(0, dtype=torch.int)
+        self.num_updates = torch.tensor(0, dtype=torch.int).to(get_device())
```
jpdunc23, line 94: This change just makes the init location of these tensors consistent with the init location for our stepper modules, which we always send to the device on stepper init. 

This is also helpful (mostly for the tests) to avoid inconsistency between a newly constructed `EMATracker` and one that has been resumed from a checkpoint in `Trainer`, since the latter will have `decay` and `num_updates` on the device.

fme/core/generics/test_trainer.py
```diff
@@ -1142,3 +1144,37 @@ def test_no_checkpoints_saved_when_disabled(tmp_path: str):
...
+    """EMA num_updates and ema_params must survive a checkpoint save/restore cycle."""
+    n_train_batches = 4
+    _, trainer = get_trainer(
+        tmp_path,
+        max_epochs=1,
+        n_train_batches=n_train_batches,
+        stepper_module_values=np.array([1.0]),
+        checkpoint_save_epochs=Slice(start=0, stop=0),
+    )
+    trainer.train()
```
frodre, line 1159: Add a short comment that trainer.train() has a checkpoint saving for the single epoch, I was a little confused by the slice usage of (0, 0) because that signified to me an empty slice.

fme/core/generics/trainer.py
```diff
@@ -249,6 +249,7 @@ def __init__(
         self._end_of_epoch_callback = end_of_epoch_callback
         self._no_optimization = NullOptimization()
         self._aggregator_builder = aggregator_builder
+        self._ema = build_ema(stepper.modules)
```
frodre: I would add a comment that this is import to occur before the restore_checkpoint.  Also, I might detail the pathway of the failure in the PR or in this comment to make it clear why it has to be done in this order.

frodre (approved):
Thanks, @jpdunc23!  Nice find.  I would mainly add a little more commentary for future readers to understand why the ordering in the restore location is important given the current implementation.  Might save someone from making an unintended regression.

fme/core/generics/trainer.py
jpdunc23: Thanks! Added comments and updated the PR description.

PR #820, opened 2026-02-11
Add `version` option to NoiseConditionedSFNOBuilder

mcgibbon:
This PR adds features to version the code implementation underlying NoiseConditionedSFNOBuilder, allowing for code changes that break backwards compatibility with checkpoint loading since the previous code versions continue to be used by older checkpoints.

No functionality is changed in this PR. Diffusion-forward-prediction (not downscaling) is made to reference the v1 code, so backwards compatibility will be broken for diffusion checkpoints in a PR soon.

Changes:
- Moved module builder code in `fme/ace/registry/stochastic_sfno` to `fme/core/models/conditional_sfno` so that it can be versioned
- Moved existing code in `fme/core/models/conditional_sfno` to both `v0` and `v1` subdirectories (the implementations in this PR are identical).
- Added `version` option to `NoiseConditionedSFNOBuilder` which can take options `v0`, `v1`, or `latest`. Latest will be updated to the latest stable version (currently v0) when configuration is loaded.

- [ ] Tests added

### DIFF
--- a/fme/diffusion/registry/sfno.py
+++ b/fme/diffusion/registry/sfno.py
@@ -1,8 +1,11 @@
 import dataclasses
 from typing import Literal
 
-from fme.core.models.conditional_sfno.sfnonet import ContextConfig, get_lat_lon_sfnonet
-from fme.core.models.conditional_sfno.sfnonet import (
+from fme.core.models.conditional_sfno.v1.sfnonet import (
+    ContextConfig,
+    get_lat_lon_sfnonet,
+)
+from fme.core.models.conditional_sfno.v1.sfnonet import (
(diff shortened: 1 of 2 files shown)


mcgibbon:
I highly recommend reviewing this PR commit-by-commit. "copy" commits do not modify the copied code. "move" commits only modify the way the moved code is called or imported, not the moved code itself.

mcgibbon:
There are some redundant unit tests after this PR, but I'd wait to delete them until we've definitely updated to v1 / frozen the code for v0. Also this PR will have merge conflicts with the benchmarking (I'll keep the benchmark in the outer conditional_sfno dir, and have it reference the latest version / v1), but don't worry about that in review.

PR #822, opened 2026-02-11
Allow multiple static inputs

AnnaKwa:
This refactors the downscaling code to use the full `StaticInputs` class in `generate` calls. 

- `DiffusionModel._get_input_from_coarse` concats all `StaticInputs.fields.data` as additional input channels
- **breaking change**: `TrainConfig` has a new field `static_inputs` which should be a mapping of `{var_name: path_to_dataset}`. This should be used to provide the static inputs fields- they will no longer be loaded automatically from the fine dataset. In the future, `use_fine_topography` in the config should be deprecated since the `static_inputs` field is the source of this information.
- `GriddedData, PairedGriddedData` are updated to have a `StaticInputs` attribute instead of the single `Topography`
- references to variables named `topography` are updated to the more general `static_input`
- `train, predict, inference, evaluator` entrypoint code and their tests are updated to pass the `StaticInputs` to generate calls

- last three commits are minor updates to use the correct number of input channels given the number of static inputs, add the static inputs to the `from_state` method, and update tests for these changes

(Exact diff of v1 branch but used cursor to organize the commits to be more reviewable)


fme/downscaling/models.py
```diff
@@ -80,7 +74,7 @@ def load(self):
 
 @dataclasses.dataclass
 class DiffusionModelConfig:
-    """
+    f"""
```
frodre: necessary?

fme/downscaling/predictors/test_composite.py
```diff
@@ -169,11 +174,16 @@ def test_SpatialCompositePredictor_generate_on_batch_no_target(patch_size_coarse
     paired_batch_data = get_paired_test_data(
         *coarse_extent, downscale_factor=downscale_factor, batch_size=batch_size
     )
-    topography = Topography(
-        torch.randn(
-            coarse_extent[0] * downscale_factor, coarse_extent[1] * downscale_factor
-        ),
-        paired_batch_data.fine.latlon_coordinates[0],
+    static_inputs = StaticInputs(
```
frodre: I would consider a helper function for creating static inputs in tests where there are multiple instances.

fme/downscaling/data/test_topography.py
```diff
@@ -82,7 +82,7 @@ def test_Topography_generate_from_patches():
     assert torch.equal(generated_patches[1].data, torch.tensor([[2], [6]]))
 
 
-def test_StaticInputs_generate_from_patches():
+def testStaticInputs_generate_from_patches():
```
frodre: test_S
```suggestion
def test_StaticInputs_generate_from_patches():
```

fme/downscaling/data/test_topography.py
```diff
@@ -124,13 +124,13 @@ def test_StaticInputs_generate_from_patches():
     assert torch.equal(generated_patches[1][1].data, expected_topography_patch_1 * -1.0)
 
 
-def test_StaticInputs_serialize():
+def testStaticInputs_serialize():
```
frodre:
```suggestion
def test_StaticInputs_serialize():
```

fme/downscaling/_deterministic_models.py
```diff
@@ -166,14 +166,14 @@ def _run_on_batch(
...
-            if topography is None:
+            if static_inputs is None:
                 raise ValueError(
                     "Topography must be provided for each batch when use of fine "
                     "topography is enabled."
                 )
             else:
                 # Join the normalized topography to the input (see dataset for details)
-                topo = topography.data.unsqueeze(self._channel_axis)
+                topo = static_inputs.fields[0].data.unsqueeze(self._channel_axis)
```
frodre, line 176: We are going to remove _deterministic_models, right?  Only asking because the use of topography only as item[0] maybe is a bit confusing here.  Ignore if this is all going to disappear.

fme/downscaling/models.py
```diff
@@ -116,6 +112,7 @@ class DiffusionModelConfig:
     predict_residual: bool
     use_fine_topography: bool = False
     use_amp_bf16: bool = False
+    static_inputs: dict[str, str] | None = None
```
frodre: Does this field get used? There's also a static inputs on the `TrainConfig` that gets used to build and is fed into the build of this class.

fme/downscaling/data/datasets.py
```diff
@@ -329,35 +329,35 @@ def on_device(batch: BatchItem) -> BatchItem:
...
             if (
-                self.topography.shape[0] % self.shape[0] != 0
-                or self.topography.shape[1] % self.shape[1] != 0
+                self.static_inputs.shape[0] % self.shape[0] != 0
+                or self.static_inputs.shape[1] % self.shape[1] != 0
             ):
                 raise ValueError(
                     "Topography shape must be evenly divisible by data shape. "
-                    f"Got topography {self.topography.shape} and data {self.shape}"
+                    f"Got topography {self.static_inputs.shape} and data {self.shape}"
```
frodre: Error still references Topography.  Does this need to be updated?  Could be internal to the `StaticInputs` so we don't have to reference attribute specifics outside of their source.

fme/downscaling/test_predict.py
```diff
@@ -99,7 +103,7 @@ def create_predictor_config(
     return out_path
 
 
-@pytest.mark.parametrize("static_inputs_on_model", [True, False])
+@pytest.mark.parametrize("static_inputs_on_model", [True])
```
frodre: Assuming the `False` is removed because we will always expect static inputs on the model?  No need for parametrization.

frodre (changes requested):
Thanks @AnnaKwa,  overall pretty good but still some minor suggestions and light cleanup to do before it's ready to merge.

fme/downscaling/models.py
AnnaKwa: Deleted, not sure what happened there.

fme/downscaling/data/datasets.py
AnnaKwa: Updated the error message. 

>  Could be internal to the StaticInputs

Do you mean the check be internal to StaticInputs? Or something else?

fme/downscaling/predictors/test_composite.py
AnnaKwa: Added in `test_composite, test_cascade`, and `test_output`

fme/downscaling/test_predict.py
AnnaKwa: Yup, correct, removed.

fme/downscaling/_deterministic_models.py
AnnaKwa: Yes, I just changed this so that mypy doesn't complain.

fme/downscaling/models.py
AnnaKwa: Oops, no, good catch- this was left over from a previous iteration. Removed.

fme/downscaling/data/test_config.py
```diff
@@ -64,7 +66,15 @@ def test_DataLoaderConfig_build(tmp_path, very_fast_only: bool):
         lat_extent=ClosedInterval(1, 4),
         lon_extent=ClosedInterval(0, 3),
     )
-    data = data_config.build(requirements=requirements)
+    static_inputs = StaticInputs(
```
frodre: Could use a helper in this file.

fme/downscaling/inference/test_output.py
```diff
@@ -101,8 +116,10 @@ def test_event_config_build_creates_output_target_with_single_time(
         lat_extent=ClosedInterval(2.0, 6.0),
         lon_extent=ClosedInterval(2.0, 6.0),
     )
-
-    output_target = config.build(loader_config, requirements, patch_config)
+    static_inputs = static_inputs = _get_static_inputs((8, 8))
```
frodre:
```suggestion
    static_inputs = _get_static_inputs((8, 8))
```

fme/downscaling/models.py
```diff
@@ -100,6 +94,8 @@ class DiffusionModelConfig:
         use_fine_topography: Whether to use fine topography in the model.
         use_amp_bf16: Whether to use automatic mixed precision (bfloat16) in the
             UNetDiffusionModule.
+
+
```
frodre:
```suggestion
```

frodre (approved):
LGTM, thanks for the updates!  Just a few minor suggestions.

PR #828, opened 2026-02-12
Fix backwards compatibility with pre-group checkpoints

### LINKED ISSUE #826: Parameter size mismatch for `module.conditional_model.blocks.0.filter.filter.weight`
After merging main into my [experiment branch](https://github.com/ai2cm/ace/tree/exper/2025-08-08-jamesd) I encountered this error when in a coupled fine-tuning [job](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KH888JATXPZQWQQH0VPBK78A/logs?jobId=01KH888JTGPQSM7NHH39Q8ZJ8D):

```
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/ace/stepper/single_module.py", line 81, in load_weights_and_history
2026-02-12T06:27:16.977Z [rank0]:     stepper = load_stepper(path)
2026-02-12T06:27:16.977Z [rank0]:               ^^^^^^^^^^^^^^^^^^
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/ace/stepper/single_module.py", line 1764, in load_stepper
2026-02-12T06:27:16.977Z [rank0]:     stepper = Stepper.from_state(checkpoint["stepper"])
2026-02-12T06:27:16.977Z [rank0]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/ace/stepper/single_module.py", line 1358, in from_state
2026-02-12T06:27:16.977Z [rank0]:     stepper.load_state(state)
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/ace/stepper/single_module.py", line 1285, in load_state
2026-02-12T06:27:16.977Z [rank0]:     self._step_obj.load_state(state["step"])
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/core/step/single_module.py", line 383, in load_state
2026-02-12T06:27:16.977Z [rank0]:     self.module.load_state(module)
2026-02-12T06:27:16.977Z [rank0]:   File "/gantry-runtime/fme/core/registry/module.py", line 109, in load_state
2026-02-12T06:27:16.977Z [rank0]:     self._module.load_state_dict(state)
2026-02-12T06:27:16.977Z [rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 2593, in load_state_dict
2026-02-12T06:27:16.977Z [rank0]:     raise RuntimeError(
2026-02-12T06:27:16.977Z [rank0]: RuntimeError: Error(s) in loading state_dict for DistributedDataParallel:
2026-02-12T06:27:16.977Z [rank0]: 	size mismatch for module.conditional_model.blocks.0.filter.filter.weight: copying a param with shape torch.Size([384, 384, 180, 2]) from checkpoint, the shape in current model is torch.Size([1, 384, 384, 180, 2]).
```

Reverting https://github.com/ai2cm/ace/commit/c2d91514897b653d7a4d21eb745e7e58652dcbde fixed the issue for me.

mcgibbon:
A previous PR #788 introduced a change to the weight shape to include a group dimension. This prevents loading existing checkpoints that don't have this dimension. This PR adds a hook to correctly handle these previous checkpoints.

Resolves #826

### DIFF
--- a/fme/core/registry/test_module_registry.py
+++ b/fme/core/registry/test_module_registry.py
@@ -10,7 +10,6 @@ import torch
 import fme
 from fme.core.coordinates import HybridSigmaPressureCoordinate, LatLonCoordinates
 from fme.core.dataset_info import DatasetInfo
-from fme.core.dicts import to_flat_dict
 from fme.core.labels import LabelEncoding
 from fme.core.registry.module import Module
 
@@ -147,11 +146,4 @@ def test_module_backwards_compatibility(selector_name: str):
         dataset_info=dataset_info,
     )
     loaded_state_dict = load_or_cache_state(selector_name, module)
-    new_keys = set(to_flat_dict(module.get_state()).keys()).difference(
-        to_flat_dict(loaded_state_dict).keys()
-    )
     module.load_state(loaded_state_dict)
-    assert not new_keys, (
-        f"New keys added to state dict: {new_keys}. "
-        "Please delete and re-generate the regression target."
-    )


fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -223,8 +223,38 @@ def __init__(
 
         if bias:
             self.bias = nn.Parameter(torch.zeros(1, out_channels, 1, 1))
+        self.in_channels = in_channels
         self.out_channels = out_channels
 
+        # rewrite old checkpoints on load
+        self.register_load_state_dict_pre_hook(self._pre_load_hook)
```
jpdunc23: Nice!

mcgibbon:
Noting that I generated a checkpoint (still running locally, not committed) using dbc292561ae10cdca58126c60109fbf075aa5f7b which should be just before the change. Test is currently failing.

mcgibbon:
I verified the test runs using a checkpoint from before that commit (the test initially failed). The checkpoint was too large, so I regenerated it, and the code pushed includes a checkpoint from this present code.

fme/core/registry/test_module_registry.py
```diff
@@ -71,3 +78,80 @@ def test_build_conditional():
...
+    )
+    loaded_state_dict = load_or_cache_state(selector_name, module)
+    new_keys = set(to_flat_dict(module.get_state()).keys()).difference(
+        to_flat_dict(loaded_state_dict).keys()
+    )
+    module.load_state(loaded_state_dict)
+    assert not new_keys, (
+        f"New keys added to state dict: {new_keys}. "
+        "Please delete and re-generate the regression target."
+    )
```
jpdunc23: I think it's important that we retain this version of `NoiseConditionedSFNO_state_dict.pt` to avoid regressions in `_add_singleton_group_dim`, especially as we've had two model releases with the older format. 

Maybe you could call it `NoiseConditionedSFNO_state_dict_dbc2925.pt` and parameterize this test including the commit SHA?

More generally, I don't think we should encourage regeneration of the regression target when there are new keys in the state dict, since presumably those new things could lead to big changes in inference behavior. We could maybe handle new keys here on a case-by-case basis, e.g. where we know that given an existing config the new key is just a no-op / identity op, but even that seems risky since we couldn't guarantee that future changes cause that key to modify inference behavior. Ideally, future changes just wouldn't add the new keys to the module unless a new config parameter is turned on, but maybe you have something specific in mind?

mcgibbon: Rather than pick the commit hash before the break, we should retain checkpoints from the specific ACE releases that used NoiseConditionedSFNO.

I think it's important to have a "latest" checkpoint that will break if any backwards incompatibilities are introduced. If the checkpoint we're testing doesn't have a key, then a PR removing that key won't cause the test to fail, but would cause newer checkpoints having that key to no longer load. It is important that this message only happen after we've already tested the current model is backwards-compatible, so that we don't update the checkpoint in a commit that also breaks backwards compatibility.

For this PR I'll just remove this check, we can add these two things in another PR.

PR #830, opened 2026-02-12
Separate regression tests for frozen and latest checkpoints

mcgibbon:
This PR defines separate regression tests for "frozen" checkpoints from specific commits (which shouldn't be updated) and "latest" checkpoints ensuring ongoing backwards compatibility (which should be updated).

### DIFF
--- a/configs/baselines/cm4-ocean/ace-train-config.yaml
+++ b/configs/baselines/cm4-ocean/ace-train-config.yaml
@@ -153,7 +153,8 @@ stepper:
             sea_ice_fraction_name: ocean_sea_ice_fraction
             land_fraction_name: land_fraction
             remove_negative_ocean_fraction: false
-            sea_ice_thickness_name: HI
+            zero_where_ice_free_names:
+            - HI
       next_step_forcing_names:
       - DLWRFsfc
       - DSWRFsfc
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -19,13 +19,23 @@ from fme.core.typing_ import TensorDict, TensorMapping
 class SeaIceFractionConfig:
     """Correct predicted sea_ice_fraction to ensure it is always in 0-1, and
     land_fraction + sea_ice_fraction + ocean_fraction = 1. After
-    sea_ice_fraction is corrected, if the sea_ice_thickness_name is provided
-    it will be set to 0 everywhere the sea_ice_fraction is 0.
+    sea_ice_fraction is corrected, all variables listed in
+    zero_where_ice_free_names will be set to 0 everywhere
+    sea_ice_fraction is 0.
+
+    Parameters:
+        sea_ice_fraction_name: Name of the sea ice fraction variable.
+        land_fraction_name: Name of the land fraction variable.
+        zero_where_ice_free_names: List of variable names to set to 0
+            wherever sea_ice_fraction is 0.
+        remove_negative_ocean_fraction: If True, reduce sea_ice_fraction
+            to prevent ocean_fraction (1 - sea_ice_fraction - land_fraction)
+            from being negative.
     """
 
     sea_ice_fraction_name: str
     land_fraction_name: str
-    sea_ice_thickness_name: str | None = None
+    zero_where_ice_free_names: list[str] = dataclasses.field(default_factory=list)
     remove_negative_ocean_fraction: bool = True
 
     def __call__(
@@ -43,10 +53,8 @@ class SeaIceFractionConfig:
             )
             negative_ocean_fraction = negative_ocean_fraction.clip(max=0)
             out[self.sea_ice_fraction_name] += negative_ocean_fraction
-        if self.sea_ice_thickness_name:
-            thickness = gen_data[self.sea_ice_thickness_name]
-            thickness = thickness * (out[self.sea_ice_fraction_name] > 0.0)
-            out[self.sea_ice_thickness_name] = thickness
+        for name in self.zero_where_ice_free_names:
+            out[name] = gen_data[name] * (out[self.sea_ice_fraction_name] > 0.0)
         return out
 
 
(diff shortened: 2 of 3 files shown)


fme/core/registry/module.py
```diff
@@ -161,7 +161,7 @@ def build(
         n_in_channels: int,
         n_out_channels: int,
         dataset_info: DatasetInfo,
-    ) -> nn.Module:
+    ) -> Module:
```
jpdunc23, line 168: Good catch

fme/core/registry/test_module_registry.py
```diff
@@ -110,40 +146,48 @@ def load_or_cache_state(selector_name: str, module: Module) -> dict[str, torch.T
...
+    Backwards compatibility for frozen releases from specific commits.
+    """
+    set_seed(0)
+    module = FROZEN_BUILDERS[selector_name]()
+    loaded_state_dict = load_state(selector_name)
+    module.load_state(loaded_state_dict)
+
+
+LATEST_BUILDERS = {
+    "NoiseConditionedSFNO": get_noise_conditioned_sfno_module,
```
jpdunc23, line 233: Should `NoiseConditionedSFNO_state_dict.pt` also be updated?

fme/core/registry/test_module_registry.py
```diff
@@ -110,40 +146,48 @@ def load_or_cache_state(selector_name: str, module: Module) -> dict[str, torch.T
...
-        dataset_info=dataset_info,
-    )
+def test_frozen_module_backwards_compatibility(selector_name: str):
+    """
+    Backwards compatibility for frozen releases from specific commits.
+    """
+    set_seed(0)
+    module = FROZEN_BUILDERS[selector_name]()
+    loaded_state_dict = load_state(selector_name)
+    module.load_state(loaded_state_dict)
```
jpdunc23, line 229: I'm still a bit concerned that if we place no limits on new keys then we could inadvertently introduce changes in behavior that lead to regressions in inference skill with `FROZEN_BUILDERS` checkpoints. I wonder if we should also raise an error here with a message along the lines of:

```
"New module parameters {new_keys} found that were not present in the 
"frozen" checkpoint. New module parameters may be added but should 
be enabled by adding a new config parameter that, by default, does not 
add the new parameters when building the module."
```

We could also save and reload the module config dict together with the artifact to verify that the config builds the same architecture. Of course there are a million other ways to change the module code that could lead to inference regressions, but I don't see why we should allow arbitrary new parameters that weren't present when the checkpoint was saved if there is a way to avoid it.

jpdunc23 (commented):
I'm still concerned about silently allowing new parameters with frozen checkpoints, though it maybe warrants a discussion at a tech sync and I'm willing to punt for now if you prefer.

Other than that, I have one blocking comment about updating the existing artifact.

fme/core/registry/test_module_registry.py
mcgibbon: It was already a "latest" version.

fme/core/registry/test_module_registry.py
mcgibbon: That type of thing is supposed to be covered by the "produces the same result" test(s). If you're concerned about new parameters not affecting an initial prediction but affecting later ones after the first gradient update, I should add a second stage to those tests that does a second step when testing for identicality.

I see though now what you're saying, I've not been understanding it. In practice, what I have here won't catch any of the cases I care about updating the regression tests for, because we always add them in a way that sets the weights to None (which doesn't get registered in the state dict). Really what we need is to remember to update/write a new test when we add features that define new weights.

What I actually want to do is test that the _config_ has no new keys, and force the user to build a new latest checkpoint when new config keys are added, saving the asdict'd config with the checkpoint. I'll see about adding that, and also adding what you suggested about making sure the config builds the same architecture.

mcgibbon: I looked at it more, and because we load model states with `strict=True` (the default), this will already error if the built model keys differ from the checkpoint. I can see how my other (wrong) test implied this wasn't the case, though.

fme/core/registry/test_module_registry.py
```diff
@@ -140,10 +144,121 @@ def test_module_backwards_compatibility(selector_name: str):
...
+    if not state_dict_path.exists():
+        raise RuntimeError(
+            f"State dict for {selector_name} not found at {state_dict_path}. "
+            "Please make sure the checkpoint exists and is committed to the repo."
+        )
+    return torch.load(state_dict_path)
+
+
+def load_or_cache_state(
+    selector_name: str, module: Module, module_config: ModuleConfig | None = None
```
jpdunc23, line 180: The `module_config` argument to `load_or_cache_state` appears to be unused.

fme/core/registry/test_module_registry.py
```diff
@@ -79,8 +82,27 @@ def test_module_selector_raises_with_bad_config():
...
+    )
+    vertical_coordinate = HybridSigmaPressureCoordinate(
+        ak=torch.arange(7, device=device), bk=torch.arange(7, device=device)
+    )
+    dataset_info = DatasetInfo(
+        horizontal_coordinates=horizontal_coordinate,
+        vertical_coordinate=vertical_coordinate,
+        timestep=timestep,
+        all_labels=all_labels,
+    )
```
jpdunc23, line 104: Consider adding a new / reusing an existing shared helper to create the `DatasetInfo` that both `get_dbc2925_ncsfno_module` and `get_noise_conditioned_sfno_module` can reuse. 

I think it makes sense to repeat the `select = ModuleSelector(` blocks, even if they're currently identical.

fme/core/registry/test_module_registry.py
jpdunc23: Ah, good point. Maybe we should add a test that confirms this behavior for `Module.load_state` (and maybe also other parts of `Module`). But that's a preexisting issue.

PR #832, opened 2026-02-12
Run parallel distributed tests in CI

mcgibbon:
This PR adds some non-trivial tests of data and model parallelism.

The implementations are trivial for the non-model-parallel case, but these tests will provide good coverage of the upcoming model-parallel code. This PR makes it so that the infrastructure is in place to use these tests and add to them.

Changes:
- Added `get_local_slices`, `data_parallel_rank` and `total_data_parallel_ranks` attributes to Distributed and DistributedBackend classes, which are currently used only in unit tests but will be needed for spatial parallelism.

- [x] Tests added

### DIFF
--- a/fme/core/distributed/non_distributed.py
+++ b/fme/core/distributed/non_distributed.py
@@ -63,7 +63,7 @@ class NonDistributed(DistributedBackend):
         self, tensor: torch.Tensor, gather_list: list[torch.Tensor] | None
     ) -> list[torch.Tensor] | None:
         if gather_list is not None:
-            if len(gather_list) != 0:
+            if len(gather_list) != 1:
                 raise ValueError(
                     f"expected 1 element in gather_list, got {len(gather_list)}"
                 )


fme/core/distributed/distributed.py
```diff
@@ -163,6 +187,20 @@ def gather(self, tensor: torch.Tensor) -> list[torch.Tensor] | None:
...
         return self._distributed.gather(tensor)
 
+    def gather_global(self, tensor: torch.Tensor, global_shape) -> torch.Tensor | None:
+        gathered = self.gather(tensor)
+        if self.is_root():
+            if gathered is None:
+                raise RuntimeError("expected non-none gathered on root rank")
+            gathered_global = torch.zeros(
+                *global_shape, dtype=tensor.dtype, device=tensor.device
+            )
```
elynnwu, line 240: Might be a spatial parallelism PR question, do we know how often we are calling `gather_global`? If allocating 2x the global tensor becomes a memory issue, we may want to consider passing a pre-allocated buffer to the gather call and use it in-place, that way we don't hold both the temporary `gathered` and `gathered_global` at the same time. I don't know if this is an issue, but something we might want to consider in the future.

Makefile
```diff
@@ -53,6 +53,9 @@ create_environment:
 test:
 	pytest --durations 40 .
 
+test_parallel:
+	torchrun --nproc-per-node 2 -m pytest ./fme/core/distributed/parallel_tests
```
elynnwu: [nit] make the number of processor a Make variable so you can use a different number when calling `make test_parallel`

elynnwu (approved):

Makefile
mcgibbon: Updated.

fme/core/distributed/distributed.py
mcgibbon: I updated it as you suggested for performance.

PR #833, opened 2026-02-12
Disable AMP for preconditioner calculations

AnnaKwa:
Previously the entire forward pass of the preconditioner was within the AMP context. However, to preserve precision for small values in the calculation of noise and coefficients these calculations should run with autocast disabled.


fme/downscaling/modules/preconditioners.py
```diff
@@ -86,30 +86,31 @@ def forward(
...
+            class_labels = (
+                None
+                if self.label_dim == 0
+                else torch.zeros([1, self.label_dim], device=x.device)
+                if class_labels is None
+                else class_labels.to(torch.float32).reshape(-1, self.label_dim)
+            )
+            dtype = (
+                torch.float16
+                if (self.use_fp16 and not force_fp32 and x.device.type == "cuda")
```
mcgibbon: Question: Should we just remove the `use_fp16` flag? is it used?

fme/downscaling/modules/preconditioners.py
```diff
@@ -121,5 +122,6 @@ def forward(
             raise ValueError(
                 f"Expected the dtype to be {dtype}, but got {F_x.dtype} instead."
             )
-        D_x = c_skip * x + c_out * F_x.to(torch.float32)
-        return D_x
+        with torch.amp.autocast(device_type=get_device().type, enabled=False):
+            D_x = c_skip * x + c_out * F_x.to(torch.float32)
+            return D_x
```
mcgibbon:
```suggestion
        return D_x
```
nit: a little confusing what the implications are of a return within a context

fme/downscaling/modules/preconditioners.py
AnnaKwa: Removed- we never use it.

PR #834, opened 2026-02-13
Define requirements in pyproject.toml instead of .txt files

oliverwm1:
Follow the modern python style of defining requirements in `pyproject.toml` instead of in separate text files that are referenced from within `pyproject.toml`. The one exception is that `requirements-healpix.txt` must remain separate because it points to a GitHub URL, and so we can't have it be declared dependency of the `fme` package or else we can't release `fme` to PyPI.

Changes:
- delete `requirements.txt`, `requirements-dev.txt`, `requirements-deploy.txt` and `docs/requirements.txt` and move the corresponding information to `pyproject.toml`

- [x] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated (NOTE: although dependencies didn't change, I updated the deps-only image due to the change in the Dockerfile and wanting to verify that gantry submissions still work)

### DIFF
--- a/.github/workflows/beaker.yaml
+++ b/.github/workflows/beaker.yaml
@@ -26,7 +26,7 @@ jobs:
       PYTHON_VERSION: "3.11"
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
 
       - name: Get commit SHA and set job name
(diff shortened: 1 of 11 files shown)


mcgibbon (approved):
LGTM!

PR #837, opened 2026-02-13
Report gpu benchmark results to wandb

mcgibbon:
This PR adds a --wandb-project <entity>/<project> argument to `fme.core.benchmark.run`, and runs it automatically during GPU CI on merges to main.

- [ ] Tests added


mcgibbon:
Ran this on some test code https://wandb.ai/ai2cm/fme-core-benchmark-test
Note the first one was before I limited the benchmarks to max_depth=1 so there are quite a few keys.

mcgibbon:
Automatic runs go to the workspace here, dots will turn into bar charts once additional runs/commits are added: https://wandb.ai/ai2cm/fme-core-benchmarks-ci

fme/core/benchmark/run.py
```diff
@@ -77,30 +83,49 @@ def get_label(name):
...
+        wandb = WandB.get_instance()
+        wandb.configure(log_to_wandb=True)
+        wandb_name = f"{get_git_commit()}-{safe_device_name}"
+        wandb.init(
+            resumable=False,
+            project=project,
+            entity=entity,
+            name=wandb_name,
+        )
+        wandb.log(wandb_logs, commit=True)
```
oliverwm1, line 128: FWIW, if `step` arg to wandb.log is None, then the default for `commit` is True, so I don't think the change to `fme.core.wandb` is necessary. https://docs.wandb.ai/models/ref/python/experiments/run#method-run-log

oliverwm1 (approved):
LGTM—I don't think the change to `fme.core.wandb` is necessary, but it's also fine to leave in.

.github/workflows/tests.yaml
```diff
@@ -93,3 +93,12 @@ jobs:
...
             make test_parallel
+      - name: Run GPU benchmarks
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        env:
+          WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
+        run: |
+            source .venv/bin/activate
+            export PATH=/usr/local/nvidia/bin:${PATH}
+            export LD_LIBRARY_PATH=/usr/local/nvidia/lib64:${LD_LIBRARY_PATH}
+            python3 -m fme.core.benchmark.run --wandb-project ai2cm/fme-core-benchmarks-ci
```
oliverwm1, line 104: how long does this take to run?

mcgibbon: It took 32s in this PR, which will grow as we add more benchmarks. I think there's room to bring the per-benchmark time downwards with fewer repeats, but for now I didn't want to push the envelope on repeatability for the timings. https://github.com/ai2cm/ace/actions/runs/22003167949/job/63580581294

fme/core/benchmark/run.py
mcgibbon: Thanks for pointing that out! I think I will leave it since it's nice to be explicit.

We should consider adding a `commit=True` to our last inference end-of-epoch log, since those currently don't get published to the web until the next 100 batches run (which has made me wait for the logs a couple times).

PR #839, opened 2026-02-13
Refactor distributed to avoid rank arg to get_local_slices

mcgibbon:
It is difficult to implement a rank argument to get_local_slices for the spatial parallel case.

In this PR, we work around the issue by removing the rank arg to get_local_slices, and gathering local slices instead.

Changes:
- Added `gather_object` to Distributed and DistributedBackend
- Removed `rank` argument to `get_local_slices` on Distributed and DistributedBackend
- Add support for FME_FORCE_CPU=1 environment variable to force CPU usage, used to run parallel tests on machines with 1 GPU

- [x] Tests added

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -93,3 +93,12 @@ jobs:
             python3 fme/require_gpu.py
             make test
             make test_parallel
+      - name: Run GPU benchmarks
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        env:
+          WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
+        run: |
+            source .venv/bin/activate
+            export PATH=/usr/local/nvidia/bin:${PATH}
+            export LD_LIBRARY_PATH=/usr/local/nvidia/lib64:${LD_LIBRARY_PATH}
+            python3 -m fme.core.benchmark.run --wandb-project ai2cm/fme-core-benchmarks-ci
--- a/fme/core/benchmark/benchmark.py
+++ b/fme/core/benchmark/benchmark.py
@@ -40,6 +40,13 @@ class BenchmarkResult:
         except AssertionError as e:
             raise AssertionError(f"Memory results differ: {e}") from e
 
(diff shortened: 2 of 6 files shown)


fme/core/distributed/parallel_tests/test_local_slices.py
```diff
@@ -46,14 +44,19 @@ def test_local_slices_subdivide_domain():
...
-        x_global, torch.ones_like(x_global)
-    )  # the entire domain should get selected
+    local_slices = dist.get_local_slices(global_shape, data_parallel_dim=0)
+    gathered_local_slices = dist.gather_object(local_slices)
+    if dist.is_root():
+        assert gathered_local_slices is not None
+        for i in range(dist.world_size):
+            # the slices should be of the minimum size required
+            assert x_global[gathered_local_slices[i]].nelement() == expected_slice_size
+            x_global[gathered_local_slices[i]] = 1
```
elynnwu: make this a distinct value per rank instead of just 1s, that way you also test that it has the correct gather order

mcgibbon: I think we need some more thought on this. FWIW the test is pre-existing, this is just a refactor to handle the changed API for getting local slices.

Making it a distinct value per rank would make the output dependent on the rank decomposition, which I want to avoid. The decomposition-independent test I thought of for positioning was to define the global domain as a reshaped `arrange(num_elements)`, like in the `reduce_mean` test below.

While we can add an additional test, I think it's important to keep this test as it is, testing only the two properties "are the local slices of the minimum size needed" and "do the local slices tile the domain completely".

The way I've written the gather now and the "get local slices for other ranks", a test of "do we have the correct gather order" is actually a test of "do gather_object and gather return rank data in the same order as one another". I can add a test for that to this PR. A test like "rank 2 should have exactly this local slice" I think should be made as a unit test on the decomposition code, ideally one that runs in the serial tests.

elynnwu: Makes sense, thanks for clarifying and adding the additional tests.

PR #840, opened 2026-02-13
Recover sea ice thickness from volume

jpdunc23:
Adds `HI` as an ocean derived variable when `sea_ice_volume` is predicted but not the thickness.

Changes:
- `HorizontalCoordinates` now has a `area_weights_m2` property and `LatLonCoordinates.area_weights_m2` returns a 2D tensor of grid cell area in meters squared.
- `VerticalCoordinate.build_derive_function` now accepts an optional `horizontal_coordinates` argument.
- Added `area_weights_m2`, `sea_ice_thickness`, and `sea_ice_volume` properties to `OceanData`.

- [x] Tests added


fme/core/constants.py
```diff
@@ -16,3 +16,5 @@
 DENSITY_OF_WATER_CM4 = 1035.0  # kg/m^3
 
 FREEZING_TEMPERATURE_KELVIN = 273.15  # K
+
+EARTH_RADIUS = 6378000.0  # m
```
spencerkclark: This can vary by small amounts depending on the model, but we should probably go with `6371000.0` to match CM4, [as defined in the FMS constants](https://github.com/NOAA-GFDL/FMS/blob/957b7c9afd1ffff9cd9e7dd23d1ae818b1c3cdc8/constants/gfdl_constants.fh#L35).

jpdunc23: This is the value used by CM4.

fme/core/ocean_data.py
```diff
@@ -17,7 +18,7 @@
         "sea_surface_height_above_geoid": ["zos"],
         "sea_surface_temperature": ["sst"],
         "sea_ice_fraction": ["sea_ice_fraction"],
-        "sea_ice_thickness": ["sea_ice_thickness"],
+        "sea_ice_thickness": ["HI"],
```
jpdunc23, line 21: The key "sea_ice_thickness" was unused but the value was incorrect for CM4.

fme/core/constants.py
spencerkclark: It will not make that much of a difference, but I think from the perspective of cell areas the ocean model is using a value implicit in the grid loaded via [`set_grid_metrics_from_mosaic`](https://github.com/NOAA-GFDL/MOM6/blob/0f977a412b3532626be4be826a5063836586a7cb/src/initialization/MOM_grid_initialize.F90#L167-L350) rather than the default radius defined in the MOM6 code.  The loaded grid was generated with FRE-NCtools, [which uses a hard-coded radius of `6.371e6`](https://github.com/NOAA-GFDL/FRE-NCtools/blob/4a11fd6401a3e201f646ba94653f510d2aee92d9/lib/libfrencutils/constant.h#L24), which aligns with [the radius defined in FMS that is used by the atmosphere model](https://github.com/NOAA-GFDL/FMS/blob/9d25a1e4f5e4b4040b513b5d40b09d7b6f904fbf/constants/constants.F90#L102).
```suggestion
EARTH_RADIUS = 6371000.0  # m
```

fme/core/ocean_data.py
```diff
@@ -216,3 +229,50 @@ def ocean_fraction(self) -> torch.Tensor:
...
+        except KeyError:
+            sfrac = self.sea_surface_fraction
+            sea_ice_vol = self.sea_ice_volume
+            try:
+                # more accurate for CM4; see compute_coupled_sea_ice
+                # in scripts/data_process/coupled_dataset_utils.py
+                sea_ice_frac = self.ocean_sea_ice_fraction * sfrac
+            except KeyError:
+                sea_ice_frac = self.sea_ice_fraction
+                sea_ice_vol = sea_ice_vol * sfrac
```
spencerkclark: Could you clarify why the `sea_ice_vol` is scaled by `sfrac` in this case?

fme/core/test_ocean_derived_variables.py
```diff
@@ -151,3 +151,36 @@ def test_metadata_registry():
         metadata["ocean_heat_content"].long_name
         == "Column-integrated ocean heat content"
     )
+
+
+def test_sea_ice_thickness_derived_variable():
```
spencerkclark: Should you add test coverage for the try-except logic in `OceanData.sea_ice_thickness`?

fme/core/ocean_data.py
jpdunc23: I realized this was incorrect. The two paths are now equivalent*. Here's the updated version:
https://github.com/ai2cm/ace/blob/d4f4382fd939da608ac339d77b202d0dae73e437/fme/core/ocean_data.py#L252-L260

*Note that this assumes that `self.sea_ice_fraction` is the output from the [coupled data processing script](https://github.com/ai2cm/ace/blob/0e14832cf10a30ca4dc26233d935aa1609770ede/scripts/data_process/coupled_dataset_utils.py#L528-L539). If it is instead the output from `compute_ocean_dataset.py` or `compute_sea_ice_dataset.py`, then this will have large biases near coastlines, assuming that `sea_surface_fraction != 1 - land_fraction`, which is the case for CM4.

For E3SM, we have `sea_surface_fraction = 1 - land_fraction`, and the coupled `sea_ice_fraction` output comes from the atmosphere data processing which uses the `fregrid` + "remove negative ocean fraction" pipeline. However, the `sea_ice_fraction` output from `compute_sea_ice_dataset_e3sm.py` looks to me like what we call `ocean_sea_ice_fraction`, the fraction of the ocean surface with sea ice:
image

Regardless, I think both paths should now give good / equivalent results for CM4 and E3SM assuming `sea_ice_fraction` is the version from coupled data processing.

spencerkclark: Sorry, now I am a bit confused how the land fraction enters in the denominator on its own (e.g. over open ocean it will be zero, so the result would be undefined).  Should it be `1 - lfrac`?  Is this scaling meant to account for a difference between `sfrac` and `1 - lfrac`?

fme/core/constants.py
spencerkclark: Upon further inspection, I am actually not sure about the provenance of the native ocean grid file; while a tripolar grid can be generated with FRE-NCtools, the ocean grid defined in the file seems to assume a radius of `6378000`, which is consistent with the MOM6 code.  See discussion in https://github.com/m2lines/ocean_emulators/issues/84.

The atmosphere assumes a radius of `6371000`, however—we can see this in the code and infer it from its native grid files—so it's somewhat surprising that the two coupled components could be different from each other in this regard.

fme/core/ocean_data.py
jpdunc23: Oh, sorry, yes that is a typo and it should be `1 - lfrac`.

spencerkclark: Oh good, thanks for clarifying.

PR #841, opened 2026-02-14
Add UI, VI, and sfdsi to the CM4 coupled datasets

jpdunc23:
Adds sea ice velocities (`UI`, `VI`), and downward sea ice basal salt flux (`sfdsi`) to the CM4 coupled datasets and improves the usability of the coupled data processing script.

Changes:
- Adds `ocean_extra_fill_values: dict[str, float]` to `CoupledSeaSurfaceConfig`.
- Improve `scripts/create_coupled_datasets.py` by reading existing coupled datasets from disk rather than re-computing them.
- Add a couple of missing dependencies for the `create_coupled_datasets` conda env.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,3 +46,18 @@ committing code: `pre-commit install`.
 
 You can run local tests with `pytest`. As a shortcut for running only faster unit
 tests, use `make test_very_fast` from the root of the repository.
+
+## Internal Development
+
+If you are making changes directly to ai2cm/ace, please follow these internal
+development guidelines.
+
+When making new branches, use the naming convention:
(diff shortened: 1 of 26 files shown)


scripts/data_process/coupled_dataset_utils.py
elynnwu, line 298: Update doc to include `name`

scripts/data_process/requirements-coupled.txt
```diff
@@ -1,10 +1,12 @@
+bokeh>=3.1.0
```
elynnwu, line 1: Just curious where is bokeh used?

jpdunc23: This is required by dask dashboard.

PR #719, opened 2026-01-05
[WIP] Spatial parallelism unit test

elynnwu:
Short description of why the PR is needed and how it satisfies those requirements, in sentence form.

Changes:
- symbol (e.g. `fme.core.my_function`) or script and concise description of changes or added feature
- Can group multiple related symbols on a single bullet

- [ ] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

Resolves #<github issues> (delete if none)

PR #842, opened 2026-02-16
add ModelTorchDistributed backend for spatial parallelism along tests

mahf708:
add ModelTorchDistributed backend to begin the process of enabling training and inferencing with a spatial parallelized context.

Changes:
- In fme.core.distributed:
  - ModelTorchDistributed (DistributedBackend) in model_torch_distributed module
  - module model_torch_distributed_comm contains comm utilities for ModelTorchDistributed
  - model_torch_distributed_utils contains tensor utilities for ModelTorchDistributed
- Expanded fme.core.distributed.parallel_tests and verified with NPROC=1,2,3,4 (on 4xA100 pod)
- Updated signature of base classes as needed

- [x] Tests added
- [x] *Optional* (for now unpinned) dependency on physicsnemo

Closes #749

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,11 +12,11 @@ jobs:
   cpu:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Set environment variables
         run: |
           echo "CURRENT_WEEK=$(date +'%Y-%U')" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
       - uses: actions/cache@v4
@@ -38,11 +38,11 @@ jobs:
   cpu-very-fast-no-healpix:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Set environment variables
         run: |
           echo "CURRENT_WEEK=$(date +'%Y-%U')" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
       - uses: actions/cache@v4
@@ -62,11 +62,11 @@ jobs:
     # https://github.com/ai2cm/long-lived-infrastructure#installing-the-github-actions-runner-controller
     runs-on: arc-runner-set
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Set environment variables
         run: |
           echo "CURRENT_WEEK=$(date +'%Y-%U')" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
       - uses: actions/cache@v4
--- a/.github/workflows/beaker.yaml
+++ b/.github/workflows/beaker.yaml
@@ -26,7 +26,7 @@ jobs:
       PYTHON_VERSION: "3.11"
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
 
       - name: Get commit SHA and set job name
         run: |
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -12,11 +12,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Set environment variables
         run: |
           echo "CURRENT_WEEK=$(date +'%Y-%U')" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
       - uses: actions/cache@v4
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v6
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: "3.11"
 
--- a/Makefile
+++ b/Makefile
@@ -52,17 +52,17 @@ create_environment:
 	conda run --no-capture-output -n $(ENVIRONMENT_NAME) uv pip install -r analysis-deps.txt
 
 test:
-	pytest --durations 40 .
+	pytest -n 4 --durations 40 .
 
 test_parallel:
 	torchrun --nproc-per-node $(NPROC) -m pytest ./fme/core/distributed/parallel_tests
 
 # --cov must come  after pytest args to use the sources defined by config
 test_cov:
-	pytest --durations 40 --cov --cov-report=term-missing:skip-covered --cov-config=pyproject.toml .
+	pytest -n 4 --durations 40 --cov --cov-report=term-missing:skip-covered --cov-config=pyproject.toml .
 
 test_fast:
-	pytest --durations 40 --fast .
+	pytest -n 4 --durations 40 --fast .
 
 test_very_fast:
 	pytest --durations 40 --very-fast .
--- a/fme/ace/step/fcn3.py
+++ b/fme/ace/step/fcn3.py
@@ -173,6 +173,7 @@ class FCN3StepConfig(StepConfigABC):
         default_factory=lambda: AtmosphereCorrectorConfig()
     )
     next_step_forcing_names: list[str] = dataclasses.field(default_factory=list)
+    prescribed_prognostic_names: list[str] = dataclasses.field(default_factory=list)
     residual_prediction: bool = False
 
     def __post_init__(self):
@@ -203,6 +204,12 @@ class FCN3StepConfig(StepConfigABC):
             self.forcing_names + self.atmosphere_input_names + self.surface_input_names
         )
         self.out_names = self.atmosphere_output_names + self.surface_output_names
+        for name in self.prescribed_prognostic_names:
+            if name not in self.out_names:
+                raise ValueError(
+                    f"prescribed_prognostic_name '{name}' must be in out_names: "
+                    f"{self.out_names}"
+                )
 
     @property
     def n_ic_timesteps(self) -> int:
@@ -264,10 +271,11 @@ class FCN3StepConfig(StepConfigABC):
     @property
     def next_step_input_names(self) -> list[str]:
         """Names of variables provided in next_step_input_data."""
-        input_only_names = set(self.input_names).difference(self.output_names)
-        if self.ocean is None:
-            return list(input_only_names)
-        return list(input_only_names.union(self.ocean.forcing_names))
+        result = set(self.input_names).difference(self.output_names)
+        if self.ocean is not None:
+            result = result.union(self.ocean.forcing_names)
+        result = result.union(self.prescribed_prognostic_names)
+        return list(result)
 
     @property
     def loss_names(self) -> list[str]:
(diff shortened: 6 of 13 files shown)


mahf708:
This builds on top of @odiazib's PR (#749) and previous draft work (#719). I decided to significantly increase test coverage; arguably overkill, but I thought these tests would help with the planned upcoming dev (data loading/writing, training operations, layering, etc.). There are some tricky (and very not pretty) parts in how we are borrowing comm utilities from makani, but they are hidden enough for now. In particular, in order to facilitate efficient/easy testing, some nontrivial hacks were needed. Some of the tests are kind of trivial in nature (e.g., barrier) and they balloon the size of this PR by a fair bit. Hopefully we will keep future PRs to <1000 lines. Copying @odiazib, @elynnwu,  and @mcgibbon for awareness. Feedback welcome.

Tested locally with `make parallel_tests NPROC=X` for X=1,2,3,4. Also, ran full test suite on A100 GPUs as in the CI files. All passing as far as I could tell (except with timed ones running overtime).

.github/workflows/tests.yaml
```diff
@@ -53,6 +53,7 @@ jobs:
         run: |
           python -m pip install uv
           uv pip install --system -c constraints.txt -e .[dev]
+          uv pip install --system --no-build-isolation -c constraints.txt -e .[dev,spatial-parallelism]
```
mcgibbon, line 56: Is it possible to run the very-fast tests without spatial parallelism? It's nice to have a test that doesn't use them to make sure the "optional" part of optional dependency is done correctly.

fme/core/distributed/parallel_tests/test_gather_not_supported.py
```diff
@@ -0,0 +1,64 @@
...
+import pytest
+import torch
+
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+from fme.core.distributed.parallel_tests._helpers import requires_parallel
+
+
+@requires_parallel
+def test_gather_raises_not_implemented(monkeypatch):
```
mcgibbon, line 19: Suggestion: Make this a unit test of the ModelTorchDistributed backend, rather than an integration test involving Distributed.

fme/core/distributed/parallel_tests/test_gather_not_supported.py
```diff
@@ -0,0 +1,64 @@
...
+    device = get_device()
+
+    tensor = torch.ones((2, 2), dtype=torch.float32, device=device)
+
+    with pytest.raises(NotImplementedError, match="gather.*spatial parallelism"):
+        dist.gather(tensor)
+
+
+@requires_parallel
+def test_gather_irregular_raises_not_implemented(monkeypatch):
```
mcgibbon, line 35: Question: Do we need tests for these NotImplementedError's? We do plan to implement them.

fme/core/distributed/parallel_tests/test_model_torch_distributed.py
```diff
@@ -0,0 +1,100 @@
...
+        monkeypatch.setenv("H_PARALLEL_SIZE", h_parallel)
+    else:
+        monkeypatch.delenv("H_PARALLEL_SIZE", raising=False)
+
+    if w_parallel is not None:
+        monkeypatch.setenv("W_PARALLEL_SIZE", w_parallel)
+    else:
+        monkeypatch.delenv("W_PARALLEL_SIZE", raising=False)
+
+    result = ModelTorchDistributed.is_available()
```
mcgibbon, line 51: Suggestion: If the logic can be refactored to a helper in some way, or you have `is_available` take input arguments that default to the environment variables, you can unit test this without environment monkeypatching.

fme/core/distributed/distributed.py
```diff
@@ -44,6 +48,21 @@ def get_instance(cls) -> "Distributed":
             singleton = cls()
         return singleton
 
+    @classmethod
+    def reset(cls) -> None:
```
mcgibbon, line 52: I am nervous about something like this, because distributed backends generally use global state which doesn't get properly reset when we do something like this.

fme/core/distributed/distributed.py
```diff
@@ -105,6 +124,9 @@ def world_size(self) -> int:
         """
         return self._distributed.total_ranks
 
+    def get_local_rank(self) -> int:
```
mcgibbon, line 127: Question: What is a "local" rank?

fme/core/distributed/distributed.py
```diff
@@ -105,6 +124,9 @@ def world_size(self) -> int:
...
+        return self._distributed.get_local_rank()
+
     def get_sampler(
```
mcgibbon, line 130: It would be great to add a test of this, I think you're right about the changes.

fme/core/distributed/distributed.py
```diff
@@ -126,7 +148,7 @@ def local_batch_size(self, batch_size: int) -> int:
         """
         return self._distributed.local_batch_size(batch_size)
 
-    def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor:
+    def reduce_mean(self, tensor: torch.Tensor, group=None) -> torch.Tensor:
```
mcgibbon, line 151: Suggestion: We could abstract away concerns like specific group names to make it easier on users. For example, can we use a `data_parallel_only: bool = True` argument instead? or a `kind: Literal["data_parallel", "model", "all"] = "data_parallel"`?

fme/core/distributed/distributed.py
```diff
@@ -169,26 +191,28 @@ def get_local_slices(
 
     def reduce_sum(self, tensor: torch.Tensor) -> torch.Tensor:
```
mcgibbon, line 192: Generally speaking we need reductions along the data-parallel dimension to get global maps of mean outputs in our aggregators. We probably also need to do global area-weighted means for certain operations in the correctors.

mahf708: We also do need to reudce over other orthogonal dimensions (say for zonal means), but that can wait a bit

fme/core/distributed/parallel_tests/test_local_batch_size.py
```diff
@@ -0,0 +1,74 @@
...
+    dist = Distributed.get_instance()
+
+    global_batch = 32
+    data_group_size = dist.total_data_parallel_ranks
+    assert data_group_size >= 1
+    expected = global_batch // data_group_size
+    assert dist.local_batch_size(global_batch) == expected
+
+
+def test_local_batch_size_not_divisible():
```
mcgibbon, line 53: Does this test need to be run in a parallel context or would it work in serial?

fme/core/distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,275 @@
...
+    def total_data_parallel_ranks(self) -> int:
+        return comm.get_size("data")
+
+    def get_local_rank(self) -> int:
+        return self._device_id
+
+    def get_local_slices(
+        self,
+        tensor_shape,
+        rank: int | None = None,
```
mcgibbon, line 104: Issue: Both the `rank` and `data_parallel_dim` argument here are ignored.

Suggestion: I have this PR to let the `rank` argument get removed: https://github.com/ai2cm/ace/pull/839

I don't know what needs to be done yet about data_parallel_dim.

fme/core/distributed/distributed.py
```diff
@@ -311,5 +335,14 @@ def get_seed(self) -> int:
...
         return self._distributed.shutdown()
 
+    def comm_get_size(self, key: str):
+        return self._distributed.comm_get_size(key)
+
+    def comm_get_group(self, key: str):
+        return self._distributed.comm_get_group(key)
+
+    def comm_get_rank(self, key: str):
+        return self._distributed.comm_get_rank(key)
```
mcgibbon, line 345: Issue: These are low-level operations. Can we hide them inside backend methods? Other backends don't have a comm.

fme/core/distributed/model_torch_distributed_comm.py
```diff
@@ -0,0 +1,220 @@
...
+import os
+import time
+import logging
+import math
+from typing import Union
+import numpy as np
+
+# we are using the distributed manager from physicsnemo
+from physicsnemo.distributed.manager import DistributedManager
+from physicsnemo.distributed.config import ProcessGroupNode, ProcessGroupConfig
```
mcgibbon, line 44: If this is all we're using from `physicsnemo`, we could consider vendorizing/forking the code, i.e. copy-pasting it into a subdirectory here. The last time I checked these seemed pretty isolated / didn't depend on other infrastructure in physicsnemo, and it would avoid a dependency.

fme/core/distributed/non_distributed.py
```diff
@@ -82,3 +90,12 @@ def barrier(self):
 
     def shutdown(self):
         return
+
+    def comm_get_size(self, key: str):
+        return 1
+
+    def comm_get_group(self, key: str):
+        return None
```
mcgibbon, line 98: Question: Is this a valid return value for `comm_get_group`?

fme/core/distributed/base.py
```diff
@@ -105,6 +111,15 @@ def wrap_module(self, module: torch.nn.Module) -> torch.nn.Module:
         """
         ...
 
+    @abstractmethod
+    def comm_get_size(self, key: str): ...
```
mcgibbon, line 115: Issue: These need return value type hints.

fme/core/distributed/distributed.py
```diff
@@ -311,5 +335,14 @@ def get_seed(self) -> int:
     def shutdown(self):
         return self._distributed.shutdown()
 
+    def comm_get_size(self, key: str):
```
mcgibbon, line 338: These comm_ methods are only used in tests, do we need them / do we need them public on Distributed?

fme/core/distributed/non_distributed.py
```diff
@@ -40,13 +40,21 @@ def total_ranks(self) -> int:
...
         return tuple(slice(None, None) for _ in tensor_shape)
 
+    def get_local_rank(self):
+        return 0
+
     def local_batch_size(self, batch_size: int) -> int:
         return batch_size
 
-    def reduce_mean(self, tensor: torch.Tensor) -> torch.Tensor:
+    def reduce_mean(self, tensor: torch.Tensor, group=None) -> torch.Tensor:
```
mcgibbon, line 57: I would like to avoid adding low-level information like the group names above this PR, but we should talk about it. For an initial PR we should avoid features that need this.

fme/core/gridded_ops.py
```diff
@@ -10,6 +10,7 @@
 from fme.core.cuhpx.sht import SHT as CuHpxSHT
 from fme.core.cuhpx.sht import iSHT as CuHpxiSHT
 from fme.core.device import get_device
+from fme.core.distributed import Distributed
```
mcgibbon, line 13: These changes to `gridded_ops` and their associated tests in parallel_tests can/probably should be split into their own PR.

fme/core/distributed/parallel_tests/test_coordinates_sp.py
```diff
@@ -0,0 +1,81 @@
+import logging
```
mcgibbon, line 1: This test should probably be split off with the gridded_ops changes into its own PR.

fme/core/distributed/parallel_tests/test_get_local_slices.py
```diff
@@ -0,0 +1,82 @@
+import pytest
```
mcgibbon, line 1: What coverage does this file add that isn't already covered by `test_local_slices.py`?

fme/core/distributed/parallel_tests/test_local_batch_size.py
mcgibbon, line 1: It would be nice if we could test this code in a unit test without the full process of setting up Distributed. But it's probably a test of DistributedManager.

fme/core/distributed/parallel_tests/test_loss_sp.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+    error_tol = 1e-7
+    nx = 8
+    ny = 8
+
+    torch.manual_seed(0)
+
+    # Create test data
+    data_tensor_host = torch.randn(1, 2, nx, ny, device="cpu")
+    area_weights_host = torch.ones(nx, ny).to("cpu") * 5
```
mcgibbon, line 41: Issue: Please update area weights to not use ones (maybe 1 plus a random uniform?) to cover how area is applied.

fme/core/distributed/parallel_tests/test_local_slices.py
mcgibbon, line 74: I don't understand how this test passes under spatial parallelism, given this isn't a reduction over the data-parallel group, and the test is supposed to require that it is.

mahf708: this never passes under sp :) because it is slyly being skipped by the pytest hackery 😺 (which we will get rid of in the next edit)

mahf708: For the record, I am not surprised at all the cpu tests bombed. I didn't even run a single one of these tests, and I totally forgot about that until after I stopped working on this. In the revision, will address those too

PR #843, opened 2026-02-16
Add `zero_where_ice_free_names` to ocean corrector's `SeaIceFractionConfig`

jpdunc23:
Deprecates `sea_ice_thickness_name` of the ocean corrector's `SeaIceFractionConfig` in favor of the more general `zero_where_ice_free_names` to allow multiple outputs to be corrected based on the predicted sea ice.

- [x] Tests added

### DIFF
--- a/configs/baselines/cm4-ocean/ace-train-config.yaml
+++ b/configs/baselines/cm4-ocean/ace-train-config.yaml
@@ -153,7 +153,8 @@ stepper:
             sea_ice_fraction_name: ocean_sea_ice_fraction
             land_fraction_name: land_fraction
             remove_negative_ocean_fraction: false
-            zero_where_ice_free_names: [HI]
+            zero_where_ice_free_names:
+            - HI
       next_step_forcing_names:
       - DLWRFsfc
       - DSWRFsfc


fme/core/corrector/ocean.py
```diff
@@ -99,6 +107,14 @@ def remove_deprecated_keys(cls, state: Mapping[str, Any]) -> dict[str, Any]:
...
             else:
                 state_copy["ocean_heat_content_correction"] = None
+        if "sea_ice_fraction_correction" in state_copy:
+            sif = state_copy["sea_ice_fraction_correction"]
+            if isinstance(sif, dict) and "sea_ice_thickness_name" in sif:
+                thickness_name = sif.pop("sea_ice_thickness_name")
+                if thickness_name is not None:
+                    sif.setdefault("zero_where_ice_free_names", []).append(
+                        thickness_name
+                    )
```
jpdunc23, line 117: Since we use `OceanCorrectorConfig.from_state` during training when the ocean vertical coordinate builds the corrector, this preserves backwards compatibility for training as well as inference. This is helpful for our coupled training scripts which fill in the stepper configs from the component pretraining checkpoints' yaml config. @elynnwu

configs/baselines/cm4-ocean/ace-train-config.yaml
```diff
@@ -153,7 +153,7 @@ stepper:
             sea_ice_fraction_name: ocean_sea_ice_fraction
             land_fraction_name: land_fraction
             remove_negative_ocean_fraction: false
-            sea_ice_thickness_name: HI
+            zero_where_ice_free_names: [HI]
```
elynnwu: [nit] might be better to stick with the same style of itemizing this to be consistent with the rest of the config
```
zero_where_ice_free_names:
- HI
```

PR #846, opened 2026-02-17
Allow downscaling to load merged datasets

AnnaKwa:
These changes allow the downscaling data loader configs to load merged datasets. 

- [x] Tests added

### DIFF
--- a/.github/workflows/beaker.yaml
+++ b/.github/workflows/beaker.yaml
@@ -15,7 +15,8 @@ jobs:
   ace-evaluator:
     runs-on: ubuntu-latest
     # Only deploy on manual trigger or push to main
-    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+    # Additionally, skip if the target repo isn't ai2cm/ace, lest it tries to run in forks
+    if: ${{ github.repository == 'ai2cm/ace' && (github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')) }}
     env:
       CONFIG_PATH: docs/evaluator-config.yaml
       BEAKER_TOKEN: ${{ secrets.BEAKER_TOKEN }}
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,7 +34,29 @@ jobs:
         run: |
           source .venv/bin/activate
           make test_cov
-          make test_parallel
+          # matrix for parallel tests
+          # | backend | WS | H | W |
+          # | ------- | -- | - | - |
+          # | torch   |  2 | 1 | 1 |
+          # | torch   |  3 | 1 | 1 |
+          # | model   |  3 | 1 | 1 |
+          # | model   |  2 | 2 | 1 |
+          # | model   |  2 | 1 | 2 |
+          # | model   |  3 | 3 | 1 |
+          # | model   |  3 | 1 | 3 |
+          # | model   |  4 | 2 | 2 |
+          # | model   |  4 | 2 | 1 |
+          # | model   | 18 | 3 | 3 |
+          NPROC=2  FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=2 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=3 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=3 make test_parallel
+          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=2 make test_parallel
+          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=18 FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=3 FME_DISTRIBUTED_W=3 make test_parallel
   cpu-very-fast-no-healpix:
     runs-on: ubuntu-latest
     steps:
@@ -59,6 +81,8 @@ jobs:
   gpu:
     # These steps run on our self-hosted k8s GPU runners. See below for setup notes
     # https://github.com/ai2cm/long-lived-infrastructure#installing-the-github-actions-runner-controller
+    # Additionally, skip if the target repo isn't ai2cm/ace, lest it tries to run in forks
+    if: ${{ github.repository == 'ai2cm/ace' }}
     runs-on: arc-runner-set
     steps:
       - uses: actions/checkout@v6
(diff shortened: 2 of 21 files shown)


fme/downscaling/data/config.py
```diff
@@ -365,6 +410,18 @@ class PairedDataLoaderConfig:
...
+    def _first_data_config(
+        self,
+        config: XarrayDataConfig | MergeDatasetConfig,
+    ) -> XarrayDataConfig:
+        """Return the first XarrayDataConfig for data_path/file_pattern lookup."""
+        if isinstance(config, XarrayDataConfig):
+            return config
+        first = config.merge[0]
+        if isinstance(first, ConcatDatasetConfig):
+            return first.concat[0]
```
jpdunc23: Since you already handle concatenation in `build_from_config_sequence`, you could maybe use `MergeNoConcatDatasetConfig` instead of `MergeDatasetConfig`. We use `MergeNoConcatDatasetConfig` for training in `fme.coupled` since we also handle concatenation there.

fme/downscaling/inference/output.py
```diff
@@ -165,8 +166,9 @@ def build(
 
     @staticmethod
     def _single_xarray_config(
-        coarse: list[XarrayDataConfig]
-        | Sequence[XarrayDataConfig | XarrayEnsembleDataConfig],
+        coarse: Sequence[
```
AnnaKwa, line 169: Currently inference contains a check that the coarse data config is a single `XarrayDataConfig`. This can be updated later in a subsequent PR to allow a merged config as well.

fme/downscaling/data/config.py
```diff
@@ -462,12 +547,14 @@ def build(
                 # TODO: change to use full static inputs list
                 fine_topography = static_inputs_from_checkpoint[0]
             elif self.topography is None:
-                data_path = self.fine[0].data_path
-                file_pattern = self.fine[0].file_pattern
-                raw_paths = get_raw_paths(data_path, file_pattern)
+                first_config = self._first_data_config(self.fine[0])
```
AnnaKwa, line 532: This block will be removed in a subsequent PR where the static inputs are loaded using the paths provided in a training config field.

fme/downscaling/data/config.py
```diff
@@ -547,10 +634,10 @@ def build(
             dataset_coarse_subset.variable_metadata
         )
         assert all(
-            dataset_fine_subset.variable_metadata[key]
-            == dataset_coarse_subset.variable_metadata[key]
+            dataset_fine_subset.variable_metadata[key].units
```
AnnaKwa, line 619: For surface pressure, the fine and coarse datasets have different long_name.

fme/downscaling/train.py
```diff
@@ -449,12 +450,16 @@ def build(self) -> Trainer:
...
 
         # load full spatial range of topography to save with model
-        # TODO: this will be replaced in the future with a more general call
-        # to get normalized static inputs from a model config field
+        # TODO: this block will be removed in a following PR that updates the
+        # the training config to have a specified static inputs field with paths.
+        if isinstance(self.train_data.fine[0], MergeNoConcatDatasetConfig):
+            first_fine_config = self.train_data.fine[0].merge[0]
+        else:
+            first_fine_config = self.train_data.fine[0]
```
jpdunc23, line 458: Nit: You could reuse `_first_data_config()` here, though since this is being removed in the future feel free to ignore.

fme/downscaling/data/config.py
```diff
@@ -158,8 +202,13 @@ def mp_context(self):
...
             return None
         for config in self.full_config:
-            if config.engine == "zarr":
+            if isinstance(config, XarrayDataConfig):
+                if config.engine == "zarr":
+                    context = "forkserver"
+                    break
+            elif getattr(config, "zarr_engine_used", False):
                 context = "forkserver"
+                break
```
jpdunc23: `XarrayDataConfig` sets its `zarr_engine_used` attr in its `__post_init__` so I think the following works:

```suggestion
            if getattr(config, "zarr_engine_used", False):
                context = "forkserver"
                break
```

fme/downscaling/data/config.py
```diff
@@ -85,6 +87,46 @@ def expand(self) -> list[XarrayDataConfig]:
...
+        if isinstance(config, XarrayEnsembleDataConfig):
+            expanded.extend(config.expand())
+        else:
+            expanded.append(config)
+    xarray_configs = [c for c in expanded if isinstance(c, XarrayDataConfig)]
+    merge_configs = [c for c in expanded if isinstance(c, MergeNoConcatDatasetConfig)]
+    datasets: list[DatasetABC] = []
+    properties: DatasetProperties | None = None
+    if xarray_configs:
+        ds, prop = get_dataset(
```
jpdunc23: It's a bit awkward that `get_dataset` returns `XarrayConcat` which then later on gets nested in another `XarrayConcat`. Maybe instead of using `get_dataset` and `get_merged_datasets` you can reuse the existing `build` methods on `XarrayDataConfig` and `MergeNoConcatDatasetConfig`? Then you should be able to avoid having to filter `expanded` by config type, since you can just loop over it and call `config.build(names, n_timesteps)` without worrying what type of config it is.

TBH, `get_dataset` and `get_merged_datasets` should probably be made private, though no need to worry about that here.

fme/downscaling/data/config.py
```diff
@@ -373,21 +435,35 @@ def _mp_context(self):
...
         for config in self.fine:
-            if config.engine == "zarr":
-                mp_context = "forkserver"
-        for config in self.coarse_full_config:
-            if config.engine == "zarr":
+            if getattr(config, "engine", None) == "zarr" or getattr(
+                config, "zarr_engine_used", False
+            ):
                 mp_context = "forkserver"
+                break
```
jpdunc23: Here again I think you can check `zarr_engine_used` for both `XarrayDataConfig` and `MergeNoConcatDatasetConfig`. I don't think you need the safety of `getattr` unless you're worried about `self.fine` having types other than what we expect.

fme/downscaling/data/config.py
jpdunc23: `zarr_engine_used` should probably be turned into an `@property` of `XarrayDataConfig` etc. to make it a bit more visible, but that's an existing issue.

fme/downscaling/data/config.py
```diff
@@ -373,21 +435,35 @@ def _mp_context(self):
...
+        if mp_context is None:
+            for coarse_config in self.coarse:
+                if isinstance(coarse_config, XarrayEnsembleDataConfig):
```
jpdunc23: Maybe add an `@property` for `zarr_engine_used` to `XarrayEnsembleDataConfig`? That should simplify the logic here a lot.

fme/downscaling/data/config.py
```diff
@@ -373,21 +435,35 @@ def _mp_context(self):
...
         return mp_context
 
     @property
-    def coarse_full_config(self) -> Sequence[XarrayDataConfig]:
-        # Expands the coarse dataset configs so that any XarrayEnsembleDataConfig
-        # is converted to the equivalent sequence of XarrayDataConfig.
-        coarse_configs = []
+    def coarse_full_config(
+        self,
+    ) -> Sequence[XarrayDataConfig | MergeNoConcatDatasetConfig]:
```
jpdunc23, line 450: This method seems more or less identical to `DataLoaderConfig.full_config`. Maybe create a shared helper?

fme/downscaling/data/config.py
jpdunc23: How important is this check? Seems like units could also easily become mismatched in "harmless" ways, e.g. `m^2` vs `m**2`.

fme/downscaling/data/utils.py
```diff
@@ -5,11 +5,28 @@
...
+) -> XarrayDataConfig | MergeNoConcatDatasetConfig:
+    if isinstance(config, XarrayDataConfig):
+        return dataclasses.replace(config, subset=subset)
+    elif isinstance(config, MergeNoConcatDatasetConfig):
+        merge_configs = [
+            dataclasses.replace(_config, subset=subset) for _config in config.merge
+        ]
+        return MergeNoConcatDatasetConfig(merge=merge_configs)
+    else:
+        raise ValueError(f"Invalid config type: {type(config)}")
```
jpdunc23: `XarrayDataConfig` and `MergeNoConcatDatasetConfig` actually both already have methods `update_subset`, so you could maybe reuse that here and not have to worry about the `isinstance` checks.

fme/downscaling/inference/output.py
jpdunc23: This is maybe getting off-track for this PR, but why does `_single_xarray_config` take a list as input when it only wants a single config? Maybe adding an `InferenceDataLoaderConfig` would help clean this up somewhat.

jpdunc23 (approved):
Mostly looks good to me! I have various suggestions that I think would help simplify the changes a bit, but feel free to treat as optional.

fme/downscaling/inference/output.py
AnnaKwa: My guess is that it is to make the configuration for this entrypoint as similar as possible to the existing `evaluator` and `predict` entrypoints, which can take a list of data configs (@frodre correct me if this is wrong). It's nice to be able to copy/paste the sections from an existing `predict` config to save out the full outputs.

fme/downscaling/data/config.py
AnnaKwa: Thanks, this is much tidier.

fme/downscaling/data/config.py
AnnaKwa: added it as a property since I was adding this property to the ensemble config in this PR as well

PR #847, opened 2026-02-18
add ModelTorchDistributed with tests

mahf708:
add ModelTorchDistributed with tests

Changes:
- fme.core.distributed has a new ModelDistributedBackend that allows for parallelism over spatial dimensions as well as batch/data.
- torch is pinned with a minimum of 2.4.0 to use new facilities for distributed, etc.

- [x] Tests added

Closes #749
Closes #842

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -81,6 +81,8 @@ jobs:
   gpu:
     # These steps run on our self-hosted k8s GPU runners. See below for setup notes
     # https://github.com/ai2cm/long-lived-infrastructure#installing-the-github-actions-runner-controller
+    # Additionally, skip if the target repo isn't ai2cm/ace, lest it tries to run in forks
+    if: ${{ github.repository == 'ai2cm/ace' }}
     runs-on: arc-runner-set
     steps:
       - uses: actions/checkout@v6
@@ -125,7 +127,7 @@ jobs:
             NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=2 make test_parallel
 
       - name: Run GPU benchmarks
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        if: ${{ github.repository == 'ai2cm/ace' && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
         env:
           WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
         run: |
--- a/.github/workflows/beaker.yaml
+++ b/.github/workflows/beaker.yaml
@@ -15,7 +15,8 @@ jobs:
   ace-evaluator:
     runs-on: ubuntu-latest
     # Only deploy on manual trigger or push to main
-    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+    # Additionally, skip if the target repo isn't ai2cm/ace, lest it tries to run in forks
+    if: ${{ github.repository == 'ai2cm/ace' && (github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')) }}
     env:
       CONFIG_PATH: docs/evaluator-config.yaml
       BEAKER_TOKEN: ${{ secrets.BEAKER_TOKEN }}
--- a/scripts/downscaling/plot_beaker_histograms.py
+++ b/scripts/downscaling/plot_beaker_histograms.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python
+"""
+Fetch netCDF event files from a beaker dataset and generate histogram plots
+comparing ensemble predictions against targets for each variable.
+
+This will work for saved event outputs from `fme.downscaling.evaluator`
+from a beaker experiment.  It downloads the experiment files to a temporary
+directory and then parses the filenames for <event_name>_YYYYMMDD.nc to look
+for single-event outputs.
+
+Usage:
+    python plot_beaker_histograms.py <beaker_dataset_id> [--output-dir <path>]
+
+Requires:
+    beaker CLI to be installed and authenticated (https://github.com/allenai/beaker).
+"""
+
+import argparse
+import re
+import subprocess
+import tempfile
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+import xarray as xr
+
+# Matching for <event_name>_YYYYMMDD.nc
+_EVENT_FILE_RE = re.compile(r"(.+)_(\d{8})\.nc$")
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Generate histogram plots from beaker dataset event files"
+    )
+    parser.add_argument(
+        "beaker_dataset_id",
+        help="The beaker dataset ID to fetch",
+    )
+    parser.add_argument(
+        "--output-dir",
+        default="./histogram_outputs",
+        help="Output directory for figures (default: ./histogram_outputs)",
+    )
+    return parser.parse_args()
+
+
+def fetch_beaker_dataset(dataset_id: str, target_dir: str) -> None:
+    """Fetch a beaker dataset to the specified directory."""
+    subprocess.run(
+        ["beaker", "dataset", "fetch", dataset_id, "--output", target_dir],
+        check=True,
+    )
+
+
+def find_event_files(directory: str) -> dict[str, Path]:
+    """Find netCDF files matching the event naming pattern, keyed by event name."""
+    event_files = {}
+    for p in sorted(Path(directory).glob("*.nc")):
+        # extract event name
+        matched = _EVENT_FILE_RE.match(p.name)
+        if matched:
+            event_files[matched.group(1)] = p
+    return event_files
+
+
+def detect_variable_pairs(ds: xr.Dataset) -> list[str]:
+    """Detect variables that have both _predicted and _target versions."""
+    predicted = {
+        v[: -len("_predicted")] for v in ds.data_vars if v.endswith("_predicted")
+    }
+    target = {v[: -len("_target")] for v in ds.data_vars if v.endswith("_target")}
+    return sorted(predicted & target)
+
+
+def plot_histogram_lines(
+    ds: xr.Dataset,
+    key_prefix: str,
+    title_prefix: str,
+    save_path: Path,
+) -> None:
+    """
+    Plot histogram comparing ensemble predictions against target.
+    """
+    fig, ax = plt.subplots(figsize=(10, 6))
+
+    # Gather all predicted and target data
+    target_data = ds[f"{key_prefix}_target"]
+    predicted_data = ds[f"{key_prefix}_predicted"]
+    all_data = np.concatenate([target_data.values[None], predicted_data.values])
+
+    # Compute bin edges from min/max of all data
+    bins = 50
+    data_min, data_max = np.min(all_data), np.max(all_data)
+    bin_edges = np.linspace(data_min, data_max, bins + 1)
+    bin_centers = 0.5 * (bin_edges[1:] + bin_edges[:-1])
+
+    sample_data = predicted_data.values
+    if sample_data.ndim != 3:
+        raise ValueError(
+            f"Expected predicted data to be 3D (samples, lat, lon), "
+            f"got shape {sample_data.shape}"
+        )
+
+    # Calculate the tail percentile values for each generated sample
+    # and generate a 95% confidence interval
+    lower_bounds = np.percentile(sample_data, 0.01, axis=(1, 2))
+    upper_bounds = np.percentile(sample_data, 99.99, axis=(1, 2))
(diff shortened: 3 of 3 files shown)


fme/core/distributed/distributed.py
```diff
@@ -26,22 +28,62 @@ class Distributed:
...
+        The backend can be overridden at the session level via environment
+        variables, which is useful for switching between backends
+
+        - ``FME_DISTRIBUTED_BACKEND=torch`` – force :class:`TorchDistributed`
+          (the default when ``torch.distributed`` is available)
+        - ``FME_DISTRIBUTED_BACKEND=spatial`` – force
+          :class:`ModelTorchDistributed`. Requires ``FME_SPATIAL_H`` and
+          ``FME_SPATIAL_W`` to also be set.
+        - ``FME_DISTRIBUTED_BACKEND=non_distributed`` – force
+          :class:`NonDistributed`
```
mahf708: i'm not thrilled with these names or this general method of forcing; I if we have to go this route, I prefer to call the new one: "model" (since it can parallelize over all sorts of dims/tags)

mcgibbon: For now, can you keep the behavior Oscar was going to use? That is, if (let's call it) `FME_DISTRIBUTED_H` or `_W` is set > 1 then the spatial backend is used? If you need a way to force it to get used, perhaps if one or both is set we can use the spatial backend and if not use the torch backend, and only use the torch backend if both are unset? I don't think we currently need a way to force non-distributed from the CLI so we shouldn't add that feature.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+                process_group=self._data_group,
+            )
+        return DummyWrapper(module)
+
+    def barrier(self):
+        """Global barrier across all ranks."""
+        logger.debug("Barrier on rank %d", self._rank)
+        torch.distributed.barrier(device_ids=self._device_ids)
+
+    def shutdown(self):
```
mcgibbon, line 269: Not for this PR or necessarily related to spatial parallelism, but we should think about defining a context manager for parallelism that makes sure cleanup happens when the context exits, kind of like we do with GlobalTimer. I think we currently don't call it properly in the unit tests for example.

fme/core/distributed/parallel_tests/test_local_slices.py
```diff
@@ -13,7 +13,8 @@ def test_gather_tensor_from_local_slices():
     "batch" parallelism in this test.
     """
     dist = Distributed.get_instance()
-    global_shape = (2, 4, 4)
+    n_dp = dist.total_data_parallel_ranks
```
mcgibbon, line 16: Is this desired? I feel like we should set a constant shape like 4 that covers the cases we plan to run with, so that special values like 3 data parallel ranks are more interesting (which my test may need to be refactored to properly manage, idk).

fme/core/distributed/parallel_tests/test_backend.py
```diff
@@ -0,0 +1,196 @@
...
+def test_reduce_max_constant():
+    """Max of a constant tensor is the same constant."""
+    dist = Distributed.get_instance()
+    t = torch.full((4,), 3.0, device=get_device())
+    result = dist.reduce_max(t)
+    torch.testing.assert_close(result, t.clone())
+
+
+def test_reduce_mean_rank_offset():
+    """Each dp-rank adds its rank index.  Mean should equal base + avg(ranks)."""
```
mcgibbon, line 52: Does this (or another test) capture the failure case where `reduce_mean` improperly reduces across all ranks and not just the data parallel ranks?

fme/core/distributed/parallel_tests/test_backend.py
mcgibbon, line 1: Good tests!

fme/core/distributed/external/pnd_manager.py
mcgibbon, line 1: It's fine this time, but generally speaking when you PR a copy of a file that has changes, first commit the copied file without changes (with the pre-commit ignore flag), and then commit the changes that make the file work. That way I can tell what was changed. For this review I'll just treat this like new code, but with a lower standard for style since it's an external copy.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+
+from .base import DistributedBackend
+from .external.pnd_manager import DistributedManager
+from .non_distributed import DummyWrapper
+from .torch_distributed import _gather_irregular
+
+logger = logging.getLogger(__name__)
+
+
+def _build_mesh(
```
mcgibbon: Question (nit/optional): There's also mesh initialization logic in `pnd_manager.py`, is there a reason to have it split between these two files at different levels of specialization?

fme/core/distributed/model_torch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+    def total_ranks(self) -> int:
+        """Total number of processes (world size)."""
+        return self._world_size
+
+    @property
+    def total_data_parallel_ranks(self) -> int:
+        """Number of data-parallel ranks."""
+        return self._data_size
+
+    def get_local_slices(self, tensor_shape, rank: int, data_parallel_dim: int | None):
```
mcgibbon: Note I have a PR about to merge to main that removes the `rank` arg.

fme/core/distributed/external/pnd_manager.py
mahf708: This is a good idea, I'd like to rebase and do exactly as you're describing. I am significantly thinning this file, and it is not transparent; it is better to thin it in a transparent way

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
     "tensorly",
     "tensorly-torch",
     "torch-harmonics==0.8.0",
-    "torch>=2",
+    "torch>=2.4.0",
```
elynnwu, line 42: No action needed here, I think our dependency only image already uses 2.7.1 and I don't believe there's anything in this PR that warrants re-rebuilding the image again.

fme/core/distributed/parallel_tests/test_backend.py
elynnwu: +1, I think it'll be good to have a test that fails if it reduces across all ranks

fme/core/distributed/external/pnd_manager.py
mahf708: done in 4b11ab1b5821772f1493ea194f7cda027e7dd2a9 and 9d226c21ddc63fc9f38ae342b4a77960e014dc9f

fme/core/distributed/model_torch_distributed.py
mahf708: noted

pyproject.toml
mahf708: thanks!

fme/core/distributed/model_torch_distributed.py
mahf708: done: [3f0bbf5](https://github.com/ai2cm/ace/pull/847/commits/3f0bbf52d2fca1f66c2111f695197213942a2454)

fme/core/distributed/parallel_tests/test_backend.py
mahf708: done: [3f0bbf5](https://github.com/ai2cm/ace/pull/847/commits/3f0bbf52d2fca1f66c2111f695197213942a2454)

fme/core/distributed/parallel_tests/test_local_slices.py
mahf708: For now, I think it is an easy copout to get things going and running with arbitrary tests. If we hardcode the batch dimension, we will need to instrument pytest skipping and such. I like the idea that these tests can run successfully with 5,000,000 ranks ;)

fme/core/distributed/model_torch_distributed.py
mahf708: FWIW, the torch backend is the one with problematic teardown for some reason; the new one is slightly cleaner. I can look into why...

fme/core/distributed/distributed.py
mahf708: Addressed mostly, but I'd rather force people to pick for one, we can streamline defaults later, etc.

.github/workflows/tests.yaml
```diff
@@ -92,7 +114,28 @@ jobs:
...
+            # | ------- | -- | - | - |
+            # | torch   |  2 | 1 | 1 |
+            # | torch   |  3 | 1 | 1 |
+            # | model   |  3 | 1 | 1 |
+            # | model   |  2 | 2 | 1 |
+            # | model   |  2 | 1 | 2 |
+            # | model   |  3 | 3 | 1 |
+            # | model   |  3 | 1 | 3 |
+            # | model   | 4 | 2 | 2 |
+            # | model   | 4 | 2 | 1 |
```
mahf708: a push is stuck to fix these two :( forgot we can't run 4 ranks on gpus

fme/core/distributed/distributed.py
elynnwu: Per our slack conversation, Jeremy had suggested a minimal flag like `FME_FORCE_NON_DISTRIBUTED`, but I think `FME_DISTRIBUTED_BACKEND=none` is also very self-explanatory. So all good here

.github/workflows/tests.yaml
```diff
@@ -34,7 +34,29 @@ jobs:
...
+          NPROC=2  FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=2 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=3 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=3 make test_parallel
+          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=2 make test_parallel
+          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
+          NPROC=18 FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=3 FME_DISTRIBUTED_W=3 make test_parallel
```
mcgibbon, line 59: I'd like to condense these into a single make target. Will do in a later PR.

fme/core/distributed/distributed.py
```diff
@@ -29,22 +31,62 @@ class Distributed:
...
+            if backend_env == "model":
+                h = int(os.environ["FME_DISTRIBUTED_H"])
+                w = int(os.environ["FME_DISTRIBUTED_W"])
+                singleton = cls(spatial_parallelism=(h, w))
+            elif backend_env == "none":
+                singleton = cls(force_non_distributed=True)
+            elif backend_env == "torch":
+                singleton = cls()
+            elif backend_env is not None:
+                raise ValueError(
```
mcgibbon, line 84: Is this the expected behavior for an explicitly unset environment variable`FME_DISTRIBUTED_BACKEND=`?

```
$ UNSET="" python3
Python 3.11.14 | packaged by conda-forge | (main, Oct 22 2025, 22:46:25) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> e = os.environ.get("UNSET")
>>> e
''
>>> type(e)
<class 'str'>
```

mahf708: That's a good catch, but it should default to the other branch, right?

mahf708: all the way down to:
```py
            else:
                singleton = cls()
```

mahf708: Oh, no

mahf708: Wait, this was intentional. Thinking: If someone plays with this env variable, they will need to choose between the three provided options. Otherwise, they should not touch it.

mahf708: Meta point: don't let users play around!

mcgibbon: I feel like I would expect the default behavior, and it's ideal to do what's expected. But we're unlikely to run into this scenario.

In certain script contexts it's easier to set an env var to '' than it is to unset it, but I also don't know if that would come up with this.

PR #848, opened 2026-02-18
Make benchmark CI env more similar to docker image

oliverwm1:
We want benchmarks to run in an environment as similar as possible to our production env using the deps-only image. This makes one change particularly relevant for GPU benchmarks. Alternatively we could run these benchmarks via a beaker job, but then it is harder to guarantee that they will run consistently on the same GPU type, which makes interpretation harder.


oliverwm1:
> Could we do beaker jobs on only one cluster?

We could, just can't guarantee it will run before timing out given how busy most of the clusters are.

PR #856, opened 2026-02-19
Coupled evaluator with prediction loader 

### LINKED ISSUE #805: `prediction_loader` for coupled eval
We have this feature for `fme.ace.inference` but not coupled. This is useful to compare things like `inference/mean/weighted_mean_gen`, where the current workaround is to manually add in the corresponding `target` from a different panel, but this is cumbersome and requires manually doing this for all variables. If we can have `prediction_loader` for coupled eval, we can simply do a run that's target data only and use it as a standalone wandb run to compare with.

Image

elynnwu:
`prediction_loader` is an already supported feature in ace evaluator, we add this feature for coupled evaluator so that we can easily compare with target data on wandb.

Resolves #805


fme/coupled/inference/evaluator.py
```diff
@@ -321,13 +365,40 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
...
+            aggregator=aggregator,
+            prediction_data=prediction_data,
+            target_data=data,
+            deriver=deriver,
+            writer=writer,
+            record_logs=record_logs,
+            restrict_to_output_names=(
+                stepper.ocean.out_names,
+                stepper.atmosphere.out_names,
+            ),
```
elynnwu: We did not have to do this in ace evaluator because there was no merge operation, we need this here because we can have overlap in atmosphere and ocean in_names (e.g., land_fraction)

elynnwu:
Job [here](https://beaker.org/orgs/ai2/workspaces/ace/work/01KHVZYNT60QFC0BD0BMDWYREZ?taskId=01KHVZYP1VX477GSH2PW8AT4CA&jobId=01KHVZYP5TK846ZHKT9F77P0HN)
wandb [here](https://wandb.ai/ai2cm/ace/runs/1d2dsx5j)

fme/coupled/inference/loop.py
```diff
@@ -0,0 +1,110 @@
...
+    aggregator: InferenceAggregatorABC[
+        CoupledPairedData | CoupledPrognosticState,
+        CoupledPairedData,
+    ],
+    prediction_data: InferenceGriddedData,
+    target_data: InferenceGriddedData,
+    deriver: CoupledDeriverABC,
+    writer: CoupledPairedDataWriter | NullDataWriter | None = None,
+    record_logs: Callable[[InferenceLogs], None] | None = None,
+    restrict_to_output_names: tuple[list[str], list[str]] | None = None,
```
jpdunc23: Can you please add a new dataclass to replace `tuple[list[str], list[str]]` in fme/coupled/typing_.py:

```python
@dataclasses.dataclass
class CoupledNames:
    ocean: list[str]
    atmosphere: list[str]
```

~~You could also add an `@property` called `out_names` to `CoupledStepperConfig` which gets the names from the component steppers and returns a `CoupledNames` object.~~ See my [other comment](https://github.com/ai2cm/ace/pull/856/changes#r2834978875) where I suggest instead adding `all_names` to `CoupledStepperConfig`.

fme/coupled/inference/test_evaluator.py
```diff
@@ -200,6 +207,11 @@ def inference_helper(
         "ak_0" not in ds_atmos.coords
     ), "TODO: update this assertion now that vertical coords are written"
     assert "ak_0" not in ds_ocean.coords
+    if use_prediction_data:
+        assert not os.path.exists(tmp_path / "atmosphere/restart.nc")
+        assert not os.path.exists(tmp_path / "atmosphere/initial_condition.nc")
+        assert not os.path.exists(tmp_path / "ocean/restart.nc")
+        assert not os.path.exists(tmp_path / "ocean/initial_condition.nc")
```
jpdunc23, line 214: Would be good to add a check similar to the following in the fme.ace test:

```python
                # if these are off by something like 90% then probably the stepper
                # is being used instead of the prediction_data
                assert log[f"inference/mean/weighted_rmse/{var}"] == 0.0
                assert log[f"inference/mean/weighted_bias/{var}"] == 0.0
```

fme/coupled/inference/loop.py
```diff
@@ -0,0 +1,110 @@
...
+                    initial_condition=CoupledPairedData(
+                        ocean_data=PairedData.from_batch_data(
+                            prediction=pred_ic.ocean_data,
+                            reference=target_ic.ocean_data,
+                        ),
+                        atmosphere_data=PairedData.from_batch_data(
+                            prediction=pred_ic.atmosphere_data,
+                            reference=target_ic.atmosphere_data,
+                        ),
+                    ),
```
jpdunc23: Can you use `CoupledPairedData.from_coupled_batch_data` here?

fme/coupled/inference/loop.py
```diff
@@ -0,0 +1,110 @@
...
+from fme.coupled.data_loading.batch_data import (
+    CoupledBatchData,
+    CoupledPairedData,
+    CoupledPrognosticState,
+)
+from fme.coupled.data_loading.gridded_data import InferenceGriddedData
+from fme.coupled.inference.data_writer import CoupledPairedDataWriter
+
+
+class CoupledDeriverABC(Protocol):
```
jpdunc23: I think using `Protocol` rather than `abc.ABC` as in fme.ace is good, but we should call it `CoupledDeriverProtocol` or simply `CoupledDeriver`.

fme/coupled/inference/evaluator.py
jpdunc23: In the fme.ace version we log all variables including input-only forcings. I think this can be useful for examining the forcings and for post-hoc analyses.

See my [other comment](https://github.com/ai2cm/ace/pull/856/changes#r2834868549). I think we can instead add an  `@property` called `all_names` to `CoupledStepperConfig` which returns a `CoupledNames` where:
- `atmosphere` has its out names + `atmosphere_forcing_exogenous_names` + `shared_forcing_exogenous_names`
- `ocean` has its out names + `ocean_forcing_exogenous_names`

elynnwu: Good point, added

fme/coupled/inference/loop.py
elynnwu: Yes, updated

fme/coupled/inference/test_evaluator.py
elynnwu: Added

fme/coupled/inference/loop.py
elynnwu: Changed to `CoupledDeriver`

fme/coupled/data_loading/test_data_loader.py
```diff
@@ -65,6 +65,7 @@ def _save_netcdf(
     timestep_size=1,
     timestep_start=0,
     nz=3,
+    masked_fill_value: float = float("nan"),
```
jpdunc23, line 68: Do you know why this is needed for `prediction_loader` inference but not evaluator inference?

fme/coupled/stepper.py
```diff
@@ -383,6 +383,19 @@ def shared_forcing_exogenous_names(self) -> list[str]:
...
         return self._shared_forcing_exogenous_names
 
+    @property
+    def all_names(self) -> CoupledNames:
+        """All variable names to log (outputs plus input-only forcings)."""
+        atmosphere_names = (
+            self.atmosphere.stepper.output_names
+            + self._atmosphere_forcing_exogenous_names
+            + self._shared_forcing_exogenous_names
+        )
```
jpdunc23, line 392: Shouldn't matter in practice, but agent review flagged that `_shared_forcing_exogenous_names` is a subset of `_atmosphere_forcing_exogenous_names`, so probably best to remove `_shared_forcing_exogenous_names` here.

fme/coupled/inference/loop.py
```diff
@@ -0,0 +1,105 @@
...
+    aggregator: InferenceAggregatorABC[
+        CoupledPairedData | CoupledPrognosticState,
+        CoupledPairedData,
+    ],
+    prediction_data: InferenceGriddedData,
+    target_data: InferenceGriddedData,
+    deriver: CoupledDeriver,
+    writer: CoupledPairedDataWriter | NullDataWriter | None = None,
+    record_logs: Callable[[InferenceLogs], None] | None = None,
+    restrict_to_output_names: CoupledNames | None = None,
```
jpdunc23: Nit: Maybe call this `all_names` since it is no longer restricted to just output names.

fme/coupled/data_loading/test_data_loader.py
elynnwu: Okay this is because we didn't write `mask_2d` so variables like `o_prog` doesn't have the right mask, I reverted back to using nans as masked fill values and added the proper mask.

PR #857, opened 2026-02-20
Add parallel equivalent of test_step

mcgibbon:
This PR adds unit and regression tests for some StepABC implementations, specifically the configurations using NoiseConditionedSFNO.

- [x] Tests added

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,3 +46,18 @@ committing code: `pre-commit install`.
 
 You can run local tests with `pytest`. As a shortcut for running only faster unit
 tests, use `make test_very_fast` from the root of the repository.
+
+## Internal Development
+
+If you are making changes directly to ai2cm/ace, please follow these internal
+development guidelines.
+
+When making new branches, use the naming convention:
+`<type>/<short-description>`, where `<type>` is one of:
+- feature: Any new functionality in the repository including workflows and scripts but not including configurations. Should be the "default" type if it's unclear which to use.
+- refactor: No changes to features, just code restructuring or simplification.
+- fix: Bug fixes.
+- exp: Branch used for running experiments that is likely not intended to merge with mainline code. Functionality changes in these branches should first be PR'd using a feature branch.
+- config: Changes to baseline and experimental configurations under config/.
+- scripts: Changes isolated to a single script under scripts/ and subject to less rigorous review than changes to core code.
+- docs: Documentation changes only.
--- a/fme/ace/data_loading/benchmark.py
+++ b/fme/ace/data_loading/benchmark.py
@@ -43,22 +43,16 @@ class BenchmarkConfig:
             requirements=DataRequirements(self.names, self.n_timesteps),
         )
 
-    def configure_wandb(self, env_vars: dict | None = None, **kwargs):
-        config = to_flat_dict(dataclasses.asdict(self))
-        # our wandb class requires "experiment_dir" to be in config
-        config["experiment_dir"] = TMPDIR
-        os.makedirs(TMPDIR)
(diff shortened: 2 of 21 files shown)


fme/core/distributed/parallel_tests/test_step.py
```diff
@@ -0,0 +1,494 @@
...
+    img_shape = (20, 40)
+    n_samples = 2
+    selector = get_config(None)
+    if selector.config.get("conditional", False):
+        labels = BatchLabels.new_from_set(
+            {"a", "b"}, n_samples=n_samples, device=fme.get_device()
+        )
+        labels.tensor[:] = np.random.randint(
+            0, 2, (n_samples,), device=fme.get_device()
+        )
```
elynnwu: Don't think numpy takes in device

fme/core/distributed/parallel_tests/test_step.py
```diff
@@ -0,0 +1,494 @@
...
+                    secondary_diagnostic_names=["diagnostic_rad"],
+                    network=ModuleSelector(type="MLP", config={}),
+                ),
+                normalization=normalization,
+            ),
+        ),
+    )
+
+
+def get_label_conditioned_selector(
```
elynnwu: I know this is just copied, but this is unused, maybe remove?

elynnwu (commented):
It's unclear to me how we want to maintain these duplicated tests in the future, I think it's fine that we keep the parallel tests in a separate folder but I do worry that there're many duplicated calls. Is it possible to share some of the utils at least? for example the `get_*` functions

fme/core/distributed/parallel_tests/test_step.py
mcgibbon: Removed

fme/core/distributed/parallel_tests/test_step.py
mcgibbon: Updated

mcgibbon:
> It's unclear to me how we want to maintain these duplicated tests in the future, I think it's fine that we keep the parallel tests in a separate folder but I do worry that there're many duplicated calls. Is it possible to share some of the utils at least? for example the `get_*` functions

It's difficult to import symbols form test files in a different directory, I would need to put them into the ace package itself. And I kind of want the ability to define these configurations independently, for example the non-parallel corrector selector uses the old SFNO architecture while this one uses the NCSFNO since it will support parallelism.

Maybe eventually we can ensure everything supports parallelism and won't need duplicates for these tests?

elynnwu:
> Maybe eventually we can ensure everything supports parallelism and won't need duplicates for these tests?

Okay this would be great. We will probably have to think about running in parallel for old architectures, maybe they can just be skipped in some configurations?

mahf708:
> > Maybe eventually we can ensure everything supports parallelism and won't need duplicates for these tests?
> 
> Okay this would be great. We will probably have to think about running in parallel for old architectures, maybe they can just be skipped in some configurations?

This is something we are discussing in E3SM. Our (ultimate) vision in implementing the spatial ||ism stuff is that the distributed backends have this strict hierarchy: ModelTorchDistributed is a superset of TorchDistributed is a superset of NonDistributed. In other words, for a given problem size (if it fits on a machine in all configs) they all should yield identical up to reduction precision (cuda reductions can be flaky, which we can allow, but which we can try to disallow if needed with a perf hit). The hierarchy means that all ops downstream must be backend-agnostic, e.g., something like RealFFT2 shouldn't care about the backend (neither in API nor in numerical results). An aside, for the e3sm model itself, we ensure this in E3SM with strict testing, so when the C++/F90/etc. cases run, they are layout-invariant (i.e., get identical results with any node count)

PR #861, opened 2026-02-20
Improve performance of dhconv contraction

mcgibbon:
This PR changes the shape of the weight matrix in SpectralConvS2 from [group, in, out, lat] to [group, lat, out, in] so that the matrix multiplication dimensions [out, in] are the fast dimensions, with [in] being the fastest. Backwards compatibility is included for existing checkpoints.

Changes:
- No public API changes

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/s2convolutions.py
+++ b/fme/core/models/conditional_sfno/s2convolutions.py
@@ -59,8 +59,6 @@ def _contract_lora(
     """
     lora_A = torch.view_as_complex(lora_A)
     lora_B = torch.view_as_complex(lora_B)
-    # tmp = torch.einsum("gxri,bgixy->bgxry", lora_A, x)
-    # out = torch.einsum("gxor,bgxry->bgoxy", lora_B, tmp)
     return torch.einsum("gxri,gxor,bgixy->bgoxy", lora_A, lora_B, x)
 
 


mcgibbon:
Commits are done such that there's verification the model produces identical behavior. I made a separate commit updating the regression target which only trivially changes the weight initialization (making it cleaner, but breaking regression).

Still need to write a test that ensures previous checkpoints have their behavior unchanged, but it should be implemented/working using the same method we use to support old checkpoints without a group dim.

Before and after (note the change to dhconv and the total):
image
image
Note that with 8 groups, we no longer see an improvement as the runtime is already low (15ms on T4).

mcgibbon:
~~When reviewing, please check that the regression file updates are valid by looking per-commit. The weight shape refactor is done first in a way that maintains backwards compatibility and passes regression tests, with the regression targets only updated afterwards with a trivial change to the way the weights are initialized (which changes the values they get initialized to).~~

I added a checkpoint-load regression test in another PR, so this is no longer an issue.

mcgibbon:
I added a regression test for the output from a loaded checkpoint being the same, using an input checkpoint generated from main. The test caught a bug, so it's working properly.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -59,7 +59,9 @@ def _contract_lora(
     """
     lora_A = torch.view_as_complex(lora_A)
     lora_B = torch.view_as_complex(lora_B)
-    return torch.einsum("girx,grox,bgixy->bgoxy", lora_A, lora_B, x)
+    # tmp = torch.einsum("gxri,bgixy->bgxry", lora_A, x)
```
Arcomano1234: I believe this is left over commented out code from testing

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -179,38 +181,40 @@ def __init__(
...
             self.modes_lat_local,
+            out_channels // num_groups,
+            in_channels // num_groups,
         ]
 
         assert factorization == "ComplexDense"
         self.weight = nn.Parameter(scale * torch.randn(*weight_shape, 2))
-        self.weight.is_shared_mp = ["matmul", "w"]
 
+        self.lora_rank = lora_rank
```
Arcomano1234, line 198: `self.lora_rank `looks unused

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -228,6 +232,68 @@ def __init__(
 
         # rewrite old checkpoints on load
         self.register_load_state_dict_pre_hook(self._add_singleton_group_dim)
+        self.register_load_state_dict_pre_hook(self._reorder_weight_dims)
+
+    @staticmethod
+    def _reorder_weight_dims(
```
Arcomano1234, line 236: Optional: Its a little hard for me to say for sure this is working 100% correctly. Did we want to add a test with the old ckpt order to make sure this working? If you have already tested this then its fine but I just want to make sure there isn't any subtle bugs like weight naming bugs etc.

mcgibbon: I added (possibly in an earlier pr) a regression test that asserts if you load a presaved checkpoint and call it, the result should be the same as the presaved target. It’s maybe worth convincing yourself of this for the review.

Arcomano1234: I see, yes this was in an early PR. Seems like that test passes so `_reorder_weight_dims` must be doing its job correctly.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: It's used as `module.lora_rank` in `_reorder_weight_dims`

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Sort of, it's an important performance optimization I'll be merging in the next PR. Still, I suppose I can remove it.

Arcomano1234: I see in that case it doesn't matter I just didnt know if this was intentional

fme/core/models/conditional_sfno/s2convolutions.py
Arcomano1234: Ahh I forgot about that

PR #862, opened 2026-02-20
Deprecate training-only attributes on `StepperConfig`

jpdunc23:
Adds `stepper_training: TrainStepperConfig` to the `fme.ace` training config and removes  `loss`, `optimize_last_step_only`, `n_ensemble`, `parameter_init`, and `train_n_forward_steps` from `StepperConfig`.

**WARNING**: This is a breaking change for existing `fme.ace` and `fme.coupled` training configs.

Changes:
- Updates baseline and example ACE configs.
- Renames `train_stepper: CoupledTrainStepperConfig` to `stepper_training: CoupledTrainStepperConfig` on the `fme.coupled` training config.

- [x] Tests added


fme/ace/test_train.py
```diff
@@ -369,6 +366,11 @@ def _get_test_yaml_files(
                 ),
             ),
         ),
+        train_stepper=TrainStepperConfig(
```
oliverwm1: I'm a bit late on this, but what about something like `StepperTrainerConfig` for the name? "TrainStepper" is a bit confusing to me, since this isn't really the configuration for a stepper.

oliverwm1: Or `StepperTrainConfig`?

mcgibbon: An issue with a different name is, this is explicitly the implementation of the TrainStepperABC generic class, so currently it is a TrainStepper whether that's a confusing name or not.

Another option is to put Adapter in the name, since this is an adapter over the Stepper to make it supply the APIs needed for a TrainStepper. For example, `TrainStepperAdapter` and `TrainStepperAdapterConfig`.

Yet another option is to rename Stepper somehow, but idk, TrainStepper really is a Train-api version of the Stepper, and contains a Stepper, so it seems like an appropriate name.

mcgibbon (commented):
IMO this is ready to merge, holding off on review in case we need to discuss names

fme/ace/test_train.py
mcgibbon: Is the issue with the configuration key `train_stepper`, with the class name, or both @oliverwm1 ? If the configuration key name is good I would suggest merging and then gathering feedback for a potential rename PR, since the class name is internal and easy to change backwards-compatibly.

jpdunc23: I'm hesitant to use "trainer" here since we already use that term when referring to an important class `Trainer`. I'd suggest `stepper_training` as an alternative name for the `TrainConfig` attribute.

From the perspective of `Trainer`, it's `stepper: TrainStepperABC` is the same object it always was, so I feel that `TrainStepper` is an appropriate name. I think it's a nice convention to append `Config` to the name of the thing being built by the config as in `TrainStepperConfig` so I would also lean toward keeping the config class name as is.

PR #863, opened 2026-02-21
Make `DatasetInfo` checks on `MaskProvider` more flexible

jpdunc23:
We occasionally add new masks to existing datasets which can lead to an `IncompatibleDatasetInfo` when running inference with checkpoints created before new masks were added. This PR allows `DatasetInfo`'s  checks of `MaskProvider` to pass when the checkpoint masks are a subset of the loaded dataset's masks.

Changes:
- Adds the `assert_compatible_with` method to `MaskProvider`.

- [x] Tests added

### DIFF
--- a/fme/core/dataset_info.py
+++ b/fme/core/dataset_info.py
@@ -120,11 +120,10 @@ class DatasetInfo:
                     f"{self._vertical_coordinate} != {other._vertical_coordinate}"
                 )
         if self._mask_provider is not None and other._mask_provider is not None:
-            if not self._mask_provider.is_compatible_with(other._mask_provider):
-                issues.append(
-                    f"mask_provider is not compatible, "
-                    f"{self._mask_provider} not a subset of {other._mask_provider}"
-                )
+            try:
+                self._mask_provider.assert_compatible_with(other._mask_provider)
+            except AssertionError as e:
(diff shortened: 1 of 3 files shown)


fme/core/dataset_info.py
```diff
@@ -120,10 +120,10 @@ def assert_compatible_with(self, other: "DatasetInfo"):
                     f"{self._vertical_coordinate} != {other._vertical_coordinate}"
                 )
         if self._mask_provider is not None and other._mask_provider is not None:
-            if self._mask_provider != other._mask_provider:
+            if not self._mask_provider.is_compatible_with(other._mask_provider):
                 issues.append(
                     f"mask_provider is not compatible, "
-                    f"{self._mask_provider} != {other._mask_provider}"
+                    f"{self._mask_provider} not a subset of {other._mask_provider}"
```
AnnaKwa: This error message covers one of the cases where `is_compatible_with` returns False but it could be due to the `isinstance(other, MaskProvider)` check or the mask values mismatch check. If it's important to you to have the specific issue reported, you could have that method return a different message for each possible failure.

jpdunc23: Good point, updated the `MaskProvider` method to `assert_compatible_with` which throws a more informative error which `DatasetInfo.assert_compatible_with` and catch and report.

PR #865, opened 2026-02-23
Remove hard-coded resumable arg from downscaling configure_logging

mcgibbon:
In the downscaling inference and evaluator scripts, we hard-code a `resumable=True` argument when calling `configure_logging`, which is the only place `configure_logging` is used. Since we always want downscaling inference/evaluation to be resumable, we should hard-code it in the configuration object itself. Additionally, the default value is False despite always being called with True, leading to confusion.

This PR removes the `resumable: bool = False` argument and instead moves the hard-code inside the `configure_logging` helper. This matches the behavior of similar `ace` objects. 

Changes:
- EvaluatorConfig.configure_logging and InferenceConfig.configure_logging in downscaling no longer take a `resumable: bool = False` arg and instead hard-code it internally to True, which matches how it is currently called.

- [ ] Tests added


mcgibbon:
This change was discussed in https://github.com/ai2cm/ace/pull/777 with mixed conclusions, so I moved it out into its own PR. Let me know if this makes it any cleaner from your perspectives @AnnaKwa @frodre , otherwise we don't need to merge it.

PR #868, opened 2026-02-23
Add `optimize_last_step_only` to coupled `LossContributionsConfig`

jpdunc23:
Adds `optimize_last_step_only` boolean to `fme.coupled`'s `LossContributionsConfig` to allow for optimization of the final output step's loss while ignoring earlier steps. This can be configured independently for the atmosphere and ocean components of the coupled stepper.

NOTE: This PR deprecates configuration of `n_coupled_steps` at the top level of `TrainConfig` in `fme.coupled`, in favor of configuration as part of `TrainConfig.stepper_training: CoupledTrainStepperConfig`.

Changes:
- Removed `TrainConfig.n_coupled_steps` and added `CoupledTrainStepperConfig.n_coupled_steps`. Baseline coupled training configs updated.
- Refactored `CoupledStepper.get_prediction_generator()` to remove the `step_is_optimized` argument to reduce coupling between the `CoupledStepper` and `CoupledTrainStepperConfig`, since `step_is_optimized` is a `CoupledStepperTrainLoss` concern.

- [x] Tests added

### DIFF
--- a/configs/baselines/cm4-piControl/finetune-config-template.yaml
+++ b/configs/baselines/cm4-piControl/finetune-config-template.yaml
@@ -4,7 +4,6 @@ validate_using_ema: true
 ema:
   decay: 0.999
 max_epochs: 20
-n_coupled_steps: 4
 inference:
   n_coupled_steps: 1456
   coupled_steps_in_memory: 8
@@ -122,6 +121,7 @@ optimization:
   scheduler:
     type: CosineAnnealingLR
 stepper_training:
+  n_coupled_steps: 4
   parameter_init:
     checkpoint_path: /ckpt.tar
   ocean:
--- a/configs/baselines/cm4-piControl/finetune-config.yaml
+++ b/configs/baselines/cm4-piControl/finetune-config.yaml
@@ -4,7 +4,6 @@ validate_using_ema: true
 ema:
   decay: 0.999
 max_epochs: 20
-n_coupled_steps: 4
 inference:
   n_coupled_steps: 1456
   coupled_steps_in_memory: 8
@@ -122,6 +121,7 @@ optimization:
   scheduler:
     type: CosineAnnealingLR
 stepper_training:
+  n_coupled_steps: 4
   parameter_init:
     checkpoint_path: /ckpt.tar
   ocean:
--- a/configs/baselines/cm4-piControl/train-config-template.yaml
+++ b/configs/baselines/cm4-piControl/train-config-template.yaml
@@ -4,7 +4,6 @@ validate_using_ema: true
 ema:
   decay: 0.999
 max_epochs: 20
-n_coupled_steps: 4
 inference:
   n_coupled_steps: 1456
   coupled_steps_in_memory: 8
@@ -120,6 +119,7 @@ optimization:
     weight_decay: 0.01
   use_gradient_accumulation: true
 stepper_training:
+  n_coupled_steps: 4
   ocean:
     parameter_init:
       weights_path: /ocean_ckpt.tar
--- a/configs/baselines/cm4-piControl/train-config.yaml
+++ b/configs/baselines/cm4-piControl/train-config.yaml
@@ -4,7 +4,6 @@ validate_using_ema: true
 ema:
   decay: 0.999
 max_epochs: 20
-n_coupled_steps: 4
 inference:
   n_coupled_steps: 1456
   coupled_steps_in_memory: 8
@@ -120,6 +119,7 @@ optimization:
     weight_decay: 0.01
   use_gradient_accumulation: true
 stepper_training:
+  n_coupled_steps: 4
   ocean:
     parameter_init:
       weights_path: /ocean_ckpt.tar


fme/coupled/stepper.py
```diff
@@ -1630,36 +1658,47 @@ def _accumulate_loss(
...
+        n_outer_steps = data.ocean_data.n_timesteps - self.n_ic_timesteps
+        n_total_atmos_steps = n_outer_steps * self.n_inner_steps
+        for i_outer in range(n_outer_steps):
+            for i_inner in range(self.n_inner_steps):
+                global_atmos_step = i_outer * self.n_inner_steps + i_inner
+                optimize = self._loss.step_is_optimized(
+                    "atmosphere",
+                    global_atmos_step,
+                    n_total_atmos_steps,
+                )
```
jpdunc23, line 1674: This was previously handled in `CoupledStepper.get_prediction_generator()` but refactored to be handled here in `CoupledTrainStepper._accumulate_loss()`.

fme/coupled/loss.py
```diff
@@ -29,9 +29,13 @@ class StepLossABC(abc.ABC):
     def effective_loss_scaling(self) -> TensorDict: ...
 
     @abc.abstractmethod
-    def step_is_optimized(self, step: int) -> bool:
-        """Returns True if the step is less than to the number of
-        steps contributing to the loss.
+    def step_is_optimized(self, step: int, n_total_steps: int | None = None) -> bool:
```
elynnwu: [nit] you can store `n_total_steps` in `LossContributions` instead of having to pass it per call, would also keep this ABC class clean. But I see that we only really have one implementation of `StepLossABC` so it's not necessary either

fme/coupled/test_stepper_integrations.py
```diff
@@ -325,45 +311,6 @@ def test_unoptimized_steps_detached(atmos_n_steps):
             assert has_grad, f"ocean step {step.step} should require grad"
 
 
-@pytest.mark.parametrize("atmos_n_steps", [1, 2])
-def test_unoptimized_steps_detached_with_gradient_accumulation(atmos_n_steps):
-    """Optimized steps should require grad even when gradient accumulation
-    detaches tensors between steps."""
```
elynnwu, line 331: Is there a replacement test with `use_gradient_accumulation` and `optimize_last_step_only`?

fme/coupled/loss.py
```diff
@@ -104,30 +114,39 @@ def __init__(
...
+        return step < self._n_steps
 
     def __call__(
         self, prediction: StepPredictionABC, target_data: TensorMapping
     ) -> torch.Tensor:
-        if self.step_is_optimized(prediction.step):
-            return self._weight * self._loss(
-                prediction.data, target_data, prediction.step
-            )
-        return torch.tensor(0.0, device=get_device())
```
elynnwu, line 133: If `LossContributions` ever gets called directly in a test or somewhere else, it will get a real loss value for a step that should not be optimized right? Probably better to still check `step_is_optimized`?

fme/coupled/stepper.py
```diff
@@ -1630,36 +1658,47 @@ def _accumulate_loss(
...
+                grad_context = contextlib.nullcontext() if optimize else torch.no_grad()
+                with grad_context:
+                    gen_step = next(output_iterator)
```
elynnwu, line 1677: Consider adding a check like: 
```python
assert gen_step.realm == "atmosphere" and gen_step.step == global_atmos_step
```
since we now assume the generator always does `n_inner_steps` of atmosphere then 1 ocean step. This matches `get_prediction_generator`, but since they are in two different spots, better to add a check to make sure it's not mismatched.

fme/coupled/stepper.py
```diff
@@ -1630,36 +1658,47 @@ def _accumulate_loss(
...
+                        n_total_steps=n_total_atmos_steps,
+                        n_ensemble=n_ensemble,
+                        optimization=optimization,
+                        metrics=metrics,
+                        output_list=output_list,
+                    )
+            optimize = self._loss.step_is_optimized("ocean", i_outer, n_outer_steps)
+            grad_context = contextlib.nullcontext() if optimize else torch.no_grad()
+            with grad_context:
+                gen_step = next(output_iterator)
```
elynnwu, line 1694: Same as the inner loop, also check here:
```python
assert gen_step.realm == "ocean" and gen_step.step == i_outer
```

fme/coupled/loss.py
jpdunc23: This is now stored. To do so I moved `n_coupled_steps` from the top level of `TrainConfig` to `CoupledTrainStepperConfig`. I think this is a good change, but it does mean that we'll have to update coupled training configs since the old way of configuring `n_coupled_steps` would be deprecated.

Lmk if that deprecation sounds OK to you.

fme/coupled/test_stepper_integrations.py
jpdunc23: Added `test_optimize_last_step_only_with_gradient_accumulation`

fme/coupled/loss.py
jpdunc23: Restored the `step_is_optimized` check.

fme/coupled/stepper.py
jpdunc23: Added

fme/coupled/stepper.py
jpdunc23: Added

fme/coupled/loss.py
elynnwu: I think this change makes sense, I'm fine with moving `n_coupled_steps` to  `CoupledTrainStepperConfig`

elynnwu (approved):
Should we put a note on this breaking change (`n_coupled_steps`), maybe just in the PR description?

PR #869, opened 2026-02-23
Randomly sampled coupled `LossContributions.n_steps`

jpdunc23:
Allow configuration of `LossContributions.n_steps` with type `TimeLengthProbabilities` for a randomly sampled number of steps (per realm) contributing to the optimization.

Changes:
- `LossContributionsConfig.n_steps` now has signature `n_steps: TimeLengthProbabilities | float = float("inf")`. `LossContributions` has a new method `sample_n_steps()`.

- [x] Tests added


fme/coupled/loss.py
```diff
@@ -51,8 +60,10 @@ class LossContributionsConfig:
     Configuration for loss contributions.
 
     Parameters:
-        n_steps: (optional) The number of consecutive steps contributing to the loss,
-            starting from the first.
+        n_steps: The number of consecutive steps contributing to the loss,
+            starting from the first. Can be a float (including ``inf`` for all
```
AnnaKwa: suggestion to indicate this is still optional
```suggestion
            starting from the first. Can be a float (defaults to ``inf`` for all
```

AnnaKwa (commented):
The implementation LGTM. Is it intended for this new feature to be allowed to interact with `optimize_last_step_only=True` to effectively use a single random step for the loss? If not, can you raise an error for this? If yes, can you add a test?

jpdunc23:
> The implementation LGTM. Is it intended for this new feature to be allowed to interact with `optimize_last_step_only=True` to effectively use a single random step for the loss? If not, can you raise an error for this? If yes, can you add a test?

Yes, this is intended. Test added!

AnnaKwa (approved):

PR #870, opened 2026-02-24
Add regression test for csfno checkpoint output being unchanged

mcgibbon:
These changes are used to test performance improvements in https://github.com/ai2cm/ace/pull/861

Changes:
- Added `asdict` and `from_dict` methods to Context
- Added regression test which ensures the output of the conditional sfno after loading checkpoint data from disk is unchanged

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/test_layers.py
+++ b/fme/core/models/conditional_sfno/test_layers.py
@@ -119,6 +119,9 @@ def test_context_round_trip(
     )
     round_trip = Context.from_dict(context.asdict())
     for attr_name in dir(context):
+        # this ensures the test fails if we ever add additional attributes to
+        # Context and forget to update the test, which would be a problem since
+        # those attributes would not be tested for correct round-tripping
         if not attr_name.startswith("_") and not callable(getattr(context, attr_name)):
             original = getattr(context, attr_name)
             round_tripped = getattr(round_trip, attr_name)


fme/core/models/conditional_sfno/test_layers.py
```diff
@@ -64,3 +64,41 @@ def test_conditional_layer_norm(
...
+    context_embedding_pos = torch.randn(1, embed_dim_pos, *img_shape, device=device)
+    context = Context(
+        embedding_scalar=context_embedding_scalar,
+        noise=context_embedding_noise,
+        labels=context_embedding_labels,
+        embedding_pos=context_embedding_pos,
+    )
+    round_trip = Context.from_dict(context.asdict())
+    for attr_name in dir(context):
+        if not attr_name.startswith("_") and not callable(getattr(context, attr_name)):
```
frodre, line 125: Could use an intermediate variable / comment to hint at what you're targeting here in the conditional or just spell out the names you want to check in a sequence since they are also specified above.  I had to go back into the `Context` definition to try and get more information.

frodre: It seems like it's just checking the top-level dataclass attributes, if it's testing more than that it would be good to know.

fme/core/models/conditional_sfno/test_layers.py
```diff
@@ -64,3 +64,41 @@ def test_conditional_layer_norm(
...
+            original = getattr(context, attr_name)
+            round_tripped = getattr(round_trip, attr_name)
+            if original is None:
```
frodre, line 128: Does this case happen?

fme/core/models/conditional_sfno/test_layers.py
frodre: the ai companion suggested `dataclass.fields` could be an option to extract just the top-level info rather than `dir()`, if that's the intention

fme/core/models/conditional_sfno/test_sfnonet.py
```diff
@@ -155,6 +161,38 @@ def test_sfnonet_output_is_unchanged():
     )
 
 
+def cache_model_state(
```
frodre: Not a big deal since this is a test helper, but loading supersedes inputs which "cache" doesn't imply necessarily
```suggestion
def load_or_cache_model_state(
```

fme/core/models/conditional_sfno/test_layers.py
mcgibbon: > It seems like it's just checking the top-level dataclass attributes, if it's testing more than that it would be good to know.

Currently it is, but I'd like the test to fail if we ever add additional attributes (which may or may not be dataclass fields).

fme/core/models/conditional_sfno/test_layers.py
mcgibbon: I thought it did when the dim=0 but I don't think it does. Will update.

fme/core/models/conditional_sfno/test_layers.py
frodre: Cool, I'd spell that out at least in a comment

PR #871, opened 2026-02-24
Add benchmark for SongUnetv2

AnnaKwa:
This PR adds benchmarks for the SongUNetv2 module. The variants are with/without using bfloat16 and apex group norm.
It also adds the option for recording additional diagnostics beyond timer and memory. For this benchmark I wanted to record the number of memory format conversions that occur to check that this is zero for properly configured (allowed set of nchannels) SongUNetv2 models using apex group norm. The output also includes which layers trigger the conversion if this is nonzero. 

e.g. 
```
  "diagnostics": {
    "n_channels_last_weights": 119,
    "n_total_4d_weights": 119,
    "n_conversions_to_channels_last": 0,
    "n_conversions_from_channels_last": 0,
    "conversion_layers": []
  }
```
image

example of child timings within decoder block:
image

The regression tests only cover the cases without apex group norm because the testing environment doesn't have apex installed.


fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def _get_amp_context(self):
+        if self.use_amp_bf16:
+            return torch.amp.autocast(get_device().type, dtype=torch.bfloat16)
+        return torch.amp.autocast(get_device().type, enabled=False)
+
+    def get_diagnostics(self) -> dict:
+        return self._channels_last_diagnostics.to_dict()
+
+    def run_instance(self, timer) -> TensorDict:
```
mcgibbon: Issue: This API gives two ways to get diagnostics, and it's unexpected that they're gathered separately from `run_instance`.

Suggestion: rather than add a new `get_diagnostics` method whose results aren't specific to any `run_instance` invocation, update the return type of `run_instance` to incorporate these diagnostics somehow. I guess the two types of data returned are "regression" and "diagnostic" data, so maybe a type that has two `dict[str: Tensor]` attributes, one for each of these?

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -604,88 +612,92 @@ def forward(self, x, noise_labels, class_labels, augment_labels=None):
...
-                        > self.checkpoint_threshold
-                    ):
-                        # self.checkpoint = checkpoint?
-                        # else: self.checkpoint  = lambda(block,x,emb:block(x,emb))
-                        x = checkpoint(block, x, emb, use_reentrant=False)
+        with timer.child("encoder") as enc_timer:
+            skips = []
+            aux = x
+            for name, block in self.enc.items():
+                with enc_timer.child(name):
```
mcgibbon: Is it helpful to have a child timer for each block here? Maybe the answer is yes? But alternatively you might consider just having the encoder and decoder children and leaving it there. I found the child blocks were most helpful when they were a) large enough to be a meaningful amount of the time, where I might consider optimizing it, and b) understandable and interpretable by a name for a set of operations. Right now the plot has a lot of unreadable boxes.

The boxes are in order of execution though - it looks like what's happening is a lot of time is being spent in the first and last blocks (the highest resolution/least coarsening?). So maybe this is a useful piece of information to get out of the benchmark.

mcgibbon:
Maybe it would be more helpful to have blocks for "###x###_blocks" all together for example, separate from the _up time.

If you have the called modules also take in Timer, they would be capable of invoking the .child call inside their own forward function, is also a possible way to accomplish it. That way a given "kind" of call always goes into the same bucket to be optimized.

fme/core/benchmark/test_benchmark.py
```diff
@@ -18,6 +18,7 @@
 def test_run_benchmark():
     def benchmark_fn(timer):
         torch.cuda._sleep(100_000_000)
+        return {}
```
AnnaKwa, line 21: Expects tensor dict returned, if not will have an error when test_regression checks for key.

mcgibbon: Suggestion (optional): Maybe we can update the return type to TensorDict | None, and have None act as a sentinel that indicates regression is not implemented for this checkpoint? Like, it should not be None (and we should check-raise) if the "get regression initial condition" method is implemented.

fme/core/benchmark/test_benchmark.py
```diff
@@ -50,7 +53,20 @@ def test_regression(benchmark_name: str):
         # If run_regression returns something,
         # we expect it to be a TensorDict of results
         assert isinstance(regression_result, dict)
-        validate_tensor_dict(
-            regression_result,
-            os.path.join(DIR, "testdata", f"{benchmark_name}-regression.pt"),
-        )
+        if "diagnostics" in regression_result:
+            # Split into two files so we don't have to check nested dicts
```
mcgibbon: Issue: There's no constraint that regression_result contain "output" as a key, and if it doesn't this will error.

Suggestion: Update validate_tensor_dict to work on nested dicts instead, using a recursive helper function. It makes this code a bit cleaner/clearer, and avoids needing twice as many regression files.

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+            out_channels=4,
+            label_dim=0,
+            model_channels=64,
+            channel_mult=[1, 2, 2, 2],
+            use_apex_gn=False,
+        )
+
+class SongUNetv2BenchmarkBf16(SongUNetv2Benchmark):
+    @classmethod
+    def new(cls) -> Self:
```
mcgibbon, line 282: Comment: I am wary of any inheritance with features added/changed but this is kinda nice.

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+
+    @classmethod
+    def new_for_regression(cls) -> Self | None:
+        return None
+
+
+register_benchmark("songunetv2")(SongUNetv2Benchmark)
+register_benchmark("songunetv2_bf16")(SongUNetv2BenchmarkBf16)
+register_benchmark("songunetv2_apex")(SongUNetv2BenchmarkApex)
+register_benchmark("songunetv2_apex_bf16")(SongUNetv2BenchmarkApexBf16)
```
mcgibbon, line 352: Question: How long do these added benchmarks take to run?

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -604,88 +613,98 @@ def forward(self, x, noise_labels, class_labels, augment_labels=None):
...
-                else:
-                    x = block(x)
-                skips.append(x)
+        with timer.child("encoder") as enc_timer:
+            skips = []
+            aux = x
+            for level_name, level_blocks in itertools.groupby(
+                self.enc.items(), key=lambda item: item[0].split("_", 1)[0]
+            ):
+                with enc_timer.child(level_name):
```
mcgibbon: Comment: I don't suggest changing it this way, but wanted to note that if you call `enc_timer.child(level_name):` multiple times with the same level_name, it will add the times together when it shows the plot. For example if you had kept the loops as you had it the first time but transformed the names to level names. The main difference is the json logs will indicate the child was called more times. I do slightly prefer this way but wanted to point it out as a feature.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
```diff
@@ -604,88 +613,98 @@ def forward(self, x, noise_labels, class_labels, augment_labels=None):
...
+                    for name, block in level_blocks:
+                        if "aux_down" in name:
+                            aux = block(aux)
```
mcgibbon, line 653: Suggestion (optional): Your timings before showed the down/up blocks and the block blocks were taking the most compute. It won't show up in the plot, but you could consider adding `.child()` timer calls with names "down", "up", and "block" around the corresponding blocks here. They'll already be nested under the resolution, so there's no need to make the names more complex. This could help if you ever try to speed up the block operations for example, or if they get slower due to some change.

fme/downscaling/modules/physicsnemo_unets_v2/unets.py
mcgibbon: If you decide to add further child timers you can do `with enc_timer.child(level_name) as level_timer:`

mcgibbon:
A test seems to be failing.

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
AnnaKwa: ~14 s for the regular benchmark and twice as long for the bf16 regression. Since it's on CPU that is very slow for the unets and apparently bf16 on CPU is also extra slow.

fme/core/benchmark/benchmark.py
```diff
@@ -132,7 +130,17 @@ def blend_with_white(
                 if part not in root.children:
                     raise ValueError(f"Child '{child}' not found in timer results.")
                 root = root.children[part]
-        root_avg = avg_time(root)
+        root_count = root.count
+
+        def avg_total_time_per_iter(t: TimerResult) -> float:
```
AnnaKwa: Edited this function to add up the times if a timer was repeated. Moved so that root.count was available to reference.

fme/core/benchmark/test_benchmark.py
AnnaKwa: Updated to work on nested dicts

fme/core/benchmark/benchmark.py
mcgibbon: Issue: I'm a little confused when I read "average total time", since these are contradictory, and it's not obvious to me what "per iter" means at first read (I can tell you mean per benchmark iteration, but at first I thought it meant per child call).

Suggestions: rename to `avg_time_per_root` or `avg_time_per_root_iter` to make it clearer, or back to `avg_time` to leave it vague but not misleading/confusing.

fme/core/benchmark/benchmark.py
```diff
@@ -308,13 +318,14 @@ def run_benchmark(cls, iters=10, warmup=1) -> BenchmarkResult:
             cpu_timer.record_start()
             for _ in range(iters):
                 with timer:
-                    benchmark.run_instance(timer)
+                    benchmark_result = benchmark.run_instance(timer)
```
mcgibbon: Issue: It's a little confusing having this called a `benchmark_result` but not being the same type as the returned value which is a BenchmarkResult. Maybe `iter_result` or `last_iter_result` instead?

Suggestion (optional, better in a later PR if at all): Which also raises a question, do we want just the last diagnostics or do we want the total/average across results? You could keep a running total owned by this method (instead of by the benchmark), similarly to how the local instance of `timer` is meant to keep the running total of times.

fme/core/benchmark/run.py
```diff
@@ -12,6 +12,14 @@
 from fme.core.benchmark.benchmark import get_benchmarks
 from fme.core.wandb import WandB
 
+
+def _json_default(obj):
+    """json.dumps ``default`` hook that converts tensors to Python scalars."""
+    if isinstance(obj, torch.Tensor):
+        return obj.item()
```
mcgibbon: Question: What will happen if the tensor is not a scalar? Is that behavior expected?

fme/core/benchmark/test_benchmark.py
mcgibbon: Re-noting this optional suggestion since you didn't reply on it, it's still optional.

fme/core/benchmark/test_benchmark.py
```diff
@@ -40,7 +41,9 @@ def test_benchmarks_are_not_empty():
 
 
 @pytest.mark.parametrize("benchmark_name", BENCHMARKS.keys())
-def test_regression(benchmark_name: str):
+def test_regression(benchmark_name: str, very_fast_only: bool):
```
mcgibbon: Question: Why is this now required? It means we no longer get very-fast coverage for the existing benchmarks also. Is it possible to set up the regression with a small enough problem (maybe a smaller spatial domain in this case) that it runs fast enough?

fme/core/testing/regression.py
```diff
@@ -16,12 +15,51 @@ def validate_tensor(x: torch.Tensor, filename: str, **assert_close_kwargs):
...
+    for k, v in x.items():
+        if isinstance(v, torch.Tensor):
+            result[k] = v.cpu()
+        else:
+            result[k] = _to_cpu(v)
+    return result
+
+
+def _assert_close(
+    x: NestedTensorDict, y: NestedTensorDict, prefix: str, **assert_close_kwargs
```
mcgibbon: Issue: The prefix logic is confusing, and I don't understand what its use case is. It seems to be hard-coded to "" below. Can we remove it?

fme/core/testing/regression.py
```diff
@@ -16,12 +15,51 @@ def validate_tensor(x: torch.Tensor, filename: str, **assert_close_kwargs):
...
+):
+    for k, v in x.items():
+        key_path = f"{prefix}.{k}" if prefix else k
+        assert k in y, f"Key '{key_path}' missing from regression file"
+        y_val = y[k]
+        if isinstance(v, torch.Tensor):
+            assert isinstance(
+                y_val, torch.Tensor
+            ), f"Expected tensor at '{key_path}' but got dict"
```
mcgibbon: Issue: This assertion error will be wrong/misleading if `y_val` is another non-tensor type, such as None. I know in practice the way you're currently using this function that can't happen (since it always comes from a loaded dict), but the call signature of this helper function doesn't know that.

Suggestion: Maybe use a "got {type(y_val)}"?

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
mcgibbon: I thought the benchmarks only run on GPU / that it raises an error if we try to benchmark on CPU? The CUDA timer will raise an error if it's run on CPU.

Are you saying the regression tests take 14s? That's quite long, is there any way to get code pathway coverage with a smaller config?

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
```diff
@@ -0,0 +1,352 @@
...
+        return cls._new_with_params(
+            img_resolution=32,
+            B=1,
+            in_channels=6,
+            out_channels=4,
+            label_dim=0,
+            model_channels=64,
+            channel_mult=[1, 2, 2, 2],
+            use_apex_gn=False,
+        )
```
mcgibbon, line 278: Sorry for not catching this earlier.

Issue: the configuration here is identical to the configuration for `new`. You want to make the `new` configuration one that is helpful for debugging/diagnosing optimization effects on GPU and fully occupies the GPU. In `new_for_regression`, you want a config that runs sufficiently fast for regression testing on CPU, and produces a small enough checkpoint to save to the repo.

Suggestion:

```suggestion
        return cls._new_with_params(
            img_resolution=8,
            B=1,
            in_channels=6,
            out_channels=4,
            label_dim=0,
            model_channels=16,
            channel_mult=[1, 2],
            use_apex_gn=False,
        )
```
(assuming len(channel_mult) is what determines the depth of the u-net)

I would suggest similar changes in the other benchmark(s). With the faster runtime you could consider including regression on more of the benchmarks.

fme/core/benchmark/benchmark.py
AnnaKwa: > do we want just the last diagnostics or do we want the total/average across results
Good point, probably the latter but in a later PR

fme/core/benchmark/run.py
AnnaKwa: Good catch, fixed so non-scalar tensors are also handled.

This is a good example of why I was wary of using AI code tools to help write functions... I had a feeling I'd get complacent and stuff like this would slip through.

fme/downscaling/modules/physicsnemo_unets_v2/benchmark.py
AnnaKwa: Thanks, the existing two are now on par with the duration of the others (<.1s). I removed the --very-fast-only part. 

I also updated the regular `new` params to match the apex benchmarks for better comparison.

I'll still exclude the 'with apex' benchmarks from regression testing as the unit testing environment doesn't have apex installed (it takes a while).

AnnaKwa:
> Suggestion (optional): Maybe we can update the return type to TensorDict | None, and have None act as a sentinel that indicates regression is not implemented for this checkpoint? Like, it should not be None (and we should check-raise) if the "get regression initial condition" method is implemented.

for some reason could reply to original comment

Left as is, since this isn't a regression test and the later references in `BenchmarkResult` assume a dict type return (`diagnostics=benchmark_result.get("diagnostics", {})`)

fme/core/benchmark/benchmark.py
mcgibbon: Suggestion (optional): Renaming to `last_iter_result` is still a good idea for this PR.

PR #873, opened 2026-02-25
Simplify ERA5 data pipeline and update to zarr v3

oliverwm1:
Given some new zarr's available from the google ARCO-ERA5 datset, we can substantially simplify the ERA5 data processing pipeline by reducing the number of input streams. This PR does that, and additionally switches to saving a zarr v3, and uses xESMF instead of metview for regridding

Notebook here compares a snapshot of data: https://github.com/ai2cm/explore2/blob/main/oliverwm/2026-03-02-compare-new-era5-pipeline.ipynb


brianhenn:
It looks like we did already [update](https://github.com/ai2cm/ace/commit/a83130821c045e9bec665270bffb2345281c74fa) the container registry name prior to this, so I don't think there is anything else to be done re: my comment about getting the dataflow cloud runner to work.

scripts/era5/Makefile
spencerkclark: I'm assuming these additional packages are mainly for interactive development—should we remove `metview` and `cfgrib` since I expect they are no longer needed?
```suggestion
	conda run --no-capture-output -n era5-ingestion pip install jupyterlab matplotlib
```

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -185,136 +128,103 @@ def grid_attribute_fix(ds, names_to_fix, reference_name):
...
-    if os.path.exists(path):
-        os.remove(path)
-    del fs
+# ---------------------------------------------------------------------------
+# Physics utilities
+# ---------------------------------------------------------------------------
 
 
 def _saturation_vapor_pressure(t: xr.DataArray) -> xr.DataArray:
-    """https://metview.readthedocs.io/en/latest/api/functions/saturation_vapour_pressure.html"""  # noqa: E501
```
spencerkclark, line 317: This reference is still useful.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -325,320 +235,179 @@ def _saturation_vapor_pressure(t: xr.DataArray) -> xr.DataArray:
 def _specific_humidity_from_dewpoint(
     dewpoint: xr.DataArray, pressure: xr.DataArray
 ) -> xr.DataArray:
-    """https://metview.readthedocs.io/en/latest/api/functions/specific_humidity_from_dewpoint.html
-    https://metview.readthedocs.io/en/latest/api/functions/vapour_pressure.html"""
```
spencerkclark, line 329: These references are still useful.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -325,320 +235,179 @@ def _saturation_vapor_pressure(t: xr.DataArray) -> xr.DataArray:
...
-) -> xr.Dataset:
-    n_output_layers = len(output_layer_indices) - 1
-    rename_dict = _get_native_rename_dict(n_output_layers)
+def open_full_37(variables, time_slice) -> xr.Dataset:
+    """Open variables from the full_37 ARCO-ERA5 store."""
+    ds = xr.open_zarr(_make_zarr_store(URL_FULL_37), chunks=None)
+    ds = ds[variables]
+    if time_slice is not None:
+        _check_time_bounds(ds, time_slice)
+        ds = ds.sel(time=time_slice)
```
spencerkclark: nit: I suspect we will not call this without specifying the `time_slice`, so we can likely remove this check similar to `open_model_level`.
```suggestion
    _check_time_bounds(ds, time_slice)
    ds = ds.sel(time=time_slice)
```

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -649,128 +418,332 @@ def _process_quarter_degree_data_sfc_an(ds, invariant_ds, output_grid):
...
 
 
-def process_quarter_degree_data_mean_flux(key, ds, output_grid=DEFAULT_OUTPUT_GRID):
-    output_ds = _process_quarter_degree_data_mean_flux(ds, output_grid)
-    new_key = key.replace(vars=frozenset(output_ds.keys()))
-    return new_key, output_ds
+def process_pressure_level_data(key, ds, output_grid=DEFAULT_OUTPUT_GRID):
+    output = _process_pressure_level_data(ds, output_grid)
+    new_key = key.replace(
+        offsets={"time": key.offsets["time"], "latitude": 0, "longitude": 0},
```
spencerkclark: Is respecifying the `offsets` needed?  Mainly just going off of symmetry with how this is addressed in `process_surface_analysis`.

spencerkclark, line 477: Just curious, is the reason we need to redefine the `offsets` that we need to remove the `"level"` dimension?  Would it be clearer to write this as:
```suggestion
    offsets = {dim: key.offsets[dim] for dim in output.dims}
    new_key = key.replace(
        offsets=offsets,
```
This could make it more obvious that we are not changing anything about the `offsets` of the original `key` other than to filter out values for dimensions not present in the `output` dataset.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -649,128 +418,332 @@ def _process_quarter_degree_data_sfc_an(ds, invariant_ds, output_grid):
...
+    ak=None,
+    bk=None,
+    output_grid=DEFAULT_OUTPUT_GRID,
+    output_layer_indices=DEFAULT_OUTPUT_LAYER_INDICES,
+):
+    output = _process_model_level_data(
+        ds, ds_surface.sel(time=ds.time), ak, bk, output_grid, output_layer_indices
+    )
+    new_key = key.replace(
+        offsets={"time": key.offsets["time"], "latitude": 0, "longitude": 0},
```
spencerkclark, line 662: Similar question here regarding the offsets (I think this time we are just removing the `"hybrid"` dimension?):
```suggestion
    offsets = {dim: key.offsets[dim] for dim in output.dims}
    new_key = key.replace(
        offsets=offsets,
```

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -185,136 +128,103 @@ def grid_attribute_fix(ds, names_to_fix, reference_name):
...
+
+
+def _regrid(ds: xr.Dataset, output_grid: str) -> xr.Dataset:
+    """Regrid a dataset from 0.25° regular lat-lon to a Gaussian target grid."""
+    regridder = _get_regridder(output_grid)
+    # Rename coords to lat/lon for xESMF
+    ds = ds.rename({"latitude": "lat", "longitude": "lon"})
+    # Ensure south-to-north
+    if ds.lat.values[0] > ds.lat.values[-1]:
+        ds = ds.reindex(lat=ds.lat[::-1])
```
spencerkclark: nit: I find `sortby` to be slightly more readable for this use-case:
```suggestion
    ds = ds.sortby("lat")
```

scripts/era5/Makefile
oliverwm1: yes agreed

scripts/era5/pipeline/xr-beam-pipeline.py
oliverwm1: thanks, Claude got a bit overeager with deleting these

scripts/era5/pipeline/xr-beam-pipeline.py
oliverwm1: yes, good idea.

scripts/era5/pipeline/xr-beam-pipeline.py
oliverwm1: Updated as you suggest

scripts/era5/pipeline/xr-beam-pipeline.py
oliverwm1: actually, this change led to an error during the chunk consolidate step, so I am going to revert it.

PR #874, opened 2026-02-25
Configure training noise distribution

AnnaKwa:
Currently, downscaling diffusion model training is hard coded to use a log normal distribution for the preconditioning noise. This is configured via the `p_mean, p_std` fields in the top level of the `DiffusionModelConfig`.

This PR
- updates the `condition_with_noise_for_training` function to take a more general `NoiseDistribution` object and call its `.sample` method instead of taking in `p_mean, p_std` args and hard coding the sampling from a log normal distribution
- creates `LogNormalNoiseDistribution` (replicates previous behavior) and `LogUniformNoiseDistribution` (new) classes
 - adds a `training_noise_distribution` field to `DiffusionModelConfig`. The distribution to use is inferred by dacite as `LogNormalNoiseDistribution` is parametrized by `p_mean, p_std` while `LogUniformNoiseDistribution` is parameterized by `p_min, p_max`. 
 

Backwards compatibility with old style configs is maintained. If both `p_mean, p_max` and `training_noise_distribution` are configured, an error will be raised.


fme/downscaling/models.py
```diff
@@ -204,51 +249,6 @@ def _separate_interleaved_samples(tensor: torch.Tensor, n_samples: int) -> torch
     return tensor.reshape(n_batch, n_samples, *tensor.shape[1:])
 
 
-@dataclasses.dataclass
```
AnnaKwa, line 212: Deleted code is moved to noise.py

fme/downscaling/noise.py
```diff
@@ -0,0 +1,77 @@
...
+        return (rnd * self.p_std + self.p_mean).exp()
+
+
+@dataclasses.dataclass
+class LogUniformNoiseDistribution(NoiseDistribution):
+    p_min: float
+    p_max: float
+
+    def sample(self, batch_size: int, device: torch.device) -> torch.Tensor:
+        sigma = np.exp(
```
frodre: LogNormal imports from `fme.core.rand`.  Could we put similar distributions in there?  I just want to make sure that we adhere to all of the seed setting infrastructure that exists there.

fme/downscaling/models.py
```diff
@@ -86,31 +91,40 @@ class DiffusionModelConfig:
         in_names: The input variable names for the diffusion model.
         out_names: The output variable names for the diffusion model.
         normalization: The normalization configurations for the diffusion model.
-        p_mean: The mean of noise distribution used during training.
-        p_std: The std of the noise distribution used during training.
+
```
frodre:
```suggestion
```

fme/downscaling/noise.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class NoiseDistribution(abc.ABC):
+    @abc.abstractmethod
+    def sample(self, batch_size: int, device: torch.device) -> torch.Tensor:
+        pass
+
+
+@dataclasses.dataclass
+class LogNormalNoiseDistribution(NoiseDistribution):
```
frodre, line 32: Commenting here, but I think since we have a new component, we could add some basic unit tests that covers the list of available distributions.  Checks that it outputs tensors of the correct shape / type.

fme/downscaling/test_models.py
```diff
@@ -403,3 +404,42 @@ def test_DiffusionModel_generate_on_batch_no_target_arbitrary_input_size():
             n_ensemble,
             *fine_shape,
         )
+
+
```
frodre, line 450: I think it would also be worthwhile to test the mixed config usage errors.

fme/downscaling/models.py
```diff
@@ -119,6 +133,37 @@ def __post_init__(self):
...
+        if self.training_noise_distribution is None and (
+            self.p_mean is None or self.p_std is None
+        ):
+            raise ValueError(
+                "Noise distribution must be specified in training_noise_distribution "
+                "field or in p_mean and p_std fields."
+            )
+
+    @property
+    def noise(self) -> LogNormalNoiseDistribution | LogUniformNoiseDistribution:
```
frodre: Name is a bit non-descript.  I would expand to `noise_distribution` and/or write a docstring description to highlight what's being returned and what it's used for.

frodre (changes requested):
Thanks @AnnaKwa.  Looking good, just a few adjustments and I think it'll be ready to go.

PR #875, opened 2026-02-25
Parameterize compute_enso_index.py with CLI arguments

brianhenn:
Parameterizes `scripts/compute_enso_index/compute_enso_index.py` with CLI arguments so it can be run on different datasets without code changes, and adds a dedicated Makefile target for the ERA5 AIMIP dataset.                                 
                                         
  Changes:                                                                                                                                                                                                                                          
  - `scripts/compute_enso_index/compute_enso_index.py` — added CLI arguments for SST variable, dimension names, ocean mask source, time range, output file, and optional detrending; added dask progress bar; fixed cftime/datetime64 compatibility
  - `scripts/compute_enso_index/Makefile` — added `era5_aimip_enso_index` target with explicit arguments for the ERA5 AIMIP zarr dataset                                                                                                            
                                                            
  - [ ] Tests added

### DIFF
--- a/fme/ace/data_loading/benchmark.py
+++ b/fme/ace/data_loading/benchmark.py
@@ -15,6 +15,7 @@ import yaml
 
 from fme.core import logging_utils
 from fme.core.dicts import to_flat_dict
+from fme.core.distributed.distributed import Distributed
 from fme.core.timing import GlobalTimer
 from fme.core.wandb import WandB
 
@@ -122,4 +123,5 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    with Distributed.context():
+        main()
--- a/fme/core/benchmark/run.py
+++ b/fme/core/benchmark/run.py
@@ -10,6 +10,7 @@ import sys
 import torch
 
 from fme.core.benchmark.benchmark import get_benchmarks
+from fme.core.distributed.distributed import Distributed
 from fme.core.wandb import WandB
 
 RESULTS_PATH = pathlib.Path(os.path.abspath(os.path.dirname(__file__))) / "results"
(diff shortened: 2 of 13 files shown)


scripts/compute_enso_index/compute_enso_index.py
```diff
@@ -0,0 +1,171 @@
...
+        .rename({mask_lon_dim: lon_dim, mask_lat_dim: lat_dim})
+    )
+    ocean_fraction_1deg = ocean_fraction.interp_like(template).load()
+    ocean_mask = xr.where(ocean_fraction_1deg > 0.5, True, False)
+    return ocean_mask
+
+
+def get_time_average(da):
+    # this version of xarray's resample method doesn't allow
+    # data shifting to create a centered 3-month average, so do it manually
```
elynnwu: [nit] Is this a requirement for ace's xarray version? I know for example under `scripts/data_process` we have a different xarray version.

scripts/compute_enso_index/.gitignore
```diff
@@ -0,0 +1,2 @@
+*index.py
```
elynnwu, line 1: Just curious why we don't want to keep the files in git, is this purely for analysis only? In the past when I did this for E3SM, I ended up overriding `historical_index.py` file so that they show up on wandb. Maybe that's not your intent here?

scripts/compute_enso_index/compute_enso_index.py
brianhenn: Good point, I am running this locally off of the `fme` conda environment, and at some point I hit the issue that required this. I think maybe the best is to just be clearer what environment is expected to be used to run the script; I'll add a comment.

scripts/compute_enso_index/.gitignore
brianhenn: For the immediate AIMIP use case I am putting the output in a notebook/other script somewhere. Perhaps we should have a catalog of indices in `historical_index.py` for different SST datasets, but not sure how pressing that need is.

elynnwu: Not pressing. We can add it when there's a use case for that, just wanted to check.

elynnwu (approved):

PR #877, opened 2026-02-26
Remove unused sht copy

mcgibbon:
In an earlier PR I made a copy of RealSHT and InverseRealSHT to allow local changes to the function without changing the API in torch_harmonics. However, in a PR after that I updated the patched torch_harmonics object, meaning the copy is unused except in a benchmark.

This PR removes the copy and directs the benchmark back to the updated implementation.

We should consider refactoring to use the sht_fix implementation directly instead of monkey-patching it into torch_harmonics, now that we're changing its API. Also worth noting that changes to torch_harmonics break regression backwards compatibility (though we should consider porting these to our code).

Changes:
- No public API changes

- [ ] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/benchmark.py
+++ b/fme/core/models/conditional_sfno/benchmark.py
@@ -1,7 +1,6 @@
 from typing import Self
 
 import torch
-from torch_harmonics import InverseRealSHT, RealSHT
 
 from fme.core.benchmark.benchmark import BenchmarkABC, register_benchmark
 from fme.core.benchmark.timer import Timer


fme/core/models/conditional_sfno/benchmark.py
```diff
@@ -7,8 +7,8 @@
 from fme.core.device import get_device
 from fme.core.models.conditional_sfno.layers import Context, ContextConfig
 from fme.core.models.conditional_sfno.sfnonet import FourierNeuralOperatorBlock
-from fme.core.models.conditional_sfno.sht import InverseRealSHT, RealSHT
 from fme.core.typing_ import TensorDict
+from fme.sht_fix import InverseRealSHT, RealSHT
```
mcgibbon, line 11: I had to import this directly because when running this test directly, `fme` isn't imported, so sht_fix doesn't monkey patch these objects, and `torch_harmonics` has a different implementation now that breaks regression tests.

Issue #693, opened 2025-11-09
SSR ensemble metric not getting logged to Wandb correctly


Arcomano1234:
As part of ai2cm/full-model#2264 I noticed that SSR was not being logged to Wandb ([example](https://wandb.ai/ai2cm/ace/runs/r1shdbfb?nw=nwusertroya)). I went back to my previous training runs and found that only 9 out of ~60 variables were being reported on Wandb during inference ([example](https://wandb.ai/ai2cm/ace/runs/4ikfm6oj?nw=nwusertroya) and [example](https://wandb.ai/ai2cm/ace/runs/nskivquh?nw=nwusertroya)). Note it is always the same 9 variables being plotted. This has been happening since at least July 21st.

PR #878, opened 2026-02-26
Add `ValidationConfig` for evaluator inference

jpdunc23:
Adds `validation_config: ValidationConfig` to `InferenceEvaluatorConfig` in `fme.ace`. When configured, validation runs first and logs metrics to step 0 of the wandb run.

Changes:
- Adds generic `run_validation` helper than can be reused by other modules.

- [x] Tests added

### DIFF
--- a/fme/ace/data_loading/batch_data.py
+++ b/fme/ace/data_loading/batch_data.py
@@ -12,6 +12,7 @@ from torch.utils.data import default_collate
 
 from fme.core.dataset.dataset import DatasetItem
 from fme.core.device import get_device
+from fme.core.distributed import Distributed
 from fme.core.labels import BatchLabels, LabelEncoding
 from fme.core.tensors import repeat_interleave_batch_dim, unfold_ensemble_dim
 from fme.core.typing_ import EnsembleTensorDict, TensorDict, TensorMapping
@@ -168,6 +169,18 @@ class BatchData:
             horizontal_dims=self.horizontal_dims,
             epoch=self.epoch,
             labels=self.labels.to(device) if self.labels is not None else None,
+        )
+
+    def scatter_spatial(self, global_img_shape: tuple[int, int]) -> "BatchData":
+        """Slice data tensors to the local spatial chunk."""
+        dist = Distributed.get_instance()
+        return self.__class__(
+            data=dist.scatter_spatial(dict(self.data), global_img_shape),
+            time=self.time,
+            horizontal_dims=self.horizontal_dims,
+            epoch=self.epoch,
+            labels=self.labels,
+            n_ensemble=self.n_ensemble,
         )
 
     def to_cpu(self) -> "BatchData":
--- a/fme/ace/data_loading/config.py
+++ b/fme/ace/data_loading/config.py
@@ -78,10 +78,11 @@ class DataLoaderConfig:
 
     def __post_init__(self):
         dist = Distributed.get_instance()
-        if self.batch_size % dist.world_size != 0:
+        if self.batch_size % dist.total_data_parallel_ranks != 0:
             raise ValueError(
-                "batch_size must be divisible by the number of parallel "
-                f"workers, got {self.batch_size} and {dist.world_size}"
+                "batch_size must be divisible by the number of data-parallel "
+                f"workers, got {self.batch_size} and "
+                f"{dist.total_data_parallel_ranks}"
             )
         self._zarr_engine_used = self.dataset.zarr_engine_used
         if self.time_buffer < 0:
--- a/fme/ace/data_loading/getters.py
+++ b/fme/ace/data_loading/getters.py
@@ -49,8 +49,13 @@ def _get_sampler(
 ) -> torch.utils.data.Sampler:
     dist = Distributed.get_instance()
     if sample_with_replacement_dataset_size is not None:
+        dist.require_no_spatial_parallelism(
+            "sample_with_replacement is not supported with spatial "
+            "parallelism. Spatial co-ranks would draw different samples, "
+            "producing corrupted data after scatter_spatial reassembly."
+        )
         local_sample_with_replacement_dataset_size = (
-            sample_with_replacement_dataset_size // dist.world_size
+            sample_with_replacement_dataset_size // dist.total_data_parallel_ranks
         )
         sampler = torch.utils.data.RandomSampler(
             dataset,
--- a/fme/ace/data_loading/gridded_data.py
+++ b/fme/ace/data_loading/gridded_data.py
@@ -50,7 +50,9 @@ class GriddedData(GriddedDataABC[BatchData]):
             will be on the current device.
         """
         self._loader = loader
-        self._properties = properties.to_device()
+        shape = properties.horizontal_coordinates.shape
+        self._global_img_shape: tuple[int, int] = (shape[-2], shape[-1])
+        self._properties = properties.to_device().localize()
         self._timestep = self._properties.timestep
         self._vertical_coordinate = self._properties.vertical_coordinate
         self._mask_provider = self._properties.mask_provider
@@ -72,7 +74,8 @@ class GriddedData(GriddedDataABC[BatchData]):
         self, base_loader: DataLoader[BatchData]
     ) -> DataLoader[BatchData]:
         def modify_and_on_device(batch: BatchData) -> BatchData:
-            return self._modifier(batch).to_device()
+            batch = self._modifier(batch)
+            return batch.to_device().scatter_spatial(self._global_img_shape)
 
         return SizedMap(modify_and_on_device, base_loader)
 
@@ -174,21 +177,23 @@ class InferenceGriddedData(InferenceDataABC[PrognosticState, BatchData]):
             will be on the current device.
         """
         self._loader = loader
-        self._properties = properties.to_device()
+        shape = properties.horizontal_coordinates.shape
+        self._global_img_shape: tuple[int, int] = (shape[-2], shape[-1])
+        self._properties = properties.to_device().localize()
         self._n_initial_conditions: int | None = None
         if isinstance(initial_condition, PrognosticStateDataRequirements):
             self._initial_condition: PrognosticState = get_initial_condition(
-                loader, initial_condition
+                self.loader, initial_condition
             )
         else:
             self._initial_condition = initial_condition.to_device()
 
     @property
     def loader(self) -> DataLoader[BatchData]:
-        def on_device(batch: BatchData) -> BatchData:
-            return batch.to_device()
+        def scatter_and_on_device(batch: BatchData) -> BatchData:
+            return batch.to_device().scatter_spatial(self._global_img_shape)
 
-        return SizedMap(on_device, self._loader)
+        return SizedMap(scatter_and_on_device, self._loader)
 
     @property
     def variable_metadata(self) -> dict[str, VariableMetadata]:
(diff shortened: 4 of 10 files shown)


jpdunc23:
Example wandb run: https://wandb.ai/ai2cm/ace2-cm4-atmos-noLSM/runs/2rlir6tb

jpdunc23:
I looked into logging the validation metrics in the last step rather than step 0, which would be helpful since Wandb considers last-step metrics as [log summary metrics](https://docs.wandb.ai/models/track/log/log-summary), but I couldn't quite come up with a clean way to do it. It might be worth looking into it again.

fme/ace/inference/evaluator.py
```diff
@@ -127,6 +165,10 @@ class InferenceEvaluatorConfig:
             This should be used with caution, as it may allow the stepper to make
             scientifically invalid predictions, but it can allow running inference with
             incorrectly formatted or missing grid information.
+        validation: Optional configuration for running a one-step validation loop
```
oliverwm1, line 209: I don't think "validation" the right word here. A user could run this part of the eval on any dataset, including the training data of the model being evaluated. The more relevant distinction is that this will compute/log "one step" metrics instead of the autoregressive inference metrics otherwise computed by the inference evaluator.

jpdunc23: > A user could run this part of the eval on any dataset, including the training data of the model being evaluated.

The same can be said of the `validation_loader` and corresponding `validation_aggregator` entries in `TrainConfig`.

> The more relevant distinction is that this will compute/log "one step" metrics

True that most of the logs will be "one step" metrics, but this supports multiple forward steps and reports "val/mean/loss", so calling this `one_step_eval` isn't quite right. 

`validation` still feels most appropriate to me since this is configured in exactly the same way as validation on `TrainConfig` (though that could change, of course).

oliverwm1: > The same can be said of the validation_loader and corresponding validation_aggregator entries in TrainConfig.

The context of `TrainConfig` is different since `TrainConfig` is actually the thing that defines the train/validation split.

> True that most of the logs will be "one step" metrics, but this supports multiple forward steps and reports "val/mean/loss", so calling this one_step_eval isn't quite right.

I agree "one step" is not quite right, but there should be some phrase that gets at "short time horizon metrics computed across many different data samples" to distinguish from the autoregressive metrics which are computed just from a smaller number of initial conditions, but potentially accumulated over long rollouts.

fme/ace/inference/evaluator.py
```diff
@@ -79,6 +88,35 @@ def resolve_variable_metadata(
...
+            to 1 for standard single-step validation.
+        aggregator: Configuration for the one-step validation aggregator.
+        loss: Loss function configuration used to compute the validation loss
+            metric. Defaults to MSE. Set this to match the training loss
+            configuration if you want ``val/mean/loss`` to be directly
+            comparable to training.
+    """
+
+    loader: DataLoaderConfig
+    n_forward_steps: int = 1
```
mcgibbon: Unfortunate that this can't be used to achieve validation metrics for stochastic-n-steps training or for n_ensemble=2.

fme/ace/inference/evaluator.py
mcgibbon: Without significantly refactoring the training code/config (which we could do, but I'm not sure is worth our time), it feels like calling this "validation" is the fastest/most obvious way to tell users what metrics these are computing. It's pretty valuable that this is called the same thing it's called in training, and the user story motivating this has been said as "we want to re-run then validation after the job is done or on different datasets". Calling this "short time horizon metrics" raises more questions than it answers, in my opinion, unless we also call it that somewhere in the training config / when we talk about training.

I do get that validation is "wrong" in the sense that this isn't a train/validation split, but I don't think a user will be confused about that, given they have to specify the dataset inside the config. To me "validation" in our discussions has been shorthand for "evaluation of an independent dataset on exactly the problem the model was trained on / run_on_batch"

fme/ace/inference/evaluator.py
```diff
@@ -281,6 +324,38 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
...
+            n_forward_steps=config.validation.n_forward_steps
+        )
+        valid_data = get_gridded_data(
+            config.validation.loader,
+            requirements=data_requirements,
+            train=False,
+        )
+
+        logging.info("Building validation stepper and aggregator")
+        train_stepper_config = TrainStepperConfig(loss=config.validation.loss)
```
mcgibbon: Why limit to only being able to configure the loss? We could expose the entire TrainStepperConfig, or at least more of it. `train_n_forward_steps` and `n_ensemble` are both useful here.

fme/ace/inference/evaluator.py
Arcomano1234: Yes I was thinking about making that as a follow on PR

fme/ace/inference/evaluator.py
jpdunc23: OK, I think we can probably include that in this PR.

jpdunc23:
Have a run that does stochastic ACE validation with `TimeLengthProbabilities`: https://wandb.ai/ai2cm/ace2-cm4-atmos-noLSM/runs/8ours1g1

A couple of unexpected things:
- ~~There was apparently no `val/mean/loss` logged. I think I just forgot to manually grab it from the `TrainOutput` and add it to the logs.~~ Nevermind, it's there but when I looked wandb wasn't showing it for some reason.
- The [logs](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KKCZDS815B0WPJ4CTWBQTQBJ/logs?jobId=01KKCZDSRKCBDCY16TNKA3PC9S) this time for some reason don't include any entries like `Inference: processing output from window 207 of 1166.` This wasn't an issue with the [other run](https://beaker.org/orgs/ai2/workspaces/ace/work/01KJRCA1F150WAGAXD6HVA241Z/logs?jobId=01KJRCA1RMYS9D06QYSX1W4S98).

Arcomano1234:
> * ~There was apparently no `val/mean/loss` logged. I think I just forgot to manually grab it from the `TrainOutput` and add it to the logs.~ Nevermind, it's there but when I looked wandb wasn't showing it for some reason.
> * The [logs](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KKCZDS815B0WPJ4CTWBQTQBJ/logs?jobId=01KKCZDSRKCBDCY16TNKA3PC9S) this time for some reason don't include any entries like `Inference: processing output from window 207 of 1166.` This wasn't an issue with the [other run](https://beaker.org/orgs/ai2/workspaces/ace/work/01KJRCA1F150WAGAXD6HVA241Z/logs?jobId=01KJRCA1RMYS9D06QYSX1W4S98).

I wonder if the first issue is some how related to #693 

That second issue is interesting. I have not seen that before not sure why this would be related to a stochastic model with `TimeLengthProbabilities `

fme/ace/inference/evaluator.py
mcgibbon: Whatever works for you.

fme/ace/inference/evaluator.py
```diff
@@ -81,6 +93,78 @@ def resolve_variable_metadata(
...
+            to 1 for standard single-step validation.
+        aggregator: Configuration for the one-step validation aggregator.
+        stepper_training: Training-specific configuration including loss, ensemble
+            settings, and forward step scheduling. Set this to match the training
+            configuration if you want ``val/mean/loss`` to be directly comparable.
+
+    """
+
+    loader: DataLoaderConfig
+    n_forward_steps: int | None = None
```
mcgibbon: Question: Why include n_forward_steps if we already have `stepper_training.n_forward_steps`? You don't need to maintain backwards compatibility here, and it would simplify the code to use just one.

fme/ace/inference/evaluator.py
```diff
@@ -286,6 +375,38 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
         stepper_all_names=stepper_config.all_names,
     )
     dataset_info = data.dataset_info.update_variable_metadata(variable_metadata)
+
+    if config.validation is not None:
+        timer.stop()
```
mcgibbon, line 373: Question: Why does this not count towards initialization?

fme/core/generics/validation.py
```diff
@@ -0,0 +1,82 @@
...
+        compute_derived_variables: Whether to compute derived variables.
+
+    Returns:
+        Dictionary of validation metrics (keys prefixed by the label).
+    """
+    if record_logs is None:
+        record_logs = _get_record_to_wandb()
+
+    timer = GlobalTimer.get_instance()
+    timer.start("validation")
```
mcgibbon: Suggestion (optional): Use a `timer.context("validation"):` block to make it clearer what's timed and that the timer gets stopped appropriately.

fme/ace/inference/evaluator.py
mcgibbon: Maybe do this in a follow on PR to get this one merged without needing another re-review, but the timer.stop and restart is reflecting the code here being a bit spaghetti - we're doing initialization, then taking a pause to do validation, then resuming initialization. I see it's because some local symbols like `aggregator` are re-used, which is a symptom of too much logic going on here and suggests we could use some helper methods.

If you can refactor this in a way that keeps initialization together such that we could use one `timer.context("initialization"):` instead of these stop-starts, it would be a bit easier to understand. Or using two `timer.context("initialization"):` blocks (and maybe one `timer.context("validation"):` block) would make the timer arithmetic here a little easier to follow.

fme/ace/inference/evaluator.py
Arcomano1234: Wouldn't this allow you to have a validation `n_forward_steps` be a value different than what was used during training? I could see this being useful for weather forecasts where we want to know the day 3 skill or something.

fme/ace/inference/evaluator.py
```diff
@@ -81,6 +93,78 @@ def resolve_variable_metadata(
...
+        """
+        train_n = self.stepper_training.train_n_forward_steps
+        if train_n is not None:
+            if isinstance(train_n, int):
+                return train_n
+            assert isinstance(train_n, TimeLengthProbabilities)  # already validated
+            return train_n.max_n_forward_steps
+        if self.n_forward_steps is not None:
+            return self.n_forward_steps
+        return 1
```
Arcomano1234: Optional Suggestion/Question: Do we want to have it log in the case where either `n_forward_steps` nor `stepper_training.train_n_forward_steps`  are configured that its using `n_forward_steps` of 1 ?

fme/ace/inference/evaluator.py
jpdunc23: This should work in this PR now.

fme/ace/inference/evaluator.py
jpdunc23: > Wouldn't this allow you to have a validation n_forward_steps be a value different than what was used during training?

You should be able to do this regardless with the `ValidationConfig.stepper_training.n_forward_steps`, so I agree that `ValidationConfig.n_forward_steps` is redundant here. Will remove.

fme/ace/inference/evaluator.py
jpdunc23: OK, planning to handle via follow-on PRs.

PR #879, opened 2026-02-27
Downscaling dataset production script.

frodre:
To create the downscaling datasets, we do a bit of light processing to gather variables from various files and renaming of fields/coordinates.  This PR formalizes that process into a script `scripts/downscaling/process_from_raw_zarrs.py`.  I used this to produce the +4K datasets under `gs://vcm-ml-scratch/andrep/downscaling-xshield-amip-plus-4k` that now reside on weka under `/climate-default/2026-02-23-X-SHiELD-AMIP-plus-4K-downscaling`.

For now, all the configuration is within the script, but if we do this for other data, we should probably split that out into YAML-based configs or something.  The script allows for `--dry-run` mode to preview writes, processing any combination of the dataset resolutions (100km, 25km, 3km), it allows for partial variable appends, and complete overwrites, but errors on output existence if no special zarr write mode is specified.

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -140,6 +140,11 @@ class DiffusionModelConfig:
         n_in_channels = len(self.in_names)
         if static_inputs is not None:
             n_in_channels += len(static_inputs.fields)
+        elif self.use_fine_topography:
+            # Old checkpoints may not have static inputs serialized, but if
+            # use_fine_topography is True, we still need to account for the topography
+            # channel, which was the only static input at the time
+            n_in_channels += 1
 
         module = self.module.build(
             n_in_channels=n_in_channels,
--- a/fme/downscaling/test_models.py
+++ b/fme/downscaling/test_models.py
@@ -116,6 +116,48 @@ def test_module_serialization(tmp_path):
     )
 
 
+def test_from_state_backward_compat_fine_topography():
+    coarse_shape = (8, 16)
+    fine_shape = (16, 32)
+    downscale_factor = 2
+    static_inputs = StaticInputs(
+        fields=[
+            StaticInput(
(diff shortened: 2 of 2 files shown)


scripts/downscaling/process_from_raw_zarrs.py
```diff
@@ -0,0 +1,460 @@
...
+            "grid_yt": "latitude",
+        },
+    )
+
+    mid_res_config = DatasetPerVariableConfig(
+        name="25km",
+        base_path=MED_RES_BASE_PATH,
+        filename_map=RAW_FILENAMES,
+        per_file_variables=FILENAME_KEY_TO_VARIABLES,
+        rename_map={"grid_xt": "longitude", "grid_yt": "latitude"},
```
AnnaKwa, line 428: The data loading works with `grid_xt, grid_yt` which is what the current datasets use- do you know if the data loading concats datasets with different lat/lon coord names? If not could we stay with the existing grid_* names in case we want to use +0 and +4K datasets together?

frodre: My reading of the current input pipeline is that all of the horizontal coordinate handling is name agnostic--- only depending whatever the last two dimension names are for lat/lon coordinates.  I also didn't see any xarray logic for concat/merge where the dimension naming would be an issue.  I decided to adjust the names since I often rename for analysis purposes, but I'm OK transferring back to the old naming as well if it will be an issue.

frodre: The [horizontal coordinate creation](https://github.com/ai2cm/ace/blob/53ef316f76abcaeedd1c0e4d310ec2b3bf08fa48/fme/core/dataset/utils.py#L224) for reference.

PR #882, opened 2026-02-27
Add tests that Distributed.context is used in all main files using fme

mcgibbon:
Changes:
- Updated many entrypoints to call Distributed.context for proper Distributed cleanup
- Added tests that ensure files with main guards contain Distributed.context() call

- [ ] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/s2convolutions.py
+++ b/fme/core/models/conditional_sfno/s2convolutions.py
@@ -77,9 +77,7 @@ def _contract_dhconv(
     Returns:
         Complex output tensor of shape (batch_size, group, out_channels, nlat, nlon)
     """
-    wc = torch.view_as_complex(
-        weight.to(torch.float32)
-    )  # einsum does not support complexhalf
+    wc = torch.view_as_complex(weight)
     return torch.einsum("bgixy,giox->bgoxy", xc, wc)
 
 


scripts/test_distributed_context.py
mcgibbon, line 1: This is mostly duplicated from the test in `fme/`, but I felt like it's important that `fme` as a package have this test, and I also felt like it's important `fme` tests don't access `scripts`. The logic needs are also slightly different because it isn't a given that the scripts here use fme, most don't.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -77,7 +77,9 @@ def _contract_dhconv(
     Returns:
         Complex output tensor of shape (batch_size, group, out_channels, nlat, nlon)
     """
-    wc = torch.view_as_complex(weight)
+    wc = torch.view_as_complex(
```
Arcomano1234: Is this related to the PR?

mcgibbon: No, just a dirty change I had when starting. Reverted.

Arcomano1234 (approved):
LGTM, did a quick check to make sure there were not any other fme files that needed to be excluded what you have seems to be good. Also the tests should find everything.

PR #883, opened 2026-02-27
Configure which inference steps to log

mcgibbon:
This PR allows setting `log_step_means: list[StepMeanEntry]` on the InferenceEvaluatorAggregatorConfig, allowing different steps to be logged and allowing the user to chose the name for the log prefix.

This choice allows us to evaluate more than one of the timesteps, to keep backwards comparisons with existing runs, and to choose to name different steps with the same name when comparing models using different timesteps.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -316,15 +316,11 @@ class InferenceEvaluatorAggregator(
                 variable_metadata=dataset_info.variable_metadata,
             )
         for step_mean_entry in log_step_means:
+            step_mean_entry.validate(n_forward_steps)
             step = step_mean_entry.step
             name = step_mean_entry.get_name()
             # -1 because step 0 (after IC) is the first forward step
             target_time = step + n_ic_steps - 1
-            if step > n_forward_steps:
-                raise ValueError(
-                    f"Cannot log mean at step {step} since it is greater than the "
-                    f"number of forward steps {n_forward_steps}"
-                )
             self._aggregators[name] = OneStepMeanAggregator(
                 ops,
                 target_time=target_time,


fme/ace/aggregator/inference/main.py
```diff
@@ -126,6 +144,8 @@ class InferenceEvaluatorAggregatorConfig:
             time series metrics.
         monthly_reference_data: Path to monthly reference data to compare against.
         time_mean_reference_data: Path to reference time means to compare against.
+        timestep_hours: Timestep of the data in hours, used for determining
+            which timestep corresponds to 5 days of evolution.
```
oliverwm1: Should we be requiring the user to configure this, when it is really just a property of the dataset? I don't think there is anywhere else in the codebase where this is a user configurable option.

mcgibbon:
> Instead of having the user configure the dataset timestep—which is a property of the dataset—how about having the user configure the number of steps at which the RMSE is computed (which would default to 20, maintaining backwards compatibility)?

Hmmm yeah that's a good idea, I didn't like it originally because it's a lot more finnicky/error-prone to be configuring integer step counts like this than timestep lengths. But it would resolve the issue of aligning this change with ocean inference without lying. The only thing I'm still not sure about is the name, as I do want to compare 5-step inference against 20-step inference on their respective runs, and we have places where we explicitly look for the "mean_step_20" name in our code.

mcgibbon:
How about adding a `weather_eval_step: int = None` configuration option which if not given keeps the `mean_step_20` metrics, or if given replaces them with a `weather_step` set of metrics using the indicated forward step? This keeps the current default behavior, but lets me put new daily-step and 6h-step runs under the same metric.

jpdunc23:
> How about adding a `weather_eval_step: int = None` configuration option which if not given keeps the `mean_step_20` metrics, or if given replaces them with a `weather_step` set of metrics using the indicated forward step? This keeps the current default behavior, but lets me put new daily-step and 6h-step runs under the same metric.

If not too expensive, I think it would be nice to keep the `mean_step_20` metrics even when `weather_eval_step` is provided.

oliverwm1:
> as I do want to compare 5-step inference against 20-step inference on their respective runs

It is easy enough to make wandb panels comparing two metrics with different names, so I don't think this should be a blocker

mcgibbon:
> > as I do want to compare 5-step inference against 20-step inference on their respective runs
> 
> It is easy enough to make wandb panels comparing two metrics with different names, so I don't think this should be a blocker

Creating ~50 panels, one for each output, is a bit onerous and means I won’t look at this comparison in practice, except by flipping back and forth for a few key variables.

mcgibbon:
> If not too expensive, I think it would be nice to keep the `mean_step_20` metrics even when `weather_eval_step` is provided.

What's the use case for this? Aside from the expense, I'd like to avoid doubling the number of wandb keys I need to flip through related to these metrics. Depending on the use case, it also may make more sense to allow `weather_eval_steps: list[int]` instead.

jpdunc23:
> > If not too expensive, I think it would be nice to keep the `mean_step_20` metrics even when `weather_eval_step` is provided.
> 
> What's the use case for this?

For the ocean, I want to keep `step_20` (100-day) metrics for comparability with older runs, but would be nice to also have metrics for shorter horizons.

> Aside from the expense, I'd like to avoid doubling the number of wandb keys I need to flip through related to these metrics.

Agreed that it can be a lot, but with the corresponding channel-mean metrics you can quickly get a sense of where things stand in terms of weather / climate, and then decide what to dig into. 

> Depending on the use case, it also may make more sense to allow `weather_eval_steps: list[int]` instead.

I think this would be great! But even so, I hope we can keep including the `step_20` metrics for backwards comparability even when `weather_eval_steps` is configured.

jpdunc23:
> > > as I do want to compare 5-step inference against 20-step inference on their respective runs
> > 
> > 
> > It is easy enough to make wandb panels comparing two metrics with different names, so I don't think this should be a blocker
> 
> Creating ~50 panels, one for each output, is a bit onerous and means I won’t look at this comparison in practice, except by flipping back and forth for a few key variables.

I agree with Jeremy here, this can be a real pain. Would be nice if wandb had some tooling to make this less onerous but I'm not aware of any.

Also, if the log key ends up being generic like `weather_eval_step` then for some runs `step_20` and `weather_eval_step` metrics will be comparable but for others it won't.

fme/ace/aggregator/inference/main.py
jpdunc23: `fme.coupled` requires configuration of the ocean / atmosphere timestep as a string like "5D" / "6h".

fme/ace/aggregator/one_step/reduced.py
```diff
@@ -36,12 +36,14 @@ def __init__(
         include_grad_mag_percent_diff: bool = True,
         target: Literal["norm", "denorm"] = "denorm",
         channel_mean_names: Sequence[str] | None = None,
+        log_loss: bool = True,
```
mcgibbon, line 39: This change avoids needing to pop NaN keys with unpredictable names.

fme/ace/aggregator/inference/main.py
```diff
@@ -273,21 +301,27 @@ def __init__(
...
-                include_bias=False,
-                include_grad_mag_percent_diff=False,
-                channel_mean_names=self._channel_mean_names,
-            )
+        for step_mean_entry in log_step_means:
+            step = step_mean_entry.step
+            name = step_mean_entry.get_name()
+            # -1 because step 0 (after IC) is the first forward step
+            target_time = step + n_ic_steps - 1
+            if step < n_forward_steps:
```
jpdunc23: Should we add an error or warning when the configured step is > n_forward_steps?

fme/coupled/aggregator.py
```diff
@@ -368,12 +369,20 @@ def __init__(
         ocean_channel_mean_names: Sequence[str] | None = None,
         atmosphere_channel_mean_names: Sequence[str] | None = None,
     ):
-        self._record_ocean_step_20 = n_timesteps_ocean >= 20
-        self._record_atmos_step_20 = n_timesteps_atmosphere >= 20
+        if n_timesteps_ocean >= 20:
+            ocean_day_5 = 20
```
jpdunc23: I think the variable names `ocean_day_5` and `atmosphere_day_5` are a bit misleading here and could cause some confusion. It's also somewhat inconsistent, e.g. in the "default" case of 5-day ocean steps, step 20 is 100 days.

fme/ace/aggregator/inference/main.py
```diff
@@ -138,15 +157,18 @@ class InferenceEvaluatorAggregatorConfig:
     monthly_reference_data: str | None = None
     time_mean_reference_data: str | None = None
     log_nino34_index: bool = True
+    log_step_means: list[StepMeanEntry] = dataclasses.field(
+        default_factory=lambda: [StepMeanEntry(step=20)]
+    )
```
jpdunc23, line 176: This PR seems like a natural place to also make this configurable on the coupled `InferenceEvaluatorAggregatorConfig` in fme/coupled/aggregator.py. 

Can you make it separately configurable for the atmosphere / ocean, e.g. via a `CoupledStepMeanEntries` dataclass?

fme/ace/aggregator/inference/main.py
```diff
@@ -273,21 +301,27 @@ def __init__(
...
-                target="norm",
-                include_bias=False,
-                include_grad_mag_percent_diff=False,
-                channel_mean_names=self._channel_mean_names,
-            )
+        for step_mean_entry in log_step_means:
+            step = step_mean_entry.step
+            name = step_mean_entry.get_name()
+            # -1 because step 0 (after IC) is the first forward step
+            target_time = step + n_ic_steps - 1
```
jpdunc23, line 323: From agent review:

> StepMeanEntry docstring says "Step at which to log" but doesn't clarify that this is a forward-step index (0-based from after IC). The `target_time = step + n_ic_steps - 1` computation in the constructor implies step 0 is the first forward step, which could be confusing. A brief note in the docstring would help.

I'm now confused about the configuration. If I want to log the metrics for the very first forward step, should I use `StepMeanEntry.step = 1` or `StepMeanEntry.step = 0`? I think it's the former (`step = 1`) which I definitely prefer because we always think of `step = 0` as the IC, do you agree? 

I think this would be less confusing with improved documentation of the `fme.ace.aggregator.one_step.reduced.MeanAggregator` args, e.g. what is the meaning of `target_time` (?) and does `record_batch` expect data with or without initial conditions? These are preexisting issues, but since you're thinking about it now might be a good time to begin documenting assumptions.

jpdunc23 (commented):
Looking good. I think it would be good to add some additional test coverage for the new feature (non-default "name", `len(log_step_means) > 1`, etc).

The step labeling vs step indexing is a bit confusing and I've come away from the review unsure of how I should configure `StepMeanEntry.step` to get the desired forward step.

I also think it would be good to make this configurable in `fme.coupled` in this PR, but I won't block if you prefer not to do it here. If so and you aren't planning to immediately create a new PR to do it, then could you open an issue so we can remember?

fme/ace/aggregator/inference/main.py
mcgibbon: Can we separate this into another PR? It's cleanly separable, and while I want to get this in while it's fresh, I have two higher priority tasks right now (data processing and spatial distributed). The code changes may be easy but updating the tests will take some time, and config changes usually need multiple rounds of review (see this PR). I'm also happy to pass it on to you if you have the time.

fme/ace/aggregator/inference/main.py
mcgibbon: Huh, I wonder why this ignore wasn't necessary before. I'll see if I can remove it.

fme/ace/aggregator/inference/main.py
mcgibbon: Updated.

fme/ace/aggregator/inference/main.py
mcgibbon: OK, I looked into it and this is necessary to maintain existing behavior. Currently this isn't a configurable option at all, the aggregator just records step 20 if it's there and not if it's not, passing it through as a `.build` argument.

Instead of adding a warn or error here, I'm going to add an exception in the inference config post inits, where we have access to n_forward_steps.

For what it's worth this will rarely if ever come up, but it would be nice to know if we've mis-configured a 5-day run before submission time.

mcgibbon: OK updated. While this did need to be copy-pasted to a few configs, this version made it much easier to update tests and lets configs fail at validation which I think is important. I also updated here to raise on init if it's configured improperly, as a failsafe (and to be more informative for unit tests like of the diagnostics netcdf files, that don't use those configs).

fme/coupled/aggregator.py
mcgibbon: I've reverted to something much more like what was originally here. I passed the `name=` arg for completeness since it's used explicitly below to pop the channel means.

mcgibbon:
> I think it would be good to add some additional test coverage for the new feature (non-default "name", len(log_step_means) > 1, etc).

Updated the evaluator log name test to include both of these conditions.

PR #884, opened 2026-02-27
Turn off rolling on zonal mean images

### LINKED ISSUE #881: Minor zonal mean plotting bug
Zonal mean images appear to have a few pixels rolled horizontally, which you can see in the left edge of the attached image. Haven't checked but I think this due to changes in https://github.com/ai2cm/ace/commit/5b5676f0b92b53d554a64c8316e43b1118f2f4c8

Image

Arcomano1234:
Small PR to turn off the `roll_lon` for the zonal_mean images. 

Changes:
- Sets` roll_lon=False ` for zonal mean gen and error plots
- Updates regression tests for inference aggregators

Resolves #881


fme/ace/aggregator/inference/testdata/test_evaluator-test_logs_labels_exist-time_mean-bias_map-a.png
jpdunc23, line 1: Why does this file get updated? I thought only the zonal mean files should be affected by this change.

Arcomano1234: Yeah that is a great question visually they look similar but might be slightly different. Just to confirm to recreate these all I was supposed to do is delete [test_evaluator-test_logs_labels_exist-regression.pt‎](https://github.com/ai2cm/ace/pull/884/changes#diff-c2010b885200eba03c448b7426b477561f5d6eae4ea2c3be373a74105fa8e0c9) right?

mcgibbon: I can confirm. The real regression file is the `.pt`, while the `.png`'s get auto-generated when the `.pt` is updated. Since one of the `.pt` outputs got updated, all outputs needed to get re-generated. To review, confirm that only the expected plots changed meaningfully, which tells us that the `.pt` data also did not change meaningfully for targets we don't expect to change.

PR #887, opened 2026-03-02
Speed up and reduce memory usage of spectral LoRA

mcgibbon:
This PR significantly speeds up the spectral convolution LoRA by forcing the einsum to take a more efficient pathway.

Changes:
- No public API changes.

- [x] Tests added


mcgibbon:
If desired, I can delete the test and just update the existing LoRA code in-place, but I needed the test for testing and it makes this easier to review.

fme/core/models/conditional_sfno/test_s2convolutions.py
```diff
@@ -90,6 +90,50 @@ def contract_grouped():
...
 
 
+@pytest.mark.skipif(
+    get_device().type != "cuda",
+    reason=(
+        "This test is only relevant for CUDA since "
+        "it's testing speed of DHConv groups on GPU."
+    ),
+)  # noqa: E501
+def test_lora_lowmem_is_faster():
```
Arcomano1234: I assume `_contract_lora_lowmem` and `_contract_lora` are bitwise identical right and we don't need to test that here right?

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -340,7 +365,7 @@ def forward(self, x, timer: Timer = NullTimer()):  # pragma: no cover
 
         if self.lora_A is not None and self.lora_B is not None:
             with timer.child("lora_update"):
-                lora_update = _contract_lora(
```
Arcomano1234, line 343: Question: Do we eventually want to remove the original `_contract_lora` method?

Arcomano1234 (approved):
Left 2 questions but approved.

fme/core/models/conditional_sfno/test_s2convolutions.py
mcgibbon: I doubt they're bitwise identical, but they should be allclose identical, yes. I don't think we need to test it.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: I can do it in this PR if you like (and remove the test). It's mostly the state it's in now so the performance improvement can be reviewed.

Actually, I'll do it now since you've already reviewed, and you can ask me to revert it if you prefer.

PR #890, opened 2026-03-03
add Distributed.reduce_spatial_gradients to accumulate grads

mahf708:
add Distributed.reduce_spatial_gradients to accumulate grads, in anticipation of use with spatial parallelism training.

- [x] Tests added

### DIFF
--- a/fme/core/distributed/base.py
+++ b/fme/core/distributed/base.py
@@ -106,6 +106,10 @@ class DistributedBackend(ABC):
         """
         ...
 
+    def reduce_spatial_gradients(self, module: torch.nn.Module) -> None:
+        """All-reduce parameter gradients across spatial groups. No-op default."""
+        ...
+
     @abstractmethod
     def barrier(self): ...
 
--- a/fme/core/distributed/model_torch_distributed.py
+++ b/fme/core/distributed/model_torch_distributed.py
@@ -244,9 +244,10 @@ class ModelTorchDistributed(DistributedBackend):
     def wrap_module(self, module: torch.nn.Module) -> torch.nn.Module:
         """Wrap with DDP over the **data** process group.
 
-        For now, we assume spatial communication is expected to be handled
-        inside the model layers themselves. If we need to change course, we
-        can revisit...
+        DDP handles gradient all-reduce across data-parallel ranks only.
+        Spatial gradient synchronization across h/w ranks is handled
+        separately by reduce_spatial_gradients below, which must be called
+        after backward and before the optimizer step.
         """
         if any(p.requires_grad for p in module.parameters()):
             if using_gpu():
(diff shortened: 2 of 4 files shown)


fme/core/distributed/model_torch_distributed.py
```diff
@@ -261,6 +262,23 @@ def wrap_module(self, module: torch.nn.Module) -> torch.nn.Module:
...
         return DummyWrapper(module)
 
+    def reduce_spatial_gradients(self, module: torch.nn.Module) -> None:
+        """All-reduce parameter gradients across spatial (h, w) process groups.
+
+        After loss.backward(), each spatial rank holds partial parameter
+        gradients computed only from its local spatial slab. This method
+        sums those partial gradients across all spatial ranks so that every
+        rank has the full gradient before the optimizer step.
+        """
```
mcgibbon, line 272: Is this true? Gradient reduction doesn't use our own `dist` reductions, it does something internally in pytorch's objects. I think it's the DistributedDataParallel wrapper we apply over the module, which automatically reduces gradients across all ranks at the time when they're getting applied.

But past that, not all gradients should be reduced. For example, the positional embedding should not get spatially reduced. I believe makani and modulus have logic for this.

Can you explain to me the full plan for implementing parameter optimizations before we implement it? It seems that simply making a `.forward` pass work is a blocker before we can implement the backward pass, should we focus on that before PR'ing these changes?

mcgibbon: Past that, it seems like `dist` should be responsible for a more general `reduce_spatial(self, tensor)` method, and that the optimizer, module wrapper, or other code should handle the use of it for parameter updates. So these changes don't go stale, you could in theory refactor this into that method. Though I'm still not sure what APIs we will want for these reductions, e.g. a `reduce(tensor, tensor_dims: list[Dim], output_dims: list[Dim])` method might better serve us (though I suppose `reduce_spatial` could become a shortcut that calls such a reduce).

We very well might want a DistributedDataAndModelParallel wrapper instead of a loss reduction.

mahf708: Maybe you're right, but and we can always change course from loss/grad hooks to a redone wrapper. 

> But past that, not all gradients should be reduced. For example, the positional embedding should not get spatially reduced. I believe makani and modulus have logic for this.

Yeah, that will be addressed near where the reduction is happening (which is not added in this PR on purpose)

> Can you explain to me the full plan for implementing parameter optimizations before we implement it?

The preliminary plan is to keep everything in place as-is, identify ones that will require spatial reduction and address them via hooks. But of course the plan can change as needs become clearer later on.


Is it possible to review the PR in abstract/isolation or is that too difficult for you because the immediate need isn't clear?

fme/core/distributed/parallel_tests/test_reduce_spatial_gradients.py
```diff
@@ -0,0 +1,162 @@
...
+"""
+
+import pytest
+import torch
+
+from fme.core import get_device
+from fme.core.distributed import Distributed
+
+
+def _simple_module(in_features: int = 4, out_features: int = 4) -> torch.nn.Module:
```
mcgibbon, line 12: Not needed right now, but note you will need tests for a module whose weights contain spatial dimensions, in the future.

fme/core/distributed/base.py
```diff
@@ -106,6 +106,10 @@ def wrap_module(self, module: torch.nn.Module) -> torch.nn.Module:
         """
         ...
 
+    def reduce_spatial_gradients(self, module: torch.nn.Module) -> None:
```
mcgibbon, line 109: Issue: This is a concrete method without an implementation. Either make it abstract and implement it on all inheritors (more complete, and the implementations will be fairly simple), or delete it from the ABC and the corresponding method from Distributed and leave it implemented and tested on only your backend for now (easier/faster to get to merge).

As it is, if the method on Distributed ever were to get called, it would break (with a Nonetype-related error) on any non-model-parallel backend. Even if we're adding methods that aren't getting called for now, I'd prefer to avoid adding methods we know are broken - it's hard to develop on top of code when you can't assume what's pre-existing is working.

mcgibbon:
It looks like the GPU tests are failing unrelated to your changes, I'll look into it.

mcgibbon:
I suspect the failure is a noisy failure since we started running unit tests in parallel, and the GPU just happened to be more occupied when running the grouped version. If the test ends up blocking merge, I will override it. I'll bring the issue up with the team.

mahf708:
Let me take a look as well into the fails. Also, let me try to get some more on the forward part for us to be clearer about what's needed so that you're [plural] in a better position to decide if we want to push/integrate things. Converting this to draft for now while I try to get other small PRs up soon...

mahf708:
Close to prioritize the forward stepping for now, as there might be a better way to handle this later

PR #892, opened 2026-03-03
Port latest updates to fft from torch-harmonics

mcgibbon:
We overwrite the implementation of RealSHT and InverseRealSHT in torch-harmonics, meaning we don't get their updates. Since our original port, they have added automatic handling of zeroing of the imaginary part of the 0 harmonic, which is important for certain GPU hardware that selects an optimized backend sensitive to this change.

These changes are covered by existing regression tests.

Changes:
- No public API changes
- Explicitly zero the imaginary part of the zeroth-order harmonics when performing `irfft`

- [ ] Tests added


fme/fft.py
mcgibbon, line 1: This file is ported without code changes (just one spelling fix in a docstring that was bothering me).

PR #896, opened 2026-03-03
Make test_xarray_loader parallelism-aware

### LINKED ISSUE #914: || test: get_local_slices test is now polluted
Issue: this is no longer a unit test of `get_local_slices` because it also will be broken by errors in `spatial_reduce_sum` and `reduce_sum`.

Suggestion: Copy-paste the original test and make this a new test (possibly in a different file).

_Originally posted by @mcgibbon in https://github.com/ai2cm/ace/pull/900#discussion_r2892674705_

mcgibbon:
Adds a regression test in test_xarray_loader that ensures the loaded data matches the checkpoint value, regardless of parallel decomposition.

Changes:
- Added `Distributed.scatter_object`
- Fixed type hints on `Distributed.gather_object`
- Fixed issue where comparing tensor dicts in fme.core.testing.regression would not show metrics on how badly the assert_close failed during failure, clobbered by msg arg

- [x] Tests added

Resolves #914

### DIFF
--- a/fme/ace/data_loading/gridded_data.py
+++ b/fme/ace/data_loading/gridded_data.py
@@ -182,7 +182,6 @@ class InferenceGriddedData(InferenceDataABC[PrognosticState, BatchData]):
             )
         else:
             self._initial_condition = initial_condition.to_device()
-        self._initial_time: xr.DataArray | None = None
 
     @property
     def loader(self) -> DataLoader[BatchData]:
@@ -243,13 +242,7 @@ class InferenceGriddedData(InferenceDataABC[PrognosticState, BatchData]):
 
     @property
     def initial_time(self) -> xr.DataArray:
-        if self._initial_time is None:
-            for batch in self.loader:
-                self._initial_time = batch.time.isel(time=0)
-                break
-            else:
-                raise ValueError("No data found in loader")
-        return self._initial_time
+        return self.initial_condition.as_batch_data().time.isel(time=0)
 
 
 class PSType:
--- a/fme/ace/inference/data_writer/file_writer.py
+++ b/fme/ace/inference/data_writer/file_writer.py
@@ -5,7 +5,9 @@ import os
 from collections.abc import Mapping, Sequence
 from typing import TypeAlias, TypeGuard, Union
 
+import cftime
 import numpy as np
+import numpy.typing as npt
 import torch
 import xarray as xr
 
@@ -270,7 +272,7 @@ class FileWriterConfig:
     def build_paired(
         self,
         experiment_dir: str,
-        n_initial_conditions: int,
+        initial_condition_times: npt.NDArray[cftime.datetime],
         n_timesteps: int,
         timestep: datetime.timedelta,
         variable_metadata: Mapping[str, VariableMetadata],
@@ -284,7 +286,7 @@ class FileWriterConfig:
             prediction_label = f"{self.label}_{prediction_suffix}"
             reference_writer = dataclasses.replace(self, label=reference_label).build(
                 experiment_dir=experiment_dir,
-                n_initial_conditions=n_initial_conditions,
+                initial_condition_times=initial_condition_times,
                 n_timesteps=n_timesteps,
                 timestep=timestep,
                 variable_metadata=variable_metadata,
@@ -296,7 +298,7 @@ class FileWriterConfig:
             reference_writer = None
         prediction_writer = dataclasses.replace(self, label=prediction_label).build(
             experiment_dir=experiment_dir,
-            n_initial_conditions=n_initial_conditions,
+            initial_condition_times=initial_condition_times,
             n_timesteps=n_timesteps,
             timestep=timestep,
             variable_metadata=variable_metadata,
@@ -310,7 +312,7 @@ class FileWriterConfig:
     def build(
         self,
         experiment_dir: str,
-        n_initial_conditions: int,
+        initial_condition_times: npt.NDArray[cftime.datetime],
         n_timesteps: int,
         timestep: datetime.timedelta,
         variable_metadata: Mapping[str, VariableMetadata],
@@ -322,7 +324,8 @@ class FileWriterConfig:
 
         Args:
             experiment_dir: The directory where experiment outputs are saved.
-            n_initial_conditions: The number of initial conditions or ensemble members.
+            initial_condition_times: 1D array of initial condition times
+                (start time for each inference run).
             n_timesteps: Total number of inference forward steps.
             timestep: The time delta between each timestep.
             variable_metadata: Metadata for each variable.
@@ -334,6 +337,8 @@ class FileWriterConfig:
             spatial_dims = DIM_INFO_HEALPIX
         else:
             spatial_dims = DIM_INFO_LATLON
+
+        n_initial_conditions = len(initial_condition_times)
 
         if (self.lat_extent and LAT_NAME not in coords) or (
             self.lon_extent and LON_NAME not in coords
@@ -372,8 +377,10 @@ class FileWriterConfig:
         if isinstance(self.format, ZarrWriterConfig):
             if isinstance(self.time_coarsen, TimeCoarsenConfig):
                 n_timesteps_write = n_timesteps // self.time_coarsen.coarsen_factor
+                timestep_write = self.time_coarsen.coarsen_factor * timestep
             else:
                 n_timesteps_write = n_timesteps
+                timestep_write = timestep
 
             zarr_writer_cls: type[SeparateICZarrWriterAdapter | ZarrWriterAdapter]
 
@@ -387,8 +394,9 @@ class FileWriterConfig:
                 path=os.path.join(experiment_dir, f"{self.label}.zarr"),
                 dims=dims,
                 data_coords=ensure_numpy_coords(subselect_coords_),
+                timestep=timestep_write,
                 n_timesteps=n_timesteps_write,
-                n_initial_conditions=n_initial_conditions,
+                initial_condition_times=initial_condition_times,
                 data_vars=self.names,
                 variable_metadata=variable_metadata,
                 dataset_metadata=dataset_metadata,
@@ -415,7 +423,7 @@ class FileWriterConfig:
                 raw_writer = RawDataWriter(
                     path=experiment_dir,
                     label=self.label,
-                    n_initial_conditions=n_initial_conditions,
+                    initial_condition_times=initial_condition_times,
                     save_names=self.names,
                     variable_metadata=variable_metadata,
                     coords=subselect_coords_,
--- a/fme/ace/inference/data_writer/main.py
+++ b/fme/ace/inference/data_writer/main.py
@@ -5,7 +5,9 @@ import warnings
 from collections.abc import Mapping, Sequence
 from typing import TypeAlias
 
+import cftime
 import numpy as np
+import numpy.typing as npt
 import torch
 import xarray as xr
 
@@ -76,7 +78,7 @@ class DataWriterConfig:
     def build_paired(
         self,
         experiment_dir: str,
-        n_initial_conditions: int,
+        initial_condition_times: npt.NDArray[cftime.datetime],
         n_timesteps: int,
         timestep: datetime.timedelta,
         variable_metadata: Mapping[str, VariableMetadata],
@@ -85,7 +87,7 @@ class DataWriterConfig:
     ) -> "PairedDataWriter":
         return PairedDataWriter(
             path=experiment_dir,
-            n_initial_conditions=n_initial_conditions,
+            initial_condition_times=initial_condition_times,
             n_timesteps=n_timesteps,
             timestep=timestep,
             variable_metadata=variable_metadata,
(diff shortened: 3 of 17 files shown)


mcgibbon:
Pinging @mahf708 for visibility. I should add some tests for the new `scatter_object` before we merge.

Note I wrote the test for `gather_object` incorrectly originally, it should be performing a global gather rather than a data parallel gather because the purpose is for scatter/gather from/to a global root that interfaces with the filesystem. I don't really like that the name isn't consistent with how other bare names do data-parallel operations, but I think maybe it is what it is until we figure out a better way. I could rename them to e.g. `gather_global_object` perhaps.

fme/ace/data_loading/test_data_loader.py
```diff
@@ -190,21 +198,57 @@ def test_ensemble_loader_n_samples(tmp_path, num_ensemble_members=3, n_samples=1
...
 
+@pytest.mark.parallel
 def test_xarray_loader(tmp_path):
     """Checks that vertical coordinates are present."""
-    _create_dataset_on_disk(tmp_path)
+    dist = Distributed.get_instance()
+    tmp_path = dist.scatter_object(tmp_path)  # get the root value
+    global_batch_size = 8
+    if dist.is_root():
+        # TODO: investigate why this test fails if we
```
mcgibbon, line 213: This is concerning to me, but I may not have the time to investigate it.

fme/ace/data_loading/test_data_loader.py
```diff
@@ -63,7 +67,7 @@ def _save_netcdf(
     filename, dim_sizes, variable_names, calendar, timestep_size=1, timestep_start=0
 ):
     data_vars = {}
-    for name in variable_names:
+    for name in sorted(variable_names):
```
elynnwu, line 71: [nit] why do you need sorted here?

elynnwu: I see the explanation below now

fme/core/testing/regression.py
```diff
@@ -38,7 +39,6 @@ def _assert_close(x: NestedTensorDict, y: NestedTensorDict, **assert_close_kwarg
             torch.testing.assert_close(
                 v,
                 y_val.to(v.device),
-                msg=f"Mismatch at key {k}",
```
elynnwu, line 41: [nit] can we keep this error message?

fme/ace/data_loading/test_data_loader.py
elynnwu: If I understand correctly, I think you're saying this fails for `n_times=global_batch_size + 2` right? I was able to get the test to fail that way and I asked AI to investigate, and this is what I got:
```
The test gathers the batch and then sorts by x[0, 0, 0].item() for the regression. If two samples have the same (or numerically equal) value at [0,0,0], sorted() can order them differently across runs or environments. The regression file was almost certainly generated with n_times = 7 (exactly 6 samples). With n_times = 8 you still feed 6 samples (0–5), but the RNG state and the extra time step might change the pattern of values so that a tie at [0,0,0] appears (or disappears). Then:
One run produces order A → matches the saved regression.
Another run (or the “+2” case) produces order B → different ordering of the same 6 samples → validate_tensor_dict fails.
```

elynnwu:
> I should add some tests for the new scatter_object before we merge

I agree, we should do that

fme/ace/data_loading/test_data_loader.py
mcgibbon: I don't particularly believe the AI - the ordering can't be wrong because of the sort.

What I would expect is that with one extra item, there are 9 potential items to choose from to put in the first batch, and perhaps one of them will change between the two configurations due to RNG. However, what I find is that the 4 samples coming from the second rank are all ones that weren't seen, which I didn't think would be possible.

Also, I intentionally changed the loading to use a non-train mode, so that shuffling shouldn't happen.

fme/core/testing/regression.py
mcgibbon: If we set this msg argument, I don't get to see the default message which describes exactly how badly it failed by rtol and atol. It literally just gives this message. It's easy for me to `--pdb` to check what key it's failing on, but not easy to use it to type in this torch testing command and manually get the atol/rtol failure.

.github/workflows/tests.yaml
```diff
@@ -34,29 +34,7 @@ jobs:
...
-          NPROC=3  FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
-          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
-          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
-          NPROC=2  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=2 make test_parallel
-          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=3 FME_DISTRIBUTED_W=1 make test_parallel
-          NPROC=3  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=3 make test_parallel
-          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=2 make test_parallel
-          NPROC=4  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 make test_parallel
-          NPROC=8  FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=2 make test_parallel
+          make cpu_test_all_parallel
```
mcgibbon, line 37: This made it much easier for me to launch local tests.

fme/core/distributed/parallel_tests/test_backend.py
```diff
@@ -181,35 +181,6 @@ def test_local_batch_size_divisibility():
     assert local_bs * n_dp == global_bs
 
 
-@pytest.mark.parallel
```
mcgibbon, line 185: These were moved to test_local_slices.py

fme/core/distributed/parallel_tests/test_local_slices.py
```diff
@@ -32,6 +33,39 @@ def test_gather_tensor_from_local_slices():
         assert gathered is None
 
 
+@pytest.mark.parallel
+def test_local_slices_cover_full_domain():
+    """Union of slices from every rank should cover every element exactly once."""
+    dist = Distributed.get_instance()
+    if isinstance(dist._distributed, ModelTorchDistributed):
```
mcgibbon: There was a chain of events (my fault) that led to this not being implemented correctly.

`gather_object` is supposed to gather/scatter globally (to/from global root), not just along the data parallel group. This wasn't clear enough when it got implemented. This meant the test of full domain coverage was passing despite local_slices not slicing along spatial axes as intended (and I hadn't written a docstring for local_slices to clarify this behavior).

`local_slices` is meant to give the slice of the local rank within a full global domain (both spatial and data-parallel). It will need to be updated and these tests re-enabled.

Makefile
```diff
@@ -56,24 +57,49 @@ create_environment:
 	conda run --no-capture-output -n $(ENVIRONMENT_NAME) uv pip install -r analysis-deps.txt
 
 test:
-	pytest -n 4 --durations 40 .
+	pytest -n 4 --durations 40 $(TEST_PATH)
```
mcgibbon, line 60: This let me run e.g.

```
 TEST_PATH="fme/core/distributed/parallel_tests/test_get_sampler.py::test_get_sampler_covers_all_indices" make cpu_test_all_parallel
```

Makefile
```diff
@@ -56,24 +57,49 @@ create_environment:
...
+# | model   |  3 | 1 | 1 |
+# | model   |  2 | 2 | 1 |
+# | model   |  2 | 1 | 2 |
+# | model   |  3 | 3 | 1 |
+# | model   |  3 | 1 | 3 |
+# | model   |  4 | 2 | 2 |
+# | model   |  4 | 2 | 1 |
+# | model   |  8 | 2 | 2 |
+cpu_test_all_parallel:
+	FME_FORCE_CPU=1 NPROC=2 FME_DISTRIBUTED_BACKEND=torch FME_DISTRIBUTED_H=1 FME_DISTRIBUTED_W=1 make test_parallel
```
mcgibbon, line 84: Same as what was in the git action, except FME_FORCE_CPU=1 makes this work on my GPU VM.

fme/core/testing/regression.py
elynnwu: Sounds good, let's remove this msg then.

fme/core/distributed/test_distributed.py
```diff
@@ -81,6 +81,17 @@ def test_distributed_gather():
     mp.spawn(run_gather_test, args=(world_size,), nprocs=world_size, join=True)
 
 
+@pytest.mark.parallel
+def test_scatter_object():
```
elynnwu, line 85: [nit] This does not get tested for the non-distributed backend right? It's trivial so up to you if you want to add an explicit non-parallel test.

elynnwu (approved):

fme/core/distributed/test_distributed.py
mcgibbon: Yeah we don't need one. Similarly to how we won't have an explicit parallel backend one.

mahf708:
@mcgibbon + @elynnwu, I took a brief look at this. I think we should try to redo this in steps:

First, Merge #900 with the stuff it brings, including a get_local_slices impl that work both on the batch and spatial dims (see below) and some sketchy gather/scatter ops.

<details><summary>new get_local_slices</summary>
<p>

```py
    def get_local_slices(self, tensor_shape, data_parallel_dim: int | None = None):
        """Return index slices for this rank's local chunk.

        Slices the ``data_parallel_dim`` across data-parallel ranks and
        the last two dimensions across spatial (h, w) model-parallel ranks.
        """
        return_list = [slice(None, None) for _ in tensor_shape]
        if data_parallel_dim is not None:
            n_dp = self.total_data_parallel_ranks
            if tensor_shape[data_parallel_dim] % n_dp != 0:
                raise ValueError(
                    f"expected global data parallel dim to be "
                    f"divisible by data parallel ranks, got "
                    f"global shape {tensor_shape} with "
                    f"{n_dp} data parallel ranks"
                )
            per_rank = tensor_shape[data_parallel_dim] // n_dp
            return_list[data_parallel_dim] = slice(
                self._data_rank * per_rank, (self._data_rank + 1) * per_rank
            )
        # Spatial slicing on the last two dimensions (H, W).
        if len(tensor_shape) >= 2:
            return_list[-2], return_list[-1] = self._get_local_spatial_shape(
                tensor_shape[-2], tensor_shape[-1]
            )
        return tuple(return_list)
```

</p>
</details> 

We should then aim to coalesce around how we should impl get_local_slices and the gather/scatter methods. I agree with Jeremy that the impl in the new PR is not good.

I will try to draft something that will make test_xarray_loader ||ism-aware on top of the other PR. The changes here don't seem to be significant so far, so I can probably start over. I will definitely need help with the regression part of that test, so I won't touch that.

mcgibbon:
> I will try to draft something that will make test_xarray_loader ||ism-aware on top of the other PR. The changes here don't seem to be significant so far, so I can probably start over. 

What would the motivation be for starting over? After merging #900 into main and resolving two small merge conflicts in test files, it looks like the tests on this PR are mostly passing. I was able to remove the xfail marker from every test except the data loading (which I intended to keep marked xfail for this PR).

fme/core/distributed/parallel_tests/test_backend.py
```diff
@@ -182,37 +182,6 @@ def test_local_batch_size_divisibility():
...
-    rows = 4 * n_dp
-    global_shape = (rows, 6, 8)
-    local_slices = dist.get_local_slices(global_shape, data_parallel_dim=0)
-    canvas = torch.zeros(global_shape, device=get_device())
-    canvas[local_slices] += 1.0
-    # Combine spatial contributions then data-parallel contributions
-    canvas = dist.spatial_reduce_sum(canvas)
-    canvas = dist.reduce_sum(canvas)
-    # The entire domain should be covered exactly once
-    torch.testing.assert_close(canvas, torch.ones_like(canvas))
```
mcgibbon, line 199: Reverted to the original version of this test, resolving #914

fme/core/distributed/parallel_tests/test_backend.py
```diff
@@ -182,37 +182,6 @@ def test_local_batch_size_divisibility():
...
-    """Without a dp dim, the data-parallel dims are untouched."""
-    dist = Distributed.get_instance()
-    shape = (8, 6, 8)  # (batch, H, W) — 3D so batch is separate from spatial
-    slices = dist.get_local_slices(shape)
-    # batch dim should be full (no dp dim specified)
-    assert slices[0] == slice(None, None)
-    # spatial dims should produce a valid sub-tensor
-    local = torch.zeros(shape, device=get_device())[slices]
-    assert local.shape[0] == shape[0]  # batch untouched
-    assert local.numel() > 0  # local portion is non-empty
```
mcgibbon, line 213: Removed the last part of the test which is just testing how slices work (kept up to the assert slices[0] == slice(None, None)), and moved to `test_local_slices.py`

PR #898, opened 2026-03-04
Add SamudrACE CM4 piControl baseline configs

jpdunc23:
Adds scripts and configs for baseline training of SamudrACE on the CM4 piControl dataset.

### DIFF
--- a/fme/coupled/inference/evaluator.py
+++ b/fme/coupled/inference/evaluator.py
@@ -157,6 +157,11 @@ def load_stepper(
     return load_coupled_stepper(checkpoint_path)
 
 
+def _validate_coupled_steps_config(n_coupled_steps: int, coupled_steps_in_memory: int):
+    if n_coupled_steps % coupled_steps_in_memory:
+        raise ValueError("n_coupled_steps must be divisible by coupled_steps_in_memory")
+
+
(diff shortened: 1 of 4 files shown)


oliverwm1:
Took a brief look—I think it generally looks good and appreciate the readme. A couple thoughts:
- how do these scripts related to the existing `cm4-ocean`, `cm4-atmosphere` and `cm4/coupled-inference-no-target` directories in `ace/configs/baselines`? It is important to avoid duplication and have a single source of truth for configs, so I suspect that maybe all of those directories should be deleted in favor of this new one?
- why the nested `coupled` folder? Could just have `configs/baselines/cm4-piControl` right?

elynnwu:
> - how do these scripts related to the existing cm4-ocean, cm4-atmosphere and cm4/coupled-inference-no-target directories in ace/configs/baselines? It is important to avoid duplication and have a single source of truth for configs, so I suspect that maybe all of those directories should be deleted in favor of this new one?

Deleted `cm4-ocean, cm4-atmosphere, cm4/coupled-inference-no-target`
> - why the nested coupled folder? Could just have configs/baselines/cm4-piControl right?

Updated to `configs/baselines/cm4-piControl`

Issue #893, opened 2026-03-03
Enable model parallelism


mcgibbon:
We'd like our code to run in a distributed mode where different ranks are responsible for different spatial regions of the domain. This will allow us to run at higher resolution, for example, or to run inference at higher speed, especially on lower-end GPUs.

mahf708:
I can't add subissues here, but here's another one: https://github.com/ai2cm/ace/issues/909

PR #900, opened 2026-03-04
make test_step distributed backend friendly

### LINKED ISSUE #894: Run CSFNO-based Step.step with model parallelism
A regression test has been added to `fme/core/distributed/parallel_tests/test_step.py::test_step_regression` which asserts that the result of a model's forward step is equal to a regression target. This should be true regardless of the spatial decomposition of the model. When run in a model-parallel configuration, the test currently passes because it simply runs the full domain configuration on each rank. SingleModuleStep.step and its callers need to be updated to use model parallelism in a way that they still pass this test.

test_step_regression also needs to be updated, for example so the output is global-gathered across ranks, and then the cache/compare operation is only run on rank 0.

mahf708:
adjusting step ops to make sure test_step now passes for a variety of tests with sp enabled.

Changes:
- expanded api of Distributed to include get_sht, get_isht, get_disco_conv_cls (to discretely handle thd stuff)
- expanded api of Distributed to include spatial_reduce_sum (but could potentially one day be improved?)
- patched gloo backend with all_to_all needed by some thd pieces (pretty slow and hacky, but only for testing now)

- [x] Tests added

Resolves #894

### DIFF
--- a/fme/core/coordinates.py
+++ b/fme/core/coordinates.py
@@ -329,6 +329,19 @@ class HybridSigmaPressureCoordinate(VerticalCoordinate):
         return (integrand * pressure_thickness).sum(dim=-1) / GRAVITY
 
 
+def dz_from_idepth(
+    idepth: torch.Tensor, deptho: torch.Tensor | None = None
+) -> torch.Tensor:
+    if deptho is not None:
+        z_top = idepth[..., :-1]
+        z_bot = idepth[..., 1:]
+        deptho = deptho.unsqueeze(-1)
+        dz = torch.clamp(deptho, min=z_top, max=z_bot) - z_top
+    else:
+        dz = idepth.diff(dim=-1)
+    return dz
+
+
 @dataclasses.dataclass
 class DepthCoordinate(VerticalCoordinate):
     """
@@ -341,10 +354,15 @@ class DepthCoordinate(VerticalCoordinate):
             it must be one shorter than idepth. The mask may have additional dimensions
             before the vertical, which are assumed to be broadcastable to match the
             integrand when computing integrals.
+        deptho: optional sea floor depth in meters at each horizontal grid point.
+            When provided, layer thicknesses account for partial bottom cells.
+            Must be broadcastable to the spatial dimensions of ``mask``
+            (i.e. ``mask.shape[:-1]``).
     """
 
     idepth: torch.Tensor
     mask: torch.Tensor
+    deptho: torch.Tensor | None = None
 
     def __post_init__(self):
         if len(self.idepth.shape) != 1:
@@ -361,6 +379,11 @@ class DepthCoordinate(VerticalCoordinate):
                 f"Got idepth.shape: {self.idepth.shape} and mask.shape: "
                 f"{self.mask.shape}."
             )
+        self._dz = dz_from_idepth(self.idepth, self.deptho)
+
+    @property
+    def dz(self) -> torch.Tensor:
+        return self._dz
 
     def __len__(self):
         """The number of vertical layer interfaces."""
@@ -410,9 +433,11 @@ class DepthCoordinate(VerticalCoordinate):
         return {"idepth": self.idepth.cpu().numpy()}
 
     def to(self, device: str) -> "DepthCoordinate":
-        idepth_on_device = self.idepth.to(device)
-        mask_on_device = self.mask.to(device)
-        return DepthCoordinate(idepth_on_device, mask_on_device)
+        return DepthCoordinate(
+            idepth=self.idepth.to(device),
+            mask=self.mask.to(device),
+            deptho=self.deptho.to(device) if self.deptho is not None else None,
+        )
 
     def __eq__(self, other) -> bool:
         if not isinstance(other, DepthCoordinate):
@@ -422,13 +447,32 @@ class DepthCoordinate(VerticalCoordinate):
             torch.testing.assert_close(self.mask, other.mask)
         except AssertionError:
             return False
+        if (self.deptho is None) != (other.deptho is None):
+            return False
+        if self.deptho is not None and other.deptho is not None:
+            try:
+                torch.testing.assert_close(self.deptho, other.deptho)
+            except AssertionError:
+                return False
         return True
 
     def __repr__(self) -> str:
-        return f"DepthCoordinate(\n    idepth={self.idepth},\n    mask={self.mask}\n)"
+        parts = [
+            f"    idepth={self.idepth}",
+            f"    mask={self.mask}",
+        ]
+        if self.deptho is not None:
+            parts.append(f"    deptho={self.deptho}")
+        return "DepthCoordinate(\n" + ",\n".join(parts) + "\n)"
 
     def as_dict(self) -> TensorMapping:
-        return {"idepth": self.idepth, "mask": self.mask}
+        result: dict[str, torch.Tensor] = {
+            "idepth": self.idepth,
+            "mask": self.mask,
+        }
+        if self.deptho is not None:
+            result["deptho"] = self.deptho
+        return result
 
     def depth_integral(self, integrand: torch.Tensor) -> torch.Tensor:
         """Compute the depth integral of the integrand.
@@ -454,10 +498,9 @@ class DepthCoordinate(VerticalCoordinate):
                 f"Got integrand.shape: {integrand.shape} and idepth.shape: "
                 f"{self.idepth.shape}."
             )
-        layer_thickness = self.idepth.diff(dim=-1)
-        ohc = (integrand * layer_thickness * self.mask).nansum(dim=-1)
-        mask_0 = self.mask.select(dim=-1, index=0).expand(ohc.shape)
-        return ohc.where(mask_0 > 0, float("nan"))
+        integral = (integrand * self.dz * self.mask).nansum(dim=-1)
+        mask_0 = self.mask.select(dim=-1, index=0).expand(integral.shape)
+        return integral.where(mask_0 > 0, float("nan"))
 
 
 @dataclasses.dataclass
--- a/fme/core/dataset/test_xarray.py
+++ b/fme/core/dataset/test_xarray.py
@@ -981,32 +981,58 @@ def test__get_vertical_coordinate_hybrid_sigma_pressure():
     assert vertical_coordinate.bk[0] == 0.5
 
 
-def test__get_vertical_coordinate_depth_no_mask():
-    data = xr.Dataset({"idepth_0": 1.0, "idepth_1": 2.0})
+@pytest.mark.parametrize("has_deptho", [False, True], ids=["no_deptho", "with_deptho"])
+def test__get_vertical_coordinate_depth_no_mask(has_deptho):
+    data_vars: dict = {"idepth_0": 1.0, "idepth_1": 2.0}
+    if has_deptho:
+        data_vars["deptho"] = 1.5
+    data = xr.Dataset(data_vars)
     vertical_coordinate = _get_vertical_coordinate(data, dtype=None)
     assert isinstance(vertical_coordinate, DepthCoordinate)
     assert vertical_coordinate.idepth[0] == 1.0
     assert vertical_coordinate.mask[0] == 1.0
+    if has_deptho:
+        assert vertical_coordinate.deptho is not None
+        assert float(vertical_coordinate.deptho) == 1.5
+    else:
+        assert vertical_coordinate.deptho is None
 
 
-def test__get_vertical_coordinate_depth_with_lat_dependent_mask():
-    data = xr.Dataset(
-        data_vars={
-            "idepth_0": 1.0,
-            "idepth_1": 2.0,
-            "idepth_2": 3.0,
-            "mask_0": ("lat", np.array([1.0, 1.0])),
-            "mask_1": ("lat", np.array([0.0, 1.0])),
-        },
-        coords={
-            "lat": np.array([1.0, 2.0]),
-        },
-    )
+@pytest.mark.parametrize("has_deptho", [False, True], ids=["no_deptho", "with_deptho"])
+def test__get_vertical_coordinate_depth_with_lat_dependent_mask(has_deptho):
+    data_vars: dict = {
+        "idepth_0": 1.0,
+        "idepth_1": 2.0,
+        "idepth_2": 3.0,
+        "mask_0": ("lat", np.array([1.0, 1.0])),
+        "mask_1": ("lat", np.array([0.0, 1.0])),
+    }
+    if has_deptho:
+        data_vars["deptho"] = ("lat", np.array([2.5, 3.0]))
+    data = xr.Dataset(data_vars, coords={"lat": np.array([1.0, 2.0])})
     vertical_coordinate = _get_vertical_coordinate(data, dtype=None)
     assert isinstance(vertical_coordinate, DepthCoordinate)
     assert vertical_coordinate.idepth[0] == 1.0
     assert vertical_coordinate.idepth.shape == (3,)
     assert vertical_coordinate.mask.shape == (2, 2)
+    if has_deptho:
+        assert vertical_coordinate.deptho is not None
+        assert vertical_coordinate.deptho.shape == (2,)
+    else:
+        assert vertical_coordinate.deptho is None
+
+
+def test__get_vertical_coordinate_depth_with_time_dependent_deptho():
+    data = xr.Dataset(
+        data_vars={
+            "idepth_0": 1.0,
+            "idepth_1": 2.0,
+            "deptho": ("time", np.array([1.5, 1.5])),
+        },
+        coords={"time": np.array([1.0, 2.0])},
+    )
+    with pytest.raises(ValueError, match="'deptho' must be time-independent"):
+        _get_vertical_coordinate(data, dtype=None)
 
 
 def test__get_vertical_coordinate_depth_with_time_dependent_mask():
--- a/fme/core/dataset/xarray.py
+++ b/fme/core/dataset/xarray.py
@@ -103,6 +103,7 @@ def _get_vertical_coordinate(
         )
 
     coordinate: VerticalCoordinate
+    deptho = None
     if len(idepth_list) > 0:
         if "mask_0" in ds.data_vars:
             mask_layers = {
@@ -121,7 +122,18 @@ def _get_vertical_coordinate(
                 "mask set to 1 at all layers."
             )
             mask = torch.ones(len(idepth_list) - 1, dtype=dtype)
-        coordinate = DepthCoordinate(torch.as_tensor(idepth_list, dtype=dtype), mask)
+        if "deptho" in ds.data_vars:
+            if "time" in ds["deptho"].dims:
+                raise ValueError("'deptho' must be time-independent.")
+            deptho = torch.as_tensor(ds["deptho"].values, dtype=dtype)
+        else:
+            logger.warning(
+                "Dataset does not have a variable named 'deptho' (sea floor depth). "
+                "The ocean depth integral will not account for partial bottom cells."
+            )
+        coordinate = DepthCoordinate(
+            torch.as_tensor(idepth_list, dtype=dtype), mask, deptho
+        )
     elif len(ak_list) > 0 and len(bk_list) > 0:
         coordinate = HybridSigmaPressureCoordinate(
             ak=torch.as_tensor(ak_list, dtype=dtype),
--- a/fme/core/test_coordinates.py
+++ b/fme/core/test_coordinates.py
@@ -8,6 +8,7 @@ from fme.core.coordinates import (
     HEALPixCoordinates,
     HybridSigmaPressureCoordinate,
     LatLonCoordinates,
+    dz_from_idepth,
 )
 from fme.core.mask_provider import MaskProvider
 
@@ -217,6 +218,83 @@ def test_healpix_ops_raises_value_error_with_mask():
     expected_msg = "HEALPixCoordinates does not support a mask"
     with pytest.raises(NotImplementedError, match=expected_msg):
         healpix_coords.get_gridded_operations(mask_provider=mask_provider)
+
+
+@pytest.mark.parametrize(
+    "idepth, deptho, expected",
+    [
+        pytest.param(
+            torch.tensor([0.0, 10.0, 50.0, 100.0]),
+            None,
+            torch.tensor([10.0, 40.0, 50.0]),
+            id="no_deptho",
+        ),
+        pytest.param(
+            torch.tensor([0.0, 10.0, 50.0, 100.0]),
+            torch.tensor(50.0),
+            torch.tensor([10.0, 40.0, 0.0]),
+            id="deptho_at_interface",
+        ),
+        pytest.param(
+            torch.tensor([0.0, 10.0, 50.0, 100.0]),
+            torch.tensor(30.0),
+            torch.tensor([10.0, 20.0, 0.0]),
+            id="deptho_between_interfaces",
+        ),
+        pytest.param(
+            torch.tensor([0.0, 10.0, 50.0, 100.0]),
+            torch.tensor(100.0),
+            torch.tensor([10.0, 40.0, 50.0]),
+            id="deptho_at_bottom",
+        ),
+        pytest.param(
+            torch.tensor([0.0, 10.0, 50.0, 100.0]),
+            torch.tensor(101.0),
+            torch.tensor([10.0, 40.0, 50.0]),
+            id="deptho_below_bottom",
+        ),
+    ],
+)
+def test_dz_from_idepth(idepth, deptho, expected):
+    result = dz_from_idepth(idepth, deptho)
+    torch.testing.assert_close(result, expected)
+
+
+@pytest.mark.parametrize(
+    "leading_dims",
+    [(), (2,), (2, 3)],
+    ids=["no_leading", "batch", "batch_time"],
+)
+def test_depth_integral_with_deptho(leading_dims):
+    nlat, nlon = 2, 3
+    idepth = torch.tensor([0.0, 10.0, 50.0, 100.0])
+    nz = len(idepth) - 1
+    mask = torch.ones(nlat, nlon, nz)
+
+    deptho = torch.full((nlat, nlon), 100.0)
+    # cell (0, 0): deptho at intermediate interface -> layers below are zero-thickness
+    deptho[0, 0] = 50.0
+    # cell (0, 1): deptho between interfaces -> partial bottom cell
+    deptho[0, 1] = 30.0
+
+    coord = DepthCoordinate(idepth=idepth, mask=mask, deptho=deptho)
+
+    integrand = torch.ones(*leading_dims, nlat, nlon, nz)
+    result = coord.depth_integral(integrand)
+    assert result.shape == (*leading_dims, nlat, nlon)
+
+    # cell (0, 0): dz = [10, 40, 0] -> integral = 50
+    torch.testing.assert_close(
+        result[..., 0, 0], torch.tensor(50.0).expand(leading_dims)
+    )
+    # cell (0, 1): dz = [10, 20, 0] -> integral = 30
+    torch.testing.assert_close(
+        result[..., 0, 1], torch.tensor(30.0).expand(leading_dims)
+    )
+    # cell (1, 0): deptho = 100 (full depth), dz = [10, 40, 50] -> integral = 100
+    torch.testing.assert_close(
+        result[..., 1, 0], torch.tensor(100.0).expand(leading_dims)
+    )
 
 
 @pytest.mark.skipif(e2ghpx is None, reason="earth2grid is not available")


mcgibbon:
> expanded api of Distributed (could potentially be simplified?)

Noting this before I look at the code so maybe it's not an issue, but generally describing what was expanded and why in the PR description could make it easier to review the changes.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -4,9 +4,11 @@
 from typing import Literal
 
 import torch
+import torch_harmonics.distributed as thd
```
mcgibbon: Issue: Aspirationally, only `Distributed` should access distributed-aware code

Suggestion: If you need to initialize a distributed version of the SHT it means we need a `get_real_sht` or similar method on Distributed.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -39,6 +43,14 @@ def isotropic_noise(
     scale = math.sqrt(4.0 * math.pi) / lmax  # (Unsöld theorem ⇒ L = lmax)
     alm = (real + 1j * imag) * scale
 
+    # For distributed iSHT, slice to local spectral extent
+    if isinstance(isht, thd.DistributedInverseRealSHT):
```
mcgibbon: Issue: Low-level distributed-aware code should be refactored into a method on `dist`, or put into the SHT object itself which will be distributed-aware or not depending on context.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -103,11 +121,15 @@ def forward(
         else:
             raise ValueError(f"Invalid noise type: {self.noise_type}")
 
+        h_slice, w_slice = self._get_spatial_slices()
+
         if self.pos_embed is not None:
-            embedding_pos = self.pos_embed.repeat(noise.shape[0], 1, 1, 1)
+            pos_local = self.pos_embed[..., h_slice, w_slice]
```
mcgibbon, line 115: Suggestion (optional): only store the local values in `self.pos_embed`, the other values are unused and pos_embed can be quite large. The init value is zeros, so it's not an issue.

Issue with my suggestion: Will loading a checkpoint work properly? We probably need to add a regression test that loads a saved checkpoint, in addition to the from-pure-init one (or maybe add an assert that when a checkpoint is loaded, it gives the same model state as from-pure-init).

fme/ace/registry/stochastic_sfno.py
mcgibbon, line 104: General issue (not with your code): To achieve reproducibility independent of decomposition, I think we're going to need to initialize the global random state (with the same seed on each rank, or on root and then scatter) and then sub-select the local slice. This would need to be as a flag we basically just use for testing, since in production especially at high resolutions this isn't desired, and gaussian noise with a different seed on each rank is perfectly fine (I think Distributed does seed randomness differently on each rank? Maybe?)

Feel free to open issues related to this if you want to track it.

fme/core/distributed/distributed.py
```diff
@@ -393,6 +398,41 @@ def get_seed(self) -> int:
...
+
+    @property
+    def w_group(self):
+        return self._distributed.w_group
+
+    @property
+    def is_spatial_parallel(self) -> bool:
+        return self._distributed.is_spatial_parallel
+
+    def get_spatial_slices(self, h: int, w: int) -> tuple[slice, slice]:
```
mcgibbon: Issue: get_local_slices already implements this, it may not be clear but I believe it assumes the shape is (..., H, W). Please use it, feel free to update the docstring if it doesn't currently describe what it does.

fme/core/distributed/distributed.py
```diff
@@ -393,6 +398,41 @@ def get_seed(self) -> int:
...
+    def h_group(self):
+        return self._distributed.h_group
+
+    @property
+    def w_group(self):
+        return self._distributed.w_group
+
+    @property
+    def is_spatial_parallel(self) -> bool:
+        return self._distributed.is_spatial_parallel
```
mcgibbon: Issue: These are low-level details we want to insulate the rest of the code from. Our prior is that if an operation needs to use these, it should be a method on Distributed. We can discuss specific examples if it seems like that isn't the case.

For example, for now you can keep `h_rank` and `w_rank` to replicate the patterns used in Makani's module initialization code, even if we will want to delete them later.

As much as possible, please remove these methods in favor of higher-level methods like `get_sht`.

fme/core/models/conditional_sfno/layers.py
```diff
@@ -202,6 +202,15 @@ def __init__(
             self.W_scale_pos = None
             self.W_bias_pos = None
         if global_layer_norm:
+            from fme.core.distributed.distributed import Distributed
+
+            if (
+                Distributed.is_entered()
+                and Distributed.get_instance().is_spatial_parallel
+            ):
+                raise ValueError(
```
mcgibbon: You couldn't know this, but our experiments found that `global_layer_norm` is just way too slow. Feel free to update this with `raise ValueError("global_layer_norm is no longer supported")` regardless of distributed.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -256,6 +257,32 @@ def _add_singleton_group_dim(
         if weight.shape == ungrouped_shape:
             state_dict[key] = weight.view(1, *ungrouped_shape)
 
+    @staticmethod
+    def _slice_for_distributed(
```
mcgibbon: We can proceed with this temporarily, but this is going to be too finnicky/copy-paste of a solution to implement in general for loading distributed weights, I expect we will want to do something like Makani.

They require all parameters have attributes set that indicate the meaning of the trailing dimensions. Then, when weights are loaded, the loading routines automatically use this annotation to sub-select the loaded weights. These same attributes are also used to automatically reduce gradients across model parallel dimensions.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -256,6 +257,32 @@ def _add_singleton_group_dim(
...
+        missing_keys: list[str],
+        unexpected_keys: list[str],
+        error_msgs: list[str],
+    ) -> None:
+        """Slice global spectral weights to local extent for distributed mode."""
+        if not isinstance(module.inverse_transform, thd.DistributedInverseRealSHT):
+            return
+        isht = module.inverse_transform
+        l_start = sum(isht.l_shapes[: isht.comm_rank_polar])
+        l_local = isht.l_shapes[isht.comm_rank_polar]
```
mcgibbon: issue: this is relying on something other than `Distributed` for distributed information.

Suggestion: For now, please duplicate the pattern used in makani: https://github.com/NVIDIA/makani/blob/main/makani/models/common/spectral_convolution.py except use a `dist.get_rank`. Also copy their choices for names, like `lat_shapes` and "h" for the rank instead of `l` and `polar`. I get that these are different, but we really are stealing the H decomposition and applying it to the polar axis, so it's better that the code describe this behavior. I have other issues with it and will want to refactor it, but we can do it later, and at least in the meantime it has the benefit of being closer to the Makani code.

I'd also like to copy-paste over the `self.weight.is_shared_mp` etc. attributes they have throughout the module definitions and use their methods for weight scatter/gather and gradient reduction, at least as our starting point. But that can happen after this PR.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -375,28 +383,78 @@ def get_lat_lon_sfnonet(
     modes_lat = int(h * hard_thresholding_fraction)
     modes_lon = int((w // 2 + 1) * hard_thresholding_fraction)
     data_grid = params.data_grid if hasattr(params, "data_grid") else "equiangular"
-    trans_down = th.RealSHT(
+
+    dist = Distributed.get_instance()
+    spatial = dist.is_spatial_parallel
+
+    if spatial:
+        if not thd.is_initialized():
```
mcgibbon: Question: Why does this run at this stage, and not during Distributed's original init?

fme/core/distributed/distributed.py
mcgibbon: As a short-term change if you want to avoid refactoring in your PR for now, you can leave this method intact but have it call `self._distributed.get_local_slices` instead (to remove the duplicated code and tests from the backends).

fme/core/distributed/model_torch_distributed.py
```diff
@@ -266,6 +266,53 @@ def barrier(self):
...
+        return (
+            slice(h_start, h_start + h_shapes[self._h_rank]),
+            slice(w_start, w_start + w_shapes[self._w_rank]),
+        )
+
+    def spatial_reduce_sum(self, tensor: torch.Tensor) -> torch.Tensor:
+        if self._h_size > 1:
+            torch.distributed.all_reduce(tensor, group=self._h_group)
+        if self._w_size > 1:
+            torch.distributed.all_reduce(tensor, group=self._w_group)
```
mcgibbon: Question: I'm fine doing it like this for now, but would it be more computationally efficient to reduce on both groups at the same time, somehow?

Question: Do you mean for this to be all_reduce, or do we only need the result on root (and in that case can use `reduce`? I could see the latter needing significantly more refactors to achieve (since it messes up the return type), so we could leave it for a later PR and open an issue.

Our Aggregators might get significantly more complex, since scalar metrics only need to be collected on root, while maps will need to be collected on each process and then gathered to root only during report time (one var at a time to avoid memory blowout), or possibly disabled altogether due to the size.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -375,28 +383,78 @@ def get_lat_lon_sfnonet(
...
+        # TODO: if it's gloo. But we could be more robust to other backends by checking
+        # TODO: if all_to_all is implemented instead of checking the backend name.
+        # NOTE: the patch is pretty slow ... but only used in cpu testing...
+        # NOTE: we may want to consider an MPI impl at some point for inference, but can wait
+        from fme.core.distributed._gloo_patch import patch_gloo_alltoall
+
+        patch_gloo_alltoall()
+        SHTClass = thd.DistributedRealSHT
+        ISHTClass = thd.DistributedInverseRealSHT
+        disco_conv_cls = thd.DistributedDiscreteContinuousConvS2
```
mcgibbon: Suggestion: Move this to a `Distributed.get_disco_conv` method, the way we will initialize RealSHT, instead of passing the class reference.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -375,28 +383,78 @@ def get_lat_lon_sfnonet(
...
+
+    def _add_local_attrs(transform):
+        """Add local extent attributes needed by SpectralConvS2."""
+        if isinstance(
+            transform, (thd.DistributedRealSHT, thd.DistributedInverseRealSHT)
+        ):
+            transform.lmax_local = transform.l_shapes[transform.comm_rank_polar]
+            transform.mmax_local = transform.m_shapes[transform.comm_rank_azimuth]
+            transform.lpad_local = 0
+            transform.mpad_local = 0
```
mcgibbon: We should try to abstract this away in some way into Distributed and/or into wrappers over the SHT objects.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -375,28 +383,78 @@ def get_lat_lon_sfnonet(
...
+    trans = SHTClass(
         *img_shape, lmax=modes_lat, mmax=modes_lon, grid="legendre-gauss"
     ).float()
-    itrans = th.InverseRealSHT(
+    itrans = ISHTClass(
         h, w, lmax=modes_lat, mmax=modes_lon, grid="legendre-gauss"
     ).float()
 
+    for t in (trans_down, itrans_up, trans, itrans):
+        _add_local_attrs(t)
```
mcgibbon: A `Distributed.get_sht` method could handle these automatically.

fme/core/gridded_ops.py
```diff
@@ -394,6 +397,126 @@ def get_initialization_kwargs(self) -> dict[str, Any]:
         return {"area_weights": self._cpu_area}
 
 
+class DistributedLatLonOperations(LatLonOperations):
```
mcgibbon: Rather than make LatLonOperations distributed or not, we should have it internally use `Distributed` in a way that seamlessly handles these concerns regardless of distributed context.

E.g. a `get_real_sht` could return the result of `dist.get_real_sht`.

Things that aren't implemented for now would raise when `dist` is called.

fme/core/gridded_ops.py
```diff
@@ -394,6 +397,126 @@ def get_initialization_kwargs(self) -> dict[str, Any]:
...
+        # TODO: gradient_magnitude uses finite differences over the local shard,
+        # so boundary pixels have incorrect gradients (no halo exchange), and the
+        # subsequent weighted_nanmean is never all-reduced across ranks.
+        raise NotImplementedError(
+            "area_weighted_gradient_magnitude_percent_diff is not supported for "
+            "DistributedLatLonOperations."
+        )
+
+    @property
+    def zonal_mean(self) -> Callable[[torch.Tensor], torch.Tensor]:
```
mcgibbon: You could trivially implement some of these in the short term by calling `gather_global` (or more likely something that only gathers along spatial parallel dimensions) before calling the existing code. Otherwise, please update to call `dist` methods that will trigger `NotImplementedError` when they eventually reach the model-parallel backend.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -17,13 +19,15 @@
...
 def isotropic_noise(
     leading_shape: tuple[int, ...],
-    lmax: int,  # length of the ℓ axis expected by isht
-    mmax: int,  # length of the m axis expected by isht
+    lmax: int,  # length of the ℓ axis expected by isht (global)
+    mmax: int,  # length of the m axis expected by isht (global)
     isht: Callable[[torch.Tensor], torch.Tensor],
     device: torch.device,
 ) -> torch.Tensor:
+    # TODO: is this happening at every rank? are ranks sharing the same seed?
```
mcgibbon: Check Distributed.set_seed, I think it adds the rank.

fme/core/distributed/parallel_tests/test_backend.py
```diff
@@ -243,3 +243,63 @@ def test_data_parallel_rank_within_total():
...
+@pytest.mark.parallel
+def test_is_spatial_parallel_consistency():
+    """is_spatial_parallel should be True iff h_size > 1 or w_size > 1."""
+    dist = Distributed.get_instance()
+    expected = dist.h_size > 1 or dist.w_size > 1
+    assert dist.is_spatial_parallel == expected
+
+
+@pytest.mark.parallel
+def test_all_to_all_works_after_gloo_patch():
```
mcgibbon: Suggestion: Maybe also assert it doesn't work when you don't patch.

fme/core/distributed/parallel_tests/test_step.py
```diff
@@ -358,6 +363,31 @@ def test_load_is_required_for_path_config(
...
 
+# TODO: move this to ModelTorchDistributed backend and add unit tests
+def scatter_spatial(data: TensorDict, img_shape: tuple[int, int]) -> TensorDict:
+    """Slice global tensors to the local spatial chunk for this rank."""
+    h_slice, w_slice = Distributed.get_instance().get_spatial_slices(*img_shape)
+    return {k: v[..., h_slice, w_slice].contiguous() for k, v in data.items()}
+
+
+# TODO: move this to ModelTorchDistributed backend and add unit tests
+def gather_spatial(data: TensorDict, img_shape: tuple[int, int]) -> TensorDict:
```
mcgibbon: Suggestion: use `dist.gather_global` to reduce duplicated code.

fme/core/distributed/parallel_tests/test_step.py
```diff
@@ -358,6 +363,31 @@ def test_load_is_required_for_path_config(
...
+        return data
+    h_slice, w_slice = dist.get_spatial_slices(*img_shape)
+    result = {}
+    for k, v in data.items():
+        global_shape = list(v.shape)
+        global_shape[-2] = img_shape[0]
+        global_shape[-1] = img_shape[1]
+        global_tensor = torch.zeros(global_shape, dtype=v.dtype, device=v.device)
+        global_tensor[..., h_slice, w_slice] = v
+        result[k] = dist.spatial_reduce_sum(global_tensor)
```
mcgibbon: Interesting approach!

fme/core/models/conditional_sfno/layers.py
```diff
@@ -202,6 +202,15 @@ def __init__(
             self.W_scale_pos = None
             self.W_bias_pos = None
         if global_layer_norm:
+            from fme.core.distributed.distributed import Distributed
+
+            if (
+                Distributed.is_entered()
```
mcgibbon: Issue: rather than check if Distributed.is_entered(), we should just let the call fail on `.get_instance()` if it's not. The idea is to eagerly catch scripts that don't enforce distributed cleanup, and not be particularly clean about only enforcing this only in the cases where Distributed is actually used.

If we wanted the later, instead of this kind of check duplicated throughout the code we would only enforce "the context needs to be entered" when Distributed.get_instance() is using a non-null backend. But we decided to prioritize the former.

fme/core/distributed/distributed.py
```diff
@@ -33,6 +33,11 @@ class Distributed:
 
     _entered: bool = False
 
+    @classmethod
+    def is_entered(cls) -> bool:
+        """Return whether a ``Distributed.context()`` is currently active."""
+        return cls._entered
```
mcgibbon: For reasons described in the comment I posted where this is used, please delete it.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -532,6 +594,7 @@ def __init__(
             embed_dim_pos=0,
         ),
         global_layer_norm: bool = False,
+        local_img_shape: Optional[Tuple[int, int]] = None,
```
mcgibbon: Suggestion: Update to follow what Makani does for passing img_shape/local shapes. https://github.com/NVIDIA/makani/blob/main/makani/models/networks/sfnonet.py

mcgibbon (changes requested):
Thank you! Exciting that this is passing the regression test!

I have a fair number of high level comments, hopefully they are not too difficult. If there are any you feel the need to push back on to get the code in for whatever reason and iterate on in another PR let me know. It feels like most of these will help avoid code bloat and make it easier to write follow-on PRs (for distributed concerns or otherwise).

High-level comments:
- Use getter methods on Distributed for the SHT and Disco convolution
- Replicate the patterns Makani uses for transporting local and global img_shape, lat/polar slices
- Remove duplicated implementation of `get_local_slices`, transfer over or delete duplicated tests
- Remove low-level details like `is_spatial_parallel` from Distributed, as much as possible
- Rather than duplicating LatLonOperations, make LatLonOperations handle distributed concerns via calls to (possibly new) methods on `dist`, which for now can raise NotImplementedError on the spatial parallel backend.

fme/ace/registry/stochastic_sfno.py
mahf708: I like the sound of this suggestion, but it's unclear to me :( can you elaborate or can we address this with when we deal backward()?

fme/ace/registry/stochastic_sfno.py
mahf708: Will open an issue

fme/core/distributed/model_torch_distributed.py
mahf708: I think we want all_reduce here; and I tried to add a simple opt (on the spatial group together)

fme/core/gridded_ops.py
mahf708: For now, I am raising inside LatLonOperations with a check on _local, but this could be further refactored. Let me know what you think when you get a chance and we can proceed to either further refactor or punting.

fme/core/gridded_ops.py
mahf708: see above https://github.com/ai2cm/ace/pull/900#discussion_r2887424997

fme/core/distributed/parallel_tests/test_step.py
mahf708: This was difficult --- gather_global will require a deep refactor to accomodate for this specific use case OR a refactor of the test itself. I didn't feel comfortable doing either for now (since it is not part of the added code here) and in the context of already needing to tinker with existing code in several places.

Can you take a second look and let me know what you prefer? For now, the path forward in my view is either a refactor of gather_global or a refactor of the test below. Do you prefer proceeding eithe direction? Maybe you have different ideas?

fme/core/distributed/parallel_tests/test_step.py
mahf708: Mostly not mine! I don't think this is efficient or well-thought out though, so if you don't like it, maybe we should get rid of it...?

fme/core/models/conditional_sfno/sfnonet.py
mahf708: You made a similar comment about s2convs in https://github.com/ai2cm/ace/pull/900#discussion_r2884760662. I spent a bit of time thinking through both and I think your suggestion will be good to follow through when we add more stuff. For now, I'm just doing basic stuff and hiding impl detail away in dist (after your other suggestions). There are still some leftover items. I can imagine we will revisit this in later stages. However, I am happy if you push back and say we should mimic makani sooner rather than later. If so, please provide more pointers because I am not entirely sure I follow if you have specific concerns or nothing specific...

mahf708:
@mcgibbon thanks for the thorough and patient review. I pushed ff717b5 to address as much as I could. For the remaining ones, I left some comments seeking clarification or potentially punting. For comments where I felt I mostly addressed your concerns to high degree I simply resolved the comments to avoid distraction in navigation. 

Two main points worth bringing up:
- https://github.com/ai2cm/ace/pull/900#discussion_r2887430331: The two gather methods (gather_spatial and gather_global) differ somewhat significantly in signature and combining them means either refactoring gather_global or refactoring the test. Neither of which feels comfortable ... but either is doable, so let me know what you prefer. 
- https://github.com/ai2cm/ace/pull/900#discussion_r2887450271: Following maknai is the better strategy longer term, but is not readily clear to me yet in the context of existing code, so punting (while knowing we may have to refactor some things) may be good, but I am also fine if you insist on us cleanly doing this (the makani) way at the outset by anticipating code yet to exist here.

Thanks!

mcgibbon:
> Following maknai is the better strategy longer term, but is not readily clear to me yet in the context of existing code, so punting (while knowing we may have to refactor some things) may be good, but I am also fine if you insist on us cleanly doing this (the makani) way at the outset by anticipating code yet to exist here.

The broader makani thing of "set attributes on parameters and then use them to handle things more automatically" is a better long-term strategy, but I don't need you to do it in this PR. I actually don't like what makani does for longer term for shape propagation (which is what I'm saying to do on our end for now), the comment is about changing a few specific things you're doing. Specifically, I'm not proposing to increase this PR's scope.

Overall I think my comment is too confusing, I'll post some more specific low-level comments and hopefully that will make it clear.

fme/core/distributed/parallel_tests/test_step.py
mcgibbon: This surprises me, I'll take a look.

fme/ace/registry/stochastic_sfno.py
mcgibbon: Yeah we can punt on this until later.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -17,10 +18,11 @@
 )
 
 
+# TODO: consider more distributed organization to avoid large allocation?
```
mcgibbon: Suggestion: Link the issue you opened. FWIW we generally avoid adding TODOs into the code in preference of issues, a TODO for us usually indicates something that's going to be resolved before the PR is merged. If you find it helpful to have TODOs during these PRs, I think we can keep them until spatial distributed is "working" and clear them out then (or if/when someone complains about them).

fme/ace/registry/stochastic_sfno.py
mcgibbon: Opened https://github.com/ai2cm/ace/issues/907

fme/core/distributed/parallel_tests/test_step.py
```diff
@@ -37,6 +38,8 @@
 from fme.core.step.step import StepABC, StepSelector
 from fme.core.typing_ import TensorDict
 
+# TODO: consider making this a parameter
+# TODO: or adding a crazy big image like (1024, 2048) in a separate test?
 DEFAULT_IMG_SHAPE = (45, 90)
```
mcgibbon: Is there a reason this needs to be a parameter? We can open an issue if we need to discuss. If we are going to add TODO's, I'd prefer limiting it to ones that are clearly needed.
```suggestion
DEFAULT_IMG_SHAPE = (45, 90)
```

fme/core/distributed/base.py
```diff
@@ -111,3 +112,27 @@ def barrier(self): ...
 
     @abstractmethod
     def shutdown(self): ...
+
+    def get_sht(
+        self, nlat: int, nlon: int, lmax: int, mmax: int, grid: str
+    ) -> nn.Module:
+        """Create a forward SHT (possibly distributed)."""
+        import torch_harmonics as th
```
mcgibbon: Suggestion: Move this import to the top of the file.

fme/core/distributed/base.py
```diff
@@ -111,3 +112,27 @@ def barrier(self): ...
...
+
+    def get_isht(
+        self, nlat: int, nlon: int, lmax: int, mmax: int, grid: str
+    ) -> nn.Module:
+        """Create an inverse SHT (possibly distributed)."""
+        import torch_harmonics as th
+
+        return th.InverseRealSHT(nlat, nlon, lmax=lmax, mmax=mmax, grid=grid).float()
+
+    def get_disco_conv_cls(self):
```
mcgibbon: Chore: Missing type hint

fme/core/distributed/base.py
```diff
@@ -111,3 +112,27 @@ def barrier(self): ...
 
     @abstractmethod
     def shutdown(self): ...
+
+    def get_sht(
```
mcgibbon, line 117: Suggestion: Please make this abstract and require children to implement it. If you prefer you can open an issue (link this comment in the description) and we can do it in a later PR.

Inheritance is a source of many headaches, one of the ways we avoid these headaches is to try to avoid making non-final concrete implementations in an abstract base class. That is to say, either it should be abstract and always defined in the children, or it should be final and never defined in the children. That way, you always know you're reading the code that gets run.

fme/core/distributed/distributed.py
```diff
@@ -393,6 +416,18 @@ def get_seed(self) -> int:
         """
         return self._seed
 
+    def get_sht(self, nlat, nlon, lmax, mmax, grid):
+        return self._distributed.get_sht(nlat, nlon, lmax, mmax, grid)
+
+    def get_isht(self, nlat, nlon, lmax, mmax, grid):
+        return self._distributed.get_isht(nlat, nlon, lmax, mmax, grid)
+
+    def get_disco_conv_cls(self):
```
mcgibbon: Issue: It's a little awkward that this gets a class and the others get an object.

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Calling `dist.get_disco_conv` will make this consistent with SHT, avoiding the awkwardness where SHT is given a builder while disco conv is given a method that returns a class. The goal is for this class to be globally defined based on the distributed context, so there isn't a need to pass it as an argument to the init.

fme/core/distributed/base.py
```diff
@@ -40,7 +41,7 @@ def total_data_parallel_ranks(self) -> int:
     def local_batch_size(self, batch_size: int) -> int: ...
 
     @abstractmethod
-    def get_local_slices(self, tensor_shape, data_parallel_dim: int | None): ...
+    def get_local_slices(self, tensor_shape, data_parallel_dim: int | None = None): ...
```
mcgibbon, line 44: Question: What required this default to be added?

fme/core/distributed/model_torch_distributed.py
```diff
@@ -266,6 +299,45 @@ def barrier(self):
...
 
+    def spatial_reduce_sum(self, tensor: torch.Tensor) -> torch.Tensor:
+        if self._h_size > 1 or self._w_size > 1:
+            torch.distributed.all_reduce(tensor, group=self._spatial_group)
+        return tensor
+
+    def _ensure_thd_initialized(self):
+        """Lazily init torch_harmonics.distributed and gloo patch."""
+        if not self._thd_initialized:
+            import torch_harmonics.distributed as thd
```
mcgibbon: Chore: move import to top of file.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -266,6 +299,45 @@ def barrier(self):
...
+
+            if not thd.is_initialized():
```
mcgibbon: Question: Why is this not initialized during distributed init?

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -60,9 +61,11 @@ def _compute_cutoff_radius(nlat, kernel_shape, basis_type):
 
 
 class DiscreteContinuousConvS2(nn.Module):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, conv_cls=None, **kwargs):
         super().__init__()
-        self.conv = th.DiscreteContinuousConvS2(*args, **kwargs)
+        if conv_cls is None:
+            conv_cls = th.DiscreteContinuousConvS2
+        self.conv = conv_cls(*args, **kwargs)
```
mcgibbon: Suggestion: call a `dist.get_disco_conv_s2(*args, **kwargs)` here.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -354,6 +361,21 @@ def forward(self, x, context: Context):
...
+
+
+def _local_img_shape(transform: nn.Module) -> tuple[int, int]:
+    """Derive local (h, w) from an SHT transform via Distributed slices."""
+    dist = Distributed.get_instance()
+    h_slice, w_slice = dist.get_local_slices((transform.nlat, transform.nlon))
+    return (
+        _slice_len(h_slice, transform.nlat),
+        _slice_len(w_slice, transform.nlon),
+    )
```
mcgibbon: No change needed right now, but we might want a helper on Distributed for getting these local sizes.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Optional for this PR (it would simplify it if you want to), but as mentioned on slack, let's move this slicing operation to the `.forward` call and just store the full parameter for now. That means we can punt on loss reductions and on weight loading/saving until later.

Doing this requires also reverting some stuff in the init code so that the full weight is initialized.

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
         mask_provider: MaskProviderABC = NullMaskProvider,
         grid: str = "legendre-gauss",
     ):
-        # requires weights are longitudinally uniform
+        from fme.core.distributed.distributed import Distributed
```
mcgibbon: Chore: move import to top of file.

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
+
+        self._dist = Distributed.get_instance()
```
mcgibbon: Issue: setting this on init prevents certain operations from working properly, such as monkey patching Distributed or contexts that say to do operations in a non-distributed mode.

Chore: Instead of storing `self._dist`, call `dist = Distributed.get_instance()` at the top of each method using it.

fme/core/gridded_ops.py
```diff
@@ -332,9 +356,12 @@ def area_weighted_sum(
         name: str | None = None,
     ) -> torch.Tensor:
         area_weights = self._get_area_weights(data, name)
-        return metrics.weighted_sum(
+        local_sum = metrics.weighted_sum(
             data, area_weights, dim=self.HORIZONTAL_DIMS, keepdim=keepdim
         )
+        if self._is_local:
+            return self._dist.spatial_reduce_sum(local_sum)
```
mcgibbon: Issue: Low-level distributed logic outside of Distributed.

Chore: Always return `self._dist.spatial_reduce_sum(local_sum)`, just have the implementation `return local_sum` in the non-spatial backends.

This can be split into another PR if you open an issue linking to this comment.

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
-        self._cpu_mask_provider = mask_provider.to("cpu")
-        self._grid = "legendre-gauss"
 
     @property
     def zonal_mean(self) -> Callable[[torch.Tensor], torch.Tensor]:
+        if self._is_local:
+            raise NotImplementedError(
+                "zonal_mean is not yet supported under spatial parallelism."
+            )
         return self._zonal_mean
```
mcgibbon: Issue: Low-level distributed logic outside of Distributed.

Suggestion: In `def _zonal_mean(self, ...)` (which is really zonal nanmean) call some kind of new distributed reduction method which computes the mean along a dimension, but also reduces it across a group if a distributed group exists for that dimension. This would simply do a nanmean on the non-spatial backends.

This suggestion can be pulled into an issue instead of resolved in this PR as it's a larger task.

fme/core/gridded_ops.py
```diff
@@ -343,6 +370,18 @@ def area_weighted_mean(
         name: str | None = None,
     ) -> torch.Tensor:
         area_weights = self._get_area_weights(data, name)
+        if self._is_local:
```
mcgibbon: Issue: low-level distributed logic outside of Distributed.

Chore: Move this logic to a new `Distributed.area_weighted_mean(data, area_weights, keepdim)` method. I'd suggest reasonable assumptions (to put in the docstring) are that the last two dims of data are [H, W], and that area_weights is 2D of shape [H, W].

This can be split into another PR if you open an issue linking to this comment.

You could consider a similar zonal_mean (for a later PR).

fme/core/gridded_ops.py
```diff
@@ -354,6 +393,11 @@ def regional_area_weighted_mean(
         keepdim: bool = False,
         name: str | None = None,
     ) -> torch.Tensor:
+        if self._is_local:
```
mcgibbon: Issue: low-level distributed logic outside of Distributed.

Similar thoughts to the other comments about this, except that the call to `dist.area_weighted_mean` would happen somewhere inside `metrics.weighted_mean`.

fme/core/gridded_ops.py
```diff
@@ -368,6 +412,11 @@ def area_weighted_gradient_magnitude_percent_diff(
         predicted: torch.Tensor,
         name: str | None = None,
     ):
+        if self._is_local:
```
mcgibbon: Issue: low-level distributed logic outside of Distributed.

Similar thoughts to the other comments about this, except that the call to `dist.area_weighted_mean` would happen somewhere inside `metrics.weighted_mean`.

fme/core/gridded_ops.py
```diff
@@ -376,22 +425,26 @@ def area_weighted_gradient_magnitude_percent_diff(
...
-    def get_real_sht(self) -> torch_harmonics.RealSHT:
-        return torch_harmonics.RealSHT(
-            nlat=self._cpu_area.shape[-2],
-            nlon=self._cpu_area.shape[-1],
+    def get_real_sht(self) -> nn.Module:
+        return self._dist.get_sht(
+            self._nlat,
+            self._nlon,
+            lmax=self._nlat,
+            mmax=self._nlon // 2 + 1,
```
mcgibbon: Is this change needed? If not we should remove it for this PR.
```suggestion
```

fme/core/gridded_ops.py
```diff
@@ -376,22 +425,26 @@ def area_weighted_gradient_magnitude_percent_diff(
...
-    def get_real_isht(self) -> torch_harmonics.InverseRealSHT:
-        return torch_harmonics.InverseRealSHT(
-            nlat=self._cpu_area.shape[-2],
-            nlon=self._cpu_area.shape[-1],
+    def get_real_isht(self) -> nn.Module:
+        return self._dist.get_isht(
+            self._nlat,
+            self._nlon,
+            lmax=self._nlat,
+            mmax=self._nlon // 2 + 1,
```
mcgibbon: Is this change needed? If not we should remove it for this PR.
```suggestion
```

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
         if not torch.allclose(area_weights, area_weights[..., :1]):
             raise ValueError(
                 "Area weights must be longitudinally uniform, "
                 "as assumed for zonal mean."
             )
-        self._device_area = area_weights.to(get_device())
-        self._cpu_area = area_weights.to("cpu")
-        self._device_mask_provider = mask_provider.to(get_device())
-        self._cpu_mask_provider = mask_provider.to("cpu")
-        self._grid = "legendre-gauss"
```
mcgibbon: Please avoid unrelated changes, It's taking me quite some time to review this file because of them.

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
         mask_provider: MaskProviderABC = NullMaskProvider,
         grid: str = "legendre-gauss",
     ):
-        # requires weights are longitudinally uniform
+        from fme.core.distributed.distributed import Distributed
+
+        self._dist = Distributed.get_instance()
+        if area_weights.ndim >= 2:
+            self._nlat = area_weights.shape[-2]
+            self._nlon = area_weights.shape[-1]
```
mcgibbon: Suggestion: It seems like we're duplicating information here, can we just use `self._cpu_area_global.shape` below instead and minimize changes?

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
+            h_slice, w_slice = self._dist.get_local_slices((self._nlat, self._nlon))
+            local_weights = area_weights[..., h_slice, w_slice]
+            self._is_local = local_weights.shape != area_weights.shape
+        else:
+            self._nlat = 0
+            self._nlon = 0
```
mcgibbon: Question: What is the meaning of nlat == nlon == 0?

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
+            local_weights = area_weights
+            self._is_local = False
```
mcgibbon: Question: How does area weights shape < 2 mean we're nonlocal?

fme/core/gridded_ops.py
```diff
@@ -287,20 +286,45 @@ def __init__(
...
+        from fme.core.distributed.distributed import Distributed
+
+        self._dist = Distributed.get_instance()
+        if area_weights.ndim >= 2:
+            self._nlat = area_weights.shape[-2]
+            self._nlon = area_weights.shape[-1]
+            h_slice, w_slice = self._dist.get_local_slices((self._nlat, self._nlon))
+            local_weights = area_weights[..., h_slice, w_slice]
+            self._is_local = local_weights.shape != area_weights.shape
+        else:
```
mcgibbon: Issue: Generally speaking I don't understand this if-else clause. What is it handling? Can you minimize the changes to make it more obvious?

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -167,16 +167,12 @@ def __init__(
...
-            self.modes_lat_local = self.inverse_transform.lmax_local
-            self.modes_lon_local = self.inverse_transform.mmax_local
-            self.lpad_local = self.inverse_transform.lpad_local
-            self.mpad_local = self.inverse_transform.mpad_local
-        else:
-            self.modes_lat_local = self.modes_lat
-            self.modes_lon_local = self.modes_lon
-            self.lpad = 0
-            self.mpad = 0
+        dist = Distributed.get_instance()
```
mcgibbon, line 172: [Makani does](https://github.com/NVIDIA/makani/blob/main/makani/models/common/spectral_convolution.py#L73C1-L82C58):
```
        if isinstance(self.inverse_transform, thd.DistributedInverseRealSHT):
            self.modes_lat_local = self.inverse_transform.l_shapes[comm.get_rank("h")]
            self.modes_lon_local = self.inverse_transform.m_shapes[comm.get_rank("w")]
            self.nlat_local = self.inverse_transform.lat_shapes[comm.get_rank("h")]
            self.nlon_local = self.inverse_transform.lon_shapes[comm.get_rank("w")]
        else:
            self.modes_lat_local = self.modes_lat
            self.modes_lon_local = self.modes_lon
            self.nlat_local = self.inverse_transform.nlat
            self.nlon_local = self.inverse_transform.nlon
```

I don't mind you keeping what you have here instead, it seems clear/obvious enough. Or maybe better is to revert this for now and store the full weight shape.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -586,6 +617,15 @@ def __init__(
             if hasattr(params, "img_shape_x") and hasattr(params, "img_shape_y")
             else img_shape
         )
+        self.local_img_shape = (
```
mcgibbon: Issue: `local_img_shape` is odd compared to the rest of the parameters, because everything describes the full network and is identical across ranks. This would introduce an init parameter that's rank-specific, which adds complexity I think we don't need.

Suggestion: Delete `local_img_shape`, and use `dist` to get the local shape in the places where you need it.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -844,7 +885,7 @@ def __init__(
         if normalize_big_skip:
             self.norm_big_skip = ConditionalLayerNorm(
                 in_chans,
-                img_shape=self.img_shape,
+                img_shape=self.local_img_shape,
```
mcgibbon: Comment: It may seem like handling the local shape conversion here is good because it avoided needing to update the ConditionalLayerNorm, but soon ConditionalLayerNorm is going to need to be distributed-aware, otherwise it can't have its weights stored and loaded properly. It's better that we uniformly treat all our class init args like they're on one rank, and have each class be responsible for its own distributed weight handling.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -877,9 +918,11 @@ def forward(self, x: torch.Tensor, context: Context):
         else:
             x = self.encoder(x)
 
-        if hasattr(self, "pos_embed"):
-            # old way of treating unequally shaped weights
-            x = x + self.pos_embed
+        if isinstance(getattr(self, "pos_embed", None), nn.Parameter):
```
mcgibbon: pre-existing issue but wow, can't believe we were using this as a bool, parameter, and unset.

nit (optional): Mostly because isinstance makes my hair stand on end, can you write this in a way that avoids it? Possibly by not setting `self.pos_embed` to a boolean, by setting `self.pos_embed` to None if it's unused, and doing an `if self.pos_embed is not None:` check? If not I'll refactor it later, no worries.

fme/core/step/test_step.py
```diff
@@ -474,12 +474,21 @@ def get_step(
 
 @pytest.mark.parallel
 def test_label_conditioned_step():
+    from fme.core.distributed.distributed import Distributed
+    from fme.core.distributed.parallel_tests.test_step import scatter_spatial
```
mcgibbon: Chore: Move imports to top of file.

I could have sworn we had a linter check that enforces this, we should look into it.

mcgibbon: You can leave this for another PR (and open an issue), but we try to avoid importing from other test files if we can avoid it (it can't always be avoided), it's a bit of a code smell. In this case, it's pointing at the fact that `scatter_spatial` should probably be a method on `dist`.

fme/core/distributed/parallel_tests/test_local_slices.py
```diff
@@ -46,18 +46,14 @@ def test_local_slices_subdivide_domain():
...
-            x_global[gathered_local_slices[i]] += 1.0
-        torch.testing.assert_close(
-            x_global, torch.ones_like(x_global)
-        )  # the entire domain should get selected, and only once
-    else:
-        assert gathered_local_slices is None
+    canvas[local_slices] += 1.0
+    # Combine spatial contributions then data-parallel contributions.
+    canvas = dist.spatial_reduce_sum(canvas)
+    canvas = dist.reduce_sum(canvas)
```
mcgibbon, line 54: Issue: this is no longer a unit test of `get_local_slices` because it also will be broken by errors in `spatial_reduce_sum` and `reduce_sum`.

Suggestion: Copy-paste the original test and make this a new test (possibly in a different file).

mcgibbon: Question: Was the existing test breaking?

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: You can ignore this comment now, anything relevant has been posted in another comment.

fme/core/distributed/parallel_tests/test_step.py
mcgibbon: Something smells wrong here. But for now, can you just move `gather_spatial` to a method on `Distributed` defined next to `gather_global`, to make this comparison more immediately apparent? Similarly for `scatter_spatial`. We can refactor it later.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -107,6 +108,8 @@ def __init__(
             self._device_id = self._local_rank
             torch.cuda.set_device(self._device_id)
 
+        self._thd_initialized = False
```
mcgibbon:
```suggestion
        self._thd_initialized = False
        self._ensure_thd_initialized()
```
Can we initialize on... init? Would suggest refactoring/cleaning up after this suggestion but this is the important part.

mcgibbon (commented):
If a comment is too large and you don't want to block the other changes, you can make issues that link to the comment(s) and have https://github.com/ai2cm/ace/issues/893 or one of its children as a parent issue.

fme/ace/registry/stochastic_sfno.py
mahf708: https://github.com/ai2cm/ace/issues/909#issue-4034722685

fme/core/distributed/distributed.py
mahf708: ~Separation of concerns: we get disco_conv_cls from a custom place for all backends (local s2convultions.py in src code) except for the ModelTorchDistributed where we get it from an external library (thd)~ I was wrong

fme/core/models/conditional_sfno/sfnonet.py
mahf708: ~Separation of concerns: we get disco_conv_cls from a custom place for all backends (local s2convultions.py in src code) except for the ModelTorchDistributed where we get it from an external library (thd)~ I was wrong, now this is done

fme/core/distributed/base.py
mahf708: The event where we are operating on 2D tensors without an explicit data dim (i.e., deep inside layers)

fme/core/distributed/parallel_tests/test_local_slices.py
mahf708: I think it is do with some edits in get_local_slices impl ... 

issue: https://github.com/ai2cm/ace/issues/914 (but feel free to push back on fixing it here)

fme/core/gridded_ops.py
mahf708: this was related to testing only, now fixed.

fme/core/gridded_ops.py
mahf708: hopefully the revised version makes better sense :)

mahf708:
In the latest commit 8867c6b, I focused on cleaning up gridded_ops and deferring slicing to forward only.

I also opened these issues to track unfinished business:
- https://github.com/ai2cm/ace/issues/916
- https://github.com/ai2cm/ace/issues/915
- https://github.com/ai2cm/ace/issues/914
- https://github.com/ai2cm/ace/issues/917

fme/core/distributed/model_torch_distributed.py
```diff
@@ -266,6 +304,56 @@ def barrier(self):
...
+        return thd.DistributedRealSHT(
+            nlat, nlon, lmax=lmax, mmax=mmax, grid=grid
+        ).float()
+
+    def get_isht(self, nlat, nlon, lmax, mmax, grid):
+        return thd.DistributedInverseRealSHT(
+            nlat, nlon, lmax=lmax, mmax=mmax, grid=grid
+        ).float()
+
+    def get_disco_conv_s2(self, *args, **kwargs) -> nn.Module:
```
mcgibbon, line 354: Issue (optional): lost documentation for arguments.

Solution is to take explicit args like in `get_sht` (or more likely as kwargs), you only would need to add the ones we use (we can add more if/when we need to use more of them).

Can be left to another PR or not done at all. No need to track in an issue.

fme/core/distributed/torch_distributed.py
```diff
@@ -164,6 +167,45 @@ def barrier(self):
...
+        return tensor
+
+    def weighted_mean(
+        self,
+        data: torch.Tensor,
+        weights: torch.Tensor,
+        dim: tuple[int, ...],
+        keepdim: bool = False,
+    ) -> torch.Tensor:
+        return metrics.weighted_mean(data, weights, dim=dim, keepdim=keepdim)
```
mcgibbon, line 203: Issue: Import of metrics from distributed. Instead, implement the logic that needs to be distributed under `fme/core/distributed`, and call that logic from metrics.

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -340,11 +335,15 @@ def forward(self, x, timer: Timer = NullTimer()):  # pragma: no cover
...
+        # Slice global weights to the local spectral partition.
+        weight_local = self.weight[:, self._l_slice]
         if self.lora_A is not None and self.lora_B is not None:
+            lora_A_local = self.lora_A[:, self._l_slice]
+            lora_B_local = self.lora_B[:, self._l_slice]
             with timer.child("lora_update"):
                 lora_update = _contract_lora(
-                    self.lora_A,
-                    self.lora_B,
+                    lora_A_local,
```
mcgibbon: We're not slicing these anymore, right? I think you need to revert this

fme/core/gridded_ops.py
```diff
@@ -369,29 +372,30 @@ def area_weighted_gradient_magnitude_percent_diff(
...
-            nlat=self._cpu_area.shape[-2],
-            nlon=self._cpu_area.shape[-1],
-            grid=self._grid,
-        ).to(get_device())
+    def get_real_sht(self) -> nn.Module:
+        nlat = self._cpu_area_global.shape[-2]
+        nlon = self._cpu_area_global.shape[-1]
+        return (
+            Distributed.get_instance()
+            .get_sht(nlat, nlon, lmax=nlat, mmax=nlon // 2 + 1, grid=self._grid)
```
mcgibbon:
```suggestion
            .get_sht(nlat, nlon, grid=self._grid)
```
We didn't override this before, is there a particular reason to override the default now?

fme/core/gridded_ops.py
```diff
@@ -369,29 +372,30 @@ def area_weighted_gradient_magnitude_percent_diff(
...
-            nlat=self._cpu_area.shape[-2],
-            nlon=self._cpu_area.shape[-1],
-            grid=self._grid,
-        ).to(get_device())
+    def get_real_isht(self) -> nn.Module:
+        nlat = self._cpu_area_global.shape[-2]
+        nlon = self._cpu_area_global.shape[-1]
+        return (
+            Distributed.get_instance()
+            .get_isht(nlat, nlon, lmax=nlat, mmax=nlon // 2 + 1, grid=self._grid)
```
mcgibbon:
```suggestion
            .get_isht(nlat, nlon, grid=self._grid)
```

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -169,27 +169,22 @@ def __init__(
...
-            self.modes_lon_local = self.modes_lon
-            self.lpad = 0
-            self.mpad = 0
+        dist = Distributed.get_instance()
+        l_slice, m_slice = dist.get_local_slices((self.modes_lat, self.modes_lon))
+        l_start, l_stop, _ = l_slice.indices(self.modes_lat)
+        m_start, m_stop, _ = m_slice.indices(self.modes_lon)
+        self.modes_lat_local = l_stop - l_start
+        self.modes_lon_local = m_stop - m_start
+        self._l_slice = l_slice
```
mcgibbon, line 176:
```suggestion
```
We're foregoing this for now, I think.

fme/core/models/conditional_sfno/s2convolutions.py
mahf708: done [cdbbb33](https://github.com/ai2cm/ace/pull/900/commits/cdbbb33e2e90ddd0fa6e19e25454ff0e88da4d4e)

fme/core/distributed/distributed.py
```diff
@@ -307,6 +307,11 @@ def gather_global(
         """
         Gathers tensor data into a single tensor with the data from all ranks.
 
+        Each rank places its local portion (determined by :meth:`get_local_slices`)
+        into a zero tensor of ``global_shape``.  Spatial chunks are combined via
+        :meth:`spatial_reduce_sum`, then the data-parallel portions are gathered
+        to root via :meth:`gather`.
```
mcgibbon, line 313: This was necessary because `get_local_slices` is broken in the case of the spatial parallel backend. I would suggest reverting to the previous implementation once that's fixed (i.e. soon), which uses one gather instead of a sum and then gather.

mahf708:
in cdbbb33, i accidentially removed the slicing for dhconv when I intended to remove them only for lora blocks; c24d64a minimally restores that to allow the tests to pass.

PR #903, opened 2026-03-04
Incorporate sea floor depth in `DepthCoordinate`

jpdunc23:
Adds `deptho` to `DepthCoordinate` to account for partial bottom cells in the depth integral.

Changes:
- Adds `dz_from_idepth` helper to compute the 1D (no `deptho`) or 3D (with `deptho`) layer thickness tensor for the depth integral.

- [x] Tests added


fme/core/coordinates.py
```diff
@@ -334,6 +334,19 @@ def vertical_integral(
...
 
 
+def dz_from_idepth(
+    idepth: torch.Tensor, deptho: torch.Tensor | None = None
+) -> torch.Tensor:
+    if deptho is not None:
+        z_top = idepth[..., :-1]
+        z_bot = idepth[..., 1:]
+        deptho = deptho.unsqueeze(-1)
+        dz = torch.clamp(deptho, min=z_top, max=z_bot) - z_top
```
spencerkclark, line 339: Clever.

PR #904, opened 2026-03-04
Deprecate `DepthCoordinate` mask providing

### LINKED ISSUE #902: Deprecate mask provider behavior of `DepthCoordinate`
Before `MaskProvider` was added, `DepthCoordinate` was the main mask provider. Except for very old Samudra checkpoints that we probably no longer need to support, the mask-provider-like API of `DepthCoordinate` is not used and can be safely removed.

jpdunc23:
Removes unused mask providing API from `DepthCoordinate`.

Changes:
- Simplified `HasOceanDepthIntegral` protocol for `OceanData`.

- [x] Tests added

Resolves #902


fme/core/dataset/xarray.py
elynnwu: [nit] should we change this to `name.startswith("mask_")` now that `surface_mask` is not an option?

fme/core/coordinates.py
```diff
@@ -372,27 +370,6 @@ def __len__(self):
...
         return len(self.idepth)
 
-    def get_mask(self) -> torch.Tensor:
-        return self.mask
-
-    def get_mask_level(self, level: int) -> torch.Tensor:
-        return self.mask.select(dim=-1, index=level)
-
-    def get_mask_tensor_for(self, name: str) -> torch.Tensor | None:
-        match = LEVEL_PATTERN.search(name)
```
elynnwu, line 382: I think you can also remove `LEVEL_PATTERN` in this file now since this was the only place that was using it

PR #905, opened 2026-03-05
Fix coupled prediction loader dist context

### LINKED ISSUE #899: Prediction loader currently broken
While running coupled evaluator with prediction loader, got
```sh
RuntimeError: Distributed.get_instance() called before entering context. Please use Distributed.context() to wrap the training process, or ensure that get_instance() is only called within the context.
```

Job: https://beaker.org/orgs/ai2/workspaces/ace/work/01KJV276RGDFA1H79Q78ATCEQJ?taskId=01KJV276YJAPQC59B4BHHGRJG3&jobId=01KJV2771ZYF5D1G99WAKNDAPJ

elynnwu:
We currently have two issues with prediction loader:
1. Duplicated forcing names in atmosphere/ocean -> we remove them from ocean if it's already in atmosphere
2. Distributed.get_instance() called before entering context -> we force coupled data loader worker to use the non distributed backend

Resolves #899


fme/coupled/data_loading/getters.py
```diff
@@ -40,6 +41,19 @@
 
 from .inference import ExplicitIndices
 
+_COUPLED_WORKER_DIST_CX = None
+
+
+def _coupled_forkserver_worker_init_fn(worker_id: int) -> None:
```
elynnwu, line 46: Here DataLoader worker processes are forced to non-distributed and can be different from the main process. I still don't totally understand why we only get this error when we turn on `prediction_loader`. And I think this might be wrong if we run on more than 1 rank. 
@mcgibbon do you mind taking a look?

mcgibbon: Do you need to force them to non-distributed? I didn't do that for the similar worker init function in ace's data loading: https://github.com/E3SM-Project/ace/blob/ff717b57dec39a11a2333d3226bcc068fa653a2d/fme/ace/data_loading/getters.py#L141

It's because the inference data loading accesses `dist` while any other data loading does not.

fme/coupled/stepper.py
```diff
@@ -393,7 +393,8 @@ def all_names(self) -> CoupledNames:
         ocean_names = (
             self.ocean.stepper.output_names + self._ocean_forcing_exogenous_names
         )
-        return CoupledNames(ocean=ocean_names, atmosphere=atmosphere_names)
+        ocean_names_no_shared = list(set(ocean_names).difference(atmosphere_names))
+        return CoupledNames(ocean=ocean_names_no_shared, atmosphere=atmosphere_names)
```
mcgibbon: Is this a related change? I can't really review this / tell you if it's right or not.

mcgibbon:
My suggestion would be to remove the force-non-distributed environment call (unless it's needed) and do what we did for the ace inference data loading, and separate the variable name stuff into another PR.

fme/coupled/data_loading/getters.py
elynnwu: Thanks. Looking at the comment in ace/data_loading again, it clearly says what needed to be done, somehow I missed it. Updated

fme/coupled/stepper.py
elynnwu: Removed and will open in a separate PR

PR #906, opened 2026-03-05
Fix SongUNetv2 benchmark OOM

AnnaKwa:
- Reduces the size of the default non-apex `songunetv2, songunetv2_bf16` benchmarks
- Adds --ignore CLI flag  to `fme.core.benchmark.run` so the workflow tests yaml can ignore the apex tests (no apex in testing environment)

I tested that the new benchmark configs run on a T4 which is what the failing tests used.

### DIFF
--- a/fme/fft.py
+++ b/fme/fft.py
@@ -1,0 +1,96 @@
+# flake8: noqa
+# fmt: off
+# isort: skip_file
+
+"""
+We forked this code from torch-harmonics [*].
+
+[*] https://github.com/NVIDIA/torch-harmonics/blob/0d08559a30d292164e1fa64dd3daeeff9b8efac9/torch_harmonics/fft.py
+"""
+
+# coding=utf-8
+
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -135,4 +135,4 @@ jobs:
             export PATH=/usr/local/nvidia/bin:${PATH}
             export LD_LIBRARY_PATH=/usr/local/nvidia/lib64:${LD_LIBRARY_PATH}
             export TORCH_ALLOW_TF32_CUBLAS_OVERRIDE=1
-            python3 -m fme.core.benchmark.run --wandb-project ai2cm/fme-core-benchmarks-ci
+            python3 -m fme.core.benchmark.run --wandb-project ai2cm/fme-core-benchmarks-ci --ignore songunetv2_apex songunetv2_apex_bf16
```
mcgibbon, line 138: I assume these are ignored because using apex requires a model size that causes OOM?

AnnaKwa: Yes, and also apex is not installed in this test environment so an error would get raised.

PR #910, opened 2026-03-06
Record initial condition times upon data writer initialization

### LINKED ISSUE #886: Inaccurate `"init_time"` and `"time"` coordinates in outputs from raw data writer
I would expect the `"init_time"` coordinate in outputs from the raw data writer to correspond to the times of the initial conditions of the runs in the ensemble.  Instead, however, it corresponds to the time of the first predicted step.  This bleeds into the lead `"time"` coordinate as well, in that the first outputs are assigned a lead time of `"0 days 00:00:00"`, when in reality they are the first step in the run, so they should have a lead time of `"0 days 06:00:00"`, i.e. six hours.

See the following outputs from [this example short evaluator run](https://beaker.org/orgs/ai2/workspaces/ace/work/01KJQFTEY11FAX3QKVYW6PAFSR?taskId=01KJQFTF4DXEEMKVN1ZSF91JA1&jobId=01KJQFTFCAG27NK5W1A435E9TS). The initial condition time in the configuration file is given as `"1940-01-01T12:00:00"`, but the `"init_time"` in the output `autoregressive_predictions.nc` file is `"1940-01-01T18:00:00"` and the first `"time"` coordinate is `"0 days 00:00:00"`:
```
>>> xr.open_dataset("autoregressive_predictions.nc", decode_timedelta=True).load()
<xarray.Dataset> Size: 3MB
Dimensions:     (sample: 1, time: 12, lat: 180, lon: 360)
Coordinates:
    init_time   (sample) datetime64[ns] 8B 1940-01-01T18:00:00
  * time        (time) timedelta64[ns] 96B 0 days 00:00:00 ... 2 days 18:00:00
    valid_time  (sample, time) datetime64[ns] 96B 1940-01-01T18:00:00 ... 194...
  * lat         (lat) float32 720B -89.24 -88.25 -87.25 ... 87.25 88.25 89.24
  * lon         (lon) float32 1kB 0.5 1.5 2.5 3.5 ... 356.5 357.5 358.5 359.5
Dimensions without coordinates: sample
Data variables:
    TMP2m       (sample, time, lat, lon) float32 3MB 242.4 242.4 ... 245.3 245.3
Attributes:
    source.git_sha:   7b9438a
    source.job_id:    5g382v6y
    history.created:  2026-03-02T14:43:35.314490
    title:            ACE autoregressive predictions data file
```
The `"valid_time"` coordinate is correct, however.

spencerkclark:
This PR fixes #886 by requiring that we explicitly pass the times associated with the initial conditions when initializing data writers, rather than just passing the number of initial conditions and attempting to (erroneously) infer them from the first predicted batch.

Changes:
- Switch from requiring `n_initial_conditions` to the more information-rich `initial_condition_times` in the netCDF and Zarr data writer build methods.
- The `ZarrWriterAdapter` and `SeparateICZarrWriterAdapter` classes now both require passing the `timestep` at initialization.
- Time outputs from the raw netCDF data writer will now have units of `"microseconds since 1970-01-01"` instead of the equivalent-meaning-but-more-verbose `"microseconds since 1970-01-01 00:00:00"`.  This harmonizes their value with what is used in the zarr data writer.

- [x] Tests added

Resolves #886

### DIFF
--- a/fme/ace/inference/data_writer/test_file_writer.py
+++ b/fme/ace/inference/data_writer/test_file_writer.py
@@ -617,6 +617,13 @@ def test_netcdf_file_writer_with_non_local_experiment_dir(
 
 
 @pytest.mark.parametrize(
+    "n_timesteps",
+    [
+        pytest.param(2, id="steps-in-memory-equal-to-coarsen-factor"),
+        pytest.param(4, id="steps-in-memory-greater-than-coarsen-factor"),
+    ],
+)
+@pytest.mark.parametrize(
     "format",
     [
         pytest.param(NetCDFWriterConfig(), id="netCDF"),
@@ -624,11 +631,10 @@ def test_netcdf_file_writer_with_non_local_experiment_dir(
     ],
 )
 def test_coarsened_file_writer_time_coordinates(
-    tmpdir, format: NetCDFWriterConfig | ZarrWriterConfig
+    tmpdir, n_timesteps: int, format: NetCDFWriterConfig | ZarrWriterConfig
 ):
     initial_condition_times = np.array([cftime.DatetimeGregorian(2020, 1, 1)])
     n_initial_conditions = len(initial_condition_times)
-    n_timesteps = 4
     coarsen_factor = 2
     timestep = datetime.timedelta(days=1)
 
@@ -667,14 +673,17 @@ def test_coarsened_file_writer_time_coordinates(
     expected_init_time = xr.DataArray(
         [cftime.DatetimeGregorian(2020, 1, 1)], dims="sample"
     )
-    expected_valid_time = xr.DataArray(
-        [
+    if n_timesteps == 2:
+        expected_valid_time_values = [[cftime.DatetimeGregorian(2020, 1, 2, 12, 0, 0)]]
+    elif n_timesteps == 4:
+        expected_valid_time_values = [
             [
                 cftime.DatetimeGregorian(2020, 1, 2, 12, 0, 0),
                 cftime.DatetimeGregorian(2020, 1, 4, 12, 0, 0),
             ]
-        ],
-        dims=["sample", "time"],
+        ]
+    expected_valid_time = xr.DataArray(
+        expected_valid_time_values, dims=["sample", "time"]
     )
     expected_valid_time = expected_valid_time.assign_coords(
         time=expected_lead_times,


fme/ace/inference/data_writer/raw.py
```diff
@@ -138,8 +141,15 @@ def __init__(
         self.dataset.createDimension(IC_DIM, n_initial_conditions)
         self.dataset.createVariable(INIT_TIME, "i8", (IC_DIM,))
         self.dataset.variables[INIT_TIME].units = INIT_TIME_UNITS
```
jpdunc23: Can this be replace with `fme.core.writer.DATETIME_ENCODING_UNITS`? They seem to be slightly different, but would be nice to consolidate:

```
INIT_TIME_UNITS = "microseconds since 1970-01-01 00:00:00"
DATETIME_ENCODING_UNITS = " "microseconds since 1970-01-01"
```

Similarly, maybe `LEAD_TIME_DIM` can be replaced with `fme.core.writer.TIMEDELTA_ENCODING_UNITS`.

fme/ace/inference/data_writer/raw.py
```diff
@@ -314,3 +300,18 @@ def get_batch_lead_time_microseconds(
     if not np.all(lead_time_microseconds == lead_time_microseconds[0, :]):
         raise ValueError("Lead times are not the same for each sample in the batch.")
     return lead_time_microseconds[0, :]
+
+
+def infer_calendar(array: npt.NDArray[cftime.datetime]) -> str:
```
jpdunc23: Nit: make private to the module i.e. `_infer_calendar`.

fme/coupled/data_loading/gridded_data.py
```diff
@@ -247,3 +250,21 @@ def horizontal_coordinates(self) -> CoupledHorizontalCoordinates:
...
+                np.testing.assert_array_equal(
+                    atmosphere_initial_time,
+                    ocean_initial_time,
+                    err_msg=err_msg,
+                )
+                self._initial_time = atmosphere_initial_time
+                break
+            else:
+                raise ValueError("No data found in loader")
+        return self._initial_time
```
jpdunc23: Setting `self._initial_time` seems somewhat related to what the helper `get_initial_condition` does. Maybe the two could be combined somehow.

Also, does this properly handle the case where the `initial_condition` passed on init has type `CoupledPrognosticState`? I think yes since `get_forcing_data` in fme/coupled/data_loading/getters.py does the same time slicing to get the initial condition, but we should avoid that assumption by just directly getting the initial time from the `CoupledPrognosticState`.

fme/ace/inference/data_writer/zarr.py
```diff
@@ -30,38 +34,56 @@ def _variable_metadata_to_dict(
...
+    # timestep whether the batch includes multiple times or just one. This
+    # approach is safe, because we currently do not support time subselection
+    # when using the zarr writer. A more robust approach would be to pass the
+    # timestep to the writer at initialization time, but that is maybe better
+    # saved for a future refactor, since it requires a little extra care around
+    # time coarsening.
+    times_with_prepended_initial_condition = np.insert(
+        batch_time.isel(sample=0).to_numpy(), 0, initial_condition_times[0]
+    )
+    dt_timedelta = _get_timestep(times_with_prepended_initial_condition)
```
jpdunc23: Nit: Maybe this is a good opportunity to rename this to `get_timestep` to acknowledge the preexisting non-private usage.

fme/ace/inference/data_writer/test_data_writer.py
```diff
@@ -35,6 +35,36 @@
 TIMESTEP = datetime.timedelta(hours=6)
 
 
+def get_initial_condition_times(
+    batch_start_time: tuple[int, int, int, int, int, int],
+    calendar: str,
+    n_initial_conditions: int,
+    separation_timedelta: datetime.timedelta = datetime.timedelta(hours=0),
+    model_timestep: datetime.timedelta = TIMESTEP,
+) -> np.ndarray[cftime.datetime]:
```
jpdunc23: Nit: Should this also be `npt.NDArray[cftime.datetime]`?

fme/ace/inference/data_writer/zarr.py
```diff
@@ -30,38 +34,56 @@ def _variable_metadata_to_dict(
...
-            attrs={"units": "microseconds", "dtype": "timedelta64[us]"},
-        )
-    elif n_timesteps == 1:
-        lead_times_coord = xr.DataArray(
-            np.array([0], dtype=np.int64),
-            dims=["time"],
-            attrs={"units": "microseconds", "dtype": "timedelta64[us]"},
-        )
-    init_times_datetime = batch_time.isel(time=0).values
+def _get_encoded_lead_times(
```
jpdunc23, line 36: Is there a subtle bug when combining `ZarrWriterConfig` and `TimeCoarsenConfig` with `forward_steps_in_memory` equal to `coarsen_factor`? E.g.:

Say the base timestep is 6h, IC is at time 0, and `coarsen_factor=2`. The batch has 2 original timesteps that get coarsened to 1:
- Original batch_time: [6h, 12h]
- After `coarsen_batch()` (block mean): [mean(6h, 12h)] = [9h]
- `_get_encoded_lead_times` prepends IC: [0h, 9h] -- 2 elements, `_get_timestep` passes
- Inferred dt = 9h
- lead_times = 9h * [1, 2, 3, 4] = [9h, 18h, 27h, 36h]

But the correct coarsened lead times should be:

- Coarsened windows: mean(6h,12h)=9h, mean(18h,24h)=21h, mean(30h,36h)=33h, mean(42h,48h)=45h
- Actual spacing: 12h (= `coarsen_factor` * 6h)
- Correct lead times: [9h, 21h, 33h, 45h]

spencerkclark: Indeed thanks for bringing this up—I had forgotten we label time-coarsened output with interval centered times (which is different than what we do with input data).  I'll fix this and address your other comments on Monday.

fme/ace/inference/data_writer/raw.py
spencerkclark: Yes, maybe this is a reasonable time to harmonize these. Switched to using `DATETIME_ENCODING_UNITS` and `TIMEDELTA_ENCODING_UNITS` throughout `raw.py` in bfc00d7.

fme/ace/inference/data_writer/raw.py
spencerkclark: Done.

fme/ace/inference/data_writer/test_data_writer.py
spencerkclark: Thanks, fixed—bad Cursor autocomplete...

fme/ace/inference/data_writer/zarr.py
spencerkclark: I added a failing test in d720af7 and then fixed it in 7680bb4.

fme/coupled/data_loading/gridded_data.py
spencerkclark: Indeed I mostly just followed the uncoupled implementation here: https://github.com/ai2cm/ace/blob/4746f518a74c800c6e6036465340bdcafdab1deb/fme/ace/data_loading/gridded_data.py#L244-L252

I refactored both to use the `InferenceGriddedData.initial_condition` directly in 5564bc6.  This exposed a minor bug in the segmented run test, in that the initial condition and forcing data used different calendars.

fme/ace/inference/data_writer/zarr.py
spencerkclark: This is moot now that I pass the `timestep` directly to this function.

fme/ace/inference/data_writer/zarr.py
```diff
@@ -30,38 +33,56 @@ def _variable_metadata_to_dict(
...
+    # context we will infer it from the difference between the first batch time
+    # and the initial condition time. All subsequent lead times will be coarse
+    # timestep increments on top of that. This is safe to do, because we do not
+    # support time subselection when using the zarr writer.
+    first_lead_time = (
+        batch_time.isel(sample=0, time=0).item() - initial_condition_times[0]
+    )
+    subsequent_lead_times = first_lead_time + timestep * np.arange(1, n_timesteps)
+    lead_times = np.concatenate([[first_lead_time], subsequent_lead_times])
+    return lead_times.astype("timedelta64[us]").astype(np.int64)
```
jpdunc23, line 54: Nice fix!

fme/ace/inference/data_writer/test_file_writer.py
```diff
@@ -592,13 +603,96 @@ def test_netcdf_file_writer_with_non_local_experiment_dir(
...
+        pytest.param(ZarrWriterConfig(), id="Zarr"),
+    ],
+)
+def test_coarsened_file_writer_time_coordinates(
+    tmpdir, format: NetCDFWriterConfig | ZarrWriterConfig
+):
+    initial_condition_times = np.array([cftime.DatetimeGregorian(2020, 1, 1)])
+    n_initial_conditions = len(initial_condition_times)
+    n_timesteps = 4
+    coarsen_factor = 2
```
jpdunc23: It would be good to add an additional case where `n_timesteps == coarsen_factor` to cover the (now fixed) silent bug when `n_forward_steps_in_memory == coarsen_factor`.

spencerkclark: Thanks, that's a good idea—I added a test of this kind in 2adf7b3.  I'll go ahead now and merge.

PR #912, opened 2026-03-06
Allow `WANDB_NOTES` for non-beaker jobs

elynnwu:
When training outside beaker, we normally log the output directory and some basic info such as the commit used for training that job. This was set in `WANDB_NOTES`, but a recent refactor caused it to be override by beaker url behavior. This PR makes it possible to set `WANDB_NOTES`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -116,6 +116,15 @@ def pytest_runtest_call(item):
 
 
 @pytest.fixture(autouse=True)
+def reset_global_timer():
+    import fme.core.timing
+
+    fme.core.timing.singleton = None
+    yield
+    fme.core.timing.singleton = None
+
+
+@pytest.fixture(autouse=True)
 def mock_gc_collect(monkeypatch):
     def mock_collect(*args, **kwargs):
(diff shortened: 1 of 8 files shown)


fme/core/logging_utils.py
```diff
@@ -144,14 +144,15 @@ def _configure_wandb(
         # must ensure wandb.configure is called before wandb.init
         wandb = WandB.get_instance()
         wandb.configure(log_to_wandb=self.log_to_wandb)
+        notes = _get_beaker_url(_get_beaker_id())
```
spencerkclark: Should we split off a separate `_get_wandb_notes` function, since `_get_beaker_url` is also used in another place that seems more beaker-centric?

https://github.com/ai2cm/ace/blob/14007d459bfea67ed936e1399fbfe15c8d42bd4e/fme/core/logging_utils.py#L205-L208

elynnwu: Good catch, updated to split off `_get_wandb_notes`

PR #919, opened 2026-03-06
Use lock to run benchmark test in serial

mcgibbon:
A benchmark test is currently flaking due to being run in parallel with other tests (noisy neighbors). This PR adds a `@pytest.mark.serial` to specify that a test should only be run on its own with no other tests, using a read-write lock.

Changes:
- Tests now use the `fasteners` package (a [pure-python](https://github.com/harlowja/fasteners/blob/main/pyproject.toml) package)
- Resolved flaky benchmark test failure in `fme/core/models/conditional_sfno/test_sfnonet.py` due to xdist causing noisy timings.
- Added "serial" marker to pytest config which, if specified, requires no other tests run at the same time as other tests even if using xdist. 

- [ ] Tests added


mcgibbon:
Apparently mock-type tests don't need to be marked, because xdist runs tests in separate processes. Only tests that modify/are sensitive to the base system (e.g. modifying environment variables) matter. Claude suggested the following tests to mark as they affect global state:

```
● All good. Here's a summary of what I added:

  4 tests marked @pytest.mark.serial:

  ┌─────────────────────────────────────────────┬────────────────────────────┬────────────────────────────────────────────────────┐
  │                    File                     │            Test            │                       Reason                       │
  ├─────────────────────────────────────────────┼────────────────────────────┼────────────────────────────────────────────────────┤
  │                                             │                            │ Spawns processes via mp.spawn that bind to         │
  │ fme/core/distributed/test_distributed.py    │ test_distributed_gather    │ hardcoded port 12355 and set os.environ for        │
  │                                             │                            │ distributed training                               │
  ├─────────────────────────────────────────────┼────────────────────────────┼────────────────────────────────────────────────────┤
  │ fme/ace/test_train.py                       │ test_resume_two_workers    │ Runs torchrun subprocesses that bind distributed   │
  │                                             │                            │ ports                                              │
  ├─────────────────────────────────────────────┼────────────────────────────┼────────────────────────────────────────────────────┤
  │ fme/downscaling/test_train.py               │ test_resume_two_workers    │ Runs torchrun subprocesses that bind distributed   │
  │                                             │                            │ ports                                              │
  ├─────────────────────────────────────────────┼────────────────────────────┼────────────────────────────────────────────────────┤
  │                                             │                            │ Runs torchrun and sets                             │
  │ fme/downscaling/inference/test_inference.py │ test_generation_entrypoint │ os.environ["TORCH_DISTRIBUTED_DEBUG"] without      │
  │                                             │                            │ cleanup                                            │
  └─────────────────────────────────────────────┴────────────────────────────┴────────────────────────────────────────────────────┘

  Why other candidates were excluded:
  - GlobalTimer tests (fme/core/test_timing.py) — already handled by the reset_global_timer autouse fixture in conftest.py that
  resets the singleton before and after every test
  - mock_wandb() / mock_distributed() users — these context managers properly save/restore the module-level singleton, and within a
  single xdist worker tests run sequentially
  - test_force_non_distributed — uses Distributed.force_non_distributed() context manager which properly restores state; singleton is
   process-local so no cross-worker risk
```

fme/ace/test_train.py
```diff
@@ -888,6 +888,7 @@ def test_restore_checkpoint(
             )
 
 
+@pytest.mark.serial
```
oliverwm1, line 891: are these tests now just skipped in CI? (since we use pytest-xdist there)

mcgibbon: No, the serial mark is only used by the new `_serialize_when_needed` fixture. This fixture communicates a lock file across xdist workers so only one can run at a given time. A lock is allowed to be held by multiple readers, but for a write lock to be held there must be no readers.

I could rename the mark if "serial" is too close to "parallel", given they have fairly different (and not opposite) meanings.

mcgibbon: More explicitly, serial tests still run when xdist is active, they just run in serial even if xdist is active.

oliverwm1: Got it, thanks

PR #921, opened 2026-03-07
ClosedInterval slice from coordinates

frodre:
This is the first small task of the StaticInputs refactor that adds some functionality to `ClosedInterval` simplifying how we slice coordinates.  `ClosedInterval.slice_of(coords)` is used to generate a slice given the provided coordinates, which can be used to subset the coordinates and the field data in a batch.  I'm planning on using this to match the static input region with the current batch within the model.

Changes:
- add `slice_of` method to `ClosedInterval`
- add string representation and docstring `ClosedInterval`
- Replace fancy-index subsetting in `HorizontalSubsetDataset` (datasets.py) with slice-based subsetting using `slice_of`
- `ClosedInterval` now errors with `ValueError` if it would generate an empty slice, which we used check in the horizontal subsetting in `datasets.py` as an `AssertionError`

- [x] Tests added


fme/downscaling/inference/test_output.py
```diff
@@ -113,8 +113,8 @@ def test_event_config_build_creates_output_target_with_single_time(
         event_time="2000-01-01T00:00:00",
         n_ens=4,
         save_vars=["var0", "var1"],
-        lat_extent=ClosedInterval(2.0, 6.0),
-        lon_extent=ClosedInterval(2.0, 6.0),
+        lat_extent=ClosedInterval(0.0, 6.0),
```
frodre, line 116: The `ClosedInterval` class now checks that the resulting slice is non-empty. Expanding the range so it covers the values used in the coordinate tensors.

fme/downscaling/data/utils.py
```diff
@@ -27,6 +37,16 @@ def __post_init__(self):
     def __contains__(self, value: float):
         return self.start <= value <= self.stop
 
+    def slice_of(self, coords: torch.Tensor) -> slice:
+        mask = (coords >= self.start) & (coords <= self.stop)
```
AnnaKwa, line 57: Can you add a docstring that explains the returned slice? Maybe also a note that it is assumed the `ClosedInterval.start, .stop` are same units as the coordinate tensor.

PR #923, opened 2026-03-07
Rename topography utilities to reflect general static input usage

frodre:
This PR adjusts the naming for a few leftover items that were still tailored for "topography" only operation on static inputs.

Changes:
- Rename `topography.py` → `static.py`
- Rename `get_normalized_topography` → `get_normalized_static_input` with parameter `topography_name` → `field_name`
- `field_name` is now a positional argument that gets specified by the current data loading topography injectio pathway
- Remove `get_topography_downscale_factor` and replace with `_check_fine_res_static_input_compatibility` helper, which enforces exact shape matching (the previous function computed a downscale factor that was only ever asserted to be 1)

### DIFF
--- a/fme/downscaling/data/static.py
+++ b/fme/downscaling/data/static.py
@@ -22,7 +22,7 @@ class StaticInput:
             self.coords.lon
         ):
             raise ValueError(
-                f"Topography data shape {self.data.shape} does not match lat/lon "
+                f"Static inputs data shape {self.data.shape} does not match lat/lon "
                 f"coordinates shape {(len(self.coords.lat), len(self.coords.lon))}"
             )
 


fme/downscaling/data/static.py
```diff
@@ -22,7 +22,7 @@ def __post_init__(self):
             self.coords.lon
         ):
             raise ValueError(
-                f"Topography data shape {self.data.shape} does not match "
+                f"Topography data shape {self.data.shape} does not match lat/lon "
```
AnnaKwa:
```suggestion
                f"Static inputs data shape {self.data.shape} does not match lat/lon "
```

fme/downscaling/data/config.py
AnnaKwa: Can also rename `fine_topography` variable defined here

PR #925, opened 2026-03-07
Add `SurfaceEnergyFluxCorrectionConfig`

jpdunc23:
Adds `surface_energy_flux_correction: SurfaceEnergyFluxCorrectionConfig | None = None` to `OceanCorrectorConfig` to correct the generated `hfds` or `hfds_total_area`, with two options:
- "residual_prediction": `corrected_hfds = gen_hfds + ocean_fraction * net_flux`
- "prescribed": `corrected_hfds = net_flux * ocean_fraction + gen_hfds * (1 - ocean_fraction)`

Changes:
- Improves naming of ocean component constants: `SPECIFIC_HEAT_OF_SEA_WATER_CM4` and `DENSITY_OF_SEA_WATER_CM4`.

- [x] Tests added


fme/core/corrector/ocean.py
```diff
@@ -160,6 +197,69 @@ def __call__(
...
+    )  # missing: - calving * LATENT_HEAT_OF_FREEZING
+    mass_heat_flux = (
+        SPECIFIC_HEAT_OF_SEA_WATER_CM4
+        * (
+            atmos.precipitation_rate
+            + atmos.frozen_precipitation_rate
+            - (atmos.latent_heat_flux / LATENT_HEAT_OF_VAPORIZATION)
+        )  # missing: + river runoff + calving
+        * (sst - FREEZING_TEMPERATURE_KELVIN)
+    )
```
jpdunc23, line 222: Originally (and in existing experimental runs) I didn't include `frozen_precipitation_rate` here, but including it provides a better match to `hfds`. See [this notebook](https://github.com/ai2cm/explore2/blob/main/jamesd/2026-02-05-CM4-surface-heat-flux/2026-03-13-frozen-precip-mass-transfer-heat-flux.ipynb).

spencerkclark: Nice, thanks for the notebook / link to the `FMSCoupler` code.

fme/core/constants.py
```diff
@@ -12,8 +12,8 @@
 
 SPECIFIC_HEAT_OF_DRY_AIR_CONST_VOLUME = SPECIFIC_HEAT_OF_DRY_AIR_CONST_PRESSURE - RDGAS
 
-SPECIFIC_HEAT_OF_WATER_CM4 = 3992.0  # J/kg/K
-DENSITY_OF_WATER_CM4 = 1035.0  # kg/m^3
+SPECIFIC_HEAT_OF_SEA_WATER_CM4 = 3989.24495292815  # J/kg/K
```
spencerkclark: I see you are doing this to be consistent with what is in FMS; do you know whether [the value in MOM6](https://github.com/NOAA-GFDL/MOM6/blob/22890625dcd3c1cd229f8d2d5494e5c319a52144/src/core/MOM.F90#L2643-L2647) would be more accurate?

fme/core/corrector/ocean.py
```diff
@@ -160,6 +197,69 @@ def __call__(
...
+    out = dict(gen_data)
+    if "hfds" in gen_data:
+        hfds_name = "hfds"
+    else:
+        hfds_name = "hfds_total_area"
+    gen_hfds = gen_data[hfds_name]
+    if method == "residual_prediction":
+        out[hfds_name] = net_flux * ocean_fraction + gen_hfds
+    elif method == "prescribed":
+        out[hfds_name] = net_flux * ocean_fraction + gen_hfds * (1 - ocean_fraction)
```
spencerkclark, line 257: In both of these cases, `if hfds_name == "hfds_total_area"` would it be more consistent to also scale `net_flux` by the sea surface fraction, similar to [how `"hfds_total_area"` was originally constructed](https://github.com/ai2cm/ace/blob/3246f8e7048f0f0106aab44abfd85ab888274cbb/scripts/data_process/coupled_dataset_utils.py#L637)?

fme/core/constants.py
jpdunc23: I don't, but probably good to check before making this change.

jpdunc23: Here's a comparison [notebook](https://github.com/ai2cm/explore2/blob/main/jamesd/2026-01-20-samudra-ohc-corrector/2026-03-16-1pctCO2-FMS_CP_OCEAN-vs-MOM6_C_P.ipynb).

The difference is quite small, but the value from FMS seems to be slightly more accurate. What do you think, does this justify the change?

At some point I wonder if it makes most sense to have these sort of constants embedded into the data itself.

spencerkclark: Thanks for looking into this more carefully; not surprising the difference is small. Upon a further look back at the history, [since these values originally came from Alistair](https://github.com/ai2cm/full-model/pull/1566#issuecomment-2672701936), I think we should likely stick with what we have rather than make a change. Even if the FMS value provides a slightly better budget closure, it might be for the wrong reason.

> At some point I wonder if it makes most sense to have these sort of constants embedded into the data itself.

Potentially, if we find these differ meaningfully between different models we are emulating.

jpdunc23: Good catch, thanks.

PR #926, opened 2026-03-09
Deprecate downscaling data configuration topography pathway

frodre:
The `topography` field on `DataLoaderConfig` and `PairedDataLoaderConfig` is a legacy pathway for loading what are now `StaticInputs`. This PR deprecates that field on both configs by raising a `ValueError` at construction time if it is set. While this is a breaking change for older configurations, it clarifies usage for any models to be trained before we eventually drop the field entirely.

Changes:
- `DataLoaderConfig`, `PairedDataLoaderConfig`: raise `ValueError` in `__post_init__` if `topography` is set, with a message explaining the replacement
- `PairedDataLoaderConfig.build()`: removed the old topography auto-loading fallback (loading `HGTsfc` from fine data files or `topography` path); now raises `ValueError` if `use_fine_topography=True` and no `static_inputs` are provided
- `CheckpointModelConfig`: added `static_inputs: dict[str, str] | None` field for specifying static inputs at evaluation time; deprecated `fine_topography_path` with a `DeprecationWarning` that maps it to `static_inputs`
- `load_static_inputs`: new helper exported from `fme.downscaling.data` that constructs `StaticInputs` from a `dict[str, str]` mapping field names to file paths; replaces direct use of `get_normalized_static_input` at call sites
- Remove `get_normalized_static_input` from the `fme.downscaling.data` top-level export in favor of `load_static_inputs`
- Remove `topography` usage from tests and configurations

- [x] Tests added

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -500,7 +500,7 @@ class DiffusionModel:
 
     def get_state(self) -> Mapping[str, Any]:
         if self.static_inputs is not None:
-            static_inputs_state = self.static_inputs.to_state()
+            static_inputs_state = self.static_inputs.get_state()
         else:
             static_inputs_state = None
 
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -1263,7 +1263,7 @@ class Stepper:
         """
         return {
             "config": self._config.as_loaded_dict(),
-            "dataset_info": self._dataset_info.to_state(),
+            "dataset_info": self._dataset_info.get_state(),
             "step": self._step_obj.get_state(),
             "training_history": self._training_history.get_state(),
         }
--- a/fme/core/coordinates.py
+++ b/fme/core/coordinates.py
@@ -692,7 +692,7 @@ class HorizontalCoordinates(abc.ABC):
         pass
 
     @abc.abstractmethod
-    def to_state(self) -> TensorMapping:
+    def get_state(self) -> TensorMapping:
         pass
 
 
@@ -786,7 +786,7 @@ class LatLonCoordinates(HorizontalCoordinates):
     def shape(self) -> tuple[int, int]:
         return (len(self.lat), len(self.lon))
 
-    def to_state(self) -> TensorMapping:
+    def get_state(self) -> TensorMapping:
         return {"lat": self.lat, "lon": self.lon}
 
 
@@ -937,7 +937,7 @@ class HEALPixCoordinates(HorizontalCoordinates):
     def shape(self) -> tuple[int, int, int]:
         return (len(self.face), len(self.width), len(self.height))
 
-    def to_state(self) -> TensorMapping:
+    def get_state(self) -> TensorMapping:
         return {"face": self.face, "height": self.height, "width": self.width}
 
 
--- a/fme/core/dataset_info.py
+++ b/fme/core/dataset_info.py
@@ -216,9 +216,9 @@ class DatasetInfo:
             all_labels=self._all_labels,
         )
 
-    def to_state(self) -> dict[str, Any]:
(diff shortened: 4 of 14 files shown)


fme/downscaling/test_predict.py
```diff
@@ -109,43 +106,22 @@ def test_predictor_runs(tmp_path, very_fast_only: bool):
...
+        downscale_factor=downscale_factor,
+        static_inputs=load_static_inputs({"HGTsfc": fine_data_path}),
     )
     with open(predictor_config_path) as f:
         predictor_config = yaml.safe_load(f)
     os.makedirs(
         os.path.join(predictor_config["experiment_dir"], "checkpoints"), exist_ok=True
     )
 
-    # ensure model static inputs shape is consistent with the test data
```
frodre, line 126: No longer needed since topography cannot be provided through data loader.

fme/downscaling/models.py
```diff
@@ -568,6 +575,17 @@ def __post_init__(self) -> None:
         self._rename = self.rename or {}
         if "module" in (self.model_updates or {}):
             raise ValueError("'module' cannot be updated in model_updates.")
+        if self.fine_topography_path is not None:
+            warnings.warn(
```
frodre: Opted for a warning instead of error because I didn't want to break recent evaluation configurations.

fme/downscaling/data/test_config.py
```diff
@@ -111,7 +110,7 @@ def test_PairedDataLoaderConfig_sample_with_replacement(tmp_path):
         fine_names=["var0"],
         coarse_names=["var0"],
         n_timesteps=1,
-        use_fine_topography=True,
+        use_fine_topography=False,
```
frodre, line 113: The DataRequirements handling of topography will be addressed in follow-ons after all of the passing between different data/model objects is removed.

fme/downscaling/inference/conftest.py
```diff
@@ -6,29 +6,30 @@
...
     # .      of importing from test_train in each location
-    test_data_path = data_paths_helper(path)
-    topography_path = (
-        f"{test_data_path.fine}/data.nc" if add_topography_path is True else None
-    )
+    # NOTE: data_paths_helper creates fine and coarse data
+    #   including the static input topography field
+    path = tmp_path / "test_data"
+    path.mkdir()
+    return data_paths_helper(path)
```
AnnaKwa, line 18: Slightly out of scope of this PR, but it might be easier to adjust future tests if the `data_paths_helper` func gets updated to write the static inputs to a separate static_inputs path (could add optional attr to `FineCoarseResPair`

fme/downscaling/models.py
```diff
@@ -594,10 +612,18 @@ def _checkpoint(self) -> Mapping[str, Any]:
     def build(
         self,
     ) -> DiffusionModel:
+        static_inputs: StaticInputs | None
         if self._checkpoint["model"]["static_inputs"] is not None:
+            if self.static_inputs is not None:
+                raise ValueError(
+                    "Cannot override static_inputs: the checkpoint already has "
```
AnnaKwa: Suggestion to clarify the first part of the error:
```suggestion
                    "The model checkpoint already has static inputs from training. "
                    "static_inputs should not be provided in checkpoint model config."
```

fme/downscaling/models.py
AnnaKwa: I'm fine with keeping this backwards compatible, but I'm also ok with just removing this attr and breaking older configs since I don't use this field, and the dacite error should be pretty straightforward and it's an easy config fix. The model uploaded to hf has the static_inputs saved so this shouldn't affects its usage.

fme/downscaling/inference/conftest.py
frodre: I'll add this as a coding task.

PR #927, opened 2026-03-09
Make `scripts/data_process/append_dataset.py` a bit more flexible

jpdunc23:
Allows `append_dataset.py` to use simplified configs for the `--dataset-config`, including reusing the same file as `--append-config`.

### DIFF
--- a/fme/downscaling/data/__init__.py
+++ b/fme/downscaling/data/__init__.py
@@ -12,7 +12,7 @@ from .datasets import (
     PairedBatchItem,
     PairedGriddedData,
 )
-from .topography import StaticInput, StaticInputs, get_normalized_topography
+from .static import StaticInput, StaticInputs, get_normalized_static_input
 from .utils import (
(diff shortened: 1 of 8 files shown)


scripts/data_process/append_dataset.py
```diff
@@ -25,6 +25,9 @@ def from_file(cls, path: str) -> "DatasetAppendConfig":
         with open(path, "r") as file:
             data = yaml.safe_load(file)
 
+        if "dataset_computation" in data:
```
Arcomano1234, line 28: Question: Do we want to just remove `strict=True` and then we don't need this right?

jpdunc23: I was thinking it might be helpful to keep it strict to guard against typos in `renaming`.

Arcomano1234: Yeah I agree this safer.

PR #930, opened 2026-03-09
Add optional LR tuning trials during training

mcgibbon:
At the start of configurable epochs, the trainer forks the current model into a baseline and a candidate copy, trains both for a short number of batches (candidate at a scaled LR), validates both, and adopts the candidate LR if it wins by a configurable margin. This enables automatic learning rate adaptation without manual tuning or restarts.

Changes:
- `fme.core.generics.lr_tuning.LRTuningConfig` and `run_lr_tuning_trial`: new dataclass and function implementing isolated LR comparison trials with deep-copied stepper/optimization/EMA state
- `fme.core.generics.trainer.Trainer`: adds `_maybe_tune_lr` called at the start of each epoch; `validate_one_epoch` now uses `run_validation`; tracks `_last_val_loss` across epochs for tuning threshold
- `LRTuningConfig` added as an optional field on `ace`, `coupled`, and `diffusion` `TrainConfig` dataclasses, and exported from `fme.ace`

- [x] Tests added

### DIFF
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -299,6 +299,11 @@ class TrainConfig:
                 "train_loader and weather_evaluation loader must both use labels or "
                 "both not use labels"
             )
+        if self.lr_tuning is not None and self.optimization.has_lr_schedule:
+            raise ValueError(
+                "lr_tuning and optimization.scheduler cannot both be specified; "
+                "lr_tuning is an alternative form of learning rate scheduling"
+            )
         if not is_local(self.experiment_dir):
             raise ValueError(
                 f"During training, experiment_dir must currently be a local "
--- a/fme/core/generics/test_lr_tuning.py
+++ b/fme/core/generics/test_lr_tuning.py
@@ -14,8 +14,8 @@ from fme.core.generics.data import DataLoader, GriddedDataABC
 from fme.core.generics.lr_tuning import LRTuningConfig, run_lr_tuning_trial
 from fme.core.generics.optimization import OptimizationABC
 from fme.core.generics.train_stepper import TrainOutputABC, TrainStepperABC
-from fme.core.optimization import Optimization
-from fme.core.scheduler import SchedulerConfig
+from fme.core.optimization import Optimization, OptimizationConfig
+from fme.core.scheduler import SchedulerConfig, SequentialSchedulerConfig
 from fme.core.training_history import TrainingJob
 from fme.core.typing_ import Slice, TensorDict
 
@@ -575,3 +575,26 @@ def test_trial_does_not_mutate_original_optimizer_state():
                 assert (
                     buf == current_state["optimizer_state_dict"]["state"][key][buf_name]
                 ), f"Optimizer state[{key}][{buf_name}] was mutated by the trial"
+
+
+def test_optimization_config_has_lr_schedule_default():
+    """Default OptimizationConfig has no LR schedule."""
+    config = OptimizationConfig()
+    assert config.has_lr_schedule is False
+
+
+def test_optimization_config_has_lr_schedule_with_scheduler():
+    """OptimizationConfig with a scheduler type has an LR schedule."""
+    config = OptimizationConfig(scheduler=SchedulerConfig(type="CosineAnnealingLR"))
+    assert config.has_lr_schedule is True
+
+
+def test_optimization_config_has_lr_schedule_with_sequential_scheduler():
+    """OptimizationConfig with a SequentialSchedulerConfig has an LR schedule."""
+    config = OptimizationConfig(
+        scheduler=SequentialSchedulerConfig(
+            schedulers=[SchedulerConfig(type="CosineAnnealingLR")],
+            milestones=[5],
+        )
+    )
+    assert config.has_lr_schedule is True
--- a/fme/core/optimization.py
+++ b/fme/core/optimization.py
@@ -263,6 +263,13 @@ class OptimizationConfig:
                 DeprecationWarning,
             )
 
+    @property
+    def has_lr_schedule(self) -> bool:
+        """Whether a learning rate scheduler is configured."""
+        if isinstance(self.scheduler, SequentialSchedulerConfig):
+            return True
+        return self.scheduler.type is not None
+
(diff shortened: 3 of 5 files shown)


fme/core/generics/lr_tuning.py
```diff
@@ -0,0 +1,161 @@
...
+from fme.core.typing_ import Slice
+
+
+@dataclasses.dataclass
+class LRTuningConfig:
+    """
+    Configuration for periodic learning rate tuning trials.
+
+    At the start of epochs contained in ``epochs``, the trainer forks the
+    current model into a baseline and a candidate copy. Both are trained for
```
oliverwm1, line 23: how does using this feature affect memory usage? Would be helpful to see metrics from some actual production runs you've done.

mcgibbon: Here's the workspace I linked in the team meeting notes: https://wandb.ai/ai2cm/ace?nw=fc7nmzc9w7j you'd want to enable train-amip-c96-shield-daily-lr1e-3 to get an apples to apples comparison (both are on Titan, and the only difference between the two is the tuning configuration).

The memory usage can't be directly compared because the daily model is trained on Jupiter while the tuning model is trained on Titan.

The memory usage increases only by 5%. Perhaps much less than you'd expect from keeping two additional copies of the neural network, but it's because the large majority of our memory use is in the model state and activations during training. If this for some reason became an issue, we could reduce or remove this impact by keeping the models not currently in use on the CPU.

image

fme/core/generics/lr_tuning.py
```diff
@@ -0,0 +1,161 @@
...
+        improvement_threshold: The candidate must beat the baseline's
+            validation loss by at least this fraction of the pre-trial
+            validation loss (e.g. 0.01 means the candidate must be lower
+            by at least 1% of the pre-trial loss).
+    """
+
+    lr_factor: float
+    num_batches: int
+    epochs: Slice = dataclasses.field(default_factory=Slice)
+    improvement_threshold: float = 0.0
```
oliverwm1: What did you use for the best run that used this feature? I think we should use that as the default here, if it was different than 0.0

fme/core/generics/lr_tuning.py
```diff
@@ -0,0 +1,161 @@
...
+        lr_factor: Multiply the current LR by this to get the candidate LR.
+        num_batches: Number of training batches for each fork in the trial.
+        improvement_threshold: The candidate must beat the baseline's
+            validation loss by at least this fraction of the pre-trial
+            validation loss (e.g. 0.01 means the candidate must be lower
+            by at least 1% of the pre-trial loss).
+    """
+
+    lr_factor: float
+    num_batches: int
```
oliverwm1: Can we provide a default? The right choice might depend on dataset/loss function, but some guidance either as a default here or in the docstring would be helpful.

fme/core/generics/lr_tuning.py
```diff
@@ -0,0 +1,161 @@
...
+        aggregator=candidate_agg,
+        ema=candidate_ema,
+        validate_using_ema=validate_using_ema,
+    )
+    candidate_val_logs = candidate_agg.get_logs(label="val")
+
+    baseline_val_loss = baseline_val_logs["val/mean/loss"]
+    candidate_val_loss = candidate_val_logs["val/mean/loss"]
+
+    threshold = baseline_val_loss - config.improvement_threshold * pre_trial_val_loss
```
oliverwm1: Why not use 
```suggestion
    threshold = baseline_val_loss - config.improvement_threshold * baseline_val_loss
```
instead? This would simplify integration into the trainer, and I'm not clear on what the advantage of computing the diff based on pre trial val loss is.

fme/core/generics/trainer.py
```diff
@@ -245,12 +249,15 @@ def __init__(
...
 
+        self._build_optimization = build_optimization
+        self._build_ema = build_ema
         self.optimization = build_optimization(stepper.modules)
         self._end_of_batch_callback = end_of_batch_callback
         self._end_of_epoch_callback = end_of_epoch_callback
         self._no_optimization = NullOptimization()
         self._aggregator_builder = aggregator_builder
         self._ema = build_ema(stepper.modules)  # build before restore_checkpoint
+        self._last_val_loss: float | None = None
```
oliverwm1: shouldn't this now also be serialized in case of training interruption? Or maybe better, following my other suggestion we could avoid the need for tracking this `self._last_val_loss` entirely.

fme/core/generics/lr_tuning.py
mcgibbon: OK, let's set the defaults to what I used the first time that seemed to work. I would suggest we update these defaults if we get a chance to tune it and find something better. Those values were 200 for num_batches, and 0.001 for improvement_threshold, with 0.5 for lr_factor - we should set each of these defaults.

fme/core/generics/lr_tuning.py
mcgibbon: Yeah, I think that's reasonable, and avoids needing that value to be cached/stored. It shouldn't really affect the results.

fme/core/generics/trainer.py
mcgibbon: I think we should get rid of this entirely according to your other suggestion.

fme/ace/train/train_config.py
```diff
@@ -267,6 +268,7 @@ class TrainConfig:
     )
     evaluate_before_training: bool = False
     save_best_inference_epoch_checkpoints: bool = False
+    lr_tuning: LRTuningConfig | None = None
```
oliverwm1, line 271: question: should we enforce that no LR scheduler is specified in the optimization config if this is provided? Using both at once seems rather confusing since this is just another form of LR scheduler.

Related question: should this option be configured as a LR schedule on the optimizer instead of TrainConfig top level?

mcgibbon: I think that would be good to enforce. I don't think it's possible for the optimizer to handle this - the owner must be able to run these trial experiments, and the optimizer shouldn't have that kind of scope. But I could see moving the learning rate schedule to an explicit section of the trainer, in the same place as this lr tuning. I'll leave any of these options for later consideration.

PR #933, opened 2026-03-10
Write proper `"init_time"` in `MonthlyDataWriter`

spencerkclark:
With the existing code in our monthly data writer, we write the `"init_time"` as the 1st of the first month of the time series for each ensemble member. This is inconsistent with how we log the initialization time in the raw data writers, where it represents the true initialization time of the runs, and is no easier to square with the `"valid_time"` and lead `"time"` coordinates, which represent the 15th of each month and the integer number of months since the start, respectively.

Changes:
- The `MonthlyDataWriter` and `PairedMonthlyDataWriter` now take `initial_condition_times` as an argument in their constructors rather than `n_samples`.
- The true initialization time of each run rather than the 1st of the first month is now logged to the `"init_time"` coordinate in the `MonthlyDataWriter`.  This harmonizes the behavior with the change made to the other data writers in #910.

### DIFF
--- a/fme/downscaling/data/__init__.py
+++ b/fme/downscaling/data/__init__.py
@@ -12,7 +12,7 @@ from .datasets import (
     PairedBatchItem,
     PairedGriddedData,
 )
-from .topography import StaticInput, StaticInputs, get_normalized_topography
+from .static import StaticInput, StaticInputs, get_normalized_static_input
 from .utils import (
     BatchedLatLonCoordinates,
     ClosedInterval,
--- a/fme/downscaling/data/config.py
+++ b/fme/downscaling/data/config.py
@@ -23,11 +23,7 @@ from fme.downscaling.data.datasets import (
     PairedBatchData,
     PairedGriddedData,
 )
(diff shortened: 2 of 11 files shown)


fme/ace/inference/data_writer/monthly.py
```diff
@@ -197,8 +195,8 @@ def _get_variable_names_to_save(
         return get_all_names(*data_varnames, allowlist=self._save_names)
 
     def _extend_lead_time(self, old_size: int, new_size: int):
-        init_time = self.dataset.variables[LEAD_TIME_DIM]
-        init_time[old_size:new_size] = np.arange(old_size, new_size)
+        lead_time = self.dataset.variables[LEAD_TIME_DIM]
+        lead_time[old_size:new_size] = np.arange(old_size, new_size)
```
spencerkclark, line 199: This change is not strictly necessary in this PR, but the naming was an unintentional mistake for these local variables introduced in https://github.com/ai2cm/full-model/pull/2372, since we are not changing the `"init_time"` in this method.

PR #944, opened 2026-03-10
scripts to plot event maps and compare histograms

AnnaKwa:
- add script `plot_compared_histograms.py` that utilizes existing functions in `plot_beaker_histograms.py` to overlay histograms of events that exist in multiple output eval event datasets
```
    python compare_histograms.py <dataset_id_1> <dataset_id_2> [<dataset_id_3> ...]
        [--labels label1 label2 ...] [--output-dir <path>]
```
- add script `plot_events.py` that takes single beaker dataset and makes map plot of the coarse, target, and generated samples as well as histogram 
```
    python plot_events.py <beaker_dataset_id> [--output-dir <path>]
        [--coarse-data <path>] [--variables VAR1 VAR2 ...]

```

example outputs
image

image


scripts/downscaling/plot_beaker_histograms.py
```diff
@@ -25,8 +25,8 @@
 import numpy as np
 import xarray as xr
 
-# Matching for <event_name>_YYYYMMDD.nc
-_EVENT_FILE_RE = re.compile(r"(.+)_(\d{8})\.nc$")
+# Matching for <event_name>_YYYYMMDD*.nc
```
AnnaKwa: Some of the event labels are of the format YYYYMMDDTHH

scripts/downscaling/plot_compared_histograms.py
```diff
@@ -0,0 +1,269 @@
...
+                    var_prefix,
+                    event_name,
+                    save_path=fig_path,
+                )
+                print(f"  Saved: {fig_path}")
+
+            for ds in opened_datasets:
+                ds.close()
+    finally:
+        import shutil
```
frodre: Move import to the top of file.

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+from cartopy.io import shapereader
+from cartopy.mpl.gridliner import LATITUDE_FORMATTER, LONGITUDE_FORMATTER
+
+warnings.filterwarnings("ignore")
+
+from plot_beaker_histograms import plot_histogram_lines
+
+TIME_SEL = slice(cftime.DatetimeJulian(2023, 1, 1), None)
+# Matching for *YYYYMMDD*.nc (date can appear anywhere in the filename)
+_EVENT_FILE_RE = re.compile(r"(.+?)[\._-]?(\d{8})[\._-]?(.*)\.nc$")
```
frodre: I would probably make a new `beaker_utils.py` file that the other plotting scripts can import from.

Should we make this consistent or expand the match search criteria (as done here) in the `plot_beaker_histograms`.  It is totally dependent on the user configuration of the event name.

```suggestion
_EVENT_FILE_RE = re.compile(r"(.+)_(\d{8}).*\.nc$")
```

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+import xarray as xr
+from cartopy.feature import ShapelyFeature
+from cartopy.io import shapereader
+from cartopy.mpl.gridliner import LATITUDE_FORMATTER, LONGITUDE_FORMATTER
+
+warnings.filterwarnings("ignore")
+
+from plot_beaker_histograms import plot_histogram_lines
+
+TIME_SEL = slice(cftime.DatetimeJulian(2023, 1, 1), None)
```
frodre, line 42: Comment what this time is selecting for

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+            "gs://vcm-ml-raw-flexible-retention/2025-07-25-X-SHiELD-AMIP-FME/regridded-zarrs/gaussian_grid_180_by_360/control/instantaneous_physics_fields.zarr"
+        ).sel(time=time_sel)[
+            ["eastward_wind_at_ten_meters", "northward_wind_at_ten_meters"]
+        ]
+        prate = xr.open_zarr(
+            "gs://vcm-ml-raw-flexible-retention/2025-07-25-X-SHiELD-AMIP-FME/regridded-zarrs/gaussian_grid_180_by_360/control/fluxes_2d.zarr"
+        ).sel(time=time_sel)["PRATEsfc"]
+        pres = xr.open_zarr(
+            "gs://vcm-ml-raw-flexible-retention/2025-07-25-X-SHiELD-AMIP-FME/regridded-zarrs/gaussian_grid_180_by_360/control/column_integrated_dynamical_fields.zarr"
+        ).sel(time=time_sel)["PRESsfc"]
```
frodre: A little bit of separation between path specification and dataset opening would make this segment more readable.

Also, I could make a consolidated coarse dataset using the processing script (or upload our current weka coarse data), if we don't have one up on GCS yet for the control dataset.

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+    ds_ = ds[vars]
+
+    if sel:
+        ds_ = ds_.sel(sel)
+
+    if len(samples) == 0:
+        samples = [0]
+
+    if var_name == "PRMSL":
+        # fill PRMSL_coarse with nans
```
frodre, line 187: Why?

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+        ds_["PRMSL_coarse"].values[:] = np.nan
+
+    vmax = ds_.to_array().max()
+    if ds_.to_array().min() < -0.2:
+        plot_kwargs["cmap"] = "RdBu_r"
+    else:
+        plot_kwargs["cmap"] = "turbo"
+        plot_kwargs["vmin"] = min(0, ds_.to_array().min())
```
frodre: Is it worth it to extend to negative values if they exist in this case?  Maybe just cap at 0

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+        row = (i + 2) // n_cols
+        col = (i + 2) % n_cols
+        add_outer_latlon_grid(
+            ax,
+            show_left=(col == 0),
+            show_bottom=(row == n_rows - 1),
+        )
+    cbar_ax = fig.add_axes([0.99, 0.25, 0.01, 0.5])  # [left, bottom, width, height]
+    cbar = fig.colorbar(img, cax=cbar_ax)
+    cbar.set_label(f"{var_name} [m/s]")
```
frodre: Fixed units don't work for all variables

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+    # plt.tight_layout()
+
+    return fig, axes
+
+
+def fetch_beaker_dataset(dataset_id: str, target_dir: str) -> None:
```
frodre: Some functions down here repeated from other file, but seem like good candidates to be in the new utils file.

scripts/downscaling/plot_events.py
```diff
@@ -0,0 +1,374 @@
...
+        subplot_kw={"projection": ccrs.PlateCarree()},
+    )
+
+    axes = axes.ravel()  # 1D array, easy to index
+    # Use only the first N axes
+    for ax in axes[N:]:
+        ax.set_visible(False)
+
+    suffixes = ["coarse", "target", "predicted"]
+    vars = [f"{var_name}_{suffix}" for suffix in suffixes]
```
frodre: Not really an issue, but overrides the python `vars`, maybe expand

scripts/downscaling/plot_events.py
AnnaKwa: Coarse data only has PRESsfc, which is what is actually used as input for the model. If the surface pressure is shown though the range often is very different from the PRMSL range which messes with the colorbar and makes it hard to see the generated features.

scripts/downscaling/plot_events.py
AnnaKwa: Actually, I meant for to be max(0, min_data_value). Otherwise, pressure fields don't have much range on the colorbar.

PR #948, opened 2026-03-10
Fill NaNs in the 3D ocean layer thickness `dz`

### LINKED ISSUE #949: NaN gradients from `depth_integral` due to 0 * NaN in backward pass
`DepthCoordinate.depth_integral` produces NaN gradients for all variables involved in the depth integral (e.g. `thetao_0`, `thetao_1`) during training with the OHC conservation corrector.

The root cause is the multiplication of the integrand with `dz` containing NaNs at land-masked points:

`integral = (integrand * self.dz * self.mask).nansum(dim=-1)`

While `nansum` correctly ignores these in the forward pass, the backward pass computes `d(nansum)/d(product) * d(product)/d(integrand) = 0 * NaN = NaN` at masked points. These NaN gradients propagate through the OHC correction ratio back to the network output, eventually corrupting all model weights.

jpdunc23:
Fixes #949 by filling `dz` with 0 on all land-masked points.

Changes:
- Updates `test_ocean_train.py` to use more realistic synthetic data. The most important change is the addition of `deptho` with NaNs.
- Fixes an unrelated bug where `DepthCoordinate` equality checks failed due to NaNs in `deptho`.
- `DepthCoordinate.dz` is now always 3D even when `deptho` is missing from the data.

- [x] Tests added

Resolves #949

### DIFF
--- a/fme/ace/test_ocean_train.py
+++ b/fme/ace/test_ocean_train.py
@@ -95,7 +95,7 @@ def save_ocean_nd_netcdf(
             f"dim_sizes.nz_interface ({dim_sizes.nz_interface}) "
             f"must be nz_levels ({nz_levels}) + 1."
         )
-    max_depth_m = 500.0 * nz_levels
+    max_depth_m = 1000.0
     interface_depths = np.linspace(0.0, max_depth_m, dim_sizes.nz_interface)
     for i in range(dim_sizes.nz_interface):
         idepth_name = f"idepth_{i}"


fme/core/coordinates.py
```diff
@@ -451,7 +453,7 @@ def __eq__(self, other) -> bool:
             return False
         if self.deptho is not None and other.deptho is not None:
             try:
-                torch.testing.assert_close(self.deptho, other.deptho)
+                torch.testing.assert_close(self.deptho, other.deptho, equal_nan=True)
```
jpdunc23, line 456: Fixes an unrelated bug.

fme/core/test_coordinates.py
```diff
@@ -49,6 +49,18 @@
...
+                idepth=torch.tensor([0.0, 10.0, 50.0]),
+                mask=torch.tensor([1.0, 0.0]),
+                deptho=torch.tensor([float("nan"), 10.0]),
+            ),
+            DepthCoordinate(
+                idepth=torch.tensor([0.0, 10.0, 50.0]),
+                mask=torch.tensor([1.0, 0.0]),
+                deptho=torch.tensor([float("nan"), 10.0]),
+            ),
+        ),
```
jpdunc23, line 63: Tests the `DepthCoordinate.__eq__` bug.

fme/core/coordinates.py
```diff
@@ -330,16 +330,18 @@ def vertical_integral(
...
 ) -> torch.Tensor:
-    if deptho is not None:
-        z_top = idepth[..., :-1]
-        z_bot = idepth[..., 1:]
-        deptho = deptho.unsqueeze(-1)
-        dz = torch.clamp(deptho, min=z_top, max=z_bot) - z_top
+    z_top = idepth[..., :-1]
+    z_bot = idepth[..., 1:]
+    if deptho is None:
+        deptho_expanded = (mask * z_bot).max(dim=-1, keepdim=True).values
```
spencerkclark: Is `.values` not needed here?
```suggestion
        deptho_expanded = (mask * z_bot).max(dim=-1, keepdim=True)
```

fme/ace/test_ocean_train.py
```diff
@@ -87,15 +95,121 @@ def save_ocean_nd_netcdf(
             f"dim_sizes.nz_interface ({dim_sizes.nz_interface}) "
             f"must be nz_levels ({nz_levels}) + 1."
         )
+    max_depth_m = 500.0 * nz_levels
```
spencerkclark: nit: is there a need for this to scale with the number of levels, or would it be safer to use a constant?

PR #951, opened 2026-03-10
Add dask/xpartition support to time_coarsen script

mcgibbon:
When dask and xpartition are available, uses lazy loading via xr.open_zarr and chunked parallel writes via xpartition. Falls back to eager loading and xarray's to_zarr when they are not installed, preserving testability in CI environments without dask.

Also extracts shared coarsening logic into a public `coarsen()` function and replaces asserts with proper ValueError raises.


scripts/data_process/time_coarsen.py
```diff
@@ -62,40 +84,46 @@ def main(config: Config, run: int, dry_run: bool = False):
...
+    Works with both eager (numpy) and lazy (dask) arrays.
+    """
     constant_names = [
         name
         for name in ds.data_vars
         if any(name.startswith(prefix) for prefix in config.constant_prefixes)
     ]
-    assert set(constant_names).intersection(set(config.snapshot_names)) == set()
-    assert set(constant_names).intersection(set(config.window_names)) == set()
+    if set(constant_names).intersection(set(config.snapshot_names)):
```
spencerkclark, line 121: While we are adding these checks I suppose we might as well raise if the `snapshot_names` and `window_names` overlap as well.

scripts/data_process/time_coarsen.py
```diff
@@ -37,6 +52,13 @@ class TimeCoarsenConfig:
     snapshot_names: list[str]
     window_names: list[str]
     constant_prefixes: list[str]
+    n_split: int = 1
+    chunking: dict[str, int] = dataclasses.field(
+        default_factory=lambda: {"time": 1, "lat": -1, "lon": -1}
+    )
+    sharding: dict[str, int] | None = dataclasses.field(
+        default_factory=lambda: {"time": 360, "lat": -1, "lon": -1}
+    )
```
spencerkclark: For better or worse, our datasets do not all use the same dimension names, particularly for the latitude and longitude dimensions, which is why we abstract this out to the [`ChunkingConfig`](https://github.com/ai2cm/ace/blob/d6073ac85043da9d0ff20c4227d1ec1241dd4c56/scripts/data_process/compute_dataset.py#L196-L206) in `compute_dataset.py`.  For the purposes of this workflow, I think it is safe to assume the latitude and longitude (or other spatial dimensions) are already appropriately chunked and we only need to adjust the time chunking.  

Since we already assume the time dimension is named `"time"` in the stats workflow, it is probably safe to assume that here for now too:
```suggestion
    chunking: dict[str, int] = dataclasses.field(
        default_factory=lambda: {"time": 1}
    )
    sharding: dict[str, int] | None = dataclasses.field(
        default_factory=lambda: {"time": 360}
    )
```

scripts/data_process/time_coarsen.py
```diff
@@ -113,23 +141,64 @@ def process_path_pair(
...
+
+def process_path_pair(
+    input_path: str, output_path: str, config: TimeCoarsenConfig, dry_run: bool
+):
+    logging.info(f"Processing input: {input_path} to output: {output_path}")
+    if os.path.exists(output_path):
+        logging.warning(f"Output path {output_path} already exists. Skipping.")
+        return
+    if _HAS_XPARTITION:
+        ds = xr.open_zarr(input_path)
```
spencerkclark: For performance reasons described [here](https://allenai.slack.com/archives/C052FC4NU0G/p1770134280875889), we will want to open these datasets with larger chunks when using dask:
```suggestion
        with dask.config.set({"array.chunk-size": "128MiB"}):
            ds = xr.open_zarr(input_zarr, chunks={"time": "auto"})
```

scripts/data_process/time_coarsen.py
mcgibbon: Done.

PR #954, opened 2026-03-10
Downscaling model handles static input instead of DataLoader

frodre:
Previously, the `DataLoader` was responsible for subsetting and device-placing static inputs (e.g. fine-resolution topography) to match each batch's spatial extent before passing them to the model. This moves that responsibility into `DiffusionModel` itself: the model stores the full-domain static inputs on construction and subsets them per-batch using the batch's coordinate metadata.

  Changes:
  - `DiffusionModel.__init__` now calls `.to_device()` on static inputs at construction
  - New `DiffusionModel._subset_static_inputs` encapsulates fine lat/lon subsetting for train_on_batch and generate_on_batch
  - `generate_on_batch_no_target` derives the fine coordinate interval from coarse batch coordinates via `adjust_fine_coord_range`, then subsets stored static inputs
  - `BatchData` gains `lat_interval` and `lon_interval` properties for use in subsetting
  - `adjust_fine_coord_range` now raises a clear `ValueError` when the coordinate range is too close to the domain boundary for the required number of fine points to exist (documents the implicit ±88° latitude requirement)
  - The `static_inputs` parameter on all public model methods is retained but ignored — removal is deferred to a follow-on PR

- [x] Tests added


fme/downscaling/data/datasets.py
```diff
@@ -441,6 +441,16 @@ def __post_init__(self):
     def horizontal_shape(self) -> tuple[int, int]:
         return self._horizontal_shape
 
+    @property
+    def lat_interval(self) -> ClosedInterval:
+        lat = self.latlon_coordinates.lat[0]  # all batch members identical; use first
```
frodre, line 446: Given that we're no longer doing per-item random patch definitions, we could update `BatchData` to just use `LatLonCoordinates`

fme/downscaling/models.py
```diff
@@ -415,9 +449,11 @@ def train_on_batch(
     def generate(
         self,
         coarse_data: TensorMapping,
-        static_inputs: StaticInputs | None,
+        static_inputs: StaticInputs | None,  # TODO: remove in follow-on PR
```
AnnaKwa: This should be kept since `generate` will be passed subsetted static_inputs from `generate_on_batch` or `generate_on_batch_no_target`, right?

fme/downscaling/models.py
```diff
@@ -415,9 +449,11 @@ def train_on_batch(
...
         n_samples: int = 1,
     ) -> tuple[TensorDict, torch.Tensor, list[torch.Tensor]]:
+        # static_inputs receives an internally-subsetted value from the calling method;
+        # external callers should use generate_on_batch / generate_on_batch_no_target.
```
AnnaKwa, line 456: Out of scope for this PR, but the only external caller is `CascadePredictor.generate`, which uses this instead of `generate_on_batch_no_target` because it was simpler to just pass the first model's output tensor instead of making a new BatchData object out of it to input to the next model.  It would be worth adding a helper function to construct a new `BatchData` object out of `generate`'s output; then we could make this method private.

fme/downscaling/models.py
```diff
@@ -467,22 +503,54 @@ def generate(
...
         self,
         batch: BatchData,
-        static_inputs: StaticInputs | None,
+        static_inputs: StaticInputs | None,  # TODO: remove in follow-on PR
         n_samples: int = 1,
     ) -> TensorDict:
-        generated, _, _ = self.generate(batch.data, static_inputs, n_samples)
+        # Ignore the passed static_inputs; derive the fine lat/lon interval from coarse
+        # batch coordinates via adjust_fine_coord_range, then subset self.static_inputs.
+        if self.config.use_fine_topography:
```
AnnaKwa, line 511: Is this check necessary? Could this instead just check if `self.static_inputs` is None? 

Since the previous PR removed the old option of loading HGTsfc from the fine dataset, this config option can be deprecated and downstream checks could be removed.

fme/downscaling/data/datasets.py
```diff
@@ -441,6 +441,16 @@ def __post_init__(self):
...
+        return ClosedInterval(lat.min().item(), lat.max().item())
+
+    @property
+    def lon_interval(self) -> ClosedInterval:
+        lon = self.latlon_coordinates.lon[0]  # all batch members identical; use first
+        return ClosedInterval(lon.min().item(), lon.max().item())
+
```
AnnaKwa, line 453: Could you add a test that the coordinates are as expected when `BatchData.generate_from_patches` is called? It looks like the existing tests for that usage only check data values.

fme/downscaling/models.py
frodre: I mentioned in the #959 , but I ran into the problem that inference needs knowledge of the output sizes that forces some awkward handling within that module.  I think this could be solved by having a richer set of output information (e.g., coordinates) passed along in the generation/prediction. That would also allow for some smarter handling in CascadedModels as well.

fme/downscaling/models.py
frodre: Yes, I agree and was thinking this and the `DataRequirements.use_fine_topography` would be handled in another PR to simplify things.

fme/downscaling/data/datasets.py
frodre: Added tests for the data and coordinates for generate_from_patches under `test_datasets.py`.  I did not adjust the tests in `test_static.py` since the patch generation will be removed in #956.

fme/downscaling/data/datasets.py
```diff
@@ -464,7 +474,7 @@ def to_device(self) -> "BatchData":
 
     def __getitem__(self, k):
         return BatchItem(
-            {key: value[k].squeeze() for key, value in self.data.items()},
```
frodre, line 467: Tests uncovered an error for patches of length 1 in the x/y dimension!

Issue #909, opened 2026-03-06
reproducibility in s_sfno wrt to decomp


mahf708:
General issue (not with your code): To achieve reproducibility independent of decomposition, I think we're going to need to initialize the global random state (with the same seed on each rank, or on root and then scatter) and then sub-select the local slice. This would need to be as a flag we basically just use for testing, since in production especially at high resolutions this isn't desired, and gaussian noise with a different seed on each rank is perfectly fine (I think Distributed does seed randomness differently on each rank? Maybe?)

Feel free to open issues related to this if you want to track it.

_Originally posted by @mcgibbon in https://github.com/ai2cm/ace/pull/900#discussion_r2884708845_

PR #955, opened 2026-03-11
enable spatial parallel data loading

mahf708:
enable simplistic spatially parallel data laoding 

Changes:
- BatchData has scatter_spatial method; properties, coords, and masks have localize method
- Distributed has require_no_spatial_parallelism error handling
- [x] Tests added

### DIFF
--- a/fme/core/mask_provider.py
+++ b/fme/core/mask_provider.py
@@ -136,11 +136,11 @@ class MaskProvider(MaskProviderABC):
     def localize(self) -> "MaskProvider":
         if not self._masks:
             return self
-        dist = Distributed.get_instance()
-        example_mask = next(iter(self._masks.values()))
-        img_shape = example_mask.shape[-2:]
         return MaskProvider(
-            {k: v[dist.get_local_slices(v.shape)].contiguous() for k, v in self._masks.items()}
+            {
+                k: v[Distributed.get_instance().get_local_slices(v.shape)].contiguous()
+                for k, v in self._masks.items()
+            }
         )
 
     def update(self, other: "MaskProvider") -> None:


fme/ace/data_loading/gridded_data.py
```diff
@@ -9,16 +9,51 @@
...
+        )
+    return DatasetProperties(
+        properties.variable_metadata,
+        properties.vertical_coordinate,
+        coords,
+        mask_provider,
+        properties.timestep,
+        properties.is_remote,
+        properties.all_labels,
+    )
```
mahf708: silently returning global healpix data, should we document this in an issue?

fme/ace/data_loading/gridded_data.py
```diff
@@ -174,21 +216,27 @@ def __init__(
...
         self._loader = loader
-        self._properties = properties.to_device()
+        img_shape = properties.horizontal_coordinates.shape
+        self._global_img_shape: tuple[int, int] = (img_shape[-2], img_shape[-1])
+        self._properties = _scatter_properties(properties.to_device())
         self._n_initial_conditions: int | None = None
         if isinstance(initial_condition, PrognosticStateDataRequirements):
             self._initial_condition: PrognosticState = get_initial_condition(
-                loader, initial_condition
+                self.loader, initial_condition
```
mahf708, line 186: In general, all diffs in this file are maybe a bit non-transparent, this is one of them. Note how we have `loader`, `self._loader`, and `self.loader` --- the last being distinct from the first two

fme/ace/data_loading/getters.py
```diff
@@ -50,7 +50,7 @@ def _get_sampler(
     dist = Distributed.get_instance()
     if sample_with_replacement_dataset_size is not None:
         local_sample_with_replacement_dataset_size = (
-            sample_with_replacement_dataset_size // dist.world_size
+            sample_with_replacement_dataset_size // dist.total_data_parallel_ranks
         )
         sampler = torch.utils.data.RandomSampler(
```
mahf708, line 60: spatial ranks in the same data group get different results, making this less than ideal (maybe similar to #909?) --- open issue or fix pattern?

fme/ace/data_loading/test_data_loader.py
```diff
@@ -1173,3 +1177,51 @@ def test_pinned_memory(tmp_path, time_buffer: int):
     for batch in loader:
         tensor = next(iter(batch.data.values()))
         assert tensor.is_pinned() is using_gpu()
+
+
+@pytest.mark.parallel
+def test_scatter_properties():
```
mahf708: potentially useless test

fme/ace/data_loading/getters.py
mcgibbon: Meaning the spatial map of data rank 0 will contain patches from different samples? If so this is a bug, can you please update the tests (ideally both the sampler test and the data loading test) so they fail due to this bug, and keep them marked xfail? Then we can merge this, open an issue, and fix it in another PR (or you can consider the issue already covered by the "implement data loading" issue).

fme/ace/data_loading/gridded_data.py
```diff
@@ -9,16 +9,51 @@
...
     DataLoader,
     GriddedDataABC,
     InferenceDataABC,
     SizedMap,
 )
+from fme.core.mask_provider import MaskProvider
+
+
+def _scatter_properties(properties: DatasetProperties) -> DatasetProperties:
+    """Slice horizontal coordinates and masks to the local spatial chunk."""
```
mcgibbon: nit: "scatter" to me implies communication, can you name this something like `localize_properties`?

~~Question: Will this mess up downstream code? e.g. `img_shape` still will need to be the full domain later on.~~

fme/ace/data_loading/gridded_data.py
mcgibbon: Maybe we should add a `dist.require_no_model_parallelism(msg: str)` helper function that raises an exception? Something that discourages an if-else type check to implement low-level features, but can raise if we encounter code we know is going to be wrong in spatially parallel configs?

fme/ace/data_loading/gridded_data.py
```diff
@@ -50,7 +85,9 @@ def __init__(
             will be on the current device.
         """
         self._loader = loader
-        self._properties = properties.to_device()
+        img_shape = properties.horizontal_coordinates.shape
+        self._global_img_shape: tuple[int, int] = (img_shape[-2], img_shape[-1])
+        self._properties = _scatter_properties(properties.to_device())
```
mcgibbon: Ah I see.

fme/ace/data_loading/test_data_loader.py
mcgibbon: Yeah, the issue is the test implements the logic.

You could instead implement this test with scatter-gather, i.e. call scatter_properties and then gather the results onto root, where you can make sure the properties satisfy some criteria (like no latitude or longitude is repeated).

Note scatter should probably be renamed to "localize", as long as we're still following the "all ranks give their own input" model. We can move to "scatter" when only root is loading the data. If you think that will be very soon/immediately, we can just not rename it maybe.

fme/ace/data_loading/getters.py
mcgibbon: I see, this code is only encountered when using this flag. Looks good to me, I don't think we need to support this intersection right now.

fme/ace/data_loading/gridded_data.py
```diff
@@ -9,16 +9,55 @@
...
+from fme.core.mask_provider import MaskProvider
+
+
+def _localize_properties(properties: DatasetProperties) -> DatasetProperties:
+    """Slice horizontal coordinates and masks to the local spatial chunk."""
+    dist = Distributed.get_instance()
+    coords = properties.horizontal_coordinates
+    if isinstance(coords, LatLonCoordinates):
+        h_slice, w_slice = dist.get_local_slices(coords.shape)
+        coords = LatLonCoordinates(
```
mcgibbon: Suggestion: Could do it in another PR, but probably we should move this logic to a `coords.localize()` helper function instead of doing low-level internal operations on the coordinates from here. Then the `require_no_spatial_parallelism` call can be inside the healpix coordinates class instead of here, as well.

fme/ace/data_loading/getters.py
mahf708: I only could find this flag to be used in downsampling (in yaml files). But I think we should document this in an issue. The "proper fix" is likely going to be quite complex if we want to keep the same exact behavior it is doing now, but I must admit, i am very confused by this, so I will need to think much more deeply about it

fme/ace/data_loading/gridded_data.py
```diff
@@ -71,8 +112,13 @@ def subset_loader(
     def _get_gpu_loader(
         self, base_loader: DataLoader[BatchData]
     ) -> DataLoader[BatchData]:
+        dist = Distributed.get_instance()
+        img_shape = self._global_img_shape
```
mcgibbon: nit: move these into modify_and_on_device so they access current base instances when called, instead of these function-local cached copies.

fme/ace/data_loading/gridded_data.py
```diff
@@ -174,21 +220,27 @@ def __init__(
...
             self._initial_condition = initial_condition.to_device()
 
     @property
     def loader(self) -> DataLoader[BatchData]:
-        def on_device(batch: BatchData) -> BatchData:
+        dist = Distributed.get_instance()
+        img_shape = self._global_img_shape
+
+        def scatter_and_on_device(batch: BatchData) -> BatchData:
+            batch.data = dist.scatter_spatial(dict(batch.data), img_shape)
```
mcgibbon: Suggestion: It feels like this should be moved to a `batch.scatter_spatial().to_device()` set of calls. Perhaps more than data will need to be scattered, especially later if you have only root load the base properties, and this is easier to manage in a BatchData method.

Also, you probably need `batch.to_device().scatter_spatial()` ordering soon, as when running on GPU, communication routines will fail if you run them on CPU data.

fme/core/distributed/distributed.py
```diff
@@ -183,6 +183,15 @@ def world_size(self) -> int:
...
         return self._distributed.total_ranks
 
+    def require_no_spatial_parallelism(self, msg: str) -> None:
+        """Raise NotImplementedError if spatial parallelism is active.
+
+        Use this to guard code paths that are known to be incorrect
+        when spatial co-ranks exist (world_size > total_data_parallel_ranks).
+        """
+        if self.world_size != self.total_data_parallel_ranks:
+            raise NotImplementedError(msg)
```
mcgibbon: nit: You could make this error slightly more informative and catchable if you define a `class SpatialParallelismNotImplemented(NotImplementedError): pass` class and raise that instead.

fme/ace/data_loading/gridded_data.py
mcgibbon: It looks like that's mostly a pre-existing issue. I think it's fine.

fme/ace/data_loading/getters.py
mcgibbon: It would likely require we define a separate type of random generator that's the same across data ranks, or that we scatter random numbers. However we didn't really end up using this feature, don't worry too much about it. It was mostly supposed to be a way for us to checkpoint more often, but we implemented that feature with the current training strategy, so this is somewhat obsolete.

fme/core/mask_provider.py
```diff
@@ -124,6 +133,17 @@ def to(self, device: str) -> "MaskProvider":
...
 
+    def localize(self) -> "MaskProvider":
+        if not self._masks:
+            return self
+        dist = Distributed.get_instance()
+        example_mask = next(iter(self._masks.values()))
+        img_shape = example_mask.shape[-2:]
+        slices = dist.get_local_slices(img_shape)
+        return MaskProvider(
+            {k: v[(..., *slices)].contiguous() for k, v in self._masks.items()}
```
mcgibbon:
```suggestion
            {k: v[dist.get_local_slices(v.shape)].contiguous() for k, v in self._masks.items()}
```
The call is cheap and this is only done during init, I'd suggest we do the safe thing and just get the slice for each value.

Also, get_local_slices is already meant to handle the full shape including leading dims.

fme/ace/data_loading/batch_data.py
```diff
@@ -170,6 +171,18 @@ def to_device(self) -> "BatchData":
             labels=self.labels.to(device) if self.labels is not None else None,
         )
 
+    def scatter_spatial(self, global_img_shape: tuple[int, int]) -> "BatchData":
+        """Slice data tensors to the local spatial chunk."""
+        dist = Distributed.get_instance()
+        return self.__class__(
+            data=dist.scatter_spatial(dict(self.data), global_img_shape),
```
mcgibbon, line 178: Neat, I guess in a follow-on PR we'll figure out how to handle this where BatchData doesn't even exist on the non-root ranks.

PR #956, opened 2026-03-11
Remove static_input from data pipeline and model call signatures

frodre:
This PR finalizes the removal of the `StaticInput` handling by the data pipeline.  The passing of static_input objects are removed from the data configuration, batch iteration, and model call signatures in favor of the direct model handling introduced in the previous downscaling PR (#954).

Changes:
- add `get_fine_coords_for_batch` to facilitate translation of an input batch domain to output coordinates via the models stored information.  For now, this relies on the model's `static_inputs`, but will be switched to model's stored coordinates in (#971)
- inference `Downscaler` now takes the batch `input_shape` instead of `static_inputs` to check the domain size and model type (regular `DiffusionModel` or `PatchPredictor`
- downscaling `torch.datasets` generators for `BatchData` no longer include `StaticInputs`
- removed `_apply_patch` and `_generate_from_patches` from `StaticInputs` 
- `config.py` no longer references static inputs as an argument

- [x] Tests added


fme/downscaling/predict.py
```diff
@@ -238,30 +235,28 @@ def save_netcdf_data(self, ds: xr.Dataset):
...
-
     def run(self):
-        aggregator = NoTargetAggregator(
-            downscale_factor=self.model.downscale_factor,
-            latlon_coordinates=self._fine_latlon_coordinates,
-        )
-        for i, (batch, static_inputs) in enumerate(self.batch_generator):
+        aggregator: NoTargetAggregator | None = None
+        for i, batch in enumerate(self.batch_generator):
+            if aggregator is None:
```
frodre, line 241: Lazy initialization, like done with `inference.py` to wait for availability on the output domain information

fme/downscaling/data/datasets.py
```diff
@@ -769,17 +727,4 @@ def patched_batch_gen_from_paired_loader(
...
-
-        if static_inputs is not None:
-            if fine_patches is None:
-                raise ValueError(
-                    "Static inputs provided but downscale_factor is None, cannot "
-                    "generate fine patches."
-                )
-            static_inputs_patches = static_inputs.generate_from_patches(fine_patches)
-        else:
-            static_inputs_patches = null_generator(len(coarse_patches))
```
AnnaKwa, line 782: Can remove `null_generator` from utils now

fme/downscaling/models.py
```diff
@@ -332,13 +334,44 @@ def _subset_static_inputs(
             )
         return self.static_inputs.subset_latlon(lat_interval, lon_interval)
 
+    def get_fine_coords_for_batch(self, batch: BatchData) -> LatLonCoordinates:
```
AnnaKwa, line 337: Suggestion: take the `ClosedInterval` and `LatLonCoordinates` as args rather than the `BatchData` to remove the dependence on how `BatchData` stores this info (e.g. won't have to change this code if switching `BatchData` to store the coords as `LatLonCoordinates` instead of `BatchedLatLonCoordinates`)

frodre: I'm going to leave the single argument method rather than expand into 3 arguments since I think the consolidated passing works in our favor. Expanding would force the repeat of the access pattern that wherever the method is used (and then we'd have to make those updates anyways in multiple locations when we to `LatLonCoordinates` at the batch level).

PR #957, opened 2026-03-11
Add separated channel module registry and step implementation

mcgibbon:
Adds a new module interface where modules explicitly take separate forcing and prognostic tensors as input and return separate prognostic and diagnostic tensors as output, instead of a single concatenated tensor. This makes channel semantics explicit at the module level.

  Changes:
  - `fme.core.registry.SeparatedModuleConfig`, `SeparatedModule`, `SeparatedModuleSelector`: new module registry mirroring `ModuleSelector` but with separated channel interface
  - `fme.core.registry.LegacyModuleAdapter`, `LegacyWrapper`: adapter wrapping old-style single-tensor modules for backwards compatibility
  - `fme.core.step.SeparatedModuleStepConfig`, `SeparatedModuleStep`: new `StepABC` implementation using the separated channel interface, with explicit `forcing_names`, `prognostic_names`, and `diagnostic_names` fields
  - `StepConfigABC.prognostic_names` and `StepABC.prognostic_names`: refactored from `@final @property` to `get_prognostic_names()` getter method, enabling subclasses to use `prognostic_names` as a dataclass field name

  - [x] Tests added


fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,271 @@
...
+        self._instance = self.registry.get(self.type, self.config)
+
+    @property
+    def module_config(self) -> SeparatedModuleConfig:
+        return self._instance
+
+    @classmethod
+    def register(
+        cls, type_name: str
+    ) -> Callable[[Type[SeparatedModuleConfig]], Type[SeparatedModuleConfig]]:  # noqa: UP006
```
mcgibbon: Question: Why is a noqa needed here?

fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,271 @@
...
+from typing import Any, ClassVar, Type  # noqa: UP035
+
+import dacite
+import torch
+from torch import nn
+
+from fme.core.dataset_info import DatasetInfo
+from fme.core.labels import BatchLabels, LabelEncoding
+
+from .module import CONDITIONAL_BUILDERS, ModuleSelector
```
mcgibbon: Question: Is it appropriate to re-use the CONDITIONAL_BUILDERS from .module? Or should we be defining these independently for SeparatedModule? Or should all new module types here support conditioning (we don't need backwards compatibility)? At a minimum labels can always be given as broadcasted input variables, right?

fme/core/step/separated_module.py
```diff
@@ -0,0 +1,395 @@
...
+    def get_state(self):
+        return {
+            "module": self.module.get_state(),
+            "secondary_decoder": self.secondary_decoder.get_module_state(),
+        }
+
+    def load_state(self, state: dict[str, Any]) -> None:
+        self.module.load_state(state["module"])
+        if "secondary_decoder" in state:
+            self.secondary_decoder.load_module_state(state["secondary_decoder"])
```
mcgibbon: Question: Do we need this check? There aren't any existing checkpoints to maintain backwards compatibility for.

Suggestion: Remove this check if you can.

fme/core/step/test_separated_module.py
```diff
@@ -0,0 +1,407 @@
...
+from fme.core.dataset_info import DatasetInfo
+from fme.core.normalizer import NetworkAndLossNormalizationConfig, NormalizationConfig
+from fme.core.registry import ModuleSelector, SeparatedModuleSelector
+from fme.core.step.args import StepArgs
+from fme.core.step.separated_module import SeparatedModuleStepConfig
+from fme.core.step.single_module import SingleModuleStepConfig
+from fme.core.step.step import StepSelector
+
+IMG_SHAPE = (16, 32)
+TIMESTEP = __import__("datetime").timedelta(hours=6)
```
mcgibbon: Suggestion: Properly import timedelta instead of this weird line.

fme/core/step/test_separated_module.py
```diff
@@ -0,0 +1,407 @@
...
+        # SingleModuleStep wraps as DummyWrapper(SFNO) while
+        # SeparatedModuleStep wraps as DummyWrapper(LegacyWrapper(SFNO)),
+        # so we copy at the SFNO level to avoid key prefix mismatches.
+        single_sfno = single_step.module.torch_module.module
+        separated_sfno = separated_step.module.torch_module.module.inner
+        separated_sfno.load_state_dict(single_sfno.state_dict())
+
+        return single_step, separated_step, in_names
+
+    def test_equivalence(self):
```
mcgibbon: Suggestion: `test_equivalence_with_(something)`, "single_step"?

fme/core/step/test_separated_module.py
```diff
@@ -0,0 +1,407 @@
...
+
+        torch.manual_seed(42)
+        separated_step = separated_config.get_step(dataset_info, lambda _: None)
+
+        # Copy weights from single to separated at the inner module level.
+        # SingleModuleStep wraps as DummyWrapper(SFNO) while
+        # SeparatedModuleStep wraps as DummyWrapper(LegacyWrapper(SFNO)),
+        # so we copy at the SFNO level to avoid key prefix mismatches.
+        single_sfno = single_step.module.torch_module.module
+        separated_sfno = separated_step.module.torch_module.module.inner
```
mcgibbon: Issue: I don't like digging into these internal properties, is it possible to do this with public API like the `.modules` attribute of StepABC?

mcgibbon:
I ended up going the route of enabling prognostic/forcing/output tensors with a new StepABC implementation. Thinking about it, it's just too much of a fundamental change in the sense that it makes no sense to run an existing checkpoint with the new code - the input and output packers are fundamentally changed, and there's no guarantee the variable name ordering for existing checkpoints is valid under the new system.

fme/core/registry/separated_module.py
mcgibbon: It's because of type being an attribute, so the type built-in is no longer accessible.

fme/core/registry/separated_module.py
mcgibbon: Answer: Good question. Right now we reuse it because the only way to get a conditional separated module is through LegacyModuleAdapter, which delegates to ModuleSelector, so the same builder types apply. When we register native separated modules that support conditioning, we should define an independent CONDITIONAL_BUILDERS list (or a different mechanism) for this registry. For now, reusing it is correct since LegacyModuleAdapter is the only registered type.

mcgibbon:
There are a bunch of weird questions for the Legacy wrapper, like conditional being set in both the inside and outside module selectors. I don't think we actually need this support - I'll just register a new separated builder when we have the local model that uses this feature. So for now this PR will just have a class/config used for testing.

fme/core/step/test_separated_module.py
mcgibbon: It wasn't really avoidable.

fme/ace/step/fcn3.py
```diff
@@ -390,7 +390,6 @@ def __init__(
         self.module = dist.wrap_module(module)
         self._img_shape = dataset_info.img_shape
         self._config = config
-        self._no_optimization = NullOptimization()
```
mcgibbon, line 389: This was unused.

fme/ace/inference/inference.py
```diff
@@ -305,7 +305,7 @@ def run_inference_from_config(config: InferenceConfig):
         logging.info("Loading initial condition data")
         initial_condition = get_initial_condition(
             config.initial_condition.get_dataset(),
-            stepper_config.prognostic_names,
+            stepper_config.get_prognostic_names(),
```
mcgibbon, line 308: This property method needed to be refactored to a getter so that `prognostic_names` could be a dataclass attribute instead, on the new Step.

fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Builds an nn.Module that takes separate forcing and prognostic tensors
+    as input, and returns separate prognostic and diagnostic tensors as output.
+
+    The built nn.Module must have the forward signature::
+
+        forward(
+            forcing: Tensor,
+            prognostic: Tensor,
+            labels: Tensor | None = None,
```
mcgibbon, line 29: I went with requiring all modules of this new type take in a labels argument, since there's no need for backwards compatibility with ones that don't. For network types that don't have conditioning, we can broadcast these to the domain size and just treat them as additional input variables.

fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,237 @@
...
+        ...
+
+    @classmethod
+    def from_state(cls, state: Mapping[str, Any]) -> "SeparatedModuleConfig":
+        return dacite.from_dict(
+            data_class=cls, data=state, config=dacite.Config(strict=True)
+        )
+
+
+class SeparatedModule:
```
mcgibbon, line 67: This handles conversion of BatchLabels to Tensor, and also provides a strongly typed interface (unlike nn.Module which is treated as Any by mypy). We do the same thing for Module.

fme/core/step/separated_module.py
```diff
@@ -0,0 +1,388 @@
...
+        next_step_forcing_names: Names of forcing variables for the next timestep.
+        prescribed_prognostic_names: Prognostic variable names to overwrite from
+            forcing data at each step (e.g. for inference with observed values).
+        residual_prediction: Whether to use residual prediction.
+    """
+
+    builder: SeparatedModuleSelector
+    forcing_names: list[str]
+    prognostic_names: list[str]
+    diagnostic_names: list[str]
```
mcgibbon: TODO: make forcing and diagnostic names optional, defaulting to an empty list, to make it clearer they are not required / can be an empty list.

fme/core/registry/separated_module.py
jpdunc23: Would it work to define `SeparatedModuleConfigType = type[SeparatedModuleConfig]` at the module level to avoid the clash?

fme/ace/inference/inference.py
jpdunc23: ~~What do you think about making this refactor a separate PR? Fairly minor change, but does touch a lot of files.~~

fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def __call__(
+        self,
+        forcing: torch.Tensor,
+        prognostic: torch.Tensor,
+        labels: BatchLabels | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        if labels is not None and self._label_encoding is not None:
+            encoded_labels = labels.conform_to_encoding(self._label_encoding)
+            return self._module(forcing, prognostic, labels=encoded_labels.tensor)
```
jpdunc23, line 96: Similar to `fme.core.registry.module.Module`, should this raise an error if `labels is not None and self._label_encoding is None` or vice versa?

Also, I think this branch is untested.

fme/core/step/separated_module.py
jpdunc23:
```suggestion
    prognostic_names: list[str]
    forcing_names: list[str] = dataclasses.field(default_factory=list)
    diagnostic_names: list[str] = dataclasses.field(default_factory=list)
```

fme/core/registry/separated_module.py
```diff
@@ -0,0 +1,237 @@
...
+        def build(
+            self,
+            n_forcing_channels,
+            n_prognostic_channels,
+            n_diagnostic_channels,
+            dataset_info,
+        ):
+            return _SimpleSeparatedModule(
+                n_forcing_channels, n_prognostic_channels, n_diagnostic_channels
+            )
```
jpdunc23: Could this can be moved to `fme/core/registry/test_separated_module.py`? I think so since we have existing tests where the registry is updated in the test file (e.g., see `MockStepConfig` in `fme/core/step/test_step_registry.py`). I think it would be cleaner to keep the testing modules isolated to the test files.

fme/core/step/separated_module.py
```diff
@@ -0,0 +1,388 @@
...
+    normalization: NetworkAndLossNormalizationConfig
+    secondary_decoder: SecondaryDecoderConfig | None = None
+    ocean: OceanConfig | None = None
+    corrector: AtmosphereCorrectorConfig | CorrectorSelector = dataclasses.field(
+        default_factory=lambda: AtmosphereCorrectorConfig()
+    )
```
jpdunc23, line 65: Since we don't need to support backwards compatibility I wish we could avoid this bit of atmosphere specificity:

```suggestion
    corrector: CorrectorSelector | None = None
```

I'm working on some refactors now to make this change possible, though I don't necessarily think this PR should be held up by it.

jpdunc23 (commented):
This is looking good. My main concern is the handling of combinations of missing labels and encoding.

When I heard you talking about this PR I had a somewhat different idea of what the changes would look like. I thought that you would implement a multi-module rather than single-module approach. The multi-module approach might be impractical for certain architectures, e.g. if you want a shared embedding layer for all inputs. But it would also be more readily able to take advantage of existing `nn.Module`s. The `StepABC` would just be responsible for piping the appropriate args to the appropriate `nn.Module` without any single `nn.Module` having to know about "business logic" (i.e., distinctions between forcing, prognostic, and diagnostic variables).

Anyways, the approach here is also reasonable.

fme/ace/inference/inference.py
jpdunc23: Feel free to ignore, I don't think it's really necessary to separate off.

mcgibbon: I also went back and forth on this. Will keep it.

fme/core/registry/separated_module.py
mcgibbon: Claude wanted to avoid one test file importing from another test file, which I agreed with. The issue being this is used in more than one test file.

fme/core/step/separated_module.py
mcgibbon: If you’re able to get those changes in soon, it wouldn’t be unreasonable to break backwards compatibility on this in the very near future before we have many checkpoints.

mcgibbon:
> without any single nn.Module having to know about "business logic" (i.e., distinctions between forcing, prognostic, and diagnostic variables).

The issue where this breaks down is that we want to use the architecture differently for these types of variables. For example, the way Nvidia does residual updates for prognostic variables in the Makani code, or the way forcing variables are used as context instead of normal inputs. These features are impossible to implement without the module knowing about this distinction. Well, maybe not impossible, the Step could do most of it, but these really do feel like architectural choices and not like physical choices.

I do think it’s good practice to continue using modules that take single tensors as much as possible when they don’t need to make this distinction, and concatenating to call them.

fme/core/step/separated_module.py
mcgibbon: At the least I think we can require CorrectorSelector now? I'm less clear on how we can run without an OceanConfig.

mcgibbon:
Mh as I'm working in tandem on the Local network, I'm finding an interesting question that is leading me to see the appeal of doing this stuff in Step... I'll need to think on it some more.

fme/core/registry/separated_module.py
jpdunc23: Perhaps a separate `fme/core/registry/testing.py` module?

mcgibbon: I think that would make sense.

fme/core/registry/separated_module.py
jpdunc23: Noting that the `if labels is not None and self._label_encoding is not None:` path is still untested. Maybe this could be covered in `fme/core/step/test_separated_module.py`?

fme/core/step/separated_module.py
jpdunc23: While it would be nice, requiring `CorrectorSelector` now would involve a bit of refactoring that probably isn't worth tackling in this PR.

jpdunc23 (approved):
Left a couple of new comments, but LGTM. There are left over references to `LegacyModuleAdapter` and `LegacyWrapper` in the PR description, so please be sure to remove before merge.

To some extent it feels like `fme/core/step/separated_module.py` copies a lot of boilerplate from `fme/core/step/single_module.py` when the differences come down to the config and the `network_calls` argument to `step_with_adjustments`. I think the two could be unified in a future refactor since `SeparatedModuleStep` is a more general version of `SingleModuleStep`.

PR #959, opened 2026-03-11
Refactor `CascadePredictor.generate` so generate methods can be private

AnnaKwa:
I made some changes to address this comment since they were fresh in my mind: https://github.com/ai2cm/ace/pull/954#discussion_r2920307878

This PR refactors CascadePredictor.generate to wrap previous step's output into a BatchData object so it can call the underlying model's generate_on_batch_no_target instead of generate.

This allows us to make the DiffusionModel.generate method private.

### DIFF
--- a/fme/downscaling/predictors/cascade.py
+++ b/fme/downscaling/predictors/cascade.py
@@ -2,6 +2,7 @@ import dataclasses
 import math
 
 import torch
+import xarray as xr
 
 from fme.core.coordinates import LatLonCoordinates
 from fme.core.device import get_device
@@ -15,6 +16,7 @@ from fme.downscaling.data import (
     adjust_fine_coord_range,
     scale_tuple,
 )
+from fme.downscaling.data.utils import BatchedLatLonCoordinates
 from fme.downscaling.metrics_and_maths import filter_tensor_mapping
 from fme.downscaling.models import CheckpointModelConfig, DiffusionModel, ModelOutputs
 from fme.downscaling.requirements import DataRequirements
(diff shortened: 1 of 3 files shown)


fme/downscaling/predictors/cascade.py
```diff
@@ -116,30 +138,34 @@ def modules(self) -> torch.nn.ModuleList:
...
+                n_samples_cascade_step,
             )
             generated = {
                 k: v.reshape(batch_size * n_samples_cascade_step, *v.shape[-2:])
                 for k, v in generated.items()
             }
             current_coarse = generated
         generated = _restore_batch_and_sample_dims(generated, n_samples)
-        return generated, generated_norm, latent_steps
+        return generated
```
frodre, line 168: Outputs need to be preserved for the `generate_on_batch` method of the cascade predictor, which complicates the use of non-public `_generate` for the underlying model.

fme/downscaling/predictors/cascade.py
```diff
@@ -86,6 +88,26 @@ def _restore_batch_and_sample_dims(data: TensorMapping, n_samples: int):
     return unfold_ensemble_dim(squeezed, n_samples)
 
 
+def _batch_data_with_unused_coords(data: TensorMapping) -> BatchData:
+    # wrapper function so that we can call each level's
```
frodre, line 92: Would be nice if this was some kind of subclass of `BatchData` or easily identifiable that this is not actual data.   But on second thought, if I am using the coarse coordinates to define the static input usage, this would break that contract.

frodre (commented):
Thanks, @AnnaKwa.  I think this is on the right track that we need to include more rich set of information with the outputs of the models (e.g., coordinates).  I ran into something related to the output information issue in the last PR of the static input refactor.  For the cascading code, there is a requirement for the `generate_on_batch` to have a `ModelOutput` with latent steps, etc., but that can't be provided with `_generate` as it's set up.

a) It's not clear whether  we'll use `CascadePredictor` again, so we could just opt to deprecate this model and remove it from the options of inference models and stop updating it for now

--or--

b) generate stays public in order to satisfy the output needs for `generate_on_batch`, but we will still need to include real coordinate information for the static_input handling within the model

I'm open to a) since we haven't been using these types of models recently.

AnnaKwa:
> a) It's not clear whether we'll use `CascadePredictor` again, so we could just opt to deprecate this model and remove it from the options of inference models and stop updating it for now

I'm also fine with this. I'll close this PR and make an issue to deprecate this code.

PR #961, opened 2026-03-11
Refactor `GlobalTimer` to support hierarchical outer/inner timing

jpdunc23:
`GlobalTimer` has "outer" and "inner" timer concepts, but inner timers share a flat namespace regardless of which outer timer they belong to. This makes granular timing of multiple tasks in the same job difficult because inner timer categories like `data_loading`, `aggregator`, and `wandb_logging` collide. 

This PR reverses the current constraints in the `GlobalTimer` API:
- Before: multiple outer timers can overlap; only one inner timer at a time.
- After: only one outer timer at a time; multiple inner timers can overlap.

Inner timers are namespaced under the active outer timer:
- With outer timer "inference" active: `timer.context("data_loading")` creates a timer keyed "inference/data_loading".
- Without an outer timer: `timer.context("data_loading")` creates a timer keyed "data_loading" (backward compatible for callers like the benchmark and trainer that don't use outer timers).

The outer timer itself still gets its own cumulative timer entry (e.g., "inference").

Changes:
- `GlobalTimer.stop()` now has a required `category: str` arg.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -465,8 +465,6 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
         f"{total_steps_per_second:.2f} steps/second"
     )
 
-    breakpoint()
-
     summary_logs = {
         "total_steps_per_second": total_steps_per_second,
         **timer.get_durations(),


mcgibbon:
I think a more drastic change might be warranted, to remove the limitation that there's exactly 1 level of depth in timings - how about all contexts become ones that could have new contexts inside of them, and the global timer tracks up and down the chain in a hierarchy of timers? This worked well with the benchmark timers, perhaps GlobalTimer could track a reference to the "current" timer which changes to a child when context (analog of `timer.child`) is called, and goes back to the parent when the child is exited, using a list / LIFO queue? To make sure things stay consistent, i'd suggest this only work with `context` based timing, not with start/stop.

Even with multi-depth timings, I think it's still a good idea to limit depth for the logs, using something like what's in the benchmark Timer.

What do you think? Does that sound worth it, or should I more thoroughly review the changes here? We could always do this and then follow-on with that, if this unblocks your current issue.

jpdunc23:
> I think a more drastic change might be warranted, to remove the limitation that there's exactly 1 level of depth in timings - how about all contexts become ones that could have new contexts inside of them, and the global timer tracks up and down the chain in a hierarchy of timers? This worked well with the benchmark timers, perhaps GlobalTimer could track a reference to the "current" timer which changes to a child when context (analog of `timer.child`) is called, and goes back to the parent when the child is exited, using a list / LIFO queue? To make sure things stay consistent, i'd suggest this only work with `context` based timing, not with start/stop.
> 
> Even with multi-depth timings, I think it's still a good idea to limit depth for the logs, using something like what's in the benchmark Timer.
> 
> What do you think? Does that sound worth it, or should I more thoroughly review the changes here? We could always do this and then follow-on with that, if this unblocks your current issue.

Thanks, I think it would be nice to allow more deeply nested categories. I haven't yet looked closely at how the benchmark timers work, but I agree that what you're proposing sounds like it could be done as a natural extension to in a follow-on PR. 

So if it sounds good to you I'd say let's review this more limited refactor as-is to unblock the generic `run_validation` timing.

fme/ace/inference/inference.py
```diff
@@ -369,7 +369,7 @@ def run_inference_from_config(config: InferenceConfig):
     timer.stop_outer("inference")
     total_steps = config.n_forward_steps * data.n_initial_conditions
     inference_duration = timer.get_duration("inference")
-    wandb_logging_duration = timer.get_duration("wandb_logging")
+    wandb_logging_duration = timer.get_duration("inference/wandb_logging")
     total_steps_per_second = total_steps / (inference_duration - wandb_logging_duration)
```
mcgibbon, line 373: This is fine (and pre-existing), but noting later/in your follow on it would be nice if this directly retrieved an inference timing, instead of subtracting everything that isn't inference from a total timing.

fme/core/test_timing.py
```diff
@@ -188,12 +188,66 @@ def test_GlobalTimer_nested_outer_context():
...
+            with timer.context("work"):
+                time.sleep(0.01)
+        with timer.outer_context("second"):
+            with timer.context("work"):
+                time.sleep(0.01)
+    durations = timer.get_durations()
+    assert "first" in durations
+    assert "first/work" in durations
+    assert "second" in durations
+    assert "second/work" in durations
```
mcgibbon, line 255: Suggestion (optional): May want to add some basic assertions about the relationships between these timings, e.g. that the work total is less than or equal to the outer context totals (would catch a hypothetical bug where they silently use the same timing but with different labels).

fme/core/timing.py
```diff
@@ -71,13 +80,28 @@ def __exit__(cls, type, value, traceback):
     def __init__(self):
         self._timers: dict[str, CumulativeTimer] = {}
         self._active = False
-        self._current_category: str | None = None
+        self._current_outer: str | None = None
+        self._current_inner: str | None = None
+
+    def _inner_key(self, category: str) -> str:
+        if self._current_outer is not None:
```
mcgibbon, line 87: Question: Do we allow / want to be allowing the case where an inner timer is defined without an outer timer encapsulating it?

jpdunc23: Yes, I think so. After merging `main` you can see how I'm now using inner timers without an outer timer in order to maintain the status quo of the timing keys in `run_evaluator_from_config`. I also had to allow multiple inner timers running simultaneously, which I think is a useful feature in general (similar to how `start_outer` previously worked), but it does require passing `category: str` to `GlobalTimer.stop()`.

The change to allow simultaneous inner timers was necessary to avoid changing the inference keys of the `timer.get_durations()` dict (no nesting, e.g., "inference", "wandb_logging", and "validation"). 

I plan to switch to nested keys like "inference/wandb_logging", "inference/aggregator", etc. and add more granular validating loop timing ("validation/wandb_logging", "validation/aggregator", etc.) in https://github.com/ai2cm/ace/pull/963

mcgibbon: Okie doke. This should disappear anyways if we move to arbitrarily-deeply-nested timers.

PR #962, opened 2026-03-12
Expand variable set and supported resolutions of ERA5 pipeline

spencerkclark:
This PR adds a number of new variables to our ERA5 data processing pipeline.  It also enables processing to a 4° Gaussian grid.  A summary of the added variables and other minor updates can be found in [this slide deck](https://docs.google.com/presentation/d/1BDCjYc_uk_hFnmmIRFhcieXvEFX5AglxmXhYlBbaoLU/edit?slide=id.g2ed0744072d_0_124#slide=id.g2ed0744072d_0_124).

Changes:
- Fixes a bug where attributes were not propagated through regridding, since this does not occur by default with xESMF.
- Enables processing to an `"F22.5"` grid, which corresponds to a 45 by 90 Gaussian grid.
- Adds a `--check-data-validity` flag.  If used, all raw variables touched will first be checked for the presence of NaNs and an error will be raised if an unexpected amount are found.
- Removes the unnecessary [template chunking step](https://github.com/ai2cm/ace/blob/e709930a0b85ce7113090016c4bbd9849feeb8f9/scripts/era5/pipeline/xr-beam-pipeline.py#L737) which otherwise blows up memory when scaling up the workflow by creating a large dask graph.
- Doubles the memory of the Dataflow workers, since we were running out of memory when running the production 1° workflow.
- Adds several new raw and derived variables.
  - `"sea_surface_temperature"`
  - `"merged_sea_surface_and_skin_temperature"`
  - `"undefined_soil_type_fraction"`
  - `"coarse_soil_type_fraction"`
  - `"medium_soil_type_fraction"`
  - `"medium_fine_soil_type_fraction"`
  - `"fine_soil_type_fraction"`
  - `"very_fine_soil_type_fraction"`
  - `"organic_soil_type_fraction"`
  - `"tropical_organic_soil_type_fraction"`
  - `"soil_temperature_0"`
  - `"soil_temperature_1"`
  - `"soil_temperature_2"`
  - `"soil_temperature_3"`
  - `"surface_snow_amount"`
  - `"surface_snow_area_fraction"`
  - `"surface_snow_thickness"`
  - `"runoff_flux"`
  - `"total_frozen_precipitation_rate"`
  - `"UCSWRFtoa"`
  - `"UCLWRFtoa"`
  - `"DCSWRFsfc"`
  - `"UCSWRFsfc"`
  - `"DCLWRFsfc"`
  - `"UCLWRFsfc"`
  - `"eastward_surface_stress"`
  - `"nortward_surface_stress"`
  - `"PRMSL"`
  - `"significant_height_of_combined_wind_waves_and_swell"`
  - `"Q"`, `"TMP"`, `"UGRD"`, and `"VGRD"` at 1000, 700, 250, 100, 50, and 10 hPa
  - `"h"` at 100, 50, and 10 hPa
  - `"global_mean_co2"` extended through 2026-07-01.

### DIFF
--- a/fme/ace/inference/data_writer/test_data_writer.py
+++ b/fme/ace/inference/data_writer/test_data_writer.py
@@ -349,6 +349,7 @@ class TestDataWriter:
         [
             pytest.param(None, id="None"),
             pytest.param(["temp", "humidity", "pressure"], id="subset"),
+            pytest.param(["temp", "missing_var"], id="missing_var"),
         ],
     )
     def test_append_batch_save_names(
--- a/fme/ace/inference/data_writer/utils.py
+++ b/fme/ace/inference/data_writer/utils.py
@@ -1,3 +1,4 @@
+import warnings
 from collections.abc import Iterable
 from dataclasses import dataclass
 from typing import TypeVar
@@ -17,8 +18,13 @@ def get_all_names(
         variables = variables.union(set(varnames))
     if allowlist is None:
         return variables
-    else:
-        return variables.intersection(set(allowlist))
+    allowlist = set(allowlist)
+    missing = allowlist - variables
+    if missing:
+        warnings.warn(
+            f"Requested variable(s) not present in data: {list(missing)}",
+        )
+    return variables.intersection(allowlist)
 
 
 @dataclass
--- a/fme/ace/registry/stochastic_sfno.py
+++ b/fme/ace/registry/stochastic_sfno.py
@@ -11,6 +11,7 @@ from fme.core.distributed.distributed import Distributed
 from fme.core.models.conditional_sfno.sfnonet import (
     Context,
     ContextConfig,
+    SFNONetConfig,
     get_lat_lon_sfnonet,
 )
 from fme.core.models.conditional_sfno.sfnonet import (
@@ -145,10 +146,10 @@ class NoiseConditionedSFNOBuilder(ModuleConfig):
     Noise is provided as conditioning input to conditional layer normalization.
 
     Attributes:
-        spectral_transform: Type of spherical transform to use.
-            Kept for backwards compatibility.
+        spectral_transform: Unused, kept for backwards compatibility only.
         filter_type: Type of filter to use.
-        operator_type: Type of operator to use. Only "dhconv" is supported.
+        operator_type: Unused, kept for backwards compatibility only.
+            Must be "dhconv".
         residual_filter_factor: Factor by which to downsample the residual.
         embed_dim: Dimension of the embedding.
         noise_embed_dim: Dimension of the noise embedding.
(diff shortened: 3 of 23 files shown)


scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -261,6 +356,95 @@ def _to_geopotential_height(geopotential: xr.DataArray) -> xr.DataArray:
...
+    """
+    output = DENSITY_OF_LIQUID_WATER * snow_depth
+    output.attrs = {"long_name": "Surface snow amount", "units": "kg/m**2"}
+    return output
+
+
+def _to_surface_snow_area_fraction(
+    snow_depth: xr.DataArray, snow_density: xr.DataArray
+) -> xr.DataArray:
+    """See Guidelines Section 11 on this page for the ERA5 formula for the
```
brianhenn, line 405: Interesting that a specific relationship between snow amount and snow cover is assumed here (this can vary greatly depending on topography, vegetation, etc.). But presumably this matches what the IFS's land surface model does.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -261,6 +356,95 @@ def _to_geopotential_height(geopotential: xr.DataArray) -> xr.DataArray:
...
 
 
+def _to_merged_sea_surface_and_skin_temperature(
+    sea_surface_temperature: xr.DataArray,
+    skin_temperature: xr.DataArray,
+    ocean_fraction: xr.DataArray,
+) -> xr.DataArray:
+    """Merge the sea surface and skin temperature into a single variable.
+
+    Note this is meant to be called after regridding, which is the only time we
```
brianhenn, line 371: Presumably it'll still be possible to determine which temperature is used for a given cell after postprocessing, from the SST and skin temperature variables as well as from the ocean_fraction variable?

brianhenn (commented):
Thanks, looks like this all of the variables I was looking for. Is the plan to update the end date of the dataset in a subsequent PR or should we do that here?

scripts/era5/pipeline/xr-beam-pipeline.py
spencerkclark: Indeed, since I am merging after regridding, it will be possible to exactly reconstruct the `(ocean_fraction < 0.5) | sea_surface_temperature.isnull()` condition from variables in the output dataset.

scripts/era5/pipeline/xr-beam-pipeline.py
spencerkclark: Yes, clearly quite heuristic, but I take ECMWF's word that this is how they compute it.

Image

spencerkclark:
> Is the plan to update the end date of the dataset in a subsequent PR or should we do that here?

Yes, I will update `run-dataflow.sh` once we settle on the final workflow code.  The expected bounds will be `1940-01-01T12:00:00` to `2025-12-31T18:00:00`.

brianhenn:
It looks like the end of 2025 should be [available](https://github.com/google-research/arco-era5) in in early April? I am assuming we want the regular version of the dataset and not the preliminary "T" version:
```
ds=xr.open_zarr('gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3')
ds.attrs.get('valid_time_stop')
# out: '2025-11-30'
ds.attrs.get('valid_time_stop_era5t')
# out: '2026-03-06'
```

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -385,11 +602,13 @@ def process_mean_flux(key, ds, output_grid=DEFAULT_OUTPUT_GRID):
 
 
 def _process_invariant(ds: xr.Dataset, output_grid: str) -> xr.Dataset:
-    """Process invariant fields (land_sea_mask, geopotential_at_surface)."""
+    """Process invariant fields (land_sea_mask, geopotential_at_surface, soil_type)."""
     logging.info("Processing invariant data")
     output = xr.Dataset()
     output["HGTsfc"] = ds["geopotential_at_surface"] / GRAVITY
     output["land_fraction"] = ds["land_sea_mask"]
+    soil_type_fractions = _to_soil_type_fractions(ds["soil_type"])
```
Arcomano1234, line 651: I did just quickly check and it seems there are not trivial amounts of grid cells at the 0.25 deg dataset that have soil types that changed during the full record of ERA5. I am not sure if this is expected

Arcomano1234: I looked into this a little bit further there seems to be ~118 grid points that are problematic and have different values depending on the year. For example in 2020 they have a certain value and in 2024 they have a different one.

Arcomano1234: See [notebook]( https://github.com/ai2cm/explore2/blob/main/troya/2026-03-13-ERA5-refactor/2026-03-13-soil-type-invariance.ipynb)

spencerkclark: Interesting, thanks for looking into that.  That is somewhat unexpected, since it shows up as an invariant field in [Table 1 on this documentation page](https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation#ERA5:datadocumentation-Table1) along with the `land_sea_mask` and `geopotential_at_surface` and others.  

I agree it is probably not worth worrying about, given how few fine grid cells this affects (0.01 %) and that they seem relatively scattered.  Oddly the `geopotential_at_surface` also changes subtly between those two dates as well; the implied surface height at the South Pole is about 35 m higher in 2024 than in 2020.

Arcomano1234: Yeah I suspect this is not an ERA5 problem but related to the pre-processing done by ARCO. But as you said, either way its probably not worth worrying about.

spencerkclark:
Both the 1° and 4° workflows have now succeeded, so I am going to go ahead and merge this PR.

PR #963, opened 2026-03-12
Add `WandBStepLogger` for inference logging

jpdunc23:
`get_record_to_wandb(label="inference")` in `fme/core/generics/inference.py` returns a closure that couples two concerns:
- Step counting -- monotonically incrementing the wandb step
- Key prefixing -- prepending "inference/" to every log key

This PR replaces the closure with a class `WandBStepLogger` that separates step tracking from prefixing. `WandBStepLogger` has a `log` method that replaces the previous closure, with support for more flexible key labeling. In addition, `WandBStepLogger.log_to_current_step` logs to wandb without incrementing the step, allowing for a bit more flexibility when logging summary metrics.

Changes:
- Inference timing metrics are now prefixed with "inference/", e.g. "inference/wandb_logging". The overall timing key "inference" is unchanged.
- Validation within `fme.ace` inference evaluator jobs now logs granular timing metrics like "validation/wandb_logging".

- [x] Tests added

### DIFF
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -472,5 +472,7 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
         "total_steps_per_second": total_steps_per_second,
         **aggregator.get_summary_logs(),
     }
-    logger.log_to_step(summary_logs)  # prefix "inference/"
-    logger.log_to_step(timer.get_durations(), label="")  # durations already prefixed
+    logger.log_to_current_step(summary_logs)  # prefix "inference/"
+    logger.log_to_current_step(
+        timer.get_durations(), label=""
+    )  # durations already prefixed
--- a/fme/ace/inference/inference.py
+++ b/fme/ace/inference/inference.py
@@ -380,8 +380,8 @@ def run_inference_from_config(config: InferenceConfig):
         "total_steps_per_second": total_steps_per_second,
         **aggregator.get_summary_logs(),
     }
-    logger.log_to_step(summary_logs)
-    logger.log_to_step(timer.get_durations(), label="")
+    logger.log_to_current_step(summary_logs)
+    logger.log_to_current_step(timer.get_durations(), label="")
 
 
 def run_segmented_inference(config: InferenceConfig, segments: int):
--- a/fme/core/generics/inference.py
+++ b/fme/core/generics/inference.py
@@ -91,19 +91,19 @@ class WandBStepLogger:
     ) -> dict[str, Any]:
         if label is None:
             label = self._label
-        if label:
+        if label:  # not None and not ""
             log_dict = {f"{label}/{k}": v for k, v in log_dict.items()}
         return log_dict
 
     def log(self, logs: InferenceLogs, label: str | None = None) -> None:
         """Log each step in a sequence of logs."""
-        for j, log_dict in enumerate(logs):
-            if len(log_dict) > 0:
-                log_dict = self._prefix_label(log_dict, label)
-                self._wandb.log(log_dict, step=self._step + j)
-        self._step += len(logs)
+        for log_dict in logs:
+            self.log_to_current_step(log_dict, label)
+            self._step += 1
 
-    def log_to_step(self, log_dict: dict[str, Any], label: str | None = None) -> None:
+    def log_to_current_step(
+        self, log_dict: dict[str, Any], label: str | None = None
+    ) -> None:
         """Log to the current step without incrementing."""
         if len(log_dict) > 0:
             log_dict = self._prefix_label(log_dict, label)
--- a/fme/coupled/inference/evaluator.py
+++ b/fme/coupled/inference/evaluator.py
@@ -408,5 +408,5 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
         "total_steps_per_second": total_steps_per_second,
         **aggregator.get_summary_logs(),
     }
-    logger.log_to_step(summary_logs)
-    logger.log_to_step(timer.get_durations(), label="")
+    logger.log_to_current_step(summary_logs)
+    logger.log_to_current_step(timer.get_durations(), label="")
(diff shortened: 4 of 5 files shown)


fme/ace/inference/inference.py
```diff
@@ -351,13 +351,13 @@ def run_inference_from_config(config: InferenceConfig):
...
     logging.info("Starting inference")
-    record_logs = get_record_to_wandb(label="inference")
+    logger = get_record_to_wandb(label="inference")
     run_inference(
         predict=stepper.predict_paired,
         data=data,
         writer=writer,
         aggregator=aggregator,
-        record_logs=record_logs,
+        record_logs=logger,
```
mcgibbon: Suggestion: It took me a minute to realize that the API of run_inference hasn't changed, and that it only needs the __call__ functionality. ~~I'd suggest making this call method a `record_logs` method instead, and explicitly passing `logger.record_logs` here to make it clear that only this subset of the logger is getting passed/used.~~

On reviewing the WandBStepLogger, I'd suggest deleting the `__call__` method and instead passing
```suggestion
        record_logs=logger.log,
```
This also satisfies the API (I believe), and sets the label.

fme/core/generics/inference.py
```diff
@@ -69,20 +69,40 @@ def get_prognostic_state(self) -> PS:
         return self._prognostic_state
 
 
-def get_record_to_wandb(label: str = "") -> Callable[[InferenceLogs], None]:
-    wandb = WandB.get_instance()
-    step = 0
-
-    def record_logs(logs: InferenceLogs):
```
mcgibbon, line 76: Comment: I'm ambivalent to this, but I wanted to note you could keep this in mostly the form it was already in by adding a `label: str = None` argument to this returned helper, and renaming the get_record_to_wandb arg to `default_label`. It makes it a bit clearer that this "class" is really one method with some inner state, though a class with one method might or might not feel more comfortable.

fme/coupled/inference/evaluator.py
```diff
@@ -404,9 +404,12 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
...
+        **{f"inference/{k}": v for k, v in aggregator.get_summary_logs().items()},
+    }
     summary_logs = {
-        "total_steps_per_second": total_steps_per_second,
         **timer.get_durations(),
-        **aggregator.get_summary_logs(),
+        **inference_summary,
     }
-    record_logs([summary_logs])
+    logger.log([summary_logs], label="")
```
mcgibbon: It's not obvious to me what change this causes in our logs. Can you update the PR description to make this clearer?

fme/ace/inference/evaluator.py
```diff
@@ -352,9 +352,12 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
...
+        **{f"inference/{k}": v for k, v in aggregator.get_summary_logs().items()},
+    }
     summary_logs = {
-        "total_steps_per_second": total_steps_per_second,
         **timer.get_durations(),
-        **aggregator.get_summary_logs(),
+        **inference_summary,
     }
-    record_logs([summary_logs])
+    logger.log([summary_logs], label="")
```
mcgibbon: Suggestion (optional and/or follow-on): This 4x-duplicated code could be a helper method somewhere, or pulled into `run_inference` if that were to make sense.

fme/ace/inference/evaluator.py
```diff
@@ -370,35 +370,38 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
...
-        data_requirements = stepper_config.get_evaluation_window_data_requirements(
-            n_forward_steps=config.validation.get_n_forward_steps()
-        )
-        valid_data = get_gridded_data(
-            config.validation.loader,
-            requirements=data_requirements,
-            train=False,
-        )
+        timer.stop_outer("inference")
+        timer.start_outer("validation")
```
mcgibbon, line 374: Suggestion (optional or follow up): context blocks would make it clearer how these are nested, especially given the multiple levels of nesting involved here. If that makes the function too deeply nested, it's a reflection of needing some helper functions.

Probably ignore this for this PR, but consider doing refactor PRs when you encounter a lot of timer.stop/start calls in a function.

fme/core/generics/inference.py
```diff
@@ -69,20 +69,49 @@ def get_prognostic_state(self) -> PS:
...
+            if len(log_dict) > 0:
+                log_dict = self._prefix_label(log_dict, label)
+                self._wandb.log(log_dict, step=self._step + j)
+        self._step += len(logs)
+
+    def log_to_step(self, log_dict: dict[str, Any], label: str | None = None) -> None:
+        """Log to the current step without incrementing."""
+        if len(log_dict) > 0:
+            log_dict = self._prefix_label(log_dict, label)
+            self._wandb.log(log_dict, step=self._step)
```
mcgibbon: Suggestion (optional): It feels like `log` should call `log_to_step` with `self._step += 1` within the loop over `j`, to avoid the two duplicated lines.

mcgibbon: Suggestion (optional): name `log_to_current_step` to make it clearer that while `log` increments the steps, `log_to_current_step` does not. If you don't think this is clearer, feel free to ignore.

fme/ace/inference/evaluator.py
jpdunc23: Agreed, I think the context blocks are a great feature of `GlobalTimer` so thanks for including that. The `stop_outer`/`start_outer` API sadly doesn't currently support `with` context blocks for outer timers, so this would require a `GlobalTimer` API addition, but I think a worthwhile one in a future PR.

fme/core/generics/inference.py
jpdunc23: The `WandBStepLogger` API is a bit more meaty now.

fme/core/generics/inference.py
jpdunc23: Thanks, much cleaner now. I like `log_to_current_step`.

PR #964, opened 2026-03-12
Add code and review guidelines to CONTRIBUTING.md and AGENTS.md

mcgibbon:
Document the team's coding standards and review conventions derived from reviewing PR history.

Key additions: responsibility isolation principles, polymorphism over isinstance chains, config validation, testing practices, naming conventions, and review comment labeling.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,11 +45,6 @@ single-rank `python -m pytest` run, then verify under spatial parallel via
 torchrun. Generating baselines under the same backend you test against does
 not validate cross-backend correctness.
 
-### GitHub MCP Server Setup
-
-To use the PR review rules, configure the GitHub MCP server in Cursor's "Tools & MCP" settings.
-You will need a read-only personal access token with the following permissions: Pull requests, Issues, Contents, Metadata.
-
 ## Code Guidelines for Agents
 
 ### Naming


mcgibbon:
IDK if this is detail we want to include, but I was gathering this information for my own benefit and found it pretty helpful. We can discuss whether to merge or not.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
...
+
+### Code Organization
+
+- Consolidate duplicated code to shared locations (e.g. `fme/core/`).
+- Remove unused code, flags, and imports proactively.
+- Use `if condition: raise` instead of `assert` in production code (asserts
+  can be stripped by `python -O`).
+- Use context managers for resource cleanup (timers, distributed contexts).
+- Pass composed objects rather than their parts (e.g. `stepper.loss_scaling`
+  instead of two separate arguments).
```
mcgibbon: Prefer to keep this example abstract in case the code changes, so this doesn't go out-of-date, or maybe remove the example altogether.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
...
+  misconfigurations before jobs are submitted.
+- Don't add config options that duplicate existing ones.
+- Remove unused fields rather than leaving them around.
+- Use deprecation warnings rather than hard errors when removing config
+  options, unless a breaking change has been communicated to the team.
+
+### Testing
+
+- **Test helpers over copy-paste.** Create helper functions to build common
+  test fixtures. If the same setup appears 3+ times, extract a helper.
```
mcgibbon: Also worth noting we want to avoid using pytest fixtures, use helpers instead.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
...
+- **Demonstrate bugs with failing tests.** When fixing a bug, add a test
+  that fails without the fix, then fix it.
+- **Test behavior, not implementation.** If a test re-implements the logic
+  it's testing, it isn't actually verifying anything.
```
mcgibbon: Also important to have tests that cover important user-story-level details, rather than a lot of tests covering subjective API details, since those make it harder to change those details later and should be used in moderation. Not to say we can't have both, though.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
...
+new class. This avoids strange intermediate states where partially-implemented
+features block on breaking YAML changes.
+
+### Naming
+
+- Names should accurately describe behavior. "scatter" implies inter-process
+  communication; use "localize" when each rank computes its own local view.
+- Config class names: append `Config` to the name of the thing being built
+  (e.g. `TrainStepperConfig` builds `TrainStepper`).
+- Prefer descriptive names over abbreviations (`noise_distribution` not
```
mcgibbon: Worth noting descriptions should for the most part only include information that's available in the context of its scope, for example we would avoid calling a function `normalize_loss(loss: Tensor)` if its internals have nothing to do with loss and could normalize anything - instead it would be `normalize(x: Tensor)`.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
...
+
+### Design Principles
+
+#### Isolate responsibilities to as few abstraction levels as possible
+
+When designing a change, think first about what absolutely must change, then
+about what level of abstraction that change could be handled in. Choose the
+option that splits the concern across the fewest levels of abstraction. When a
+decision changes in the future, as little code as possible should need to be
+touched.
```
mcgibbon: Worth noting this does not mean you should be changing as few lines as possible when putting in the feature. You might need to modify APIs at several levels to isolate the feature into one level of abstraction.

AGENTS.md
```diff
@@ -31,6 +31,70 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
...
+  (`dist.require_no_model_parallelism(msg)`) over scattered if-else checks.
+- **Training concerns stay in training code.** DDP wrapping, weight freezing,
+  and loss config should not leak into inference-capable code paths.
+- **Facade pattern for multi-PR refactors.** Implement the new class
+  internally, keep the old class as a translation layer, swap in a final PR.
+
+### Naming
+
+- Config classes: append `Config` to the built type (`TrainStepperConfig`).
+- Prefer descriptive names (`noise_distribution` not `distribution`).
```
mcgibbon: Similar thing here, avoid description that isn't available in the present context / level of abstraction. Names shouldn't be decided based on the caller's context, unless there's a good reason it makes the code more readable.

AGENTS.md
```diff
@@ -31,6 +31,70 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
...
+- When fixing a bug, add a failing test first.
+- Tests must test behavior, not re-implement logic.
+- Use `xfail` for known bugs, not silent skips.
+- Use non-trivial values (not all-ones) so tests exercise real behavior.
+
+### Code organization
+
+- `if/raise` instead of `assert` in production code.
+- Context managers for cleanup (timers, distributed contexts).
+- Pass composed objects, not their parts.
```
frodre: Maybe add the specification "if multiple attributes would be use within the function".  Something to highlight that I shouldn't pass an object just to grab a single value from it.

CONTRIBUTING.md
jpdunc23: Agreed that fixtures can quickly become unwieldy, but can also be very useful e.g. for sharing scope across tests.

CONTRIBUTING.md
```diff
@@ -26,16 +26,119 @@ If you're an external user and want to implement a new feature, **please open an
 issue first to discuss it with us**. This helps ensure the feature aligns with
 the project's direction and prevents wasted effort.
 
+## Code Guidelines
```
jpdunc23: For much of the general advice below, rather than commit to developing and maintaining our own set of detailed coding guidelines and design principles, I think it would be better to simply include links to external style guides and guidelines that match our expectations for the quality of contributions from sources that we know aren't going anywhere. 

This way the contributing guidelines focus on things that are specific to `fme`.

AGENTS.md
```diff
@@ -31,6 +31,70 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
...
+- Commit vendorized code unmodified first, then modifications separately.
+
+### Review comment conventions
+
+One guide our team finds helpful is [Conventional Comments](https://conventionalcomments.org/).
+Label every comment with a severity prefix:
+- **Issue**: must fix before merge.
+- **Suggestion (optional)**: non-blocking improvement.
+- **Question**: seeking clarification.
+- **nit**: minor style; does not need re-review.
```
jpdunc23: This duplicates similar content added to `CONTRIBUTING.md`

AGENTS.md
```diff
@@ -31,6 +31,70 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
...
+
+- Config classes: append `Config` to the built type (`TrainStepperConfig`).
+- Prefer descriptive names (`noise_distribution` not `distribution`).
+- "scatter" implies communication; "localize" when no communication occurs.
+- Private functions get a `_` prefix.
+
+### Config design
+
+- Validate in `__post_init__`, not at runtime.
+- No redundant options; remove unused fields.
```
frodre: Do we want to have any guidance on backwards compatiblity for configuration parameters of training/eval/inference?  It slightly conflicts if we remove top-level configuration fields instead of deprecation warnings.  Not that this is a frequent occurence.

CONTRIBUTING.md
frodre: Yeah, I like tests for lower level operations with tricky logic, but once those are set they kind of feel like dead code compared to the user-story-level details of using the higher-level objects to perform ops.

CONTRIBUTING.md
frodre: Like @brianhenn mentioned, this feels overly verbose for CONTRIBUTING.md since the wall of text may dissuade humans from reading it!  I like the sort of FAQ type high-level stuff under the first abstraction section.  I'd vote for keeping that and the testing section since those would be most broadly relevant.

frodre (commented):
Thanks for putting this forward @mcgibbon.

AGENTS.md
```diff
@@ -31,6 +31,70 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
...
+- Backwards compatibility for checkpoint loading is critical; use deprecation
+  warnings for config removal.
+
+### Testing
```
oliverwm1, line 60: I think we should indicate a preference for fast-running and parsimonious tests.

frodre:
>As a general rule, I think we should try to keep AGENTS.md as lightweight as possible to avoid filling up context with things that may not be relevant or may even distract from the task at hand. 

I've also seen references to a [research](https://www.engineerscodex.com/agents-md-making-ai-worse) suggesting that machine generated context files or overly verbose/convoluted context files end up hurting agent performance rather than helping it.  I've seen suggestion that targeted instructions / tool specifications usually lead to better outcomes, but also you actually have to evaluate whether it's helping.  Iike @jpdunc23 said, it would be useful to know what of the updates were borne out of agent mistakes

CONTRIBUTING.md
mcgibbon: Do you have examples of external style guides to include? These came specifically from looking at our code reviews, rather than an external reference, so I don't have one (except for conventional comments).

I do feel like while some of these are general guidelines, they are ones that are specifically relevant to our code. It's hard to use a lot of CS-focused style guidelines because the nature of our code (our being the end user and developer) means a lot of the rules can be bent or broken. I find it hard in using them to pick out the "right" ones.

For example, isolating responsibilities to one abstraction level is specifically relevant to us because the freedom to make changes to our code in the future is high-value (because our requirements constantly change).

mcgibbon: I think I agree. As a first pass, I'll keep those, but move everything else (including the part about refactoring using facades) to AGENTS.md and add a reference to it from here.

It's hard because I feel like this is text all humans should read >.< but I do agree with you.

AGENTS.md
mcgibbon: Added below.

AGENTS.md
mcgibbon: Added

CONTRIBUTING.md
mcgibbon: Updated.

mcgibbon: Updated.

CONTRIBUTING.md
mcgibbon: Updated (see extension of bullet).

mcgibbon:
Ready for another iteration if anyone has feedback.

jpdunc23:
I agree with @brianhenn that there is a lot of nice content here that would be good to preserve somewhere. One possibility is to add a new page to the docs which we can link to from `CONTRIBUTING.md`.

mcgibbon:
I'm going to do another pass of the AGENTS.md based on the [research](https://www.engineerscodex.com/agents-md-making-ai-worse) @frodre linked. Basically, I'll treat added constraints as further constraints the agents will need to follow, which might make their job harder. I'll focus on adding constraints that actually are things we want to specify (e.g. we want to avoid large numbers of tests of low-level APIs), as well as lines that specify tooling that is available if the agent so chooses to use it (like our regression tools in `fme/core/testing/regression.py`. I'll also try to remove constraints that are things I expect the agent to already do (e.g. not writing tests that re-implement logic, or using if/raise instead of assert in production code).

Another option is to hold off entirely on the updates to AGENTS.md and just merge the CONTRIBUTING.md updates for now, continuing to work on a PR for the AGENTS changes.

One thing I do want to add but will hold off on right now is to tell agents to handle larger tasks by 1) planning out the full change, 2) splitting the plan into a series of unit-tested commits of individual changes/features that are easier to review, and 3) implementing those commits one by one. I have been doing this manually in each new session I make, and it's been extremely helpful in having Claude do larger PRs. Without this instruction, Claude has a tendency to tackle the entire change in one mega-commit, which I find I can't review well. Part of me wonders if some things like this should go in CLAUDE.md instead of AGENTS.md, I don't know how Cursor makes use of that file.

AGENTS.md
mcgibbon: I'm removing this for now in light of @frodre 's article, because I've so far found Claude is very good at writing fast-running tests already.

AGENTS.md
```diff
@@ -29,7 +29,34 @@ Pre-commit hooks run ruff, ruff-format, and mypy. If ruff-format modifies files,
 
 ### GitHub MCP Server Setup
 
-To use the PR review rules, configure the GitHub MCP server in Cursor's "Tools & MCP" settings. You will need a read-only personal access token with the following permissions: Pull requests, Issues, Contents, Metadata.
+To use the PR review rules, configure the GitHub MCP server in Cursor's "Tools & MCP" settings.
+You will need a read-only personal access token with the following permissions: Pull requests, Issues, Contents, Metadata.
```
mcgibbon: Updated the file to follow "one sentence per line" formatting, which it already mostly followed.

jpdunc23: We should probably just remove this section since it isn't relevant to agents.

AGENTS.md
```diff
@@ -47,7 +47,34 @@ not validate cross-backend correctness.
 
 ### GitHub MCP Server Setup
 
-To use the PR review rules, configure the GitHub MCP server in Cursor's "Tools & MCP" settings. You will need a read-only personal access token with the following permissions: Pull requests, Issues, Contents, Metadata.
+To use the PR review rules, configure the GitHub MCP server in Cursor's "Tools & MCP" settings.
```
frodre: Maybe this should be in a README under a "Working with agents" section or something?

AGENTS.md
mcgibbon: Done.

AGENTS.md
mcgibbon: It's only specifically relevant to Cursor, and IMO we should rely on Cursor's docs for this as it can change.

mcgibbon:
> One thing I still find more difficult is drift between each PR as sometimes I make stacked PRs (chained branches), but then the rebasing during the human review process can get messy and hard to manage with the follow on branches if there isn't a completely clean separation.

I've actually found this much easier with Claude, because I no longer really need these to be separate branches. What I would normally do as a branch/PR chain when writing my own code, I instead do as a series of git commits when working with Claude. PRs do let you look at code commit-by-commit, I just can't really make those individual commits as well as Claude can, and from this experience I've found separate branches/PRs are in a way a byproduct of the way I write code.

PR #965, opened 2026-03-12
Add `CorrectorConfigABC`

jpdunc23:
Refactors corrector config typing with `CorrectorConfigABC` with a `get_corrector` abstract method which accepts `DatasetInfo` and returns a `CorrectorABC`. This allows the corrector configs to build their corresponding correctors, removing that responsibility from the vertical coordinates.

Changes:
-  `CorrectorSelector.registry` was previously unused but is now once again active via `CorrectorSelector.__post_init__`.
- Adds properties `atmosphere_vertical_coordinate` and `ocean_vertical_coordinate` to `DatasetInfo`. 

- [x] Tests added

### DIFF
--- a/fme/core/dataset_info.py
+++ b/fme/core/dataset_info.py
@@ -187,13 +187,23 @@ class DatasetInfo:
     def atmosphere_vertical_coordinate(self) -> HasAtmosphereVerticalIntegral | None:
         if isinstance(self._vertical_coordinate, HasAtmosphereVerticalIntegral):
             return self._vertical_coordinate
-        return None
+        elif isinstance(self._vertical_coordinate, NullVerticalCoordinate):
+            return None
+        raise RuntimeError(
+            f"{self._vertical_coordinate} cannot be used as an atmosphere vertical "
+            "coordinate."
+        )
 
     @property
     def ocean_vertical_coordinate(self) -> HasOceanDepthIntegral | None:
         if isinstance(self._vertical_coordinate, HasOceanDepthIntegral):
             return self._vertical_coordinate
-        return None
+        elif isinstance(self._vertical_coordinate, NullVerticalCoordinate):
+            return None
+        raise RuntimeError(
+            f"{self._vertical_coordinate} cannot be used as an ocean vertical "
+            "coordinate."
+        )
 
     @property
     def mask_provider(self) -> MaskProvider:
--- a/fme/core/test_dataset_info.py
+++ b/fme/core/test_dataset_info.py
@@ -6,6 +6,7 @@ import pytest
 import torch
 
 from fme.core.coordinates import (
+    DepthCoordinate,
     HybridSigmaPressureCoordinate,
     LatLonCoordinates,
     NullVerticalCoordinate,
@@ -410,3 +411,46 @@ def test_masked_gridded_ops():
         mask_provider=mask_provider,
     )
     assert dataset_info.gridded_operations == expected_gridded_ops
+
+
+def _make_hybrid_sigma_pressure_coordinate():
+    return HybridSigmaPressureCoordinate(ak=torch.arange(10), bk=torch.arange(10))
+
+
+def _make_depth_coordinate():
+    return DepthCoordinate(
+        idepth=torch.arange(4, dtype=torch.float),
+        mask=torch.ones(3),
+    )
+
+
+class TestAtmosphereVerticalCoordinate:
+    def test_returns_coordinate_for_atmosphere(self):
+        vc = _make_hybrid_sigma_pressure_coordinate()
+        info = DatasetInfo(vertical_coordinate=vc)
+        assert info.atmosphere_vertical_coordinate is vc
+
+    def test_returns_none_for_null(self):
+        info = DatasetInfo(vertical_coordinate=NullVerticalCoordinate())
+        assert info.atmosphere_vertical_coordinate is None
+
+    def test_raises_for_ocean_coordinate(self):
+        info = DatasetInfo(vertical_coordinate=_make_depth_coordinate())
+        with pytest.raises(RuntimeError, match="atmosphere vertical coordinate"):
+            info.atmosphere_vertical_coordinate
+
+
+class TestOceanVerticalCoordinate:
+    def test_returns_coordinate_for_ocean(self):
+        vc = _make_depth_coordinate()
+        info = DatasetInfo(vertical_coordinate=vc)
+        assert info.ocean_vertical_coordinate is vc
+
+    def test_returns_none_for_null(self):
+        info = DatasetInfo(vertical_coordinate=NullVerticalCoordinate())
+        assert info.ocean_vertical_coordinate is None
+
+    def test_raises_for_atmosphere_coordinate(self):
+        info = DatasetInfo(vertical_coordinate=_make_hybrid_sigma_pressure_coordinate())
+        with pytest.raises(RuntimeError, match="ocean vertical coordinate"):
+            info.ocean_vertical_coordinate


fme/core/registry/corrector.py
```diff
@@ -1,56 +1,65 @@
...
-# Note we can either type hint register, which prevents registering
-# from overwriting the type with a generic type or Any, _or_ we can
-# type hint the registry attribute below, which prevents `instance`
-# in `build` from being typed as `Any`. We cannot do both, unfortunately.
 T = TypeVar("T")
 
 
 @dataclasses.dataclass
-class CorrectorSelector:
+class CorrectorSelector(CorrectorConfigABC):
```
mcgibbon, line 14: Does CorrectorSelector need to be typed as a CorrectorConfigABC? I understand that it might implement that interface because of symmetry with its registered objects, but if we don't want this to be passed to places where `arg: CorrectorConfigABC` is used (for example, we don't want CorrectorSelector to get registered as a CorrectorConfig), I think we shouldn't inherit here.

fme/core/corrector/atmosphere.py
```diff
@@ -138,6 +138,26 @@ def from_state(cls, state: Mapping[str, Any]) -> "AtmosphereCorrectorConfig":
             data_class=cls, data=state, config=dacite.Config(strict=True)
         )
 
+    def get_corrector(
+        self,
+        gridded_operations: GriddedOperations,
+        vertical_coordinate: Any | None,
```
mcgibbon:
```suggestion
        vertical_coordinate: HasAtmosphereVerticalIntegral | None,
```
If we can do this change (it seems valid), is there any need for the isinstance check below?

fme/core/registry/corrector.py
jpdunc23, line 1: This file could probably be moved to `fme/core/corrector` / combined with `fme/core/corrector/registry.py`.

fme/core/corrector/atmosphere.py
mcgibbon: Hm, I see why this doesn't work. The typing here feels not great, we have to type this as Any because we don't know the type of the vertical coordinate in general.

Suggestion: I think `get_corrector` should receive `DatasetInfo` instead. That's the object responsible for saying "we're trying to initialize for the atmosphere but this dataset isn't of an atmosphere" through exceptions on its property methods, and it makes the typing clearer. It also might reduce this down to 1 input argument.

mcgibbon: Claude had a similar comment:

1. vertical_coordinate: Any | None typing (fme/core/corrector/registry.py:14, and all get_corrector signatures) — The Any type for
  vertical_coordinate defeats type checking at the ABC level. Consider using a VerticalCoordinate | None type (or a Protocol that captures what's
   needed). Each concrete implementation then narrows the type via isinstance, which is fine. The current Any means callers get no type-safety on
   what they pass.

fme/core/corrector/atmosphere.py
```diff
@@ -138,6 +138,26 @@ def from_state(cls, state: Mapping[str, Any]) -> "AtmosphereCorrectorConfig":
...
+        timestep: datetime.timedelta,
+    ) -> "AtmosphereCorrector":
+        if vertical_coordinate and not isinstance(
```
mcgibbon: Issue:

Truthiness check on vertical_coordinate (atmosphere.py:147, ocean.py:126) — if vertical_coordinate and not isinstance(...) uses truthiness
  to check for None. This is subtly fragile — if a vertical coordinate ever has a custom __bool__ or __len__, this could behave unexpectedly.
  Prefer if vertical_coordinate is not None and not isinstance(...).

fme/core/coordinates.py
```diff
@@ -136,14 +132,21 @@ def __repr__(self) -> str:
...
     def build_corrector(
         self,
-        config: AtmosphereCorrectorConfig | CorrectorSelector,
+        config: CorrectorConfigABC,
         gridded_operations: GriddedOperations,
         timestep: timedelta,
     ) -> CorrectorABC:
-        pass
+        vertical_coord: VerticalCoordinate | None = self
+        if isinstance(vertical_coord, NullVerticalCoordinate):
```
mcgibbon: Suggestion (optional):

3. isinstance check in build_corrector (coordinates.py:143) — if isinstance(vertical_coord, NullVerticalCoordinate): vertical_coord = None. Per
   the contributing guidelines (polymorphism over type-checking), this could be handled by giving NullVerticalCoordinate a property that returns
  None for the vertical coordinate value, or by having the ABC define a method like as_optional() that returns self for real coordinates and None
   for null. That said, this is a single check in a @final method, so it's borderline acceptable.

fme/core/registry/corrector.py
```diff
@@ -1,56 +1,65 @@
...
-    # In practice the corrector is selected based on the type of the vertical
-    # coordinate in the loaded data.
     registry: ClassVar[Registry] = Registry()
 
     def __post_init__(self):
-        if not isinstance(self.registry, Registry):
-            raise ValueError("CorrectorSelector.registry should not be set manually")
+        self._corrector_config_instance: CorrectorConfigABC = cast(
+            CorrectorConfigABC, self.registry.get(self.type, self.config)
+        )
```
mcgibbon, line 41: I think there's some weird python typing around Registry that makes this unavoidable, so if you've copied this pattern from somewhere else I think it's fine. But noting this Claude comment:

4. cast in CorrectorSelector.__post_init__ (fme/core/registry/corrector.py:42-44) — The cast to CorrectorConfigABC papers over the fact that
  Registry.get doesn't know about CorrectorConfigABC. If a registered class doesn't subclass CorrectorConfigABC, this would silently pass type
  checking and fail at runtime. Could the register method enforce CorrectorConfigABC subclassing, or could Registry.get be generic-typed?

fme/core/registry/corrector.py
```diff
@@ -1,56 +1,65 @@
...
-        type: the type of the CorrectorConfigProtocol
-        config: data for a CorrectorConfigProtocol instance of the indicated type
+        type: the type of the CorrectorConfigABC
+        config: data for a CorrectorConfigABC instance of the indicated type
+
     """
 
     type: str
     config: Mapping[str, Any]
-    # Note this registry is no longer used, but is kept for backwards compatibility.
```
mcgibbon, line 40: Suggestion: Update PR description about this.

5. Removed comment about registry being unused — The old code had a comment: "Note this registry is no longer used, but is kept for backwards
  compatibility." The registry is now actively used again (via __post_init__ delegation). Good that the stale comment is gone, but it would be
  worth noting in the PR description that this re-activates the registry for corrector construction.

fme/core/corrector/ice.py
```diff
@@ -195,6 +194,18 @@ def from_state(cls, state: Mapping[str, Any]) -> "IceCorrectorConfig":
             data_class=cls, data=state, config=dacite.Config(strict=True)
         )
 
+    def get_corrector(
+        self,
+        gridded_operations: GriddedOperations,
+        vertical_coordinate: Any | None,  # ignored
```
mcgibbon: Passing DatasetInfo would avoid the need for comments like this, Claude had a similar comment:

7. nit: IceCorrectorConfig.get_corrector has vertical_coordinate: Any | None,  # ignored — the "ignored" comment is helpful, but if the ABC
  signature were more specific (point 1), IceCorrector could accept None explicitly and the mismatch would be more apparent.

mcgibbon (commented):
`code` comments are from Claude, the rest are mine or my commentary. I found using Claude Opus on its own (without /code-review) to be much more helpful.

Claude likes your suggestion for moving corrector.py:
  - @jpdunc23 self-commented suggesting fme/core/registry/corrector.py could be moved to or combined with fme/core/corrector/registry.py. This is
   unresolved.

  6. nit: James's existing review comment about moving CorrectorSelector from fme/core/registry/corrector.py to fme/core/corrector/ is worth
  considering — it would co-locate all corrector config/registry code. The current split between fme/core/registry/ and fme/core/corrector/ is
  somewhat confusing.

Suggestion (optional): add get_corrector tests on each config class

  - The PR description has "Tests added" unchecked. Since this is a pure refactor (behavior should be identical), the existing test suite passing
   is reasonable. However, unit tests for get_corrector on each config class (especially the validation error paths in AtmosphereCorrectorConfig
  and OceanCorrectorConfig) would be valuable to prevent regressions.

jpdunc23:
> `code` comments are from Claude, the rest are mine or my commentary. I found using Claude Opus on its own (without /code-review) to be much more helpful.

Sorry, removed the `code` formatting from Claude's comments since I found it unreadable. Still not great due to weird line breaks. 

Might be worth discussing at Monday's tech sync how we can best incorporate agent comments into PR reviews, e.g., a carefully-scoped GitHub PAT to allow commenting via MCP might be helpful.

fme/core/registry/corrector.py
jpdunc23: I think the current answer is "yes" but maybe we can do something to change it to "no". It's "yes" because we currently have a lot of `StepConfigABC` implementations that have `corrector: AtmosphereCorrectorConfig | CorrectorSelector`, so `VerticalCoordinate.build_corrector` has a `config: CorrectorConfigABC` arg which in practice can be either an `AtmosphereCorrectorConfig` or a `CorrectorSelector` object.

If we instead do `corrector: CorrectorSelector` then I think we can just type the `VerticalCoordinate.build_corrector` arg as `config: CorrectorSelector` and then `CorrectorSelector` doesn't need to inherit `CorrectorConfigABC`. I'll see about doing that refactor in this PR.

BTW, this is the same pattern used by `StepSelector`. Now I'm wondering: Why does `StepSelector` need to be typed as `StepConfigABC`?

jpdunc23: The change from `corrector: AtmosphereCorrectorConfig | CorrectorSelector` to `corrector: CorrectorSelector` is probably best done in a follow up PR.

fme/core/registry/corrector.py
jpdunc23: Will do in a follow up PR to avoid bloating the diff here.

fme/core/registry/corrector.py
jpdunc23: Yes, this is a pattern copied from other registry patterns (see `StepSelector`). I think worth looking into Claude's question here for a follow up PR.

fme/core/corrector/atmosphere.py
```diff
@@ -138,6 +139,23 @@ def from_state(cls, state: Mapping[str, Any]) -> "AtmosphereCorrectorConfig":
...
+    def get_corrector(
+        self,
+        dataset_info: DatasetInfo,
+    ) -> "AtmosphereCorrector":
+        vc = dataset_info.vertical_coordinate
+        vertical_coordinate: HasAtmosphereVerticalIntegral | None
+        if isinstance(vc, HasAtmosphereVerticalIntegral):
+            vertical_coordinate = vc
+        else:
+            vertical_coordinate = None
```
jpdunc23: Note that this is a bit more permissive than before since previously there would've been a `ValueError` from the vertical coord while building the corrector if the corrector type didn't match the vertical coord. We could maybe add some sort of `assert_corrector_compatible(corrector_type: str)` method to `DatasetInfo` to retain the previous behavior.

fme/core/corrector/ocean.py
```diff
@@ -117,6 +118,23 @@ def remove_deprecated_keys(cls, state: Mapping[str, Any]) -> dict[str, Any]:
...
+    def get_corrector(
+        self,
+        dataset_info: DatasetInfo,
+    ) -> "OceanCorrector":
+        vc = dataset_info.vertical_coordinate
+        vertical_coordinate: HasOceanDepthIntegral | None
+        if isinstance(vc, HasOceanDepthIntegral):
+            vertical_coordinate = vc
+        else:
+            vertical_coordinate = None
```
jpdunc23: [Same as above](https://github.com/ai2cm/ace/pull/965/changes#r2944365312)

fme/core/corrector/ice.py
jpdunc23: Passing `DatasetInfo` now.

fme/core/coordinates.py
jpdunc23: Removed `VerticalCoordinate.build_corrector` in favor of building the corrector directly from the `CorrectorConfigABC`.

fme/core/corrector/atmosphere.py
jpdunc23: This check removed.

fme/core/corrector/atmosphere.py
jpdunc23: Typing should be better now.

fme/core/corrector/atmosphere.py
```diff
@@ -138,6 +139,23 @@ def from_state(cls, state: Mapping[str, Any]) -> "AtmosphereCorrectorConfig":
             data_class=cls, data=state, config=dacite.Config(strict=True)
         )
 
+    def get_corrector(
+        self,
+        dataset_info: DatasetInfo,
+    ) -> "AtmosphereCorrector":
+        vc = dataset_info.vertical_coordinate
```
mcgibbon: Suggestion: Use a `dataset_info.atmosphere_vertical_coordinate` attribute, adding it if it doesn't already exist, which returns `HasAtmosphereVerticalIntegral | None`. Or, if it matches the patterns on DatasetInfo better, have it raise when it's missing, but catch the exception and turn it to None here.

One of the original motivations for DatasetInfo was to clear up this "vertical coordinate doesn't have any real type information" issue via a property like this for the atmosphere and one for the ocean.

fme/core/corrector/atmosphere.py
mcgibbon: It's difficult because not all atmosphere corrector configurations require the vertical coordinate, and we want to support running no-coordinate runs when only using e.g. the non-negative correctors. So I do think we need the permissive version at this level, raising inside AtmosphereCorrector if we need the VC but it's None.

fme/core/corrector/ocean.py
mcgibbon: Similarly, add a `ocean_vertical_coordinate` attribute to DatasetInfo.

fme/core/registry/corrector.py
mcgibbon: I don't fully understand it right now but Claude seems to think this flow has some merit, and that isinstance checks would be needed to avoid it, so we can keep it. Yes let's still change the corrector type to not include AtmosphereCorrectorConfig in your follow-on PR.

fme/core/registry/corrector.py
mcgibbon: Got Claude to make a PR based from this one, if you can review: https://github.com/ai2cm/ace/pull/984

fme/core/corrector/atmosphere.py
jpdunc23: In the end I made this more restrictive by throwing an error in `atmosphere_vertical_coordinate` and `ocean_vertical_coordinate` when the vertical coordinate type is incompatible with the property (e.g., vertical coordinate is `DepthCoordinate` and user tries to access the `atmosphere_vertical_coordinate` property).

jpdunc23:
Before this PR, the dependency chain was `fme.core.corrector` => `fme.core.coordinates` => `fme.core.dataset_info` (where => means "imported by") and the protocols  `HasOceanDepthIntegral` and `HasAtmosphereVerticalIntegral` helped us to avoid a circular import. 

Changes in this PR removed the dependency `fme.core.corrector` => `fme.core.coordinates`, but added `fme.core.dataset_info` => `fme.core.corrector`, which is a more intuitive dependency.

This implies that we don't actually need `HasOceanDepthIntegral` and `HasAtmosphereVerticalIntegral` anymore, though I think they're useful to keep around for now to communicate the minimal API needed by the correctors.

PR #971, opened 2026-03-13
Add fine coordinates to the model for easier inference handling

frodre:
`DiffusionModel` previously relied on `StaticInput.coords` to store the fine-resolution lat/lon grid, coupling spatial metadata to individual topography fields. This made coordinate handling awkward since models without any static inputs had no coordinate information and would either fail or required a `_downscale_coord` bandaid to approximate fine-resolution coordinates from coarse ones.

### Changes:
- `StaticInputs` now carries a required `coords: LatLonCoordinates` field representing the full fine-resolution domain. `coords` is removed from individual `StaticInput` fields.
- `DiffusionModel` always receives a non-optional `StaticInputs` (fields may be empty when no static data is needed). `full_fine_coords` is a property delegating to `static_inputs.coords`.
- `DiffusionModelConfig.build` and `DiffusionModel.__init__` no longer accept `None` for `static_inputs`.
- `StaticInputs.subset(lat_interval, lon_interval)` replaces `subset_latlon`, computing slices internally and subsetting both fields and coords together.
- `ClosedInterval.slice_of` renamed to `slice_from`; new `subset_of` convenience method returns the coordinate values within the interval.
- All checkpoint backwards-compatibility loading logic is consolidated in `StaticInputs.from_state_backwards_compatible` in `static.py`. `CheckpointModelConfig` gains an optional `fine_coordinates_path` for old checkpoints with no stored coordinates.
- `load_fine_coords_from_path` added to `fme.downscaling.data` and tested.
- `PairedGriddedData` carries `fine_coords`, passed to the model at build time to use as a fallback if no static input's with coordinates in the dataset are specified.
- Removed `_downscale_coord` from `predict.py`; `model.get_fine_coords_for_batch` is used instead.

- [x] Tests added


fme/downscaling/models.py
```diff
@@ -615,10 +632,40 @@ def from_state(
             static_inputs = StaticInputs.from_state(state["static_inputs"]).to_device()
         else:
             static_inputs = None
+
+        # Load fine_coords: new checkpoints store it directly; old checkpoints
+        # that had static_inputs with coords can auto-migrate from raw state.
+        fine_coords = state.get("fine_coords")
+        if fine_coords is not None:
```
frodre: This block is the pathway executed during training resumption and will fail if we try and resume training for a model without any static inputs or fine coordinates.  Not totally sure if this sort of backwards compat is really necessary, since we'll likely just be training new models w/ fine_coords.

AnnaKwa: I'm ok with breaking backwards compatibility here. I think the oldest checkpoint we'd possibly want to continue to use would be the released precip-only model checkpoint, which has static inputs saved.

fme/downscaling/models.py
```diff
@@ -293,8 +297,12 @@ def __init__(
                 coarse to fine.
             sigma_data: The standard deviation of the data, used for diffusion
                 model preconditioning.
+            fine_coords: the full-domain fine-resolution coordinates to use
```
AnnaKwa: Suggestion: call this `full_fine_coords` or something to that effect so it's obvious that this is the full domain and doesn't need to be updated if training is resumed on some different domain.

fme/downscaling/models.py
```diff
@@ -181,6 +182,7 @@ def build(
         self,
         coarse_shape: tuple[int, int],
         downscale_factor: int,
+        fine_coords: LatLonCoordinates,
```
AnnaKwa: For this and other places where this arg is added, see comment about naming to describe it's the full domain.

fme/downscaling/models.py
```diff
@@ -310,6 +318,19 @@ def __init__(
         self.static_inputs = (
             static_inputs.to_device() if static_inputs is not None else None
         )
+        device = get_device()
+        self.fine_coords = LatLonCoordinates(
```
AnnaKwa:
```suggestion
        self.fine_coords = fine_coords.to(device)
```

fme/downscaling/models.py
```diff
@@ -310,6 +318,19 @@ def __init__(
...
+            lat=fine_coords.lat.to(device),
+            lon=fine_coords.lon.to(device),
+        )
+        if static_inputs is not None:
+            expected = (len(fine_coords.lat), len(fine_coords.lon))
```
AnnaKwa:
```suggestion
            expected = fine_coords.shape
```

fme/downscaling/models.py
```diff
@@ -332,36 +353,31 @@ def _subset_static_inputs(
                 "Static inputs must be provided for each batch when use of fine "
                 "static inputs is enabled."
             )
-        return self.static_inputs.subset_latlon(lat_interval, lon_interval)
+        # Fine coords and static inputs are expected to be on the same grid
```
AnnaKwa:
```suggestion
        # Static inputs are expected to be on the same grid
```

fme/downscaling/models.py
```diff
@@ -276,6 +279,7 @@ def __init__(
         coarse_shape: tuple[int, int],
         downscale_factor: int,
         sigma_data: float,
+        fine_coords: LatLonCoordinates,
```
AnnaKwa: Suggestion: make this arg optional, only to be used if building from a checkpoint that has an optional `fine_coords_path` arg to allow for models w/o static inputs to have correct coords in the saved predict/evaluate output. Otherwise for the standard case where the model uses static inputs, set the attribute `self.fine_coords` using the static inputs coords.

fme/core/coordinates.py
```diff
@@ -736,7 +736,7 @@ def __eq__(self, other) -> bool:
     def __repr__(self) -> str:
         return f"LatLonCoordinates(\n    lat={self.lat},\n    lon={self.lon}\n"
 
-    def to(self, device: str) -> "LatLonCoordinates":
+    def to(self, device: str | torch.device) -> "LatLonCoordinates":
```
frodre, line 622: To make my VSCode linter happy

fme/downscaling/data/static.py
```diff
@@ -93,96 +68,179 @@ def _get_normalized_static_input(path: str, field_name: str):
...
+    return LatLonCoordinates(
+        lat=torch.tensor(ds[lat_name].values, dtype=torch.float32),
+        lon=torch.tensor(ds[lon_name].values, dtype=torch.float32),
     )
 
 
 @dataclasses.dataclass
 class StaticInputs:
     fields: list[StaticInput]
+    coords: LatLonCoordinates
```
frodre, line 135: Not named `full_coords` because we do produce subsets with this class.

AnnaKwa:
Ah, I'm very sorry for the confusion- when I said "I do prefer to keep the the coordinates on the StaticInputs" I meant I preferred them kept where there were currently were (within the static inputs but on individual `StaticInput` objects not the higher level `StaticInputs`). The potential mismatch in coords I was concerned about in the previous iteration was between the static inputs coords (post init ensures all `StaticInput` objects have same coords) and the `DiffusionModel`'s coords attribute getting out of sync through mixing of saved checkpoints and updated configs. I think the static inputs coords are best kept in the lowest level object `StaticInput` so it's clear they are associated with the data information there.

> PairedGriddedData carries fine_coords, passed to the model at build time.

Is this so that models without static inputs have the fine coords information? Could we instead set a `_fine_coords_from_gridded_data` attribute and fall back to using this if there are no static inputs (rather than setting the static inputs coords from the fine dataset coords)? See comment on the `full_fine_coords` property. 
It should amount the same thing since the fine coords are usually the same from both sources, but this way it's 100% guaranteed that the static inputs coords are direct from their underlying datasets.

fme/downscaling/models.py
```diff
@@ -307,62 +306,39 @@ def __init__(
...
-            return None
-        if self.static_inputs is None:
-            raise ValueError(
-                "Static inputs must be provided for each batch when use of fine "
-                "static inputs is enabled."
-            )
-        return self.static_inputs.subset_latlon(lat_interval, lon_interval)
+    @property
+    def full_fine_coords(self) -> LatLonCoordinates:
+        return self.static_inputs.coords
```
AnnaKwa: Could we do something like have an attribute `self._full_fine_coords_from_gridded_data: None | LatLonCoordinates` that is set in the build method, and then

```
if len(self.static_inputs.fields) > 0: # or maybe add a len method to the class
    return self.static_inputs.coords
else:
    return self._full_fine_coords_from_gridded_data
```

AnnaKwa: Can the fallback be put here, rather than within the `StaticInputs`?

fme/downscaling/data/static.py
```diff
@@ -232,15 +247,42 @@ def from_state_backwards_compatible(
...
+            coords_to_use = coords
+        elif coords is not None and validate_coords:
+            assert coords_to_use is not None  # for type checker
+            _validate_coords(field_name, coords, coords_to_use)
+
+    if coords_to_use is None:
+        # no coords found with static inputs, use provided fallback
+        coords_to_use = fallback_coords
+    elif validate_coords:
+        _validate_coords("fallback", coords_to_use, fallback_coords)
```
AnnaKwa: Could this validation also get moved up to be done at the model level when it gets built?

fme/downscaling/data/static.py
```diff
@@ -58,22 +85,30 @@ def _get_normalized_static_input(path: str, field_name: str):
...
-    if "time" in static_input.dims:
-        static_input = static_input.isel(time=0).squeeze()
-    if len(static_input.shape) != 2:
+        ds = xr.open_dataset(path, mask_and_scale=False)
+
+    da = ds[field_name]
+    try:
+        coords = _load_coords_from_ds(ds)
+    except ValueError:
+        # no coords available
```
AnnaKwa: Up to you, but in the data loading (get_horizontal_coordinates) it's assumed the last two dims are lat, lon so I think it's ok to assume this here as well. If for some reason there are no usable coords I think we should just raise the error here (i.e. if there are static inputs, we expect them to have valid coords).

frodre:
Split into #997, #998, and #999.  Thanks @AnnaKwa for your feedback on this update!

PR #974, opened 2026-03-15
Fix dataset_info to use global (non-localized) properties

### LINKED ISSUE #972: bug in data_loading/gridded_data: dataset_info shouldn't be localized
introduced in https://github.com/ai2cm/ace/pull/955

GriddedData.__init__ calls properties.to_device().localize(), dataset_info is built from these localized properties, so dataset_info.img_shape holds the local shape. But the SFNO model receives this local shape and slices/localizes it again via get_local_slices(local_shape), thus double partitioning

The fix: store self._global_properties = properties.to_device() before calling .localize(). Then dataset_info is constructed from self._global_properties

applies to both GriddedData and InferenceData

should add test either in test_dataset_info (maybe annoying) or test_data_loader (maybe easer)

mahf708:
This pull request ensures that the `dataset_info` property in both `GriddedData` and `InferenceGriddedData` uses global (non-localized) dataset properties, which prevents models from accidentally double-partitioning data under spatial parallelism. It also adds tests to verify this behavior.

- [x] Tests added

Resolves #972

### DIFF
--- a/fme/ace/data_loading/gridded_data.py
+++ b/fme/ace/data_loading/gridded_data.py
@@ -202,7 +202,7 @@ class InferenceGriddedData(InferenceDataABC[PrognosticState, BatchData]):
 
     @property
     def dataset_info(self) -> DatasetInfo:
-    """Always returns global datasets regardless of model parallelism"""
+        """Always returns global datasets regardless of model parallelism."""
         return DatasetInfo(
             horizontal_coordinates=self._global_properties.horizontal_coordinates,
             vertical_coordinate=self._global_properties.vertical_coordinate,


fme/ace/data_loading/gridded_data.py
```diff
@@ -202,11 +204,11 @@ def variable_metadata(self) -> dict[str, VariableMetadata]:
     @property
     def dataset_info(self) -> DatasetInfo:
```
elynnwu, line 204: [nit] add a doc string in `DatasetInfo` to say that it always returns the global dataset info regardless of domain/model parallelism

fme/ace/data_loading/gridded_data.py
```diff
@@ -50,9 +50,10 @@ def __init__(
...
         self._loader = loader
-        shape = properties.horizontal_coordinates.shape
+        self._global_properties = properties.to_device()
+        shape = self._global_properties.horizontal_coordinates.shape
         self._global_img_shape: tuple[int, int] = (shape[-2], shape[-1])
-        self._properties = properties.to_device().localize()
+        self._properties = self._global_properties.localize()
         self._timestep = self._properties.timestep
         self._vertical_coordinate = self._properties.vertical_coordinate
         self._mask_provider = self._properties.mask_provider
```
elynnwu: `self._mask_provider` is not used, can remove now

mahf708:
```suggestion
```

fme/ace/data_loading/gridded_data.py
mahf708:
```suggestion
    def dataset_info(self) -> DatasetInfo:
    """Always returns global datasets regardless of model parallelism"""
```

elynnwu (approved):

PR #975, opened 2026-03-15
add model-parallel metrics with tests

mahf708:
Implement both zonal_mean and gradient_magnitude_percent_diff for ModelTorchDistributed backend. The gradient_magnitude_percent_diff impl is quite inefficient, and tha'ts okay because it's not really on a critical path (and maybe we should just diable it?)

Changes:
- a minor API change in the base backend class exposing a gather_spatial_tensor, and then building on that with gather_spatial in distributed. Not 100% sure this is better or worse, it makes life slightly easier for downstream use (see comment below)

- [x] Tests added (in the "right" place)

### DIFF
--- a/scripts/era5/Makefile
+++ b/scripts/era5/Makefile
@@ -2,6 +2,20 @@ VERSION ?= 2026-03-03-era5-xarray-beam-pipelines
 IMAGE_NAME ?= us-central1-docker.pkg.dev/vcm-ml/full-model/era5-ingest-dataflow:$(VERSION)
 LOCAL_ENVIRONMENT ?= era5-ingestion
 
+# Production run parameters
+START_TIME=1940-01-01T12:00:00
+END_TIME=2025-12-31T18:00:00
+OUTPUT_PATH_ONE_DEGREE=gs://vcm-ml-intermediate/2026-03-19-era5-1deg-8layer-1940-2025.zarr
+OUTPUT_GRID_ONE_DEGREE=F90
+OUTPUT_PATH_FOUR_DEGREE=gs://vcm-ml-intermediate/2026-03-19-era5-4deg-8layer-1940-2025.zarr
+OUTPUT_GRID_FOUR_DEGREE=F22.5
+
+# Test run parameters
+START_TIME_TEST_RUN ?= 2025-12-30T00:00:00
+END_TIME_TEST_RUN ?= 2025-12-31T18:00:00
+OUTPUT_PATH_TEST_RUN ?= gs://vcm-ml-scratch/spencerc/test-updated-era5-pipeline/era5-4deg-8layer.zarr
+OUTPUT_GRID_TEST_RUN ?= F22.5
+
 create_environment:
 	conda env create -f environment.yaml
 	conda run --no-capture-output -n era5-ingestion pip install -r dataflow-requirements.txt
@@ -19,5 +33,41 @@ enter:
 enter_google:
 	docker run --rm -v $$(pwd):/era5 -w /era5 --entrypoint "/bin/bash" -it gcr.io/weather-tools-prod/weather-tools:0.0.0
 
-era5_dataflow:
-	cd pipeline && conda run --no-capture-output -n $(LOCAL_ENVIRONMENT) ./run-dataflow.sh
+.PHONY: era5_dataflow
+era5_dataflow: era5_dataflow_one_degree era5_dataflow_four_degree
+
+.PHONY: era5_dataflow_one_degree
+era5_dataflow_one_degree:
+	cd pipeline && \
+		conda run --no-capture-output -n $(LOCAL_ENVIRONMENT) \
+		./run-dataflow.sh \
+		DataflowRunner \
+		$(OUTPUT_PATH_ONE_DEGREE) \
+		$(OUTPUT_GRID_ONE_DEGREE) \
+		$(START_TIME) \
+		$(END_TIME) \
+		--check_data_validity
+
+.PHONY: era5_dataflow_four_degree
+era5_dataflow_four_degree:
+	cd pipeline && \
+		conda run --no-capture-output -n $(LOCAL_ENVIRONMENT) \
+		./run-dataflow.sh \
+		DataflowRunner \
+		$(OUTPUT_PATH_FOUR_DEGREE) \
+		$(OUTPUT_GRID_FOUR_DEGREE) \
+		$(START_TIME) \
+		$(END_TIME) \
+		--check_data_validity
+
+.PHONY: era5_dataflow_test_run
+era5_dataflow_test_run:
+	cd pipeline && \
+		conda run --no-capture-output -n $(LOCAL_ENVIRONMENT) \
+		./run-dataflow.sh \
+		DataflowRunner \
+		$(OUTPUT_PATH_TEST_RUN) \
+		$(OUTPUT_GRID_TEST_RUN) \
+		$(START_TIME_TEST_RUN) \
+		$(END_TIME_TEST_RUN) \
+		--check_data_validity
--- a/scripts/era5/README.md
+++ b/scripts/era5/README.md
@@ -25,3 +25,14 @@ image has been pushed, the pipeline can be launched with:
 ```
 make era5_dataflow
 ```
+
+This will run the production workflow on the full range of data targeting one
+degree and four degree grids. During development it can be helpful to run the
+workflow for a shorter period of time targeting just a single grid. For this
+purpose use the
+
+```
+make era5_dataflow_test_run
+```
+
+rule as a starting point.
--- a/scripts/era5/pipeline/run-dataflow.sh
+++ b/scripts/era5/pipeline/run-dataflow.sh
@@ -3,14 +3,19 @@
 # Options,
 # DirectRunner - run pipeline locally, good for testing and debugging
 # DataflowRunner - run pipeline on Google Cloud Dataflow, good for large scale processing
-RUNNER=${1:-DataflowRunner}
+RUNNER="${1}"
+OUTPUT_PATH="${2}"
+OUTPUT_GRID="${3}"
+START_TIME="${4}"
+END_TIME="${5}"
+EXTRA_FLAGS=("${@:6}")
 
 # Common args shared by all runners
 COMMON_ARGS=(
-    gs://vcm-ml-scratch/oliwm/test-updated-era5-pipeline/era5-1deg-8layer-2013-2022.zarr
-    2013-01-01T00:00:00
-    2022-12-31T18:00:00
-    --output_grid F90
+    $OUTPUT_PATH
+    $START_TIME
+    $END_TIME
+    --output_grid $OUTPUT_GRID
     --output_time_chunksize 1
     --output_time_shardsize 240
     --process_time_chunksize 2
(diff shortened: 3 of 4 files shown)


fme/core/distributed/distributed.py
```diff
@@ -469,14 +469,10 @@ def gather_spatial(
...
-        slices = self.get_local_slices(img_shape)
-        result = {}
-        for k, v in data.items():
-            global_shape = (*v.shape[:-2], *img_shape)
-            global_tensor = torch.zeros(global_shape, dtype=v.dtype, device=v.device)
-            global_tensor[(..., *slices)] = v
-            result[k] = self.spatial_reduce_sum(global_tensor)
-        return result
+        return {
+            k: self._distributed.gather_spatial_tensor(v, img_shape)
```
mahf708: this now gets something from base.py, which felt a lil odd because that base.py was bare. I think the new is better, but I am not 100% sure. It helps in later gather ops in the backends...

fme/core/distributed/model_torch_distributed.py
```diff
@@ -361,9 +366,18 @@ def gradient_magnitude_percent_diff(
...
+        h_total = torch.tensor(truth.shape[-2], device=truth.device)
+        w_total = torch.tensor(truth.shape[-1], device=truth.device)
+        torch.distributed.all_reduce(h_total, group=self._h_group)
+        torch.distributed.all_reduce(w_total, group=self._w_group)
+        img_shape = (int(h_total), int(w_total))
+
+        return metrics.gradient_magnitude_percent_diff(
+            self.gather_spatial_tensor(truth, img_shape),
+            self.gather_spatial_tensor(predicted, img_shape),
+            weights=self.gather_spatial_tensor(weights, img_shape),
```
mahf708: this is a "correct" impl, but not efficient. we have some paths forward for a distributed future:
1. a fully correct impl with halo exchanges (will need to code up that stuff carefully, etc.)
2. something fully correct, but in between (e.g., avoid excessive compute on all ranks, avoid excessive alloc on all ranks, etc.)
3. a nearly correct impl that simply ignores boundaries, see below for a sketch

```py
  def gradient_magnitude_percent_diff(                                                                       
        self,                                                                                                
        truth: torch.Tensor,                                                                                 
        predicted: torch.Tensor,                                                                             
        weights: torch.Tensor,                                                                               
        dim: tuple[int, ...],                                                                                
    ) -> torch.Tensor:                                                                                       
        from fme.core.metrics import gradient_magnitude                                                      
                                                                                                             
        truth_grad_mag = gradient_magnitude(truth, dim)                                                      
        predicted_grad_mag = gradient_magnitude(predicted, dim)                                              
        truth_mean = self._weighted_nanmean(truth_grad_mag, weights, dim)                                    
        predicted_mean = self._weighted_nanmean(predicted_grad_mag, weights, dim)                            
        return 100 * (predicted_mean - truth_mean) / truth_mean                                              
                                                                                                             
    def _weighted_nanmean(                                                                                   
        self,                                                                                                
        data: torch.Tensor,                                                                                  
        weights: torch.Tensor,                                                                               
        dim: tuple[int, ...],                                                                                
    ) -> torch.Tensor:                                                                                       
        expanded_weights = weights.expand(data.shape)                                                        
        valid_weights = torch.where(torch.isnan(data), 0.0, expanded_weights)                                
        local_weighted_sum = torch.nan_to_num(data * expanded_weights).sum(dim=dim)                          
        local_weight_sum = valid_weights.sum(dim=dim)                                                        
        return self.spatial_reduce_sum(local_weighted_sum) / self.spatial_reduce_sum(                        
            local_weight_sum                                                                                 
        ) 
```


let's open an issue or ... shall we tuck this under the rug 👀 I'd go for the last option and forget about it if this were up to me

mahf708:
@elynnwu the one fail is a timer fail --- is it something i should look into?

fme/core/distributed/model_torch_distributed.py
elynnwu: Can you open an issue for this? I think it's fine for this to go in now but we should address it later.

fme/core/distributed/base.py
```diff
@@ -153,6 +153,23 @@ def spatial_reduce_sum(self, tensor: torch.Tensor) -> torch.Tensor:
         """All-reduce sum across spatial (h, w) ranks. Identity for non-spatial."""
         ...
 
+    def gather_spatial_tensor(
```
elynnwu: I see what you mean by base not being purely abstract. The logic here is technically shared across all backends, does that make it a good exception? I'm deferring this to @mcgibbon, specifically if this is a concern if we start to add more concrete methods in the future?

mahf708: Happy to wait on this until later; in the meanwhile, I put this PR to skip the main metric needing special treatment (it's not strictly needed). See: https://github.com/ai2cm/ace/pull/983

mahf708:
Setting this to draft, and we can live with https://github.com/ai2cm/ace/pull/983 in the meanwhile.

fme/core/distributed/base.py
mcgibbon: Suggestion: Move this to `Distributed` instead of on the backend ABC.

Question: Why is `gather_global` not sufficient for this purpose? I'd like to avoid two ways to do the same thing, if possible, and it seems like they do the same thing except without the data-parallel portion of the gather. If we really need to be able to do a spatial-only gather, we could update `gather_global` to allow None for the data-parallel dim, and do spatial-only in this case.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -361,9 +366,18 @@ def gradient_magnitude_percent_diff(
...
     ) -> torch.Tensor:
-        raise NotImplementedError(
-            "gradient_magnitude_percent_diff is not yet supported "
-            "under spatial parallelism."
+        # Recover global spatial shape (2 scalar all-reduces).
+        h_total = torch.tensor(truth.shape[-2], device=truth.device)
+        w_total = torch.tensor(truth.shape[-1], device=truth.device)
+        torch.distributed.all_reduce(h_total, group=self._h_group)
+        torch.distributed.all_reduce(w_total, group=self._w_group)
+        img_shape = (int(h_total), int(w_total))
```
mcgibbon: Probably this should be a helper function, or more likely this code should go into `gather_spatial_tensor` as a way of supporting `img_shape=None` (you could consider still _requiring_ the argument, so the user has to explicitly say "yeah I don't have the img_shape in this scope", given how important it is to pass along.

We probably could also speed this up by passing the img_shape through here, we have access to it where we call this function (at least, it gets passed in to the aggregator init's in general).

fme/core/distributed/model_torch_distributed.py
mcgibbon: Yes let's open an issue for this and address it later.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -361,9 +366,18 @@ def gradient_magnitude_percent_diff(
...
+
+        return metrics.gradient_magnitude_percent_diff(
```
mcgibbon: Not now: Later I'd like to flip this so that `metrics.gradient_magnitude_percent_diff` calls helper methods on `dist`, instead of vice-versa. E.g. the efficient version of this probably involves `dist.weighted_mean_diff_h` and `dist.weighted_mean_diff_w` functions called inside `gradient_magnitude_percent_diff`.

For example, you could do it this way:
- pass your left data to/receive your right data from the rank next to you along the h group
- compute weighted_mean_diff_h locally on each rank for its internal data, including the right side halo
- gather these diffs to the global root

The mean_diff_w is identical except with a different dimension and comm group, so they can both use the same helper function.

zonal_mean feels like an atomic distributed function that long-term should exist on Distributed (though maybe later in a more general form), but grad mag percent diff doesn't.

fme/core/distributed/base.py
mahf708: > Question: Why is gather_global not sufficient for this purpose?  I'd like to avoid two ways to do the same thing, if possible [...]

The only difference is the signature/return ... gather_spatial returns a stack of tensors, but gather_spatial_tensor returns a single tensor (from that stack). That's why I made gather_spatial use gather_spatial_tensor ... but maybe there's an easier/cleaner way without editing all files using gather_spatial (which are only tests anyway)?

mahf708:
```py
    def gather_spatial(
        self, data: dict[str, torch.Tensor], img_shape: tuple[int, int]
    ) -> dict[str, torch.Tensor]:
        return {
            k: self._distributed.gather_spatial_tensor(v, img_shape)
            for k, v in data.items()
        }
```

mahf708: Sorry I misread your question 💀 

the difference between gather_global and this: gather_global does stuff on the root only whereas this does all ranks, right? I did consider this earlier and settled on the hierarchical approach in gather_spatial borrowing from gather_spatial_tensor

fme/core/distributed/distributed.py
```diff
@@ -453,9 +455,13 @@ def gradient_magnitude_percent_diff(
...
         weights: torch.Tensor,
         dim: tuple[int, ...],
+        img_shape: tuple[int, int],
     ) -> torch.Tensor:
-        return self._distributed.gradient_magnitude_percent_diff(
-            truth, predicted, weights, dim
+        truth = self.gather_spatial_tensor(truth, img_shape)
+        predicted = self.gather_spatial_tensor(predicted, img_shape)
+        weights = self.gather_spatial_tensor(weights, img_shape)
+        return metrics.gradient_magnitude_percent_diff(
```
mcgibbon, line 463: Very simple!

Do we expect this to return a globally reduced value or just a spatially-reduced value? Currently, this returned value needs to afterwards also get passed into a `dist.reduce_mean` call, which I think is fine/expected/congruent with all our other metrics, but I wanted to check.

fme/core/testing/distributed.py
```diff
@@ -56,12 +56,20 @@ def weighted_mean(self, data, weights, dim, keepdim=False) -> torch.Tensor:
...
     def gradient_magnitude_percent_diff(
-        self, truth, predicted, weights, dim
+        self, truth, predicted, weights, dim, img_shape
     ) -> torch.Tensor:
-        return metrics.gradient_magnitude_percent_diff(
+        result = metrics.gradient_magnitude_percent_diff(
             truth, predicted, weights=weights, dim=dim
         )
+        result.fill_(self.fill_value)
+        self.reduce_called = True
```
mcgibbon: This line got added, but I don't see any other references to `.reduce_called`. Is this used? If these changes aren't used, please remove them.

fme/core/distributed/distributed.py
mahf708: I think we want the global value, right?

fme/core/testing/distributed.py
mahf708:
```suggestion
```

fme/core/distributed/distributed.py
mcgibbon: Well, here you're returning the spatially-global but not rank-global value. Assuming the gather is allgather, each rank gets the grad mag percent diff for its data-parallel segment, and then later when we call dist.reduce_mean those get reduced to a truly global value for grad mag percent diff.

PR #979, opened 2026-03-16
Don't filter derived variables in coupled `prediction_loader` eval

### LINKED ISSUE #980: Missing derived variables in coupled `prediction_loader` eval
Derived variables such as `ocean_heat_content` are filtered out by the logic that ensures no overlapping forcing names in ocean vs atmosphere metrics.

jpdunc23:
Fixes #980 by delaying the filtering of overlapping ocean / atmosphere forcings until the inference loop. That way, we can determine the overlapping forcings that need to be removed from the ocean data and remove those variables specifically, rather than remove all variables not found in `CoupledStepper.all_names`.

Changes:
- The `CoupledStepperConfig` property `all_names` no longer filters the ocean names for overlap with atmosphere names.
- Fixed an unrelated bug in `fme/coupled/inference/test_evaluator.py` where the saved `CoupledStepper` checkpoint could have a different ocean mask than the mock inference data.

- [x] Tests added

Resolves #980


fme/coupled/inference/loop.py
```diff
@@ -39,21 +39,25 @@ def run_coupled_dataset_comparison(
...
 ) -> None:
     if record_logs is None:
         record_logs = get_record_to_wandb(label="inference")
     if writer is None:
         writer = NullDataWriter()
 
-    def _restrict_to_all_names(batch: CoupledBatchData) -> CoupledBatchData:
-        if restrict_to_all_names is None:
+    if all_names is not None:
+        overlapping_names = set(all_names.ocean) & set(all_names.atmosphere)
```
elynnwu, line 50: Okay I understand now, this is pretty subtle. So the issue is that all names do not include derived names and we were subsetting it based on the out names which remove all the derived names.

PR #981, opened 2026-03-16
Add per variable loss to Stepper and Log using TrainAggs

### LINKED ISSUE #485: Add metrics to track variable-specific contributions to the training loss
https://github.com/ai2cm/full-model/pull/1239 proposed adding these "per-variable" metrics via `TrainAggregator` but we decided that these would be best to include in validation rather than training.

Arcomano1234:
Currently, we do not log the individual components of the loss (e.g., variable contributions to the overall loss), although we do log weighted RMSE for training, val, and inference (which is very rarely the actual loss). This can make diagnosing overfitting difficult. This PR uses the new created a per_channel (per variable) LossOutput method in `fme/core/loss.py` to add a `per_channel_losses` to `TrainOutput`. This gets passed to `TrainAggregator` where it gets logged by default. 

Example of run on [wandb](https://wandb.ai/ai2cm/ace/runs/sb5izqb1?nw=as78mf7ewuo).

Changes:
- Add `per_channel_losses` to `TrainOutput`
- Add the necessary logging logic in `TrainAggregator` to log per-channel loss 

- [x] Tests added

Resolves #485


fme/ace/aggregator/train.py
```diff
@@ -66,6 +67,16 @@ def __init__(self, config: TrainAggregatorConfig, operations: GriddedOperations)
     def record_batch(self, batch: TrainOutput):
         self._loss += batch.metrics["loss"]
         self._n_loss_batches += 1
+        for key, value in batch.metrics.items():
+            if key.startswith("loss/") and key != "loss":
```
Arcomano1234: Not a fan of this

fme/ace/aggregator/train.py
```diff
@@ -13,6 +13,9 @@
 from fme.core.tensors import fold_ensemble_dim, fold_sized_ensemble_dim
 from fme.core.typing_ import TensorMapping
 
+# Metric key prefix for per-variable loss (must match stepper's metrics["loss/<var>"]).
+PER_CHANNEL_LOSS_PREFIX = "loss/"
```
oliverwm1: not ideal to have this coupling with the naming in the stepper metrics, but this already exists for the the other loss terms so I think it's okay.

fme/ace/aggregator/train.py
```diff
@@ -22,10 +25,13 @@ class TrainAggregatorConfig:
     Attributes:
         spherical_power_spectrum: Whether to compute the spherical power spectrum.
         weighted_rmse: Whether to compute the weighted RMSE.
+        per_channel_loss: Whether to accumulate and report per-variable (per-channel)
+            loss in get_logs (e.g. train/mean/loss/<var_name>).
     """
 
     spherical_power_spectrum: bool = True
     weighted_rmse: bool = True
+    per_channel_loss: bool = False
```
oliverwm1: I think this metric is more important than the other two above—so a bit backwards to have it off by default while the others are on. To avoid metric explosion, we could set the 'weighted_rmse' to False by default.

fme/ace/stepper/single_module.py
```diff
@@ -1628,8 +1629,23 @@ def _accumulate_loss(
...
+                else:
+                    for k in per_channel_sum:
+                        per_channel_sum[k] = (
+                            per_channel_sum[k] + per_channel[k].detach()
+                        )
             if optimize_step:
                 optimization.accumulate_loss(step_loss)
+        if per_channel_sum is not None and n_forward_steps > 0:
+            for k, v in per_channel_sum.items():
+                metrics[f"loss/{k}"] = (v / n_forward_steps).detach()
```
oliverwm1: why the divide by `n_forward_steps` here? I'm not following why this code is necessary.

fme/ace/stepper/single_module.py
```diff
@@ -1628,8 +1629,23 @@ def _accumulate_loss(
                 )
                 step_loss = self._loss_obj(gen_step, target_step, step=step)
                 metrics[f"loss_step_{step}"] = step_loss.detach()
+                per_channel = self._loss_obj.forward_per_channel(
```
oliverwm1: Is there a way to avoid computing this when it is not needed? I.e. it seems we only should do this calculation when computing metrics for the train/val aggregator, but not when actually training.

fme/ace/aggregator/train.py
Arcomano1234: Yeah I was not a fan of this at all either but Claude and I couldn't think of a good way. I guess the one thing I can do is make this an aggregator it self and decouple anything from the stepper. This would also help reduce the need to record it when we aren't using it during training.

Arcomano1234: The problem becomes then getting the loss function to the aggregator which has it's own complications. I defer to you or Jeremy on whether its worth decoupling this from the stepper and just pass a loss_fn to an "PerChannelLossAggregator".

fme/ace/stepper/single_module.py
Arcomano1234: Yeah for some reason claude is convinced we are doing the loss normalization wrong for multi-step losses and wants to normalize by the `n_forward_steps` for training but I fixed that.

fme/ace/stepper/single_module.py
Arcomano1234: Agreed its a little hacky but the newest commit has a way around this.

fme/core/generics/trainer.py
```diff
@@ -490,12 +490,17 @@ def train_one_epoch(self):
...
         with torch.no_grad(), self.validation_context():
             for batch in self.train_data.subset_loader(
                 stop_batch=self.config.train_evaluation_batches
             ):
                 with GlobalTimer():
-                    stepped = self.stepper.train_on_batch(batch, self._no_optimization)
+                    stepped = self.stepper.train_on_batch(
+                        batch,
+                        self._no_optimization,
+                        compute_per_channel_metrics=compute_per_channel,
```
oliverwm1: I think it would be fine to just hard-code this to True instead of coupling to the aggregator. Important thing is that it's only done for the `train_evaluation_batches` not the full training dataset.

fme/ace/stepper/single_module.py
```diff
@@ -1573,6 +1578,8 @@ def _accumulate_loss(
         target_data: BatchData,
         optimization: OptimizationABC,
         metrics: dict[str, float],
+        *,
```
oliverwm1: why *?

fme/ace/aggregator/train.py
oliverwm1: Yeah, like I said it's a pre-existing issue so I don't think we should worry about decoupling in this PR. But open to other thoughts from @mcgibbon on this.

fme/ace/stepper/single_module.py
Arcomano1234: Claude has been trying to add this back in every time I've made changes happen, for some reason it likes this style.

fme/core/generics/trainer.py
Arcomano1234: Yeah I agree should simply the code.

Arcomano1234: Updated

configs/baselines/era5/ace-train-config.yaml
```diff
@@ -2,8 +2,10 @@
 experiment_dir: /results
 save_checkpoint: true
 validate_using_ema: true
-max_epochs: 120
+max_epochs: 2
```
mcgibbon: Revert this before merge

configs/baselines/era5/run-ace-train.sh
```diff
@@ -52,4 +52,4 @@ run_training() {
     -- torchrun --nproc_per_node $N_GPUS -m fme.ace.train $CONFIG_PATH
 }
 
-run_training "ace-train-config.yaml" "ace2-era5-train" "ace2-era5"
+run_training "ace-train-config.yaml" "ace2-era5-train-test-per-var-loss" "ace2-era5"
```
mcgibbon: Revert this before merge

fme/ace/stepper/single_module.py
```diff
@@ -1628,8 +1636,24 @@ def _accumulate_loss(
...
+                    )
+                    if per_channel_sum is None:
+                        per_channel_sum = {
+                            k: v.detach().clone() for k, v in per_channel.items()
+                        }
+                    else:
+                        for k in per_channel_sum:
+                            per_channel_sum[k] = (
+                                per_channel_sum[k] + per_channel[k].detach()
+                            )
```
mcgibbon: Rather than re-computing the per-channel metrics, I suggest updating the code so `_loss_obj` returns a 1-D vector over the channel dimension. Then at this level you can use the unpacker to turn that into a dict of per-channel losses. while still using `step_loss.sum()` as the optimized/accumulated loss. This should make it cheap enough that you can just always "compute" these per-channel metrics, instead of needing new boolean flags.

Additionally, right now you have low-level accumulation code in the middle of the optimize function. I suggest instead using a simple `Aggregator` implementation that takes in a dict of values and keeps a running mean of those values (maybe the existing batch metrics aggregator already handles this) to accumulate these channel losses. You could either attach them as an additional attribute on TrainOutput that then gets passed to this aggregator above this scope (probably better? at least more consistent with what we currently do), or you could take the aggregator as an Optional input argument and record them in this scope.

fme/core/loss.py
```diff
@@ -73,6 +73,50 @@ def __call__(
 
         return self.loss(predict_tensors, target_tensors)
 
+    def call_per_channel(
```
mcgibbon: I think this function can be avoided if loss_obj were to return a 1D vector of per-channel losses.

fme/core/loss.py
```diff
@@ -470,6 +514,22 @@ def forward(
         step_weight = (1.0 + self.sqrt_loss_decay_constant * step) ** (-0.5)
         return self.loss(predict_dict, target_dict) * step_weight
 
+    def forward_per_channel(
```
mcgibbon: What's the meaning of "forward" here? In any case, I think this function can be avoided if loss_obj were to return a 1D vector of per-channel losses.

fme/ace/aggregator/train.py
mcgibbon: I'd suggest using a new attribute on the TrainOutput instead of a string label.

fme/ace/stepper/single_module.py
mcgibbon: This forces kwargs to be passed as kwargs and not as positional arguments, which IMO is nice though not something we enforce.

fme/ace/stepper/single_module.py
mcgibbon: Make sure that when refactoring this, you maintain the sum-ness of loss across steps (perhaps make sure this is tested).

mcgibbon (commented):
I think we can iterate on this design to reduce the amount of code needed to handle this concern, and avoid re-computing the loss N times (with N GPU kernel dispatches). I'd suggest making a substantial change to the API of the loss object, having them return 1D vectors across channel, to avoid most of the code here as well as to avoid recomputing the loss (which can be expensive, I expect the GPUs to have very low occupancy during these calls). If we need to allow support for cross-channel losses that don't nicely fit into per-channel losses, we can iterate later by having loss return a data type allowing both per-channel and scalar losses (each optional).

Also, it feels a little odd to have TrainAggregator responsible so much for these logs. At the least, the validation aggregator should also log them, and it should be easy in principle (even if we don't want to run it that way as a matter of course) to include these in the per-batch metrics we get during training. The suggested change(s) would make this easier to implement if we choose, as it avoids the low-level prefix arithmetic in the aggregator.

fme/ace/stepper/single_module.py
Arcomano1234: Yeah at least for me claude really wants us to use this style. There is some benefit IMO to this way of passing kwargs but thats outside the scope of this PR

fme/ace/stepper/single_module.py
```diff
@@ -346,6 +346,7 @@ class TrainOutput(TrainOutputABC):
     derive_func: Callable[[TensorMapping, TensorMapping], TensorDict] = lambda x, _: (
         dict(x)
     )
+    per_channel_losses: dict[str, torch.Tensor] | None = None
```
Arcomano1234, line 348: The idea is in future PRs for this to be `LossOutput` and allow the aggs to have access to this. Just felt to much for this PR

PR #984, opened 2026-03-18
Parameterize Registry with config ABCs enforce type safety

mcgibbon:
Parameterize `Registry` with config ABCs so that `Registry.get()` returns the correct type directly, eliminating `cast()` calls and enabling the existing `issubclass` check in `Registry.register` to enforce that registered classes are subtypes of the config ABC.

Changes:
- `CorrectorConfigABC`: add default `from_state` (to satisfy `HasFromState` bound on `Registry`'s TypeVar)
- `StepConfigABC.from_state`: widen parameter type from `dict` to `Mapping` (to satisfy `HasFromState`)
- `CorrectorSelector`, `StepSelector`: parameterize `Registry`, remove `cast`, clean up unused imports

- [ ] Tests added

### DIFF
--- a/fme/ace/registry/stochastic_sfno.py
+++ b/fme/ace/registry/stochastic_sfno.py
@@ -11,6 +11,7 @@ from fme.core.distributed.distributed import Distributed
 from fme.core.models.conditional_sfno.sfnonet import (
     Context,
     ContextConfig,
+    SFNONetConfig,
     get_lat_lon_sfnonet,
 )
 from fme.core.models.conditional_sfno.sfnonet import (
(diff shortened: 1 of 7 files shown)


fme/core/corrector/registry.py
```diff
@@ -1,10 +1,18 @@
...
+import dacite
 
 from fme.core.dataset_info import DatasetInfo
 from fme.core.typing_ import TensorDict, TensorMapping
 
 
 class CorrectorConfigABC(abc.ABC):
+    @classmethod
+    def from_state(cls, state: Mapping[str, Any]) -> Self:
+        return dacite.from_dict(cls, state, config=dacite.Config(strict=True))
```
jpdunc23, line 16: `OceanCorrectorConfig` overrides this with a version that calls its `remove_deprecated_keys` on the state dict. This has been helpful to allow refactoring of the ocean corrector config attributes. Just noting since we have mostly tried to avoid overrides, so I would support having this be an abstract classmethod. Fine with leaving as is if you prefer.

fme/core/corrector/registry.py
```diff
@@ -1,10 +1,29 @@
...
 
 from fme.core.dataset_info import DatasetInfo
 from fme.core.typing_ import TensorDict, TensorMapping
 
 
 class CorrectorConfigABC(abc.ABC):
+    @classmethod
+    @final
+    def from_state(cls, state: Mapping[str, Any]) -> Self:
+        cls.remove_deprecated_keys(state)
```
jpdunc23:
```suggestion
        state = cls.remove_deprecated_keys(state)
```

PR #986, opened 2026-03-18
Add warning when requested output variable not present 

### LINKED ISSUE #985: No error if requested name output does not exist
Had an evaluation run where I accidentally requested the wrong output variable name but did not realize until after the run completed. Would be better if we add a warning or raise an error.

elynnwu:
Resolves #985

### DIFF
--- a/fme/ace/registry/stochastic_sfno.py
+++ b/fme/ace/registry/stochastic_sfno.py
@@ -11,6 +11,7 @@ from fme.core.distributed.distributed import Distributed
 from fme.core.models.conditional_sfno.sfnonet import (
     Context,
     ContextConfig,
+    SFNONetConfig,
     get_lat_lon_sfnonet,
 )
 from fme.core.models.conditional_sfno.sfnonet import (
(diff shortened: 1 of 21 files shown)


elynnwu:
I would be in favor to raise an exception and not let the evaluation keep running, presumably if someone requests a specific variable, they actually plan to use it and so the job should fail.

Arcomano1234:
> I would be in favor to raise an exception and not let the evaluation keep running, presumably if someone requests a specific variable, they actually plan to use it and so the job should fail.

I don't know about everyone else but I tend to ignore most warnings in our logs and usually prefer things being raised. I assume its not possible to have this check during validate config right?

fme/ace/inference/data_writer/utils.py
```diff
@@ -17,8 +18,12 @@ def get_all_names(
         variables = variables.union(set(varnames))
     if allowlist is None:
         return variables
-    else:
-        return variables.intersection(set(allowlist))
+    missing = set(allowlist) - variables
```
Arcomano1234: Very optional: you could declare` allowlist = set(allowlist)` above this line and avoid making it a set twice

PR #992, opened 2026-03-19
Add metrics_log_dir option to LoggingConfig

mcgibbon:
This PR adds the ability to log wandb metrics to a local directory. Currently a jsonl file of scalar metrics is written, but this may be updated to include other metric types in later PRs.

Changes:
- Added DiskMetricLogger for logging scalar metrics to disk
- Wired DiskMetricLogger into WandB, MockWandB, and LoggingConfig
- Added `metrics_log_dir` option to LoggingConfig which, if given, will have wandb metrics logged to it.
- Require `step` when logging to WandB to facilitate disk logging and avoid missing logs when step is not passed.

- [x] Tests added

### DIFF
--- a/fme/core/benchmark/test_timer.py
+++ b/fme/core/benchmark/test_timer.py
@@ -23,6 +23,7 @@ def test_new_if_available(is_available: bool):
     not torch.cuda.is_available(),
     reason="CUDA is not available, skipping CUDATimer tests.",
 )
+@pytest.mark.serial  # should not run in parallel with xdist
 def test_timer_with_child():
     timer = CUDATimer()
     with timer:
--- a/scripts/data_process/Makefile
+++ b/scripts/data_process/Makefile
@@ -194,7 +194,7 @@ shield_som_c96_spin_up_dataset:
 
 .PHONY: shield_som_ensemble_c96_dataset
 shield_som_ensemble_c96_dataset:
-	./compute_dataset.sh --config configs/shield-som-ensemble-c96-$(RESOLUTION)-$(LAYERS).yaml
+	./compute_dataset.sh --dataset --stats --time-coarsen --config configs/shield-som-ensemble-c96-$(RESOLUTION)-$(LAYERS).yaml
 
 .PHONY: shield_som_abrupt_co2_increase_c96_dataset
 shield_som_abrupt_co2_increase_c96_dataset:
--- a/scripts/data_process/configs/shield-som-ensemble-c96-4deg-8layer.yaml
+++ b/scripts/data_process/configs/shield-som-ensemble-c96-4deg-8layer.yaml
@@ -22,10 +22,10 @@ runs:
   4xCO2-ic_0003: gs://vcm-ml-raw-flexible-retention/2024-07-03-C96-SHiELD-SOM/regridded-zarrs/gaussian_grid_45_by_90/4xCO2-ic_0003
   4xCO2-ic_0004: gs://vcm-ml-raw-flexible-retention/2024-07-03-C96-SHiELD-SOM/regridded-zarrs/gaussian_grid_45_by_90/4xCO2-ic_0004
   4xCO2-ic_0005: gs://vcm-ml-raw-flexible-retention/2024-07-03-C96-SHiELD-SOM/regridded-zarrs/gaussian_grid_45_by_90/4xCO2-ic_0005
-data_output_directory: gs://vcm-ml-intermediate/2024-07-09-vertically-resolved-4deg-c96-shield-som-ensemble-fme-dataset
+data_output_directory: gs://vcm-ml-intermediate/2026-01-28-vertically-resolved-4deg-c96-shield-som-ensemble-fme-dataset
 stats:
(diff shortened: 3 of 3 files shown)


mcgibbon:
Note the metrics logging does not require log_to_wandb=True.

fme/core/disk_metric_logger.py
```diff
@@ -0,0 +1,108 @@
...
+                        self._high_water_mark = step
+
+    def log(self, data: dict[str, Any], step: int) -> None:
+        """Log scalar metrics for a given step.
+
+        If ``step`` is at or below the high-water mark from a previous run,
+        the call is silently skipped.  Non-serializable values are dropped.
+        """
+        if self._high_water_mark is not None and step <= self._high_water_mark:
+            return
```
jpdunc23, line 61: WandB logging to a past step issues a warning. I think it would be good to do the same here for visibility.

fme/core/testing/wandb.py
jpdunc23, line 1: Agent review suggested the following to avoid leaking disk logger file handles:

```python
# fme/core/testing/wandb.py, inside mock_wandb()
try:
    yield wandb.singleton
finally:
    if hasattr(wandb.singleton, '_disk_logger') and wandb.singleton._disk_logger is not None:
        wandb.singleton._disk_logger.close()
    wandb.singleton = original
```

fme/core/disk_metric_logger.py
```diff
@@ -0,0 +1,108 @@
...
+        if self._file is not None:
+            self._file.close()
+            self._file = None
+
+
+def _extract_scalars(data: dict[str, Any]) -> dict[str, Any]:
+    """Return only JSON-serializable scalar entries from *data*."""
+    result: dict[str, Any] = {}
+    for key, value in data.items():
+        if isinstance(value, int | float | bool):
```
jpdunc23, line 80: From agent review:

> In `_extract_scalars`, `isinstance(value, float)` matches NaN and Inf, which bypass the `json.dumps` validation path. Python's `json.dumps` produces non-standard `NaN` / `Infinity` tokens (allowed by default via `allow_nan=True`), making the JSONL unreadable by strict JSON parsers. ML training metrics can legitimately be NaN (e.g., loss explosion).

Apparently this may lead to parsing errors for certain versions of `jq` and possibly for some `pandas.read_json` usages. I don't think we should change the behavior here, but maybe a brief note in the docs for `DiskMetricLogger` would be helpful to say that these values are possible or maybe just encourage use of the `read_metrics` util.

I guess NaN / Inf values are also not included in tests, might be worthwhile to add them.

fme/core/disk_metric_logger.py
```diff
@@ -0,0 +1,108 @@
...
+            result[key] = value
+        elif isinstance(value, str):
+            result[key] = value
+        else:
+            try:
+                json.dumps(value)
```
jpdunc23, line 86: From agent review:

> The fallback json.dumps(value) path accepts lists, dicts, None, etc. -- not just scalars. Consider renaming to `_extract_serializable`.

fme/core/disk_metric_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    Each line in the file is a JSON object with a "step" key and scalar metric
+    key-value pairs. On construction, any existing file is read to determine the
+    high-water mark (maximum step already logged). Subsequent calls to ``log``
+    with a step at or below that mark are silently skipped, which makes this
+    logger safe for job resumption.
+
+    Non-JSON-serializable values (e.g. images, tensors) are silently dropped.
+    """
+
+    def __init__(self, directory: str):
```
jpdunc23: From agent review:

> Accepting `str | os.PathLike` would be a minor ergonomic improvement.

fme/core/test_disk_metric_logger.py
```diff
@@ -0,0 +1,181 @@
...
+
+from fme.core.disk_metric_logger import METRICS_FILENAME, DiskMetricLogger, read_metrics
+
+
+@pytest.fixture
+def log_dir(tmp_path):
+    return str(tmp_path / "metrics")
+
+
+class TestDiskMetricLogger:
```
jpdunc23: Nit: Remove the `class` def here. 

I see agents using these test class defs a lot, which I think it can be useful to group related tests (as in `test_wandb.py` in this PR) or to simplify test naming since it allows for reuse of the same `test_` method name in distinct classes. But in this case I think it just adds unnecessary nesting.

fme/core/testing/wandb.py
mcgibbon: Done, added cleanup in the finally block of mock_wandb.

fme/core/disk_metric_logger.py
mcgibbon: Done, added a logging.warning with the step and high-water mark values.

fme/core/disk_metric_logger.py
mcgibbon: Added a test for NaN/Inf round-tripping through read_metrics, with a docstring noting the strict-parser caveat.

fme/core/disk_metric_logger.py
mcgibbon: Done, renamed to _extract_serializable.

PR #993, opened 2026-03-19
add backward pass for spatial prallellism

mahf708:
add backward pass for spatial prallellism with three changes:
- an autograd version of spatial_reduce_sum to deal with the loss path
- hook up the parameters with a grad hook
- setting broadcast_buffers to False, otherwise corrupting/mutating in-place

there might be some assumptions about what types of loss we are supporting thus far, and a careful auditing will be required. An alternative to this formulation would be an explicit and clear spatial awarness in the loss calculation elsewhere ... but that may be invasive?

- [x] Tests added


fme/ace/stepper/test_single_module_csfno.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+    output_dict = get_train_outputs_tensor_dict(result1, result2)
+    validate_tensor_dict(
+        output_dict,
+        os.path.join(
+            DIR,
+            "testdata/csfno_stepper_train_on_batch_with_optimization_regression.pt",
+        ),
+        atol=1e-2,
+        rtol=1e-2,
```
mahf708, line 218: seems too low, and thus I have low confidence in all of this for now

fme/core/distributed/model_torch_distributed.py
```diff
@@ -42,6 +43,35 @@
...
+        group: torch.distributed.ProcessGroup,
+    ) -> torch.Tensor:
+        output = input.clone()
+        torch.distributed.all_reduce(output, group=group)
+        return output
+
+    @staticmethod
+    @custom_bwd(device_type="cuda")
+    def backward(ctx, grad_output: torch.Tensor):
+        return grad_output.clone(), None
```
mahf708, line 73: the cloning is probably unnecessary here

fme/core/distributed/model_torch_distributed.py
```diff
@@ -307,31 +337,73 @@ def _device_ids(self) -> list[int] | None:
...
+
+        def _hook(grad: torch.Tensor) -> torch.Tensor:
+            if grad is None:
+                return grad
+
+            reduced = grad.contiguous().clone()
+            torch.distributed.all_reduce(reduced, group=spatial_group)
+
+            # If we want mean gradient instead of sum, we want:
+            # reduced /= (self._h_size * self._w_size)
```
mahf708, line 392: I can't quite wrap my head around which one we really need tbh, and I think this is linked with how we do losses here...

fme/core/distributed/model_torch_distributed.py
```diff
@@ -42,6 +43,35 @@
...
+    ) -> torch.Tensor:
+        output = input.clone()
+        torch.distributed.all_reduce(output, group=group)
+        return output
+
+    @staticmethod
+    @custom_bwd(device_type="cuda")
+    def backward(ctx, grad_output: torch.Tensor):
+        return grad_output.clone(), None
+
```
mahf708, line 74: identity in backward may become an issue... we may need an all-reduce in backward, but that's kind of what the hook below is doing

fme/core/distributed/parallel_tests/test_backward_step.py
```diff
@@ -0,0 +1,135 @@
...
+    torch.manual_seed(0)
+
+    # Run forwards/backwards
+    loss, grad = _run_forward_backward(img_shape)
+
+    # Only root does I/O
+    if not dist.is_root():
+        return
+
+    if not BASELINE_FILE.exists():
```
mcgibbon, line 95: Please use the regression helper in `fme/core/testing/regression.py` for this, to reduce duplication of this kind of regression logic. It should be general-purpose and support this use case.

mcgibbon: A "bug" here is that this will pass when no baseline exists.

fme/ace/stepper/test_single_module_csfno.py
```diff
@@ -0,0 +1,245 @@
...
+from fme.core.registry.module import ModuleSelector
+from fme.core.step import SingleModuleStepConfig, StepSelector
+from fme.core.testing.regression import validate_tensor_dict
+from fme.core.typing_ import EnsembleTensorDict
+
+DIR = os.path.abspath(os.path.dirname(__file__))
+TIMESTEP = datetime.timedelta(hours=6)
+
+
+def get_dataset_info(
```
mcgibbon, line 42: This is a lot of duplicated testing logic, do we need a new testing file or can we make the existing test in test_single_module.py parallel-enabled / put this parallel-enabled test in that file to use its helpers?

scripts/testing/test_spatial.sh
mcgibbon, line 1: I'd like to avoid new ways of executing tests, if we need something can we put this in the Makefile, or if it's temporary for this PR can you leave it uncommitted? I think the current make targets support this in two executions (in serial and then in parallel), using the TEST_PATH argument.

mcgibbon (commented):
It seems like the core contribution of this script is that it adds proper gradient reductions to `spatial_reduce_sum`. It would be nice to see a unit test added that fails before you add those changes and passes after you add those changes. I think we could and should merge a PR with just those changes.

It feels like the large rtol and atol are needed because the test isn't really passing, and that we'll need some finer-grained tests to tell us what part of the code is causing the failure. Splitting the spatial_reduce_sum changes out will let us get them merged without blocking on this, which could take a while and will need a lot more test code to be written/reviewed.

fme/ace/stepper/test_single_module_csfno.py
mahf708: That test uses the legacy SFNO --- could we change it to use CSFNO?

mahf708:
@mcgibbon note that large tol is only need for GPU/CPU inconsistency, but it is ok otherwise (i.e., it is ok if we run everything on cpu or gpu --- that may simplify the hunt or be reassiuring? idk)

also, @mcgibbon, thoguhts on this potential gotcha: besides the forward ops that we want to be differentiable, etc., I don't think we want this thing to be differentiable, right? If so, we can potentially change this to use a bare all_reduce here. What do you think? I think this is largely ok for forward, but if we ever add backward to the spatial_reduce_sum, it may cause problems?

```py
    def gather_spatial(
        self, data: dict[str, torch.Tensor], img_shape: tuple[int, int]
    ) -> dict[str, torch.Tensor]:
        """Gather local spatial chunks back to global tensors via all-reduce."""
        return {k: self.gather_spatial_tensor(v, img_shape) for k, v in data.items()}

    def gather_spatial_tensor(
        self, tensor: torch.Tensor, img_shape: tuple[int, int]
    ) -> torch.Tensor:
        """Reassemble a spatially-sharded tensor on every rank via all-reduce.

        Args:
            tensor: Local spatial shard.
            img_shape: Global ``(H, W)`` spatial dimensions.
        """
        if img_shape == tensor.shape[-2:]:
            return tensor
        global_shape = (*tensor.shape[:-2], *img_shape)
        slices = self.get_local_slices(img_shape)
        buf = torch.zeros(global_shape, dtype=tensor.dtype, device=tensor.device)
        buf[(..., *slices)] = tensor
        return self.spatial_reduce_sum(buf)
```

fme/core/distributed/model_torch_distributed.py
```diff
@@ -42,6 +44,35 @@
...
+    This makes ``spatial_reduce_sum`` differentiable so that gradients
+    flow correctly through the loss computation path::
+        AreaWeightedMSELoss → area_weighted_mean → weighted_mean
+            → spatial_reduce_sum (uses this function)
+    Without this, the raw ``torch.distributed.all_reduce`` would break
+    the autograd graph because it is an in-place, non-differentiable op.
+    """
+
+    @staticmethod
+    @custom_fwd(device_type="cuda")
```
mcgibbon, line 60: Why is custom_fwd needed? Don't you need this on CPU? Is this why CPU and GPU are giving different results?

mcgibbon:
> @mcgibbon note that large tol is only need for GPU/CPU inconsistency, but it is ok otherwise (i.e., it is ok if we run everything on cpu or gpu --- that may simplify the hunt or be reassiuring? idk)
> 

We should be able to get pretty close results on both. Just make sure that initializations always happen on CPU, and then get transferred to get_device(), rather than being initialized directly on-device. I marked one location that might be causing this difference with a line comment.

> also, @mcgibbon, thoguhts on this potential gotcha: besides the forward ops that we want to be differentiable, etc., I don't think we want this thing to be differentiable, right? If so, we can potentially change this to use a bare all_reduce here. What do you think? I think this is largely ok for forward, but if we ever add backward to the spatial_reduce_sum, it may cause problems?
> 
> ```python
>     def gather_spatial(
>         self, data: dict[str, torch.Tensor], img_shape: tuple[int, int]
>     ) -> dict[str, torch.Tensor]:
>         """Gather local spatial chunks back to global tensors via all-reduce."""
>         return {k: self.gather_spatial_tensor(v, img_shape) for k, v in data.items()}
> 
>     def gather_spatial_tensor(
>         self, tensor: torch.Tensor, img_shape: tuple[int, int]
>     ) -> torch.Tensor:
>         """Reassemble a spatially-sharded tensor on every rank via all-reduce.
> 
>         Args:
>             tensor: Local spatial shard.
>             img_shape: Global ``(H, W)`` spatial dimensions.
>         """
>         if img_shape == tensor.shape[-2:]:
>             return tensor
>         global_shape = (*tensor.shape[:-2], *img_shape)
>         slices = self.get_local_slices(img_shape)
>         buf = torch.zeros(global_shape, dtype=tensor.dtype, device=tensor.device)
>         buf[(..., *slices)] = tensor
>         return self.spatial_reduce_sum(buf)
> ```

Does your added code make anything slower? I don't really expect so? When we are in non-differentiable places we use a no_grad context to handle not running extra logic. I'd say don't worry about using the differentiable version.

fme/core/distributed/model_torch_distributed.py
mahf708: No idea, I saw this in the makani/physicsnemo repos and I copied it blindly (testing with it and without didn't really make any difference, or at least I didn't notice it)

mcgibbon: They don’t have spatial parallelism on cpu, so it wouldn’t cause issues for them, but we should remove it. I did in my branch incorporating this code.

mcgibbon:
https://github.com/ai2cm/ace/pull/995 is merged to main which I think includes the (passing) critical components of this PR - let me know what you think. It does seem like the CSFNO step is a "fail" in the current state of this PR.

Next steps from current main are to add in a test or a RegressionCase which covers the single module CSFNO the way you have here, and for it to pass with good tolerances. If you're mostly getting differences between CPU and GPU, then merging main and using the version of your autograd function that's in main may help the issue - I think the special logic is needed for both CPU and GPU, not just GPU.

frodre:
Closing due to inactivity

PR #994, opened 2026-03-20
Compute optional left/two-tailed pdf metrics

AnnaKwa:
- This adds `ComparedDynamicTailsHistograms` which inherits from `ComparedDynamicHistograms` but adds the option to specify which variables are left/both-tailed. The percentile-based metrics will be computed for the complementary 100-p percentiles for each percentile in `ComparedDynamicTailsHistograms.percentiles`. By default, variables not in either list are assumed to be upper-tailed. The percentiles arg is assumed to be provided for the upper tail, and the complementary values are calculated by the class. I made a new class to minimize changes to the existing class (I cleaned up some of the `get_wandb` code for calculating percentiles) since it is used outside of downscaling. The new optional args could be instead put into the existing `ComparedDynamicHistogram` class instead of creating a new subclass- I am open to opinions. 

- Changed the downscaling trainer to instead use the `"generation/histogram/prediction_frac_of_target/99.9999th-percentile"` field for choosing the best histogram tail checkpoint. Small/zero target bins can have an outsize weight on the value of the previously used abs_norm_tail_bias metric. For this reason I did not include the `abs_norm_tail_bias` metrics in the new histogram class. I can remove them from the existing `ComparedDynamicHistograms` as well in a later PR if people are ok with this. I don't think anyone else uses this metric.


fme/downscaling/aggregators/main.py
```diff
@@ -758,6 +758,11 @@ def __init__(
         ssim_kwargs: Mapping[str, Any] | None = None,
         variable_metadata: Mapping[str, VariableMetadata] | None = None,
         include_positional_comparisons: bool = True,
+        two_tailed_variables: list[str] | None = [
```
frodre: Missing additional arg descriptions above.

frodre: nit from claude: using mutable lists means if it gets adjusted by any subsequent code then it will get adjusted for all follow on default initializations.  Safer to use default `None` and then set w/ lists in the `__init__`.  I would reference the defaults in the docstring since this reduces visibility.

fme/downscaling/train.py
```diff
@@ -465,10 +471,42 @@ def configure_logging(self, log_filename: str):
         )
 
 
+def _get_complement_percentile_prefix(prefix):
+    """Given a prefix containing a percentile value, return a prefix with
+    100 minus that percentile. Returns None if no percentile pattern is found.
```
frodre, line 475: Maybe add a note about which sources of keys this is currently set to cover since I can't read the regexp w/ out an explainer.

fme/core/histogram.py
```diff
@@ -512,6 +536,104 @@ def get_dataset(self) -> xr.Dataset:
...
+    lower tail.
+    """
+
+    def __init__(
+        self,
+        n_bins: int,
+        percentiles: list[float] | None = None,
+        compute_percentile_frac: bool = True,
+        two_tailed_variables: list[str] | None = None,
+        left_tailed_variables: list[str] | None = None,
```
frodre, line 555: I think it's fine to default to "upper" for non-specified but for completeness, maybe also allow the option for specification `right_tailed_variables` so intention could be clear within the instantiation.  We could also consider allowing users to change the default `default_tail = Literal["upper", ...]` depending on which class holds the most variables.

fme/core/histogram.py
```diff
@@ -512,6 +536,104 @@ def get_dataset(self) -> xr.Dataset:
...
+    def _variable_distribution_tail(
+        self, field_name: str
+    ) -> Literal["upper", "lower", "both"]:
+        if field_name in self._two_tailed_variables:
+            return "both"
+        if field_name in self._left_tailed_variables:
+            return "lower"
+        return "upper"
+
+    def _percentile_entries_for_field(self, field_name: str) -> list[tuple[float, str]]:
```
frodre: Looks like this isn't used anywhere

fme/core/histogram.py
```diff
@@ -512,6 +536,104 @@ def get_dataset(self) -> xr.Dataset:
...
+            if tail in ["lower", "both"]:
+                p_key = _format_percentile_for_metric_key(100.0 - p, p)
+                percentile_metrics[f"{p_key}th-percentile/{field_name}"] = quantile(
+                    histogram.bin_edges, histogram.counts, (100.0 - p) / 100.0
+                )
+        return percentile_metrics
+
+    def get_wandb(self) -> dict[str, float]:
+        return_dict: dict[str, matplotlib.figure.Figure | float] = {}
+        target_dict: dict[str, float] = {}
```
frodre: Rename to something like `target_metrics` and `return_results` rather than `_dict` to make the usage clearer in the fraction calc below and since that's already specified by the typing.

fme/downscaling/aggregators/main.py
frodre: Or we could do the hard coding at the point of usage in `train.py`

frodre (approved):
Thanks, @AnnaKwa.  Approved with a few small suggestion; nothing I need to look at again.  I'm okay with a follow up removing the bias from the other aggregator.

PR #995, opened 2026-03-20
Add spatial parallel regression tests and fix differentiable reduce

mcgibbon:
Fix spatial parallelism backward pass and add regression test infrastructure.

`spatial_reduce_sum` used an in-place `all_reduce` that broke the autograd graph,
preventing gradients from flowing through the loss computation path. Spatial ranks
also did not aggregate parameter gradients, so each rank applied different weight
updates.

Changes:
- `fme.core.distributed.model_torch_distributed._AutogradAllReduce`: new
  `torch.autograd.Function` wrapping `all_reduce` with identity backward, making
  `spatial_reduce_sum` differentiable
- `fme.core.distributed.model_torch_distributed.ModelTorchDistributed.wrap_module`:
  register per-parameter gradient hooks that all-reduce across spatial ranks;
  set `broadcast_buffers=False` to protect SHT/iSHT Legendre buffers
- `fme.core.distributed.parallel_tests.test_regression`: new parameterized
  regression test framework (`RegressionCase` base class) that validates
  forward → backward → forward correctness across spatial decompositions
- `AGENTS.md`: document parallel test commands, env vars, and baseline workflow

### DIFF
--- a/fme/core/benchmark/run.py
+++ b/fme/core/benchmark/run.py
@@ -144,7 +144,7 @@ def main(
             entity=entity,
             name=wandb_name,
         )
-        wandb.log(wandb_logs, commit=True)
+        wandb.log(wandb_logs, step=0, commit=True)
     return 0
 
 
--- a/fme/core/disk_metric_logger.py
+++ b/fme/core/disk_metric_logger.py
@@ -1,0 +1,114 @@
+import io
+import json
+import logging
+import os
+from typing import Any
+
+METRICS_FILENAME = "metrics.jsonl"
+
+
+class DiskMetricLogger:
+    """Logs scalar metrics to a JSONL file on disk.
+
+    Each line in the file is a JSON object with a "step" key and scalar metric
+    key-value pairs. On construction, any existing file is read to determine the
+    high-water mark (maximum step already logged). Subsequent calls to ``log``
+    with a step at or below that mark are silently skipped, which makes this
+    logger safe for job resumption.
+
+    Non-JSON-serializable values (e.g. images, tensors) are silently dropped.
+    """
+
+    def __init__(self, directory: str | os.PathLike):
+        os.makedirs(directory, exist_ok=True)
+        self._path = os.path.join(directory, METRICS_FILENAME)
+        self._high_water_mark: int | None = None
+        self._file: io.TextIOWrapper | None = None
+        self._read_high_water_mark()
+        self._file = open(self._path, "a")
+
+    def _read_high_water_mark(self):
+        """Read the existing file (if any) to find the max step logged."""
+        if not os.path.exists(self._path):
+            return
+        with open(self._path) as f:
+            for line in f:
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    record = json.loads(line)
+                except json.JSONDecodeError:
+                    continue
+                step = record.get("step")
+                if isinstance(step, int):
+                    if self._high_water_mark is None or step > self._high_water_mark:
+                        self._high_water_mark = step
+
+    def log(self, data: dict[str, Any], step: int) -> None:
+        """Log scalar metrics for a given step.
+
+        If ``step`` is at or below the high-water mark from a previous run,
+        the call is silently skipped.  Non-serializable values are dropped.
+        """
+        if self._high_water_mark is not None and step <= self._high_water_mark:
+            logging.warning(
+                "DiskMetricLogger: skipping log for step %d "
+                "(at or below high-water mark %d)",
+                step,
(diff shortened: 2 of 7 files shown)


fme/core/distributed/model_torch_distributed.py
```diff
@@ -43,6 +44,35 @@
...
+    This makes ``spatial_reduce_sum`` differentiable so that gradients
+    flow correctly through the loss computation path::
+        AreaWeightedMSELoss → area_weighted_mean → weighted_mean
+            → spatial_reduce_sum (uses this function)
+    Without this, the raw ``torch.distributed.all_reduce`` would break
+    the autograd graph because it is an in-place, non-differentiable op.
+    """
+
+    @staticmethod
+    @custom_fwd(device_type="cuda")
```
elynnwu: Is this strictly a GPU test? Does it get skipped on CPU?

fme/core/distributed/parallel_tests/test_regression.py
```diff
@@ -0,0 +1,156 @@
...
+    def forward(self, data: TensorDict) -> torch.Tensor:
+        return self.conv(data["x"])
+
+
+@dataclasses.dataclass
+class _LinearCase(RegressionCase):
+    """1x1 Conv2d applied channel-wise; verifies basic gradient flow."""
+
+    n_channels: int = 4
+    _img_shape: tuple[int, int] = (8, 16)
```
elynnwu: [nit] can you just have this as `img_shape` and not have a property?

AGENTS.md
```diff
@@ -27,6 +27,24 @@ This is a Python machine learning project for atmospheric modeling (ACE - AI2 Cl
...
+FME_FORCE_CPU=1 FME_DISTRIBUTED_BACKEND=model FME_DISTRIBUTED_H=2 FME_DISTRIBUTED_W=1 \
+  torchrun --nproc-per-node 2 -m pytest -m parallel .
+```
+Full matrix (8 configs, ~3-4 min): `make cpu_test_all_parallel`
+Narrow to specific tests: `make cpu_test_all_parallel TEST_PATH=fme/core/distributed/parallel_tests/test_step.py`
+
+Regression tests using `.pt` baselines must generate the baseline with a
+single-rank `python -m pytest` run, then verify under spatial parallel via
+torchrun. Generating baselines under the same backend you test against does
+not validate cross-backend correctness.
```
elynnwu, line 46: [nit] I don't understand this comment to the agent, we don't allow the agent to generate or delete the baseline `.pt` right?

fme/core/distributed/model_torch_distributed.py
```diff
@@ -305,31 +335,73 @@ def _device_ids(self) -> list[int] | None:
...
+
+        def _hook(grad: torch.Tensor) -> torch.Tensor:
+            if grad is None:
+                return grad
+
+            reduced = grad.contiguous().clone()
+            torch.distributed.all_reduce(reduced, group=spatial_group)
+
+            # If we want mean gradient instead of sum, we want:
+            # reduced /= (self._h_size * self._w_size)
```
elynnwu: We always want the sum right? Doesn't each rank get the partial gradient that is already area weighted, so summing them should be the correct mean?

AGENTS.md
mcgibbon: We do, or at least I do - if the file is deleted (which you can let it do on a per-command basis) and the test is re-run, the file gets re-generated.

mcgibbon: Also, if Claude adds a new test, it needs to generate the baseline pt file.

fme/core/distributed/model_torch_distributed.py
mcgibbon: Yes, deleted.

fme/core/distributed/model_torch_distributed.py
mcgibbon: Ah yes I had fixed this locally but not pushed. Now it's pushed.

fme/core/distributed/parallel_tests/test_regression.py
mcgibbon: Done.

fme/core/distributed/model_torch_distributed.py
mahf708: Not always, right? Some averages aren't weighted... or maybe I misread the code. For the production ones, like AreaWeightedMSE, we are covered

mahf708: I kept it in the other PR because I wasn't 100% sure which losses are used and which ones etc. and what general case we should support

PR #998, opened 2026-03-21
Update StaticInputs coordinate handling

frodre:
PR 2/3 for #971 split

Changes:

- Move lat/lon coordinates from individual `StaticInput` fields to the parent `StaticInputs` container
- `StaticInput.subset` now takes `slice` arguments directly instead of `ClosedInterval`s
- Extract `_load_coords_from_ds` with flexible coordinate name matching (`lat`/`latitude`/`grid_yt`, etc.)
- Add `StaticInputs.from_state_backwards_compatible` to handle legacy checkpoints (coords per-field) and config-based loading (no coords in checkpoint)
- `load_static_inputs` now requires at least one field and validates coordinate consistency across fields

- [x] Tests added

### DIFF
--- a/fme/downscaling/data/test_static.py
+++ b/fme/downscaling/data/test_static.py
@@ -127,7 +127,7 @@ def test_from_state_backwards_compatible_raises_state_and_config():
         )
 
 
-def test_load_fine_coords_from_path():
+def test__load_coords_from_ds():
     lat = [0.0, 1.0, 2.0]
     lon = [10.0, 20.0, 30.0, 40.0]
     ds = xr.Dataset(coords={"lat": lat, "lon": lon})


fme/downscaling/data/static.py
```diff
@@ -139,48 +160,88 @@ def subset(
...
+                StaticInput.from_state(field_state) for field_state in state["fields"]
+            ],
+            coords=LatLonCoordinates(
+                lat=state["coords"]["lat"],
+                lon=state["coords"]["lon"],
+            ),
         )
 
+    @classmethod
+    def from_state_backwards_compatible(
```
frodre, line 196: This method gets used in #999, but there will be a `CheckpointModelConfig`  gate on calling this without `coords` available in the state or the configuration.

fme/downscaling/data/static.py
```diff
@@ -34,100 +28,127 @@ def shape(self) -> tuple[int, int]:
...
+        lon=torch.tensor(ds[lon_name].values, dtype=torch.float32),
+    )
 
-    def get_state(self) -> dict:
-        return {
-            "data": self.data.cpu(),
-            "coords": self.coords.get_state(),
-        }
+
+def load_fine_coords_from_path(path: str) -> LatLonCoordinates:
```
AnnaKwa: It looks like this code is duplicated in `_get_normalized_static_input`. Can this helper and its tests be removed? I think `_load_coords_from_ds` is simple enough that it doesn't need multiple tests.

AnnaKwa (approved):
Mostly LGTM, one request that doesn't need to be reviewed.

PR #999, opened 2026-03-21
Add fine_coordinates to model with validation against StaticInputs

frodre:
PR 3/3 for #971 split

Changes
- Add `full_fine_coords: LatLonCoordinates` to `DiffusionModel` as the canonical source of truth for the fine-resolution grid, decoupled from `StaticInputs`
- Validate at build time that `static_inputs.coords` matches `full_fine_coords` when both are provided
- Remove dead runtime guard in `_get_input_from_coarse` for missing static inputs when `use_fine_topography=True`; this is now enforced at build time
- Add `fine_coordinates_path` to `CheckpointModelConfig` for backwards-compatible loading of old checkpoints that lack stored fine coordinates
- Add `StaticInputs.from_state_backwards_compatible` for checkpoint loading that handles both old and new state formats
- Remove `_downscale_coord` bandaid from `predict.py`; `get_fine_coords_for_batch` now uses `full_fine_coords` directly
- Add `fine_coords` field to `PairedGriddedData` and thread it through to `TrainerConfig.build`

- [x] Tests added

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -387,28 +387,21 @@ class DiffusionModel:
         )
         interpolated = interpolate(normalized, self.downscale_factor)
 
-        if self.config.use_fine_topography:
-            if static_inputs is None or not static_inputs.fields:
+        if self.config.use_fine_topography and static_inputs is not None:
+            expected_shape = interpolated.shape[-2:]
+            if static_inputs.shape != expected_shape:
                 raise ValueError(
-                    "Static inputs must be provided for each batch when flag "
-                    "use_fine_topography is enabled, but no static input fields "
-                    "were found."
+                    f"Subsetted static input shape {static_inputs.shape} does not "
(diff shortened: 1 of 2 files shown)


fme/downscaling/data/static.py
```diff
@@ -62,9 +62,22 @@ def _load_coords_from_ds(ds: xr.Dataset) -> LatLonCoordinates:
...
 
+def _open_ds_from_path(path: str) -> xr.Dataset:
+    if path.endswith(".zarr"):
+        ds = xr.open_zarr(path, mask_and_scale=False)
+    else:
+        ds = xr.open_dataset(path, mask_and_scale=False)
+    return ds
+
+
+def load_coords_from_path(path: str) -> LatLonCoordinates:
```
frodre, line 73: Re-added to use in `CheckpointModelConfig`.  Added the small helper `_open_ds_from_path` to de-dupe code

PR #1001, opened 2026-03-23
Add AnkurLocalNet and LocalNet model implementations

mcgibbon:
## Summary

- Add `AnkurLocalNet`: a simple 3-hidden-layer MLP replicating Ankur's diagnostic architecture from `ColumnDiagnosticSphericalFourierNeuralOperatorNet`, with optional DISCO convolution encoder and positional embedding.
- Add `LocalNet`: a local neural operator network using DISCO convolutions and/or 1x1 convolutions with encoder/decoder structure, conditional layer normalization, skip connections, and LoRA support.
- Both modules accept the `(x, context: Context)` forward signature used by the conditional SFNO, enabling composition with noise/label conditioning wrappers.
- Add regression baselines for both networks.
- Remove `HybridNet` (superseded by secondary network support in the follow-on PR).

- [x] Tests added

### DIFF
--- a/fme/downscaling/data/__init__.py
+++ b/fme/downscaling/data/__init__.py
@@ -12,7 +12,7 @@ from .datasets import (
     PairedBatchItem,
     PairedGriddedData,
 )
-from .static import StaticInput, StaticInputs, load_static_inputs
+from .static import StaticInput, StaticInputs, load_coords_from_path, load_static_inputs
 from .utils import (
     BatchedLatLonCoordinates,
     ClosedInterval,
--- a/fme/downscaling/data/config.py
+++ b/fme/downscaling/data/config.py
@@ -23,7 +23,11 @@ from fme.downscaling.data.datasets import (
     PairedBatchData,
     PairedGriddedData,
 )
-from fme.downscaling.data.utils import ClosedInterval, adjust_fine_coord_range
+from fme.downscaling.data.utils import (
+    ClosedInterval,
+    adjust_fine_coord_range,
+    get_latlon_coords_from_properties,
+)
 from fme.downscaling.requirements import DataRequirements
 
 
@@ -529,6 +533,7 @@ class PairedDataLoaderConfig:
             dims=example.fine.latlon_coordinates.dims,
             variable_metadata=variable_metadata,
             all_times=all_times,
+            fine_coords=get_latlon_coords_from_properties(properties_fine),
         )
 
     def _get_sampler(
--- a/fme/downscaling/data/datasets.py
+++ b/fme/downscaling/data/datasets.py
@@ -337,6 +337,7 @@ class PairedGriddedData:
     dims: list[str]
     variable_metadata: Mapping[str, VariableMetadata]
     all_times: xr.CFTimeIndex
+    fine_coords: LatLonCoordinates
 
     @property
     def loader(self) -> DataLoader[PairedBatchItem]:
--- a/fme/downscaling/data/static.py
+++ b/fme/downscaling/data/static.py
@@ -62,9 +62,22 @@ def _load_coords_from_ds(ds: xr.Dataset) -> LatLonCoordinates:
     )
 
 
+def _open_ds_from_path(path: str) -> xr.Dataset:
+    if path.endswith(".zarr"):
+        ds = xr.open_zarr(path, mask_and_scale=False)
+    else:
+        ds = xr.open_dataset(path, mask_and_scale=False)
+    return ds
+
+
+def load_coords_from_path(path: str) -> LatLonCoordinates:
+    ds = _open_ds_from_path(path)
+    return _load_coords_from_ds(ds)
+
+
 def _get_normalized_static_input(
     path: str, field_name: str
-) -> tuple["StaticInput", LatLonCoordinates]:
+) -> tuple[StaticInput, LatLonCoordinates]:
     """
     Load a static input field from a given file path and field name and
     normalize it.
@@ -74,11 +87,7 @@ def _get_normalized_static_input(
 
     Raises ValueError if lat/lon coordinates are not found in the dataset.
     """
-    if path.endswith(".zarr"):
-        ds = xr.open_zarr(path, mask_and_scale=False)
-    else:
-        ds = xr.open_dataset(path, mask_and_scale=False)
-
+    ds = _open_ds_from_path(path)
     coords = _load_coords_from_ds(ds)
     da = ds[field_name]
 
--- a/fme/downscaling/inference/test_inference.py
+++ b/fme/downscaling/inference/test_inference.py
@@ -277,7 +277,12 @@ def checkpointed_model_config(
     # loader_config is passed in to add static inputs into model
     # that correspond to the dataset coordinates
     static_inputs = load_static_inputs({"HGTsfc": f"{data_paths.fine}/data.nc"})
-    model = model_config.build(coarse_shape, 2, static_inputs=static_inputs)
+    model = model_config.build(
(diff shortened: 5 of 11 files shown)


fme/core/models/conditional_sfno/ankur.py
```diff
@@ -0,0 +1,162 @@
...
+            out_chans,
+            in_shape=img_shape,
+            out_shape=img_shape,
+            kernel_shape=kernel_shape,
+            basis_type="morlet",
+            basis_norm_mode="mean",
+            groups=math.gcd(in_chans, out_chans),
+            grid_in=data_grid,
+            grid_out=data_grid,
+            bias=False,
```
Arcomano1234, line 65: I know this is coming from Ankur/ Nvidia, but it is interesting that they don't include biases in this conv as this is used for the encoder

fme/core/models/conditional_sfno/localnet.py
```diff
@@ -0,0 +1,487 @@
...
+            embed_dim,
+            embed_dim,
+            1,
+            bias=True,
+            lora_rank=lora_rank,
+            lora_alpha=lora_alpha,
+        )
+
+    def forward(self, x, timer: Timer = NullTimer()):
+        return self.conv(x), x
```
Arcomano1234, line 86: Question: Why is this returning both conv(x) and x?

Arcomano1234: Never mind its to match the behavior of `DiscreteContinuousConvS2`

fme/core/models/conditional_sfno/ankur.py
```diff
@@ -0,0 +1,162 @@
...
+    return AnkurLocalNet(
+        params,
+        img_shape=img_shape,
+        in_chans=in_chans,
+        out_chans=out_chans,
+        data_grid=data_grid,
+    )
+
+
+class AnkurLocalNet(nn.Module):
```
Arcomano1234, line 107: High level thought. Compared to `LocalNet` there is no way to get context into Ankur's network do we think this is important for training a stochastic model?

fme/core/models/conditional_sfno/hybrid.py
```diff
@@ -0,0 +1,168 @@
...
+        img_shape: Spatial dimensions (lat, lon) of the input data.
+        embed_dim_labels: Dimension of label embeddings for conditional
+            layer normalization. 0 disables label conditioning.
+
+    Returns:
+        A configured HybridNet instance.
+    """
+    n_in = n_forcing_channels + n_prognostic_channels
+
+    context_config = ContextConfig(
```
Arcomano1234: Question: Having the context for both the backbone and local network not have noise context intentional? Wouldn't this mean we can't train a stochastic backbone?

fme/core/models/conditional_sfno/localnet.py
```diff
@@ -0,0 +1,487 @@
...
+        filter_type="disco",
+        data_grid="equiangular",
+        lora_rank: int = 0,
+        lora_alpha: float | None = None,
+    ):
+        super().__init__()
+
+        if filter_type == "disco":
+            nlat, nlon = img_shape
+            theta_cutoff = 2 * _compute_cutoff_radius(
```
Arcomano1234, line 105: Question: this is 2x larger than what Ankur is there a reason to double the cut off threshold?

fme/core/models/conditional_sfno/ankur.py
```diff
@@ -0,0 +1,162 @@
...
+class AddPosEmbed(nn.Module):
+    def __init__(self, embed_dim, img_shape):
+        super().__init__()
+        h, w = img_shape
+        self.pos_embed = nn.Parameter(torch.zeros(1, embed_dim, h, w))
+        self.pos_embed.is_shared_mp = ["matmul"]
+        trunc_normal_(self.pos_embed, std=0.02)
+
+    def forward(self, x):
+        return x + self.pos_embed
```
Arcomano1234, line 82: Should be have this be distributed "ready". E.g. `x = x + self.pos_embed[..., self._spatial_h_slice, self._spatial_w_slice]`

mcgibbon: Maybe, but also we should assume any code that is not distributed-tested is not distributed-enabled.

fme/core/models/conditional_sfno/hybrid.py
mcgibbon: This file is deleted in the follow-on PR, can we ignore it for now? I'd like to merge this code and then merge the deletion, just to not have to handle the merge conflict(s).

fme/core/models/conditional_sfno/localnet.py
mcgibbon: Not sure, it doesn't look like makani doubles it either (e.g. https://github.com/NVIDIA/makani/blob/main/makani/models/networks/fourcastnet3.py). I'll check with Claude.

fme/core/models/conditional_sfno/localnet.py
```diff
@@ -0,0 +1,487 @@
...
+                basis_type="morlet",
+            )
+            self.filter = DiscreteContinuousConvS2(
+                embed_dim,
+                embed_dim,
+                in_shape=img_shape,
+                out_shape=img_shape,
+                kernel_shape=(3, 3),
+                basis_type="morlet",
+                basis_norm_mode="mean",
```
mcgibbon, line 117: Ooh, I'm curious to know if basis_norm_mode leads to zero-mean on these filters.

fme/core/models/conditional_sfno/ankur.py
mcgibbon: Yes, it would be important for the component being stochastic. Ankur's network is strictly deterministic, and it has no norms so there isn't an obvious way to condition it on noise.

fme/core/models/conditional_sfno/ankur.py
mcgibbon: Agreed

fme/core/models/conditional_sfno/localnet.py
mcgibbon: OK, in FCN3 they use 1x for the encoder and decoder but 2x for the blocks within the network. Not clear to me why, but that's what they do. 2x is more consistent with FCN3, so I think we should use that for now.

Honestly, IMO this should be a configurable parameter, not determined automatically based on the domain size. The input receptive field (as an angle on the sphere) should be determined by the length of the forward timestep, not the resolution of the data.

fme/core/models/conditional_sfno/hybrid.py
mcgibbon: Ok actually I'll delete it in this PR.

fme/core/models/conditional_sfno/localnet.py
Arcomano1234: I agree but for now lets keep it consistent with FCN3

fme/core/models/conditional_sfno/ankur.py
Arcomano1234: Agreed I would say it doesn't make sense to use this module when training a stochastic model. It might be worth warning the user that maybe?

fme/core/models/conditional_sfno/ankur.py
Arcomano1234: Yeah that is true. This is a discuss for another topic / PR but this will error out if used in a domain parallel distributed training, are we planning to potentially valid configurations once the distributed code is more mature?

PR #1002, opened 2026-03-24
Add secondary network support to SingleModuleStep and register local network types

mcgibbon:
## Summary

- Extend `SingleModuleStepConfig` with optional `secondary_builder`, `secondary_out_names`, and `secondary_residual_names` to support a secondary network that receives the same input as the backbone. The secondary network can predict diagnostic-only variables and/or add residual corrections to the backbone's prognostic outputs.
- Add `AnkurLocalNet` (simple MLP) and `LocalNet` (DISCO/conv1x1 local neural operator) model implementations, with full test coverage and regression baselines.
- Register both as `ModuleSelector` types (`"AnkurLocalNet"` and `"LocalNet"`), usable as either `builder` or `secondary_builder` in config YAML. `LocalNet` supports noise and label conditioning via `NoiseConditionedModule`.
- Extract shared `NoiseConditionedModule` base used by both `LocalNet` and `NoiseConditionedSFNO`, using composition (noise generator callback) rather than inheritance.
- Remove `HybridNet` which is superseded by the secondary network support in `SingleModuleStep`.

- [x] Tests added


fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,199 @@
...
+    mlp_ratio: float = 2.0
+    activation_function: str = "gelu"
+    encoder_layers: int = 1
+    pos_embed: bool = True
+    big_skip: bool = True
+    normalize_big_skip: bool = False
+    affine_norms: bool = False
+    lora_rank: int = 0
+    lora_alpha: float | None = None
+    data_grid: Literal["legendre-gauss", "equiangular"] = "equiangular"
```
mcgibbon: data_grid should be taken from the dataset_info passed to build, not given as a configuration flag.

fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,199 @@
...
+        data_grid: Grid type for spherical harmonic transforms used by
+            DISCO convolutions.
+    """
+
+    embed_dim: int = 256
+    use_disco_encoder: bool = False
+    disco_kernel_size: int = 3
+    pos_embed: bool = False
+    activation_function: str = "gelu"
+    data_grid: Literal["legendre-gauss", "equiangular"] = "equiangular"
```
mcgibbon: data_grid should be taken from dataset info on build.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -48,91 +52,53 @@ def isotropic_noise(
...
-                        embed_dim_pos,
-                        img_shape[0],
-                        img_shape[1],
-                        requires_grad=True,
-                    )
-                )
-                torch.nn.init.trunc_normal_(self.label_pos_embed, std=0.02)
-        else:
-            self.pos_embed = None
+def _make_sfno_noise_generator(
```
mcgibbon: We want to allow isotropic noise for the local net as well. Refactor so it can share this code.

fme/core/step/test_step.py
```diff
@@ -374,6 +374,41 @@ def get_fcn3_selector(
...
+                        "scale_factor": 1,
+                        "embed_dim": 4,
+                        "num_layers": 2,
+                    },
+                ),
+                in_names=["forcing_a", "prog_a", "prog_b"],
+                out_names=["prog_a", "prog_b"],
+                normalization=normalization,
+                secondary_builder=ModuleSelector(type="MLP", config={}),
+                secondary_out_names=["prog_a", "diag_a"],
```
mcgibbon: Oh, hm. I had meant for it to be that "out" names are full field outputs only, and the residual names exist only in that list. Please rename the second attribute to secondary_residual_out_names and give it that behavior.

fme/ace/registry/local_net.py
mcgibbon: Actually leaving this as-is, it's a pre-existing issue on NoiseConditionedSFNOBuilder. We should fix it in another PR.

fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        img_shape = dataset_info.img_shape
+        if self.noise_type == "isotropic":
+            from torch_harmonics import InverseRealSHT
+
+            grid_mapping = {
+                "legendre-gauss": "legendre-gauss",
+                "equiangular": "equiangular",
+            }
+            isht = InverseRealSHT(*img_shape, grid=grid_mapping[self.data_grid])
```
mcgibbon: Use the `dist` getter method for this rather than retrieving it directly.

fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,221 @@
...
+            out_chans=n_out_channels,
+            img_shape=dataset_info.img_shape,
+            data_grid=self.data_grid,
+            context_config=context_config,
+        )
+        img_shape = dataset_info.img_shape
+        if self.noise_type == "isotropic":
+            from torch_harmonics import InverseRealSHT
+
+            grid_mapping = {
```
mcgibbon: This is an identity mapping, right? I think we should remove it.

fme/ace/registry/noise_conditioned.py
```diff
@@ -0,0 +1,186 @@
...
+        noise_type: Type of noise to generate.
+        isht: Inverse SHT callable, required for isotropic noise.
+        lmax: Maximum degree, required for isotropic noise.
+        mmax: Maximum order, required for isotropic noise.
+    """
+    if noise_type == "gaussian":
+        return gaussian_noise
+    elif noise_type == "isotropic":
+        if isht is None:
+            raise ValueError("isht is required for isotropic noise")
```
mcgibbon: I don't really like the lack of type safety that leads to the need for this ValueError, is there a way to avoid it? For example, perhaps this helper should be deleted and we should just directly implement this chain one level higher, where we construct the isht?

fme/ace/registry/stochastic_sfno.py
```diff
@@ -19,120 +17,28 @@
...
-    alm = (real + 1j * imag) * scale
-
-    # --- for distributed iSHT, slice to local spectral extent --------------
-    l_slice, m_slice = Distributed.get_instance().get_local_slices((lmax, mmax))
-    alm = alm[..., l_slice, m_slice]
-
-    return isht(alm)
-
-
-class NoiseConditionedSFNO(torch.nn.Module):
```
mcgibbon: I'd like to reduce the diff of this change to make it easier to review. First of all, it's not critical that NoiseConditionedSFNO keep being called that as a class - this is never used in serialization. You can just rename the module to NoiseConditionedModel, and pass `inverse_sht` as an optional (can be none) argument during init. Then, we can have the local files import this symbol from this file. This would make it so any changes you've made show up as a diff, instead of seeing the entire module is deleted and added and not knowing what's a copy vs a change.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -48,21 +45,42 @@ def isotropic_noise(
...
     def __init__(
         self,
-        conditional_model: ConditionalSFNO,
+        module: torch.nn.Module,
         img_shape: tuple[int, int],
-        noise_type: Literal["isotropic", "gaussian"] = "gaussian",
         embed_dim_noise: int = 256,
         embed_dim_pos: int = 0,
         embed_dim_labels: int = 0,
+        inverse_sht: Callable[[torch.Tensor], torch.Tensor] | None = None,
```
mcgibbon, line 74: It's a little non-obvious that passing an inverse_sht would change the noise from being gaussian to being isotropic. Please add a docstring to this init (using consistent style for this repo) that describes this behavior in the arg docstring.

fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,211 @@
...
+        lora_rank: Rank of LoRA adaptations. 0 disables LoRA.
+        lora_alpha: Strength of LoRA adaptations. Defaults to lora_rank
+            if None.
+    """
+
+    embed_dim: int = 256
+    noise_embed_dim: int = 256
+    noise_type: Literal["gaussian", "isotropic"] = "gaussian"
+    context_pos_embed_dim: int = 0
+    block_types: list[BlockType] = dataclasses.field(
```
Arcomano1234, line 155: Question: Is this what we want for default? It seems a little odd to default to a rather deep (12) layer local model.

fme/ace/registry/stochastic_sfno.py
```diff
@@ -48,21 +45,58 @@ def isotropic_noise(
     return isht(alm)
 
 
-class NoiseConditionedSFNO(torch.nn.Module):
+class NoiseConditionedModel(torch.nn.Module):
```
Arcomano1234, line 48: Nit: I see why it makes sense to have 2 doc strings at the class level and at init but it does feel redundant. Is it possible to remove this documentation to init

fme/core/step/single_module.py
```diff
@@ -355,6 +433,23 @@ def network_call(input_norm: TensorDict) -> TensorDict:
...
+                if self.secondary_out_packer is None:
+                    raise RuntimeError("secondary_out_packer is unexpectedly None")
+                secondary_tensor = self.secondary_module.wrap_module(wrapper)(
+                    input_tensor,
+                    labels=args.labels,
+                )
+                secondary_dict = self.secondary_out_packer.unpack(
+                    secondary_tensor, axis=self.CHANNEL_DIM
+                )
+                for name in self._config.secondary_out_names:
```
Arcomano1234, line 446: Claude pointed this out as a potential bug,  when a `secondary_residual_out_name` is in `in_names` but not in `out_names`, and `residual_prediction=True`, the input gets added twice. I can't really envisioning how you could misconfig that/ if the code checks for this in other parts but I did want to point this out.

Arcomano1234 (commented):
I think this approach is better than with a `HybridNet`. Left a few questions / comments but this is looking good.

fme/ace/registry/local_net.py
mcgibbon: Agreed. I think we will want to tune this and I have no idea what it will look like in the end, but for now I think I want to start at 4 disco blocks followed by 4 conv1x1 blocks. This gives similar depth to what we have in the SFNO case, while biasing it towards more local analysis. I do think we should also test with as low as 2 blocks (and of course Ankur's case which only uses 1 DISCO layer) and compare. For what it's worth, this is analogous to the SFNO class where the default config uses 12 blocks but we always override it to 8.

Unfortunately to some degree this will be resolution-dependent as well, so we may have to tune both on 4 and 1-degree data.

fme/ace/registry/stochastic_sfno.py
mcgibbon: I meant to just have it on init (when I read the init I don't notice there's a class-level docstring), but sure let's remove the init copy and keep the class-level copy.

fme/core/step/single_module.py
mcgibbon: Yeah that's a good point. We should add more thorough testing of the configuration permutations and their corresponding programmatic input/output names property methods. For example, I think it should raise if a residual name is not present in any input names (while this is technically possible I don't think it's very scientifically valid).

From reading the code I don't actually think there's a bug in the way Claude pointed out there, but there should at least be a test to cover this.

mcgibbon: From Claude:

> For Arcomano1234's double-add concern (id=2985531730): The code is correct — when secondary_residual_out_name is in out_names, the secondary residual is added to the backbone's output (line 450). The residual_prediction input-adding happens separately in step_with_adjustments on prognostic names (line 544-545), and secondary_residual_out_names that are already in out_names get their backbone output added to the input there, then the secondary residual added on top — which is the intended behavior (backbone residual + secondary correction). I've added a test covering this case.

PR #1004, opened 2026-03-24
Add ERA5 training data statistical analysis scripts and findings

mcgibbon:
Investigate statistical properties of the ERA5 1-degree 8-layer dataset used for ACE2 pretraining, focusing on features that could explain why upper atmospheric variables are difficult to learn. Key findings include pathological distributions in stratospheric humidity, near-unpredictable stratospheric meridional wind, and normalization scale mismatches for upper atmosphere variables.


scripts/era5_investigation/f01_initial_findings.md
```diff
@@ -0,0 +1,193 @@
...
+
+**Implication**: This is likely a major contributor to why upper atmosphere variables are hard to learn. Switching to `residual_prediction: true` would directly address this by having the network predict the small delta rather than the large full field. Alternatively, level-specific or variable-specific normalization strategies could help.
+
+---
+
+## Finding 4: ERA5 data discontinuities affect stratospheric training data
+
+ERA5 has known discontinuities where parallel data streams were merged (documented in Hersbach 2020). The training config already excludes some problematic periods (1996-2010, 2020) but **includes 1979-1995**, which spans two significant discontinuities:
+
+| Year | Variable most affected | Jump magnitude |
```
mcgibbon, line 103: @Arcomano1234 these may indicate more periods we need to skip. However it sounds like these are year-over-year discontinuities, they might just be years with strong jumps due to volcanic events.

Should we have volcanic activity / aerosol as part of our forcings? Though Claude said they’re reanalysis artifacts below. Should investigate further.

scripts/era5_investigation/f01_initial_findings.md
```diff
@@ -0,0 +1,193 @@
...
+
+---
+
+## Finding 5: Radiation variables are bimodal and zero-inflated
+
+| Variable | % Exact Zero | % Negative | Residual Std / Full Std |
+| -------- | ------------ | ---------- | ----------------------- |
+| DSWRFsfc | 29.1%        | 0%         | 1.28                    |
+| USWRFtoa | 29.0%        | 0%         | 1.13                    |
+| USWRFsfc | 0.8%         | 23.1%      | 0.82                    |
```
mcgibbon, line 122: @spencerkclark any idea why upwelling shortwave surface radiation has so many small negative values, while these other similar variables do not?

scripts/era5_investigation/f01_initial_findings.md
mcgibbon: Got Claude to investigate, these are real discontinuities. Looks like the global map basically gets scaled.

scripts/era5_investigation/f01_initial_findings.md
spencerkclark: I think it is related to the fact that we compute the upward components as the difference between the downward and net fluxes, since they are not stored directly in ERA5. I cannot immediately explain why this does not show up in the upward shortwave flux at TOA in this older version of the dataset, but it is present in that variable too in the new dataset.

If we are concerned about it, we could consider clipping negative values at zero.

scripts/era5_investigation/f01_initial_findings.md
Arcomano1234: Thats really interesting I wonder if ECMWF is aware of this I want to dive a little more into these dates

scripts/era5_investigation/s07_spectral_power_periods.py
```diff
@@ -0,0 +1,296 @@
...
+from fme.sht_fix import RealSHT
+
+ZARR_PATH = "gs://vcm-ml-intermediate/2024-06-20-era5-1deg-8layer-1940-2022.zarr"
+PLOT_DIR = os.path.join(os.path.dirname(__file__), "plots")
+os.makedirs(PLOT_DIR, exist_ok=True)
+CACHE_DIR = "/tmp/era5_samples"
+os.makedirs(CACHE_DIR, exist_ok=True)
+
+# Variables to analyze — a mix of upper-atmosphere and surface
+VARIABLES = [
```
Arcomano1234, line 35: Can you add PRATEsfc?

scripts/era5_investigation/f01_initial_findings.md
mcgibbon: Clipping them may be worse, since then the budget wouldn't close. Not sure if we need to be concerned about it.

spencerkclark: Yeah, the deviations are exceedingly small, so I do not think it would make a material difference either way, considering the already large energy budget non-closure in ERA5, but I agree it is not worth worrying about.  It would just introduce unnecessary complexity.

PR #1008, opened 2026-03-25
Extract run_validation_loop and refactor Trainer to use run_validation

climate-ci-github:
## Summary
- Extract `run_validation_loop` as a minimal validation loop (eval, no_grad, EMA, iterate, record) without flush/WandB side effects, for reuse by callers that need only the core loop
- Refactor `run_validation` to delegate to `run_validation_loop` instead of having an inline loop
- Switch `Trainer.validate_one_epoch` from an inline validation loop to using `run_validation`
- Add `ema` and `validate_using_ema` parameters to `run_validation` so callers can pass EMA config through

## Test plan
- [x] New unit tests for `run_validation_loop` covering basic operation, EMA application, no-EMA case, and verifying no diagnostic flush
- [x] Existing `test_trainer.py` and `test_evaluator.py` tests pass unchanged

### DIFF
--- a/fme/core/generics/test_validation.py
+++ b/fme/core/generics/test_validation.py
@@ -1,0 +1,92 @@
+import unittest.mock
+
+import torch
+
+from fme.core.ema import EMATracker
+from fme.core.generics.test_trainer import TrainData, TrainStepper, ValidationAggregator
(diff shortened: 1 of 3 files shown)


mcgibbon:
Closing to open as author.

PR #1011, opened 2026-03-25
Extract run_validation_loop and refactor Trainer to use run_validation

mcgibbon:
## Summary
- Extract `run_validation_loop` as a minimal validation loop (eval, no_grad, EMA, iterate, record) without flush/WandB side effects, for reuse by callers that need only the core loop
- Refactor `run_validation` to delegate to `run_validation_loop` instead of having an inline loop
- Switch `Trainer.validate_one_epoch` from an inline validation loop to using `run_validation`
- Add `ema` and `validate_using_ema` parameters to `run_validation` so callers can pass EMA config through

## Test plan
- [x] New unit tests for `run_validation_loop` covering basic operation, EMA application, no-EMA case, and verifying no diagnostic flush
- [x] Existing `test_trainer.py` and `test_evaluator.py` tests pass unchanged

### DIFF
--- a/fme/core/distributed/torch_distributed.py
+++ b/fme/core/distributed/torch_distributed.py
@@ -32,7 +32,7 @@ class TorchDistributed(DistributedBackend):
                     torch.distributed.init_process_group(
                         backend="nccl",
                         init_method="env://",
-                        timeout=timedelta(minutes=20),
+                        timeout=timedelta(minutes=60),
                     )
                 else:
                     torch.distributed.init_process_group(


fme/core/generics/validation.py
```diff
@@ -16,6 +20,46 @@
...
+    Args:
+        stepper: The train stepper to evaluate.
+        valid_data: The validation dataset.
+        aggregator: The aggregator to record batch results into.
+        ema: The EMA tracker, or None if EMA is not used.
+        validate_using_ema: Whether to use EMA parameters during validation.
+        compute_derived_variables: Whether to compute derived variables.
+    """
+    stepper.set_eval()
+    ema_context: contextlib.AbstractContextManager = (
```
mcgibbon, line 49: I was thinking maybe the ema context entering should be pulled higher into the lr candidate code, since the inference doesn't interact with the ema context, but it's good to have this deduplicated between the trainer and lr candidate code. In a way this is making sure we apply that context correctly, and consistently between when we do the validation step and the lr tuning step during training.

jpdunc23: Agreed!

fme/core/generics/validation.py
```diff
@@ -16,6 +20,46 @@
...
+    )
+    no_opt = NullOptimization()
+    with torch.no_grad(), ema_context:
+        for batch in valid_data.loader:
+            stepped = stepper.train_on_batch(
+                batch,
+                optimization=no_opt,
+                compute_derived_variables=compute_derived_variables,
+            )
+            aggregator.record_batch(batch=stepped)
```
jpdunc23: Was it intentional to remove with timer.context("aggregator"):? IMO it would be good to retain the aggregator timing at least for inference jobs. I don't think we have validation aggregator timing metrics otherwise?

fme/core/generics/validation.py
```diff
@@ -57,18 +108,14 @@ def run_validation(
 
     logging.info("Starting validation loop")
 
-    no_opt = NullOptimization()
-    n_batches = len(validation_data.loader)
-
-    for i, batch in enumerate(validation_data.loader):
-        logging.info(f"Validation: processing batch {i + 1} of {n_batches}.")
```
jpdunc23, line 64: I think logging the progress of validation is helpful in validation-within-inference-eval since otherwise there is no feedback on the progress of the job during this stage. I don't think it would be harmful to also log during `validate_one_epoch` or even the "LR tuning trials", given that the number of validation batches are likely to be small relative to training batches. What do you think?

Otherwise `run_validation_loop` might benefit from having a boolean args to log or not.

fme/core/generics/test_validation.py
jpdunc23, line 1: Thanks for adding dedicated tests.

fme/core/generics/test_validation.py
```diff
@@ -0,0 +1,92 @@
+import unittest.mock
+
+import torch
+
+from fme.core.ema import EMATracker
+from fme.core.generics.test_trainer import TrainData, TrainStepper, ValidationAggregator
+from fme.core.generics.validation import run_validation_loop
+
+
+def test_run_validation_loop_returns_none():
```
jpdunc23: Nit: Maybe just `test_run_validation_loop`?

fme/core/generics/test_validation.py
```diff
@@ -0,0 +1,92 @@
...
+        validate_using_ema=True,
+    )
+
+    # After run_validation_loop, the original weights should be restored
+    assert torch.allclose(stepper.modules[0].weight.data, weight_before)
+    logs = aggregator.get_logs(label="val")
+    assert "val/mean/loss" in logs
+
+
+def test_run_validation_loop_without_ema_none():
```
jpdunc23: Nit: Maybe just `test_run_validation_loop_without_ema`? Or `test_run_validation_loop_without_ema_returns_none` if you want to keep `test_run_validation_loop_returns_none`

fme/core/generics/test_validation.py
```diff
@@ -0,0 +1,92 @@
...
+
+    run_validation_loop(
+        stepper=stepper,
+        valid_data=valid_data,
+        aggregator=aggregator,
+        ema=None,
+        validate_using_ema=False,
+    )
+
+    aggregator.flush_diagnostics.assert_not_called()
```
jpdunc23: This is sort of an odd test since it verifies behavior that `run_validation_loop` doesn't do. I guess the goal is to guard against any changes that would move flush_diagnostics from `run_validation` to `run_validation_loop`, but I don't see that as a huge risk. The `run_validation_loop` docstring is probably sufficient to guard against that change.

fme/core/generics/validation.py
```diff
@@ -16,6 +20,46 @@
...
+    valid_data: GriddedDataABC,
+    aggregator: AggregatorABC,
+    ema: EMATracker | None = None,
+    validate_using_ema: bool = False,
+    compute_derived_variables: bool = True,
+) -> None:
+    """Run the core validation loop: iterate batches and record to aggregator.
+
+    This is the minimal validation loop used by both `run_validation` and
+    LR tuning trials. It does NOT call `aggregator.get_logs`,
```
jpdunc23: Nit: Since https://github.com/ai2cm/ace/pull/930 will come later maybe remove "and LR tuning trials" for the time being.

fme/core/generics/test_validation.py
mcgibbon: Good idea.

fme/core/generics/test_validation.py
mcgibbon: I like your first suggestion.

fme/core/generics/test_validation.py
mcgibbon: Agreed, I don't think we need to enforce this with a unit test. This test can get deleted.

fme/core/generics/validation.py
```diff
@@ -16,6 +20,46 @@
...
+    `aggregator.flush_diagnostics`, or log to WandB — callers are responsible
+    for those.
```
mcgibbon: I think we should avoid referencing what calls this at all, since that's subject to change.

```suggestion
    This is the minimal validation loop. It does NOT call `aggregator.get_logs`,
    `aggregator.flush_diagnostics`, or log to WandB — callers are responsible
    for those.
```

fme/core/generics/validation.py
mcgibbon: Not intentional, let's update this so that it gives the same timing information as before. Part of this I think got lost in the parallel implementations with your change, and then the merge to main requiring a large refactor, which was then pulled out to this PR.

fme/core/generics/validation.py
mcgibbon: It seems a bit excessive to me to log it on every single batch, maybe we can log (batches {i + 1} through {min(n_batches, i+10)}) every 10 batches instead? Or even each 100 batches like we do for training? I think I'd require this if we're going to log this at all during training.

For right now, let's restore the previous behavior using a boolean arg on whether to log, and only using it in validation-within-inference-eval.

jpdunc23: Fair enough. Logging every 100 batches sounds good.

PR #1017, opened 2026-03-26
Register AnkurLocalNet and LocalNet in the module registry

mcgibbon:
Register the AnkurLocalNet and LocalNet model architectures in the ACE module registry so they can be selected via YAML configuration, and generalize `NoiseConditionedSFNO` into a reusable `NoiseConditionedModel` wrapper.

Changes:
- `fme.ace.registry.local_net` new module with `AnkurLocalNetBuilder` and `LocalNetBuilder` registry configs
- `fme.ace.registry.stochastic_sfno.NoiseConditionedSFNO` renamed to `NoiseConditionedModel`, decoupled from SFNO internals by accepting `inverse_sht` callable instead of reaching into model attributes; backward-compatible alias kept
- `fme.core.models.conditional_sfno.localnet.LocalNetConfig` add `kernel_shape` and `basis_type` config fields (previously hardcoded)
- `fme.core.registry.module.CONDITIONAL_BUILDERS` add `LocalNet`
- Default `LocalNetBuilder` config updated to better match latest prod settings

- [x] Tests added


fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,236 @@
...
+            if None.
+    """
+
+    embed_dim: int = 256
+    kernel_shape: tuple[int, int] = (3, 3)
+    basis_type: BasisType = "morlet"
+    noise_embed_dim: int = 256
+    noise_type: Literal["gaussian", "isotropic"] = "gaussian"
+    context_pos_embed_dim: int = 0
+    block_types: list[BlockType] = dataclasses.field(
```
Arcomano1234, line 155: Nit: Some of the defaults in `LocalNetBuilder` are different than the defaults `LocalNetConfig`. I didn't look at all of them but I noticed `block_types`, `normalize_big_skip `, and `affine_norms ` had different default values.

fme/ace/registry/local_net.py
```diff
@@ -0,0 +1,236 @@
...
+        use_disco_encoder: Whether to use a DISCO convolution for the first
+            layer instead of a 1x1 convolution.
+        disco_kernel_size: Kernel size for the DISCO convolution if used.
+        pos_embed: Whether to add a learned positional embedding after the
+            first layer.
+        activation_function: Activation function name ('relu', 'gelu', 'silu').
+    """
+
+    embed_dim: int = 256
+    use_disco_encoder: bool = True
```
Arcomano1234, line 68: Nit: I believe this default is different than what is in the `AnkurLocalNetConfig`

fme/ace/registry/stochastic_sfno.py
```diff
@@ -48,21 +45,42 @@ def isotropic_noise(
...
         embed_dim_noise: int = 256,
         embed_dim_pos: int = 0,
         embed_dim_labels: int = 0,
+        inverse_sht: Callable[[torch.Tensor], torch.Tensor] | None = None,
+        lmax: int = 0,
+        mmax: int = 0,
     ):
         super().__init__()
-        self.conditional_model = conditional_model
+        self.conditional_model = module
```
Arcomano1234: Question: I assume this is for backwards compatibility but this could be a source of confusion if some one sees the module in the docs while its assigned conditional_model. Maybe make a note of it somewhere

fme/ace/registry/local_net.py
mcgibbon: Ah yes we should update the AnkurLocalNetConfig. I forgot.

fme/ace/registry/local_net.py
mcgibbon: Similarly we should update the Config. I updated the values here but forgot those.

fme/ace/registry/stochastic_sfno.py
mcgibbon: We should revert the change to the argument name. The model being wrapped is still a conditional model (one that accepts the context for conditioning).

Arcomano1234: Agreed

PR #1018, opened 2026-03-26
feature/vector_filter_basis

mcgibbon:
Opening this draft PR to discuss with Claude.

Changes:

- [ ] Tests added


vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
+# DISCO Convolution with Vector-Typed Hidden Features
+
+## Overview
+
+This document specifies an approach for handling directional (vector-valued) data in DISCO convolution on the sphere. Rather than constructing a continuous directional filter basis (as in the cube-gauged approach), this design uses **isotropic radial filters** combined with a **frame-rotation matrix** that correctly transforms vector inputs from each input point's local meridian frame into the output point's meridian frame.
```
mcgibbon: The filters the design uses aren't isotropic, right? They will have an angular dependence. Or are they rotationally invariant, just they properly handle vector inputs and outputs?

I think it is a good idea that the physics be rotationally invariant, in the sense that for example if there is a convergence term for advection in x depending on u and the x-oriented filter, there must also be a corresponding convergence term in y depending on v and the y-oriented filter. Similarly for rotation-like filters. Is there a way to capture this? Describe how such filters would be represented and how it results in this desired behavior. Would it mean we store one set of filters that gets applied in some coherent way to both the x- and y- axis, using a rotation operator?

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+
+The frame-rotation angle between two points depends only on `(lat_out, lat_in, lon_in - lon_out)`, making it invariant to translations along longitude. This means the rotation can be baked into the filter tensor without breaking the FFT-based cross-convolution optimization.
+
+## Hidden Representation
+
+The network maintains three types of hidden feature channels at each grid point:
+
+- **Scalar channels** `s(lat, lon)`: Frame-independent quantities (temperature, pressure, geopotential, etc.). No special treatment needed in convolution.
+- **u channels** `u(lat, lon)`: East component of a vector in the local meridian frame (eastward wind, etc.).
+- **v channels** `v(lat, lon)`: North component of a vector in the local meridian frame (northward wind, etc.).
```
mcgibbon: Note the number of u and v channels are the same, but the number of scalar channels is independent of the number of vector channels. The vectors should be stored as a single tensor with a final [2] dimension.

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+Before the filter integrates over input points, each input vector is rotated from the input point's meridian frame into the output point's meridian frame. This rotation is characterized by a single angle `γ(lat_out, lat_in, Δlon)` — the angle between the two points' north directions, as measured via parallel transport along the connecting geodesic.
+
+The rotation acts as:
+
+```
+[u_rotated]   [cos γ  −sin γ] [u_in]
+[v_rotated] = [sin γ   cos γ] [v_in]
+```
+
+where `(u_in, v_in)` are the input vector components in the input's meridian frame, and `(u_rotated, v_rotated)` are the same vector expressed in the output's meridian frame.
```
mcgibbon: Note that this rotation is represented by being stored within the pre-computed filter itself.

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+- **Vector norm → scalar:** `n = sqrt(u² + v²)` is frame-invariant and produces a true scalar.
+- **Norm-gated vectors:** `σ(n) · (u, v) / n` applies a nonlinearity to the vector magnitude while preserving direction.
+- **Scalar × vector → vector:** `s · (u, v)` multiplies a scalar channel with a vector pair to produce a new vector pair. This is how the network creates new vector features from scalar-vector interactions.
+- **Vector dot product → scalar:** `u₁·u₂ + v₁·v₂` produces a frame-invariant scalar from two vector pairs.
+- **2D cross product → scalar:** `u₁·v₂ − u₂·v₁` produces a pseudo-scalar (changes sign under reflection but is frame-rotation invariant).
+
+### Unsafe Nonlinearities
+
+- **Independent nonlinearities on u and v:** Applying `ReLU(u)`, `ReLU(v)` independently breaks rotational structure — the result depends on the frame orientation.
+- **Adding scalar to vector component:** `u + s` breaks vector transformation rules.
```
mcgibbon: Let's only allow nonlinearities to be applied to the scalars. Don't apply nonlinearities to the vector channels. This should still be fine, as the vectors can have nonlinear behavior by virtue of their dependence on the scalar channels, right?

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+
+Two additional banded FFT tensors (`psi_cos_fft`, `psi_sin_fft`) of the same shape as `psi_scalar_fft` are stored per layer. Total filter storage is 3× the scalar-only case. The extra memory is typically small relative to activations and weights.
+
+### Comparison with Cube-Gauged Directional Filters
+
+The cube-gauged approach (documented in `cube_filter_basis.md`) adds angular modes `sin(mφ)`, `cos(mφ)` to the filter basis, giving `1 + 2M` basis functions per radial scale. The post-convolution rotation to the cube frame is cheap (pointwise), and the cube construction provides a globally smooth (up to Z₄ vertex singularities) directional filter.
+
+This approach is different in character:
+
+- **No angular filter modes.** The filters remain isotropic (radial only). Directional information enters through the vector channels, not through the filter shape.
```
mcgibbon: Let's discuss this some more. The filters won't be purely isotropic, right? We need to take d/dx of scalar fields, it's just that this operation will be handled in a way that aligns it with the u component, while d/dy would be aligned with the v component of the vectors.

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+- **Simpler but less expressive filters.** Isotropic filters cannot detect angular structure in scalar fields. Directional sensitivity comes only from vector inputs. This is sufficient if the physically important directional information is carried by vector quantities (wind, currents, etc.), which is typically the case in atmospheric modeling.
+
+## Design Parameters
+
+- **K (radial basis size):** Number of radial filter basis functions per convolution layer. Same as the current `kernel_shape` for `IsotropicMorletFilterBasis`. Controls the radial resolution of the filter.
+- **N_s, N_v (channel counts):** Number of scalar and vector channel pairs in the hidden representation. The network architecture determines these. A reasonable starting point is to match the current total channel count, reserving some for vector pairs.
+- **Nonlinearity design:** The choice of how to combine scalar and vector channels between layers (norm-gating, scalar-vector products, etc.) is an architectural decision. Simple norm-gating plus scalar×vector products provide the essential interactions.
+
+## Summary of Implementation Steps
```
mcgibbon: Can we also outline the testing plan for this new feature? Especially testing that various properties of certain operations are satisfied using user-story level tests of that behavior. It is hard to fully understand how the mathematical details lead to these properties, so it would help to have this also proven out by tests.

vector_filter_basis.md
```diff
@@ -0,0 +1,237 @@
...
+
+All intermediate features are in the output point's meridian frame, so they can be freely mixed by the learned weight tensor. The weight produces output scalar, u, and v channels:
+
+- **Scalar output from scalar input:** Standard — weight contracts over `(c_in, k)`.
+- **Scalar output from vector input:** Weight contracts over `(c_in, k)` separately for `conv_u` and `conv_v`, producing a scalar from vector components (analogous to divergence or a directional projection).
+- **Vector output from vector input:** Weight contracts over `(c_in, k)` for `conv_u` and `conv_v` to produce output `u` and `v` components. The weight can independently scale and mix the u/v components.
+- **Vector output from scalar input:** The weight assigns a scalar intermediate feature to a vector output channel. This implicitly creates a preferred direction in the meridian frame, and is therefore not rotationally equivariant. But if `u_out` and `v_out` channels are created from the same scalar feature with appropriate weights, the network can learn latitude-dependent directional structures (e.g., patterns aligned with Coriolis deflection).
```
mcgibbon: How is this weight defined? It should correspond to rotationally-invariant (but not isotropic) physics.

mcgibbon: The vectors should not be assigned values directly from scalar inputs, rather they should only be assigned values that correspond to specific filters (which may have an orientation or be isotropic) being applied to the scalars, and then that value being mapped into the vectors. I think this is what you mean, but it's not entirely clear. Requiring this mapping be rotationally invariant means that the isotropic filter would be assigned into the u and v components with exactly the same contribution - there's no directional dependence.

vector_filter_basis.md
mcgibbon: Claude: You're right — the effective filters are not isotropic. The radial basis functions ψ_k(r) are isotropic (depend only on geodesic distance), but the full filter tensors include angular dependence through cos(γ) and sin(γ), where γ is the frame rotation angle. So the filter tensor `ψ_k(r)·cos(γ)` has a dipole-like angular pattern — it picks out the component of the input aligned with one direction, while `ψ_k(r)·sin(γ)` picks out the perpendicular component.

The rotational invariance comes from how these filters are used in matched pairs. Yes — we store one set of radial basis functions ψ_k(r), and from it derive two filter tensors (psi_cos and psi_sin) that are always applied together via the rotation matrix structure. For example:

- **Scalar→vector (gradient-like):** Applying psi_cos to a scalar field s gives the u-component, and psi_sin gives the v-component. Because both use the same ψ_k(r) and the angular factors cos(γ)/sin(γ) are geometrically determined (not learned), this automatically ensures that d/dx and d/dy are locked together — you can't have convergence in x without the corresponding convergence in y.

- **Vector→vector (frame rotation):** The same psi_cos/psi_sin pair forms the 2×2 rotation matrix [cos γ, −sin γ; sin γ, cos γ] that transforms input vectors into the output frame. Again, one set of filters handles both components coherently.

- **Vector→scalar (divergence/curl-like):** The cos/sin filters contract vector components into rotationally invariant scalars (div-like: cos·u + sin·v, curl-like: cos·v − sin·u).

The learned weights are shared between the u and v pathways — the weight can't independently scale the x vs y contribution, so the network can't break the rotational symmetry. I'll revise the document to clarify this and stop calling the filters "isotropic."

vector_filter_basis.md
mcgibbon: Claude: Agreed. Forward signature will be `forward(x_scalar, x_vector)` where `x_scalar: (B, N_s, H, W)` and `x_vector: (B, N_v, H, W, 2)` with the last dimension being (u, v). Output follows the same convention. Will update the document.

vector_filter_basis.md
mcgibbon: Claude: Yes — cos(γ) and sin(γ) are multiplied into ψ_k(r) during precomputation to produce `psi_cos_fft` and `psi_sin_fft`. No runtime rotation needed.

vector_filter_basis.md
mcgibbon: Claude: Correct. Vectors gain nonlinear behavior through the scalar pathway: scalars get nonlinearities applied, then the next layer's scalar→vector filters (psi_cos/psi_sin) produce new vector features that depend nonlinearly on the original inputs. This avoids frame-dependent nonlinearity issues entirely. I'll replace the nonlinearity section with a note about this constraint.

vector_filter_basis.md
mcgibbon: Claude: Right — as discussed in the thread on line 5, the effective filters have angular dependence via cos(γ)/sin(γ). For scalar→vector specifically: `psi_cos` applied to a scalar gives the d/dx-like component (→ u), `psi_sin` gives d/dy-like (→ v). Same ψ_k(r) underlies both, geometrically locked together. I'll replace "isotropic" throughout.

vector_filter_basis.md
mcgibbon: Claude: The weight has separate blocks per type interaction, with u/v sharing weights to enforce rotational invariance. Each interaction that involves vectors has two independent weight components (analogous to a complex weight: one "real" and one "imaginary" or equivalently one "stretch" and one "90°-rotation"):

- **W_ss** `(N_s_out, N_s_in, K)`: scalar→scalar via psi_scalar. Standard, 1 weight.
- **W_sv** `(N_v_out, N_s_in, K, 2)`: scalar→vector. Component 0 ("gradient"): u = w·contraction(psi_cos, s), v = w·contraction(psi_sin, s). Component 1 ("perp gradient"): u = w·(-contraction(psi_sin, s)), v = w·contraction(psi_cos, s). The perpendicular gradient is physically meaningful — e.g. geostrophic wind from pressure.
- **W_vs** `(N_s_out, N_v_in, K, 2)`: vector→scalar. Component 0 ("divergence"): s = w·(psi_cos·u + psi_sin·v). Component 1 ("curl"): s = w·(psi_cos·v − psi_sin·u).
- **W_vv** `(N_v_out, N_v_in, K, 2)`: vector→vector. Component 0 ("stretch"): same weight on both components of frame-rotated vector. Component 1 ("rotate"): applies 90° rotation (-conv_v, conv_u) with shared weight.

Key constraint: within each block, u and v pathways share the same scalar weight, so the network can't break rotational symmetry. The 2-component structure for vector interactions captures the full space of rotationally-invariant linear maps (divergence+curl, gradient+perp-gradient, stretch+rotation).

mcgibbon: Claude: Agreed. In the updated design, scalar→vector exclusively uses the oriented filters: `u = W_sv · contraction(psi_cos, s)`, `v = W_sv · contraction(psi_sin, s)`. The directional structure comes from the geometry (cos γ, sin γ), not the weights. There is no path from psi_scalar to vector outputs — an isotropic filter applied to a scalar carries no directional information and can't meaningfully populate a vector channel.

vector_filter_basis.md
mcgibbon: Claude: Proposed testing plan (user-story-level behavioral tests):

1. **Uniform vector field is preserved.** Create a constant vector field (e.g., everywhere pointing east). After vector→vector convolution, the output should still be uniform and pointing east — frame rotations cancel for a constant field.

2. **Longitude shift equivariance.** Shift scalar and vector inputs by N longitude points. Verify outputs shift by N points (both scalar and vector channels). Tests that the FFT-based implementation preserves translational symmetry.

3. **Scalar gradient produces correct vector.** Apply scalar→vector to a scalar field with a known gradient (e.g., linear ramp in longitude → expect u output, zero v). Verify the output vector direction matches the gradient direction.

4. **Divergence recovery.** Create a radially diverging vector field from a point. Apply vector→scalar with divergence-like weights. Verify the output scalar is positive at the source.

5. **Curl recovery.** Create a rotating vector field. Apply vector→scalar with curl-like weights. Verify the result reflects the rotation.

6. **Rotational invariance.** Rotate input fields (shift longitude), verify that scalar outputs are identical and vector outputs rotate consistently.

7. **Scalar-only consistency.** When N_v=0, verify the module produces identical results to the existing `DiscreteContinuousConvS2`.

8. **Regression baselines.** Save reference outputs from known inputs/weights, verify they don't change across refactors (using `validate_tensor_dict`).

mcgibbon: Your "longitude shift" and "rotational" tests seem to be the same test? Consider also doing a test that involves rotating the north pole somewhere else. The output field should be rotated the same way.

mcgibbon: Claude: You're right, those were redundant. I'll merge them and add a proper full-sphere rotation test. Updated plan in the document:

- **Longitude shift equivariance**: shift inputs by N longitude points, verify outputs shift identically.
- **Arbitrary rotation equivariance**: rotate the entire input field (move the north pole elsewhere), verify the output rotates consistently — scalar outputs match under the rotation, vector outputs transform correctly into the new frame.

The arbitrary rotation test is stronger because longitude shifts are already guaranteed by the FFT structure, while a pole-moving rotation exercises the frame rotation geometry at every latitude.

mcgibbon:
The shallow water model produces plausible videos of evolution. For no-rotation they look exactly like what I'd expect. For with-rotation, I have much less intuition on what the solution should look like, but the solution looks plausible.

PR #1020, opened 2026-03-26
Have losses return non-reduced tensors

Arcomano1234:
Part of a larger PR to allow for per channel loss to be logged. Here, we change how the loss is reduced, specifically having `StepLoss` return `LossOutput` instead of a torch.Tensor. `LossOutput` has several methods to allow the return of a total() which is the normal behavior of return mean over all of the dimensions or get_channel_losses() which allows for the per-channel loss in the follow PR. 
Changes:
- Add `LossOutput` to `loss.py`
- Have `StepLoss` return `LossOutput`

- [x] Tests added


fme/ace/stepper/single_module.py
```diff
@@ -1627,9 +1627,10 @@ def _accumulate_loss(
                     }
                 )
                 step_loss = self._loss_obj(gen_step, target_step, step=step)
-                metrics[f"loss_step_{step}"] = step_loss.detach()
+                step_total = step_loss.sum()
```
mcgibbon: nit: step total what?

```suggestion
                step_total_loss = step_loss.sum()
```

May want to rename step_loss to step_channel_loss if you do this, but also if you do the all reductions here (like we discussed earlier), having them here should make it clear.

fme/coupled/loss.py
```diff
@@ -127,7 +127,8 @@ def __call__(
         self, prediction: StepPredictionABC, target_data: TensorMapping
     ) -> torch.Tensor:
         if self.step_is_optimized(prediction.step):
-            return self._weight * self._loss(
-                prediction.data, target_data, prediction.step
+            return (
+                self._weight
+                * self._loss(prediction.data, target_data, prediction.step).sum()
```
mcgibbon: What we discussed was having the loss take in a `reduce` argument, and passing `reduce=False` for the ace case, which would avoid needing to do anything here (though coupled wouldn't get access to these improvements, either).

fme/ace/stepper/single_module.py
```diff
@@ -1626,10 +1626,13 @@ def _accumulate_loss(
                         for k, v in target_data.data.items()
                     }
                 )
-                step_loss = self._loss_obj(gen_step, target_step, step=step)
-                metrics[f"loss_step_{step}"] = step_loss.detach()
+                step_loss = self._loss_obj(
+                    gen_step, target_step, step=step, reduce=False
+                )
+                step_total_loss = step_loss.sum()
+                metrics[f"loss_step_{step}"] = step_total_loss.detach()
```
Arcomano1234, line 1637: These changes are temporary and will be moved to per-channel reductions with the other per-channel loss PR

fme/core/test_loss.py
```diff
@@ -384,21 +419,113 @@ def test_WeightedMappingLoss_with_target_nans():
...
+        channel_dim,
+        torch.nn.MSELoss()(x, y),
+    )
+
+
+def test_mse_per_channel_sum_5d():
+    torch.manual_seed(42)
+    x = torch.randn(4, 2, 5, 8, 8, device=get_device())
+    y = torch.randn(4, 2, 5, 8, 8, device=get_device())
+    _assert_per_channel_sum_matches(
```
mcgibbon, line 461: TL;DR: I think we should use a container class for the returned loss, with `.total()` and `.get_channel_losses()` methods.

Before I left we talked about updating this so that the loss returns a non-reduced tensor (not even reduced to 1D channel, but rather keeping all the dimensions, except for _maybe_ the batch dimension if that). Then the aggregator receives the non-summed losses. That gives us the option down the line to do neat things like plot the worst example in an aggregator (which I could see doing with Claude in a one-off experiment, for example).

Did you run into an issue that makes this difficult/not feasible?

We could maybe just merge with this route so we have the feature available to us, but I do feel like it hard codes a fair bit of a specific present low-level concern (we need the aggregator to produce per-channel losses) instead of implementing the more general feature (we want an aggregator that can analyze higher-order features of the loss, and not just its mean).

The current implementation (and wide array of tests ensuring the returned losses are per-channel) also may make it a bit harder to support non-channel losses (like gan losses, or regularizer losses) in the future, where I could see saying the returned loss may be scalar or N-dimensional, and only reporting per-channel losses if the loss is higher-dimensional. This gets even more complicated if we have multiple differently-dimensioned losses (like the global-mean loss or a gan loss).

What this is kind of pointing me to thinking is that we should maybe have a LossValue or LossOutput type with a `.total()` method and any necessary methods for the aggregator (like `.get_channel_losses()`), which is what gets passed to this train aggregator. If you had that, it would kind of be fine for the current LossOutput to always be a 1D channel-varying tensor on the inside, because changing this in the future is isolated to changing just that class, instead of manually searching for a bunch of low-level tensor operations.

fme/core/loss.py
```diff
@@ -13,6 +13,70 @@
...
+
+    Wraps the raw unreduced loss tensor and provides convenience methods
+    for obtaining the scalar total or per-channel breakdowns. This
+    isolates reduction logic so callers and aggregators don't need to
+    know the internal tensor layout.
+    """
+
+    _loss: torch.Tensor
+    _channel_dim: int
+    _channel_names: list[str]
```
mcgibbon: Comment: We don't usually use private dataclass attributes, however I do think these should be private.

Question: Does this need to be a dataclass? I don't think we'll serialize/deserialize it.

Suggestion (optional): Make this a normal class, not a dataclass, and set the attributes privately on init.

fme/core/test_loss.py
```diff
@@ -68,15 +69,14 @@ def test_loss_of_zeros_is_variance():
     x = torch.zeros(10, 10, 10, 10, 10, device=get_device())
     y = torch.randn(10, 10, 10, 10, 10, device=get_device())
     result = loss(x, y)
-    assert result.shape == ()
     assert isinstance(result, torch.Tensor)
```
mcgibbon, line 72: Question: I would have expected the result would be a LossOutput type, not a tensor. Is that because only WeightedMappingLoss returns LossOutput? I guess that's fine if so, but maybe we should make the non-weighted-mapping losses private in another PR (I would guess we'll never get around to it, which is also fine).

mcgibbon (commented):
Mostly looks good to me, only comment is to use a non-dataclass class for less awkward control of private state initialization.

fme/core/loss.py
Arcomano1234: Thanks this is a good suggestion. I don't think this needs to be a dataclass.

fme/core/test_loss.py
Arcomano1234: Yes right now only WeightedMappingLoss returns LossOutput. I agree its a little awkward let me think about this for a minute to see if there is a nice solution (either make the other losses private in another PR or maybe we can have LossOutput for all losses with out changing too many things)

mcgibbon (approved):
LGTM

PR #1021, opened 2026-03-27
Build torch-harmonics with CUDA kernels in deps-only image

mcgibbon:
The default pip install of torch-harmonics does not compile custom CUDA extensions because the runtime Docker base image lacks nvcc. This means the optimized CUDA kernels for spherical harmonic transforms and DISCO convolutions are not available at runtime, falling back to slower pure-PyTorch implementations.

Changes:
- `docker/install_torch_harmonics_cuda.sh`: new install script (following the Apex pattern) that rebuilds torch-harmonics with `FORCE_CUDA_EXTENSION=1`. Targets T4 (sm_75), A100 (sm_80), L40 (sm_89), H100 (sm_90), and B200 (sm_100) architectures.
- `docker/install_cuda_extensions.sh`: new wrapper script that installs the CUDA toolkit once, runs both the torch-harmonics and Apex GroupNorm build scripts, then removes the toolkit — avoiding a redundant install/remove cycle.
- `docker/install_apex_groupnorm.sh`: removed CUDA toolkit install/removal (now handled by the wrapper).
- `docker/Dockerfile`: replaced separate `RUN` steps for torch-harmonics and Apex with a single step using the wrapper script.

- [ ] Tests added


docker/install_torch_harmonics_cuda.sh
```diff
@@ -0,0 +1,34 @@
...
+# Step 1: Install CUDA toolkit via conda
+echo ""
+echo "[1/3] Installing CUDA toolkit (nvcc) via conda..."
+conda install -y -c nvidia cuda-nvcc=12.8 cuda-toolkit=12.8
+
+# Step 2: Reinstall torch-harmonics with CUDA extensions
+echo ""
+echo "[2/3] Building torch-harmonics with CUDA extensions..."
+FORCE_CUDA_EXTENSION=1 \
+    TORCH_CUDA_ARCH_LIST="7.5 9.0 10.0+PTX" \
```
mcgibbon: Corresponds to T4, H100, and B200's - would be good to document this in a comment.

docker/Dockerfile
```diff
@@ -52,6 +52,9 @@ RUN apt-get update && apt-get install -y \
 RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq &&\
 chmod +x /usr/bin/yq
 
+COPY docker/install_torch_harmonics_cuda.sh /tmp/install_torch_harmonics_cuda.sh
+RUN chmod +x /tmp/install_torch_harmonics_cuda.sh && /tmp/install_torch_harmonics_cuda.sh && rm /tmp/install_torch_harmonics_cuda.sh
```
mcgibbon: Comment for other reviewers: Note this was put in the same place as apex group norm, which means it's not installed in the "production" image (not sure what that's used for).

mcgibbon:
Note build time for torch harmonics was 852s on my CPU VM, compared with 1631s for apex group norm. If I use the full list of target architectures, this goes up to more like 1600s.

docker/install_torch_harmonics_cuda.sh
```diff
@@ -0,0 +1,34 @@
...
+# CUDA extensions enabled, then removes the toolkit to save space.
+
+set -e
+
+echo "=== Reinstalling torch-harmonics with CUDA extensions ==="
+
+# Step 1: Install CUDA toolkit via conda
+echo ""
+echo "[1/3] Installing CUDA toolkit (nvcc) via conda..."
+conda install -y -c nvidia cuda-nvcc=12.8 cuda-toolkit=12.8
```
AnnaKwa: Do you know how long this step takes? I don't recall how much of the apex install was the toolkit, but it was definitely at least a few Gb in size. If it is long- should the temporary install and removal take place before and after the apex and torch harmonics scripts are run so it doesn't have to occur twice?

mcgibbon: I don't know how long it takes, but this seems like a reasonable optimization to make regardless.

PR #1022, opened 2026-03-27
Add AIMIP evaluation IC and modified forcing workflow steps

brianhenn:
Two workflow steps for creating ACE AIMIP inference datasets (separate from the training and eval, which used the previous ACE2-ERA5 dataset):
- The IC script creates a 5-member lagged ensemble by relabeling atmospheric state snapshots to a common start timestamp (`1978-09-30T18:00:00`), chosen so the first model prediction step lands at `1978-10-01T00:00:00`.
- The forcing script repeats the first-timestep forcing data from `1978-10-01T00:00:00` at `1978-09-30T18:00:00` so that the first predicted step is also `1978-10-01T00:00:00`. This is a bit of a hack/oversight needed to get ACE's first full predicted day to be 0, 6, 12, 18Z on 1978-10-01 for daily data output purposes, since the data writer can't actually coarsen by day but only by integer factors, so having the first predicted step be 6Z on 1978-10-01 doesn't work for AIMIP. For AIMIP, we could have chosen to make the forcing data to start on 1978-09-30T18:00:00 but we didn't do that -- not sure how other groups handled this wrinkle.

Changes:                                                                                                                                                                                                                                          
  - `scripts/aimip_forcing/create_aimip_ic_datasets.py`: new CLI script to generate 5 AIMIP evaluation IC NetCDF files from ERA5 GCS data
  - `scripts/aimip_forcing/prepend_first_timestep_forcing.py`: new CLI script to prepend a `1978-09-30T18:00:00` timestep to the AIMIP forcing zarr


scripts/aimip_forcing/prepend_first_timestep_forcing.py
```diff
@@ -0,0 +1,84 @@
+import logging
+
+import click
+import numpy as np
+import xarray as xr
+from dask.diagnostics import ProgressBar
+from encoding import clear_encoding, set_shards_chunks
```
Arcomano1234, line 7: I know I've used similar functions before it would be nice to eventually get a lot of these toolings into a common place somewhere (maybe `scripts/ace-tools` or `scripts/common`). But thats outside of this PR.

brianhenn: Will open an issue; I've also been unclear on zarr v3 best practices.

brianhenn: https://github.com/ai2cm/ace/issues/1029

PR #1023, opened 2026-03-27
Add scripts/aimip_postprocessing for CMIP6-compliant AIMIP result processing

brianhenn:
Notebook from explore2 [2025-11-26-process-aimip-inference-results.ipynb](https://github.com/ai2cm/explore2/blob/main/brianh/2025-11-26-aimip-inference-results/2025-11-26-process-aimip-inference-results.ipynb) contains a one-off workflow for post-processing ACE2.1-ERA5 AIMIP inference results into CMIP6-compliant netCDF files. This PR converts it to a script alongside the related `scripts/aimip_forcing/` pipeline. 

To some extent, this PR is coupled to the configurations used to run inference on ACE2.1-ERA5 and specify the names of the data writer files, which is being merged in [a separate PR](https://github.com/ai2cm/ace/pull/1027). Maybe there's useful added functionality to handle these "CMIP-style" runs automatically in ACE without the need for a postprocessing script, but leaving that for later.

Changes:
- `scripts/aimip_postprocessing/postprocess.py` — script with typed `SimulationConfig`/`FileConfig` dataclasses, YAML-backed defaults, and all CF-convention transformation helpers from the notebook
- `scripts/aimip_postprocessing/simulations.yaml` / `files.yaml` — default configuration lists (15 simulations, 48 variable/table/grid entries)
- `scripts/aimip_postprocessing/test_postprocess.py` — unit tests for key helpers
- `scripts/aimip_postprocessing/README.md` / `Makefile` — usage docs and convenience targets

- [x] Tests added


scripts/aimip_postprocessing/postprocess.py
```diff
@@ -0,0 +1,594 @@
...
+        "standard_name": "time",
+        "long_name": "time",
+        "bounds": "time_bnds",
+        "axis": "T",
+    }
+    ds[time_name] = ds[time_name].assign_attrs(time_attrs)
+    return ds
+
+
+def stack_vertical_dimension(
```
elynnwu, line 206: Feels like we have some version of the same function in many places (e.g., [this](https://github.com/ai2cm/ace-tools/blob/7255634ebf92614c61d2d309905f5255c8d15c6e/ace_tools/utils.py#L302) is used for some of the coupled analysis work), would it make sense to consolidate them? I think it's kind of the idea for that `ace-tools` repo though we didn't want to rigorously maintain it.

In fact, this applies to many functions in this file. Just wondering if want to consider unify them.

scripts/aimip_postprocessing/test_postprocess.py
```diff
@@ -0,0 +1,197 @@
...
+def test_stack_vertical_dimension_sorted_by_level():
+    ds = xr.Dataset(
+        {
+            "ta850": xr.DataArray(np.full((2,), 2.0), dims=["x"]),
+            "ta100": xr.DataArray(np.full((2,), 1.0), dims=["x"]),
+            "ta500": xr.DataArray(np.full((2,), 0.5), dims=["x"]),
+        }
+    )
+    result = stack_vertical_dimension(ds, "plev", "ta", r"[0-9]+$")
+    assert list(result["plev"].values) == [100.0, 500.0, 850.0]
```
elynnwu, line 42: Does it sort correctly if you have `100., 500., 850., 1000.`?

pyproject.toml
```diff
@@ -22,6 +22,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "cartopy>=0.22.0",
+    "cftime",
```
elynnwu, line 25: Should this be added here or `analysis-deps.txt`?

elynnwu:
Don't need to address making the common functions widely available in this PR, but I think it makes sense to have some sort of shared common ace analysis tools. Still not sure what's the best way.

scripts/aimip_postprocessing/postprocess.py
brianhenn: Agree that there's a lot of duplication overall, and if we make some of these postprocessing steps part of the the real ACE workflow we'd consolidate some of the code.

At least here, I'll look into whether using the existing `Stacker` might work.

pyproject.toml
brianhenn: While it's unlikely that `xarray` would ever drop `cftime` as a transitive dependency in our repo, we directly import `cftime` in a some places already in `ace` (eg [here](https://github.com/ai2cm/ace/blob/0894f63af9d4e9fd0b1c042306a859d06fd698ad/fme/ace/inference/data_writer/file_writer.py#L8)) because we like to write CF calendar time coordinates, so I figured it was probably worth adding explicitly. But probably not critical either way.

scripts/aimip_postprocessing/test_postprocess.py
brianhenn: I think it does, but can test, or use the stacker.

scripts/aimip_postprocessing/postprocess.py
brianhenn: It looks like the current `Stacker` hardcodes that variable names are `<name>_<number>` which doesn't work here because we lack the underscore. But this is good motivation to maybe try to refactor this code into the data writer and use a common set of tools in a future PR.

brianhenn:
Closing this in favor of a near-identical PR in the ACE2.1-ERA5-AIMIP repo: https://github.com/ai2cm/ACE2.1-ERA5-AIMIP/pull/1

PR #1025, opened 2026-03-30
`LossVsNoiseAggregator`

AnnaKwa:
It is useful to be able to see plots of loss vs training sample noise to check if the training noise distribution is appropriately set, see how changes in noise-weighted loss affect training, etc. This PR adds that option for downscaling wandb reports but makes it optional since generates a bunch of new 2d figures for each epoch.

- Adds class `LossVsNoiseAggregator`.
- Adds an option to log loss vs noise to wandb via `TrainerConfig.log_loss_vs_noise`. Defaults to False. If true, the new aggregator is used during training.
- adds attributes `ModelOutputs.sigma` (1d tensor of noise used in each training sample) and `ModelOutputs.per_sample_channel_loss` (tensor dict where keys are channel names and values are 1d tensors of loss per sample) so that this information can be passed to the aggregator.

This returns a wandb.Image since wandb's plotly sometimes chokes on plots with two axes and makes the bin counts horizontal bars. 

example output:
[wandb](https://wandb.ai/ai2cm/annak-downscaling/runs/bomshtxf)
image

### DIFF
--- a/scripts/data_process/configs/CM4-piControl-200yr-coupled.yaml
+++ b/scripts/data_process/configs/CM4-piControl-200yr-coupled.yaml
@@ -8,6 +8,9 @@ coupled_datasets:
   coupled_sea_surface:
     sst_threshold: 291.75
     ocean_extra_masked_names: ["HI", "sea_ice_volume", "UI", "VI", "sfdsi"]
+    ocean_extra_fill_values:
+      UI: 0.0
+      VI: 0.0
     surface_flux_window_avg:
       window_timedelta: 120h
       first_timestamp: "0151-01-03T12:00:00"
--- a/scripts/data_process/create_coupled_datasets.py
+++ b/scripts/data_process/create_coupled_datasets.py
@@ -74,7 +74,6 @@ def _get_stats(
     else:
         logging.info(f"Computing statistics")
         logging.info(f"  Output directory: {stats_dir}")
-        out_dir = stats_dir + "/" + run_name
         get_stats(
             config=StatsConfig(
                 output_directory=os.path.dirname(input_zarr_path),
@@ -83,7 +82,7 @@ def _get_stats(
                 end_date=end_date,
             ),
             input_zarr=input_zarr_path,
-            out_dir=out_dir,
+            out_dir=stats_dir,
             debug=False,
         )
     return stats_dir


fme/downscaling/aggregators/main.py
```diff
@@ -6,6 +6,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import torch
+import wandb
```
frodre: Instead of importing wandb and sidestepping the FME wrapper interaction, use `WandB.get_instance()` in the aggregator

fme/downscaling/aggregators/main.py
```diff
@@ -50,6 +51,165 @@ def _tensor_mapping_to_numpy(data: TensorMapping) -> TensorMapping:
...
+        self._channel_count[name].scatter_add_(
+            0, bin_indices, torch.ones_like(bin_indices, dtype=torch.int64)
+        )
+
+    @torch.no_grad()
+    def record_batch(self, outputs: ModelOutputs) -> None:
+        if outputs.sigma is None or not outputs.per_sample_channel_loss:
+            return
+
+        sigma = outputs.sigma.detach().flatten().cpu()
```
frodre: `detach()` should be covered by the no_grad() declaration.

fme/downscaling/aggregators/main.py
```diff
@@ -50,6 +51,165 @@ def _tensor_mapping_to_numpy(data: TensorMapping) -> TensorMapping:
...
+        lines, labels = ax.get_legend_handles_labels()
+        lines2, labels2 = ax2.get_legend_handles_labels()
+        ax.legend(lines + lines2, labels + labels2)
+
+        image = wandb.Image(fig)
+        plt.close(fig)
+        return image
+
+    def get_wandb(self, prefix: str = "") -> Mapping[str, Any]:
+        prefix = ensure_trailing_slash(prefix)
```
frodre: To simplify the amount of code, I've been implementing (mostly private where possible) `_get` methods which reduce and move the final values to CPU.  That way `get_wandb` directly handles the wandb data prep and handoff and there's less device handling in separate locations.

fme/downscaling/aggregators/main.py
```diff
@@ -50,6 +51,165 @@ def _tensor_mapping_to_numpy(data: TensorMapping) -> TensorMapping:
...
+            color="gray",
+            label="bin count",
+        )
+        ax2.set_ylabel("bin count")
+
+        lines, labels = ax.get_legend_handles_labels()
+        lines2, labels2 = ax2.get_legend_handles_labels()
+        ax.legend(lines + lines2, labels + labels2)
+
+        image = wandb.Image(fig)
```
frodre, line 168: e.g., `wandb = WandB.get_instance()` then `wandb.Image(fig)` follows the same standard path.

fme/downscaling/aggregators/main.py
```diff
@@ -50,6 +51,165 @@ def _tensor_mapping_to_numpy(data: TensorMapping) -> TensorMapping:
...
+
+        sigma = outputs.sigma.detach().flatten().cpu()
+        if torch.any(sigma <= 0):
+            raise ValueError("Sigma must be strictly positive for log10 binning")
+        log_sigma = torch.log10(sigma)
+        in_range = (log_sigma >= self._log10_sigma_min) & (
+            log_sigma <= self._log10_sigma_max
+        )
+        if not torch.any(in_range):
+            return
```
frodre, line 114: Could move this out into a local helper method `_maybe_get_in_range` that returns `in_range | None` to focus the logic under this method on batch recording.

fme/downscaling/aggregators/main.py
```diff
@@ -50,6 +51,165 @@ def _tensor_mapping_to_numpy(data: TensorMapping) -> TensorMapping:
...
+        lines2, labels2 = ax2.get_legend_handles_labels()
+        ax.legend(lines + lines2, labels + labels2)
+
+        image = wandb.Image(fig)
+        plt.close(fig)
+        return image
+
+    def get_wandb(self, prefix: str = "") -> Mapping[str, Any]:
+        prefix = ensure_trailing_slash(prefix)
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
frodre: Any reason to not use the `fme` function `get_device()`?

fme/downscaling/aggregators/main.py
frodre: I would consider leaving the aggregator values on device until `_get` is called to reduce the amount of moving back and forth.

frodre (approved):
Thanks.  Overall, LGTM.   The only change I think needs to happen is the usage of `WandB.get_instance` instead of importing `wandb`. I made suggestions to try and improve code readability, but feel free to ignore.

PR #1031, opened 2026-04-02
Add `SmoothFloodFill` to allow for global ocean spectra

jpdunc23:
Add the callable `fme.core.fill.SmoothFloodFill` for filling NaN regions (e.g. land in ocean fields) with smooth boundary transitions suitable for spectral analysis. The algorithm mean-fills deep interior pixels, iteratively blends edges inward via neighbor-averaging with circular longitude padding, and applies a separable Gaussian blur to smooth the valid/filled boundary. The first time `SmoothFloodFill` is called for a previously unseen variable the interior NaN mask is computed and cached for later calls. For fields with no NaN regions, `SmoothFloodFill` simply returns the original tensor. 

Changes:
- Adds a `nan_fill_fn` argument to `SphericalPowerSpectrumAggregator` and `PairedSphericalPowerSpectrumAggregator` with default no-op value. The `"power_spectrum"` aggregators now use `nan_fill_fn = SmoothFloodFill(num_steps=4)`.

- [x] Tests added

### DIFF
--- a/fme/core/fill.py
+++ b/fme/core/fill.py
@@ -1,11 +1,10 @@
-import dataclasses
+from typing import NamedTuple
 
 import torch
 import torch.nn.functional as F
 
 
-@dataclasses.dataclass
-class _Masks:
+class _Masks(NamedTuple):
     interior: torch.Tensor
     valid: torch.Tensor
     blurred_valid: torch.Tensor


jpdunc23:
See [these slides](https://docs.google.com/presentation/d/1MLiVisBcKD4E5_hAc188hyzzzy-VJlvXizuL2jxf7ck/edit?slide=id.g3dcf82e5540_0_6#slide=id.g3dcf82e5540_0_6) for more background.

fme/core/fill.py
```diff
@@ -0,0 +1,287 @@
...
+    neighbor-averaging that grows valid pixels inward from the boundary, and
+    (3) Gaussian blur interpolation across the original valid/NaN boundary to
+    avoid sharp seams.
+
+    Interior masks are computed once per variable name and cached.
+    """
+
+    def __init__(
+        self,
+        num_steps: int,
```
elynnwu: It looks like you always use 4, maybe just defaults to 4?

fme/core/fill.py
```diff
@@ -0,0 +1,287 @@
...
+    tensor = torch.where(tensor.isnan(), mean, tensor)
+    return tensor
+
+
+def _fast_flood_fill(
+    tensor: torch.Tensor,
+    num_steps: int | None = None,
+    blur_kernel_size: int = 5,
+    blur_sigma: float = 1.0,
+    interior_mask: torch.Tensor = None,
```
elynnwu: Surprised you don't need Optional tensor here if it defaults to None?

fme/core/fill.py
```diff
@@ -0,0 +1,287 @@
...
+    blurred_valid_mask = _separable_gaussian_blur(
+        original_valid_mask, blur_kernel_size, blur_sigma
+    )
+
+    # smoothly interpolate across the boundary
+    tensor = (tensor * blurred_valid_mask) + (
+        blurred_tensor * (1.0 - blurred_valid_mask)
+    )
+
+    return tensor, steps_taken
```
elynnwu: It looks like steps_taken is not actually used, maybe don't return it?

fme/core/fill.py
```diff
@@ -0,0 +1,287 @@
...
+class SmoothFloodFill:
+    """Fill NaN regions using flood fill with smooth boundary transitions.
+
+    Operates in three phases: (1) mean-fill deep interior NaN pixels that
+    would not be reached within ``num_steps`` of edge expansion, (2) iterative
+    neighbor-averaging that grows valid pixels inward from the boundary, and
+    (3) Gaussian blur interpolation across the original valid/NaN boundary to
+    avoid sharp seams.
+
+    Interior masks are computed once per variable name and cached.
```
elynnwu: Maybe add a note that this assumes mask region doesn't change in time

fme/core/test_fill.py
```diff
@@ -0,0 +1,69 @@
...
+    nan_cols = [13, 14, 15, 0, 1, 2]
+
+    data = torch.arange(W, dtype=torch.float32).expand(1, 1, H, W).clone()
+    original = data.clone()
+    data[:, :, :, nan_cols] = float("nan")
+
+    flood_fill = SmoothFloodFill(
+        num_steps=num_steps, blur_kernel_size=5, blur_sigma=1.0
+    )
+    filled = flood_fill(data, "name")
```
elynnwu, line 18: Can you call this again to test that the same mask region is used in the subsequent calls?

fme/core/fill.py
jpdunc23: Hmm, maybe `torch.Tensor` is treated as `Any` by mypy?

fme/ace/aggregator/one_step/deterministic.py
```diff
@@ -91,8 +92,10 @@ def __init__(
             ),
         }
         try:
+            flood_fill = SmoothFloodFill(num_steps=4)
```
elynnwu, line 95: [nit] no need to specify since it defaults to 4 now in the class

fme/core/fill.py
```diff
@@ -0,0 +1,306 @@
+from typing import NamedTuple
+
+import torch
+import torch.nn.functional as F
+
+
+class _Masks(NamedTuple):
```
elynnwu, line 7: Nice that this is immutable once created

jpdunc23:
Example run with "spectrum/nan_filled" snapshots showing gen / target after flood fill: https://wandb.ai/ai2cm/ace-samudra-coupled-cm4/runs/k25xt6af?nw=i1xx728vwb

fme/ace/aggregator/one_step/deterministic.py
jpdunc23: True but I like to keep it for documentation purposes.

PR #1032, opened 2026-04-02
Bugfix for pre-processing E3SM data

elynnwu:
This PR fixes bugs in preprocessing E3SM data:

1. Correct reconstruction of precipitation: `PRECT = PRECC + PRECL`
2. Ocean vertical interface levels mistake

### DIFF
--- a/scripts/data_process/compute_dataset_e3smv2.py
+++ b/scripts/data_process/compute_dataset_e3smv2.py
@@ -264,13 +264,23 @@ def compute_surface_precipitation_rate(
     liquid_precip_density: float = LIQUID_PRECIP_DENSITY,
     output_name: str = SURFACE_PRECIPITATION,
     standard_names=None,
+    include_snow_rate: bool = False,
 ):
     if total_precip_rate_name not in ds.variables:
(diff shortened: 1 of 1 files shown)


scripts/data_process/compute_dataset_e3smv2.py
```diff
@@ -266,10 +266,9 @@ def compute_surface_precipitation_rate(
     standard_names=None,
 ):
     if total_precip_rate_name not in ds.variables:
+        # see https://github.com/E3SM-Project/e3sm_diags/blob/f80253a063fde83dd8b588a7e144622d702718fb/e3sm_diags/derivations/formulas.py#L230-L233
         precip = (
-            ds[standard_names.surface_snow_rate]
-            + ds[standard_names.surface_ice_rate]
-            + ds[standard_names.convective_snow_rate]
+            ds[standard_names.surface_ice_rate]
```
jpdunc23: Maybe we should put this behind a config flag that is by default turned on with a note that for use with the ACE-EAMv2 checkpoint the flag should be turned off?

jpdunc23 (approved):
LGTM. Just left one suggestion but I don't feel super strongly about it since it's probably an unlikely edge case.

scripts/data_process/compute_dataset_e3smv2.py
elynnwu: Good point - updated

PR #1033, opened 2026-04-03
add zonal interpolation fill for NaNs

Arcomano1234:

Changes:

- [ ] Tests added

### DIFF
--- a/fme/core/dataset/utils.py
+++ b/fme/core/dataset/utils.py
@@ -145,6 +145,39 @@ def _load_all_variables(
     return ds[variables].compute()
 
 
+def _zonal_interp_periodic(array: np.ndarray, pad_width: int = 3) -> np.ndarray:
+    """Fill NaNs via periodic linear interpolation along the last axis.
+
+    Wraps the longitude dimension periodically, then uses linear interpolation
+    to fill NaN gaps. Rows that are entirely NaN are left unchanged.
+
+    Args:
+        array: Array of any shape; interpolation is along the last axis.
+        pad_width: Number of elements to mirror from each end for periodicity.
+
+    Returns:
+        Copy of array with NaNs filled where possible.
+    """
+    n_lon = array.shape[-1]
+    left = array[..., -pad_width:]
+    right = array[..., :pad_width]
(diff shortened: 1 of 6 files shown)


fme/core/dataset/utils.py
```diff
@@ -153,10 +186,14 @@ class FillNaNsConfig:
     Parameters:
         method: Type of fill operation. Currently only 'constant' is supported.
         value: Value to fill NaNs with.
+        zonal_interp_variables: Variables to fill via periodic zonal (longitude)
```
brianhenn, line 189: If this is getting merged it seems like we'd want to add 'zonal_interp` as a method in addition to 'constant' rather than adding a new arg

fme/core/dataset/utils.py
```diff
@@ -145,6 +145,39 @@ def _load_all_variables(
...
+    right = array[..., :pad_width]
+    padded = np.concatenate([left, array, right], axis=-1)
+
+    orig_shape = padded.shape
+    flat = padded.reshape(-1, orig_shape[-1]).copy()
+
+    x = np.arange(flat.shape[-1])
+    for i in range(flat.shape[0]):
+        row = flat[i]
+        mask = np.isnan(row)
```
brianhenn, line 172: This method is greedy, where does this fall in the call stack relative to where we actually load the tensors? I guess not actually wrong to load data earlier, but a bit confusing in terms of the data model

fme/core/dataset/utils.py
```diff
@@ -145,6 +145,39 @@ def _load_all_variables(
...
+
+    Returns:
+        Copy of array with NaNs filled where possible.
+    """
+    n_lon = array.shape[-1]
+    left = array[..., -pad_width:]
+    right = array[..., :pad_width]
+    padded = np.concatenate([left, array, right], axis=-1)
+
+    orig_shape = padded.shape
```
brianhenn, line 166: Nit use a different name than "orig" here since it has padding on it

brianhenn (commented):
Looks good, some minor suggestions

Arcomano1234:
Closing for now did not see any obvious improvement or need for this feature as of July 2026.

PR #1035, opened 2026-04-06
Move validate_amp calls into init

AnnaKwa:
A bunch of modules call `validate_amp` every single forward pass. This only needs to be checked once and can be done at initialization.


frodre (commented):
Was there a performance degradation or something else annoying associated with this update? 

I can't find any documentation on `torch.is_autocast_enabled` (maybe from older API?), but wouldn't it not work during init since autocasting mode happens within the forward pass not during init during training?

A minimal failing test that then gets fixed using this change would be illustrative.

AnnaKwa:
> Was there a performance degradation or something else annoying associated with this update?
> 
> I can't find any documentation on `torch.is_autocast_enabled` (maybe from older API?), but wouldn't it not work during init since autocasting mode happens within the forward pass not during init during training?
> 
> A minimal failing test that then gets fixed using this change would be illustrative.

No, I just asked Claude if there were any inefficiencies in the diffusion unet and it brought this up. It seemed like a lot of redundant checks but as you point out, the context is set at runtime so I don't think there's a clean way of just moving it to init. Closing PR

PR #1038, opened 2026-04-07
Provide default args for get_initial_condition

oliverwm1:
Inspired by making an example notebook for colab. This would allow us to remove a couple lines of code in that notebook.


elynnwu (approved):
lgtm, thanks for updating

PR #1042, opened 2026-04-09
Avoid having two copies of weights on GPU when loading checkpoints

oliverwm1:
I noticed when calling the `load_stepper` function that the GPU memory used was 2x the size of the checkpoint being used. This PR implements some changes to avoid this.

Changes:

Don't have a great testbed, but compared running `load_stepper` on the ACE2-ERA5 checkpoint in a colab notebook with 2026.1.1 fme release and this branch.

v2026.1.1:
Screenshot 2026-04-09 at 9 44 53 AM

This branch:
Screenshot 2026-04-09 at 9 22 34 AM

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -343,8 +343,8 @@ class TrainOutput(TrainOutputABC):
     target_data: EnsembleTensorDict
     time: xr.DataArray
     normalize: Callable[[TensorDict], TensorDict]
-    derive_func: Callable[[TensorMapping, TensorMapping], TensorDict] = (
-        lambda x, _: dict(x)
+    derive_func: Callable[[TensorMapping, TensorMapping], TensorDict] = lambda x, _: (
+        dict(x)
     )
 
     def __post_init__(self):
@@ -1350,6 +1350,16 @@ class Stepper:
         stepper.load_state(state)
         return stepper
 
+    def to(self, device: str) -> "Stepper":
+        """Move the stepper's tensors and modules to the specified device."""
+        self._step_obj.to(device)
+        self._dataset_info = self._dataset_info.to(device)
+        if hasattr(self._input_process_func, "to"):
+            self._input_process_func.to(device)
+        if hasattr(self._output_process_func, "to"):
+            self._output_process_func.to(device)
+        return self
+
     def set_eval(self) -> None:
         for module in self.modules:
             module.eval()
@@ -1795,10 +1805,10 @@ def load_stepper(
     if override_config is None:
         override_config = StepperOverrideConfig()
 
-    checkpoint = torch.load(
-        checkpoint_path, map_location=get_device(), weights_only=False
-    )
+    checkpoint = torch.load(checkpoint_path, map_location="cpu", weights_only=False)
     stepper = Stepper.from_state(checkpoint["stepper"])
+    del checkpoint
+    stepper.to(get_device())
 
     if override_config.ocean != "keep":
         logging.info(
--- a/fme/ace/step/fcn3.py
+++ b/fme/ace/step/fcn3.py
@@ -507,3 +507,10 @@ class FCN3Step(StepABC):
             state: The state to load.
         """
         self.module.load_state_dict(state["module"])
+
+    def to(self, device: str) -> "FCN3Step":
+        """Move the step's tensors and modules to the specified device."""
+        self.module = self.module.to(device)
+        self._normalizer = self._normalizer.to(device)
+        self._corrector.to(device)
+        return self
--- a/fme/core/atmosphere_data.py
+++ b/fme/core/atmosphere_data.py
@@ -58,6 +58,9 @@ class HasAtmosphereVerticalIntegral(Protocol):
     def get_bk(self) -> torch.Tensor:
         pass
 
+    def to(self, device: str) -> "HasAtmosphereVerticalIntegral":
+        pass
+
 
 class AtmosphereData:
     """Container for atmospheric data for accessing variables and providing
--- a/fme/core/coordinates.py
+++ b/fme/core/coordinates.py
@@ -108,6 +108,9 @@ PostProcessFnType = Callable[[TensorMapping], TensorDict]
 
 
 class NullPostProcessFn:
+    def to(self, device: str) -> "NullPostProcessFn":
+        return self
+
     def __call__(self, data: TensorMapping) -> TensorDict:
         return dict(data)
 
--- a/fme/core/corrector/atmosphere.py
+++ b/fme/core/corrector/atmosphere.py
@@ -162,6 +162,11 @@ class AtmosphereCorrector(CorrectorABC):
         else:
             self._dry_air_precision = torch.float64
 
+    def to(self, device: str) -> "AtmosphereCorrector":
+        if self._vertical_coordinate is not None:
+            self._vertical_coordinate = self._vertical_coordinate.to(device)
+        return self
+
     def __call__(
         self,
         input_data: TensorMapping,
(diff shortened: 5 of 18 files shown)


fme/ace/stepper/single_module.py
```diff
@@ -1795,10 +1805,10 @@ def load_stepper(
     if override_config is None:
         override_config = StepperOverrideConfig()
 
-    checkpoint = torch.load(
-        checkpoint_path, map_location=get_device(), weights_only=False
-    )
+    checkpoint = torch.load(checkpoint_path, map_location="cpu", weights_only=False)
     stepper = Stepper.from_state(checkpoint["stepper"])
+    del checkpoint
+    stepper.to(get_device())
```
jpdunc23, line 1811: I wonder if there is a way to avoid these many `to()` methods, which may be a bit error-prone and annoying to maintain. For example, the motivating issue here could be avoided by just refactoring `load_stepper_config()` to load the checkpoint on CPU.

A util like [`_tensor_to_device()`](https://github.com/ai2cm/ace/blob/a456ae867156cfea676ca7292d8a047900c938a5/fme/core/generics/trainer.py#L803-L811)  (from a [draft PR](https://github.com/ai2cm/ace/pull/1043) I'm working on) might also be helpful to allow subsets of the checkpoint state dicts to be transferred to GPU without loading the full model into GPU memory.

oliverwm1: I think our current practice of moving stuff to device within init methods is not ideal (e.g. doing this in the StandardNormalizer made a lot more tests flaky back in the day, by requiring test data to also be moved to device). Having the `.to()` methods would provide useful flexibility to move tensors to GPU at the appropriate time. But yes, I can look into whether we can resolve the issue I had in a less intrusive way.

> For example, the motivating issue here could be avoided by just refactoring load_stepper_config() to load the checkpoint on CPU.

The issue I ran into doesn't have anything to do with `load_stepper_config`. The 2x memory usage happens even when just calling `load_stepper`.

jpdunc23: > The issue I ran into doesn't have anything to do with load_stepper_config. The 2x memory usage happens even when just calling load_stepper.

No? I thought it stemmed from the fact that `run_evaluator_from_config()` calls both `load_stepper_config()` ([here](https://github.com/ai2cm/ace/blob/366f1171d728bc5aa6fd6df94a5cbce053a25d86/fme/ace/inference/evaluator.py#L333)) and `load_stepper()` (a few lines later, [here](https://github.com/ai2cm/ace/blob/366f1171d728bc5aa6fd6df94a5cbce053a25d86/fme/ace/inference/evaluator.py#L348)), thus leading to 2x `torch.load()` calls. 

But maybe I'm misunderstanding your issue.

oliverwm1: The memory usage difference which I show in the PR description above was measured only be running `load_stepper()` itself, not by running `run_evaluator_from_config()`. Maybe I can use the benchmarking code to make a test for this to show more clearly...

jpdunc23: OK, thanks, ya based on our convo in slack it's clear how this happens with `load_stepper` alone due to having the loaded copy on GPU as well as the initialized Stepper with it's module weights sent to GPU via e.g. `SingleModuleStep.__init__` calling `self.module = module.to(get_device())`. This is one example of a `Tensor.to()` call we'd want to remove when centralizing in `SingleModuleStep.to()`

These `to()` methods might be a bit repetitive and a pain to maintain, but on the other hand that might be worthwhile for the extra documentation they provide as a centralized source of truth for what needs to be on the device, as well as solving the issue here and in my draft PR.

I think this PR could be reframed as a refactoring to support `to()`, in which case this bugfix (and likely others, e.g. when resuming from a checkpoint in `Trainer` or fine-tuning with `ParameterInitializationConfig`) is just a free consequence of the refactoring.

fme/ace/stepper/single_module.py
```diff
@@ -1350,6 +1350,16 @@ def from_state(cls, state) -> "Stepper":
...
         return stepper
 
+    def to(self, device: str) -> "Stepper":
+        """Move the stepper's tensors and modules to the specified device."""
+        self._step_obj.to(device)
+        self._dataset_info = self._dataset_info.to(device)
+        if hasattr(self._input_process_func, "to"):
+            self._input_process_func.to(device)
+        if hasattr(self._output_process_func, "to"):
+            self._output_process_func.to(device)
```
mcgibbon, line 1360: I don't love the `hasattr` calls, it smells like input_process_func should be defined as an API where some of our funcs have a does-nothing `.to` method. I'm inclined to say I'm nit picking, and I'm fine if you keep it like this (might be a pain to fix), but I also want to say I'm frequently surprised by the ways in which writing type-unsafe code causes issues down the line, and the ways that writing the same code type-safely makes it easier to refactor down the line.

It does seem like you've already added the do-nothing methods, in which case this should be formalized so you don't need the hasattr's. Maybe define a helper `FuncPostProcessFn(func)` with a dummy `.to`, if you need that for some stragglers.

fme/ace/stepper/single_module.py
mcgibbon: If you wanted to fix the current issue without resorting to adding all the `.to` methods, I would suggest updating `Stepper.from_state` so that while it continues to initialize itself using GPU tensors, the assignment operations support CPU inputs. GPT seems to believe you can `gpu_module.load_state_dict(cpu_state)` without any issues, for example, so this might be easier than you'd think. Then while we'd load the data with `map_location="cpu"`, we wouldn't need a `.to` call.

I'd probably suggest doing that for this fix if it works, regardless of whether we want to explicitly control `.to` methods for other reasons, so that the `.to` change can be evaluated solely for those other reasons.

mcgibbon: Actually, this brings up an interesting question, how much work is `stepper.to(get_device())` actually doing here? I notice no changes are made here to the initialization of any objects, so presumably the ones that make the double-copy tensors are still initializing their state on GPU. It's only ones that don't properly copy tensors in their `from_state` that would need to be changed (like LatLonCoordinates), and probably they should be changed anyways so they make their own (on-device) copy of the data passed to `from_state` to avoid weird memory coupling issues I've run into in some other PRs.

For what it's worth, all tests in `fme/ace/stepper` pass even if I delete this `stepper.to` line (the train test fails, on a LatLonCoordinate device not matching error)

mcgibbon:
One thing to maybe worry about is, a big way we stay sane right now is that everything we initialize initializes itself on-device. The only exception is DataLoader-loaded data, which we cast at a specific place (when it's yielded from the DataLoader). I think being able to `Object.from_state(cpu_state)` would be very helpful here, but it is probably nice if it creates an object that's already on-device (and we should probably update our existing from_state methods to do that consistently).

fme/ace/stepper/single_module.py
jpdunc23: > I notice no changes are made here to the initialization of any objects

This is something we would need to tackle in this PR (or whatever PR adds `to()` methods, if we go that route), just hasn't been done here yet given the "Draft" status.

jpdunc23: > `gpu_module.load_state_dict(cpu_state)`

This sounds like a reasonable approach to me, if it works. I'd suggest we simultaneously update anywhere else where we're using `map_location=get_device()`, e.g. in `Trainer`'s checkpoint restoring, `fme.coupled`, `fme.downscaling`, etc.

mcgibbon: That sounds good to me.

oliverwm1:
superseded by #1061

PR #1043, opened 2026-04-09
Add `OptimizationConfig.resume_optimizer_ckpt_path` for restoring the optimizer state for fine-tuning

jpdunc23:
Adds `resume_optimizer_ckpt_path` to `OptimizationConfig`. This allows for restoring the optimizer state in a fine-tuning (i.e., non-resuming) job.

Changes:
- Adds `Optimization.load_optimizer_state_for_finetuning()`. Unlike `Optimization.load_state()`, loads per-parameter optimizer state from a checkpoint but allows the LR scheduler and per-group hyperparameters (`lr`, `weight_decay`, `betas`, etc.) to be set by the training config, overriding the checkpoint values.

- [x] Tests added


fme/core/generics/trainer.py
```diff
@@ -787,6 +798,42 @@ def _restore_checkpoint(trainer: Trainer, checkpoint_path):
...
 
+def _tensors_to_device(obj, device: torch.device):
+    """Recursively move all tensors in a nested dict/list to *device*."""
+    if isinstance(obj, torch.Tensor):
+        return obj.to(device)
+    elif isinstance(obj, dict):
+        return {k: _tensors_to_device(v, device) for k, v in obj.items()}
+    elif isinstance(obj, list | tuple):
+        return type(obj)(_tensors_to_device(v, device) for v in obj)
+    return obj
```
jpdunc23: Possibly this should be made public / moved to another module, or maybe we don't actually need it.

fme/ace/train/train_config.py
```diff
@@ -222,6 +222,7 @@ class TrainConfig:
     evaluate_before_training: bool = False
     save_best_inference_epoch_checkpoints: bool = False
     lr_tuning: LRTuningConfig | None = None
+    resume_optimization_ckpt_path: str | None = None
```
mcgibbon: Question: Any reason not to put this in the OptimizationConfig? That's probably where I'd expect to find it. It seems like if it lived there, we could avoid needing any logic on the generic Trainer since the optimization would handle the loading during OptimizationConfig.build, before the generic code is called. It's a bit out of place for the Trainer to handle initialization concerns.

fme/core/optimization.py
```diff
@@ -195,6 +195,51 @@ def set_learning_rate(self, lr: float):
         for param_group in self.optimizer.param_groups:
             param_group["lr"] = lr
 
+    def load_optimizer_state_for_finetuning(self, state: dict):
```
mcgibbon, line 198: Question: Why does this differ from `load_state`?

fme/ace/train/train_config.py
mcgibbon: Claude: +1 on moving this to `OptimizationConfig` — having `build()` handle the loading would keep optimization initialization co-located and remove this concern from the Trainer and `TrainConfigProtocol`. During preemption resume the finetune state would be loaded in `build()` and then immediately overwritten by `restore_checkpoint()`, but that one-time extra read seems acceptable.

fme/core/optimization.py
mcgibbon: Claude: This differs from `load_state` in two ways: (1) it skips the scheduler, so the new training phase's schedule starts from scratch, and (2) it preserves the fresh config's per-group hyperparameters (`lr`, `weight_decay`, `betas`, etc.) rather than adopting the checkpoint's values. Only the per-parameter running state (e.g. Adam moment estimates) and grad scaler are restored from the checkpoint.

A more precise opening line for the docstring could be something like: "Load per-parameter optimizer running state (e.g. Adam moment estimates) and grad scaler state from a checkpoint for fine-tuning."

fme/core/generics/trainer.py
mcgibbon: Claude: If the loading logic moves to `OptimizationConfig.build()` per Jeremy's suggestion above, this helper could be encapsulated there or co-located with `load_optimizer_state_for_finetuning`. I don't think it needs to be public unless there's another use case for it.

fme/core/optimization.py
mcgibbon: Suggestion: Update the short-description for the method to avoid saying you're loading "the optimizer state" (something like Claude suggested, or more generic like "non-config optimizer state" or "per-parameter optimizer state").

fme/ace/train/train_config.py
jpdunc23: The main reasons to keep it on `TrainConfigProtocol` is to avoid extra unnecessary checkpoint reads and to unify handling of restoring the `ckpt_xxxx.tar` optimizer state and (in a later PR) the EMA state as well.

But I agree having this on `OptimizationConfig` is cleaner, will implement here and (in the later PR) a similar feature on `EMAConfig`.

mcgibbon: If we get particularly concerned about checkpoint reads (I'm not particularly worried, it's startup cost), we should consider making the training checkpoint more modular, storing the data for these different components in different files for example or in some kind of packed file that allows reading part of the data.

fme/core/optimization.py
```diff
@@ -255,6 +309,7 @@ class OptimizationConfig:
     checkpoint: CheckpointConfig = dataclasses.field(
         default_factory=lambda: CheckpointConfig()
     )
+    resume_optimizer_ckpt_path: str | None = None
```
mcgibbon, line 312: I like the name!

Optional: You could expand ckpt to checkpoint to be more style-consistent but I'm fine with it as-is (it's already long).

fme/core/optimization.py
```diff
@@ -292,6 +352,42 @@ def from_state(cls, state: Mapping[str, Any]) -> "OptimizationConfig":
         return cls(**state)
 
 
+def _tensors_to_device(obj, device: torch.device):
```
mcgibbon: nit: Could add typing to obj and the return value, to make the code more obviously correct. I think you need TypeAlias to do so because it's recursive, e.g.:

```
from typing import TypeAlias

NestedTensor: TypeAlias = "torch.Tensor | dict[str, NestedTensor] | list[NestedTensor] | tuple[NestedTensor]"
```

fme/core/test_optimization.py
```diff
@@ -535,6 +537,241 @@ def test_load_state_then_set_learning_rate():
...
+    )
+
+    for _ in range(3):
+        with optimization.autocast():
+            loss = model(x).sum()
+        optimization.accumulate_loss(loss)
+        optimization.step_weights()
+
+    saved_state = optimization.get_state()
+    assert saved_state["gscaler_state_dict"] != {}
```
mcgibbon, line 686: Good check!

mcgibbon (approved):
LGTM, just optional comments.

fme/core/optimization.py
jpdunc23: OK, will leave as is to keep it a bit shorter for now.

PR #1044, opened 2026-04-10
Ensure `IceCorrectorConfig` is added to `CorrectorSelector` registry

William-gregory:
Needed to register ice corrector after recent changes to CorrectorConfigABC

Changes:
- added `from fme.core.corrector.ice import IceCorrectorConfig` to `fme.ace.__init__.py`


jpdunc23 (approved):
LGTM, thank you!

PR #1050, opened 2026-04-15
Change default preemptible option in job runner

elynnwu:
Getting this warning in the latest gantry:
```sh
DeprecationWarning: The option 'preemptible' is deprecated. Use --min-runtime and/or --no-auto-resume instead.
```

I think `--min-runtime 0` keeps the job preemptible.


jpdunc23:
LGTM! Merging

PR #1051, opened 2026-04-16
Update dataset path for era5 data processing config

mcgibbon:
The ERA5 data config was pointing to an old version of the data that is missing frozen precipitation, leading to the time-coarsening run failing. This PR updates that configuration so it runs.


scripts/data_process/configs/era5-1deg-8layer-1940-2025.yaml
spencerkclark, line 1: Could you update the filename to end with `2025`?

spencerkclark (approved):
Thanks for updating / running this—could you update the `Makefile` rule for the new `compute_dataset.sh` flags too?

https://github.com/ai2cm/ace/blob/1ff21b305b01ba8f381275ac8f981988045950c6/scripts/data_process/Makefile#L246-L248

scripts/data_process/Makefile
```diff
@@ -245,11 +245,11 @@ e3smv2_1deg_climSST_dataset:
 
 .PHONY: era5_1deg_stats_beaker_dataset
 era5_1deg_stats_beaker_dataset:
-	./compute_stats.sh --config configs/era5-1deg-8layer-1940-2022.yaml
+	./compute_stats.sh --config configs/era5-1deg-8layer-1940-2025.yaml
 
 .PHONY: era5_1deg_16layer_stats_beaker_dataset
 era5_1deg_16layer_stats_beaker_dataset:
-	./compute_stats.sh --config configs/era5-1deg-16layer-1940-2022.yaml
+	./compute_stats.sh --config configs/era5-1deg-16layer-1940-2025.yaml
```
spencerkclark: This should stay the same since we have not regenerated a 16-layer dataset yet.
```suggestion
	./compute_stats.sh --config configs/era5-1deg-16layer-1940-2022.yaml
```

scripts/data_process/Makefile
```diff
@@ -245,11 +245,11 @@ e3smv2_1deg_climSST_dataset:
 
 .PHONY: era5_1deg_stats_beaker_dataset
 era5_1deg_stats_beaker_dataset:
-	./compute_stats.sh --config configs/era5-1deg-8layer-1940-2022.yaml
+	./compute_stats.sh --config configs/era5-1deg-8layer-1940-2025.yaml
```
spencerkclark, line 248: Could optionally update this to include the time-coarsening, but fine to keep it as stats only too.

mcgibbon: I think I'll leave it for now, this way is more instructional (instead of time coarsening being a side-effect) and helps demonstrate how to run `compute_dataset.sh` directly with flags (which is what I do, personally).

PR #1052, opened 2026-04-16
Remove deprecated n_forward_steps from TrainConfig

mcgibbon:
TrainConfig.n_forward_steps has been deprecated for a while, as we've moved to specify this in the stepper configuration and now in the stepper training configuration, where stochastic sampling of forward step lengths is supported. This PR removes the deprecated option, requiring the new method of configuration.

Changes:
- TrainConfig.n_forward_steps is removed, and validation will fail if its TrainStepperConfig.train_n_forward_steps is not set.
- Renamed `TrainStepperConfig.train_n_forward_steps` to `TrainStepperConfig.n_forward_steps`, a breaking change for all train configs

- [ ] Tests added

### DIFF
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -298,6 +298,11 @@ class TrainConfig:
                 f"During training, experiment_dir must currently be a local "
                 f"directory, got {self.experiment_dir!r}."
             )
+        if self.stepper_training.train_n_forward_steps is None:
+            raise ValueError(
+                "train_n_forward_steps must be specified in stepper_training "
+                "to determine data loading requirements."
+            )
 
     def set_random_seed(self):
         if self.seed is not None:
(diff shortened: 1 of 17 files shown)


fme/ace/train/train_config.py
```diff
@@ -309,6 +298,11 @@ def __post_init__(self):
                 f"During training, experiment_dir must currently be a local "
                 f"directory, got {self.experiment_dir!r}."
             )
+        if self.stepper_training.train_n_forward_steps is None:
```
jpdunc23: Minor point, but for config validation there's an implicit assumption that `self.stepper_training.train_n_forward_steps_schedule` is not None so long as `self.stepper_training.train_n_forward_steps` is not None, otherwise we'll have to wait for the runtime call of `_get_n_forward_steps()` to see an error. If you wanted to be extra conservative you could instead check `if self.stepper_training.train_n_forward_steps_schedule is None`.

jpdunc23 (commented):
Thanks for this! Aside from failing tests, looks good.

What do you think about renaming `TrainStepperConfig.train_n_forward_steps` to `TrainStepperConfig.n_forward_steps`? I think it would be nicer, but fine with leaving as is if you prefer.

`fme/ace/test_ocean_train.py` may need to be updated since it uses hard-coded `n_forward_steps: 2` in its `_TRAIN_CONFIG_TEMPLATE`.

fme/ace/train/train_config.py
mcgibbon: Added an additional check for that, since train_n_forward_steps_schedule isn't something set by a user. The value error should always catch this, unless there's a bug in the code (in which case the new RuntimeError will indicate that).

jpdunc23 (approved):
LGTM! Reminder to update the description noting the change from `TrainStepperConfig.train_n_forward_steps` to `TrainStepperConfig.n_forward_steps`

PR #1054, opened 2026-04-17
Add new `test_ice_train` script to catch bugs which affect GraphCast and `IceCorrector`

William-gregory:
Have added a new `test_ice_train.py` script in `fme.ace`, which is built from `test_ocean_train.py`. This tests training and inference using `FloeNet` and `fme.core.corrector.ice`, so should catch any bugs which impact these workflows in the future.


fme/ace/test_ice_train.py
```diff
@@ -0,0 +1,516 @@
...
+    if "mask_2d" in variable_names:
+        mask_data_values = np.ones(horizontal_shape, dtype=np.float32)
+        mask_data_values[1, 1] = 0.0  # small island
+        masks_data[0] = xr.DataArray(
+            mask_data_values.astype(np.float32),
+            dims=horizontal_dims,
+            coords=horizontal_coords,
+            name="mask_2d",
+        )
+        ds["mask_2d"] = masks_data[0]
```
jpdunc23: Maybe don't bother with `masks_data` since you don't need 3D masks.

fme/ace/test_ice_train.py
```diff
@@ -0,0 +1,516 @@
...
+        )
+    xr.Dataset(data_vars=data_vars).to_netcdf(filename, format="NETCDF4_CLASSIC")
+
+
+_TRAIN_CONFIG_TEMPLATE = """
+experiment_dir: {results_dir}
+save_checkpoint: true
+save_per_epoch_diagnostics: {save_per_epoch_diagnostics}
+max_epochs: {max_epochs}
+n_forward_steps: 2
```
jpdunc23: Heads up with https://github.com/ai2cm/ace/pull/1052 merged you'll need to move `n_forward_steps` to `stepper_training`.

fme/ace/test_ice_train.py
```diff
@@ -0,0 +1,516 @@
...
+        kind="mean",
+    )
+    _save_ice_scalar_stats_netcdf(
+        stats_dir / "stats-stddev.nc",
+        variable_names=stats_variable_names,
+        kind="std",
+    )
+
+    monthly_dim_sizes = get_ice_dim_sizes(n_time=10 * 12, nz_levels=1)
+    monthly_out_vars = [v for v in out_variable_names if not re.match(r".+_\d+", v)]
```
jpdunc23: Probably no need to use the list comprehension here since you don't have variables that end with `_\d+` (i.e., 3D levels). You can just create the `MonthlyReferenceData` with `out_variable_names`.

TBH, this probably isn't critical to test here given it is already tested in `test_train.py`, so feel free to remove as well.

fme/ace/test_ice_train.py
```diff
@@ -0,0 +1,516 @@
...
+      normalization:
+        network:
+          global_means_path: '{global_means_path}'
+          global_stds_path: '{global_stds_path}'
+      builder:
+        type: FloeNet
+        config:
+          latent_dimension: 4
+          activation: SiLU
+          meshes: 6
```
jpdunc23:
```suggestion
          meshes: 4
```

This reduced the test runtime from about 25s to 7s on my laptop.

jpdunc23 (approved):
A few nits / suggestions, but otherwise LGTM. Please change `meshes: 6` to `meshes: 4` as I found this sped up the test a lot.

After merging main you'll also need to update the train config template per my other comment, but aside from those two changes feel free to merge when ready. No need for re-review.

PR #1055, opened 2026-04-17
Enable configuring time subsetting prior to time coarsening

spencerkclark:
Some of our datasets start at unusual times.  For instance the ERA5 dataset starts on `"1940-01-01T12:00:00"`.  When time coarsening, this can lead to unconventional intervals, which could be inconsistent with other datasets.  For example for default daily coarsening of the ERA5 dataset this leads to times of `"1940-01-02T06:00:00"`, `"1940-01-03T06:00:00"`, and so on.

A way to address this is to apply a time slice prior to coarsening.  In the case of ERA5 that would be to provide a start time of `"1940-01-02T06:00:00"`, such that the daily coarsened data would start cleanly on `"1940-01-03T00:00:00"`.

We take the opportunity to add `"h10"`, `"Q10"`, `"TMP10"`, `"UGRD10"`, and `"VGRD10"` as snapshot variables to the ERA5 coarsening configurations as well.

Changes:
- Adds an optional `input_time_slice` attribute to `TimeCoarsenConfig`.
- Updates the ERA5 time coarsening configurations to include a `time_slice` as well as `"h10"`, `"Q10"`, `"TMP10"`, "UGRD10", and `"VGRD10"` as `snapshot_names`.

Argo workflows:
- 1° ERA5 time-coarsening/stats: `compute-fme-dataset-ensemble-gckg7`
- 4° ERA5 time-coarsening/stats: `compute-fme-dataset-ensemble-j8rxk`

- [x] Tests added

### DIFF
--- a/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
@@ -9,23 +9,27 @@ stats:
   data_type: ERA5
 time_coarsen:
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-03-19-era5-4deg-8layer-daily-1940-2022
-  stats_output_directory: gs://vcm-ml-intermediate/2026-03-19-era5-4deg-8layer-daily-stats-1990-2019
+  data_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-1940-2022
+  stats_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-stats-1990-2019
   snapshot_names:
     - DPT2m
     - PRESsfc
+    - Q10
     - Q200
     - Q2m
     - Q500
     - Q850
+    - TMP10
     - TMP200
     - TMP2m
     - TMP500
     - TMP850
+    - UGRD10
     - UGRD10m
     - UGRD200
     - UGRD500
     - UGRD850
+    - VGRD10
     - VGRD10m
     - VGRD200
     - VGRD500
@@ -46,6 +50,7 @@ time_coarsen:
     - eastward_wind_5
     - eastward_wind_6
     - eastward_wind_7
+    - h10
     - h1000
     - h200
     - h250
@@ -97,3 +102,5 @@ time_coarsen:
     - land_fraction
     - latitude
     - longitude
+  time_slice:  # ensures coarsened data starts on 1940-01-03T00:00:00
+    start: "1940-01-02T06:00:00"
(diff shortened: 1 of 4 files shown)


scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
```diff
@@ -97,3 +102,5 @@ time_coarsen:
     - land_fraction
     - latitude
     - longitude
+  time_slice:  # ensures coarsened data starts on 1940-01-03T00:00:00
```
mcgibbon: Note for later, but we should consider doing something like this in all the coarsening configs (even the ones that don't necessarily need it). Maybe in the form of making this argument required, at least for the start of the slice.

mcgibbon: Suggestion (optional): maybe `input_time_slice` to make it clear this is the pre-coarsening time slice

scripts/data_process/test_time_coarsen.py
```diff
@@ -131,3 +131,33 @@ def test_process_path_pair() -> None:
...
+    ds_coarsened = coarsen(ds, config)
+    # Slicing to start at Jan 2 gives [Jan 2..6] (5 times).
+    # factor=2 with trim → 2 output snapshots at indices 1 and 3 of the sliced
+    # dataset, i.e. Jan 3 and Jan 5.
+    xr.testing.assert_equal(
+        ds_coarsened["time"],
+        xr.DataArray(times[2::2], dims=["time"]),  # Jan 3, Jan 5
+    )
+    assert ds_coarsened.attrs["time_slice.start"] == "2000-01-02"
+    assert ds_coarsened.attrs["time_slice.stop"] == "None"
```
mcgibbon: These attribute names are particularly confusing out of context if you read them when looking at the generated dataset, since they don't describe the start/stop time slice of thee generated dataset. I'd suggest putting them in a more verbose form in the description-like attribute we put in other datasets, or not including it at all (they can be inferred by the time coordinate of the generated dataset).

scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
spencerkclark: Good idea—I opened #1057 and renamed the attribute to `input_time_slice`.

scripts/data_process/test_time_coarsen.py
spencerkclark: Good point.  I went ahead and added a more verbose description as part of the `history` attribute.

mcgibbon (approved):
LGTM

PR #1056, opened 2026-04-17
Downscaling loss weighting

AnnaKwa:
This PR adds two configurable options to loss weighting in the downscaling training config, contained in the optional field `loss_weights`. Both were found to improve the generated outputs.

- `loss_weights.output_channels`: dict with key, value corresponding to output channel name and weight to multiply that variable's loss by. Output names not in this dict are not adjusted. Downweighting precipitation improves the overall outputs' CRPS and extreme values with no adverse effect on precip skill.
- `loss_weights.noise_weight_exponent`: power to raise the default EDM noise-weighted loss by. By default 1.0. Experiments indicate lowering this to ~0.75 improves the generation of pressure and wind extremes; these fields tend to have their noise vs loss dominated by small noise values if using the default EDM weighting.

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -413,6 +413,7 @@ class DiffusionModel:
         self,
         batch: PairedBatchData,
         optimizer: Optimization | NullOptimization,
+        loss_weight_exponent: float = 1.0,
     ) -> ModelOutputs:
         """Performs a denoising training step on a batch of data."""
         _static_inputs = self._subset_static_if_available(batch.coarse)
@@ -435,7 +436,10 @@ class DiffusionModel:
             targets_norm = targets_norm - base_prediction
 
         conditioned_target = condition_with_noise_for_training(
-            targets_norm, self.config.noise_distribution, self.sigma_data
+            targets_norm,
+            self.config.noise_distribution,
+            self.sigma_data,
+            loss_weight_exponent=loss_weight_exponent,
         )
 
         denoised_norm = self.module(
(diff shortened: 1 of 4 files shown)


fme/downscaling/models.py
```diff
@@ -59,6 +59,33 @@ def _rename_normalizer(
...
+@dataclasses.dataclass
+class LossWeightsConfig:
+    """
+    Configuration for loss weighting during training.
+
+    Parameters:
+        output_channels: Per-variable multiplicative weights applied to the loss.
+            Keys are variable names from out_names; variables not listed default to 1.0.
+        noise_weight_exponent: Exponent applied to the EDM noise-level loss weight
+            ``(sigma^2 + sigma_data^2) / (sigma * sigma_data)^2``. The default
```
frodre, line 85: I would add the value found helpful for variables dominated by the lower-noise losses for easy reference like you did docs for the noise code.

fme/downscaling/models.py
```diff
@@ -59,6 +59,33 @@ def _rename_normalizer(
     return StandardNormalizer(means=new_means, stds=new_stds)
 
 
+def _build_variable_loss_weight_tensor(
+    weights: dict[str, float], out_names: list[str]
+) -> torch.Tensor:
+    values = [weights.get(name, 1.0) for name in out_names]
```
frodre, line 70: I would consider putting something in here that errors if a requested loss weight name is not in `out_names` so that we get a fast turnaround on typos, etc.

frodre (approved):
LGTM! Thanks for the addition.  Just a couple minor suggestions.

PR #1059, opened 2026-04-20
Enable `DenoisingMoEPredictor` in inference

AnnaKwa:
This PR allows a `DenoisingMoEPredictor` to be used in the various inference entrypoints. It also allows the `CompositePredictor` to use one to generate patch predictions.


frodre (approved):
LGTM

PR #1060, opened 2026-04-20
Refactor `DiffusionModel.generate`

AnnaKwa:
This separates out helper methods from the original `DiffusionModel.generate` to `prepare_generation_inputs` and `postprocess_generation`. This refactor makes the code for an upcoming mixture of experts predictor much cleaner as it needs to call `edm_sampler` using with its own module but can reuse these methods from its underlying `DiffusionModel`s.
 
Related but outside the scope of this PR: it would be nice to separate out the full denoising concerns (`generate*` methods) out of `DiffusionModel` into a predictor class.

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -523,38 +523,38 @@ class DiffusionModel:
             latent_steps=[],
         )
 
-    @torch.no_grad()
-    def generate(
+    def prepare_generation_inputs(
         self,
         coarse_data: TensorMapping,
         static_inputs: StaticInputs | None,
-        n_samples: int = 1,
-    ) -> tuple[TensorDict, torch.Tensor, list[torch.Tensor]]:
-        # Internal method; external callers should use generate_on_batch /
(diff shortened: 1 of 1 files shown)


fme/downscaling/models.py
```diff
@@ -523,38 +523,38 @@ def train_on_batch(
...
-            sigma_max=self.config.sigma_max,
-            num_steps=self.config.num_diffusion_generation_steps,
-        )
-
+    def postprocess_generated(
+        self,
+        generated_norm: torch.Tensor,
+        coarse_data: TensorMapping,
+        n_samples: int,
+        latent_steps: list[torch.Tensor],
```
frodre: Latent steps isn't used in the post-processing. Can we omit? Or is it used in the follow-on?

frodre (approved):
Looks good to me and I agree on that it would be good to separate out some of the data handling concerns.  I think we'll need to do that for distillation as well.

PR #1062, opened 2026-04-20
[job_runner] Refactor `inference.sh`

jpdunc23:
Refactors `job_runner/inference.sh` to use the same `experiments.txt` file as `evaluate.sh`, but with `inference-config-*.yaml` files rather than `evaluator-config-*.yaml` files. Inference jobs are launched with `run_inf_*` in `experiments.txt`, to distinguish from `run_*` for evaluator jobs.


elynnwu (commented):
I'm good with the change, it makes it easier to keep track of the exps. The only catch is that if you accidentally call evaluate instead of inference because experiments have jobs meant for inference only, it'll error. Maybe it's better to separate them with something like `run_eval_ICx1` vs. `run_inf_ICx1`? or just `run_ICx1` vs. `run_inf_ICx1`

PR #1063, opened 2026-04-20
Updates repo name in perlmutter example make-venv

yikwill:
This PR updates the urls and directory names in `make-venv.sh` of the Perlmutter example to match the new repository name (`full-model` --> `ace`).


jpdunc23 (approved):
LGTM, thanks for the fix!

PR #1066, opened 2026-04-21
Fork DISCO convolution with FFT-based contraction into fme/core/disco

mcgibbon:
Fork the DISCO (discrete-continuous) spherical convolution from torch-harmonics into `fme/core/disco`, replacing the CUDA-kernel and sparse-torch contraction paths with a single FFT-based contraction. This gives us a fast pure-PyTorch implementation that works on both CPU and GPU without requiring the optional CUDA extension, and lets us add filter basis types (like isotropic Morlet) that upstream does not support.

When torch-harmonics' optimized CUDA kernels *are* installed and we're running on GPU, we fall back to `th.DiscreteContinuousConvS2` so that the CUDA kernels are still used for the basis types it supports.

### Changes versus upstream torch-harmonics

- **Removed** CUDA kernel wrappers and torch sparse-matrix contraction — the sole forward path is FFT-based (`_disco_s2_contraction_fft`), we fall back on torch-harmonics for the case where we're using GPU with a kernel it supporst and the optimized CUDA kernels are available.
- **Added** banded FFT precomputation (`_precompute_psi_banded`) to reduce memory versus full densification
- **Added** `IsotropicMorletFilterBasis` in `_filter_basis.py` — a truly isotropic variant of the Morlet basis that commutes with latitude reflection
- **Stored** integer index tensors (`psi_ker_idx`, `psi_row_idx`, `psi_col_idx`) as plain attributes instead of buffers, with a `_apply` override for device tracking, to avoid DDP crashes on non-float dtypes
- **Removed** the torch-harmonics package dependency from the DISCO module (quadrature, cache, FFT helpers are self-contained)

### Changes to existing code

- `fme.core.distributed.base._use_torch_harmonics_disco` — selects torch-harmonics when CUDA extension is available, CUDA is present, and basis is not "isotropic morlet"; otherwise uses the fork
- `fme.core.distributed.non_distributed` / `torch_distributed` — `get_disco_conv_s2` routes through the above logic
- `fme.core.models.conditional_sfno.localnet` / `sfnonet` — add `"isotropic morlet"` to `BasisType` and `_compute_cutoff_radius`

- [x] Tests added


fme/core/distributed/non_distributed.py
```diff
@@ -115,7 +115,9 @@ def get_isht(
         return th.InverseRealSHT(nlat, nlon, lmax=lmax, mmax=mmax, grid=grid).float()
 
     def get_disco_conv_s2(self, *args, **kwargs) -> nn.Module:
-        return th.DiscreteContinuousConvS2(*args, **kwargs).float()
+        from fme.core.disco import DiscreteContinuousConvS2
```
mcgibbon: Imports should happen at top-of-file.

fme/core/disco/test_disco.py
```diff
@@ -0,0 +1,120 @@
...
+        bias=True,
+        theta_cutoff=0.5,
+    )
+    x = torch.randn(2, in_channels, *img_shape)
+    with torch.no_grad():
+        y = conv(x)
+    assert y.shape == (2, out_channels, *img_shape)
+
+
+def test_matches_torch_harmonics_reference():
```
mcgibbon: This test is the main coverage of this code contribution. We should parameterize it so that it covers all basis types/norm modes/other configs that we use. I think we also normally use the gauss-legendre grid for example. We don't need to cover every intersection of these configs, just enough to cover all our use patterns.

fme/core/disco/_fft.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def irfft(
+    x: torch.Tensor, n: int | None = None, dim: int = -1, **kwargs
+) -> torch.Tensor:
+    """Inverse real FFT with Hermitian symmetry enforcement."""
+    if n is None:
+        n = 2 * (x.size(dim) - 1)
+
+    x[..., 0].imag = 0.0
```
Arcomano1234, line 46: Question: Should this whole operation use a clone of x instead? 

I had claude look at this part of the code to make sure everything made sense and it flagged a "critical" bug here which is the in-place mutation of x. I suspect this doesn't change anything when we are just using forward operations (which is why the tests that check this pass) but I think this could potentially break things during a backward pass as the imaginary parts of x are now zeroed out.

Arcomano1234: After playing around with some tests this doesn't seem to be a major problem with how `irfft` is currently used.

Arcomano1234 (commented):
LGTM. I did have one comment / question but after playing with the code a little I don't think its necessary to change it from what you have implemented.

fme/core/disco/_fft.py
mcgibbon: FWIW this is copied verbatim from torch-harmonics https://github.com/NVIDIA/torch-harmonics/blob/main/torch_harmonics/fft.py#L83

Arcomano1234 (approved):
LGTM

PR #1067, opened 2026-04-21
Remove barrier from ModelTorchDistributed.shutdown()

mcgibbon:
The `barrier()` call in `ModelTorchDistributed.shutdown()` causes surviving ranks to hang indefinitely when a peer has already exited (e.g., due to OOM or an unhandled exception). This contradicts the design of `Distributed.context()`, which intentionally skips `shutdown()` on exceptions to avoid this kind of hang. `TorchDistributed.shutdown()` already does not have a barrier.

Changes:
- `fme.core.distributed.ModelTorchDistributed.shutdown`: removed `self.barrier()` call before cleanup

- [ ] Tests added


frodre (approved):
LGTM

PR #1068, opened 2026-04-21
Exit with nonzero code in signal handler

mcgibbon:
The signal handler used `sys.exit(0)`, which makes torchrun/SLURM think the process exited successfully. This prevents the launcher from killing sibling ranks, leaving them to hang on the next collective operation when they try to communicate with the terminated rank.

Changes:
- `fme.core.generics.trainer.chain_signal_handler`: changed `sys.exit(0)` to `sys.exit(1)`

- [ ] Tests added


Arcomano1234 (approved):
I see no reason why this shouldn't be approved, but it would be nice to see if we can launch an experiment with this on to make sure it actually fixes the problem.

mcgibbon:
> I see no reason why this shouldn't be approved, but it would be nice to see if we can launch an experiment with this on to make sure it actually fixes the problem.

Yes, I think this should be merged regardless but we should ask whoever has been having the issue to retry it.

PR #1069, opened 2026-04-21
Set explicit 30-minute timeout on all init_process_group calls

mcgibbon:
The Gloo+torchrun and srun `init_process_group` paths had no explicit timeout, relying on PyTorch defaults (30 minutes). This makes all paths consistent with the NCCL+torchrun path, ensuring collective operations time out after 60 minutes when a peer dies.

Changes:
- `fme.core.distributed.TorchDistributed.__init__`: added `timeout=timedelta(minutes=60)` to the Gloo+torchrun and srun `init_process_group` calls

- [ ] Tests added


jpdunc23 (approved):
LGTM. Per discussion on Slack, maybe update to 30 min in this PR?

PR #1070, opened 2026-04-21
Allow Distributed.get_instance() without context for single-rank

### LINKED ISSUE #1049: Distributed.context() required even in non-distributed inference setting
When developing [this notebook](https://colab.research.google.com/drive/1M6y9UBnb8pJLgO6PrJydfRogBFkdd2oL), I found that I had to use our `Distributed` context even when just loading the checkpoint (and ultimately, trying to do single GPU inference). The error message is also confusing, because it references a "training process" even though I was just trying to do inference. This makes our model harder to use, so it would be better if we did not require using this context.

Code:
```
from fme.ace.stepper import load_stepper
stepper = load_stepper("ace2_era5/ace2_era5_ckpt.tar")
```

Traceback:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
[/tmp/ipykernel_17023/543238903.py](https://localhost:8080/#) in <cell line: 0>()
      1 from fme.ace.stepper import load_stepper
      2 
----> 3 stepper = load_stepper("ace2_era5/ace2_era5_ckpt.tar")
      4 stepper.set_eval()
      5 print(f"Loaded model with {len(stepper.prognostic_names)} prognostic variables.")

5 frames
[/usr/local/lib/python3.12/dist-packages/fme/core/distributed/distributed.py](https://localhost:8080/#) in get_instance(cls)
    106         """
    107         if not cls._entered:
--> 108             raise RuntimeError(
    109                 "Distributed.get_instance() called before entering context. "
    110                 "Please use Distributed.context() to wrap the training process, "

RuntimeError: Distributed.get_instance() called before entering context. Please use Distributed.context() to wrap the training process, or ensure that get_instance() is only called within the context.
```

mcgibbon:
Single-rank environments like Jupyter notebooks previously could not call `Distributed.get_instance()` without wrapping code in `Distributed.context()`, even though the context manager is only needed for multi-rank distributed shutdown coordination.

Changes:
- `fme.core.distributed.Distributed.get_instance`: moved the `_entered` check to after singleton creation and only raises when `world_size > 1`

- [ ] Tests added

Fixes #1049

### DIFF
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -104,12 +104,6 @@ class Distributed:
         - ``FME_DISTRIBUTED_BACKEND=none`` to force
           :class:`NonDistributed`
         """
-        if not cls._entered:
-            raise RuntimeError(
-                "Distributed.get_instance() called before entering context. "
-                "Please use Distributed.context() to wrap the training process, "
-                "or ensure that get_instance() is only called within the context."
-            )
         global singleton
         if singleton is None:
             backend_env = os.environ.get("FME_DISTRIBUTED_BACKEND")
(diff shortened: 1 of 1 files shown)


oliverwm1 (approved):
Update the PR description to link the existing open issue before merging. Otherwise looks good.

PR #1071, opened 2026-04-21
`DenoisingMoEPredictor`

AnnaKwa:
This adds a predictor class `DenoisingMoEPredictor` and related config classes. It uses different checkpoints for different parts of the denoising schedule.

A followup PR will update the various other parts of the code to allow this predictor to be used at generation time.

I also made a minor fix to the code checking for equality in `LatLonCoordinates`. Previously, `__eq__` would raise a runtime error instead of returning False if the size of compared tensors was different.


fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+    DiffusionModel,
+    ModelOutputs,
+    _repeat_batch_by_samples,
+)
+from fme.downscaling.requirements import DataRequirements
+from fme.downscaling.samplers import stochastic_sampler as edm_sampler
+
+
+@dataclasses.dataclass
+class DenoisingExpertCheckpointConfig:
```
frodre, line 20: Are the denoising experts saved with these values and they are used/verified upon initialization here?

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    One expert checkpoint and the sigma interval (inclusive) it handles.
+    """
+
+    checkpoint_config: CheckpointModelConfig
+    sigma_min: float
+    sigma_max: float
+
+
+def _validate_sigma_ranges(sigma_ranges: list[tuple[float, float]]) -> None:
```
frodre, line 30: Seems pretty complicated check if we are manually specifying this information in the configuration.  Perhaps we could introduce a list of sigma expert models and specify a single list of dividing levels (always enforced to be size `n - 1` where `n` is the number of models) that would have to satisfy this list of tuple constraints.

For example:
models: [m1, m2, m3]
sigma_split: [sigma1, sigma2]

m1: s <= sigma1
m2: sigma1 < s <= sigma2
m3: s > sigma 3

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+                "Sigma ranges must be contiguous: "
+                f"sigma_ranges[{i}] max ({sigma_ranges[i][1]}) must equal "
+                f"sigma_ranges[{i + 1}] min ({sigma_ranges[i + 1][0]}). "
+                "List ranges in ascending sigma order."
+            )
+
+
+def _validate_experts_compatible(experts: list[DiffusionModel]) -> None:
+    primary = experts[0]
+    for m in experts[1:]:
```
frodre, line 53: Seems like it could be in some kind of softer __eq__ check within the  `DiffusionModel` (doesn't check that weights match, just metadata)

frodre: Discussing w/ Claude, which could be a good option.  Code lives closer to the diffusion models, and I think these classes could just use the `!=` check.  Perhaps better in a separate PR.

A metadata class:
```
  @dataclasses.dataclass(frozen=True)
  class DiffusionModelMetadata:
      in_names: tuple[str, ...]
      out_names: tuple[str, ...]
      coarse_shape: tuple[int, int]
      downscale_factor: int
      sigma_data: float
      predict_residual: bool
      use_fine_topography: bool
      full_fine_coords: LatLonCoordinates
      has_static_inputs: bool
      static_inputs_coords: LatLonCoordinates | None
```

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+    Assumes the sigma ranges and corresponding modules are sorted by
+    sigma_min ascending.
+    """
+
+    def __init__(
+        self,
+        sigma_ranges: list[tuple[float, float]],
+        modules: list[torch.nn.Module],
+    ) -> None:
+        if len(sigma_ranges) != len(modules):
```
frodre, line 77: If this stores a breakpoint list, and a module list, the `__call__` could simplify to use `bisect_left` to determine the module to call.

```
  def __call__(self, x, x_lr, sigma):
      idx = bisect.bisect_left(self._breakpoints, float(sigma.item()))
      return self._modules[idx](x, x_lr, sigma)
```

fme/downscaling/predictors/serial_denoising.py
frodre: See comments below about potentially removing need for this configuration object.

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+    The overall schedule bounds are the minimum ``sigma_min`` and maximum ``sigma_max``
+    across all ranges.
+
+    Parameters:
+        denoising_expert_configs: One entry per expert (checkpoint + sigma range).
+        num_diffusion_generation_steps: EDM sampler step count for the full schedule.
+        churn: EDM sampler churn (stochasticity) for the full schedule. Default 0.0.
+    """
+
+    denoising_expert_configs: list[DenoisingExpertCheckpointConfig]
```
frodre, line 134: Using a sigma breakpoint list, we no longer need manual sorting of checkpoints.  Instead just check that list is sorted and monotonic.  

```
 @dataclasses.dataclass
  class DenoisingMoEConfig:
      checkpoint_configs: list[CheckpointModelConfig]
      sigma_breakpoints: list[float]   # exactly len(checkpoint_configs) - 1
      sigma_min: float                 # passed to EDM sampler
      sigma_max: float                 # passed to EDM sampler
      num_diffusion_generation_steps: int
      churn: float = 0.0
```

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -0,0 +1,310 @@
...
+    @property
+    def data_requirements(self) -> DataRequirements:
+        return self.denoising_expert_configs[0].checkpoint_config.data_requirements
+
+
+class DenoisingMoEPredictor:
+    """
+    Mixture of ``DiffusionModel`` experts, each used for part of the EDM sigma
+    schedule. Behaves like ``DiffusionModel`` for generation and patching.
+    """
```
frodre, line 165: There's a lot of overlap here w/ the underlying model class when all we really need to replace is the module used during generation.   As long as we validate the experts metadata like you're already doing, we can initialize a diffusion model with the correct "primary" values.  Not required, but could re-implement this as a subclass of `DiffusionModel` that just handles using the dispatched module like a `torch.nn.Module` and not have to rewrite all the generation.  Something like this.

```
  class MoEDiffusionModel(DiffusionModel):
      @classmethod
      def from_experts(
          cls,
          experts: list[DiffusionModel],
          sigma_breakpoints: list[float],
          sigma_min: float,
          sigma_max: float,
          churn: float,
          num_diffusion_generation_steps: int,
      ) -> "MoEDiffusionModel":
          _validate_experts_compatible(experts)
          dispatch = _SigmaDispatchModule(sigma_breakpoints, [e.module for e in experts])
          primary = experts[0]
          moe_config = dataclasses.replace(
              primary.config,
              sigma_min=sigma_min,
              sigma_max=sigma_max,
              churn=churn,
              num_diffusion_generation_steps=num_diffusion_generation_steps,
          )
          return cls(
              config=moe_config,
              module=dispatch,
              normalizer=primary.normalizer,
              loss=primary.loss,
              coarse_shape=primary.coarse_shape,
              downscale_factor=primary.downscale_factor,
              sigma_data=primary.sigma_data,
              full_fine_coords=primary.full_fine_coords,
              static_inputs=primary.static_inputs,
          )
```

frodre (changes requested):
Thanks @AnnaKwa!  This is a solid start to adding feature in, but I had some reservations about the complexity for certain components.  I used claude to flesh those out and try and build examples of simplification ideas to iterate on.

I think the highest priority change I'd like to see is using a list of sigma breakpoints and a checkpoint model list instead of tying each individual together in a single config object.  I don't think that will reduce any visibility of what's happening in the model and would simplify a lot of the validation/logic.

I also think moving the equivalence validation into the `DiffusionModel` itself (perhaps via a metadata class as show) would be good to make sure it's tightly coupled to co-located code.

A larger change would be refactoring the MoEPredictor as a subclass of a diffusion model given the duplication, but open to discussion on whether or not this is a good idea, and not required for this PR to merge.

AnnaKwa:
As discussed in-person, I added the update to use a metadata class for comparison and am leaving the rest of the PR code as is for now.

fme/downscaling/models.py
```diff
@@ -688,6 +702,23 @@ def from_state(
         model.module.load_state_dict(state["module"], strict=True)
         return model
 
+    @property
+    def metadata(self):
```
frodre, line 708: nit: this recreates the object for every property call. Could be stored under `self._metadata` in `__post_init__` or use the [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) from functools

fme/downscaling/models.py
```diff
@@ -688,6 +702,23 @@ def from_state(
...
+            in_names=self.config.in_names,
+            out_names=self.config.out_names,
+            coarse_shape=self.coarse_shape,
+            downscale_factor=self.downscale_factor,
+            sigma_data=self.sigma_data,
+            predict_residual=self.config.predict_residual,
+            use_fine_topography=self.config.use_fine_topography,
+            full_fine_coords=self.full_fine_coords,
+            has_static_inputs=self.static_inputs is not None,
+            static_inputs_coords=self.static_inputs.coords
```
frodre: `static_inputs_coords` are, by construction, the same as `full_fine_coords` and `has_static_inputs` covers whether they exist.  We could add a length parameter to ensure they have the same number of static_inputs since we don't store name information.

fme/downscaling/models.py
```diff
@@ -122,6 +122,20 @@ def load(self):
         self.coarse.load()
 
 
+@dataclasses.dataclass(frozen=True)
```
frodre, line 126: Small thing flagged by Claude, which prompted more about the `frozen=True`. This flag ensures that these attrs can't be changed after initialization, but also means the `__hash__` method will be automatically created (because `eq=True` by default in the init of a dataclass).  We have a few fields which are unhashable, which would create some runtime failures if a hash is ever attempted.  Explicitly setting the class attribute `__hash__ = None` is the suggested remedy, especially since we rarely use this.

frodre (approved):
A couple minor nits/ideas but otherwise LGTM!  Added issues related to ideas discussed in person.

fme/downscaling/models.py
AnnaKwa: Also removed `has_static_inputs`, new attr `num_static_inputs=0` if no static inputs.

PR #1075, opened 2026-04-22
Set seed in `fme.coupled` training integration test

jpdunc23:
Sets the seed to avoid possible flakiness when the atmosphere component is stochastic.


mcgibbon (approved):
LGTM, at least shouldn't hurt anything to be setting the seed.

PR #1076, opened 2026-04-22
Put 3D variable name regex into a function for reuse

### LINKED ISSUE #605: Consolidate definitions of `LEVEL_PATTERN` in `fme.core`
Our `LEVEL_PATTERN` regular expression [is defined in 5 separate places currently](https://github.com/search?q=repo%3Aai2cm%2Ffull-model%20LEVEL_PATTERN&type=code).  It would be good to consolidate these into a single definition somewhere in `fme.core` that could be imported where needed.

xref: https://github.com/ai2cm/full-model/pull/1803#pullrequestreview-2794911053

nknish:
Changes:
- `parse_3d_varname`: regex that checks if a variable name is 3D, i.e. has a level number, returning the name and level number split apart if so and `None` if not.
  - Implementation consolidated from a few existing versions scattered throughout the codebase. All previous implementations are replaced with a call to this function
  - Added a small test

Is `fme.core.wildcard` a good place for this to live?

- [x] Tests added

Resolves #605


fme/core/step/_multi_call.py
```diff
@@ -37,10 +36,10 @@ def get_multi_call_name(name: str, suffix: str) -> str:
...
        'bar_with_quartered_co2_0'
     """
-    match = LEVEL_PATTERN.search(name)
-    if match is not None:
-        name = name[: match.start()]
-        suffix = suffix + match.group(0)
+    parsed = parse_3d_varname(name)
+    if parsed is not None:
+        name = parsed[0]
+        suffix = suffix + "_" + str(parsed[1])
```
mcgibbon, line 42: Issue: `parsed[0]` and `parsed[1]` are a little hard to understand when reading the code.

Suggestion: Have `parse_3d_varname` return a new type `ParsedVerticalName` (or None), which you can define as a `@dataclasses.dataclass` containing attributes `name: str` and `level: int`.

Suggestion (optional): Use `f"_{parsed.level}"`, f-strings are a very readable way to do string interpolation.

fme/core/wildcard.py
```diff
@@ -113,3 +113,14 @@ def apply_by_exclude(
         f"Skipped parameters: {set(model.state_dict().keys()) - applied_names}"
     )
     return model
+
+
+def parse_3d_varname(name: str) -> tuple[str, int] | None:
```
mcgibbon, line 118: Suggestion: Rename to something like `parse_vertical_varname`, as our contains-vertical data is not always 3D (for the healpix grid it's 4D, for example).

fme/core/wildcard.py
```diff
@@ -113,3 +113,14 @@ def apply_by_exclude(
...
+    """Return (name, level_num) from a 3D variable name, or None if not 3D."""
+    regex_pattern = re.compile(r"_(\d+)$")
```
mcgibbon, line 120: Issue: re.compile is expensive, it exists so that you can perform the compilation once when the code is imported and then the compiled regex pattern can do searches very quickly.

Suggestion: Store this on a `_LEVEL_PATTERN` file-local global variable.

fme/core/wildcard.py
mcgibbon, line 1: I don't think this really belongs in `wildcard.py`, which was made primarily to parse strings containing "*". You could put it in `stacker.py`, which is probably the most appropriate place it was already in, but that might cause circular imports. Another option is to make a new file `fme/core/vertical_names.py` for it.

fme/core/mask_provider.py
```diff
@@ -120,9 +118,9 @@ def get_mask_tensor_for(self, name: str) -> torch.Tensor | None:
         if mask_name in self.masks:
             return self.masks[mask_name]
         # level specific for 3D vars
-        match = LEVEL_PATTERN.search(name)
-        if match:
-            level = int(match.group(1))
+        parsed = parse_3d_varname(name)
+        if parsed:
```
mcgibbon, line 122: Suggestion: Use `if parsed is None` wherever you want to none-check, rather than the boolean-style check for None.

PR #1078, opened 2026-04-23
Log power spectrum diagnostics in inference entrypoint

### LINKED ISSUE #1077: Power spectrum diagnostics are not logged for rollouts with the inference entrypoint
We log power spectrum diagnostics for simulations with the evaluator entrypoint, but not the inference entrypoint.  We obviously do not have target data so we cannot compute error metrics in that context, but for various reasons it can be useful to have the raw spectra themselves. An example run can be found [here](https://wandb.ai/ai2cm/ace-shield/runs/2xrxtau4).

spencerkclark:
This PR implements what is necessary to add the `SphericalPowerSpectrumAggregator` to the `InferenceAggregator`, so that we can now log power spectra in runs launched with the inference entrypoint.

Changes:
- Adds `get_logs` and `get_dataset` methods to the `SphericalPowerSpectrumAggregator`.
- Adds `SphericalPowerSpectrumAggregator` to the `InferenceAggregator`.

- [x] Tests added

Resolves #1077

### DIFF
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -32,7 +32,10 @@ from .enso import (
 from .histogram import HistogramAggregator
 from .reduced import MeanAggregator, SingleTargetMeanAggregator
 from .seasonal import SeasonalAggregator
-from .spectrum import PairedSphericalPowerSpectrumAggregator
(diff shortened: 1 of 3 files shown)


jpdunc23 (approved):
LGTM!

PR #1079, opened 2026-04-23
Implement `get_dataset` for power spectrum aggregators

spencerkclark:
This PR implements `get_dataset` for the power spectrum aggregators to enable writing power spectra diagnostics to netCDF.  For now in the paired case we do not write the additional scalar metrics, since these can be computed from the raw spectra themselves.

Changes:
- Adds a `report_plot` option to the `SphericalPowerSpectrumAggregator` constructor to aid in writing faster tests.
- Implements meaningful `get_dataset` methods for the `SphericalPowerSpectrumAggregator` and `PairedSphericalPowerSpectrumAggregator` classes.
- Passes variable metadata through the `SphericalPowerSpectrumAggregator` and `PairedSphericalPowerSpectrumAggregator` constructors to enable assigning `"units"` and `"long_name"` attributes to the resulting datasets.

- [x] Tests added


fme/ace/aggregator/inference/spectrum.py
```diff
@@ -86,12 +109,13 @@ def __init__(
         gridded_operations: GriddedOperations,
         report_plot: bool,
         nan_fill_fn: Callable[[torch.Tensor, str], torch.Tensor] = lambda x, _: x,
+        variable_metadata: Mapping[str, VariableMetadata] | None = None,
     ):
         self._gen_aggregator = SphericalPowerSpectrumAggregator(
-            gridded_operations, nan_fill_fn
+            gridded_operations, nan_fill_fn, variable_metadata=variable_metadata
```
Arcomano1234: Optional: Do we want to hard code `report_plot=False` for the sub `SphericalPowerSpectrumAggregator`'s in `PairedSphericalPowerSpectrumAggregator`

Arcomano1234 (approved):
Left one optional comment, but overall LGTM.

fme/ace/aggregator/inference/spectrum.py
spencerkclark: Sure, probably no harm in doing that even though their `get_logs` method is not called anywhere in that context.

PR #1080, opened 2026-04-23
Update baseline Beaker budget to "atec-climate"

brianhenn:
Following [this announcement](https://docs.google.com/document/d/1AlQSb3dO2JcBEERu82Ix8Kn3D8bR9GKHNr7saqFIOLE/edit?tab=t.0) from beaker about new budgets, this PR updates the baseline configs to use the new budget name ("ai2/atec-climate"). Technically having this budget name isn't necessary because the workspaces will be assigned to a group with the right default budget, but in any case the old budget name won't work anymore. I verified that the new budget is valid on Beaker ([job](https://beaker.org/orgs/ai2/workspaces/ace/work/01KPXR15X6G0PMTWNW6BWWMW3T)).


mcgibbon (approved):
LGTM, thanks!

PR #1083, opened 2026-04-23
Decouple noise floor argo workflow from `fme`

spencerkclark:
This PR decouples the argo portion of the noise floor workflow from `fme` so that we can run it using the standard data processing image instead of an old `fme` image.  Old `fme` images cannot open zarr v3 stores, but new `fme` images do not have dask installed, which is necessary for this workflow.  The only coupling was in computing the area weights, which is straightforward to do without a special function.

The main motivation for this is to enable computing noise floor statistics for the `total_frozen_precipitation_rate`, which is only present in our newer intermediate zarr v3 stores.

Note I have not touched `generate_train_report.py`, which still depends on `fme`, but is typically run in a local environment, and does not require dask.

Changes:
- Computes the area weights within `scripts/noise_floor/generate_stats.get_output_datasets` so that we do not need to import a function from `fme`.
- Removes the need for the `fme.core.distributed.distributed.Distributed` context, since we are no longer calling anything in `fme`.
- Adds auto-chunking along the time dimension to ensure more performant reads from zarr v3 stores with inner chunk sizes of one.
- Updates the SHiELD-AMIP and SHiELD-SOM equilibrium-climate configurations to point to new reference datasets, adds `total_frozen_precipitation_rate` to the `names`, and updates the output `stats_path` to new locations for output from the re-run workflow for each.

Successful argo jobs (the longer one is the AMIP case):
```
compute-noise-floor-kgrsf            Succeeded   21h    5h         0
compute-noise-floor-vgbc7            Succeeded   21h    1h         0
compute-noise-floor-rp2r5            Succeeded   21h    1h         0
compute-noise-floor-vw7hs            Succeeded   21h    1h         0
compute-noise-floor-jrssh            Succeeded   21h    1h         0
```


Arcomano1234 (approved):
LGTM!

PR #1084, opened 2026-04-23
Remove unused fme/diffusion folder

oliverwm1:
This was previously used for experimental work using diffusion for forward stepping of ACE. This code has not been used in over a year, so this PR removes it, which reduces maintenance burden and removes some tests (including 1 slow integration test).


mcgibbon (approved):
I was planning to ask for permission to do this at our next technical sync. Approving, but may be good to announce there before committing, or at stand up today.

PR #1085, opened 2026-04-24
Fix/n ensemble attributes batch data

Arcomano1234:
This PR fixes some of the issues in the original implementation of `n_ensemble` in `BatchData`. The two main bugs were: 1.) some of the methods in `batch_data.py` were missing the `n_ensemble` priorities and 2.) there was a subtle CPU/ device issue that comes up during dataloading. Lastly, this PR extends `broadcast_ensemble()`to `PairedData`. 

Changes:
- Adds `n_ensemble` attributes to the appropriate methods in `BatchData`
- Fixes the device bug that broadcasts the ensembles during dataloading (previously always initializes on device)
- Adds `broadcast_ensemble()` to `PairedData`

- [x] Tests added


mcgibbon (approved):
LGTM. I wonder if there's a more general way to test that all batch data attributes are being tested.

Arcomano1234:
> LGTM. I wonder if there's a more general way to test that all batch data attributes are being tested.

Yeah I didn't look too hard at other tests but I assume we do something better for another tests right?

PR #1088, opened 2026-04-25
Clamp SSR calculation that was leading to NaNs and silently dropped from Wandb

### LINKED ISSUE #693: SSR ensemble metric not getting logged to Wandb correctly

Arcomano1234:
`SSRBiasMetric.get()` produces NaN values when the unbiased MSE correction overshoots to negative values (common with small ensembles or few batches), causing SSR to silently drop from WandB logs. This PR clamps the unbiased MSE before taking sqrt and returns a defined value (-1) when skill is zero, and also fixes the diverging colormap check in `_EnsembleAggregator` which was comparing against variable names instead of metric names.

Changes:
-` fme.ace.aggregator.one_step.ensemble.SSRBiasMetric.get`: clamp `total_unbiased_mse` to min=0.0 before sqrt to avoid NaN, return -1 via torch.where when skill is zero
-` fme.ace.aggregator.one_step.ensemble._EnsembleAggregator._get_data`: fix diverging colormap check to use metric in `self._diverging_metrics` instead of key in `self._diverging_metrics`

- [x] Tests added

Resolves #693


configs/experiments/2025-11-15-ace2s-x-shield/test-config.yaml
mcgibbon, line 1: Should clean this file up (delete it and revert the sh script) before merge.

fme/ace/aggregator/inference/test_main.py
```diff
@@ -132,28 +130,101 @@ def test_inference_evaluator_aggregator_ensemble():
...
     gen_data = BatchData.new_for_testing(
         names=["a", "b", "c"],
-        n_samples=1,
-        n_timesteps=n_timesteps,
+        n_samples=n_ensemble,
+        n_timesteps=n_forward_steps,
         img_shape=(nx, ny),
     )
-
-    target_data = target_data.broadcast_ensemble(n_ensemble=n_ensemble)
```
mcgibbon, line 149: Why did this test initialization logic get changed?

fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -134,8 +134,14 @@ def get(self) -> torch.Tensor:
         if self._total_unbiased_mse is None or self._total_variance is None:
             raise ValueError("No batches have been recorded.")
         spread = self._total_variance.sqrt()
-        skill = self._total_unbiased_mse.sqrt()
-        return spread / skill - 1
+        # Clamp to avoid NaN from sqrt of negative values. The unbiased MSE
```
mcgibbon, line 137: These changes make sense and seem good.

mcgibbon (commented):
Mostly LGTM, but clean up the submission directory, revert any unnecessary changes to tests, and update the PR description.

fme/ace/aggregator/one_step/ensemble.py
Arcomano1234: Yeah I agree and I believe the NaNs are the root of this whole problem

mcgibbon (approved):
Please update the PR title, otherwise LGTM

PR #1093, opened 2026-04-27
Add isotropic Morlet filter basis for DISCO

mcgibbon:
Add an isotropic Morlet filter basis for DISCO convolutions, enabling radially symmetric learned filters on the sphere. The standard Morlet basis includes azimuthal modes that break isotropy; the new basis uses only radial Fourier harmonics windowed by a Hann envelope, guaranteeing that any learned linear combination depends only on geodesic distance.

Changes:
- `fme.core.disco.IsotropicMorletFilterBasis` and `get_filter_basis` factory registration for `"isotropic morlet"` basis type
- `fme.core.models.conditional_sfno.sfnonet._compute_cutoff_radius` extended with `"isotropic morlet"` entry
- Test verifying isotropic DISCO conv commutes with latitude reflection while anisotropic Morlet does not

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/sfnonet.py
+++ b/fme/core/models/conditional_sfno/sfnonet.py
@@ -119,6 +119,7 @@ def _compute_cutoff_radius(nlat, kernel_shape, basis_type):
     theta_cutoff_factor = {
         "piecewise linear": 0.5,
         "morlet": 0.5,
+        "isotropic morlet": 0.5,
         "zernike": math.sqrt(2.0),
     }
 
--- a/fme/core/models/conditional_sfno/test_localnet.py
+++ b/fme/core/models/conditional_sfno/test_localnet.py
@@ -9,6 +9,7 @@ from fme.core.testing.regression import validate_tensor
 
 from .layers import Context, ContextConfig
 from .localnet import LocalNetConfig, get_lat_lon_localnet
+from .sfnonet import DiscreteContinuousConvS2, _compute_cutoff_radius
 
 DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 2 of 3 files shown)


fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -119,6 +119,7 @@ def _compute_cutoff_radius(nlat, kernel_shape, basis_type):
     theta_cutoff_factor = {
         "piecewise linear": 0.5,
         "morlet": 0.5,
+        "isotropic morlet": 0.5,
```
Arcomano1234, line 122: Question: I assume isotropic morlet should have the same cutoff factor as regular morlet?

fme/core/models/conditional_sfno/test_localnet.py
```diff
@@ -265,6 +266,71 @@ def test_can_call_localnet_with_kernel_shape(kernel_shape, basis_type):
     assert output.shape == (n_samples, output_channels, *img_shape)
 
 
+def test_isotropic_disco_conv_commutes_with_latitude_reflection():
```
Arcomano1234, line 269: Nice test to actually see if it's isotropic!

Arcomano1234 (approved):
It is a little hard for me to know for sure this implementation is correct but the `test_isotropic_disco_conv_commutes_with_latitude_reflection ` test seems quite reasonable and it is passing it.

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Yep, the idea is it's the same as morlet but doesn't include the non-isotropic filters.

PR #1095, opened 2026-04-27
Add time_buffer_pool_size for decorrelated sliding window batches

mcgibbon:
When `time_buffer > 0`, consecutive training batches are highly correlated because all sub-windows come from the same pre-loaded time window. Adding `time_buffer_pool_size` holds multiple windows in memory simultaneously and interleaves output batches across them, reducing this correlation while preserving data loading throughput. The implementation uses a pre-computed deterministic schedule so that checkpoint resume via `subset()` is fully supported.

For `pool_size=1` (the default), behavior is identical to the previous implementation — backward compatible with existing configs and checkpoints.

**Note: seed-breaking change for shuffled configs.** With `shuffle=True`, the offset permutation now uses `numpy.random.default_rng` instead of `np.argsort(rng.random(...))`, so exact batch sequences differ for the same seed. Statistical properties (uniform random offset selection within each input batch) are equivalent, so this should not affect training quality.

Changes:
- `fme.ace.data_loading.config.DataLoaderConfig`: added `time_buffer_pool_size` field (default 1, backward compatible), with validation in `__post_init__`
- `fme.ace.data_loading.pool`: new module containing `PooledSequence` (generic pool-based iterator) and `build_pool_schedule` (deterministic schedule generator)
- `fme.ace.data_loading.dataloader.SlidingWindowDataLoader`: rewritten to delegate iteration and subsetting to `PooledSequence`, with subset bounds stored directly rather than reaching into pool internals
- `fme.ace.data_loading.getters.get_gridded_data`: passes `time_buffer_pool_size` through to `get_data_loader`

- [x] Tests added (unit tests in `test_pool.py`, integration tests for `pool_size > 1` in `test_subset.py`)


fme/ace/data_loading/dataloader.py
```diff
@@ -164,135 +228,127 @@ def __init__(
...
+
+        if self._slice_last_use[input_idx] == self._schedule_pos:
+            del self._cache[input_idx]
+
+        self._schedule_pos += 1
         return small_batch
 
     @property
     def batch_size(self) -> int:
-        """
```
mcgibbon, line 272: Restore these deleted docstrings.

mcgibbon: Pre-review agent: Addressed — the init docstring was restored in a later commit and is present in the squashed version. This conversation can be resolved.

mcgibbon:
Pre-review agent: Here are a few pointers for the reviewer:

**Key design decision**: The core idea is a pre-computed deterministic schedule (`build_pool_schedule` in `pool.py`). The schedule is an `(N, 2)` array of `(input_batch_idx, sub_window_offset)` pairs. This makes checkpoint-resume via `subset()` trivially correct — it's just slicing the schedule array and subsetting the source to only load the needed input batches.

**Review focus areas**:
- `pool.py:build_pool_schedule` (~60 lines) — the schedule generation logic. Worth verifying: (1) the round-robin `robin` variable tracking after slot eviction, and (2) that every `(input_batch_idx, offset)` pair appears exactly once.
- `pool.py:PooledSequence.subset` — the most complex part. It narrows both the schedule window and the underlying source, and correctly computes `_source_start_idx` for items that need to be skipped during iteration.
- `pool.py:PooledSequence.__next__` — the `_use_counts` dict drives eviction: source items not in the active schedule slice are consumed from the iterator but never cached in the pool (line 173).

**Backward compat**: `pool_size=1` + `shuffle=False` produces the same sequential `(0,0),(0,1),...,(1,0),(1,1),...` order as the old code.

**Cleanup commit**: Removed the now-unused `_n_samples_per_dataset_item` method.

mcgibbon:
This will help us train much faster on 4-degree datasets.

fme/ace/data_loading/dataloader.py
```diff
@@ -301,27 +264,24 @@ def subset(
...
+                self._subset_stop
+                if self._subset_stop is not None
+                else self._total_input_batches * self._n_new_batches
+            )
+            new_stop: int | None = min(self._subset_start + stop_batch, effective_stop)
+        else:
+            new_stop = self._subset_stop
         loader = SlidingWindowDataLoader(
-            cast(TorchDataLoader, subsetted_loader),
+            self._loader,
```
jpdunc23, line 277: Issue: `SlidingWindowDataLoader.__init__` calls `self._loader.set_epoch(0)`, whereas previously the argument passed as `loader` to `SlidingWindowDataLoader` was a different object from `self._loader`. There is no bug in current usage, as far as I can tell, but it would be best to avoid shared state between original and subsetted `SlidingWindowDataLoader`, if possible.

fme/ace/data_loading/pool.py
jpdunc23, line 1: For reference, no action required.

<details>
<summary>Agent testing of `PooledSequence.subset` + Source Subsetting</summary>
Here is my evaluation.

## Evaluation of `PooledSequence.subset` + Source Subsetting

### Overview

`PooledSequence.subset(start, stop)` must accomplish three things:
1. **Narrow the schedule** to only positions `[start:stop]` (relative to the current view)
2. **Narrow the source** so only the needed input items are loaded from disk
3. **Track which global source index the narrowed source starts at**, so `__next__` can map items from `iter(source)` to the correct schedule positions

### How it works

Taking the shuffled schedule (pool_size=2, n_input=4, n_sub=3) as a concrete example:

```
pos=0: src=1,off=1    pos=6: src=2,off=2
pos=1: src=1,off=0    pos=7: src=2,off=0
pos=2: src=0,off=0    pos=8: src=3,off=2
pos=3: src=0,off=1    pos=9: src=3,off=1
pos=4: src=0,off=2    pos=10: src=3,off=0
pos=5: src=2,off=1    pos=11: src=1,off=2
```

If we call `subset(start=3, stop=9)`, the subset sees schedule positions 3..8:

```
(0,1), (0,2), (2,1), (2,2), (2,0), (3,2)
```

The method computes:
- `_use_counts = {0: 2, 2: 3, 3: 1}` — source 1 is absent from this slice
- `min_source_idx = 0`, `max_source_idx = 3`
- Source subsetted to `[0:4]` (items 0 through 3)
- `_source_start_idx = 0`

During iteration of the subset, `__next__` walks `iter(source)` sequentially. When it encounters source item 1, it checks `1 in _use_counts` → False → skips it (doesn't pool it). This is the key mechanism: items in the source range but not in the schedule slice are consumed from the iterator and discarded.

### Verified properties

I ran **8,480 test configurations** covering:
- n_items in {4, 8, 12}, n_sub in {2, 3, 5}, pool_size in {1, 2, 3, 4}
- Both shuffle=True and shuffle=False, seeds 0-4
- Various start/stop boundaries including mid-window, cross-boundary, and near-end positions

**All passed. Three properties verified:**

1. **Correctness:** `list(pooled.subset(start, stop))` always equals `list(pooled)[start:stop]`. This holds for single subsets, nested subsets (subset-of-subset), and triple-nested subsets.

2. **Pool size invariant holds:** Maximum pool occupancy during subset iteration never exceeds `pool_size`. For every `(pool_size, shuffle)` combination, the observed maximum exactly equaled `pool_size` (i.e., the pool is fully utilized but never exceeded). The `assert len(self._pool) < self._pool_size` on line 174 is safe.

3. **Source is narrowed to `[min_source_idx : max_source_idx+1]`:** The source is sliced to the contiguous range covering all source indices in the schedule slice.

### One noted inefficiency (not a bug)

The source is narrowed to the *contiguous range* of needed indices, not just the *set* of needed indices. In 889 of 8,480 cases, the subsetted source contained items not actually used by the schedule (they get iterated and discarded). The worst case: 8 wasted items (n=12, sub=2, pool_size=4, shuffle=True) where the schedule slice references source indices {4, 12} but the source contains items 4 through 12.

This is inherent to the `Subsettable` protocol which only supports contiguous `[start:stop]` ranges. It's a deliberate design choice — the real `TorchDataLoader` source doesn't support non-contiguous subsetting, and the wasted items are just consumed-and-discarded from the iterator (not held in memory via the pool). The `_use_counts` filter on line 173 ensures only items that will actually be used enter the pool.

### Nested subset correctness

The trickiest part is `subset()` of a `subset()`. When the outer subset has `_schedule_start=3, _schedule_stop=9, _source_start_idx=0`, and an inner `subset(start=2, stop=4)` is called:

- `new_start = 3 + 2 = 5`, `new_stop = min(3 + 4, 9) = 7`
- Schedule positions 5..6: `(2,1), (2,2)` — only source 2 needed
- `min_source_idx = 2`, `relative_start = 2 - 0 = 2`, `relative_stop = 3`
- Source subsetted to `outer_source[2:3]` — just one item
- `_source_start_idx = 2`

During iteration, `__next__` starts `_next_source_idx = 2`, gets the one source item, and correctly maps it. This all checks out — 18 nested-subset configurations were verified.

### Summary

The implementation is correct. The source subsetting logic properly handles:
- Schedule slices where not all source indices in the contiguous range are used (skip mechanism via `_use_counts`)
- Nested subsets with correct `_source_start_idx` propagation
- The pool size invariant under all tested configurations
</details>

fme/ace/data_loading/test_pool.py
jpdunc23, line 1: **Prompt**: Based on your testing, is there one or two additional key test cases we should add that are not already sufficiently covered by the existing tests?

> Looking at what the existing tests in `test_pool.py` cover versus what my extended testing revealed, the main gaps are:
> 
> Pool size invariant is never explicitly tested for subsets. The existing `test_cache_eviction` checks max pool occupancy for full iteration, but no test verifies that the pool invariant holds when iterating a subset (especially with `shuffle=True`, where source items must be skipped without pooling). This is the most critical untested property.
> 
> Source narrowing efficiency is untested. No test verifies that `subset()` actually narrows the source rather than passing through the entire dataset. With large GPU-resident batches this matters — a regression here would silently load and transfer unnecessary data.

Suggested additional tests:

```python
@pytest.mark.parametrize("pool_size", [2, 3])
@pytest.mark.parametrize("shuffle", [True, False])
@pytest.mark.parametrize("start", [0, 3, 5, 7])
def test_subset_pool_invariant(pool_size, shuffle, start):
    """Pool occupancy must never exceed pool_size when iterating a subset.

    This guards against a regression where source items outside the schedule
    slice are loaded into the pool because ``__next__`` iterates the source
    sequentially and must skip items not in ``_use_counts``.
    """
    pooled = make_pooled(6, 3, pool_size, shuffle=shuffle, seed=0)
    sub = pooled.subset(start=start)
    iter(sub)
    max_pool = 0
    for _ in sub:
        max_pool = max(max_pool, len(sub._pool))
    assert max_pool <= pool_size


@pytest.mark.parametrize("pool_size", [2, 3])
@pytest.mark.parametrize("shuffle", [True, False])
def test_subset_narrows_source(pool_size, shuffle):
    """Subsetting should narrow the source to only the needed range,
    not pass through the entire original source."""
    n_items, n_sub = 8, 3
    pooled = make_pooled(n_items, n_sub, pool_size, shuffle=shuffle, seed=0)
    total = n_items * n_sub
    sub = pooled.subset(start=total // 2)
    assert len(sub._source) < n_items
```

jpdunc23: Prompt: What about for `build_pool_schedule` correctness?

> The critical untested property is the pool occupancy invariant of the schedule itself — the guarantee that at most pool_size items are "alive" (have appeared but not yet been fully consumed) at any point in the schedule. This is what makes the `assert len(self._pool) < self._pool_size` in `__next__` safe. It's indirectly tested via `test_cache_eviction`, but that tests the iterator rather than the schedule directly, and only for `shuffle=False`.
> 
> The `covers_all_pairs` property is also only tested for a single configuration `(n=5, sub=3, pool=2, shuffle=False)`.

Suggested expansion of `test_cache_eviction` [here](https://github.com/ai2cm/ace/pull/1095/changes#r3229841174).

Suggested expansion of `test_build_pool_schedule_covers_all_samples` [here](https://github.com/ai2cm/ace/pull/1095/changes#r3229806441).

fme/ace/data_loading/test_pool.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def test_build_pool_schedule_covers_all_samples():
+    schedule = build_pool_schedule(
+        n_input_batches=5, n_sub_batches=3, pool_size=2, shuffle=False, seed=0
+    )
+    assert schedule.shape == (15, 2)
+    pairs = set(map(tuple, schedule.tolist()))
+    expected = {(i, j) for i in range(5) for j in range(3)}
+    assert pairs == expected
```
jpdunc23:
```suggestion
@pytest.mark.parametrize("n_input,n_sub,pool_size", [
    (4, 3, 1), (4, 3, 2), (4, 3, 4),
    (5, 3, 2), (8, 5, 3), (10, 2, 4), (3, 6, 2), (1, 5, 1), (6, 1, 3),
])
@pytest.mark.parametrize("shuffle", [True, False])
def test_build_pool_schedule_covers_all_samples(n_input, n_sub, pool_size, shuffle):
    """Every (input_idx, offset) pair appears exactly once."""
    schedule = build_pool_schedule(
        n_input, n_sub, pool_size, shuffle=shuffle, seed=7
    )
    assert schedule.shape == (n_input * n_sub, 2)
    pairs = set(map(tuple, schedule.tolist()))
    assert pairs == {(i, j) for i in range(n_input) for j in range(n_sub)}
```

fme/ace/data_loading/test_pool.py
```diff
@@ -0,0 +1,247 @@
...
+
+@pytest.mark.parametrize("pool_size", [1, 2, 3])
+def test_cache_eviction(pool_size):
+    pooled = make_pooled(n_items=6, n_sub=3, pool_size=pool_size, shuffle=False)
+    iter(pooled)
+    max_pool_size = 0
+    for _ in pooled:
+        max_pool_size = max(max_pool_size, len(pooled._pool))
+    assert max_pool_size <= pool_size
+    assert len(pooled._pool) == 0
```
jpdunc23, line 162:
```suggestion
@pytest.mark.parametrize("pool_size", [1, 2, 3])
@pytest.mark.parametrize("shuffle", [True, False])
def test_cache_eviction(pool_size, shuffle):
    pooled = make_pooled(n_items=6, n_sub=3, pool_size=pool_size, shuffle=shuffle)
    iter(pooled)
    max_pool_size = 0
    for _ in pooled:
        max_pool_size = max(max_pool_size, len(pooled._pool))
    assert max_pool_size <= pool_size
    assert len(pooled._pool) == 0
```

jpdunc23 (commented):
One issue to fix, defensively: shared `loader` state in parent and subset `SlidingWindowDataLoader` (see [here](https://github.com/ai2cm/ace/pull/1095/changes#r3229482853)). 

See suggestions below. Otherwise, looks good.

> 1. **`build_pool_schedule` correctness** (as the author's own pre-review comment suggests) — specifically verifying that every `(input_batch_idx, offset)` pair appears exactly once and that the pool invariant holds. I verified both of these empirically and they check out.

Agent review initially thought there was a bug here, but after unsuccessfully attempting to add a failing test, verified empirically that this is working as expected. Suggested additional test [here](https://github.com/ai2cm/ace/pull/1095/changes#r3229757323).

> 2. **`PooledSequence.subset` + source subsetting** — the most intricate logic in the PR. The interaction between schedule slicing, `_use_counts` recomputation, and source narrowing is correct but dense; worth a careful read.

Comprehensively tested by agent, see [this comment](https://github.com/ai2cm/ace/pull/1095/changes#r3229660168) for details. Suggested additional tests to add [here](https://github.com/ai2cm/ace/pull/1095/changes#r3229725926).

> 3. **Backward compatibility** — `pool_size=1` with `shuffle=False` produces identical ordering to the old code. With `shuffle=True`, the offset permutation uses a different RNG path (numpy `default_rng` vs the old `np.argsort(rng.random(...))`) so exact sequences differ, but the statistical properties (uniform random offset selection within each input batch) are equivalent. This is a **seed-breaking change for existing shuffled configs**, which is worth noting even though it shouldn't affect training quality.

Suggestion: Note the seed-breaking change in the PR description.

fme/ace/data_loading/pool.py
mcgibbon: > One noted inefficiency (not a bug)

Yes, this is a necessary inefficiency resulting from our choice of shuffling. Unfortunately the only way to make it more "efficient" is to reduce the pool size or make the selection cross-pool less random, both of which we don't want to do. So there's nothing to do here.

fme/ace/data_loading/test_pool.py
mcgibbon: > Pool size invariant is never explicitly tested for subsets. The existing test_cache_eviction checks max pool occupancy for full iteration, but no test verifies that the pool invariant holds when iterating a subset (especially with shuffle=True, where source items must be skipped without pooling). This is the most critical untested property.

Yes, this test should be added.

> Source narrowing efficiency is untested. No test verifies that subset() actually narrows the source rather than passing through the entire dataset. With large GPU-resident batches this matters — a regression here would silently load and transfer unnecessary data.

I thought this is already covered by the explicit subset tests, which show examples where subset leads to a smaller set of outputs, and those outputs are tested for equality in a way that guarantees those outputs aren't duplicated (so the size can't be larger). When I looked I was satisfied there were enough regression cases that this property is appropriately tested, at the least adding a test of this check with two random pools doesn't provide significant additional coverage.

fme/ace/data_loading/test_pool.py
mcgibbon: Doesn't the usage of `set` remove duplicates in a way that prevents us from testing that the pairs appear exactly once? I think we want them sorted instead?

fme/ace/data_loading/test_pool.py
mcgibbon: Looks good.

fme/ace/data_loading/dataloader.py
mcgibbon: Yes, let's avoid coupling the state if we can.

mcgibbon: Claude: Fixed — extracted `_init_epoch_state` from `set_epoch` so that `subset()` and `__init__` no longer call `set_epoch` on the shared `TorchDataLoader`. The subset now only updates its own internal epoch state via `_init_epoch_state`, and only the owning `SlidingWindowDataLoader` propagates epoch changes to the loader.

fme/ace/data_loading/test_pool.py
mcgibbon: Claude: Applied the expanded parametrize cases. Used `sorted()` instead of `set()` per mcgibbon's feedback, so duplicates would cause a length mismatch — verifying that each pair appears *exactly* once.

fme/ace/data_loading/test_pool.py
mcgibbon: Claude: Applied — also added `shuffle` parametrize.

fme/ace/data_loading/test_pool.py
mcgibbon: Claude: Added `test_subset_pool_invariant` — parametrized over pool_size, shuffle, and start offset. Verifies pool occupancy stays within `pool_size` when iterating subsets.

jpdunc23 (approved):
LGTM.

PR #1096, opened 2026-04-28
Replace weather eval with additional inference

mcgibbon:
Replaces the single optional `weather_evaluation` config with a list of named `additional_inference` entries, each able to run an independent inference evaluation during training. Also fixes two bugs in the existing weather evaluation code path.

Changes:
- `TrainBuilders.get_evaluation_inference_data`, `TrainBuilders.get_end_of_epoch_callback`: fix weather evaluation using inline inference's data requirements instead of its own
- `TrainBuilders.get_end_of_epoch_callback`: fix aggregator being built once and reused across epochs instead of rebuilt each time
- `WeatherEvaluationConfig`: removed, was identical to `InlineInferenceConfig`
- `TrainConfig.weather_evaluation` renamed to `TrainConfig.additional_inference`, changed from `InlineInferenceConfig | None` to `list[AdditionalInferenceConfig]`
- `AdditionalInferenceConfig`: new dataclass with `name` (used as wandb log prefix and output subdirectory) and `config: InlineInferenceConfig`, with duplicate name validation
- `fme.ace.AdditionalInferenceConfig`: added to public API

- [x] Tests added

### DIFF
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -39,67 +39,6 @@ from fme.core.weight_ops import CopyWeightsConfig
 
 
 @dataclasses.dataclass
-class WeatherEvaluationConfig:
-    """
-    Parameters:
-        loader: configuration for the data loader used during weather evaluation
-        n_forward_steps: number of forward steps to take
-        forward_steps_in_memory: number of forward steps to take before
-            re-reading data from disk
-        epochs: epochs on which to run weather evaluation. By default runs
-            weather evaluation every epoch.
-        aggregator: configuration of weather evaluation aggregator.
-    """
-
-    loader: InferenceDataLoaderConfig
-    n_forward_steps: int
-    forward_steps_in_memory: int
-    epochs: Slice = dataclasses.field(default_factory=lambda: Slice())
-    aggregator: InferenceEvaluatorAggregatorConfig = dataclasses.field(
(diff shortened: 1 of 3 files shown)


fme/ace/train/train_config.py
```diff
@@ -104,63 +99,13 @@ def get_inference_data(
...
-    n_forward_steps: int
-    forward_steps_in_memory: int
-    epochs: Slice = dataclasses.field(default_factory=lambda: Slice())
-    aggregator: InferenceEvaluatorAggregatorConfig = dataclasses.field(
-        default_factory=lambda: InferenceEvaluatorAggregatorConfig(
-            log_global_mean_time_series=False, log_global_mean_norm_time_series=False
-        )
-    )
+class AdditionalInferenceConfig:
+    name: str
```
mcgibbon, line 112: Wandb logs will be stored under `{name}/`, so this could be things like `weather_eval`, `inference_era5`, or `inf_era5_pre_industrial`. When I add a similar additional validation config, I'll need to check for duplicate names across both lists.

spencerkclark (approved):
Nice that you could repurpose / generalize this after fixing a few bugs.  Looks good to me!

One thing that this does not make easier yet is automating checkpoint selection based on inline inference targeting multiple datasets, but that would probably need more design discussion.

PR #1097, opened 2026-04-28
Uniform sub-aggregator interface via InferenceBatchData

mcgibbon:
Unify the inference sub-aggregator interface so that all sub-aggregators accept a single `InferenceBatchData` dataclass, replacing four separate Protocol variants. This eliminates the need for isinstance checks and prepares for configurable metrics in follow-up PRs.

Changes:
- `fme.ace.aggregator.inference.data.InferenceBatchData` new dataclass bundling prediction, target, time, and i_time_start
- `fme.ace.aggregator.inference.data.SubAggregator` new Protocol replacing `_Aggregator`, `_EvaluatorAggregator`, `_TimeDependentAggregator`, `_TimeDependentEvaluatorAggregator`
- `fme.ace.aggregator.inference.config` new module extracting `InferenceAggregatorConfig`, `InferenceEvaluatorAggregatorConfig`, `StepMeanEntry` from `main.py`
- `fme.ace.aggregator.inference.config._OneStepMeanAdapter` wraps `OneStepMeanAggregator` to accept `InferenceBatchData`
- `fme.ace.aggregator.inference.reduced.MeanAggregator`, `SingleTargetMeanAggregator` converted to `InferenceBatchData`
- `fme.ace.aggregator.inference.time_mean.TimeMeanAggregator`, `TimeMeanEvaluatorAggregator` converted to `InferenceBatchData`
- `fme.ace.aggregator.inference.spectrum.SphericalPowerSpectrumAggregator`, `PairedSphericalPowerSpectrumAggregator` converted to `InferenceBatchData`
- `fme.ace.aggregator.inference.video.VideoAggregator`, `histogram.HistogramAggregator`, `zonal_mean.ZonalMeanAggregator`, `seasonal.SeasonalAggregator` converted to `InferenceBatchData`
- `fme.ace.aggregator.inference.annual`, `enso` sub-aggregators converted to `InferenceBatchData`
- `fme.ace.aggregator.inference.main.InferenceEvaluatorAggregator` merged two aggregator dicts into one, using `_streaming_names` list instead of isinstance checks

- [x] Tests added

PR 1 of 3 in the inference aggregator refactor series (#1098, #1099).


fme/ace/aggregator/inference/main.py
```diff
@@ -296,13 +135,9 @@ def __init__(
         if save_diagnostics and output_dir is None:
             raise ValueError("Output directory must be set to save diagnostics")
         self._channel_mean_names = channel_mean_names
-        self._aggregators: dict[str, _EvaluatorAggregator] = {}
+        self._aggregators: dict[str, Any] = {}
```
mcgibbon: It would be ideal for this to retain some helpful typing information, does the value fit some kind of protocol class or ABC?

fme/ace/aggregator/inference/main.py
```diff
@@ -296,13 +135,9 @@ def __init__(
...
         self.n_ensemble_per_ic = n_ensemble_per_ic
-        if n_ensemble_per_ic > 1:
-            self._ensemble_aggregators: dict[str, _EvaluatorEnsembleAggregator] = {}
-        self._time_dependent_aggregators: dict[
-            str, _TimeDependentEvaluatorAggregator
-        ] = {}
+        self._ensemble_aggregators: dict[str, Any] = {}
```
mcgibbon: Similarly, can we improve the typing here?

fme/ace/aggregator/inference/config.py
```diff
@@ -0,0 +1,172 @@
...
+    time_mean_reference_data: str | None = None
+    log_global_mean_time_series: bool = True
+
+    def build(
+        self,
+        dataset_info: DatasetInfo,
+        n_timesteps: int,
+        output_dir: str,
+    ) -> InferenceAggregator:
+        from .main import InferenceAggregator
```
mcgibbon: Imports should exist at the top of the file. If these two files are tightly coupled preventing a top-of-file import, move the configuration class to where the code lives - we kind of think of the Config and its built class as one conceptual object, anyways.

fme/ace/aggregator/inference/config.py
```diff
@@ -0,0 +1,172 @@
...
+        n_ic_steps: int,
+        n_forward_steps: int,
+        initial_time: xr.DataArray,
+        normalize: Callable[[TensorMapping], TensorDict],
+        output_dir: str | None = None,
+        channel_mean_names: Sequence[str] | None = None,
+        save_diagnostics: bool = True,
+        n_ensemble_per_ic: int = 1,
+    ) -> InferenceEvaluatorAggregator:
+        from .main import InferenceEvaluatorAggregator
```
mcgibbon: Similar comment, import top-of-file.

fme/ace/aggregator/inference/config.py
```diff
@@ -0,0 +1,172 @@
...
+
+import dataclasses
+from collections.abc import Callable, Sequence
+from typing import TYPE_CHECKING
+
+import xarray as xr
+
+from fme.core.typing_ import TensorDict, TensorMapping
+
+if TYPE_CHECKING:
```
mcgibbon: We avoid using TYPE_CHECKING, if there's a circular import it implies an issue in how the code is organized. These configs may need to live with the code they configure. If you do decide to revert in that way, put the configs at the location in the file that minimizes the diff from main.

fme/ace/aggregator/inference/data.py
```diff
@@ -0,0 +1,38 @@
...
+
+    prediction: TensorMapping
+    prediction_norm: TensorMapping
+    target: TensorMapping | None
+    target_norm: TensorMapping | None
+    time: xr.DataArray | None
+    i_time_start: int
+
+
+@runtime_checkable
```
mcgibbon: Do we use this runtime checking, or can you remove the decorator?

fme/ace/aggregator/inference/main.py
```diff
@@ -529,31 +355,29 @@ def record_initial_condition(
...
                 f"Expected {self.n_ic_steps} initial condition steps, but got {n_times}"
             )
-        for aggregator_name in ["mean", "mean_norm"]:
-            aggregator = self._aggregators.get(aggregator_name)
+        batch = InferenceBatchData(
+            prediction=gen_data,
+            prediction_norm=self._normalize(gen_data),
+            target=target_data,
+            target_norm=self._normalize(target_data),
+            time=None,
```
mcgibbon: Do we not have time available? If we do, we should probably include it, to avoid not knowing its available in lower level code. If we can do that, we may also be able to make this a required argument on InferenceBatchData.

fme/ace/aggregator/inference/main.py
```diff
@@ -643,50 +449,12 @@ def to_inference_logs(
 
 
 def table_to_logs(table: Table) -> list[dict[str, float | int]]:
-    """
-    Converts a WandB table into a list of dictionaries.
-    """
```
mcgibbon, line 648: Why remove this docstring?

fme/ace/aggregator/inference/main.py
```diff
@@ -643,50 +449,12 @@ def to_inference_logs(
...
     logs = []
     for row in table.data:
         logs.append({table.columns[i]: row[i] for i in range(len(row))})
     return logs
 
 
-@dataclasses.dataclass
-class InferenceAggregatorConfig:
```
mcgibbon, line 656: The more I think about it, let's put these configs back where they were, if that's an option.

fme/ace/aggregator/inference/main.py
```diff
@@ -725,13 +493,15 @@ def __init__(
         self._coords = horizontal_coordinates.coords
         self._save_diagnostics = save_diagnostics
         self._output_dir = output_dir
-        aggregators: dict[str, _Aggregator] = {}
+        aggregators: dict[str, Any] = {}
```
mcgibbon: Can we avoid losing typing information? Any is not very helpful.

fme/ace/aggregator/inference/main.py
```diff
@@ -725,13 +493,15 @@ def __init__(
...
         gridded_operations = dataset_info.gridded_operations
+        self._streaming_names: list[str] = []
```
mcgibbon: I don't understand the meaning of "streaming" names, can this variable name be clearer?

fme/ace/aggregator/inference/main.py
```diff
@@ -840,37 +608,21 @@ def get_summary_logs(self) -> InferenceLog:
 
     @torch.no_grad()
     def _get_logs(self):
-        """
```
mcgibbon, line 843: Please restore deleted docstrings (including elsewhere in this PR), unless there was a good reason to delete them.

fme/ace/aggregator/inference/test_spectrum.py
```diff
@@ -32,8 +33,26 @@ def test_spherical_power_spectrum_aggregator(report_plot: bool):
...
-    agg.record_batch(data2)
+    agg.record_batch(
+        InferenceBatchData(
+            prediction=data,
+            prediction_norm={},
+            target=None,
+            target_norm=None,
+            time=None,
+            i_time_start=0,
+        )
```
mcgibbon, line 44: Can we reduce duplication/line count here and in other test files using a `wrap_as_inference_batch_data(data)` helper? Or using a new `InferenceBatchData.new_test_data` helper constructor method?

fme/ace/aggregator/inference/test_zonal_mean.py
```diff
@@ -13,6 +14,17 @@ def zonal_mean(data: torch.Tensor) -> torch.Tensor:
     return data.mean(dim=3)
 
 
+def _make_batch_data(target_data, gen_data, i_time_start=0) -> InferenceBatchData:
```
mcgibbon, line 17: Good pattern!

mcgibbon:
From Claude:

❯ I'm looking at PR 1097. Can you summarize for me the main "improvement" of this PR? Why is this change
  necessary?

● The core improvement is giving all inference sub-aggregators a uniform interface via InferenceBatchData.
   Before this PR, different sub-aggregators accept different combinations of positional/keyword arguments
   — some take target_data, gen_data, others take target_data, gen_data, target_data_norm, gen_data_norm,
  i_time_start, etc. The coordinator (InferenceEvaluatorAggregator) has to know each sub-aggregator's
  specific calling convention, which makes the loop in record_batch a mess of special cases.

  After this PR, every sub-aggregator accepts a single InferenceBatchData object and pulls out what it
  needs. The coordinator just does for agg in aggregators: agg.record_batch(batch) — it doesn't need to
  know what each aggregator cares about.

  This is necessary groundwork for PR 1099 (configurable metrics). If you want users to be able to declare
   arbitrary metrics in a config list and have the system build/wire them automatically, the coordinator
  can't have per-aggregator-type calling logic — it needs a single interface it can call uniformly on
  whatever aggregators the config produces.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done. Typed as `dict[str, SubAggregator]` using the new `SubAggregator` protocol from `data.py`.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done. Typed as `dict[str, SelectStepEnsembleAggregator]`.

fme/ace/aggregator/inference/config.py
mcgibbon: Claude: Moved all configs back to `main.py` at their original positions and deleted `config.py`. No more `TYPE_CHECKING` or late imports.

fme/ace/aggregator/inference/config.py
mcgibbon: Claude: Resolved by moving configs back to `main.py` — no more late imports.

fme/ace/aggregator/inference/config.py
mcgibbon: Claude: Resolved by moving configs back to `main.py`.

fme/ace/aggregator/inference/data.py
mcgibbon: Claude: Removed. It wasn't used for any runtime `isinstance` checks.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Good catch — time was available in both branches but not being passed. Fixed to pass `initial_condition.time` / `batch_data.time` in both `record_initial_condition` methods. Left `time` as optional on `InferenceBatchData` for now since making it required would touch many test files; happy to do that as a follow-up if you'd like.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: No good reason — restored this and all other deleted docstrings.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done. All three config classes are back in `main.py` at their original positions.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done. Typed as `dict[str, SubAggregator]`.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Replaced `_streaming_names: list[str]` with `_time_series_aggregators: dict[str, TimeSeriesLogs]`. This is a new protocol that captures the `get_logs(label, step_slice=...)` capability — the name-list was acting as a runtime stand-in for what's really a type-level distinction between aggregators that produce per-step time series logs and those that only produce summaries.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Restored all deleted docstrings.

fme/ace/aggregator/inference/test_spectrum.py
mcgibbon: Claude: Added `InferenceBatchData.new_test_data()` classmethod that generates random tensors from a shape and variable names. Existing tests that need specific tensor values still use the regular constructor directly.

fme/ace/aggregator/inference/main.py
mcgibbon: Please make it required, reduce subsequent test-touching if possible by having the testing new constructor automatically handle time with a reasonable default.

fme/ace/aggregator/inference/main.py
```diff
@@ -792,16 +732,16 @@ def record_batch(self, data: PairedData) -> InferenceLogs:
...
-                    time=data.time, data=data.prediction
-                )
-            else:
-                self._aggregators[name].record_batch(
-                    data=data.prediction,
-                    i_time_start=self._n_timesteps_seen,
-                )
+        batch = InferenceBatchData(
+            prediction=data.prediction,
+            prediction_norm=data.prediction,
```
Arcomano1234: Question: this seems potentially error prone or at best confusing to be passing raw predictions to `prediction_norm`. Is there a way to construct `InferenceBatchData` to have prediction_norm take a None or empty dictionary in such cases?

fme/ace/aggregator/inference/main.py
```diff
@@ -314,40 +271,48 @@ def __init__(
...
                 n_timesteps=n_timesteps,
                 variable_metadata=dataset_info.variable_metadata,
             )
+            self._aggregators["mean_norm"] = mean_norm_agg
+            self._time_series_aggregators["mean_norm"] = mean_norm_agg
         for step_mean_entry in log_step_means:
             step_mean_entry.validate(n_forward_steps)
             step = step_mean_entry.step
             name = step_mean_entry.get_name()
-            # -1 because step 0 (after IC) is the first forward step
```
Arcomano1234, line 335: Question: Is this deletion of the comment intentional it seems that behavior is still the same

fme/ace/aggregator/inference/spectrum.py
```diff
@@ -125,16 +127,30 @@ def __init__(
...
-        target_data: TensorMapping,
-        gen_data: TensorMapping,
-        target_data_norm: TensorMapping | None = None,
-        gen_data_norm: TensorMapping | None = None,
-        i_time_start: int = 0,
-    ):
-        self._gen_aggregator.record_batch(gen_data)
-        self._target_aggregator.record_batch(target_data)
+    def record_batch(self, data: InferenceBatchData):
+        self._gen_aggregator.record_batch(
```
Arcomano1234: nit: Is there any reason we can't construct InferenceBatchData before recording the in the same style as it's done in annual.py and other aggregators? E.g.,
```
        target_data = dataclasses.replace(data, prediction=data.target)
        gen_data = dataclasses.replace(data, prediction=data.prediction)
        self._target_aggregator.record_batch(target_data)
        self._gen_aggregator.record_batch(gen_data)
```

fme/ace/aggregator/one_step/spectrum.py
```diff
@@ -47,12 +48,16 @@ def record_batch(
...
-                gen_data,
-                target_data_norm,
-                gen_data_norm,
-                i_time_start,
+                InferenceBatchData(
+                    prediction=gen_data,
+                    prediction_norm=gen_data_norm,
+                    target=target_data,
+                    target_norm=target_data_norm,
+                    time=make_dummy_time(sample.shape[0], sample.shape[1]),
```
Arcomano1234: Question: Do we want `time` in `InferenceBatchData` to have a | None so that we don't need to construct this dummy_time?

Arcomano1234 (commented):
Left a few questions / comments but seems pretty much ready to go.

fme/ace/aggregator/inference/main.py
mcgibbon: Yeah seems like this should be restored.

fme/ace/aggregator/inference/main.py
mcgibbon: Yeah I think I agree, we should be passing None here.

fme/ace/aggregator/inference/spectrum.py
mcgibbon: Isn't `dataclasses.replace(data, prediction=data.prediction)` a no-op? But in any case, let's look at the other aggregators and see if there's a simpler style to use here.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Mmmm... maybe... though I feel a little strange making this API decision for InferenceBatchData based on its non-inference use. The problem is that while having a single unified input data type for all aggregators is nice, it does also mean they're less type-safe.

I wonder if there's a better solution here, where we could make a paired spherical power spectrum type or method that takes in only the specific arguments it needs, and then we could call that method here instead of `.record_batch`? Since For this object `self._wrapped` isn't a generic aggregator, it's a concrete type.

fme/ace/aggregator/inference/data.py
```diff
@@ -0,0 +1,81 @@
...
+        target: Denormalized target tensors, or None for single-series inference.
+        target_norm: Normalized target tensors, or None for single-series inference.
+        time: Datetime array of shape (sample, time).
+        i_time_start: Global timestep offset for this batch.
+    """
+
+    prediction: TensorMapping
+    prediction_norm: TensorMapping
+    target: TensorMapping | None
+    target_norm: TensorMapping | None
```
mcgibbon: We want to make prediction_norm optional. It's a bit tedious though having so many none-checks in the code. I would propose we make this type a non-dataclass, and set these attributes privately during init. Then have public property methods that raise an exception Unavailable(ValueError) with a descriptive message if these are accessed when they aren't defined. That way the public property methods are typed as non-none.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Restored.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done. `InferenceBatchData` is now a non-dataclass where `prediction_norm`, `target`, and `target_norm` are stored privately and accessed via properties that raise `Unavailable(ValueError)` with a descriptive message when not provided. `InferenceAggregator` now passes `prediction_norm=None` instead of the raw predictions.

fme/ace/aggregator/inference/spectrum.py
mcgibbon: Claude: Simplified. `PairedSphericalPowerSpectrumAggregator` now has a `record_paired_data(prediction, target)` method, and `SphericalPowerSpectrumAggregator` has `record_data(prediction)`. The `record_batch` methods delegate to these, and internal calls go through the direct methods instead of constructing intermediate `InferenceBatchData` objects.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Claude: Agreed — instead of making `time` optional on `InferenceBatchData`, I added `record_paired_data(prediction, target)` to `PairedSphericalPowerSpectrumAggregator`. The one_step `SpectrumAggregator` now calls that method directly with the tensor dicts it already has, so it no longer needs to construct an `InferenceBatchData` or use `make_dummy_time`.

fme/ace/aggregator/inference/data.py
mcgibbon: Claude: Done. `InferenceBatchData` is now a non-dataclass. `prediction_norm`, `target`, and `target_norm` are accepted as optional `__init__` parameters (defaulting to `None`), stored privately, and exposed via properties typed as non-`None` that raise `Unavailable(ValueError)` with a descriptive message if accessed when not provided. Added `has_target`, `has_prediction_norm`, `has_target_norm` for conditional checks, and a `.replace()` method for creating modified copies (replacing the former `dataclasses.replace` usage).

Arcomano1234 (approved):
This looks ready to merge, thanks for addresses the comments.

PR #1098, opened 2026-04-28
Thin coordinator: move builder logic to config.build()

mcgibbon:
Move all sub-aggregator construction logic from aggregator constructors into `config.build()` methods, making `main.py` a thin coordinator loop. Adding new sub-aggregators now only requires changes in `config.py`.

Changes:
- `fme.ace.aggregator.inference.config.InferenceEvaluatorAggregatorConfig.build()` now handles all sub-aggregator construction, reference data loading, and conditional logic
- `fme.ace.aggregator.inference.config.InferenceAggregatorConfig.build()` same treatment for the non-evaluator aggregator
- `fme.ace.aggregator.inference.main.InferenceEvaluatorAggregator.__init__`, `InferenceAggregator.__init__` thinned to accept pre-built aggregator dicts
- `fme.ace.aggregator.one_step.spectrum.PairedSphericalPowerSpectrumAggregator` new wrapper adapting `InferenceBatchData` interface for training use
- `fme.coupled.aggregator` updated to build ace aggregators via ace config.build(), removing duplicated reference data loading

- [x] Tests added

PR 2 of 3 in the inference aggregator refactor series (based on #1097, followed by #1099).

### DIFF
--- a/fme/ace/aggregator/one_step/spectrum.py
+++ b/fme/ace/aggregator/one_step/spectrum.py
@@ -1,12 +1,40 @@
 from collections.abc import Callable
+from typing import Any
 
 import numpy as np
 import torch
 import xarray as xr
 
-from fme.ace.aggregator.inference.spectrum import PairedSphericalPowerSpectrumAggregator
+from fme.ace.aggregator.inference.data import InferenceBatchData
+from fme.ace.aggregator.inference.spectrum import (
+    PairedSphericalPowerSpectrumAggregator as _InferencePairedSpectrumAgg,
+)
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.typing_ import TensorMapping
+
+
+class PairedSphericalPowerSpectrumAggregator:
+    """Wraps the inference PairedSphericalPowerSpectrumAggregator for one-step use."""
+
+    def __init__(self, **kwargs: Any):
+        self._inner = _InferencePairedSpectrumAgg(**kwargs)
+
+    def record_batch(self, target_data: TensorMapping, gen_data: TensorMapping) -> None:
+        batch = InferenceBatchData(
+            prediction=gen_data,
+            prediction_norm=gen_data,
+            target=target_data,
+            target_norm=target_data,
+            time=None,
+            i_time_start=0,
+        )
+        self._inner.record_batch(batch)
+
+    def get_logs(self, label: str) -> dict[str, Any]:
+        return self._inner.get_logs(label)
+
+    def get_dataset(self) -> xr.Dataset:
+        return self._inner.get_dataset()
 
 
 class SpectrumAggregator:
@@ -16,7 +44,7 @@ class SpectrumAggregator:
         target_time: int = 1,
         nan_fill_fn: Callable[[torch.Tensor, str], torch.Tensor] = lambda x, _: x,
     ):
-        self._wrapped = PairedSphericalPowerSpectrumAggregator(
+        self._wrapped = _InferencePairedSpectrumAgg(
             gridded_operations=gridded_operations,
             report_plot=False,
             nan_fill_fn=nan_fill_fn,
--- a/fme/ace/aggregator/one_step/spectrum.py
+++ b/fme/ace/aggregator/one_step/spectrum.py
@@ -1,14 +1,15 @@
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from typing import Any
 
 import numpy as np
 import torch
 import xarray as xr
 
-from fme.ace.aggregator.inference.data import InferenceBatchData
+from fme.ace.aggregator.inference.data import InferenceBatchData, make_dummy_time
 from fme.ace.aggregator.inference.spectrum import (
     PairedSphericalPowerSpectrumAggregator as _InferencePairedSpectrumAgg,
 )
+from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.typing_ import TensorMapping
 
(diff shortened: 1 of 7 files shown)


fme/ace/aggregator/one_step/spectrum.py
```diff
@@ -1,22 +1,50 @@
...
+    def __init__(self, **kwargs: Any):
+        self._inner = _InferencePairedSpectrumAgg(**kwargs)
+
+    def record_batch(self, target_data: TensorMapping, gen_data: TensorMapping) -> None:
+        batch = InferenceBatchData(
+            prediction=gen_data,
+            prediction_norm=gen_data,
+            target=target_data,
+            target_norm=target_data,
+            time=None,
```
Arcomano1234: Nit: This will throw a type error. This still needs a dummy time to be constructed

fme/ace/aggregator/one_step/spectrum.py
```diff
@@ -1,22 +1,50 @@
...
+    PairedSphericalPowerSpectrumAggregator as _InferencePairedSpectrumAgg,
+)
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.typing_ import TensorMapping
 
 
+class PairedSphericalPowerSpectrumAggregator:
+    """Wraps the inference PairedSphericalPowerSpectrumAggregator for one-step use."""
+
+    def __init__(self, **kwargs: Any):
```
Arcomano1234: Question: Is there any way have any the typing explicit here

fme/coupled/aggregator.py
```diff
@@ -260,28 +253,71 @@ def build(
                 "log_zonal_mean_images=False or "
                 "decrease n_coupled_steps to avoid this warning."
             )
-            log_zonal_mean_images = False
+            log_zonal_mean_images: bool | int = False
```
Arcomano1234, line 256: Thanks for catching this bug

fme/coupled/aggregator.py
```diff
@@ -260,28 +253,71 @@ def build(
...
-            ocean_normalize=ocean_normalize,
-            atmosphere_normalize=atmosphere_normalize,
+            monthly_reference_data=self.monthly_reference_data,
+            time_mean_reference_data=self.time_mean_reference_data,
+            log_nino34_index=True,
+            log_step_means=[StepMeanEntry(step=20, name="mean_step_20")]
+            if n_timesteps_ocean >= 20
+            else [],
+        )
+        atmosphere_ace_config = AceInferenceEvaluatorAggregatorConfig(
```
Arcomano1234: Question / nit: The `atmosphere_ace_config` and `ocean_ace_config` are nearly identical is there any way we make a shared config constructor or something?

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Strange - why doesn't mypy raise a type error here?

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Yes, I think we can and should explicitly list the arguments.

mcgibbon: Yes, I think we can and should explicitly list the arguments.

fme/ace/aggregator/one_step/spectrum.py
```diff
@@ -1,22 +1,50 @@
...
+        self._inner = _InferencePairedSpectrumAgg(**kwargs)
+
+    def record_batch(self, target_data: TensorMapping, gen_data: TensorMapping) -> None:
+        batch = InferenceBatchData(
+            prediction=gen_data,
+            prediction_norm=gen_data,
```
mcgibbon: Similar issue to one you pointed out in the last PR @Arcomano1234 , we should not send non-norm'ed gen and target data along. We should be able to pass these as None.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Yes, I think we can and should explicitly list the arguments.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Claude: Fixed — now uses `make_dummy_time(n_sample, n_time)` to construct a valid `xr.DataArray`. As for why mypy didn't catch this: `xr.DataArray` resolves to `Any` from mypy's perspective since xarray doesn't ship complete type stubs, so passing `None` silently satisfies the annotation.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Claude: Done — replaced `**kwargs: Any` with the explicit signature: `gridded_operations`, `report_plot`, `nan_fill_fn`, and `variable_metadata`.

fme/coupled/aggregator.py
mcgibbon: Claude: Addressed — extracted a `base_ace_config` with the shared fields and use `dataclasses.replace()` to create the ocean and atmosphere variants, which differ only in `log_nino34_index` and `log_step_means`.

fme/ace/aggregator/one_step/spectrum.py
mcgibbon: Claude: Fixed — `prediction_norm` and `target_norm` now default to `None` instead of incorrectly passing the non-normalized data.

fme/ace/aggregator/one_step/spectrum.py
Arcomano1234: > xr.DataArray resolves to Any from mypy's perspective since xarray doesn't ship complete type stubs, so passing None silently satisfies the annotation`

Ah that does make sense.

Arcomano1234 (approved):
Thanks for addressing the comments, LGTM.

PR #1099, opened 2026-04-29
Add configurable metrics with variable filtering

mcgibbon:
Add an optional `metrics` list to `InferenceEvaluatorAggregatorConfig` that enables fine-grained control over which metrics are computed and which variables they apply to. When set, it overrides the toggle-based defaults. Existing configs are fully backward compatible.

Changes:
- `fme.ace.aggregator.inference.config.MetricConfig` new dataclass for per-metric configuration (type, target, variables, custom name, reference data)
- `fme.ace.aggregator.inference.config._VariableFilterAdapter` wraps any sub-aggregator to scope it to specific variables
- `fme.ace.aggregator.inference.config._build_evaluator_metric()` dispatches on metric type to construct the appropriate sub-aggregator
- `fme.ace.aggregator.inference.config.InferenceEvaluatorAggregatorConfig.metrics` new optional field; when set, `build()` uses the metrics list instead of toggle-based logic
- `fme.ace.aggregator.inference.config.VALID_METRIC_TYPES` all 11 supported metric types: mean, step_mean, power_spectrum, zonal_mean, video, time_mean, histogram, seasonal, annual, enso_index, enso_coefficient

- [x] Tests added

PR 3 of 3 in the inference aggregator refactor series (based on #1098).


fme/ace/aggregator/inference/main.py
```diff
@@ -52,6 +52,54 @@
...
+        "enso_index",
+        "enso_coefficient",
+    }
+)
+
+
+class _VariableFilterAdapter:
+    """Wraps a sub-aggregator to filter InferenceBatchData to specified variables."""
+
+    def __init__(self, inner: Any, variables: Sequence[str]):
```
Arcomano1234: Nit: As with the previous PR these a way we could make this more type safe? Claude suggested something like:
> Consider typing it as inner: SubAggregator which is the protocol it must conform to anyway.

fme/ace/aggregator/inference/main.py
```diff
@@ -52,6 +52,54 @@
 NINO34_LAT = (-5, 5)
 NINO34_LON = (190, 240)
 
+VALID_METRIC_TYPES = frozenset(
```
Arcomano1234: Nit: I think we should have ensemble support here. This would require adding `ensemble` here and a few other changes.

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
...
+                f"Unknown metric type: {self.type!r}. "
+                f"Valid types: {sorted(VALID_METRIC_TYPES)}"
+            )
+        if self.type == "step_mean" and self.step is None:
+            raise ValueError("step is required for step_mean metric type")
+
+
+def _build_evaluator_metric(
+    metric: MetricConfig,
+    *,
```
Arcomano1234: Question: We haven't really used this ability in fme previously. I think its useful for constructing calls for functions with a ton of arguments. My question is do we want to try to standardize the use of * for future PRs?

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
...
+    ops: Any,
```
Arcomano1234: Nit: Does `ops` and `horizontal_coordinates` really need to be an Any type here? Can't we use `GriddedOperations` and  `HorizontalCoordinates`

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
...
+    horizontal_coordinates: Any,
+    n_timesteps: int,
+    n_ic_steps: int,
+    timestep: datetime.timedelta,
+    variable_metadata: Any,
+    channel_mean_names: Sequence[str] | None,
+    monthly_reference_data: xr.Dataset | None,
+    time_mean_reference_data: xr.Dataset | None,
+    initial_time: xr.DataArray,
+) -> Any:
```
Arcomano1234: Could change this to a type if we end up making a `SubAggregator` type

Arcomano1234 (commented):
Left a few preliminary comments for this PR.

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
...
+        target: Whether to use denormalized ("denorm") or normalized ("norm")
+            data. Only applies to "mean", "step_mean", and "time_mean" types.
+        step: Forward step number. Required for "step_mean" type.
+        enable_extended_videos: Whether to log extended statistical videos.
+            Only applies to "video" type.
+        reference_data: Path to reference data. For "time_mean" (reference
+            means) and "annual" (monthly reference).
+        channel_mean_names: Variable names for channel mean computation.
+            Applies to norm-target "step_mean" and "time_mean" types.
+        zonal_mean_max_size: Max time dimension for zonal mean images.
```
mcgibbon: I don't like all these metric-specific configurations in the MetricConfig used for all metrics. Should we have a "kwargs" section? To be fair I also don't like losing the typing insight into what settings are available. Maybe we should have a type that's a union of our configs for each metric, to retain this information while not attaching nonsensical settings to other metrics? Then each MetricConfig could use its own .build method.

fme/ace/aggregator/inference/main.py
mcgibbon: Yes, we should add what typing information we can.

fme/ace/aggregator/inference/main.py
mcgibbon: AI agents like using this. I like it and would be fine with adopting it, though for now I'd say it is something we can but don't need to use. For this long signature with so many arguments, I think it's very helpful.

fme/ace/aggregator/inference/main.py
mcgibbon: Yeah I'd like to avoid `Any`.

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
             )
 
 
+@dataclasses.dataclass
+class MetricConfig:
+    """Configuration for a single metric in the explicit metrics list.
+
+    Attributes:
+        type: Metric type. One of: "mean", "step_mean", "power_spectrum",
```
mcgibbon: If we used a union of types, I think something like `type: Literal["mean"]` could be used to select e.g. MeanMetricConfig?

fme/ace/aggregator/inference/main.py
```diff
@@ -84,6 +132,205 @@ def validate(self, n_forward_steps: int):
...
+    channel_mean_names: Sequence[str] | None,
+    monthly_reference_data: xr.Dataset | None,
+    time_mean_reference_data: xr.Dataset | None,
+    initial_time: xr.DataArray,
+) -> Any:
+    """Build a single evaluator sub-aggregator from a MetricConfig."""
+    t = metric.type
+    target = metric.target
+
+    if t == "mean":
```
mcgibbon: Is this long if-else chain a code smell, or is it helpful?

fme/ace/aggregator/inference/main.py
```diff
@@ -121,6 +368,22 @@ class InferenceEvaluatorAggregatorConfig:
     log_step_means: list[StepMeanEntry] = dataclasses.field(
         default_factory=lambda: [StepMeanEntry(step=20)]
     )
+    metrics: list[MetricConfig] | None = None
```
mcgibbon: We've added a list[MetricConfig], but no existing settings have been removed, even though I'd expect this field handles some of those same settings.

fme/ace/aggregator/inference/main.py
```diff
@@ -158,147 +421,177 @@ def build(
...
+                    n_ic_steps=n_ic_steps,
+                    timestep=timestep,
+                    variable_metadata=dataset_info.variable_metadata,
+                    channel_mean_names=channel_mean_names,
+                    monthly_reference_data=monthly_reference_data,
+                    time_mean_reference_data=time_mean_reference_data,
+                    initial_time=initial_time,
+                )
+                if metric.variables is not None:
+                    agg = _VariableFilterAdapter(agg, metric.variables)
```
mcgibbon: The "mean" adapter has a bit of an issue with this when it's computing channel mean. It needs to receive all variables to be able to compute channel mean RMSE. If we used separate Config classes for each metric, we could apply the variable filter adapter on the ones where that's a good solution, and use something different where it's not?

fme/ace/aggregator/inference/main.py
```diff
@@ -158,147 +421,177 @@ def build(
...
-            time_series_aggregators["mean_norm"] = mean_norm_agg
-        for step_mean_entry in self.log_step_means:
-            step_mean_entry.validate(n_forward_steps)
-            step = step_mean_entry.step
-            name = step_mean_entry.get_name()
-            # -1 because step 0 (after IC) is the first forward step
-            target_time = step + n_ic_steps - 1
-            aggregators[name] = _OneStepMeanAdapter(
-                OneStepMeanAggregator(
+        if self.metrics is not None:
```
mcgibbon: I'd like to fully move over to using `self.metrics` and have it support all our existing features, with a "default" that's like the current default. I don't want to maintain two pathways. Is this possible?

fme/ace/aggregator/inference/main.py
```diff
@@ -53,74 +54,449 @@
...
+                (self.channel_mean_names or ctx.channel_mean_names) if is_norm else None
+            ),
+        )
+
+
+@dataclasses.dataclass
+class HistogramMetricConfig:
+    type: Literal["histogram"] = "histogram"
+    variables: list[str] | None = None
+    name: str | None = None
```
mcgibbon: We don't need name to be configurable on these configs, make them class constants. Or, if there is a benefit to having this be a variable, make this (and on other configs) default to the default string instead of None.

fme/coupled/test_aggregator.py
```diff
@@ -248,9 +245,6 @@ def test_inference_logs_labels_exist(tmpdir):
         "power_spectrum/smallest_scale_norm_bias/ocean_var",
         "video/ocean_var",
         # atmosphere-specific keys
-        "annual/atmos_var",
-        "annual/r2/atmos_var_gen",
```
mcgibbon, line 252: Is it possible to update the test configs so these assertions don't change?

configs/baselines/cm4-piControl/uncoupled-atmos/train-config.yaml
```diff
@@ -28,7 +28,6 @@ inference:
         engine: zarr
     num_data_workers: 4
   aggregator:
```
mcgibbon: Is it really OK to have aggregator: with nothing after it? Should we remove this to have the default? If it parses fine I suppose it's fine.

fme/ace/aggregator/inference/main.py
```diff
@@ -53,74 +54,449 @@
...
+
+    def get_dataset(self) -> xr.Dataset:
+        return self._inner.get_dataset()
+
+
 @dataclasses.dataclass
-class StepMeanEntry:
-    """
-    Configuration for logging mean metrics at a particular step.
+class _MetricBuildContext:
```
mcgibbon, line 116: We don't commonly use this pattern, so please add a docstring describing what this does and why it exists.

fme/ace/aggregator/inference/main.py
```diff
@@ -53,74 +54,449 @@
...
-        step: Number of forward steps after which to log mean metrics. For example,
-            step=20 will log mean metrics at the 20th forward step
-            (i.e. time index n_ic_steps + 19).
-        name: Name to use for the logged metrics. If None, will use "mean_step_{step}".
-    """
 
+@dataclasses.dataclass
+class MeanMetricConfig:
+    type: Literal["mean"] = "mean"
+    variables: list[str] | None = None
```
mcgibbon, line 155: This list of variables seems to be unused? Similarly on other configs. How does it get used? Could we make it so the `.build` uses it instead of a higher level code applying it to the build output, if that's why I don't understand?

mcgibbon:
Superceded by https://github.com/ai2cm/ace/pull/1123

PR #1100, opened 2026-04-29
Add filter_preserves_global_mean option to SFNO

mcgibbon:
Add an option to prevent the spectral filter from modifying the l=0 (global mean) spherical harmonic coefficient, so that all global-mean impact must result from local operations (norms, MLPs, skip connections).

Changes:
- `SFNONetConfig.filter_preserves_global_mean` config field, threaded through `SpectralFilterLayer` and `FourierNeuralOperatorBlock`
- `SpectralConvS2`: after the spectral convolution + LoRA update, replaces the l=0 output with the l=0 input via `torch.cat` (autograd-safe, distributed-aware)
- Raises `NotImplementedError` if used with non-`linear` filter types

- [x] Tests added


fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -85,6 +85,10 @@ class SFNONetConfig:
             0 disables LoRA.
         spectral_lora_alpha: Strength of LoRA adaptations for spectral
             convolutions. Defaults to spectral_lora_rank if None.
+        filter_preserves_global_mean: If True, the spectral filter preserves
```
Arcomano1234, line 88: Nit: you need to also update `NoiseConditionedSFNOBuilder` in `fme/ace/registry/stochastic_sfno.py ` to have a `filter_preserves_global_mean` attribute

fme/core/models/conditional_sfno/test_sfnonet.py
```diff
@@ -280,3 +282,105 @@ def test_block_speed():
...
+    )
+    output = model(x, context)
+    output.sum().backward()
+    for block in model.blocks:
+        weight = block.filter.filter.weight
+        assert weight.grad is not None
+        assert not torch.all(weight.grad == 0)
+
+
+def test_can_call_sfnonet_with_filter_preserves_global_mean():
```
Arcomano1234: Question: Is this test necessary? I think `test_can_call_sfnonet_with_filter_preserves_global_mean` and `test_filter_preserves_global_mean_allows_grad` can be combined by just having the shape of output check in `test_can_call_sfnonet_with_filter_preserves_global_mean`

Arcomano1234 (approved):
Left one question / optional comment about consolidating some of tests but this should be good to go.

fme/core/models/conditional_sfno/test_sfnonet.py
mcgibbon: Updated

PR #1101, opened 2026-04-29
update era5 baseline to stochastic ACE with updated era5 dataset

Arcomano1234:
Update the ERA5 baseline to a stochastic ACE (ACE2S) training recipe that uses the newest ERA5 zarr3 dataset (through 2025), with a two-stage 1-step pretrain then multi-step finetune workflow. See this google slide deck for more details ([slides](https://docs.google.com/presentation/d/14XpW1eSmiVtMSspQY3gQlmL166URcyOB6qGpdmilZnc/edit?slide=id.g2ed0744072d_0_124#slide=id.g2ed0744072d_0_124)). 

Changes:
- `configs/baselines/era5/ace-train-config-1-step-pretrain.yaml`: rename from `ace-train-config.yaml`
  - switch to `NoiseConditionedSFNO` (`embed_dim: 512`, `noise_embed_dim: 32`, isotropic noise)
  - use `EnsembleLoss` (CRPS/energy-score weights) with ensemble training (`n_ensemble: 2`)
  - update to zarr ERA5 paths/stats (goes how to 2025 now and is zarr3)
  - move surface fields (`TMP2m`, `Q2m`, `UGRD10m`, `VGRD10m`) from prognostic in/out to diagnostic outputs
  - add diagnostics variables needed for ocean coupling and downscaling applications (`total_frozen_precipitation_rate`, `PRMSL`, two components of surface stress)
  - enable frozen precipitation clipping and constant-temperature energy budget correction 
  - treat `global_mean_co2` as a next-step forcing (better match other datasets where it matters more)

- `configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml`: add multi-step finetuning config initialized from the pretrained checkpoint, with probabilistic `n_forward_steps` schedule and `optimize_last_step_only`
- `configs/baselines/era5/ace-evaluator-config.yaml`: point evaluator at the new zarr ERA5 dataset and extend evaluation length through 2025; enable zonal-mean and histogram aggregators
- `configs/baselines/era5/run-ace-train.sh`: run ACE2S pretrain/finetune on titan with 4 GPUs; wire staged training job names
- `configs/baselines/era5/README.md`: document the stochastic ACE / updated ERA5 baseline and remove Google Slides links


configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml
```diff
@@ -0,0 +1,262 @@
...
+        - '1996-05-15T00:00:00'
+        - '1996-07-01T00:00:00'
+        - '1996-08-15T00:00:00'
+        - '1996-10-01T00:00:00'
+        - '1996-11-15T00:00:00'
+    dataset:
+      data_path: /climate-default/
+      file_pattern: 2026-03-19-era5-1deg-8layer-1940-2025.zarr
+      engine: zarr
+    num_data_workers: 16
```
oliverwm1: is this what we typically use? I thought there were issues with oversubscription with such a large number of data workers—can we get away with 8 per GPU?

oliverwm1: (ditto for other loaders and the 1 step pretrain config)

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -230,3 +245,5 @@ stepper:
       - tendency_of_total_water_path_due_to_advection
       - TMP850
       - h500
+      - total_frozen_precipitation_rate
+      - PRMSL
```
oliverwm1, line 239: should launch a training job and potentially adjust weightings for these new diagnostics before merging.

configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml
```diff
@@ -0,0 +1,262 @@
...
+      h500: 10
+      TMP850: 5
+      Q2m: 0.5
+      DLWRFsfc: 2
+      ULWRFsfc: 5
+      USWRFsfc: 2
+      DSWRFsfc: 2
+      USWRFtoa: 2
+      tendency_of_total_water_path_due_to_advection: 0.5
+stepper:
```
oliverwm1, line 105: I'm a bit out of the loop on mechanics of this, but is there no way to just provide a path to existing stepper that you want to fine-tune, assuming that you only want to change the dataset and/or loss function, and not modify anything about the architecture? It is a bit bug-prone to have to copy-and-paste the exact same stepper config again. Now that we've separated out the "stepper_training" section from the "stepper" config this might be easier to implement. cc @mcgibbon

configs/baselines/era5/run-ace-train.sh
```diff
@@ -28,7 +28,7 @@ run_training() {
   gantry run \
     --name "$job_name" \
     --task-name "$job_name" \
-    --description 'Run ACE2-ERA5 training' \
+    --description 'Run ACE2S-ERA5 training' \
```
oliverwm1, line 31: Should also update the clusters to what we typically use now.

configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml
Arcomano1234: Thats a good question I just copied what was in the baseline already but yeah I thought that 16 was too much as well. I typically only use 8 max.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: Yeah agreed. Ill launch the job with the config as is and see how it looks.

configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml
Arcomano1234: I am not aware of any such capabilities, but yeah I agree it is quite bug prone and cumbersome to copy and paste the exact stepper config for FTing.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
oliverwm1: Another change that we made for AIMIP which I think we should also implement here is making u10/v10/t2m/q2m diagnostic variables. @brianhenn did you add some scaling values for the loss function when making these variables diagnostic?

configs/baselines/era5/ace-train-config-multi-step-finetuning.yaml
mcgibbon: I'll look into this, ideally it should also be possible to say "I want to use the config in this stepper checkpoint, but change this one parameter". Indeed having separated the stepper from stepper training should make this more straightforward / make it make more sense.

mcgibbon: PR is up: https://github.com/ai2cm/ace/pull/1103

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: Might also want to add wind stresses for future work like the ACE-NEMO paper

oliverwm1: agreed

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -32,27 +35,35 @@ logging:
...
+    - data_path: /climate-default/
+      file_pattern: 2026-03-19-era5-1deg-8layer-1940-2025.zarr
+      engine: zarr
       subset:
         start_time: '2021-01-01'
 validation_loader:
-  batch_size: 128
-  num_data_workers: 16
-  prefetch_factor: 4
+  batch_size: 64
```
Arcomano1234: Larger stochastic model will sometimes run out of memory with val batch size greater than 64

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -76,7 +88,7 @@ stepper_training:
       specific_total_water_0: 0.5
       specific_total_water_1: 0.25
       specific_total_water_2: 0.5
-      PRATEsfc: 0.5
+      PRATEsfc: 2.0
```
Arcomano1234: I see no point in not upping this to at least 2 if not higher

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -93,15 +105,17 @@ stepper:
       builder:
         type: NoiseConditionedSFNO
         config:
-          embed_dim: 384
-          noise_embed_dim: 0
+          embed_dim: 512
```
brianhenn, line 97: Curious about the tradeoffs from increasing the embed_dim to 512 (model skill vs memory footprint for inference on smaller GPUs)?

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -88,15 +111,20 @@ stepper_training:
...
   step:
     type: single_module
     config:
       builder:
         type: NoiseConditionedSFNO
         config:
-          embed_dim: 384
-          noise_embed_dim: 0
+          embed_dim: 512
+          noise_embed_dim: 64
```
mcgibbon: Suggest using 32, I lowered it a little while ago and it doesn't seem to make any difference (but makes the model faster to run).

Arcomano1234: As in 32 `noise_embed_dim`? Yeah I realized that I have configs that go back and forth between 32 and 64 and never noticed as far as I can tell.

mcgibbon: Yes exactly.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -222,6 +214,10 @@ stepper:
       - northward_wind_5
       - northward_wind_6
       - northward_wind_7
+      - TMP2m
```
brianhenn, line 222: Support leaving these here as diagnostics (and removing them as prognostic variables). In AIMIP we removed them entirely from the output list and then learned them as secondary decoder pressure-level diagnostics, but that approach is a pain since we like TMP2m for many evals and having to run another stage of post-training adds overhead.

brianhenn (commented):
Also while we're here can you delete the Google slides links I put in the README in this directory last time? Since now on open development it resulted in random share requests.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
brianhenn, line 119: Also potentially worth adding the energy corrector given its apparent benefits in the AIMIP context ([slides](https://docs.google.com/presentation/d/19MjZOUUvl3ph2GRwOgEkZdkx4Lo5_yWy6VtXZvJ_S84/edit?slide=id.g3e8cb62190a_0_7#slide=id.g3e8cb62190a_0_7))

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: As you suspected this did not have any noticeable effects going from 64 to 32 `noise_embed_dim`. Will update this

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
brianhenn, line 10: (nit): Without an explicit "name" argument, this inference will get labeled "inference", which is fine. But if a *second* inference is added without a name for the first, the first's name will become "inference_0" and won't match prior jobs' keys on wandb (a bit of a surprise IMO).

Not sure if the right move is to change the baseline config here for redundancy/to teach people that they should still use the expected defaults, or to make a code change separately to ensure we aren't changing long-established keys by accident.

brianhenn: I made an issue https://github.com/ai2cm/ace/issues/1361. This is mostly separate from this PR, but could add the expected name back if you like.

mcgibbon: Maybe we should require explicit names be set if more than one inference is given.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: Yeah and it seems to not make any difference as far as I can see

oliverwm1 (approved):
LGTM, thanks Troy!

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -119,6 +119,9 @@ stepper:
       corrector:
         conserve_dry_air: true
         moisture_budget_correction: advection_and_precipitation
+        clip_frozen_precipitation: true
```
mcgibbon, line 122: I'm kind of worried about the negative impact this has on both the skill and PDF of frozen precipitation, I would avoid including this in a published paper right now. I'd suggest just removing the clip correction for now. But I don't think this should block merge.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
```diff
@@ -138,6 +142,7 @@ stepper:
         - USWRFtoa
       next_step_forcing_names:
       - DSWRFtoa
+      - global_mean_co2
```
mcgibbon, line 145: We want the new baseline to be with-co2 correct?

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: Yeah the feature is already merged, I can just turn it off. I will say this is just the new baseline template so people can standardize their ACE2S training, but I agree its not necessarily needed.

configs/baselines/era5/ace-train-config-1-step-pretrain.yaml
Arcomano1234: Yes I believe so right @oliverwm1 ?

oliverwm1: yes

PR #1103, opened 2026-04-30
Add CheckpointStepperConfig to load stepper config from checkpoint

mcgibbon:
When fine-tuning from an existing checkpoint, users currently need to copy-paste the entire stepper configuration and specify a `parameter_init.weights_path`. This PR adds `CheckpointStepperConfig`, which loads the stepper configuration directly from a checkpoint file, eliminating the need to duplicate the config.

Changes:
- `fme.ace.stepper.single_module.CheckpointStepperConfig`: new dataclass with a `checkpoint_path` field that loads and returns a `StepperConfig` from a saved checkpoint
- `fme.ace.stepper.single_module.load_stepper_config_from_checkpoint`: extracts the config from a checkpoint without instantiating the model
- `fme.ace.train.train_config.TrainConfig`: `stepper` field now accepts `StepperConfig | CheckpointStepperConfig`, resolved to a `StepperConfig` via `stepper_config` in `__post_init__`
- `fme.ace.train.train_config.TrainBuilders`: all methods updated to use `stepper_config`

- [x] Tests added


fme/ace/train/train_config.py
```diff
@@ -187,7 +191,7 @@ class TrainConfig:
 
     train_loader: DataLoaderConfig
     validation_loader: DataLoaderConfig
-    stepper: StepperConfig
+    stepper: StepperConfig | CheckpointStepperConfig
```
mcgibbon, line 194: Note that this stepper: config is _independent_ of the parameter_init setting in stepper_training. This means you could use this to define a new training based on an existing one but with random init, or you can fine-control how the model is initialized, frozen, etc. during fine-tuning the same way as before. The only difference is you're replacing this configuration to define the way the stepper is configured/initialized to be based on an existing checkpoint.

jpdunc23: Nice, this should be helpful!

With this feature, we may want to be a bit more careful about supporting TrainConfig backwards compatibility.

mcgibbon: > With this feature, we may want to be a bit more careful about supporting TrainConfig backwards compatibility.

What do you mean? For what it's worth, being able to use CheckpointStepperConfig on previous checkpoints will be guaranteed by inference backwards compatibility - we don't use anything that isn't used for inference loading. Am I missing something?

fme/ace/stepper/single_module.py
```diff
@@ -1780,6 +1798,22 @@ class StepperOverrideConfig:
     prescribed_prognostic_names: Literal["keep"] | list[str] = "keep"
 
 
+def load_stepper_config_from_checkpoint(
```
oliverwm1: why do we need this function in addition to the existing `load_stepper_config` that is right below it?

fme/ace/train/train_config.py
oliverwm1: Is this basically just going back to state prior to https://github.com/ai2cm/ace/commit/69289c005a1e3ac46170b2ff46c689e987a4a4f6? Or is something different now?

fme/ace/stepper/single_module.py
mcgibbon: Because overrides are only implemented on Stepper, not on StepperConfig. I looked at some quick refactors to implement it on StepperConfig, but the way multi call replacement is implemented tightly couples the override implementation for the Stepper state and the StepperConfig, making this difficult.

For now I'm leaving this as-is, renaming the helpers and making a small refactor to the function below to make this purpose clearer and to get the "don't need to make a stepper" optimization when override_config is None.

fme/ace/train/train_config.py
mcgibbon: It is different now. The prior behavior was to use the checkpoint to say "let's overwrite everything about the Trainer based on this checkpoint". The new behavior is to say "let's define the configuration of the Stepper (but not stepper_training, including not ParameterInitializationConfig) based on the checkpoint".

The old behavior wasn't compatible with fine-tuning because you couldn't re-define any aspects of the model or its training. The new behavior is much more flexible, at the cost if needing to be more explicit about your settings (e.g. saying you want to use the checkpoint for both the architecture config and for the parameter initialization).

mcgibbon:
Should be ready for re-review.

fme/ace/train/train_config.py
```diff
@@ -225,6 +229,10 @@ class TrainConfig:
     resume_results: ResumeResultsConfig | None = None
 
     def __post_init__(self):
+        if isinstance(self.stepper, CheckpointStepperConfig):
+            self.stepper_config = self.stepper.to_stepper_config()
+        else:
+            self.stepper_config = self.stepper
```
jpdunc23, line 236: From agent review:

> In `TrainConfig.__post_init__`, `self.stepper_config` is assigned but never declared as a dataclass field (or annotated at the class level). This means:
> 
> It won't appear in type-checker analysis of `TrainConfig` instances
> It won't be visible to IDE autocompletion without reading `__post_init__`
> `dataclasses.fields(TrainConfig)` won't include it
> Consider declaring it as a field with `init=False`:
> ```
> stepper_config: StepperConfig = dataclasses.field(init=False)
> ```
> This makes the attribute discoverable to type checkers and readers while still computing it in `__post_init__`.

fme/ace/stepper/test_single_module.py
```diff
@@ -2358,3 +2360,28 @@ def test_ocean_derived_variables_integration(
         # hfds is diagnostic, so the generated net_energy_flux_into_ocean
         # differs from the reference
         assert not torch.allclose(pred_flux, ref_flux)
+
+
+def test_load_stepper_config_from_checkpoint(tmp_path: pathlib.Path):
+    from fme.ace.testing import save_stepper_checkpoint
```
jpdunc23: I've noticed agents doing this lately (imports inside test definition) which leads to some repetitive imports, as here and in `test_checkpoint_stepper_config_to_stepper_config`. Maybe just add to the imports at the top of the file?

jpdunc23 (approved):
LGTM! Left a couple of comments.

fme/ace/stepper/test_single_module.py
mcgibbon: I agree.

fme/ace/train/train_config.py
mcgibbon: Good idea.

PR #1104, opened 2026-04-30
Save optimization state in `ckpt_{epoch:04d}.tar`

jpdunc23:
Include the optimization state when creating `ckpt_{epoch:04d}.tar` checkpoints with the `checkpoint_save_epochs` configuration.

- [x] Tests added


mcgibbon (commented):
Will need team discussion on this one, we explicitly avoided including this before because it makes the checkpoints much larger, which can be prohibitive for storing epoch checkpoints. Maybe it should be a config flag, or maybe what's here is best, but it depends what people think.

jpdunc23:
> Will need team discussion on this one, we explicitly avoided including this before because it makes the checkpoints much larger, which can be prohibitive for storing epoch checkpoints. Maybe it should be a config flag, or maybe what's here is best, but it depends what people think.

Sounds good. I'm curious to know how folks are currently using `checkpoint_save_epochs`, since I would think that `ema_checkpoint_save_epochs` is more useful for inference purposes.

mcgibbon (approved):
LGTM!

PR #1105, opened 2026-04-30
Add script to extract `Stepper` checkpoints from an `CoupledStepper` checkpoint

jpdunc23:
Adds `scripts/coupled/create_decoupled_checkpoint.py` to create atmosphere or ocean `Stepper` checkpoints that are usable in `fme.ace` fine-tuning or inference.

Changes:
- Adds `scripts/coupled/create_decoupled_checkpoint_beaker.sh` convenience script for downloading the `CoupledStepper` checkpoint from beaker and uploading the extracted `Stepper` checkpoint as a new beaker dataset.
- Fixes bugs in `scripts/test_distributed_context.py`.

- [ ] Tests added

### DIFF
--- a/scripts/coupled/create_decoupled_checkpoint.py
+++ b/scripts/coupled/create_decoupled_checkpoint.py
@@ -1,0 +1,41 @@
+import argparse
+import logging
+
+import torch
+
+from fme.coupled.stepper import load_coupled_stepper
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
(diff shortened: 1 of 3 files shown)


oliverwm1 (approved):

oliverwm1:
I think we should delete the failing test, since this is no longer necessary since #1070

jpdunc23:
> I think we should delete the failing test, since this is no longer necessary since #1070

Ah shoot, I wish I had seen this earlier since I had to fix some issues in the test. ~~Will delete.~~

On second thought, let's get the fixes in main and then delete in a separate PR, so we avoid bringing the bugs back in case we want a similar test for some reason.

PR #1106, opened 2026-04-30
Fix latents, inputs order in mixture of experts `generate`

AnnaKwa:
`DiffusionModel.prepare_generation_inputs` returns tensors in order `inputs, latents` but the mixture of experts predictor swapped the return order. This PR fixes that bug.


Arcomano1234 (commented):
Approved

Arcomano1234 (approved):

PR #1109, opened 2026-05-01
Avoid rolling out more steps than needed for coupled optimization

jpdunc23:
Prior to this PR, `CoupledTrainStepper` always took `n_coupled_steps` forward steps for every batch, regardless of the ocean's `LossContributionsConfig.n_steps`, which could be less than `n_coupled_steps`. This PR updates `CoupledTrainStepper.train_on_batch()` to step forward only as far as needed to compute the loss.

Changes:
- Add `n_required_forward_steps` helper property to `LossContributions`, used by new `n_required_outer_steps()` helper method on `CoupledStepperTrainLoss`.

- [x] Tests added


fme/coupled/stepper.py
```diff
@@ -1670,7 +1690,14 @@ def _accumulate_loss(
         )
         output_iterator = iter(output_generator)
         output_list: list[ComponentEnsembleStepPrediction] = []
-        n_outer_steps = data.ocean_data.n_timesteps - self.n_ic_timesteps
+        n_outer_steps_data = data.ocean_data.n_timesteps - self.n_ic_timesteps
+        # Clamp to at least 1 outer step so downstream gen_data is non-empty
```
mcgibbon, line 1707: Is this desired behavior? Or would it be better to have zero steps?

I would expect in practice we don't want this to happen, right? Like we shouldn't be configuring stochastic samplers where n_steps=0, or we should probably "re-roll" such an outcome. It feels like this would be an artificial consequence of randomizing the number of slow and subsequent fast steps independently, instead of randomizing the number of total fast steps (and using it to compute the number of slow steps). A schedule of total fast steps would never include n=0 (e.g. we could reasonably disallow it in config validation).

fme/coupled/loss.py
```diff
@@ -152,6 +176,11 @@ def sample_n_steps(self) -> None:
     def effective_loss_scaling(self) -> TensorDict:
         return self._loss.effective_loss_scaling
 
+    def n_required_forward_steps(self) -> int:
+        if self._weight == 0.0:
+            return 0
+        return int(min(self._n_steps, self._max_n_steps))
```
mcgibbon: nit: _n_steps is initialized to TimeLengthProbabilities.max_n_forward_steps (the max sampler outcome) while _max_n_steps is n_coupled_steps [* n_inner_steps] (the physical step limit from the training config). These are independent constraints and the min is correct, but the names don't make the distinction obvious. Consider renaming _max_n_steps to something like _max_data_steps or _n_steps_available to clarify that it's the data/config ceiling, not a property of the loss horizon, and renaming the init arguments / adding a docstring.

fme/coupled/stepper.py
```diff
@@ -1387,6 +1387,18 @@ def sample_n_steps(self) -> None:
...
+        """Minimum number of outer (ocean) steps needed so that every
+        component step contributing to the current batch's loss is computed.
+
+        Callers must invoke ``sample_n_steps()`` beforehand for stochastic
+        configs so the value reflects the current batch.
+        """
+        ocean_required = self._loss_objs["ocean"].n_required_forward_steps()
+        atmos_required = self._loss_objs["atmosphere"].n_required_forward_steps()
+        atmos_outer = -(-atmos_required // n_inner_steps)  # ceil division
+        return max(ocean_required, atmos_outer)
```
mcgibbon, line 1400: High level issue (no action required): This is tight coupling between the loss objects and the training stepper that results from both the training stepper and the loss objects being responsible for organizing a "window" of training and deciding what steps get called and what steps get what loss applied on them. In the ace code this coupling is avoided by the training stepper being solely responsible, with the step scheduler being the sub-object it uses to handle the concerns you have in the loss contributions object. That's why we never had the issue this PR resolves in ace.

I don't think this is a huge issue, and it's pre-existing, but I do think it's worth bringing up each time this separation makes adding a new feature more complex.

mcgibbon (commented):
LGTM, one pre-existing naming nit and a high level continuing comment that doesn't require action.

fme/coupled/stepper.py
jpdunc23: I'm actually using n=0 as a (low probability) possibility for both ocean and atmosphere, the reason being that I've found success with both FTO (ocean-only fine-tuning) and FTA (atmosphere-only fine-tuning), i.e. in optimizing only the atmosphere or the ocean component. So for joint fine-tuning I wanted some batches to optimize steps for only one of the components while holding the other fixed.

When we get unlucky and both draw 0 steps, this clamping happens. In current configs, this only happens with probability 0.00125 (once every couple of epochs, or so), so not a huge concern but I agree that we should disallow this by re-rolling one of the component's n_steps draw until we get a non-zero outcome; I can work on a follow-up PR to do that.

fme/coupled/loss.py
jpdunc23: Agreed this is confusing. I addressed in https://github.com/ai2cm/ace/pull/1141 which is in review

fme/coupled/stepper.py
jpdunc23: I agree there's tight coupling here. I can work on a refactoring PR to try to decouple this, basically by all functionality other than per-component weighting away from `LossContributions`. Will still need to query `LossContributions` for the weight to determine what the max steps are needed for the loss, but that seems less onerous.

fme/coupled/loss.py
jpdunc23: This should hopefully look a bit better now with #1141 merged

mcgibbon (approved):
LGTM

PR #1111, opened 2026-05-01
Properly propagate reduction argument to weighted and ensemble losses

Arcomano1234:
Per-channel training loss currently reports identical values for every channel for certain losses ( AreaWeightedMSELoss, EnsembleLoss) because these loss functions are reduced to a scalar internally, and `_reduce_to_per_channel` then uniformly broadcast that scalar across channels. This PR adds `reduction="none" `support to these losses so they preserve per-channel structure, enabling meaningful per-channel loss diagnostics during training.

This is a little tricky with `EnsembleLoss`. When both CRPS and ES are active in `EnsembleLoss` with reduction="none", CRPS returns (batch, channel, lat, lon) while energy score returns (batch, channel) since it operates in spectral space; the energy score output is unsqueezed to (batch, channel, 1, 1) so it broadcasts over the spatial dimensions when summed with CRPS.

Changes:
- `AreaWeightedMSELoss`, `CRPSLoss`, `EnergyScoreLoss`, and `EnsembleLoss` now take a reduction parameter
- `_find_channel_dim` is added to` loss.py` to help with edge cases of how we ask for a specific tensor dimension (positive vs negative dimension and if a tensor has a ensemble dim).

- [x] Tests added


fme/core/loss.py
```diff
@@ -481,23 +526,21 @@ def forward(
...
+            result = result + self.energy_score_weight * es
         if self.diff_crps_loss is not None:
-            diff_crps = self.diff_crps_weight * self.diff_crps_loss(
+            result = result + self.diff_crps_weight * self.diff_crps_loss(
                 gen_norm, target_norm
             )
-        else:
-            diff_crps = torch.tensor(0.0)
-        return crps + energy_score_loss + diff_crps
+        assert isinstance(result, torch.Tensor)
```
mcgibbon, line 542: Suggestion: Can we write this in a way where you don't need the assertion because static typing ensures the correct return type?

fme/core/loss.py
```diff
@@ -481,23 +526,21 @@ def forward(
...
-            crps = torch.tensor(0.0)
+            result = self.crps_weight * self.crps_loss(gen_norm, target_norm)
         if self.energy_score_weight > 0:
-            energy_score_loss = self.energy_score_weight * self.energy_score_loss(
-                gen_norm, target_norm
-            )
-        else:
-            energy_score_loss = torch.tensor(0.0)
+            es = self.energy_score_loss(gen_norm, target_norm)
+            if isinstance(result, torch.Tensor) and result.ndim > es.ndim:
```
mcgibbon, line 534: The need for checks like this seems like a code smell.

I thought we were having the losses not reduce anything, and then reducing uniformly at a higher level? That way we don't need to implement reductions on each and every loss type. This seems a bit prone to error.

mcgibbon (commented):
Maybe we should merge this to fix this issue, but it points at a larger structural issue in the loss code.

mcgibbon:
Working on a branch I think will supercede these changes, so let's hold for now.

mcgibbon:
I think this is defunct since merging 1159. Closing.

Arcomano1234:
Yes thank you I forgot to close this.

PR #1112, opened 2026-05-01
Pre-process UFS Replay ocean dataset

Arcomano1234:
Adds an xarray-beam pipeline (`scripts/ufs-replay/`) that processes the NOAA UFS GEFSv13 Replay ocean (MOM6) + atmosphere (FV3) dataset into a training-ready zarr store for SamudrACE, following the same runner/infrastructure pattern as `scripts/era5/`. Pre-processing matches CM4's ocean convention as closely as possible, with a few UFS-specific exceptions (nearest-neighbor filling for a handful of variables with sparse coastal NaNs, e.g. wind stress).

Changes:
- `scripts/ufs-replay/pipeline/ufs-replay-pipeline.py`: two independent Beam streams (ocean, atmosphere) write to the same output zarr store. Ocean does thickness-weighted vertical coarsening (75 → 19 levels matching CM4 depths) at native horizontal resolution *before* regridding — matches the CM4 convention (vertical averaging and horizontal regridding don't commute) and is cheaper, since only 19 levels then pass through the regridder. Atmosphere derives frozen precipitation rate from bucket accumulations and averages 3-hourly fields to the ocean's 6-hourly cadence before regridding. Ocean and atmosphere each keep their own native time-labeling convention (ocean: interval-center; atmosphere: interval-end); `_align_times` restricts both datasets up front to only the mutually-covered times, so the streams align purely by integer Beam chunk offsets with no per-chunk time matching.
- The pipeline always emits data at the ocean model's native 6-hourly cadence — no time coarsening happens inline. Coarser products (daily, 5-day, etc.) are produced downstream from that raw output via `scripts/data_process/time_coarsen.py` (reusing the existing shared infrastructure rather than a bespoke script), which snapshots ocean/sea-ice state variables and window-averages atmosphere forcing / ocean surface flux variables — see `scripts/ufs-replay/README.md` and `scripts/data_process/configs/ufs-replay-ocean-1deg-19level-daily.yaml`.
- `scripts/data_process/get_stats.py`: add `UFS_REPLAY` as a supported `ClimateDataType`.
- `scripts/ufs-replay/{Dockerfile,Makefile,environment.yaml,dataflow-requirements.txt,pipeline/run-dataflow.sh}`: local (DirectRunner) and production (Dataflow) run infrastructure, mirroring `scripts/era5/`.

Sanity checks:

1. Distribution of variable means and standard deviations: [notebook](https://github.com/ai2cm/explore2/blob/main/troya/2026-05-13-UFS-replay-ocean-comparison/2026-05-13-ufs-vs-cm4-statistics.ipynb)
2. 2D and 3D masks: [notebook](https://github.com/ai2cm/explore2/blob/main/troya/2026-05-13-UFS-replay-ocean-comparison/2026-05-14-processed-ufs-vs-cm4-comparison.ipynb)
3. Matching of CM4's pre-processed ocean levels to UFS: [notebook](https://github.com/ai2cm/explore2/blob/main/troya/2026-05-13-UFS-replay-ocean-comparison/2026-05-18-ufs-vertical-level-selection.ipynb)

- [x] Tests added


scripts/data_process/configs/UFS-replay-ocean-1deg.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - 43
+  - 49
+  - 55
+  - 61
+  - 66
+  - 71
+  - 74
+
+# No temporal coarsening: keep native 6-hourly resolution
+time_coarsen_factor: 1
```
Arcomano1234: This will be a good question to ask James / Oli about temporal resolution we want

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+    )
+    return new_key, output
+
+
+# ---------------------------------------------------------------------------
+# Template building
+# ---------------------------------------------------------------------------
+
+
+def _extract_invariant_fields(
```
elynnwu, line 866: There're some duplicated logic in this function and `_process_chunk` such as building the 3D mask and per-level masking. Probably better if the they share the same helper functions?

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+        sample = v.isel(time=0).values
+        need_fill = np.isnan(sample) & ocean
+        if not need_fill.any():
+            continue
+
+        n_fill = int(need_fill.sum())
+        logging.info("NN-filling %d ocean cells for '%s'", n_fill, name)
+
+        valid = ~np.isnan(sample)
+        _, nn_idx = distance_transform_edt(
```
elynnwu, line 385: Since the ocean mask doesn't change in time, you can compute the fill indices once during template creation and passed to workers instead of calling `distance_transform_edt` on every chunk

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+    logging.info("Atmo after time slice: %d timesteps", ds.sizes["time"])
+    return ds
+
+
+# ---------------------------------------------------------------------------
+# Per-chunk processing (called by each Beam worker)
+# ---------------------------------------------------------------------------
+
+
+def _process_chunk(
```
elynnwu: Better to break the different processes into different functions so that it can also be re-used in `_extract_invariant_fields`

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+    if "thetao_0" in ds_levels:
+        sst_K = ds_levels["thetao_0"] + 273.15
+        sst_K.attrs = {"long_name": "Sea surface temperature", "units": "K"}
+        ds_levels["sst"] = sst_K
+    if "zos" in ds_ocean_2d:
+        ds_ocean_2d["zos"].attrs.setdefault("long_name", "Sea Surface Height")
+
+    # Surface velocity aliases
+    if "uo_0" in ds_levels:
+        ds_levels["ssu"] = ds_levels["uo_0"].copy()
```
elynnwu: [nit] Do we need `.copy()` here?

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+    )
+    return parser
+
+
+def main():
+    import json
+
+    parser = _get_parser()
+    args, pipeline_args = parser.parse_known_args()
+    print(pipeline_args)
```
elynnwu: [nit] Use `logging.info`

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1479 @@
...
+        ds_atmo,
+        args.output_grid,
+        vert_indices,
+        time_coarsen_factor,
+        output_time,
+    )
+
+    logging.info("Template generated. Starting pipeline.")
+    output_store = _make_zarr_store(args.output_path, read_only=False)
+    print(PipelineOptions(pipeline_args).get_all_options())
```
elynnwu: [nit] use `logging.info`

scripts/ufs-replay/README.md
```diff
@@ -0,0 +1,78 @@
...
+9. Coarsen in time (6-hourly → daily)
+
+## Quick Start
+
+```bash
+# Create local conda environment
+make create_environment
+
+# Local test run (DirectRunner)
+make ufs_replay_direct_test_run
```
elynnwu, line 85: this Make target is missing

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+        siv.attrs = {"long_name": "Sea Ice Volume Per Area", "units": "m"}
+        ds["sea_ice_volume"] = siv
+    if "hfds" in ds and "sea_surface_fraction" in ds:
+        ds["hfds_total_area"] = ds["hfds"] * ds["sea_surface_fraction"]
+        ds["hfds_total_area"].attrs = {
+            "long_name": "heat flux into sea water scaled by sea surface fraction",
+            "units": "W/m2",
+        }
+
+    # --- NN fill ---
```
elynnwu: The refactor might have a bug in ordering. Can you double check?

> `_make_template` currently calls `_process_chunk(...)` first without passing `nn_fill_map`. That triggers `_process_chunk` fallback path, so it computes and applies NN fill to the template sample. Then `_make_template` computes `nn_fill_map` from that already-filled processed dataset. Since the coastal NaNs were already removed, `_compute_nn_fill_indices` likely sees nothing to fill and returns an empty map.
> 
> Then the Beam workers receive that empty-but-not-None map. Because `nn_fill_map is not None`, workers call `_apply_nn_fill(ds, {})`, which is a no-op, and they do not take the fallback path that would compute indices locally. So production chunks can retain coastal NaNs that the pipeline was supposed to fill.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+    coarsened.update(_build_land_sea_fractions(mask_2d))
+
+    # Drop stray coordinates
+    keep_coords = {"time", "lat", "lon"}
+    cleaned = {}
+    for name, da in coarsened.items():
+        stray = [c for c in da.coords if c not in keep_coords and c not in da.dims]
+        cleaned[name] = da.drop_vars(stray) if stray else da
+
+    return xr.Dataset(cleaned)
```
spencerkclark: Could you drop these coordinates right after opening the dataset at the front of the pipeline?  I'm assuming they are not used for anything (at least the ones not related to the vertical coordinate).

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+    ds_ocean = ds_ocean.drop_vars(
+        [v for v in OCEAN_DROP if v in ds_ocean], errors="ignore"
+    )
+    ds_ocean = ds_ocean.drop_vars("landsea_mask", errors="ignore")
+
+    # Rename ocean variables
+    rename_map = {k: v for k, v in OCEAN_RENAME.items() if k in ds_ocean}
+    ds_ocean = ds_ocean.rename(rename_map)
+    stress_map = {k: v for k, v in STRESS_RENAME.items() if k in ds_ocean}
+    ds_ocean = ds_ocean.rename(stress_map)
```
spencerkclark: I think this sort of data cleaning might be better to handle right after opening the dataset at the front of the pipeline.  I would also recommend doing it without the conditional logic surrounding the vertical dimension name and presence of the `"ho"` variable, since I think there is little likelihood that we would use this script on another dataset that did not have these quirks.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+            ho_ds = _regrid_dataset(
+                ho_ds,
+                output_grid,
+                source_grid_ocean,
+                skipna=True,
+                na_thres=1.0,
+            )
+            ho_ds = ho_ds.assign_coords(
+                lat=ds_ocean.lat.values, lon=ds_ocean.lon.values
+            )
```
spencerkclark: Is there a reason the regridding of `"ho"` is done separately?

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+            vdim,
+        )
+    else:
+        ds_ocean = ds_ocean.drop_vars("ho", errors="ignore")
+        ds_levels = _split_3d_to_levels(
+            ds_ocean,
+            vars_3d_present,
+            vdim,
+            mask_3d,
+        )
```
spencerkclark: In practice do you think we are likely to use this workflow without vertical coarsening indices?  It seems like it would simplify the code to always assume vertical coarsening indices were present.  In principle it should be possible to specify no coarsening via vertical coarsening indices if needed.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+        return url
+
+
+def _match_time_type(dt, reference_time):
+    """Convert *dt* to match the type of *reference_time*.
+
+    The UFS zarr stores may use different calendars (Julian for MOM6,
+    possibly standard/proleptic_gregorian for FV3).  This function
+    inspects the actual time values in the dataset and converts
+    accordingly.
```
spencerkclark: To me it looks like they use the same calendar (`"proleptic_gregorian"`) for the `"time"` dimension coordinate.  Can you provide more context?  I think the other time coordinates can be ignored.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
spencerkclark: I agree. Could the mask generation logic be isolated to the invariant context?  It is surprising to also see it in `_compute_ocean_vertical_coarsening` with different handling of the `"idepth_*"` coordinate.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1510 @@
...
+    for name in ds.data_vars:
+        if name in vars_3d or name == thickness_name:
+            continue
+        if vdim not in ds[name].dims:
+            coarsened[name] = ds[name]
+
+    # Surface mask + related fields
+    mask_2d = coarsened["mask_0"].astype(np.float32)
+    mask_2d.attrs = {"long_name": "ocean mask", "units": "0 if land, 1 if ocean"}
+    coarsened["mask_2d"] = mask_2d
```
spencerkclark: Does the mask need to be re-computed in this function or can it be passed in since it is invariant?

scripts/ufs-replay/README.md
Arcomano1234: Thanks for finding that claude literally deleted it at some point it seems

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: Thank you so much for pointing this out, yes this is a critical bug and fixed in [dffd99b](https://github.com/ai2cm/ace/pull/1112/commits/dffd99bd0dde709b2760f188a50bd2801293b4a2)

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+# FV3 atmosphere forcing variables → output names
+ATMO_FORCING_VARS = {
+    "dlwrf_ave": "DLWRFsfc",
+    "dswrf_ave": "DSWRFsfc",
+    "ulwrf_ave": "ULWRFsfc",
+    "uswrf_ave": "USWRFsfc",
+    "lhtfl_ave": "LHTFLsfc",
+    "shtfl_ave": "SHTFLsfc",
+    "prateb_ave": "PRATEsfc",
+    "pressfc": "PRESsfc",
```
spencerkclark: Do we use surface pressure as a forcing anywhere?

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+                vertical_coarsening_indices=vert_indices,
+                time_coarsen_factor=time_coarsen_factor,
+                nn_fill_map=nn_fill_map,
+            )
+            | xbeam.ConsolidateChunks(output_shards)
+            | xbeam.ChunksToZarr(
+                output_store,
+                template,
+                zarr_chunks=output_chunks,
+                zarr_shards=output_shards,
```
spencerkclark: Is the chunking and sharding logic working as expected?  It seems like the dataset was written out with a chunk size of 50 in time and no shards, but maybe I am not looking at the right path:
```
>>> xr.open_zarr("gs://vcm-ml-intermediate/2026-06-03-ufs-replay-ocean-1deg-19level-5day-1994-2023.zarr").thetao_0.encoding
{'chunks': (50, 180, 360), 'preferred_chunks': {'time': 50, 'lat': 180, 'lon': 360}, 'compressors': (ZstdCodec(level=0, checksum=False),), 'filters': (), 'shards': None, 'serializer': BytesCodec(endian=<Endian.little: 'little'>), '_FillValue': nan, 'dtype': dtype('<f4'), 'coordinates': 'cftime ftime z_l'}
```

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+        frozen_accum = sum(ds_atmo[v] for v in accum_vars)
+        frozen_rate = (frozen_accum / dt_seconds).clip(min=0)
+        frozen_rate.attrs = {
+            "long_name": "total frozen precipitation rate",
+            "units": "kg/m**2/s",
+        }
+        ds_atmo = ds_atmo.drop_vars(accum_vars)
+        ds_atmo["total_frozen_precipitation_rate"] = frozen_rate
+
+    ds_atmo_6h = ds_atmo.resample(time="6h", closed="right", label="right").mean()
```
spencerkclark: `coarsen` with a custom coordinate function might do something a bit more predictable. For this small example `resample` returns a length-7 array, where the first bin only averages a single value, which might be surprising:

```
>>> time = xr.date_range("2000", freq="3h", periods=12)
>>> da = xr.DataArray(range(1, 13), dims=["time"], coords=[time])
>>> da.resample(time="6h", closed="right", label="right").mean()
<xarray.DataArray (time: 7)> Size: 56B
array([ 1. ,  2.5,  4.5,  6.5,  8.5, 10.5, 12. ])
Coordinates:
  * time     (time) datetime64[ns] 56B 2000-01-01 ... 2000-01-02T12:00:00
```


`coarsen` simply averages every two values resulting in a length-6 array:
```
>>> def coord_func_last(x, axis=-1):
...     return x[:, -1]
...
>>> da.coarsen(time=2, coord_func=coord_func_last).mean()
<xarray.DataArray (time: 6)> Size: 48B
array([ 1.5,  3.5,  5.5,  7.5,  9.5, 11.5])
Coordinates:
  * time     (time) datetime64[ns] 48B 2000-01-01T03:00:00 ... 2000-01-02T09:...
```

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+    if not common_strs:
+        raise ValueError(
+            f"No overlapping times between averaged atmosphere and ocean. "
+            f"Atmo 6h times: {atmo_strs[:5]}..., "
+            f"Ocean times: {ocean_strs[:5]}..."
+        )
+
+    # Select by position using the string-matched indices
+    mask = [str(t) in common_strs for t in ds_atmo_6h.time.values]
+    return ds_atmo_6h.isel(time=mask)
```
spencerkclark: If this workflow is set up carefully and the proper checks are done at a higher level, I feel like this low-level time alignment should not be necessary.  I might try to avoid it if possible, since it could paper over subtle bugs.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+            )
+        ds_atmo_6h = ds_atmo_6h.coarsen(
+            time=time_coarsen_factor, boundary="trim"
+        ).mean()
+        ds_ocean_2d = ds_ocean_2d.coarsen(
+            time=time_coarsen_factor, boundary="trim"
+        ).mean()
+        ds_levels = ds_levels.coarsen(time=time_coarsen_factor, boundary="trim").mean()
+        # Snap daily-mean time labels to 12Z (the natural midpoint of a day).
+        # coarsen().mean() averages the timestamps, giving 09Z for (00,06,12,18).
```
spencerkclark: Strictly speaking, if averaging snapshots, then 09Z would be more correct as the average time of 00Z, 06Z, 12Z, and 18Z.  Bigger picture though, if we want something representative of 12Z, should we really be averaging 00Z, 06Z, 12Z, 18Z, and 24Z instead?

If taking the mean of interval averages ending on the coordinate label, i.e. -06Z to 00Z, 00Z to 06Z, 06Z to 12Z, and 12Z to 18Z, then 06Z would be the most representative as the midpoint.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
```diff
@@ -0,0 +1,1516 @@
...
+        )
+
+    # Restrict to common times (string-based to handle cross-calendar types)
+    atmo_time_strs = {str(t) for t in ds_atmo_6h.time.values}
+    ocean_mask = [str(t) in atmo_time_strs for t in ds_ocean.time.values]
+    atmo_mask = [
+        str(t) in {str(t) for t in ds_ocean.time.values[ocean_mask]}
+        for t in ds_atmo_6h.time.values
+    ]
+    ds_atmo_6h = ds_atmo_6h.isel(time=atmo_mask)
```
spencerkclark: I might find it easier to follow if the atmosphere data were processed in a separate stream than the ocean data, similar to [how the ERA5 pipeline is set up to process different classes of data in different streams](https://github.com/ai2cm/ace/blob/4729449052394a4f7591fe1655638a70cbb0f8af/scripts/era5/pipeline/xr-beam-pipeline.py#L1214-L1302).  I think this is particularly true for the time logic since the two source datasets have a different frequency.

spencerkclark (commented):
Thanks for your patience @Arcomano1234.  I think there is room to simplify this workflow in a few places.  I believe this is particularly true around the time handling, but I want to make sure I understand the big picture a bit more before providing concrete suggestions.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: Yes I was running to some problems when doing the thickness weighings. If I recall correctly its because ho has valid values at cells where tracer fields (thetao, so, etc.) were NaN. Specifically, at the deepest levels where the ocean bottom intersects a vertical layer. When regridded together with tracers, xESMF's skipna=True + na_thres=1.0 settings interact: a grid cell that is NaN in the tracers but valid in ho would have its ho contribution masked out by the NaN in the co-located tracer. Regridding ho independently preserves the correct thickness field, which is then used for the thickness-weighted vertical coarsening. Without this, the coarsened levels near the ocean floor would have incorrect weights. 

To be honest this came from back and forth's with claude, I am not sure what's the "best way" the vertical coarsening for the ocean, I would be curious how you overcame this issue in CM4 or how its done when processing the ocean data online.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: No this is some left over logic when I was early testing of the code. Will just always assume we are doing vertical coarsening indices.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: At some points I was running into 2 different calendars but at least right now it seems we can get rid of this.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: Yes that should be possible. I thought I already fixed that let me see whats going on.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: No this came from not originally having `frozen_precipitation_rate` in the training dataset and then ACE, for what ever reason, looked for `ICEsfc` and then finally `torch.zeros_like(self.surface_pressure) `but I didn't realize that was why it was looking for surface pressure and just thought the ocean model needed it. Should be fine to remove.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
spencerkclark: (The choice here obviously depends on what the raw UFS-Replay data represents).

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
Arcomano1234: Yeah so that is not what this pipeline made. That was from a notebook that definitely seems to have a bug. The daily dataset from the pipeline is `gs://vcm-ml-intermediate/ufs-replay-ocean-1deg-19level-1994-2023.zarr/`
```
xr.open_zarr("gs://vcm-ml-intermediate/ufs-replay-ocean-1deg-19level-1994-2023.zarr/").thetao_0.encoding
Out[3]:
{'chunks': (1, 180, 360),
 'preferred_chunks': {'time': 1, 'lat': 180, 'lon': 360},
 'compressors': (ZstdCodec(level=0, checksum=False),),
 'filters': (),
 'shards': (360, 180, 360),
 'serializer': BytesCodec(endian=<Endian.little: 'little'>),
 '_FillValue': nan,
 'dtype': dtype('<f4'),
 'coordinates': 'cftime ftime z_l'
```

spencerkclark: I see, thanks, so the production run that produced this dataset pre-dated the date prepending logic in the `Makefile`.

Arcomano1234: Yes, considering it was only 70$ to run the full pipeline, I figured it was better to start training once I was sure the data was fine but I was still iterating on the final version of the pipeline. I will re-run the full pipeline once this PR is ready to merge so that everything is consistent.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: I think this should be resolved now.

scripts/data_process/Makefile
```diff
@@ -251,6 +251,10 @@ era5_1deg_stats_beaker_dataset:
 era5_1deg_16layer_stats_beaker_dataset:
 	./compute_stats.sh --config configs/era5-1deg-16layer-1940-2022.yaml
 
+.PHONY: ufs_replay_ocean_stats
+ufs_replay_ocean_stats:
+	./compute_stats.sh --config configs/ufs-replay-ocean-1deg-19level.yaml
```
spencerkclark, line 260: Make sure to commit this YAML file (I know you probably meant to).

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
spencerkclark: Yeah, no worries, that makes sense.

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: Yeah this was a question I had for you. How does the time averaging work in CM4? I would like to match that as much as possible.

scripts/data_process/Makefile
Arcomano1234: Oh good call not sure why that didn't get added

scripts/data_process/configs/ufs-replay-ocean-1deg-19level.yaml
```diff
@@ -0,0 +1,9 @@
+runs:
+  ufs-replay-ocean-1deg-19level-1994-2023: ""
+data_output_directory: gs://vcm-ml-intermediate
+stats:
+  output_directory: gs://vcm-ml-intermediate/ufs-replay-ocean-1deg-19level-stats
```
Arcomano1234, line 5: Note: Will update this with the final production run job which will have the datetime in the run name

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
spencerkclark: For CM4, we average the state variables occurs online, so the numerator for interval averages is incremented at each model timestep.  

It looks like this might be the case for ocean portion of the UFS-Replay data (quoting from [their example notebook](https://nbviewer.org/gist/timothyas/f37ec28562e97c74217401fceb144f1f)):

> The atmosphere dataset is at 3 hour frequency, and ocean dataset is at 6 hour frequency. The time stamps selected correspond to the "analysis" portion of the Replay, coinciding with the IAU. While the atmosphere fields are snapshots, the ocean fields are a 6 hour temporal average over the IAU period.

So it would be more like this situation (which is simpler):

> If taking the mean of interval averages ending on the coordinate label, i.e. -06Z to 00Z, 00Z to 06Z, 06Z to 12Z, and 12Z to 18Z, then 06Z would be the most representative as the midpoint.

A question still remains though of what the time label represents.  Is it the end of the averaging interval (like we impose in CM4) or is it the center?  [This page](https://psl.noaa.gov/data/ufs_replay/) seems to suggest it is the center of the averaging window:

> Ocean model output (MOM6), available every 6 hours:
> - `ocn_<YYYY>_<MM>_<DD>_<HH>.nc` (Corrector segment and first guess) time stamp defines the middle of the averaging window.

That suggests that the average of 00Z, 06Z, 12Z, and 18Z would be representative of the average between -03Z through 21Z, which would have a center of 09Z.  

We would want to align the atmosphere data with these time intervals, though note I believe the accumulated versions of those fields are labeled with the interval endpoint (e.g., from [this page](https://psl.noaa.gov/data/ufs_replay/) again):

> fhr00 files contain data valid for the center of the previous cycle's 6-hour "corrector (replay) segment." As an example, in a 12Z folder (e.g., 1994010112), fhr00 files provide model output with incremental analysis updates applied from 03Z (1994010103) to 06Z (1994010106), with accumulated values representing 03Z to 06Z and timestamps at 06Z.

We can think about the cleanest way to do that, but before then, is that consistent with your reading of the available UFS-Replay documentation?

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
Arcomano1234: Thanks for the suggestions this should be fixed now

scripts/ufs-replay/pipeline/xr-beam-pipeline.py
Arcomano1234: Yes ocean stuff is averaged over 6hrs and time stamp is centered and why Claude was originally calling the center 09Z. For the atmosphere yes that is correct.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1474 @@
...
+def _open_ufs_zarr(url: str) -> xr.Dataset:
+    """Open a UFS zarr store with cftime support and suppress warnings."""
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", FutureWarning)
+        ds = xr.open_zarr(
+            _make_zarr_store(url),
+            chunks=None,
+            use_cftime=True,
```
spencerkclark: nit: sorry I missed this previously—I think we should be able to get away with solely using `np.datetime64` values and a standard `pandas.DatetimeIndex` for this dataset.
```suggestion
```
I believe this would remove the need for the `_match_time_type` function.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+        ds = ds.rename(rename)
+    found = [v for v in variables if v in ds]
+    missing = [v for v in variables if v not in ds]
+    if missing:
+        logging.warning("Atmo vars not found in store: %s", missing)
+    logging.info("Atmo vars found: %s", found)
+
+    # The FV3 store may use a time dimension called "time" on some
+    # variables but have other variables on different time dims (e.g.
+    # "ftime").  Select variables first, then log the time coordinate.
```
spencerkclark: nit: I think this comment can be removed, since it makes things sound more complicated than they really are (I believe all the relevant variables use `"time"` as their time dimension in the source store).
```suggestion
```

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+    ref = ds.time.values[0]
+    logging.info("Atmo time type: %s (%s)", type(ref).__name__, ref)
+    t0 = _match_time_type(start_time, ref)
+    t1 = _match_time_type(end_time, ref)
+    logging.info("Atmo time slice: %s to %s", t0, t1)
+    ds = ds.sel(time=slice(t0, t1))
+    logging.info("Atmo after time slice: %d timesteps", ds.sizes["time"])
+
+    # Drop stray non-dimension coordinates (cftime, ftime, etc.) that
+    # pollute downstream coarsen steps and the output encoding.
```
spencerkclark: Fine to keep this comment though.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+        rename["grid_xt"] = "lon"
+    if "grid_yt" in ds.dims:
+        rename["grid_yt"] = "lat"
+    if rename:
+        ds = ds.rename(rename)
+    found = [v for v in variables if v in ds]
+    missing = [v for v in variables if v not in ds]
+    if missing:
+        logging.warning("Atmo vars not found in store: %s", missing)
+    logging.info("Atmo vars found: %s", found)
```
spencerkclark: Question: is there a use case for being forgiving about this or should we let xarray raise an error about it instead?

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+def _validate_time_alignment(ocean_times, atmo_times) -> None:
+    """Check that the atmosphere times exactly interleave the ocean times.
+
+    The 6-hourly ocean fields are averages centered on their labels,
+    while the 3-hourly FV3 interval-average fields are labeled at the
+    END of their averaging window, so ocean time T is covered by the
+    atmosphere pair (T, T+3h).  Verifying this correspondence once up
+    front lets the two Beam streams align purely by integer chunk
+    offsets, with no per-chunk time matching.
+    """
```
spencerkclark: Question: as far as I can tell this and the downstream time logic are correct, but would it be slightly simpler to conform the FV3 data's time coordinate convention to the ocean data's at the start of the pipeline?  

Then I believe the logic around how to handle the time coordinate around coarsening could be identical between the two streams.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+    # first timestamp of each pair.  The FV3 interval-average fields are
+    # labeled at the END of their 3h averaging window, so the pair
+    # (T, T+3h) covers T-3h through T+3h — exactly the window of the
+    # center-labeled 6h ocean average at T.  coarsen is preferred over
+    # resample because the atmo data is regularly spaced and coarsen
+    # always produces equal-sized groups (resample can create uneven
+    # bins at boundaries).
+    atmo_times = ds_atmo.time.values
+    ds = ds_atmo.coarsen(time=2, boundary="trim").mean()
+    ds = ds.assign_coords(time=atmo_times[::2])
```
spencerkclark: Question: other than the special time coordinate handling, is there a reason to do the time coarsening in two steps rather than all at once after horizontal regridding?

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1312 @@
...
+
+    # --- Vertical coarsening ---
+    vars_3d_present = [v for v in VARS_3D if v in ds_ocean]
+    indices_as_tuples = [tuple(pair) for pair in vertical_coarsening_indices]
+    ds = _compute_ocean_vertical_coarsening(
+        ds_ocean,
+        vars_3d_present,
+        indices_as_tuples,
+        VDIM,
+    )
```
spencerkclark: Question: sorry I did not spot this earlier—to better align with the CM4 pipeline should we vertically coarsen prior to doing horizontal regridding?  I believe it would technically be more accurate.

scripts/ufs-replay/coarsen_to_5day.py
spencerkclark, line 1: A few questions regarding this script: do you normally just run this on your VM? Is it possible to re-use [the code that handles this for the atmosphere datasets](https://github.com/ai2cm/ace/blob/main/scripts/data_process/time_coarsen.py)? 

Is it mainly for cost savings to handle things in two stages this way rather than coarsen directly to 5-daily resolution in the Dataflow pipeline?

spencerkclark (commented):
Thanks @Arcomano1234 for the work to address https://github.com/ai2cm/ace/pull/1112#discussion_r3391289941; the pipeline is a lot easier for me to follow now.  A few more minor questions related to possible additional simplifications, but I think this is getting close.

scripts/ufs-replay/coarsen_to_5day.py
Arcomano1234: Yeah I have just been using my regular VM as its unclear what we want for time resolution and I didn't want it to be tied down to the full workflow.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
Arcomano1234: Yeah if anything that will save us quite a bit on computation. I didn't realize it would be more accurate in that order of regridding, I figured it was the other way.

scripts/data_process/get_stats.py
```diff
@@ -152,18 +153,23 @@ def get_stats(
...
 
     dims = DIMS[config.data_type]
+    skipna = config.data_type in ("UFS_REPLAY",)
 
     # Explicitly compute the statistics here, since xarray does not support
     # writing netCDFs with the scipy engine with the distributed scheduler.
     # There is no harm to computing here versus later, since the end result is
     # not something memory intensive.
-    centering = ds.mean(dim=dims).compute()
+    centering = ds.mean(dim=dims, skipna=skipna).compute().reset_coords(drop=True)
```
spencerkclark: A couple nits:
- `skipna=None` (the default) and `skipna=True` should have identical behavior, at least for float variables. Was there a reason for specifying it explicitly here?
- `reset_coords` feels like it should be done when creating the dataset rather than when computing the stats.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1281 @@
...
+    ds = _open_ufs_zarr(URL_ATMO)
+    ds = ds.rename({"grid_xt": "lon", "grid_yt": "lat"})[variables]
+    ds = ds.sel(time=slice(start_time, end_time))
+
+    # FV3's interval-average fields are labeled at the END of their 3h
+    # averaging window; shift to the window CENTER (the ocean stream's
+    # convention, see OCEAN_TIME_STEP usage) so that a plain
+    # coarsen(...).mean() lands exactly on the ocean output times
+    # downstream (see _process_atmo_chunk and _validate_time_alignment).
+    ds = ds.assign_coords(time=ds.time - _ATMO_HALF_STEP)
```
spencerkclark: After some further thought, now that we are considering the ocean data to be "instantaneous" I think we should retain the interval-average-center time labeling convention for the ocean data (as is done currently) and the interval-average-end time labeling convention for the atmosphere data (meaning we should not conform the data here): 

```suggestion
```

The motivation for this is to use the same conventions that we use for all of our other datasets.

I suggest we do the time alignment in a separate step with knowledge of the available ocean and atmosphere times.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1281 @@
...
+    assert args.output_time_shardsize % args.output_time_chunksize == 0, msg
+
+    output_chunks = {"time": args.output_time_chunksize}
+    output_shards = {"time": args.output_time_shardsize}
+    process_chunks = {"time": args.process_time_chunksize}
+
+    # --- Open source datasets ---
+    logging.info("Opening datasets")
+
+    ds_ocean = open_ocean(OCEAN_LOAD_VARS, start_time, end_time)
```
spencerkclark, line 1207: This is my new suggestion to ensure alignment while retaining the time labeling conventions of the ocean and atmosphere datasets (note the time bounds are applied to the ocean dataset and then are carried through implicitly by the time alignment to the atmosphere dataset):

```suggestion
    ds_ocean = open_ocean(OCEAN_LOAD_VARS, start_time, end_time)
    ds_atmo = open_atmo(atmo_load_vars)
    ds_ocean, ds_atmo = _align_times(ds_ocean, ds_atmo)
```

where `_align_times` is defined as:

```python
def _align_times(ds_ocean, ds_atmo) -> tuple[xr.Dataset, xr.Dataset]:
    ocean_times = ds_ocean.indexes["time"]
    atmosphere_times = ds_atmo.indexes["time"]

    aligned_ocean_times = []
    aligned_atmo_times = []

    for time in ocean_times:
        atmo_time_1 = time - np.timedelta64(ATMO_TIME_STEP, "h")
        atmo_time_2 = time

        if atmo_time_1 in atmo_times and atmo_time_2 in atmo_times:
            aligned_ocean_times.append(time)
            aligned_atmo_times.append(atmo_time_1)
            aligned_atmo_times.append(atmo_time_2)

    aligned_ocean_times = pd.Index(aligned_ocean_times)
    aligned_atmo_times = pd.Index(aligned_atmo_times)

    return ds_ocean.sel(time=aligned_ocean_times), ds_atmo.sel(time=aligned_atmo_times)
```

Within the atmosphere coarsening, instead of using the default time labeling for the resulting coordinate, we will then use the interval endpoints.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
```diff
@@ -0,0 +1,1281 @@
...
+    # Average consecutive pairs of 3h fields → 6h, before horizontal
+    # regridding, to minimize how much data passes through the
+    # expensive conservative regridder.  Records are already
+    # center-labeled (see open_atmo) half the atmo cadence apart, so
+    # averaging labels lands exactly on the ocean stream's output time
+    # — no manual relabeling needed.  coarsen is preferred over resample
+    # because the atmo data is regularly spaced and coarsen always
+    # produces equal-sized groups (resample can create uneven bins at
+    # boundaries).
+    ds = ds_atmo.coarsen(time=2, boundary="trim").mean()
```
spencerkclark, line 786: Suggestion to retain interval-end time labeling for the atmosphere data:
```suggestion
    # Average consecutive pairs of 3h fields → 6h, before horizontal
    # regridding, to minimize how much data passes through the
    # expensive conservative regridder. Label interval averages with
    # the time at the end of each interval.
    ds = ds_atmo.coarsen(time=2, boundary="trim").mean()
    ds = ds.assign_coords(time=ds_atmo.time.sel(time=slice(1, None, 2))
```

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
spencerkclark: Some adjustment might be needed for the `process_time_chunksize` code following this.

scripts/ufs-replay/pipeline/ufs-replay-pipeline.py
spencerkclark: Apologies, I ended up changing my mind about this.  Resolving to reduce noise.

spencerkclark (commented):
Thanks @Arcomano1234, sorry for all the back and forth—one last final set of minor suggestions regarding time alignment / labeling and then I think we'll be good!

scripts/data_process/get_stats.py
Arcomano1234: Yeah at some point it was needed but I think these changes were accidentally brought into this PR. I just checked and there is no need now.

spencerkclark (approved):
Thanks @Arcomano1234; this looks good to me now!

Arcomano1234:
@spencerkclark Thanks for helping so much to get this into a good state!

PR #1116, opened 2026-05-04
Refactor generic trainer to use validation and inference callbacks

mcgibbon:
Refactors the generic Trainer to use pluggable validation and inference callbacks instead of directly managing inference data and calling hardcoded methods. This decouples the trainer from inference specifics, setting the stage for weighted multi-inference checkpoint selection.

Changes:
- `fme.core.generics.trainer.ValidationCallback`, `InferenceCallback`: new callback protocols returning `(logs, loss/error)` tuples
- `fme.core.generics.trainer.AggregatorBuilderABC`: simplified from `Generic[PS, TO, SD]` to `Generic[TO]`, removed `get_inference_aggregator`
- `fme.core.generics.trainer.TrainConfigProtocol`: removed `get_inference_epochs` (epoch filtering moved to callbacks)
- `fme.core.generics.trainer.Trainer`: removed `inference_data` param and `validate_one_epoch`/`inference_one_epoch` instance methods; added `set_validation_callback`/`set_inference_callback` setters; `train()` now calls callbacks
- `fme.ace.train.train.build_trainer`, `fme.coupled.train.train.build_trainer`: build validation and inference callbacks as closures, set them on the trainer after construction

- [x] Tests added

### DIFF
--- a/fme/ace/train/train.py
+++ b/fme/ace/train/train.py
@@ -48,6 +48,7 @@
 # Karthik Kashinath - NVIDIA Corporation
 # Animashree Anandkumar - California Institute of Technology, NVIDIA Corporation
 
+import contextlib
 import dataclasses
 import logging
 import os
@@ -158,8 +159,6 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
             aggregator=aggregator,
             diagnostics_subdir=f"epoch_{epoch:04d}",
             record_logs=lambda logs: None,
-            ema=trainer._ema,
-            validate_using_ema=config.validate_using_ema,
         )
         return logs, logs["val/mean/loss"]
 
@@ -170,7 +169,7 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
             return {}, None
         logs = inference_one_epoch(
             stepper=stepper,
-            validation_context=trainer.validation_context,
+            validation_context=contextlib.nullcontext,
             dataset=inference_data,
             aggregator=aggregator_builder.get_inference_aggregator(),
             label="inference",
@@ -191,7 +190,7 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
         logging.info("Starting weather evaluation inference run")
         return inference_one_epoch(
             stepper=stepper,
-            validation_context=trainer.validation_context,
+            validation_context=contextlib.nullcontext,
             dataset=data,
             aggregator=aggregator,
             label=label,
--- a/fme/ace/train/train.py
+++ b/fme/ace/train/train.py
@@ -171,7 +171,6 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
         label: str,
         epoch: int,
     ):
-        logging.info("Starting weather evaluation inference run")
         return inference_one_epoch(
             stepper=stepper,
             validation_context=contextlib.nullcontext,
--- a/fme/ace/train/train.py
+++ b/fme/ace/train/train.py
@@ -68,7 +68,7 @@ from fme.ace.aggregator.inference.main import (
     InferenceEvaluatorAggregatorConfig,
 )
 from fme.ace.aggregator.train import TrainAggregatorConfig
-from fme.ace.data_loading.batch_data import BatchData, PairedData, PrognosticState
+from fme.ace.data_loading.batch_data import BatchData, PrognosticState
 from fme.ace.stepper import TrainOutput
 from fme.ace.train.train_config import TrainBuilders, TrainConfig
 from fme.core.cli import prepare_config, prepare_directory
@@ -82,6 +82,7 @@ from fme.core.generics.trainer import (
     Trainer,
     inference_one_epoch,
 )
+from fme.core.generics.validation import run_validation
 from fme.core.typing_ import TensorDict, TensorMapping
 
 
@@ -138,7 +139,6 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
     trainer = Trainer(
         train_data=train_data,
         validation_data=validation_data,
-        inference_data=inference_data,
         stepper=stepper,
         build_optimization=builder.get_optimization,
         build_ema=builder.get_ema,
@@ -148,7 +148,40 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
         do_gc_collect=do_gc_collect,
     )
 
-    def inference(
+    def validation_callback(epoch: int):
+        validation_data.set_epoch(epoch)
+        aggregator = aggregator_builder.get_validation_aggregator()
+        logs = run_validation(
+            train_stepper=stepper,
+            validation_data=validation_data,
+            aggregator=aggregator,
+            diagnostics_subdir=f"epoch_{epoch:04d}",
+            record_logs=lambda logs: None,
+            ema=trainer._ema,
+            validate_using_ema=config.validate_using_ema,
+        )
+        return logs, logs["val/mean/loss"]
+
+    inference_epochs = config.get_inference_epochs()
+
+    def inference_callback(epoch: int):
+        if epoch not in inference_epochs:
+            return {}, None
+        logs = inference_one_epoch(
+            stepper=stepper,
+            validation_context=trainer.validation_context,
+            dataset=inference_data,
+            aggregator=aggregator_builder.get_inference_aggregator(),
+            label="inference",
+            epoch=epoch,
+        )
+        error = logs.get("inference/time_mean_norm/rmse/channel_mean")
+        return logs, error
+
+    trainer.set_validation_callback(validation_callback)
+    trainer.set_inference_callback(inference_callback)
+
+    def _run_inference(
         data: InferenceDataABC[PrognosticState, BatchData],
         aggregator: InferenceEvaluatorAggregator,
         label: str,
@@ -165,7 +198,7 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
         )
 
     end_of_epoch_ops = builder.get_end_of_epoch_callback(
-        inference,
+        _run_inference,
         normalize=stepper.normalizer.normalize,
         channel_mean_names=stepper.loss_names,
         output_dir=config.output_dir,
@@ -178,7 +211,7 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
 
 
 class AggregatorBuilder(
-    AggregatorBuilderABC[PrognosticState, TrainOutput, PairedData],
+    AggregatorBuilderABC[TrainOutput],
 ):
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


fme/ace/train/train.py
```diff
@@ -148,7 +148,40 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
         do_gc_collect=do_gc_collect,
     )
 
-    def inference(
+    def validation_callback(epoch: int):
```
mcgibbon: Include return type hints for these callbacks.

fme/core/generics/test_trainer.py
```diff
@@ -439,6 +438,41 @@ def build_ema(modules: torch.nn.ModuleList) -> EMATracker:
         do_gc_collect=False,  # for much faster tests
     )
 
+    inference_epochs = config._inference_epochs
+
+    def validation_callback(epoch):
```
mcgibbon: Similar comment about return types.

fme/core/generics/trainer.py
```diff
@@ -269,10 +286,11 @@ def __init__(
         n_params = count_parameters(self.stepper.modules)
         logging.info(f"Number of trainable model parameters: {n_params}")
 
-        self._inference_data = inference_data
         self._do_gc_collect = do_gc_collect
         self._in_ema_context = False
         self._started_training = False
+        self._validation_callback: ValidationCallback | None = None
+        self._inference_callback: InferenceCallback = _null_inference_callback
```
mcgibbon: Add these as input arguments even if we aren't using it as one, to be consistent with end_of_epoch_callback and to make it clear from the init that this state exists.

fme/core/generics/trainer.py
```diff
@@ -364,17 +395,12 @@ def train(self):
...
-            else:
-                inference_logs = {}
-            valid_loss = valid_logs["val/mean/loss"]
+            logging.info("Starting evaluation before training")
+            valid_logs, valid_loss = validation_callback(self._epochs_trained)
+            inference_logs, _ = inference_callback(self._epochs_trained)
             logging.info(f"Validation loss before training: {valid_loss}")
             logging.info("Logging to wandb")
-            all_logs = valid_logs | inference_logs | {"epoch": self._epochs_trained}
+            all_logs = {**valid_logs, **inference_logs, "epoch": self._epochs_trained}
```
mcgibbon: Unrelated change, please revert unless it is needed.

fme/coupled/train/train.py
```diff
@@ -73,9 +69,43 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> Trainer:
...
+            validation_data=validation_data,
+            aggregator=aggregator,
+            diagnostics_subdir=f"epoch_{epoch:04d}",
+            record_logs=lambda logs: None,
+            ema=trainer._ema,
+            validate_using_ema=config.validate_using_ema,
+        )
+        return logs, logs["val/mean/loss"]
+
+    def inference_cb(epoch: int):
```
mcgibbon: Please expand "cb" to "callback", we use verbose naming.

fme/coupled/train/train.py
```diff
@@ -73,9 +70,43 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> Trainer:
...
+            dataset=inference_data,
+            aggregator=aggregator_builder.get_inference_aggregator(),
+            label="inference",
+            epoch=epoch,
+        )
+        error = logs.get("inference/time_mean_norm/rmse/channel_mean")
+        return logs, error
+
+    trainer.set_validation_callback(validation_callback)
+    trainer.set_inference_callback(inference_callback)
```
mcgibbon: This must be done post-init for Trainer because the callback accesses the instantiated Trainer. We already do the same thing for the end-of-epoch callback.

fme/core/generics/test_trainer.py
```diff
@@ -449,6 +448,39 @@ def build_ema(modules: torch.nn.ModuleList) -> EMATracker:
...
+            dataset=inference_data,
+            aggregator=aggregator_builder.get_inference_aggregator(),
+            label="inference",
+            epoch=epoch,
+        )
+        error = logs.get("inference/time_mean_norm/rmse/channel_mean")
+        return logs, error
+
+    trainer.set_validation_callback(validation_callback)
+    trainer.set_inference_callback(inference_callback)
```
mcgibbon: We should be able to pass the callbacks during trainer init, now that we don't use the trainer in them. Please do this and remove the `set_` helpers for these callbacks. That includes the existing end-of-epoch callback.

fme/coupled/train/train.py
```diff
@@ -73,9 +71,41 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> Trainer:
...
```
mcgibbon: We should be able to set these on init instead of using setters.

fme/core/generics/trainer.py
```diff
@@ -299,6 +319,12 @@ def on_terminate(signum, frame):
     def set_end_of_epoch_callback(self, end_of_epoch_callback: EndOfEpochCallback):
         self._end_of_epoch_callback = end_of_epoch_callback
 
+    def set_validation_callback(self, callback: ValidationCallback):
+        self._validation_callback = callback
+
+    def set_inference_callback(self, callback: InferenceCallback):
+        self._inference_callback = callback
```
mcgibbon: Do we still need these setters (including for end of epoch callback)? Delete them if not.

fme/core/generics/trainer.py
```diff
@@ -589,66 +613,20 @@ def validation_context(self):
...
 
-    def validate_one_epoch(self):
-        if self._current_epoch_num_batches_seen > 0:
-            raise RuntimeError(
-                "Validation should only be called after a full epoch has been trained"
-            )
-        logging.info(
-            f"Starting validation step for model trained for "
-            f"{self._epochs_trained} epochs"
-        )
```
mcgibbon, line 620: Some logging lines, like this one, got removed. Please add them back. This one for example should happen before the validation callback is called.

fme/ace/train/train.py
```diff
@@ -157,28 +174,43 @@ def inference(
...
+        end_of_batch_callback=end_of_batch_ops,
+        end_of_epoch_callback=end_of_epoch_ops,
+        inference_callback=inference_callback,
+        do_gc_collect=do_gc_collect,
+    )
 
 
 class AggregatorBuilder(
-    AggregatorBuilderABC[PrognosticState, TrainOutput, PairedData],
+    AggregatorBuilderABC[TrainOutput],
```
elynnwu, line 212: Can you explain why the other types are removed?

fme/ace/train/train.py
```diff
@@ -157,28 +174,43 @@ def inference(
         logging.info("Starting weather evaluation inference run")
```
elynnwu: This log is outdated right?

fme/ace/train/train.py
```diff
@@ -133,22 +136,36 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
+        return logs, logs["val/mean/loss"]
+
+    inference_epochs = config.get_inference_epochs()
+
+    def inference_callback(epoch: int) -> tuple[dict[str, Any], float | None]:
+        if epoch not in inference_epochs:
+            return {}, None
+        logs = inference_one_epoch(
+            stepper=stepper,
+            validation_context=contextlib.nullcontext,
```
elynnwu, line 159: [nit] add a short comment that Trainer wraps this externally

fme/ace/train/train.py
```diff
@@ -133,22 +136,36 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
+            validation_data=validation_data,
+            aggregator=aggregator,
+            diagnostics_subdir=f"epoch_{epoch:04d}",
+            record_logs=lambda logs: None,
+        )
+        return logs, logs["val/mean/loss"]
+
+    inference_epochs = config.get_inference_epochs()
+
+    def inference_callback(epoch: int) -> tuple[dict[str, Any], float | None]:
```
elynnwu, line 154: [nit] Since `inference_callback` and `_run_inference` have very similar args, it may be confusing to understand their difference. Consider adding a short doc here saying `inference_callback`: inline inference for checkpoint selection vs. `_run_inference`: additional inference for diagnostics

elynnwu: If this is going to change quickly in the follow-on PRs, please ignore

fme/ace/train/train.py
mcgibbon: Because now the generic Trainer no longer builds aggregators for validation or inference - these are handled by the caller in the callback function. Therefore the AggregatorBuilderABC generic no longer needs to know about the data types used during these methods.

fme/ace/train/train.py
mcgibbon: Deleted the argument instead, since now only nullcontext was being passed.

fme/ace/train/train.py
mcgibbon: I do think the current function names accurately describe what they do, but yes, mostly because this issue will disappear very soon either way, I will leave it as is.

FWIW, `inference_callback` is a callback because it accepts just the epoch and then does its own thing that isn't defined by the thing calling the callback - In other words, we use the general purpose callback to implement our own logic. On the other hand, `_run_inference` is a helper method that takes in arguments that define a very specific inference operation, and it runs that inference, so it's not a callback (the end of epoch callback that defines these args is the callback). Saying _run_inference does additional inference for diagnostics is kind of gratuitous context - it doesn't have anything specific to this in how its defined - making an "ideal" clarifying name a little more complicated.

I'd normally make at least some change to make this clearer, but given the complications and the fact that this will disappear very soon, I'll leave it as-is for now.

fme/ace/train/train.py
mcgibbon: Yes, pre-existing but I'll fix it. Moving to end_of_epoch ops and updating it.

fme/core/generics/trainer.py
```diff
@@ -149,7 +171,7 @@ def get_inference_epochs(self) -> list[int]: ...
 SD = TypeVar("SD")  # stepped data from inference
 
 
-class AggregatorBuilderABC(abc.ABC, Generic[PS, TO, SD]):
+class AggregatorBuilderABC(abc.ABC, Generic[TO]):
```
mcgibbon, line 177: PS and SD are no longer used by this class, so they can be removed.

fme/core/generics/trainer.py
```diff
@@ -206,14 +224,15 @@ def __init__(
         self,
         train_data: GriddedDataABC[BD],
         validation_data: GriddedDataABC[BD],
```
mcgibbon, line 229: Question: why is validation_data still used, now that validation is a callback?

mcgibbon: Answer: It's still used for a `.log_info` call, and it's used for LR tuning trials.

mcgibbon: I wonder if in follow-on I could use the validation callback for lr tuning, instead of the validation dataset.

elynnwu (approved):
I think you forgot to remove the log for weather evaluation inference run

PR #1118, opened 2026-05-04
Add `EMAConfig.resume_ema_ckpt_path`

jpdunc23:
Adds `resume_ema_ckpt_path` to `EMAConfig`. When pointed to a checkpoint with the saved EMA state (i.e., `ckpt.tar` or `ckpt_xxxx.tar`), restores the `EMATracker` state and number of updates from the checkpoint. The `decay` configuration is set by the training config.

Changes:
- Adds the `ema` property to `TrainConfigProtocol`. This property is already implemented by `TrainConfig` in `fme.ace` and `fme.coupled`.

- [x] Tests added

### DIFF
--- a/fme/core/ema.py
+++ b/fme/core/ema.py
@@ -54,12 +54,23 @@ class EMAConfig:
 
     Parameters:
         decay: decay rate for the moving average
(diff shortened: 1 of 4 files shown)


mcgibbon (approved):
LGTM

PR #1120, opened 2026-05-04
Ema context reentrant inside trainer

### LINKED ISSUE #1117: Additional inference with validate using ema
There's a bug with `validate_using_ema` + `additional inference`. Full log [here](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KQT8E1JE3WP4SZJ4NPZSJRDG/logs?jobId=01KQT8E1WV5T25PPQ2P52SQ93M).

With `validate_using_ema: true`, `Trainer.validation_context()` enters an EMA context. The trainer wraps `_end_of_epoch_callback` in `validation_context()`, and inside that callback `additional_inference` calls `inference_one_epoch`, which also opens `validation_context()`. The second EMA-context entry then hits `RuntimeError: Cannot nest EMA contexts`.

I'm able to reproduce this error in `fme/ace/test_train.py` with `validate_using_ema=True`.

elynnwu:
Make `Trainer._ema_context` reentrant so the EMA model is applied exactly once even when `validation_context()` is opened more than once on the same trainer. The fix is only for `_ema_context` when `_in_ema_context` is already True, so callers and library helpers can defensively wrap work in `validation_context()` without coordinating. The lower-level `EMATracker.applied_params` nesting guard is unchanged, so any direct misuse that would still error.

- [x] Tests added

Resolves #1117

### DIFF
--- a/fme/core/generics/trainer.py
+++ b/fme/core/generics/trainer.py
@@ -589,7 +589,16 @@ class Trainer:
     def _ema_context(self):
         """
         A context where the stepper uses the EMA model.
+
+        Reentrant: if we're already inside an EMA context (e.g. because
+        ``validation_context()`` was opened twice -- once by the trainer to
+        wrap an end-of-epoch callback, and once again by an inference helper
+        called from within that callback), nesting is a no-op so the EMA
+        weights are applied exactly once.
         """
+        if self._in_ema_context:
+            yield
+            return
         self._in_ema_context = True
         try:
             with self._ema.applied_params(self.stepper.modules):
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -119,6 +119,7 @@ def _get_test_yaml_files(
     time_buffer=1,
     use_time_length_probabilities=True,
     use_schedule=False,
+    validate_using_ema=False,
     derived_forcings=None,
 ):
     input_time_size = 1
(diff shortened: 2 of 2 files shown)


fme/core/generics/trainer.py
```diff
@@ -589,7 +589,16 @@ def validation_context(self):
...
+
+        Reentrant: if we're already inside an EMA context (e.g. because
+        ``validation_context()`` was opened twice -- once by the trainer to
+        wrap an end-of-epoch callback, and once again by an inference helper
+        called from within that callback), nesting is a no-op so the EMA
+        weights are applied exactly once.
         """
+        if self._in_ema_context:
+            yield
+            return
```
mcgibbon, line 601: I'm a little worried about this because when I see this context, I generally understand than when it's entered and after it exits we're not in an EMA context. This being needed kind of implies incorrectly written or confusing code. Like, in the example given it sounds like the Trainer is wrong to be forcing end-of-epoch callbacks in a validation context, since it seems we expect to be able to control that context when we write that code (and write them as though we have that control).

The other interpretation is to say we always want to do validation, inference, and end-of-epoch callbacks in a validation context (which I would be happy with, and is "safer" if we really don't need this control, which I don't think we do). If that's the case, we should be passing a null context (or letting the default be passed if it's a null context) to the validation_context argument of inference_one_epoch in train builder. This would avoid the need for callback setter functions like I ran into on #1116, since the callback itself wouldn't need to access the trainer - the trainer itself would already apply its validation context.

Could we raise an exception here instead, and make sure we never double-call this context?

mcgibbon (approved):
Approving this PR because it's a bug-fix without any issues (and with good test additions), and the improvement I suggest would be well suited for a follow-on PR. The "issue" is pre-existing.

PR #1121, opened 2026-05-04
Make `faster_decay_at_start` configurable in `EMAConfig`

jpdunc23:
Makes the `EMATracker` init arg `faster_decay_at_start` configurable via an attribute of the same name on `EMAConfig`. This is potentially useful for fine-tuning EMA checkpoints to keep the EMA state close to the initialization.

- [ ] Tests added

### DIFF
--- a/fme/core/ema.py
+++ b/fme/core/ema.py
@@ -57,9 +57,14 @@ class EMAConfig:
     """
 
     decay: float = 0.9999
+    faster_decay_at_start: bool = True
 
     def build(self, model: HasNamedParameters):
-        return EMATracker(model, decay=self.decay, faster_decay_at_start=True)
+        return EMATracker(
+            model,
+            decay=self.decay,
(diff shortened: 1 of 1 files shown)


fme/core/ema.py
```diff
@@ -53,21 +53,30 @@ class EMAConfig:
...
     decay: float = 0.9999
+    faster_decay_at_start: bool = True
     resume_ema_ckpt_path: str | None = None
 
     def build(self, model: HasNamedParameters):
-        ema = EMATracker(model, decay=self.decay, faster_decay_at_start=True)
+        ema = EMATracker(
+            model,
+            decay=self.decay,
+            faster_decay_at_start=self.faster_decay_at_start,
```
jpdunc23, line 78: Note that `faster_decay_at_start=False` is already covered by existing `EMATracker` tests.

mcgibbon (approved):
LGTM

PR #1123, opened 2026-05-05
Add TypedMetricInferenceEvaluatorAggregatorConfig

mcgibbon:
The current `InferenceEvaluatorAggregatorConfig` uses boolean flags (`log_video`, `log_histograms`, etc.) to control which metrics are computed during inference. This makes it difficult to add new metrics, configure per-metric options like variable filtering, or let users define custom metric sets in YAML. This PR adds a new `TypedMetricInferenceEvaluatorAggregatorConfig` that uses explicit typed metric config objects instead, while leaving the existing config class untouched for backwards compatibility. Higher-level configs (training and inference) accept a union of both types.

Changes:
- `fme.ace.aggregator.inference.main`: Add 12 typed metric config dataclasses (`MeanMetricConfig`, `StepMeanMetricConfig`, `PowerSpectrumMetricConfig`, `ZonalMeanMetricConfig`, `TimeMeanMetricConfig`, `VideoMetricConfig`, `HistogramMetricConfig`, `SeasonalMetricConfig`, `AnnualMetricConfig`, `EnsembleMetricConfig`, `EnsoIndexMetricConfig`, `EnsoCoefficientMetricConfig`), `MetricConfig` union type, `TypedMetricInferenceEvaluatorAggregatorConfig`, and supporting classes (`_MetricBuildContext`, `_MetricNotSupportedError`, `_VariableFilterAdapter`, `_maybe_filter`)
- `fme.ace.train.train_config.InlineInferenceConfig`, `fme.ace.train.train.AggregatorBuilder`, `fme.ace.inference.evaluator.InferenceEvaluatorConfig`: Accept union of old and new aggregator config types

- [x] Tests added


fme/ace/train/train_config.py
```diff
@@ -76,7 +81,7 @@ def __post_init__(self):
                 f"{self.loader.start_indices.n_initial_conditions} and "
                 f"{dist.world_size}."
             )
-        if (
+        if isinstance(self.aggregator, InferenceEvaluatorAggregatorConfig) and (
```
mcgibbon, line 84: Is it fine that we aren't doing this check on the new class?

mcgibbon: Good catch — it wasn't fine. The old config handles this via `__post_init__` mutation (setting the boolean flags to `False`), but the new config uses an explicit `enable_time_series` parameter on `build()` which wasn't being passed during inline training inference. Fixed in a24aaa5 by passing `enable_time_series=False` when building `TypedMetricInferenceEvaluatorAggregatorConfig` in `AggregatorBuilder.get_inference_aggregator()`.

fme/ace/aggregator/inference/main.py
```diff
@@ -53,6 +54,559 @@
...
+
+    def __init__(self, inner: SubAggregator, variables: Sequence[str]):
+        self._inner = inner
+        self._variables = frozenset(variables)
+
+    def record_batch(self, data: InferenceBatchData) -> None:
+        vs = self._variables
+        filtered = data.replace(
+            prediction={k: v for k, v in data.prediction.items() if k in vs},
+            prediction_norm={k: v for k, v in data.prediction_norm.items() if k in vs},
```
Arcomano1234: Question: Can data have a missing `data.prediction_norm` that would throw an error here?

fme/ace/train/train_config.py
```diff
@@ -8,6 +8,11 @@
 from fme.ace.aggregator import (
     InferenceEvaluatorAggregatorConfig,
     OneStepAggregatorConfig,
+    TypedMetricInferenceEvaluatorAggregatorConfig,
+)
+
+AnyAggregatorConfig = (
```
Arcomano1234: Nit: Can we move this to after the imports?

Arcomano1234 (commented):
Left a few specific comments but this looks pretty close. The other overall comment I have is `fme/ace/aggregator/inference/main.py` is now bloated with these different aggregator configs. Do we want to move them to a separate file?

mcgibbon:
> The other overall comment I have is fme/ace/aggregator/inference/main.py is now bloated with these different aggregator configs. Do we want to move them to a separate file?

Maybe they should go in the files where their aggregators are defined?

Arcomano1234:
> Maybe they should go in the files where their aggregators are defined?

I like that idea, I haven't looked at all of the individual aggregator files but assuming they can all go into them cleanly I think that is the best idea and makes `main.py `cleaner.

fme/ace/aggregator/inference/main.py
mcgibbon: I believe it can, we should add a try: except Unavailable to handle that case. Or use a `data.has_prediction_norm` attribute.

fme/ace/aggregator/inference/main.py
```diff
@@ -53,6 +54,559 @@
...
+@dataclasses.dataclass
+class SeasonalMetricConfig:
+    type: Literal["seasonal"] = "seasonal"
+    variables: list[str] | None = None
+    name: str = "seasonal"
+
+    def get_name(self) -> str:
+        return self.name
+
+    def build(self, ctx: _MetricBuildContext) -> SubAggregator:
```
mcgibbon: Let's put the _MetricBuildContext in its own file (along with the filtering, or that also in another file), and the metric configs in the files that define each metric. If it makes more sense we could put all the configs in one file, but it feels like these "live" with the built classes.

fme/ace/train/train_config.py
mcgibbon: Yes, let's do that.

fme/ace/aggregator/inference/main.py
```diff
@@ -23,35 +22,226 @@
...
+            initial_time=initial_time,
+        )
+
+        if self.metrics is not None:
+            metrics = list(self.metrics)
+        else:
+            metrics = self._default_metrics(ctx, n_ensemble_per_ic)
+
+        if not enable_time_series:
+            metrics = [m for m in metrics if not isinstance(m, MeanMetricConfig)]
```
elynnwu, line 207: Does this remove both the denorm and norm MeanMetricConfig entries entirely (not just time series)?

fme/ace/aggregator/inference/zonal_mean.py
```diff
@@ -315,3 +317,27 @@ def _initialize_zeros_zonal_mean_from_batch(
...
             for name, tensor in data.items()
         }
+
+
+@dataclasses.dataclass
+class ZonalMeanMetricConfig:
+    type: Literal["zonal_mean"] = "zonal_mean"
+    variables: list[str] | None = None
+    name: str = "zonal_mean"
+    zonal_mean_max_size: int | bool = 4096
```
elynnwu: Can this be type hinted as int only now that we control the aggregator from the metrics list?

fme/ace/aggregator/inference/main.py
Arcomano1234: Yeah I was little concerned about this as well and Claude said it didn't but yeah lets make sure this doesn't remove the mean metrics.

fme/ace/aggregator/inference/main.py
```diff
@@ -23,35 +22,226 @@
...
+            except MetricNotSupportedError:
+                if is_explicit:
+                    raise
+                logging.warning(
+                    f"{name} metric not supported for this grid type, " "omitting."
+                )
+                continue
+
+            aggregators[name] = agg
+            if isinstance(metric, MeanMetricConfig):
```
mcgibbon: Is it possible to avoid `isinstance` checks and the `type: ignore` statement below?

fme/ace/aggregator/inference/main.py
mcgibbon: If it doesn't, Claude should comment explaining how it doesn't. It's also possible this is fixed in #1126 which I believe adds a test for them giving the same logs.

fme/ace/train/train_config.py
mcgibbon: Claude: Done — moved `AnyAggregatorConfig` after all imports.

fme/ace/aggregator/inference/zonal_mean.py
mcgibbon: Claude: Good catch — changed `zonal_mean_max_size` from `int | bool` to `int` in `ZonalMeanMetricConfig`. The `bool` was only needed in the old config where `False` meant "don't log zonal means at all." In the typed config, you just omit `ZonalMeanMetricConfig` from the metrics list instead.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: This does remove both the denorm and norm `MeanMetricConfig` entries entirely, not just the time series. This matches the old config's behavior: when `InlineInferenceConfig.__post_init__` sets `log_global_mean_time_series=False` and `log_global_mean_norm_time_series=False`, the `MeanAggregator` objects are never created at all in the old `InferenceEvaluatorAggregatorConfig.build()`. The mean aggregators aren't needed without time series because they're excluded from `_summary_aggregators` (which filters out anything also in `time_series_aggregators`), so their only contribution is the per-step time series logs.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Addressed — introduced a `MetricBuildResult` dataclass in `data.py` that each metric config's `build()` now returns. Each config populates the appropriate slot(s) (`aggregator`, `time_series`, `ensemble`), so the build loop dispatches based on which slots are populated rather than using `isinstance` on the config type. This also eliminates the `type: ignore` since `time_series` is explicitly typed as `TimeSeriesLogs | None`.

fme/ace/aggregator/inference/main.py
mcgibbon: Ah, what we're missing is that what you're thinking of comes from the TimeMeanMetricConfig @elynnwu, not the MeanMetricConfig.

"mean" metric means global mean only, so its only output is a timeseries.

elynnwu (approved):

PR #1125, opened 2026-05-05
Fix shared memory handling for multi-worker epoch sync

mcgibbon:
Fix shared memory handling for epoch tensors in multi-worker data loading. The old code eagerly allocated shared memory for every dataset even without workers, and with ConcatDataset each sub-dataset created its own shared memory tensor instead of sharing one.

Changes:
- `fme.core.dataset.xarray.XarrayDataset`: defer shared memory allocation; add `enable_shared_memory()` and `set_global_epoch_tensor()` methods
- `fme.core.dataset.xarray.XarraySubset`: delegate new methods to wrapped dataset
- `fme.core.dataset.concat.XarrayConcat.enable_shared_memory`: create a single shared epoch tensor distributed to all sub-datasets
- `fme.ace.data_loading.getters.get_gridded_data`: call `enable_shared_memory()` when forkserver context is used

- [ ] Tests added (no new tests — existing 135 dataset tests pass)


jpdunc23 (approved):
LGTM!

PR #1126, opened 2026-05-05
Use typed metrics as primary InferenceEvaluatorAggregatorConfig

mcgibbon:
This is the second PR in a chain (after #1123) that completes the migration from boolean-flag to typed-metric configuration for the inference aggregator. It renames `TypedMetricInferenceEvaluatorAggregatorConfig` to `InferenceEvaluatorAggregatorConfig`, deletes the old boolean-flag class, and updates all references. The old class is then re-added as `LegacyFlagInferenceEvaluatorAggregatorConfig` for backwards compatibility with existing YAML configs.

Changes:
- `fme.ace.aggregator.inference.main.InferenceEvaluatorAggregatorConfig`: Rename from `TypedMetricInferenceEvaluatorAggregatorConfig`; this is now the primary config class
- `fme.ace.aggregator.inference.main.LegacyFlagInferenceEvaluatorAggregatorConfig`: Re-added old boolean-flag config under a new name for backwards compatibility
- `fme.ace.train.train_config.InlineInferenceConfig`, `fme.ace.train.train.AggregatorBuilder`, `fme.ace.inference.evaluator.InferenceEvaluatorConfig`: Simplified to use only `InferenceEvaluatorAggregatorConfig` (no longer a union type)
- `fme.coupled.aggregator.InferenceEvaluatorAggregatorConfig`: Translates boolean flags to typed metric configs via new `_build_metrics()` helper instead of passing them through to the old config

- [x] Tests added


fme/ace/aggregator/inference/main.py
```diff
@@ -639,27 +661,11 @@ def validate(self, n_forward_steps: int):
...
-                with a maximum of 32768 (2**15) (limited by matplotlib).
-        log_seasonal_means: Whether to log seasonal mean metrics and images.
-        log_global_mean_time_series: Whether to log global mean time series metrics.
-        log_global_mean_norm_time_series: Whether to log the normalized global mean
-            time series metrics.
-        monthly_reference_data: Path to monthly reference data to compare against.
-        time_mean_reference_data: Path to reference time means to compare against.
-        log_step_means: List of StepMeanEntry objects specifying steps at which
-            to log mean metrics.
+    Deprecated: Use InferenceEvaluatorAggregatorConfig with typed metrics instead.
```
mcgibbon, line 261: Instead of having its own build method, this config should construct a new-style config and use that to build. Can you write a regression test that ensures this is done correctly (you can use regression tools in fme.core.testing.regression), and make that update? We could delete the test as the last commit (with a commit that has it running in CI to show it passes).

fme/ace/aggregator/inference/test_evaluator.py
```diff
@@ -193,14 +201,7 @@ def test_inference_logs_labels_exist():
     ny = 45
     ds_info = get_ds_info(nx, ny)
     initial_time = (get_zero_time(shape=[n_sample, 0], dims=["sample", "time"]),)
-    agg = InferenceEvaluatorAggregatorConfig(
```
mcgibbon, line 196: Did the settings used in these tests change? If so, please justify each of the changes and why they improve the test as a comment on those lines. If they aren't justified, revert the changes.

fme/ace/aggregator/inference/test_evaluator.py
```diff
@@ -478,3 +523,78 @@ def test_raises_without_output_dir(self):
                 save_diagnostics=True,
                 output_dir=None,
             )
+
+
+def test_legacy_config_matches_typed_config():
```
mcgibbon, line 538: This is a key test.

fme/ace/train/train_config.py
```diff
@@ -81,18 +74,6 @@ def __post_init__(self):
                 f"{self.loader.start_indices.n_initial_conditions} and "
                 f"{dist.world_size}."
             )
-        if isinstance(self.aggregator, InferenceEvaluatorAggregatorConfig) and (
```
mcgibbon, line 93: These are replaced by the `enable_time_series=False` argument passed during build.

fme/ace/train/train_config.py
```diff
@@ -81,18 +74,6 @@ def __post_init__(self):
...
-            self.aggregator.log_global_mean_time_series
-            or self.aggregator.log_global_mean_norm_time_series
-        ):
-            # Both of log_global_mean_time_series and
-            # log_global_mean_norm_time_series must be False for inline inference.
-            self.aggregator.log_global_mean_time_series = False
-            self.aggregator.log_global_mean_norm_time_series = False
-
-        if isinstance(self.aggregator, InferenceEvaluatorAggregatorConfig):
-            for log_step_mean in self.aggregator.log_step_means:
```
mcgibbon, line 103: This is replaced by a build-time check, however, I don't think we've ever trained a case where this really matters.

fme/ace/aggregator/inference/main.py
```diff
@@ -639,27 +661,11 @@ def validate(self, n_forward_steps: int):
 
 
 @dataclasses.dataclass
-class InferenceEvaluatorAggregatorConfig:
+class LegacyFlagInferenceEvaluatorAggregatorConfig:
```
elynnwu, line 257: It doesn't look like this it being type hinted anywhere, so any config with the old style will break right? Should we use a custom dacite hook to build old configs?

fme/coupled/aggregator.py
```diff
@@ -231,6 +243,45 @@ class InferenceEvaluatorAggregatorConfig:
...
+        metrics.append(TimeMeanMetricConfig(target="norm"))
+        if self.log_video:
+            metrics.append(
+                VideoMetricConfig(enable_extended_videos=self.log_extended_video)
+            )
+        if self.log_histograms:
+            metrics.append(HistogramMetricConfig())
+        if self.log_seasonal_means:
+            metrics.append(SeasonalMetricConfig())
+        metrics.append(AnnualMetricConfig())
```
elynnwu: This annual metric is added unconditionally. Include the >2yr check

fme/coupled/aggregator.py
```diff
@@ -231,6 +243,45 @@ class InferenceEvaluatorAggregatorConfig:
...
+        if include_nino34:
+            metrics.append(EnsoIndexMetricConfig())
```
elynnwu: Missing `EnsoCoefficientMetricConfig` for coupled

fme/ace/aggregator/inference/test_evaluator.py
elynnwu, line 538: Do we want to also test long-run config here so things like annual and ENSO show up?

fme/ace/aggregator/inference/main.py
elynnwu: Also, should we add a deprecation warning anytime this is instantiated?

fme/ace/aggregator/inference/main.py
```diff
@@ -676,218 +682,102 @@ class InferenceEvaluatorAggregatorConfig:
...
-        n_forward_steps: int,
-        initial_time: xr.DataArray,
-        normalize: Callable[[TensorMapping], TensorDict],
-        output_dir: str | None = None,
-        channel_mean_names: Sequence[str] | None = None,
-        save_diagnostics: bool = True,
+        n_timesteps: int | None = None,
+        timestep: datetime.timedelta | None = None,
+        horizontal_coordinates: HorizontalCoordinates | None = None,
+        ops: GriddedOperations | None = None,
```
elynnwu: It looks like this is only called from build, can we remove type hint and default to None?

fme/ace/aggregator/inference/main.py
mcgibbon: We should update the configs to allow this type where we allow InferenceEvaluatorAggregatorConfig.

fme/coupled/aggregator.py
```diff
@@ -231,6 +247,49 @@ class InferenceEvaluatorAggregatorConfig:
...
+        if self.log_histograms:
+            metrics.append(HistogramMetricConfig())
+        if self.log_seasonal_means:
+            metrics.append(SeasonalMetricConfig())
+        if n_timesteps * timestep > APPROXIMATELY_TWO_YEARS:
+            metrics.append(AnnualMetricConfig())
+            if include_nino34:
+                metrics.append(EnsoIndexMetricConfig())
+        if n_timesteps * timestep > SLIGHTLY_LESS_THAN_FIVE_YEARS:
+            metrics.append(EnsoCoefficientMetricConfig())
```
elynnwu, line 290: Great! I think I realized why we probably didn't have it before since none of our datasets have the historical coefficients to regress on, but it is still good to have it. It'll just get skipped if there's no overlap time period.

elynnwu (approved):

PR #1128, opened 2026-05-05
Unify inference configs with name and weight for checkpoint selection

mcgibbon:
Replace `inference: InlineInferenceConfig | None` and `additional_inference: list[AdditionalInferenceConfig]` with a single `inference: list[InlineInferenceConfig]`, allowing multiple inference runs to inform checkpoint selection via a weighted error sum.

Changes:
- `InlineInferenceConfig`: add `name: str | None` and `weight: float` fields
- `TrainConfig`: replace `inference` and `additional_inference` fields with `inference: list[InlineInferenceConfig]`; add `inference_names` property for auto-naming (`"inference"` when single, `"inference_{i}"` when multiple); update `get_inference_epochs` for list; remove `inference_n_forward_steps` and `inference_aggregator` properties
- `TrainBuilders`: replace `get_evaluation_inference_data` and `get_additional_inference_data` with unified `get_inference_data`; remove `get_end_of_epoch_callback`
- `AggregatorBuilder`: remove inference-specific fields and `get_inference_aggregator` (aggregators now built inline in the callback)
- `build_trainer`: consolidate all inference runs into `inference_callback` with weighted error sum for checkpoint selection; remove `AdditionalInferenceConfig`, `_run_inference` helper, and end-of-epoch inference dispatch
- `fme.ace.__init__`: remove `AdditionalInferenceConfig` re-export

- [x] Tests added


fme/ace/train/train_config.py
```diff
@@ -66,8 +66,14 @@ class InlineInferenceConfig:
             log_global_mean_time_series=False, log_global_mean_norm_time_series=False
         )
     )
+    name: str | None = None
+    weight: float = 1.0
```
mcgibbon, line 77: Suggestion: add unit tests for `weight` validation — at minimum, that a negative value raises `ValueError` and that `weight=0.0` is accepted.

fme/ace/train/train_config.py
```diff
@@ -286,16 +269,16 @@ def train_evaluation_batches(self) -> int:
...
+    def inference_names(self) -> list[str]:
+        names = []
+        for i, entry in enumerate(self.inference):
+            if entry.name is not None:
+                names.append(entry.name)
+            elif len(self.inference) == 1:
+                names.append("inference")
+            else:
+                names.append(f"inference_{i}")
+        return names
```
mcgibbon, line 316: Two items here:

1. **Document the renaming behavior.** When there's a single unnamed config it gets `"inference"`, but if a second config is added both become `"inference_0"` / `"inference_1"`. This silently renames wandb metric keys and output directories. Worth a note in the docstring so users expect it.

2. **Add unit tests** for this property — it has non-trivial branching. Suggested cases:
   - single unnamed config → `["inference"]`
   - two unnamed configs → `["inference_0", "inference_1"]`
   - explicit names respected
   - mix of named and unnamed
   - duplicate name detection raises `ValueError`

fme/ace/train/train.py
```diff
@@ -149,46 +128,51 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
         return logs, logs["val/mean/loss"]
 
-    inference_epochs = config.get_inference_epochs()
+    # precompute per-config epoch sets for evaluate_before_training support
+    start_epoch = 0 if config.evaluate_before_training else 1
+    all_epochs = list(range(start_epoch, config.max_epochs + 1))
+    inference_epoch_sets = [
+        set(all_epochs[entry_config.epochs.slice])
+        for entry_config, _, _, _ in inference_entries
+    ]
```
mcgibbon: This `start_epoch` / `all_epochs` / per-config epoch set computation duplicates what `get_inference_epochs()` already does in `train_config.py`. Consider either:
- Having `get_inference_epochs` also return the per-config sets, or
- Extracting the shared `all_epochs` list into a helper

to avoid the two sites drifting apart.

fme/ace/train/train_config.py
```diff
@@ -355,21 +341,27 @@ def get_validation_data(self) -> GriddedData:
...
         self,
-    ) -> InferenceGriddedData:
-        if self.config.inference is None:
-            return ErrorInferenceData()  # type: ignore
-        else:
+        variable_metadata: Mapping[str, VariableMetadata],
+    ) -> list[tuple[InlineInferenceConfig, InferenceGriddedData, DatasetInfo, str]]:
+        names = self.config.inference_names
+        entries: list[
+            tuple[InlineInferenceConfig, InferenceGriddedData, DatasetInfo, str]
```
mcgibbon, line 388: `ErrorInferenceData` (in `fme/ace/data_loading/gridded_data.py`) is now dead code — its only import was in this file. Should be deleted in this PR.

mcgibbon (commented):
Nice simplification overall — collapsing `inference` + `additional_inference` into a single list is a clear improvement. A few items to address before merging.

Note: review generated by Claude

fme/ace/train/train_config.py
mcgibbon: Claude: Done — added `test_negative_weight_raises`, `test_zero_weight_accepted`, and `test_default_weight_is_one` in `fme/ace/train/test_train_config.py` (d7a2d83).

fme/ace/train/train_config.py
mcgibbon: Claude: Both addressed:

1. Added a note to the `name` docstring about the renaming behavior (d7a2d83).
2. Added unit tests covering all suggested cases: single unnamed, multiple unnamed, explicit names, mixed named/unnamed, empty, and duplicate name detection (d7a2d83).

fme/ace/train/train.py
mcgibbon: Claude: Extracted `get_inference_epoch_sets()` on `TrainConfig` which computes per-config sets. `get_inference_epochs()` now delegates to it, and `build_trainer` calls it instead of recomputing (9262f45).

fme/ace/train/train_config.py
mcgibbon: Claude: Deleted `ErrorInferenceData` and its helper types `PSType`/`FDType` (304ed1e).

mcgibbon:
Question for reviewers: Would we like to allow the old style in the case of single inference? i.e. would we like to allow inference to be of type `InlineInferenceConfig | list[InlineInferenceConfig]`?

Update: I went ahead with this

fme/ace/train/train.py
```diff
@@ -151,46 +131,43 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
-            dataset=inference_data,
-            aggregator=aggregator_builder.get_inference_aggregator(),
-            label="inference",
-            epoch=epoch,
-        )
-        error = logs.get("inference/time_mean_norm/rmse/channel_mean")
-        return logs, error
+        all_logs: dict[str, Any] = {}
+        weighted_error = 0.0
+        has_error = False
```
jpdunc23: Nit: `has_error` is slightly awkward, could `weighted_error: float | None = None` help?

fme/ace/train/train.py
```diff
@@ -114,29 +109,14 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
-        initial_inference_time=initial_inference_times,
-        n_ic_steps=stepper.n_ic_timesteps,
-        n_forward_steps=inference_n_forward_steps,
         loss_scaling=stepper.effective_loss_scaling,
         channel_mean_names=stepper.loss_names,
-        normalize=stepper.normalizer.normalize,
         save_per_epoch_diagnostics=config.save_per_epoch_diagnostics,
         validation_config=config.validation_aggregator,
-        n_ensemble_per_ic=n_ensemble_per_ic,
     )
```
jpdunc23, line 120: Nice to see all these args removed

fme/ace/train/train_config.py
```diff
@@ -410,62 +418,3 @@ def copy_after_batch():
 
             return copy_after_batch
         return lambda: None
-
-    def get_end_of_epoch_callback(
```
jpdunc23, line 420: I wasn't super familiar with the code here, but I think this was originally added with `weather_evaluation`? And then reused for `AdditionalInferenceConfig`?

Maybe the `EndOfEpochCallback` type could be useful for a generic `build_trainer` that would eliminate some of the duplication between `fme.ace.train` and `fme.coupled.train`. I see you haven't removed it from `trainer.py` (fine by me, though as far as I can tell it is now unused).

fme/ace/train/train_config.py
```diff
@@ -197,7 +193,9 @@ class TrainConfig:
     max_epochs: int
     save_checkpoint: bool
     experiment_dir: str
-    inference: InlineInferenceConfig | None
+    inference: InlineInferenceConfig | list[InlineInferenceConfig] = dataclasses.field(
```
jpdunc23, line 202: Nice to have this flexibility.

fme/ace/train/train.py
```diff
@@ -151,46 +131,43 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
+                dataset=data,
+                aggregator=aggregator,
+                label=name,
+                epoch=epoch,
+            )
+            all_logs.update(logs)
+            error = logs.get(f"{name}/time_mean_norm/rmse/channel_mean")
+            if error is not None:
+                weighted_error += entry_config.weight * error
+                has_error = True
```
jpdunc23: I didn't catch this myself, but agent review caught something potentially critical here (assuming the analysis is correct):

> Heterogeneous per-config epochs schedules silently break best_inference_ckpt selection

> The inference_callback returns weighted_error = Σᵢ wᵢ · errorᵢ summed only over entries that ran in this epoch:
> 
> 
> train.py
> Lines 140-170
> ```python
>         for i, (entry_config, data, entry_dataset_info, name) in enumerate(
>             inference_entries
>         ):
>             if epoch not in inference_epoch_sets[i]:
>                 continue
>             ...
>             error = logs.get(f"{name}/time_mean_norm/rmse/channel_mean")
>             if error is not None:
>                 weighted_error += entry_config.weight * error
>                 has_error = True
>         return all_logs, weighted_error if has_error else None
>```
> The trainer then compares this scalar across epochs:
> 
> 
> trainer.py
> Lines 697-709
> ```python
>             if inference_error is not None and (
>                 inference_error <= self._best_inference_error
>             ):
>                 ...
>                 self._best_inference_error = inference_error
>                 self.save_checkpoint(self.paths.best_inference_checkpoint_path)
> ```
> That's only valid if every epoch sums over the same set of entries. The new tests literally demonstrate the broken regime (test_get_inference_epoch_sets_per_config): with Slice(step=2) and Slice(step=3), epoch 1 sums two entries, epoch 3 sums one, epoch 4 sums the other. An epoch where only the cheaper/easier scenario runs almost mechanically wins best_inference_ckpt — not because the model improved, but because fewer terms got summed. 
>
> Mitigations the PR should pick: (a) require all weighted entries to share the same epochs (or weight=0 if not), (b) renormalize by sum of active weights per epoch, or at minimum (c) raise a hard error when epoch_sets differ for any entries with weight > 0.

I like option (a), personally, since it keeps flexibility of 0-weighted supplementary runs every N (possibly offset) epochs while ensuring the weighted inferences always run together.

fme/ace/train/train_config.py
```diff
@@ -66,8 +68,14 @@ class InlineInferenceConfig:
...
     )
+    name: str | None = None
+    weight: float = 1.0
 
     def __post_init__(self):
+        if self.weight < 0:
+            raise ValueError(
+                f"InlineInferenceConfig weight must be non-negative, got {self.weight}"
+            )
         dist = Distributed.get_instance()
```
jpdunc23, line 84: Another from agent review to consider:

> weight=0.0 is accepted and produces a permanent winner at inference_error = 0.0
> `__post_init__` allows weight=0.0 (and the new test test_zero_weight_accepted enshrines this), but the natural use case "log this scenario without influencing checkpoint selection" silently breaks selection:
> 
> If an epoch runs only zero-weight configs, weighted_error = 0.0, has_error = True → callback returns (logs, 0.0).
> 0.0 <= self._best_inference_error becomes True forever after, locking best_inference_ckpt at that epoch and blocking further updates.
> Compounding: in the wandb log, min(inference_error or torch.inf, ...) (trainer line 466) treats 0.0 as falsy and reports best_inference_error = inf, so the metric panel hides the bug.
> Either reject weight=0.0, or skip zero-weight entries entirely from the weighted sum (and from has_error).

fme/ace/train/train.py
jpdunc23: Another point from agent review, below. This one seems less common and unclear what the right thing to do is, but I think worth perhaps adding some docs or a warning regardless.

> Resume-after-config-edit silently corrupts _best_inference_error
> _best_inference_error is a single scalar persisted in the checkpoint and restored on resume:
> 
> 
> trainer.py
> Lines 775-776
> ```python
>     trainer._best_validation_loss = checkpoint["best_validation_loss"]
>     trainer._best_inference_error = checkpoint["best_inference_error"]
> ```
> Pre-PR this was a single inference's error; post-PR it's a weighted sum whose scale depends on len(inference_list), the chosen weights, and which aggregator keys exist. Any of these changing on resume (extremely common: add a second scenario mid-training, or bump a weight) silently shifts the metric scale. The most likely failure mode is that new errors land systematically above the saved _best_inference_error and best_inference_ckpt freezes at the pre-edit checkpoint with no warning. There is no schema/version stamp on this scalar to detect the change.
> 
> At minimum, write inference_names + weights into the checkpoint and reset _best_inference_error = inf (with a warning) when they don't match on resume.

fme/ace/train/train.py
```diff
@@ -151,46 +131,43 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
+            logs = inference_one_epoch(
+                stepper=stepper,
+                validation_context=contextlib.nullcontext,
+                dataset=data,
+                aggregator=aggregator,
+                label=name,
+                epoch=epoch,
+            )
+            all_logs.update(logs)
+            error = logs.get(f"{name}/time_mean_norm/rmse/channel_mean")
```
jpdunc23: Agent review. Too paranoid?

> Checkpoint metric depends on a hardcoded log-key string with no validation
> 
> This is a soft dependency on `InferenceEvaluatorAggregatorConfig` continuing to emit exactly that key. Two failure modes:
> 
> A user customizes the per-entry aggregator (allowed today — aggregator is a configurable field) in a way that drops `time_mean_norm` or `channel_mean`. That entry silently contributes 0 to weighted_error. If all entries miss the key, the callback returns (logs, None) and the trainer treats the epoch as "no inference ran" while wandb shows full inference metrics — confusing and silent.
> An aggregator refactor renames the key. Same silent breakage; checkpoint selection becomes a no-op with no test to catch it.
> This should be an explicit, documented, and validated contract: e.g., resolve the metric key once at config-validation time, error out if missing, and assert it appears in logs at runtime.

jpdunc23: Digging into this a bit more, this is partly a preexisting issue, which should be fixed in a separate PR; I created issue https://github.com/ai2cm/ace/issues/1136.

The issue specific to changes in this PR: "Silent partial sums". 

> With N inference entries summed into one `weighted_error`, any entry that fails to emit the expected log key is silently skipped while the others still flip `has_error = True`, so `best_inference_ckpt` can be updated based on a sum over an inconsistent subset of entries.

jpdunc23 (commented):
This looks great and will be a really nice addition. I'm definitely going to be using this feature and excited to finally have a way to break out of best_checkpoint_selection via time-mean RMSE only.

Agent review caught some issues that seem potentially critical: [this](https://github.com/ai2cm/ace/pull/1128/changes#r3198928718), [this](https://github.com/ai2cm/ace/pull/1128/changes#r3198933030), and [this one](https://github.com/ai2cm/ace/pull/1128/changes#r3199014697).

fme/ace/train/train.py
mcgibbon: Yeah I like the | None pattern.

fme/ace/train/train.py
mcgibbon: Yeah, I'd consider this a pre-existing issue, though it's true that not computing this metric does mean it can't contribute to the loss. I would suggest fixing this by a) requiring this aggregator be provided (maybe best for this PR) during config validation, or b) (maybe best for a follow-on PR) allowing configuration (that gets validated, perhaps) to set which metric determines checkpoint selection.

Thanks for documenting with an issue!

fme/ace/train/train.py
mcgibbon: > I like option (a), personally, since it keeps flexibility of 0-weighted supplementary runs every N (possibly offset) epochs while ensuring the weighted inferences always run together.

I like option a as well.

> Another point from agent review, below. 

I don't consider this an issue. We already don't guarantee that training on one commit and then "resuming training" on a different commit is valid. We only guarantee that checkpoints can be fine-tuned (e.g. with parameter initialization or your new features), and this comment isn't relevant for those cases.

In the case where this actually does come up, the current behavior is following the user's instructions, and I don't really see a user messing up those instructions (though even moreso, I don't see this case coming up at all).

fme/ace/train/train_config.py
mcgibbon: I would echo the suggestion that zero-weight entries shouldn't affect the "none-ness" of the inference metric.

fme/ace/train/train_config.py
mcgibbon: Correct, it's not used now. We use an "end of batch" callback for weight copying, rather than end-of-epoch.

> I see you haven't removed it from trainer.py (fine by me, though as far as I can tell it is now unused).

Yes, I did leave the trainer.py code for this even though it's unused, since it doesn't feel like it has much cost to maintain and feels "unified" with the end-of-batch callback. I'd also be fine with removing it, which I maybe "should" do.

> Maybe the EndOfEpochCallback type could be useful for a generic build_trainer that would eliminate some of the duplication between fme.ace.train and fme.coupled.train. 

Maybe - we can look at that if/when we're working on reducing duplication there (which we could do after these PRs). Though if neither one is using an end of epoch callback, it probably isn't warranted.

fme/ace/test_ice_train.py
```diff
@@ -216,19 +216,19 @@ def _save_ice_scalar_stats_netcdf(
               - LSNKc
               - XPRTc
 inference:
-  aggregator:
```
jpdunc23, line 219: Nit: I think this change can be reverted since the typing is `TrainConfig.inference: InlineInferenceConfig | list[InlineInferenceConfig]`.

On the other hand, the similar updates to the baseline YAML configs seems reasonable since those are often copy-pasted as a starting point for new config runs.

jpdunc23 (approved):
Looks good! Thank you for the fixes.

Maybe in a follow-up PR, I do think we should add logging of `"inference_error"` and not just `"best_inference_error"`, in addition to the individual inference metrics.

mcgibbon:
> Maybe in a follow-up PR, I do think we should add logging of "inference_error" and not just "best_inference_error", in addition to the individual inference metrics.

Would be happy to review/approve such a PR, but yeah it should happen in a follow-on.

fme/ace/test_ice_train.py
mcgibbon: I agree it could, but I intentionally had it make a commit that only allows `list[InlineInferenceConfig]` as an intermediate so that I could consistently update all the configs.

PR #1129, opened 2026-05-05
IPO TPI aggregator

elynnwu:
Adds an inference aggregator that computes the Tripole Index (TPI) for the Interdecadal Pacific Oscillation (IPO) from SST predictions and targets. The aggregator computes area-weighted SST anomalies over the three TPI regions (with NaN/ocean-mask handling), applies a 13-year Chebyshev low-pass filter, and reports the prediction-vs-target std ratio of the filtered TPI. It also plots the power spectrum of the unfiltered monthly TPI, where the spectrum is averaged across ensemble members and smoothed using 40 logarithmically-spaced frequency bins to preserve resolution at low frequencies where the IPO signal lives.

Changes:
- `fme.ace.aggregator.inference.ipo.ipo_index`: new IPOIndexAggregator computing TPI regional means, low-pass filtering, and std-normalized power spectrum diagnostics
- wire the IPO aggregator into the inference aggregator suite

- [x] Tests added


elynnwu:
Example job here: [wandb](https://wandb.ai/ai2cm/ace-samudra-coupled-cm4/runs/1b55e3k2?nw=nwuserelynnwu)

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+        data: Tensor of shape (sample, time, lat, lon), may contain NaN.
+        weights: Tensor of shape (lat, lon) with regional area weights.
+
+    Returns:
+        Tensor of shape (sample, time) with the NaN-excluded weighted mean.
+    """
+    valid_mask = ~torch.isnan(data)
+    data_filled = torch.where(valid_mask, data, torch.zeros_like(data))
+    # weights broadcast: (lat, lon) -> (1, 1, lat, lon)
+    w = weights.unsqueeze(0).unsqueeze(0)
```
jpdunc23: The weights here are not actually area weights, since the existing `LatLonRegion.regional_weights` just returns a binary tensor. This is fine for nino 3.4 index where the region has lat [-5, 5], but I wonder if it's more important to take into account here.

I think we can just update `LatLonRegion.regional_weights` to return the actual area weights on the region, as I don't expect this to greatly change the nino 3.4 index (though we should probably check as well).

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+        )
+
+    def _apply_filter_to_samples(self, tpi_da: xr.DataArray) -> xr.DataArray | None:
+        """Apply low-pass filter to each sample, trimming edge transients.
+
+        Trims one cutoff period from each end to remove filtfilt edge artifacts.
+        Returns None if the series is too short after trimming.
+        """
+        trim = int(self._cutoff_period_yrs * 12)
+        min_length = 3 * trim  # need at least one cutoff period after trimming
```
jpdunc23: I think we should probably just not report the IPO TPI index if we don't have enough data to do the filtering and therefore visualize the series, i.e. if the rollout is less than 39 years.

Thinking about this a bit more with agent review, I think this is a good point:

> A std ratio computed on 14 years of a 13-year-low-passed signal has essentially ~1 effective degree of freedom. The metric is dominated by which phase of the oscillation the run happens to sample, not by model skill. This makes the `ipo_tpi_std_ratio` metric nearly meaningless for `ICx1` runs with 40-year rollout, and potentially misleading if anyone tries to interpret it without this context. At minimum the metric should carry a caveat, but more seriously, the 30-year gate should probably be raised or the metric should only be reported for sufficiently long runs (e.g. 80+ years).

I think this becomes even more challenging when trying to interpret the spectrum, so maybe we should bump up the threshold to 80 years? That would give us more like 4+ clean IPO peaks / troughs, which seems a bit more meaningful.

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+                logs[f"{sst_name}_ipo_tpi_filtered"] = fig
+
+                std_ratio = _compute_sample_mean_std_ratio(
+                    pred_filtered.values, tgt_filtered.values
+                )
+                logs[f"{sst_name}_ipo_tpi_std_ratio"] = std_ratio
+
+            logs[f"{sst_name}_ipo_tpi_std_norm"] = _compute_sample_mean_std_ratio(
+                pred_da.values, tgt_da.values
+            )
```
jpdunc23: Is this the ratio of the unfiltered TPIs? I'm not sure this is particularly useful as an "IPO" metric since it includes so much high frequency variability.

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+            tgt_filtered = self._apply_filter_to_samples(tgt_da)
+
+            if pred_filtered is not None and tgt_filtered is not None:
+                fig = self._plot_filtered_tpi(pred_filtered, tgt_filtered)
+                logs[f"{sst_name}_ipo_tpi_filtered"] = fig
+
+                std_ratio = _compute_sample_mean_std_ratio(
+                    pred_filtered.values, tgt_filtered.values
+                )
+                logs[f"{sst_name}_ipo_tpi_std_ratio"] = std_ratio
```
jpdunc23: IMO we should only report the filtered std ratio, and call this `f"{sst_name}_ipo_tpi_std_norm"` to align with the naming of the equivalent nino3.4 metric.

Also, can we add `f"{sst_name}_ipo_tpi_std"` (i.e., the std of the filtered generated IPO TPI series without normalization by the target std).

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+            f"of {min_data_length} steps for IPO TPI power spectrum. The maximum "
+            f"input sample length is {max(data_lengths)} steps."
+        )
+    all_data = np.array(
+        [
+            data_array.isel(time=slice(0, min_data_length)).values
+            for data_array in data_arrays
+        ]
+    )
+    return compute_power_spectrum(all_data)
```
jpdunc23: This seems to be identical to the util in `enso/dynamic_index.py`. Maybe make the existing one public and reuse here?

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+        tgt_freq, tgt_power = _calculate_sample_average_power_spectrum(target_tpi)
+        pred_freq_s, pred_power_s = _smooth_spectrum(pred_freq, pred_power)
+        tgt_freq_s, tgt_power_s = _smooth_spectrum(tgt_freq, tgt_power)
+        fig, ax = plt.subplots(1, 1)
+        ax.plot(pred_freq_s, pred_power_s, label="predicted ensemble mean")
+        ax.plot(tgt_freq_s, tgt_power_s, label="target", color="orange")
+        ax.set_title("Power Spectrum of IPO TPI (unfiltered)")
+        ax.set_xlabel("Frequency [cycles/year]")
+        ax.set_ylabel("Power [K**2]")
+        ax.set(yscale="log")
```
jpdunc23: It would probably be good to use `ax.set(xscale="log")` here as well since we mostly care about the low frequency signals (< 0.077 cycles/year). 

I think we should also see what this looks like without applying `_smooth_spectrum()`. Agent review seemed to indicate this might do more harm than good. With the log-scaled x axis we will mostly ignore the high frequencies regardless, and can get a better view of it in the ENSO spectrum plot.

fme/ace/aggregator/inference/ipo/ipo_index.py
```diff
@@ -0,0 +1,445 @@
...
+    cutoff_freq = 1.0 / cutoff_period_yrs
+    wn = cutoff_freq / nyquist_freq
+
+    b, a = signal.cheby1(
+        N=filter_order, rp=passband_ripple_db, Wn=wn, btype="low", analog=False
+    )
+    return signal.filtfilt(b, a, data)
+
+
+def _compute_sample_mean_std_ratio(
```
jpdunc23: You can make public and reuse `_compute_sample_mean_std()` from dynamic_index.py here. When `target_data` is given, `_compute_sample_mean_std()` simultaneously returns the un-normalized and normalized stds.

jpdunc23 (commented):
Thanks for adding this! Will be really nice to have these metrics for longer rollouts.

Have a few suggestions. I think the main thing is to keep this focused on the decadal variability in the metrics and spectrum visualization, since the nino 3.4 index already gives a good look at the higher frequency modes.

Moving to a minimum of 80 years for rollouts might help as well.

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Returning the actual weights now, experiment on the way to check that nino 3.4 index doesn't change much

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Bumped to >80years.

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Updated to filtered.

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Updated and added the std of the filtered generated IPO TPI

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Moved to utils one level up

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Changed to log-scaled x axis

fme/ace/aggregator/inference/ipo/ipo_index.py
elynnwu: Moved to utils one level up

elynnwu:
Very little difference in Nino 3.4 index after returning the actual area weights ([wandb](https://wandb.ai/ai2cm/ace-samudra-coupled-cm4/workspace?nw=kkycuzewvlp)), see target values:

before
image

after
image

And using log x-axis works well for IPO power spectrum:
image

fme/ace/aggregator/inference/main.py
```diff
@@ -27,20 +27,22 @@
...
 from .spectrum import PowerSpectrumMetricConfig, SphericalPowerSpectrumAggregator
 from .time_mean import TimeMeanAggregator, TimeMeanMetricConfig
+from .utils import LatLonRegion
 from .video import VideoMetricConfig
 from .zonal_mean import ZonalMeanMetricConfig
 
 wandb = WandB.get_instance()
 APPROXIMATELY_TWO_YEARS = datetime.timedelta(days=730)
 SLIGHTLY_LESS_THAN_FIVE_YEARS = datetime.timedelta(days=1800)
+APPROXIMATELY_THIRTY_YEARS = datetime.timedelta(days=10950)
```
jpdunc23: I don't think this is used? Or `log_ipo_index` further down. 

Should this be `APPROXIMATELY_EIGHTY_YEARS` and block IPO computation for shorter runs?

jpdunc23 (commented):
Thanks for the updates! I think some changes to `main.py` may have gotten lost in the shuffle. Otherwise, should be ready to merge.

fme/ace/aggregator/inference/main.py
elynnwu: Yes, removed APPROXIMATELY_THIRTY_YEARS and added APPROXIMATELY_EIGHTY_YEARS

jpdunc23 (approved):
LGTM!

PR #1130, opened 2026-05-06
Reduce coupled training data I/O

jpdunc23:
Reduce coupled training data I/O by only loading atmosphere target data out to the maximum step that is used in the loss, based on the training config.

Changes:
- Add `TimePaddedMergedDataset`, similar to `MergedXarrayDataset`, which combines full-time-length targets (e.g., forcings) with short-time-horizon targets (e.g., prognostic / diagnostic variables needed for loss computation).
- Add `CoupledTrainDataRequirements` which separates forcing variable requirements from prognostic / diagnostic variable requirements. Add `get_train_window_data_requirements()` to `CoupledTrainStepperConfig` to build the `CoupledTrainDataRequirements` at training time.
- Add `get_train_dataset()` to `fme/coupled/data_loading/getters.py`. Builds a `TimePaddedMergedDataset` for atmosphere data.
- Fixed a pre-existing issue with `DatasetProperties.copy()` that led to copies sharing the same underlying variables metadata and label dictionaries.

- [x] Tests added


fme/core/dataset/merged.py
```diff
@@ -122,6 +123,159 @@ def set_epoch(self, epoch: int):
...
+        if len(combined_names) != len(set(combined_names)):
+            duplicates = list(
+                {item for item in combined_names if combined_names.count(item) > 1}
+            )
+            raise ValueError(
+                "Variable names must be unique across merged datasets. "
+                f"Duplicates found: {duplicates}"
+            )
+
+        self._sample_n_times = max(d.sample_n_times for d in self.datasets)
```
elynnwu, line 173: Agent review: the underlying `XarrayDatasets` has `_n_timesteps_schedule` that mutates `sample_n_times` at epoch milestones. So if a schedule is in use, there could be some failure modes. I don't believe the failure will be reached under the current call of `get_train_window_data_requirements`. But just to be safe, it could be good to raise an error in `CoupledTrainDataRequirements` post init for `atmosphere_target_requirements.n_timesteps`, `atmosphere_forcing_requirements.n_timesteps`, `ocean_requirements.n_timesteps` to always have constant `n_timesteps`.

```python
def _reject_nontrivial_schedule(field_name: str, n_timesteps: int | IntSchedule) -> None:
    if isinstance(n_timesteps, IntSchedule) and n_timesteps.milestones:
        raise ValueError(
            f"{field_name} must use a constant n_timesteps; "
            "TimePaddedMergedDataset caches the canonical sample length at "
            "construction and would break at the first milestone crossing. "
            f"Got milestones={n_timesteps.milestones}."
        )
```

fme/core/dataset/merged.py
```diff
@@ -122,6 +123,159 @@ def set_epoch(self, epoch: int):
...
+    """
+
+    def __init__(self, datasets: Sequence[DatasetABC]):
+        if len(datasets) == 0:
+            raise ValueError("Must provide at least one dataset.")
+        self.datasets = datasets
+
+        combined_names = [
+            item for dataset in self.datasets for item in dataset[0][0].keys()
+        ]
```
elynnwu, line 153: [optional] This is a pre-existing pattern in `MergedXarrayDataset`, but just to get the name you can get it through properties instead of forcing a sample 0 read. If you decide to update this, we should do it in `MergedXarrayDataset` too.

```python
combined_names = [
    name
    for dataset in self.datasets
    for name in dataset.properties.variable_metadata.keys()
]
```

fme/coupled/train/train_config.py
```diff
@@ -299,21 +306,25 @@ class TrainBuilders:
     def __init__(self, config: TrainConfig):
         self.config = config
 
-    def _get_train_window_data_requirements(self) -> CoupledDataRequirements:
+    def _get_train_window_data_requirements(self) -> CoupledTrainDataRequirements:
+        return self.config.stepper_training.get_train_window_data_requirements(
+            self.config.stepper
+        )
+
+    def _get_valid_window_data_requirements(self) -> CoupledDataRequirements:
```
elynnwu, line 366: Validation still loads atmosphere targets at the long horizon right? Is this intentional because validation needs the full target window?

fme/coupled/data_loading/getters.py
```diff
@@ -207,6 +225,89 @@ def get_gridded_data(
...
+        requirements.atmosphere_target_requirements.n_timesteps_schedule,
+    )
+    atmos_forcing, atmos_forcing_properties = config.atmosphere.build(
+        requirements.atmosphere_forcing_requirements.names,
+        requirements.atmosphere_forcing_requirements.n_timesteps_schedule,
+    )
+    # canonical = the long (forcing) horizon, since it provides the time
+    # coordinate that downstream code uses
+    atmosphere = TimePaddedMergedDataset([atmos_forcing, atmos_target])
+    atmosphere_properties = atmos_forcing_properties
```
elynnwu: [nit] use a `.copy()` here so that if something else uses `atmos_forcing_properties` it doesn't get updated

fme/coupled/data_loading/getters.py
```diff
@@ -207,6 +225,89 @@ def get_gridded_data(
...
+        requirements.ocean_requirements.n_timesteps_schedule,
+    )
+    atmos_target, atmos_target_properties = config.atmosphere.build(
+        requirements.atmosphere_target_requirements.names,
+        requirements.atmosphere_target_requirements.n_timesteps_schedule,
+    )
+    atmos_forcing, atmos_forcing_properties = config.atmosphere.build(
+        requirements.atmosphere_forcing_requirements.names,
+        requirements.atmosphere_forcing_requirements.n_timesteps_schedule,
+    )
```
elynnwu, line 266: Two `config.atmosphere.build` calls which calls `enable_shared_memory()` twice on the same underlying data. I'm not totally convinced that this is important, but agent seems to think that shard count matters especially due to `_get_raw_times` call. Can we time a typical `config.atmosphere.build(...)` to see if it's trivial?

elynnwu (commented):
I remember you showed some training samples per second comparison, can you attach to this PR?

jpdunc23:
> I remember you showed some training samples per second comparison, can you attach to this PR?

Sure, you can find it [here](https://docs.google.com/document/d/14lD9VhvEtfcWX25pJBiNjAzqnNFW5khLVeGD37PUKIs/edit?tab=t.0#heading=h.tj2qzwemr1gc).

fme/core/dataset/merged.py
jpdunc23: Updated to reset the canonical length with `TimePaddedMergedDataset.set_epoch()`

fme/coupled/train/train_config.py
jpdunc23: Yes, this doesn't touch validation.

fme/coupled/data_loading/getters.py
jpdunc23: Updated, and fixed existing bug in `DatasetProperties.copy()` to avoid sharing of the same underlying metadata dicts.

fme/coupled/data_loading/getters.py
jpdunc23: I haven't noticed any great slowdown in training initialization in these runs, e.g. [this run](https://beaker.org/orgs/ai2/workspaces/ace/work/01KRFRST726WS286KADG9J94PG/logs?jobId=01KRFRSTHC7F3RY6XVFMJH65XQ&recentLines=5000) took about 13 minutes from first log to "Starting Training Loop...", vs 10 minutes for [a run](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KQE6DP5009P6T7WFTNVJAZ6T/logs?jobId=01KQE6DPNC3FCKMTNDPH4T7SYY&recentLines=5000) without this feature.

fme/core/dataset/merged.py
jpdunc23: `variable_metadata` doesn't currently correctly reflect what variables are in the dataset, e.g. when variables are missing the expected metadata. Added issue https://github.com/ai2cm/ace/issues/1183 to track.

fme/core/dataset/merged.py
elynnwu: Agent: this fixes the single-process case but does not work with multiple workers.

> TimePaddedMergedDataset.set_epoch propagates the shared-tensor update (good — sub-datasets see the new epoch in workers) and recomputes its own local cache (only visible in the main process, since workers hold pickled copies). At a schedule milestone:
> 
> The main process sees consistent _sample_n_times / _canonical_idx.
> Each worker still has the pre-milestone cached values. When __getitem__ runs in the worker, sub-datasets report fresh sample_n_times via _ensure_epoch_synchronized, but self._sample_n_times is stale. _pad_tensors computes a negative pad_len exactly as before — same crash as the original bug, just reachable only under num_workers > 0.

elynnwu: Do we expect a use case where `n_timesteps` is a non-trivial `IntSchedule` with milestones? If not, maybe it's better to just raise an error at construction time?

fme/core/dataset/merged.py
```diff
@@ -122,6 +123,157 @@ def set_epoch(self, epoch: int):
...
+
+    def _pad_tensors(self, tensors: TensorDict, n_short: int) -> TensorDict:
+        if n_short == self._sample_n_times:
+            return tensors
+        pad_len = self._sample_n_times - n_short
+        padded: TensorDict = {}
+        for k, v in tensors.items():
+            pad_shape = list(v.shape)
+            pad_shape[0] = pad_len
+            nan_pad = torch.full(pad_shape, float("nan"), dtype=v.dtype)
```
elynnwu, line 200: Just double checking that this is never the case, if `v.dtype` is non-float, this will cast NaN to 0. I believe the mask variables are still float right?

fme/core/dataset/merged.py
jpdunc23: Updated again, should be fixed now following the pattern used by `XarrayDataset`. `IntSchedule` isn't used by `fme.coupled` so there isn't any risk, but in case future features use it `TimePaddedMergedDataset` should now be ready to do the right thing.

fme/core/dataset/merged.py
jpdunc23: Yes, masks are float32.

elynnwu (approved):

PR #1131, opened 2026-05-06
Update Beaker CI budget to ai2/atec-climate

mcgibbon:
The baseline configs were updated in #1080 to use the new budget name `ai2/atec-climate`, but the `ace-evaluator` Beaker CI workflow was missed and is now failing on `main` because the old `ai2/climate` budget name no longer works.

Changes:
- `.github/workflows/beaker.yaml`: update `--budget` from `ai2/climate` to `ai2/atec-climate`

- [ ] Tests added

No tests added: the workflow only runs on push to `main` or manual dispatch, so the fix can only be verified post-merge (or via `workflow_dispatch`).


AnnaKwa (approved):

mcgibbon:
Superceded by https://github.com/ai2cm/ace/pull/1134

PR #1133, opened 2026-05-06
Unify coupled inference configs with name and weight

mcgibbon:
Mirror the unified inference config changes from #1128 for the coupled training module, enabling multiple named and weighted inference runs for checkpoint selection.

Changes:
- `fme.coupled.train.train_config.InlineInferenceConfig`: add `name: str | None` and `weight: float` fields with validation
- `fme.coupled.train.train_config.TrainConfig`: change `inference` from required `InlineInferenceConfig` to optional `InlineInferenceConfig | list[InlineInferenceConfig]` (defaults to empty list); add `inference_list`, `inference_names`, `get_inference_epoch_sets` properties; remove `inference_aggregator` and `inference_n_coupled_steps` properties
- `fme.coupled.train.train_config.TrainBuilders`: replace `get_evaluation_inference_data` with `get_inference_data` returning list of `(config, data, name)` entries
- `fme.coupled.train.train.build_trainer`: refactor `inference_callback` to loop over entries with weighted error sum for checkpoint selection
- `fme.coupled.train.train.CoupledAggregatorBuilder`: remove inference-specific fields; aggregators now built inline per entry

- [x] Tests added

### DIFF
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -236,6 +236,8 @@ class TrainConfig:
             return self.stepper.to_stepper_config()
         return self.stepper
 
+    _RESERVED_NAMES = {"train", "val"}
+
     def __post_init__(self):
         if self.train_loader.using_labels != self.validation_loader.using_labels:
             raise ValueError(
@@ -245,6 +247,12 @@ class TrainConfig:
         resolved_names = self.inference_names
         if len(resolved_names) != len(set(resolved_names)):
             raise ValueError(f"Duplicate inference names: {resolved_names}")
+        reserved_overlap = set(resolved_names) & self._RESERVED_NAMES
+        if reserved_overlap:
+            raise ValueError(
+                f"Inference names {sorted(reserved_overlap)} collide with "
+                f"reserved names {sorted(self._RESERVED_NAMES)}"
+            )
         for i, entry in enumerate(self.inference_list):
             if self.train_loader.using_labels != entry.using_labels:
                 name = resolved_names[i]
--- a/fme/coupled/train/train.py
+++ b/fme/coupled/train/train.py
@@ -2,12 +2,11 @@ import contextlib
 import dataclasses
 import logging
 import os
-from collections.abc import Callable, Sequence
+from collections.abc import Sequence
 from typing import Any
 
 import dacite
 import torch
-import xarray as xr
 
 import fme
 from fme.core.cli import prepare_config, prepare_directory
@@ -15,12 +14,7 @@ from fme.core.derived_variables import get_derived_variable_metadata
 from fme.core.distributed import Distributed
 from fme.core.generics.trainer import AggregatorBuilderABC, Trainer, inference_one_epoch
 from fme.core.generics.validation import run_validation
-from fme.core.typing_ import TensorDict, TensorMapping
-from fme.coupled.aggregator import (
-    InferenceEvaluatorAggregatorConfig,
-    OneStepAggregator,
-    TrainAggregator,
-)
+from fme.coupled.aggregator import OneStepAggregator, TrainAggregator
 from fme.coupled.dataset_info import CoupledDatasetInfo
 from fme.coupled.stepper import CoupledTrainOutput
 from fme.coupled.train.train_config import TrainBuilders, TrainConfig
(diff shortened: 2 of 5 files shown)


fme/coupled/train/train.py
```diff
@@ -79,16 +66,46 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
-        )
-        error = logs.get("inference/time_mean_norm/rmse/channel_mean")
-        return logs, error
+        all_logs: dict[str, Any] = {}
+        weighted_error = 0.0
+        has_error = False
+        for i, (entry_config, data, name) in enumerate(inference_entries):
+            if epoch not in inference_epoch_sets[i]:
+                continue
+            batch = next(iter(data.loader))
```
mcgibbon, line 75: Claude: Note for reviewer — this `next(iter(data.loader))` call used to happen once during `build_trainer` setup (extracting `initial_inference_times` upfront). It now runs on every inference epoch inside the callback. This is fine in practice (inference loaders don't shuffle, and the cost is trivial vs. the full inference run that follows), but it's a behavioral change worth being aware of.

fme/coupled/train/train.py
```diff
@@ -109,30 +126,14 @@ class CoupledAggregatorBuilder(
...
         self,
-        inference_config: InferenceEvaluatorAggregatorConfig,
         dataset_info: CoupledDatasetInfo,
-        initial_inference_times: xr.DataArray,
-        n_timesteps_ocean: int,
-        n_timesteps_atmosphere: int,
         output_dir: str,
-        ocean_normalize: Callable[[TensorMapping], TensorDict],
-        atmosphere_normalize: Callable[[TensorMapping], TensorDict],
         loss_scaling: CoupledTensorMapping,
```
mcgibbon, line 131: Claude: Note for reviewer — the removed `ocean_channel_mean_names` / `atmosphere_channel_mean_names` parameters (and the removed `get_inference_aggregator` method) were never actually wired up: `build_trainer` on `main` never passed these to `CoupledAggregatorBuilder`, so they always defaulted to `None`. This is a clean dead-code removal with no behavior change.

fme/coupled/train/train_config.py
```diff
@@ -41,6 +38,13 @@ class InlineInferenceConfig:
             re-reading data from disk
         epochs: epochs on which to run inference. By default runs inference every epoch.
         aggregator: configuration of inline coupled inference aggregator.
+        name: name used as wandb log prefix and output subdirectory. If None,
+            defaults to "inference" when there is a single inference config
+            and "inference_{i}" when there are multiple. Note: adding a second
+            unnamed config will rename the first from "inference" to
+            "inference_0", changing its wandb keys and output directory.
```
elynnwu, line 86: I think it's better to keep the first one `inference` so that we can continue to compare with old jobs.

elynnwu: This is actually not an issue because you can configure the first inference to be "inference".

fme/coupled/train/train.py
```diff
@@ -79,16 +66,46 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
+                validation_context=contextlib.nullcontext,
+                dataset=data,
+                aggregator=aggregator,
+                label=name,
+                epoch=epoch,
+            )
+            all_logs.update(logs)
+            error = logs.get(f"{name}/time_mean_norm/rmse/channel_mean")
+            if error is not None:
+                weighted_error += entry_config.weight * error
```
elynnwu, line 105: I think this has the same issue documented in #1136, just flagging in here but we can address in a followup PR.

fme/coupled/train/train.py
```diff
@@ -79,16 +66,46 @@ def validation_callback(epoch: int) -> tuple[dict[str, Any], float]:
...
+            )
+            aggregator = entry_config.aggregator.build(
+                dataset_info=dataset_info,
+                n_timesteps_ocean=n_timesteps_ocean,
+                n_timesteps_atmosphere=n_timesteps_atmosphere,
+                initial_time=initial_times,
+                ocean_normalize=stepper.ocean.normalizer.normalize,
+                atmosphere_normalize=stepper.atmosphere.normalizer.normalize,
+                save_diagnostics=config.save_per_epoch_diagnostics,
+                output_dir=os.path.join(config.output_dir, name),
```
elynnwu, line 92: Do we want to add any checking for the specified name? While unlikely, if someone put `val` it would collide with the validation aggregator.

fme/coupled/train/train_config.py
```diff
@@ -93,8 +97,14 @@ class InlineInferenceConfig:
             log_global_mean_time_series=False, log_global_mean_norm_time_series=False
         )
     )
+    name: str | None = None
+    weight: float = 1.0
 
     def __post_init__(self):
+        if self.weight < 0:
```
elynnwu, line 104: I think we should also require sum of all weights to be > 0 right?

fme/coupled/train/train.py
elynnwu: Thanks for pointing this out. While the behavior doesn't change, there's possibly some overhead in each `iter(data.loader)` call. I still think it's trivial compared to inference so we probably can leave as it is.

fme/coupled/train/train_config.py
mcgibbon: Hmmm... Not necessarily. Like kind of yeah, but I could see it being valid to train a run where we aren't doing inference selection and set the weights to zero. I also just don't see this happening by accident - the default is 1, so you would have to intentionally set the weight for each config to 0.

If you like I can still update it, since you're the one using these configs?

fme/coupled/train/train.py
mcgibbon: Perhaps, though the validation aggregator name is also going to be configurable in a PR soon. Let's add some post-init configuration checks that make sure the names don't overlap across different inferences or with any other known top-level names (like "train" or "val"), and that check can get updated when we have the validation names, since they'll need the same kind of checking. I suppose this is at the TrainConfig level, so we need it in both coupled and ace training configs. It's a bit prone to error (if we update the code the checks might not get updated), but maybe this is ok since the effect would be obvious (merged wandb subtrees or filesystem errors).

fme/ace/train/train_config.py
```diff
@@ -236,6 +236,8 @@ def stepper_config(self) -> StepperConfig:
             return self.stepper.to_stepper_config()
         return self.stepper
 
+    _RESERVED_NAMES = {"train", "val"}
+
     def __post_init__(self):
```
mcgibbon, line 224: Claude: These are the only two names that need reserving — they are the hardcoded `label=` arguments passed to `aggregator.get_logs()` (which become wandb group prefixes like `train/...` and `val/...`) and the hardcoded output subdirectory name (`os.path.join(output_dir, "val")`). Other top-level wandb keys like `epoch`, `lr`, `best_val_loss`, etc. are flat scalars, not `/`-prefixed groups, so an inference name wouldn't merge with them in wandb or collide on disk.

fme/coupled/train/train_config.py
elynnwu: All good. Since the default is 1, I think this is unlikely to be accidentally misconfigured.

elynnwu (approved):

PR #1135, opened 2026-05-06
Allow any network type in SecondaryDecoderConfig

mcgibbon:
Removes the hard-coded restriction that limited `SecondaryDecoderConfig` (and the secondary decoder it builds) to the `MLP` network type, so any registered network can now be used. To support networks that need real dataset metadata (e.g. SFNO needs horizontal coordinates), `DatasetInfo` is now threaded from the parent step into the secondary decoder instead of substituting an empty placeholder.

Changes:
- `fme.core.step.secondary_decoder.SecondaryDecoderConfig`: drop the `_VALID_NETWORK_TYPES` allowlist and the `__post_init__` validator that rejected non-MLP types; `build` now takes a `dataset_info: DatasetInfo` argument.
- `fme.core.step.secondary_decoder.SecondaryDecoder.__init__`: now requires a `dataset_info` argument, which is forwarded to `network.build` instead of constructing an empty `DatasetInfo()`.
- `fme.core.step.single_module.SingleModuleStep` and `fme.core.step.secondary_module.SecondaryModuleStep`: pass the step's `dataset_info` when building the secondary decoder.
- `fme.core.step.test_secondary_decoder`: replace the "invalid network type" test with one asserting non-MLP types are accepted; update existing tests to pass `dataset_info`.

- [x] Tests added

---
_Generated by [Claude Code](https://claude.ai/code/session_01ErnQniFxAhoTCk3eeKzPf5)_

### DIFF
--- a/fme/core/step/secondary_decoder.py
+++ b/fme/core/step/secondary_decoder.py
@@ -12,8 +12,6 @@ from fme.core.registry import ModuleSelector
 from fme.core.registry.module import Module
 from fme.core.typing_ import TensorDict
 
-_VALID_NETWORK_TYPES = {"MLP"}
-
 
 @dataclasses.dataclass
 class SecondaryDecoderConfig:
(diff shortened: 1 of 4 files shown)


AnnaKwa (approved):
LGTM

PR #1138, opened 2026-05-07
Add finite difference CRPS loss for horizontal stochastic structure

mcgibbon:
Adds a `FiniteDifferenceCRPSLoss` that computes CRPS on spatial finite differences (latitude and longitude) of generated and target fields, encouraging the model to produce realistic horizontal stochastic structures. Supports multi-level coarsening: each additional level applies 2x average pooling before computing finite differences, so the loss captures structure at progressively coarser spatial scales. The total loss is averaged across levels.

Also adds an `almost_fair_crps_alpha` parameter to `EnsembleLoss` so the alpha used for both standard and finite-difference CRPS terms is configurable (defaults to 0.95, preserving existing behavior).

Changes:
- `fme.core.loss.FiniteDifferenceCRPSLoss`: new loss module computing CRPS on lat/lon finite differences with optional multi-level coarsening
- `fme.core.loss.EnsembleLoss`: add `finite_difference_crps_weight`, `finite_difference_crps_levels`, and `almost_fair_crps_alpha` parameters
- Tests for the new loss (zero loss on spatially constant fields, positive loss on random fields, multi-level, input validation)
- Training integration test exercises the new parameter

- [x] Tests added


fme/core/loss.py
```diff
@@ -383,28 +384,96 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+
 class EnsembleLoss(torch.nn.Module):
     def __init__(
         self,
         crps_weight: float,
         energy_score_weight: float,
         sht: Callable[[torch.Tensor], torch.Tensor],
+        finite_difference_crps_weight: float = 0.0,
+        finite_difference_crps_levels: int = 1,
+        almost_fair_crps_alpha: float = 0.95,
```
jpdunc23, line 445: Nit: Maybe `fair_crps_alpha`?

jpdunc23 (commented):
The results in your [wandb workspace ](https://wandb.ai/ai2cm/ace/workspace?nw=wipupcz48w) seem promising (I mostly compared train-amip-c96-shield-daily-e1c7d2-pos16-rsop to train-amip-c96-shield-daily-pos16-rsop, please lmk if there are other runs I should look at). How many levels did you use in the train-amip-c96-shield-daily-e1c7d2-pos16-rsop run?

Have you done a comparison of training throughput with and without this term using same type / number of GPUs? Curious to know if using this loss slows things down significantly, and if so whether there's anything we can do to optimize it.

mcgibbon:
I believe all of these runs are on Titan with only loss and learning rate differences, the purple/pink ones include a finite difference loss. Might be slightly slower, but it's hard to tell. Any optimizations are good regardless.

image

jpdunc23:
> I believe all of these runs are on Titan with only loss and learning rate differences, the purple/pink ones include a finite difference loss. Might be slightly slower, but it's hard to tell. Any optimizations are good regardless.

Thanks, ya I agree this doesn't look like a huge difference to be too worried about. Will keep an eye on it when trying this out for stochastic coupled ocean training.

fme/core/loss.py
```diff
@@ -383,28 +384,96 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        y_pooled = F.avg_pool2d(y_flat, kernel_size=2, stride=2, ceil_mode=True)
+        x_coarse = x_pooled.reshape(
+            *x.shape[:-2], x_pooled.shape[-2], x_pooled.shape[-1]
+        )
+        y_coarse = y_pooled.reshape(
+            *y.shape[:-2], y_pooled.shape[-2], y_pooled.shape[-1]
+        )
+        return level_crps + _get_finite_difference_crps_loss(
+            x_coarse, y_coarse, alpha=alpha, levels=levels - 1
+        )
```
jpdunc23, line 433: A couple of agent review comments:

> Pool-then-diff is partially redundant across levels
> At level k the loss is CRPS(diff(pool^k(x)), diff(pool^k(y))). Because pool is a low-pass filter and diff is a high-pass filter, what each level sees is roughly the band around scale 2^k. But this band is not orthogonal across levels — most of the level-0 signal also passes through level-1 (because the high-pass after one round of 2× smoothing still overlaps the level-0 high-pass band a lot). The result is a loss whose levels are highly correlated and whose 1/levels averaging cancels less variance than the cost suggests. You're paying compute for redundant gradient signal. A wavelet or Laplacian-pyramid decomposition would give nearly-orthogonal levels at the same cost. Worth at least benchmarking before committing to "levels" as a primary tuning knob.

and

> The finite_difference_crps_weight and finite_difference_crps_levels knobs are coupled
forward divides the recursive sum by self.levels, so each per-level term is implicitly weighted weight / levels. A user who tunes weight at levels=1 and then increases levels to 3 will silently reduce the magnitude of the level-0 contribution by 3×. This will surface as "loss got worse when I added levels," and people will retune the wrong knob. Either document this prominently in the parameter docstring, or normalize so increasing levels strictly adds signal at coarser scales (no rescaling of finer scales).

Unless you found that using `levels>1` led to better results in your experiments, it may be best to remove this feature for the time being. We can always revisit later if it isn't quite working yet.

fme/core/loss.py
```diff
@@ -383,28 +384,96 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
             raise ValueError(
                 "crps_weight and energy_score_weight must sum to a positive value, "
                 f"got {crps_weight} and {energy_score_weight}"
             )
-        self.crps_loss = CRPSLoss(alpha=0.95)
+        self.crps_loss = CRPSLoss(alpha=almost_fair_crps_alpha)
+        if finite_difference_crps_weight > 0:
+            self.diff_crps_loss: FiniteDifferenceCRPSLoss | None = (
+                FiniteDifferenceCRPSLoss(
+                    alpha=almost_fair_crps_alpha,
```
jpdunc23, line 467: Agent comment, curious to hear what you think:

> The PR description says defaulting almost_fair_crps_alpha to 0.95 preserves prior behavior. That's true for the standard CRPS path (was hardcoded CRPSLoss(alpha=0.95)), but the new FiniteDifferenceCRPSLoss has no prior behavior to preserve, and it inherits the same default. There's a strong argument the gradient-marginal CRPS should be fair (alpha=1) because finite differences halve the effective sample variance — the almost-fair correction was tuned for full fields, not their derivatives. PR #1139 already proposes flipping the standard-CRPS default to 1.0; either land that first, or expose two alpha knobs so the FD term doesn't get locked to whatever the standard term needs.

fme/core/loss.py
```diff
@@ -383,28 +384,96 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
         return get_crps(x, y, alpha=self.alpha).mean()
 
 
+class FiniteDifferenceCRPSLoss(torch.nn.Module):
+    """
+    Computes the CRPS of the x and y finite differences of the input tensors,
+    which helps with representations of horizontal stochastic structures.
```
jpdunc23, line 390: The statement "helps with representation of horizontal stochastic structures" is somewhat strong statement. In combination with the CRPS and energy score losses, it seems like it could be true, but do we have a reference or other evidence backing this up now?

jpdunc23 (approved):
Left a few more comments from agent review. Approving since this PR is self-contained and these are things we can revisit in later PRs, if needed, so feel free to merge as is.

fme/core/loss.py
mcgibbon: These two comments are not very self-consistent - they're kind of saying the code has two opposite effects, and it's also claiming both of those effects are bad. There may be a few too many adversarial terms in the prompt you're using for these reviews.

> But this band is not orthogonal across levels — most of the level-0 signal also passes through level-1 (because the high-pass after one round of 2× smoothing still overlaps the level-0 high-pass band a lot).

This is not entirely untrue, but it's an exaggeration. Consider a simple 1-D periodic case:
[ a, b, c, d]
the level=0 has a-d, b-a, c-b, and d-c differences. The level=1 has (c+d)-(b+a) differences, of which only (c-b) and (a-d) overlaps with level=0 - a 50% overlap. The effect of this is to over-count small-scale effects, but like, small scales do add up to large scales - this is accounting for that. I'm sure there is a more computationally efficient way to do this, but I'm not too concerned about performance-tuning of a loss term yet. If we saw it made our runs much slower and this is the low-hanging fruit for optimization, I could see it, but even then there are other optimizations to consider like computing the CRPS for all levels and the non-difference in the same C-loop.

> The finite_difference_crps_weight and finite_difference_crps_levels knobs are coupled
forward divides the recursive sum by self.levels, so each per-level term is implicitly weighted weight / levels.

This is intentional, I didn't want the magnitude of the loss to increase just because of the number of levels increasing, requiring LR re-tuning - I wanted this to be separately tunable.

> A user who tunes weight at levels=1 and then increases levels to 3 will silently reduce the magnitude of the level-0 contribution by 3×.

I mean, maybe (less than 3x because of the effect of the other comment), but also we're reducing the effect of CRPS overall by having any of this term, and we reduce the effect of CRPS by having energy score - but we don't worry about it because these are things we want to be optimizing.

> This will surface as "loss got worse when I added levels," and people will retune the wrong knob.

Actually this way will maintain the loss when adding levels. It could make CRPS/validation worse because of more optimization targets, but that isn't necessarily the case, or what we've seen so far with these types of loss inclusions.

fme/core/loss.py
mcgibbon: I also implemented this as fair CRPS for the finite differences originally and used fair CRPS for them in my experiments. I just don't think the more complex API is worth it for this minor detail, and refactored it in preparing this PR for a more critical review of the configuration and details. I do think we should switch to using 1.0 in both as soon as we can, though it's worth launching a a baseline first.

fme/core/loss.py
mcgibbon: The slides I linked in slack are empirical evidence. There's also strong theoretical evidence - previously the ensemble variance of stochastic structures did not enter the loss term at all (if using CRPS only), and an optimal model in terms of our loss had no guarantee of any kind of correct horizontal structure. Adding energy score requires the correct joint PDF of the spectral coefficients (to what extent isn't clear), but still doesn't strongly constrain local structures. Adding this term would mean having an optimal loss requires the PDF of the finite differences of channels to be accurate, which we see in the empirical result, and is a significant improvement.

fme/core/loss.py
jpdunc23: > There may be a few too many adversarial terms in the prompt you're using for these reviews.

Fair, ya may need to tone the agent down somewhat in this "threat analysis" style review prompt. I do think a well-calibrated adversarial approach can be helpful at times, but probably missing the mark here.

> These two comments are not very self-consistent - they're kind of saying the code has two opposite effects, and it's also claiming both of those effects are bad. 

Yes, I see your point: If there are redundant signals in the levels, then that will offset the averaging across levels somewhat. 

I do worry that the agent has a valid point here that the fine-scale skill may degrade somewhat in fine-tuning stages since that signal is discounted in the lose. If this isn't a case we plan to explore, then we probably shouldn't worry about it.

jpdunc23: Did you find that `levels > 1` was an improvement over the default?

mcgibbon: > Did you find that levels > 1 was an improvement over the default?

Yes, in exactly the way you might expect. Basically, this loss can only optimize structure at scales that are seen at the coarsest "level" of the loss. With levels=1, the small-scale structure is significantly improved, removing the artifacty-looking noise, but you still see scale-4-pixel blobs with larger structures more blobby than they should be. At levels=2, that significantly improves with coherent T0-time-difference structures twice as large, but it still wasn't perfect - I conjectured we should try with even more levels. For 4-degree runs we likely won't need as many levels, since what really matters is the receptive field size and that scales with gridcell size.

> I do worry that the agent has a valid point here that the fine-scale skill may degrade somewhat in fine-tuning stages since that signal is discounted in the lose.

There two parts of this, each of which I'm not too worried about. For grid-scale skill, I'm not too worried about this for two reasons: 1) We still separately keep the CRPS weight in the loss, so gridscale CRPS doesn't reduce in importance at all when levels are changed, and 2) we actually saw gridscale train and validation CRPS _improve_ when we added the finite difference CRPS, despite the CRPS having lower weighting in the loss. I conjectured this may be because these losses help bridge the pointwise CRPS and the global energy score.

For small-scale skill, yes it has a smaller contribution as more of the large scale is getting optimized, but the same can be said for computing the energy score loss at all scales and not just at small scales. The hierarchical dx/dy losses feel like they give a "natural" distribution of contribution across the scales they work on. Of course, if we see reduced grad_mag_percent_diff performance, it could be due to this effect, so we should keep it in mind.

mcgibbon: > I do think a well-calibrated adversarial approach can be helpful at times, but probably missing the mark here.

Where the value of finding issues or even understanding nuances of the choices made is high, I think it's especially helpful in a self-review where the cost of an inapplicable or incorrect concern is low. But it's more laborious in peer reviews where the author doesn't just have to identify that the concern doesn't apply, but also assess to what degree the reviewer thinks this concern applies, and use that assessment to diplomatically express their disagreement with the reviewer. I find that process takes a lot of mental energy, for me at least.

It helps a bit for the reviewer's stance to be clear, for example following something like the [conventional comments](https://conventionalcomments.org/) guidelines to indicate how they view the agent comment (a thought, a question, an issue, a suggestion? Optional?).

PR #1140, opened 2026-05-07
Additional ENSO specific metrics

elynnwu:
This PR adds ENSO specific metrics: integral of Nino34 power spectrum over 2-5 and 1-16yr band. 

Changes:
- `dynamic_index` now includes `compute_psd_band_power`

- [x] Tests added


elynnwu:
An earlier version of these metrics are in this [job](https://wandb.ai/ai2cm/ace-samudra-coupled-cm4/runs/h2bnnixw)

fme/core/generics/trainer.py
```diff
@@ -739,6 +780,58 @@ def save_all_checkpoints(self, valid_loss: float, inference_error: float | None)
             logging.info(f"Saving epoch checkpoint to {epoch_checkpoint_path}")
             self.save_checkpoint(epoch_checkpoint_path, include_optimization=True)
 
+    def _save_best_enso_checkpoint(
```
jpdunc23: It think it would be better to separate out the changes to `trainer.py` into a separate PR, since I think the checkpoint selection strategy here will probably need some broader discussion. On the other hand, it would be immediately useful to have the aggregator metrics added and start to get a sense of what these look like in our runs (which will also help with figuring out how to weight them).

fme/ace/aggregator/inference/enso/dynamic_index.py
```diff
@@ -379,6 +464,13 @@ def get_logs(self, label: str) -> dict[str, Any]:
                 ax.legend()
                 fig.tight_layout()
                 logs[f"{sst_name}_nino34_index_power_spectrum"] = fig
+                pred_psd = compute_psd_band_power(pred_freq, prediction_power_spectrum)
+                target_psd = compute_psd_band_power(target_freq, target_power_spectrum)
+                logs[f"{sst_name}_nino34_index_psd_2_5yr"] = pred_psd
```
jpdunc23: Maybe `f"{sst_name}_nino34_index_power_2_5yr"`?

jpdunc23 (commented):
Thank you for adding this! 

As we discussed, let's remove the "autocorr" metrics and changes to `Trainer` / `TrainConfig`, for now. Otherwise, should be good to go.

fme/core/generics/trainer.py
elynnwu: Removed ckpt selection related code

fme/ace/aggregator/inference/enso/dynamic_index.py
elynnwu: Updated

jpdunc23 (approved):
LGTM!

PR #1141, opened 2026-05-07
Config-time validation of coupled `LossContributionsConfig.n_steps`

jpdunc23:
Check the compatibility of user-configured per-component `LossContributionsConfig.n_steps` against the configured value of `n_coupled_steps`, and throw an error when config objects are built.

Changes:
- Add `is_null` and `n_steps_max` helper properties to `LossContributionsConfig`.
- Change `LossContributionsConfig.build()` arg name `max_n_steps` to `n_steps_limit` to disambiguate from the new `n_steps_max` property.
- Change `LossContributionsConfig.n_steps` type from `TimeLengthProbabilities | float = float("inf")` to `TimeLengthProbabilities | int | None = None` to avoid float-typed config of `n_steps`, with the same default meaning of "use all available steps" for loss contribution.

- [x] Tests added


fme/coupled/loss.py
elynnwu: Update to `n_steps_limit`

elynnwu: Update to `n_steps_limit`

fme/coupled/test_loss.py
```diff
@@ -404,6 +427,21 @@ def test_coupled_stepper_train_loss_sample_n_steps_delegates():
...
 
 
+@pytest.mark.parametrize(
+    "config_kwargs, expected_is_null",
+    [
+        ({}, False),
+        ({"n_steps": 0}, True),
+        ({"weight": 0.0}, True),
+        ({"n_steps": 5.0}, False),
+        ({"n_steps": 0.0}, True),
```
elynnwu: I think float is fine but we technically type hint this as `int | None`, unless this test is meant to show that float also works?

fme/coupled/loss.py
```diff
@@ -73,27 +73,60 @@ class LossContributionsConfig:
...
     optimize_last_step_only: bool = False
 
+    @property
+    def is_null(self) -> bool:
+        """True when this config produces a ``NullLossContributions`` (i.e.
+        contributes nothing to the loss).
+        """
+        if self.weight == 0.0:
+            return True
+        return self.n_steps == 0
```
elynnwu: Unlikely to be an issue, but technically if someone uses `n_steps = TimeLengthProbabilities(outcomes=[(steps=0, probability=1.0)])`, this will return False. Returning `self.n_steps_max == 0` could be better

elynnwu (approved):

PR #1142, opened 2026-05-08
Add Claude pre-review skill

mcgibbon:
Add instruction file for Claude to perform PR pre-reviews, making PRs easier for human reviewers with limited time.

Changes:
- `.claude/skills/pre-review/SKILL.md` added with instructions for Claude to pre-review PRs before human review

- [ ] Tests added


.claude/skills/pre-review/SKILL.md
```diff
@@ -0,0 +1,23 @@
...
+Issues involving readability might be particularly important, especially if the changes are hard to understand or the reason for them is not immediately clear.
+
+Begin by taking a look at the PR, as well as its description and title, and considering areas for improvement.
+Communicate with the user about whether:
+ - The PR has significant structural issues that should be fixed with significant refactors of the PR.
+ - The PR should be significantly restructured, for example by restructuring the commits to have a more logical flow or by splitting into multiple smaller PRs.
+ - The PR structure is acceptable, but there are specific issues that should be fixed before review.
+ - The PR is in good shape, perhaps you have some comments to make on the PR or updates to its title/description but the code is ready for the reviewer.
+
+When leaving comments on a PR, always start your comment with "Claude: ".
```
jpdunc23: It would be good IMO to keep this sort of thing harness-agnostic, e.g.
```suggestion
When leaving comments on a PR, always start your comment with "Pre-review agent: ".
```

jpdunc23 (commented):
I wasn't familiar with the `instruct/` folder and a quick search turned nothing up. 

In terms of organizing these skill files, I think we should try to use the [agentskills.io open specification](https://agentskills.io/client-implementation/adding-skills-support#what-to-scan-for) started by Anthropic:

```
.agents/
    skills/
        pre-review/
            SKILL.md
```

mcgibbon:
> agentskills.io open specification

Have you used this before / do you know how it works? I read the page and looked for claude code interoperability for skills for a few minutes, and while there was a lot of content, I'm still not clear on how to actually access skills from the client. As in, I don't know what I would enter into the terminal to use this. I was just going along the lines of prompting "Follow the instructions in instruct/pre-review.md" which is more immediately obvious, and worked quite well. This kind of mirrors the experience I had with MCP and "letting the agent handle it" being the easier solution.

What does this buy us? From what I can tell, the main benefit of this is avoiding loading code into context when you want to execute it as part of a skill (which isn't relevant here) or Claude accessing it as part of doing something you've asked it to do (which I don't really want to happen if I haven't explicitly invoked it). I'm also not sure it's worth having the skill name and description in our context for the agent to be able to auto-trigger it.

When I asked Claude to turn the script into a skill, it made it in `.claude`, and said it doesn't support `.agents`:
```
❯ Do you support skills in an .agents/ directory?

● claude-code-guide(Check .agents/skills support in Claude Code)
  ⎿  Done (2 tool uses · 22.0k tokens · 11s)
  (ctrl+o to expand)

● No — Claude Code only scans .claude/skills/, ~/.claude/skills/, and plugin directories. It doesn't
  currently scan .agents/skills/ despite that being the cross-client interoperability convention from
  the Agent Skills spec.

  If you want the skill available at both paths, a symlink would work (ln -s ../.claude/skills
  .agents/skills), but the native location for Claude Code is .claude/skills/ which is where I put it.
```

I'm a little annoyed that when I ran it as a skill, I was dropped in a multiple choice form that I had to confirm my answers to - I would have rather just continued the conversation. From this I'm a little worried the skills integration has other hidden optimizations that make it better suited to following a set of instructions - I specifically wanted to avoid a "workflow", and let Claude have more leeway on this one.

I do kind of like being able to just run `/pre-review` and get this going, since I mostly don't want to add more context to it, but it's also kind of nice to be able to add more context with something like "Follow the instructions in instruct/pre-review.md, I'm particularly concerned the tests have a lot of copy-paste but that might be the best way to go.".

Hmm... I did find a way to engage with it more naturally now - if I say "Follow the pre-review skill" instead of `/pre-review` I get a natural response. Maybe we can go ahead with that.

All in all I'm kind of wary of using this as a skill, but we can try it out for a while. `.agents/` sounds like it will not work.

.claude/skills/pre-review/SKILL.md
```diff
@@ -0,0 +1,26 @@
+---
+name: pre-review
+description: Prepare a PR for review by a team member. Use when the user wants to get a PR ready for review, make a PR easier to review, do a pre-review pass, or clean up a PR before requesting review. Also use when the user mentions preparing changes for a teammate to look at, or wants help making their branch review-ready.
```
mcgibbon:
```suggestion
description: Prepare a PR for review by a team member. Use when the user asks you to perform a pre-review.
```

.claude/skills/pre-review/SKILL.md
mcgibbon: I'm a little wary of this cuing it to be less natural, but we can try it.

jpdunc23:
> All in all I'm kind of wary of using this as a skill, but we can try it out for a while. .agents/ sounds like it will not work.

OK, looks like `.claude/skills` also [works fine with Cursor](https://cursor.com/docs/skills#skill-directories). Would be nice if Anthropic supported open standards, but why would they, right?

jpdunc23 (approved):

Issue #1143, opened 2026-05-08
`fme/coupled/test_train.py::test_train_and_inference[3-True]` flakiness


jpdunc23:
This test case randomly fails in CI, particularly in the GPU test suite. Re-running the CI job is usually enough to get the test passing.

jpdunc23:
This is still failing with #1157, e.g. see [here](https://github.com/ai2cm/ace/actions/runs/25748552038/job/75618370382?pr=1130).

PR #1144, opened 2026-05-08
Use `fme.core.rand.randn()` for SFNO noise

jpdunc23:
Switch to using `fme.core.rand.randn()` which respects the `use_cpu_randn()` context manager, rather than direct `torch.randn()`. 

Changes:
- Update the `fme.coupled` training integration tests with `use_cpu_randn()`. Hopefully this will address issue https://github.com/ai2cm/ace/issues/1143.

- [x] Tests added


mcgibbon (approved):
LGTM

PR #1147, opened 2026-05-08
Install cuhpx weights

yikwill:
Currently, the `.fits` and `.npy` files in `ace/fme/core/cuhpx/data/` which are used for the `cuhpx` quadrature weights are not included when installing the repository using `pyproject.toml`. This causes downstream errors for things like computing spectral transforms of HEALPix data during inline inference. This PR ensures these weight files are shipped with the rest of the package.


oliverwm1 (approved):
thanks @yikwill

PR #1148, opened 2026-05-11
Add configurable label embedding dimension to NoiseConditionedSFNO

mcgibbon:
Replace direct one-hot label usage in noise-conditioned SFNO with an optional learned linear embedding. When `label_embed_dim > 0`, a shared `nn.Linear(n_labels, label_embed_dim)` maps one-hot labels to a lower-dimensional embedding space before they reach `ConditionalLayerNorm` and positional interaction layers. When `label_embed_dim = 0` (default), one-hot labels pass through directly, preserving existing behavior and checkpoint compatibility for non-label models.

Changes:
- `NoiseConditionedModel`: replace `embed_dim_labels` parameter with `n_labels` and `label_embed_dim`; add optional `nn.Linear` embedding layer applied in `forward()` before label-position interaction and `Context` construction
- `NoiseConditionedSFNOBuilder`: add `label_embed_dim: int = 0` config field; compute `effective_label_dim` for `ContextConfig` based on whether embedding is enabled
- `LocalNetBuilder.build()`: update `NoiseConditionedModel` call to use `n_labels` kwarg (no embedding support, one-hot labels used directly)
- `Context`: update docstring to reflect labels may be one-hot or learned embeddings
- Regenerate `NoiseConditionedSFNO` backwards-compatibility test checkpoint with `label_embed_dim=3`


fme/ace/registry/local_net.py
mcgibbon, line 1: Note this file uses the NoiseConditionedModel wrapper from stochastic_sfno.py, which is why you don't see any particular application of the embedding in this file.

Arcomano1234 (approved):
LGTM, the only comment is can you add this functionality to `AnkurLocalNetBuilder`? If we are not in favor of supporting new features `AnkurLocalNet` we should think about what we want to do with that code.

mcgibbon:
> LGTM, the only comment is can you add this functionality to `AnkurLocalNetBuilder`? If we are not in favor of supporting new features `AnkurLocalNet` we should think about what we want to do with that code.

I think this is more scientifically difficult than programmatically difficult. Ankur's network doesn't have any FiLM conditioning, it completely ignores labels right now.

PR #1149, opened 2026-05-11
Refactor validation to support multiple weighted configs

mcgibbon:
Refactors validation in both ACE and coupled training to support multiple weighted validation configs, mirroring the existing `InlineInferenceConfig` pattern. Each validation entry can have its own data loader, aggregator, epoch schedule, name, and weight. The weighted sum of validation losses drives checkpoint selection and LR scheduling.

Changes:
- `fme.ace.train.train_config.InlineValidationConfig`: new dataclass with `loader`, `aggregator`, `epochs`, `name`, `weight`
- `fme.ace.train.train_config.TrainConfig`: replace `validation_loader`/`validation_aggregator` with `validation: InlineValidationConfig | list[InlineValidationConfig]`; add `validation_list`, `validation_names`, `get_validation_epoch_sets()`, `get_validation_epochs()`
- `fme.ace.train.train.build_trainer`: validation callback iterates over entries, returns weighted loss
- `fme.coupled.train.train_config.InlineValidationConfig`: coupled equivalent with `loader`, `epochs`, `name`, `weight`
- `fme.coupled.train.train_config.TrainConfig`: same refactor as ACE — `validation_loader` → `validation`
- `fme.coupled.train.train.build_trainer`: coupled validation callback iterates over entries
- All baseline YAML configs updated (`validation_loader:` → `validation:\n- loader:`)

- [x] Tests added

### DIFF
--- a/configs/baselines/amip-c96-shield/train-ace2.yaml
+++ b/configs/baselines/amip-c96-shield/train-ace2.yaml
@@ -52,20 +52,21 @@ train_loader:
       - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0002
         subset:
           start_time: '2011-01-01'
-validation_loader:
-  batch_size: 128
-  num_data_workers: 8
-  prefetch_factor: 4
-  dataset:
-    concat:
-      - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0001
-        subset:
-          start_time: '1996-01-01'
-          stop_time: '2000-12-31'
-      - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0002
-        subset:
-          start_time: '1996-01-01'
-          stop_time: '2000-12-31'
+validation:
+  loader:
+    batch_size: 128
+    num_data_workers: 8
+    prefetch_factor: 4
+    dataset:
+      concat:
+        - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0001
+          subset:
+            start_time: '1996-01-01'
+            stop_time: '2000-12-31'
+        - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0002
+          subset:
+            start_time: '1996-01-01'
+            stop_time: '2000-12-31'
 optimization:
   enable_automatic_mixed_precision: false
   lr: 0.0001
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -57,7 +57,12 @@ from fme.ace.testing import (
 )
 from fme.ace.train.train import build_trainer, prepare_directory
 from fme.ace.train.train import main as train_main
-from fme.ace.train.train_config import InlineInferenceConfig, TrainBuilders, TrainConfig
+from fme.ace.train.train_config import (
+    InlineInferenceConfig,
+    InlineValidationConfig,
+    TrainBuilders,
+    TrainConfig,
+)
 from fme.core.coordinates import (
     HEALPixCoordinates,
     HorizontalCoordinates,
@@ -317,14 +322,20 @@ def _get_test_yaml_files(
             time_buffer=time_buffer,
             sample_with_replacement=10,
         ),
-        validation_loader=DataLoaderConfig(
-            dataset=XarrayDataConfig(
-                data_path=str(valid_data_path),
-                spatial_dimensions=spatial_dimensions_str,
-                labels=["era5"] if conditional else None,
+        validation=InlineValidationConfig(
+            loader=DataLoaderConfig(
+                dataset=XarrayDataConfig(
+                    data_path=str(valid_data_path),
+                    spatial_dimensions=spatial_dimensions_str,
+                    labels=["era5"] if conditional else None,
+                ),
+                batch_size=2,
+                num_data_workers=0,
             ),
-            batch_size=2,
-            num_data_workers=0,
+            aggregator=OneStepAggregatorConfig(
+                log_snapshots=log_validation_maps,
+                log_mean_maps=log_validation_maps,
+            ),
         ),
         optimization=OptimizationConfig(
             use_gradient_accumulation=True,
@@ -382,10 +393,6 @@ def _get_test_yaml_files(
         logging=logging_config,
         experiment_dir=str(results_dir),
         save_per_epoch_diagnostics=save_per_epoch_diagnostics,
-        validation_aggregator=OneStepAggregatorConfig(
-            log_snapshots=log_validation_maps,
-            log_mean_maps=log_validation_maps,
-        ),
     )
 
     inference_config = InferenceEvaluatorConfig(
@@ -1088,7 +1095,7 @@ def test_train_with_non_local_experiment_dir_error():
             experiment_dir=non_local_experiment_dir,
             stepper=stepper,
             train_loader=dummy_data_loader,
-            validation_loader=dummy_data_loader,
+            validation=InlineValidationConfig(loader=dummy_data_loader),
             optimization=OptimizationConfig(),
             logging=LoggingConfig(),
             max_epochs=1,
@@ -1109,7 +1116,7 @@ def test_train_config_with_checkpoint_stepper(tmp_path: pathlib.Path):
         stepper=CheckpointStepperConfig(checkpoint_path=str(checkpoint_path)),
         stepper_training=TrainStepperConfig(n_forward_steps=2),
         train_loader=dummy_data_loader,
-        validation_loader=dummy_data_loader,
+        validation=InlineValidationConfig(loader=dummy_data_loader),
         optimization=OptimizationConfig(),
         logging=LoggingConfig(),
         max_epochs=1,
@@ -1152,7 +1159,7 @@ def test_train_config_with_stepper_config_sets_stepper_config(tmp_path: pathlib.
         stepper=stepper,
         stepper_training=TrainStepperConfig(n_forward_steps=2),
         train_loader=dummy_data_loader,
-        validation_loader=dummy_data_loader,
+        validation=InlineValidationConfig(loader=dummy_data_loader),
         optimization=OptimizationConfig(),
         logging=LoggingConfig(),
         max_epochs=1,
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -610,14 +610,14 @@ def _setup(
 
 @pytest.mark.parametrize(
     "nettype, crps_training, log_validation_maps, \
-        use_healpix, use_schedule, validate_using_ema",
+        use_healpix, use_schedule, validate_using_ema, multi_validation",
     [
-        ("NoiseConditionedSFNO", True, False, False, True, False),
-        ("SphericalFourierNeuralOperatorNet", False, True, False, False, False),
-        ("HEALPixRecUNet", False, False, True, False, False),
-        ("Samudra", False, False, False, False, False),
-        ("NoiseConditionedSFNO", False, False, False, False, False),
-        ("NoiseConditionedSFNO", True, False, False, True, True),
+        ("NoiseConditionedSFNO", True, False, False, True, False, False),
+        ("SphericalFourierNeuralOperatorNet", False, True, False, False, False, True),
+        ("HEALPixRecUNet", False, False, True, False, False, False),
+        ("Samudra", False, False, False, False, False, False),
+        ("NoiseConditionedSFNO", False, False, False, False, False, False),
+        ("NoiseConditionedSFNO", True, False, False, True, True, False),
     ],
 )
 def test_train_and_inference(
@@ -628,6 +628,7 @@ def test_train_and_inference(
     use_healpix: bool,
     use_schedule: bool,
     validate_using_ema: bool,
+    multi_validation: bool,
     very_fast_only: bool,
 ):
     """Ensure that ACE training and subsequent standalone inference run without errors.
@@ -658,6 +659,7 @@ def test_train_and_inference(
         use_schedule=use_schedule,
         validate_using_ema=validate_using_ema,
         log_validation_maps=log_validation_maps,
+        multi_validation=multi_validation,
     )
     # using pdb requires calling main functions directly
     with mock_wandb() as wandb:
@@ -671,7 +673,10 @@ def test_train_and_inference(
 
         epoch_logs = wandb_logs[-1]
         assert "inference_0/mean_step_20_norm/weighted_rmse/channel_mean" in epoch_logs
-        assert "val/mean_norm/weighted_rmse/channel_mean" in epoch_logs
+        primary_val_name = "val_0" if multi_validation else "val"
+        assert f"{primary_val_name}/mean_norm/weighted_rmse/channel_mean" in epoch_logs
+        if multi_validation:
+            assert "val_extra/mean/loss" in epoch_logs
         ensemble_step_20_keys = [
             k for k in epoch_logs if "inference_0/ensemble_step_20/" in k
         ]
@@ -691,7 +696,9 @@ def test_train_and_inference(
             "inference epoch log"
         )
 
-    validation_output_dir = tmp_path / "results" / "output" / "val" / "epoch_0001"
+    validation_output_dir = (
+        tmp_path / "results" / "output" / primary_val_name / "epoch_0001"
+    )
     assert validation_output_dir.exists()
     validation_diags = ["mean"]
     validation_map_diags = ["snapshot", "mean_map"]
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -91,6 +91,47 @@ JOB_SUBMISSION_SCRIPT_PATH = (
 )
 
 
+def _make_validation_entries(
+    *,
+    valid_data_path,
+    spatial_dimensions_str,
+    conditional,
+    log_validation_maps,
+    multi_validation=False,
+) -> InlineValidationConfig | list[InlineValidationConfig]:
+    primary = InlineValidationConfig(
+        loader=DataLoaderConfig(
+            dataset=XarrayDataConfig(
+                data_path=str(valid_data_path),
+                spatial_dimensions=spatial_dimensions_str,
+                labels=["era5"] if conditional else None,
+            ),
+            batch_size=2,
+            num_data_workers=0,
+        ),
+        aggregator=OneStepAggregatorConfig(
+            log_snapshots=log_validation_maps,
+            log_mean_maps=log_validation_maps,
+        ),
+    )
+    if not multi_validation:
+        return primary
+    secondary = InlineValidationConfig(
+        loader=DataLoaderConfig(
+            dataset=XarrayDataConfig(
+                data_path=str(valid_data_path),
+                spatial_dimensions=spatial_dimensions_str,
+                labels=["era5"] if conditional else None,
+            ),
+            batch_size=2,
+            num_data_workers=0,
+        ),
+        name="val_extra",
+        weight=0.0,
+    )
+    return [primary, secondary]
+
+
 def _get_test_yaml_files(
     *,
     train_data_path,
@@ -118,6 +159,7 @@ def _get_test_yaml_files(
     use_schedule=False,
     validate_using_ema=False,
     derived_forcings=None,
+    multi_validation=False,
 ):
     input_time_size = 1
     output_time_size = 1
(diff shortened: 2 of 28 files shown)


fme/coupled/train/train_config.py
```diff
@@ -227,13 +258,20 @@ def checkpoint_dir(self) -> str:
     def output_dir(self) -> str:
         return os.path.join(self.experiment_dir, "output")
 
-    @property
-    def inference_aggregator(self) -> InferenceEvaluatorAggregatorConfig:
-        return self.inference.aggregator
-
-    @property
-    def inference_n_coupled_steps(self) -> int:
-        return self.inference.n_coupled_steps
```
mcgibbon: These look like unrelated changes, please revert them.

fme/coupled/train/train_config.py
```diff
@@ -258,16 +329,6 @@ def _get_train_window_data_requirements(self) -> CoupledDataRequirements:
             self.config.n_coupled_steps
         )
 
-    def _get_evaluation_window_data_requirements(self) -> CoupledDataRequirements:
```
mcgibbon: How is this related to validation? Revert if unrelated.

configs/baselines/amip-c96-shield/train-ace2.yaml
```diff
@@ -51,20 +51,21 @@ train_loader:
...
-      - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0001
-        subset:
-          start_time: '1996-01-01'
-          stop_time: '2000-12-31'
-      - data_path: /climate-default/2024-07-24-vertically-resolved-c96-1deg-shield-amip-ensemble-dataset/netCDFs/ic_0002
-        subset:
-          start_time: '1996-01-01'
-          stop_time: '2000-12-31'
+validation:
+- loader:
```
mcgibbon: Actually, I'd like you to use the non-list version for these baselines that have only 1 validation entry.

fme/ace/train/train.py
```diff
@@ -109,27 +115,55 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
+                metric_key = f"{name}/mean/loss"
+                loss = logs.get(metric_key)
+                if loss is None:
+                    raise RuntimeError(
+                        f"Validation entry {name!r} with "
+                        f"weight={entry_config.weight} did not produce "
+                        f"expected metric key {metric_key!r}."
+                    )
+                weighted_loss += entry_config.weight * loss
+        return all_logs, weighted_loss
```
mcgibbon, line 137: No integration test exercises multiple validation entries end-to-end. The config-level naming/weight tests are thorough, but no test in `test_train.py`, `test_trainer.py`, or `coupled/test_train.py` actually runs training with 2+ validation configs. This means the `weighted_loss` summation logic, the `all_logs.update()` merging, and multi-entry aggregator construction in the validation callback closures are untested at runtime.

fme/ace/train/train.py
```diff
@@ -109,27 +115,55 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
-            record_logs=lambda logs: None,
-        )
-        return logs, logs["val/mean/loss"]
+    def validation_callback(
+        epoch: int,
+        stepper: TrainStepperABC,
+        ema: EMATracker,
+    ) -> tuple[dict[str, Any], float]:
+        all_logs: dict[str, Any] = {}
+        weighted_loss = 0.0
```
mcgibbon, line 102: No test runs training with an empty validation list. With no entries, `weighted_loss` returns `0.0` on every epoch, which becomes `valid_loss`. This means `best_val_loss` will be `0.0` from epoch 1, and `step_scheduler(valid_loss=0.0)` will feed `0.0` to the LR scheduler on every epoch. If this is intentional, it deserves a test documenting the behavior; if not, it may warrant a guard (or at minimum a warning).

fme/ace/train/train.py
```diff
@@ -109,27 +115,55 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
+                validation_data=data,
+                aggregator=aggregator,
+                label=name,
+                diagnostics_subdir=f"epoch_{epoch:04d}",
+                record_logs=lambda logs: None,
+                ema=ema,
+                validate_using_ema=config.validate_using_ema,
+            )
+            all_logs.update(logs)
+            if entry_config.weight > 0:
```
mcgibbon, line 127: The config-level `test_validation_zero_weight_accepted` exists, but there's no test verifying that a zero-weight entry's loss is correctly excluded from the `weighted_loss` sum at runtime (the `if entry_config.weight > 0` branch here). Consider adding a test that runs 2 validation entries where one has `weight=0.0` and asserts that its loss doesn't influence the returned `weighted_loss`.

fme/core/generics/trainer.py
```diff
@@ -395,9 +390,11 @@ def train(self):
             and self._current_epoch_num_batches_seen == 0
         ):
             logging.info("Starting validation before training")
+            valid_logs, valid_loss = validation_callback(
+                self._epochs_trained, self.stepper, self._ema
```
mcgibbon: Previously, `validation_callback(epoch)` was called inside `self.validation_context()`, which applied EMA params once for the entire validation+inference block. Now the callback runs outside that context and receives `stepper` and `ema` directly, passing them to `run_validation` → `run_validation_loop`, which applies EMA itself. This means the EMA context is entered/exited once *per validation entry* rather than once for all of them. Functionally correct, but worth noting: if EMA state application becomes expensive, this will multiply the cost by the number of validation entries.

fme/ace/train/train.py
```diff
@@ -109,27 +115,55 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
+                train_stepper=stepper,
+                validation_data=data,
+                aggregator=aggregator,
+                label=name,
+                diagnostics_subdir=f"epoch_{epoch:04d}",
+                record_logs=lambda logs: None,
+                ema=ema,
+                validate_using_ema=config.validate_using_ema,
+            )
+            all_logs.update(logs)
```
mcgibbon, line 126: Duplication note: The `validation_callback` closures in `fme/ace/train/train.py` and `fme/coupled/train/train.py` are nearly identical ~35-line blocks. Similarly, `InlineValidationConfig`, `validation_list`, `validation_names`, and `get_validation_data` are duplicated between ACE and coupled `train_config.py`. This mirrors the existing inference duplication. We're aware and will consolidate later.

fme/ace/train/train.py
mcgibbon: The trainer is generic over this, but let's cover the behavior in test_train.py, ideally without adding a new test.

fme/ace/train/train.py
mcgibbon: We should require at least one validation entry - zero was not supported before, and for now we shouldn't add it.

fme/ace/train/train.py
mcgibbon: We should test this. Perhaps we can refactor the code to make this more unit-testable at the callback level? Like test a new get_validation_callback function?

fme/core/generics/trainer.py
mcgibbon: I think this is a bug, right? The whole callback should happen in the context, and I thought we don't have access to or call that context from within the callback.

Oh, perhaps this issue is because we needed to add the ability for the lr tuning trials to use the validation callback? We really would like to maintain the safety of the context being called from outside, not inside the callback. Is that possible, while re-using the callback for lr tuning? Please comment.

fme/coupled/train/train_config.py
mcgibbon: Claude: Done — reverted these unrelated changes in commit f68a9b715.

fme/coupled/train/train_config.py
mcgibbon: Claude: Reverted — this was an unrelated cleanup that got pulled in during the rebase. Removed in commit f68a9b715.

configs/baselines/amip-c96-shield/train-ace2.yaml
mcgibbon: Claude: Done — updated in commit bea5e224b. Single-entry baseline configs now use the non-list format for validation.

fme/ace/train/train.py
mcgibbon: Claude: Addressed — added a `multi_validation=True` parameter to the existing `test_train_without_inline_inference` in `test_train.py`. When enabled, it configures two validation entries (the original one plus a second with `weight: 0.0`), and the test asserts that both `val_0/mean/loss` and `val_extra/mean/loss` appear in the wandb logs. This exercises the full multi-validation path (weighted loss summation, log merging, overlap check) without adding a new test function.

fme/ace/train/train.py
mcgibbon: Claude: Done — added a `__post_init__` check in both ACE and coupled `TrainConfig` that raises `ValueError("At least one validation entry is required.")` when the validation list is empty. Added corresponding `test_empty_validation_raises` tests in both test_train_config modules.

fme/ace/train/train.py
mcgibbon: Claude: Done — extracted `get_validation_callback()` and `get_validate_stepper_callback()` as standalone factory functions in both `fme/ace/train/train.py` and `fme/coupled/train/train.py`. They take a `Sequence[tuple[InlineValidationConfig, GriddedDataABC, str]]` and return a closure, making them directly unit-testable.

Added `TestGetValidationCallback` in `fme/ace/train/test_train_config.py` with three tests:
- `test_single_entry` — basic callback returns correct logs and loss
- `test_zero_weight_excluded_from_loss` — verifies `weight=0.0` entries contribute to logs but not to `weighted_loss`
- `test_multiple_weighted_entries` — verifies the weighted sum `2.0 * 0.1 + 3.0 * 0.2`

These mock `run_validation` to test the callback logic in isolation.

fme/ace/train/train.py
mcgibbon: Claude: Acknowledged — the duplication between ACE and coupled validation callbacks, config classes, and builder methods is real. Consolidating this into `fme/core/` would be a good follow-up, but it would require abstracting over the different aggregator types and dataset info types, which felt like a larger refactor than this PR should take on. Happy to tackle it separately if you agree.

fme/core/generics/trainer.py
mcgibbon: Claude: Yes, this was a bug. Fixed — `validation_callback` now takes only `(epoch: int)` again (matching the original signature on main) and is called inside `self.validation_context()`, which applies the EMA params once for the entire validation+inference block, just as before.

For LR tuning, we use a separate `ValidateStepper` callback (passed as a new constructor parameter to `Trainer`). This callback takes `(stepper, ema)` and calls `run_validation_loop` internally to manage EMA on the trial steppers — it never goes through `validation_context()`. This preserves the safety property: normal validation always runs under `validation_context()`, and LR tuning independently manages EMA via the existing `run_validation_loop` helper that was already designed for this.

The key insight is that `validation_context()` and `ValidateStepper` serve different purposes: the former wraps the main stepper's EMA for epoch-end validation, while the latter wraps trial steppers' EMA during LR search. They don't interfere with each other.

fme/coupled/train/train.py
elynnwu: This method can be removed now

fme/ace/train/train.py
elynnwu: I'm not seeing this in the PR, can you double check?

fme/ace/test_train.py
```diff
@@ -317,14 +322,20 @@ def _get_test_yaml_files(
             time_buffer=time_buffer,
             sample_with_replacement=10,
         ),
-        validation_loader=DataLoaderConfig(
-            dataset=XarrayDataConfig(
-                data_path=str(valid_data_path),
-                spatial_dimensions=spatial_dimensions_str,
-                labels=["era5"] if conditional else None,
+        validation=InlineValidationConfig(
```
elynnwu: I think we should add a test with multiple validation

elynnwu: And another test verifying that `weight=0` validation entry's loss is correctly excluded from `weighted_loss` at the callback level

fme/ace/train/train.py
```diff
@@ -85,11 +83,19 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
-    logging.info("Initializing validation data loader")
-    validation_data = builder.get_validation_data()
 
     variable_metadata = get_derived_variable_metadata() | train_data.variable_metadata
-    for data, name in zip((train_data, validation_data), ("train", "valid")):
+
+    if config.validation_list:
+        logging.info("Initializing validation data loaders")
+    else:
+        logging.info("Skipping validation")
```
elynnwu: We don't actually want to skip validation right? Should remove this after checking that at least one validation entry is there

fme/ace/train/train.py
```diff
@@ -109,27 +115,55 @@ def build_trainer(builder: TrainBuilders, config: TrainConfig) -> "Trainer":
...
-        aggregator = aggregator_builder.get_validation_aggregator()
-        logs = run_validation(
-            train_stepper=stepper,
-            validation_data=validation_data,
-            aggregator=aggregator,
-            diagnostics_subdir=f"epoch_{epoch:04d}",
-            record_logs=lambda logs: None,
-        )
-        return logs, logs["val/mean/loss"]
+    def validation_callback(
```
elynnwu: [nit] Add a comment noting that it manages its own EMA because LR tuning passes trial stepper/ema instances? This is different from inference callback and probably worth having a brief comment.

fme/ace/train/train.py
mcgibbon: Agreed. Let's ensure the config requires at least one validation entry (so we raise at config-validation time), and then convert this to an exception.

fme/ace/train/train_config.py
```diff
@@ -222,25 +254,30 @@ def stepper_config(self) -> StepperConfig:
...
-        if self.train_loader.using_labels != self.validation_loader.using_labels:
-            raise ValueError(
-                "train_loader and validation_loader must both use labels or both not "
-                "use labels"
-            )
-        resolved_names = self.inference_names
-        if len(resolved_names) != len(set(resolved_names)):
-            raise ValueError(f"Duplicate inference names: {resolved_names}")
-        reserved_overlap = set(resolved_names) & self._RESERVED_NAMES
+        resolved_val_names = self.validation_names
```
mcgibbon: nit: expand val to validation

fme/ace/train/train_config.py
```diff
@@ -222,25 +254,30 @@ def stepper_config(self) -> StepperConfig:
...
+        if len(resolved_val_names) != len(set(resolved_val_names)):
+            raise ValueError(f"Duplicate validation names: {resolved_val_names}")
+        for i, entry in enumerate(self.validation_list):
+            if self.train_loader.using_labels != entry.using_labels:
+                name = resolved_val_names[i]
+                raise ValueError(
+                    f"train_loader and validation {name!r} loader "
+                    "must both use labels or both not use labels"
+                )
+        resolved_inf_names = self.inference_names
```
mcgibbon: nit: expand inf to inference to disambiguate from infinity.

fme/ace/train/train_config.py
```diff
@@ -288,6 +325,25 @@ def set_random_seed(self):
...
         return self.train_evaluation_samples // self.train_loader.batch_size
 
+    @property
+    def validation_list(self) -> list[InlineValidationConfig]:
+        if isinstance(self.validation, InlineValidationConfig):
+            return [self.validation]
+        return self.validation
+
+    @property
+    def validation_names(self) -> list[str]:
```
mcgibbon, line 347: Nit: if this is only used privately, prefix it with an underscore.

fme/ace/test_train.py
mcgibbon: Agreed, I kind of think we should refactor the "get a callback that calls multiple evaluation routines and applies weights to their losses" logic into a function that can be unit tested, as well. But the main "test_train" type test should be updated to use multiple validation and multiple inference for at least one case, as well.

fme/core/generics/trainer.py
```diff
@@ -238,6 +236,7 @@ def __init__(
         end_of_batch_callback: EndOfBatchCallback = lambda: None,
         end_of_epoch_callback: EndOfEpochCallback = null_end_of_epoch_callback,
         inference_callback: InferenceCallback = _null_inference_callback,
+        validate_stepper: ValidateStepper | None = None,
```
mcgibbon, line 239: Let's add a docstring, specifically so we can specify the difference in usage between the validation callback and validate_stepper, and why we need both.

fme/ace/train/train.py
mcgibbon: Yes, in particular we should add info about this to a docstring on the generic Trainer init.

fme/coupled/train/train.py
mcgibbon: Claude: Done — removed `get_validation_aggregator` from `CoupledAggregatorBuilder` in commit f6015b8.

fme/ace/train/train.py
mcgibbon: Claude: It's at `fme/ace/train/train_config.py:257-258` and `fme/coupled/train/train_config.py:248-249` — `__post_init__` raises `ValueError('At least one validation entry is required.')` when `validation_list` is empty. Covered by `test_empty_validation_raises` in both test_train_config modules.

fme/ace/test_train.py
mcgibbon: Claude: Done — `test_train_and_inference` now takes a `multi_validation` parameter, and the `SphericalFourierNeuralOperatorNet` row sets it to `True` so the main integration test runs with 2 validation entries (one with `weight=0`) and 2 inference entries. `test_train_without_inline_inference` also uses `multi_validation=True`. Commit f6015b8.

mcgibbon: Claude: `TestGetValidationCallback.test_zero_weight_excluded_from_loss` covers this at the callback level for ACE. In commit f6015b8 I added `TestGetValidationCallback` and `TestGetInferenceCallback` to `fme/coupled/train/test_train_config.py` for parity, including the same zero-weight exclusion checks.

fme/ace/train/train.py
mcgibbon: Claude: Done — the 'Skipping validation' branch is gone from `build_trainer`, and `__post_init__` now raises when `validation_list` is empty, so the skip path is unreachable.

fme/ace/train/train.py
mcgibbon: Claude: Done in commit f6015b8 — added a comment above `get_validate_stepper_callback` in both ACE and coupled `train.py` noting that LR tuning passes trial stepper/EMA instances distinct from the Trainer's own stepper, so this callback manages its own EMA via `run_validation_loop` rather than relying on the Trainer's `validation_context()`. Also covered in the new `Trainer.__init__` docstring.

fme/ace/train/train_config.py
mcgibbon: Claude: Done — renamed `resolved_val_names` → `resolved_validation_names` in both ACE and coupled `train_config.py`. Commit f6015b8.

fme/ace/train/train_config.py
mcgibbon: Claude: Done — renamed `resolved_inf_names` → `resolved_inference_names`, `inf_entry` → `inference_entry`, and `inf_name` → `inference_name`. Commit f6015b8.

fme/ace/train/train_config.py
mcgibbon: Claude: Left this one for your guidance — `validation_names` is referenced externally by tests (multiple `assert config.validation_names == [...]` cases) and by `TrainBuilders.get_validation_data` in the same module; same for `inference_names`. Happy to underscore both (and update the test references) if you'd prefer treating them as private.

fme/ace/test_train.py
mcgibbon: Claude: Done — extracted `get_inference_callback` as a factory in both ACE and coupled `train.py`, mirroring `get_validation_callback`. Added `TestGetInferenceCallback` to both `test_train_config.py` files covering weighted summation, zero-weight exclusion, epoch filtering, and the missing-metric error path. `test_train_and_inference` now also runs with `multi_validation=True` for one parametrization, so multiple validation + multiple inference is exercised end-to-end. Commit f6015b8.

fme/core/generics/trainer.py
mcgibbon: Claude: Done — added an `__init__` docstring to `Trainer` in `fme/core/generics/trainer.py`. The `validation_callback` entry notes that the Trainer wraps the call in `validation_context()`, so the callback must not enter it itself. The `validate_stepper` entry explains the LR-tuning use case and that the callback receives trial stepper/EMA instances distinct from `self.stepper`/`self._ema` and manages EMA on those trial instances itself. Commit f6015b8.

elynnwu (approved):
lgtm

PR #1151, opened 2026-05-11
Install HPX reorder arrays

yikwill:
HEALPix reordering arrays aren't included when installing the repository. This PR fixes this by adding them to the `pyproject.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ find = {}
 
 [tool.setuptools.package-data]
 "fme.core.cuhpx" = ["data/*.npy", "data/*.fits"]
+"fme.core.hpx" = ["data/*.npy"]
 
 [tool.uv]
 cache-keys = [


mcgibbon (approved):
LGTM

pyproject.toml
```diff
@@ -80,6 +80,7 @@ find = {}
 
 [tool.setuptools.package-data]
 "fme.core.cuhpx" = ["data/*.npy", "data/*.fits"]
```
mcgibbon, line 82: Comment: This is pre-existing but do we need the .fits? I thought I kept this just for documentation of where the .npy's came from. Though if they're small it doesn't matter, anyways.

yikwill: I don't think they're strictly necessary given that the .npy files are included, but they come with cuhpx and are very small so I don't think it hurts to keep them around.

PR #1152, opened 2026-05-11
Bump version to 2026.5.0

brianhenn:
Bump version to 2026.5.0


brianhenn:
Actually will branch this from the actual tagged commit to avoid future confusion.

PR #1154, opened 2026-05-11
Add hierarchical inference metric configuration

mcgibbon:
Redesign `InferenceEvaluatorAggregatorConfig` so each metric is a named field with its own typed config, `enabled` flag, and `strict` flag. Defaults match the standard metric set — users only need to specify overrides.

Previously, `InferenceEvaluatorAggregatorConfig` accepted an optional `metrics` list and computed defaults at build time based on runtime information (grid type, run length, ensemble count). The new design makes all defaults explicit and static in the config, with runtime compatibility checks moved into each metric's `build()` method as `MetricNotSupportedError` exceptions.

Changes:
- **`InferenceEvaluatorAggregatorConfig`**: replace `metrics: list | None` with named fields for each metric type (`mean_denorm`, `mean_norm`, `step_means`, `ensembles`, `power_spectrum`, `zonal_mean`, `time_mean_denorm`, `time_mean_norm`, `video`, `histogram`, `seasonal`, `annual`, `enso_index`, `enso_coefficient`, `ipo_index`), plus `__post_init__` target validation
- **`build_inference_evaluator_aggregator` builder function**: extracted from the config's `build()` method; used by the config, the legacy config, and the coupled aggregator
- **`enabled` + `strict` fields** on all 15 `MetricConfig` dataclasses: `enabled` controls inclusion; `strict` controls whether `MetricNotSupportedError` raises (True) or warns-and-skips (False). Built-in metrics default to `strict=False`; user-opt-in metrics (`histogram`, `video`, `seasonal`) default to `strict=True`
- **`type` discriminator fields removed** from all metric configs (no longer needed since metrics are identified by field name)
- **Runtime checks moved into metric `build()` methods**: run-length checks for `annual`, `enso_index`, `enso_coefficient`, `ipo_index`; grid type check for `enso_coefficient`; step count checks for `ensemble` and `step_mean` (now raises `MetricNotSupportedError` instead of `ValueError`)
- **Coupled aggregator** (`fme/coupled/aggregator.py`): uses `build_inference_evaluator_aggregator` directly instead of constructing intermediate config objects
- **Evaluator** (`fme/ace/inference/evaluator.py`): expand `initial_time` for ensemble members so metric builders see the correct sample dimension
- **Baseline configs** (12 files under `configs/baselines/`): migrate from legacy boolean flags to hierarchical format
- **Public API** (`fme/ace/__init__.py`): export `InferenceEvaluatorAggregatorConfig` and all 15 metric config classes
- **Docs** (`docs/evaluator_config.rst`): document the default aggregator config YAML and all metric config types
- **Tests**: hierarchical build/enable/disable, `MetricNotSupportedError` for unsupported runtime conditions, default YAML regression test, doc-sync test for metric config coverage

- [x] Tests added


fme/ace/aggregator/inference/main.py
```diff
@@ -177,25 +186,44 @@ def build(
             initial_time=initial_time,
         )
 
+        defaults = self._default_metrics(ctx, n_ensemble_per_ic)
```
mcgibbon: This override based on lists and name matching seems confusing and prone to allowing strange/erroneous configs. I kind of want to swap to a version where every sub-aggregator type is a sub-config of the aggregator config, with the enabled: setting determining whether they get built. Would that be reasonable?

elynnwu:
To summarize, here're 3 equivalent ways of configuring the same aggregator:

### Legacy
```yaml
aggregator:
  log_zonal_mean_images: true
  log_histograms: true
``` 
### List style (`InferenceEvaluatorAggregatorConfig`, currently in main)
```yaml
aggregator:
  metrics:
    - type: mean
      target: denorm
    - type: mean
      target: norm
    - type: step_mean
      step: 20
      target: denorm
    - type: step_mean
      step: 20
      target: norm
    - type: ensemble
      step: 20
    - type: power_spectrum
    - type: zonal_mean
    - type: time_mean
      target: denorm
    - type: time_mean
      target: norm
    - type: histogram
    - type: annual
    - type: enso_index
    - type: enso_coefficient
```
### Hierarchical (`HierarchicalInferenceEvaluatorAggregatorConfig`)
```yaml
aggregator:
  histogram:
    enabled: true
```

The list style is less flexible in that you either use the default (`aggregator: {}`) or if you want anything more custom (such as turning on histogram), you will need to list everything as the above example. However, it has the benefit of being able to see exactly what is logged easily from the config. The hierarchical style is more flexible because it gives you the default plus the ability to selectively toggle individual metrics.

I found hierarchical to be more user friendly and lean towards keeping hierarchical and dropping list style. Though I'm curious to hear what others think.

mcgibbon:
Yes, I would also lean towards hierarchical style, and turn the list-style config into a builder function (so that we still use it to construct the aggregator, since I do like the way it does it).

mcgibbon:
Here's an example of a more "full" hierarchical yaml config. Note that usually you won't need to fill so many sections due to reasonable defaults. Also variable lists can be re-used in multiple sections using anchors and aliases.

```yaml
aggregator:
  # --- Global-mean time-series (enabled by default) ---
  # target is fixed by the field name (denorm/norm) and cannot be overridden.
  mean_denorm:
    enabled: true
  mean_norm:
    enabled: true

  # --- Per-step snapshot metrics ---
  # Multiple entries supported; defaults to step-20 denorm + norm.
  step_means:
    - step: 10
      target: denorm
    - step: 10
      target: norm
    - step: 20
      target: denorm
    - step: 20
      target: norm
      # Optionally restrict to specific variables:
      variables:
        - air_temperature_0
        - specific_humidity_0

  # --- Ensemble spread (skipped automatically when n_ensemble <= 1) ---
  ensembles:
    - step: 20
      log_mean_maps: true

  # --- Spatial metrics ---
  power_spectrum:
    enabled: true
    variables:
      - air_temperature_0
      - eastward_wind_0
  zonal_mean:
    enabled: true
    zonal_mean_max_size: 8192

  # --- Time-mean metrics (target is fixed by the field name) ---
  time_mean_denorm:
    reference_data: /data/time-mean-ref.nc
  time_mean_norm:
    enabled: true

  # --- Optional metrics (disabled by default, enable here) ---
  video:
    enabled: true
    enable_extended_videos: true
    variables:
      - air_temperature_0
      - precipitation_rate
  histogram:
    enabled: true
  seasonal:
    enabled: true

  # --- Climate indices ---
  annual:
    enabled: true
    reference_data: /data/annual-ref.nc
  enso_index:
    enabled: true
  enso_coefficient:
    enabled: true

  # --- Reference data ---
  monthly_reference_data: /data/monthly-ref.nc
  time_mean_reference_data: /data/time-mean-ref.nc

  ```

**Comment by jpdunc23** [comment]
> Here's an example of a more "full" hierarchical yaml config. Note that usually you won't need to fill so many sections due to reasonable defaults. Also variable lists can be re-used in multiple sections using anchors and aliases.
> 
> ```yaml
> aggregator:
>   # --- Global-mean time-series (enabled by default) ---
>   # target is fixed by the field name (denorm/norm) and cannot be overridden.
>   mean_denorm:
>     enabled: true
>   mean_norm:
>     enabled: true
> 
>   # --- Per-step snapshot metrics ---
>   # Multiple entries supported; defaults to step-20 denorm + norm.
>   step_means:
>     - step: 10
>       target: denorm
>     - step: 10
>       target: norm
>     - step: 20
>       target: denorm
>     - step: 20
>       target: norm
>       # Optionally restrict to specific variables:
>       variables:
>         - air_temperature_0
>         - specific_humidity_0
> 
>   # --- Ensemble spread (skipped automatically when n_ensemble <= 1) ---
>   ensembles:
>     - step: 20
>       log_mean_maps: true
> 
>   # --- Spatial metrics ---
>   power_spectrum:
>     enabled: true
>     variables:
>       - air_temperature_0
>       - eastward_wind_0
>   zonal_mean:
>     enabled: true
>     zonal_mean_max_size: 8192
> 
>   # --- Time-mean metrics (target is fixed by the field name) ---
>   time_mean_denorm:
>     reference_data: /data/time-mean-ref.nc
>   time_mean_norm:
>     enabled: true
> 
>   # --- Optional metrics (disabled by default, enable here) ---
>   video:
>     enabled: true
>     enable_extended_videos: true
>     variables:
>       - air_temperature_0
>       - precipitation_rate
>   histogram:
>     enabled: true
>   seasonal:
>     enabled: true
> 
>   # --- Climate indices ---
>   annual:
>     enabled: true
>     reference_data: /data/annual-ref.nc
>   enso_index:
>     enabled: true
>   enso_coefficient:
>     enabled: true
> 
>   # --- Reference data ---
>   monthly_reference_data: /data/monthly-ref.nc
>   time_mean_reference_data: /data/time-mean-ref.nc
> ```

This might be helpful to have in `config/examples` as a quick reference to the "full" hierarchical config, i.e. just this `aggregator` entry in a yaml file without the other config entries. A simple test would help ensure it doesn't get stale.

**Comment by jpdunc23** [comment]
Looks great! Is there something the list style config can do that the hierarchical style can't? If not, I agree with Elynn that we should remove the list style as an option (perhaps in a later PR).

**Comment by mcgibbon** [comment]
> This might be helpful to have in `config/examples` as a quick reference to the "full" hierarchical config, i.e. just this `aggregator` entry in a yaml file without the other config entries. A simple test would help ensure it doesn't get stale.

Perhaps I can have the sphinx docs automatically generate it? If we want to have a `.yaml` in the repo for reference I can make it a built-and-committed component of the sphinx docs that requires running a script to re-generate. I think this would be really helpful to see the "default" config.

File: fme/ace/aggregator/inference/test_evaluator.py [hunk 2]
```diff
@@ -685,3 +705,141 @@ def test_legacy_config_long_run_includes_annual_and_enso():
...
+            id="time_mean_norm_wrong_target",
+        ),
+    ],
+)
+def test_hierarchical_rejects_mismatched_target(kwargs, match):
+    with pytest.raises(ValueError, match=match):
+        HierarchicalInferenceEvaluatorAggregatorConfig(**kwargs)
+
+
+def test_default_aggregator_config_yaml():
```
mcgibbon: Please put this test in docs/ instead of in fme/.

fme/ace/aggregator/inference/test_evaluator.py
```diff
@@ -685,3 +705,141 @@ def test_legacy_config_long_run_includes_annual_and_enso():
...
+        default_flow_style=False,
+        sort_keys=False,
+    )
+    docs_path = pathlib.Path(__file__).parents[4] / "docs"
+    validate_text(content, docs_path / "default-aggregator-config.yaml")
+
+
+def test_all_metric_configs_documented():
+    """Every type in the MetricConfig union must appear in evaluator_config.rst."""
+    import fme.ace
```
mcgibbon: Put all imports at the top of the file, including in the test you move to docs/.

mcgibbon:
Pre-review agent: **Design note for reviewer — intentional silent-skip behavior**

`HierarchicalInferenceEvaluatorAggregatorConfig.build()` hardcodes `raise_on_unsupported=False` (line 317 of `main.py`). This means **all** metrics — including ones the user explicitly enabled — are silently skipped when their runtime requirements aren't met (e.g. ENSO on a HEALPix grid, step_mean on a short run).

This differs from the old list-style config where explicitly-provided metrics would raise an error if unsupported. The change is intentional: the hierarchical config's philosophy is that enabled metrics are "desired if possible" rather than "required." A warning is logged for each skipped metric.

Tests have been added to verify both the raise (default) and silent-skip (`raise_on_unsupported=False`) code paths.

docs/evaluator_config.rst
```diff
@@ -89,10 +89,68 @@ The sub-configurations are:
...
+   :language: yaml
+   :caption: Default Aggregator Configuration
+
+Metric configurations
+^^^^^^^^^^^^^^^^^^^^^
+
+The named fields on
+:class:`~fme.ace.HierarchicalInferenceEvaluatorAggregatorConfig`
+accept the following typed entries.  Each entry has an ``enabled`` field
+(default ``true``) that can be set to ``false`` to disable a metric.
```
jpdunc23: Some default to `false`, e.g. the video and histogram aggregators. This will be clear from `default-aggregator-config.yaml`, but maybe there is a way to call it out here as well?

fme/ace/aggregator/inference/main.py
```diff
@@ -64,80 +64,228 @@
...
-    entry is a typed metric configuration (e.g. ``MeanMetricConfig``,
-    ``StepMeanMetricConfig``).  When ``metrics`` is ``None``, a default set
-    of metrics is computed at build time based on the grid type and run length.
+    Each metric is a named field with its own typed configuration and an
+    ``enabled`` flag.  Defaults match the standard metric set: metrics that
+    are always desired are enabled, while optional ones (``histogram``,
+    ``video``, ``seasonal``) are disabled.
+
+    Metrics whose runtime requirements are not met (e.g. ``enso_index``
+    on a non-lat/lon grid) are silently skipped even when enabled.
```
jpdunc23: Maybe we should add a warning so it is clear in the logs why this is skipped? Could be helpful in cases where the rollout length is borderline, e.g. the IPO will be skipped for any rollout <80 years and if by accident inference is configured with number of steps just barely below the threshold, it would be hard to notice until the job finished.

This could be done in a separate PR. Added issue https://github.com/ai2cm/ace/issues/1173 to track.

fme/ace/aggregator/inference/main.py
```diff
@@ -64,80 +64,228 @@
...
+        save_diagnostics=save_diagnostics,
+        output_dir=output_dir,
+        n_ensemble_per_ic=n_ensemble_per_ic,
+        ensemble_aggregators=ensemble_aggregators,
+    )
+
+
 @dataclasses.dataclass
-class InferenceEvaluatorAggregatorConfig:
+class HierarchicalInferenceEvaluatorAggregatorConfig:
```
jpdunc23: Issue: I think `HierarchicalInferenceEvaluatorAggregatorConfig` may be missing the checks for sufficient rollout length for the ENSO and IPO indices. The checks are in `InferenceAggregatorConfig`.

fme/ace/aggregator/inference/main.py
```diff
@@ -64,80 +64,228 @@
...
+    for metric in metrics:
+        name = metric.get_name()
+        try:
+            result: MetricBuildResult = metric.build(ctx)
+        except MetricNotSupportedError:
+            if raise_on_unsupported:
+                raise
+            logging.warning(
+                f"{name} metric not supported for this configuration, " "omitting."
+            )
```
jpdunc23, line 143: It would be helpful to pass along the caught error message with this warning.

fme/ace/aggregator/inference/main.py
jpdunc23: I guess they are not really "silently" skipped, since there is a warning in the logs.

Agent review raised a good point about change in behavior here:

> The old system made a meaningful distinction: default metrics (computed at runtime) were silently skipped, but user-explicit metrics (`InferenceEvaluatorAggregatorConfig(metrics=[...])`) raised loudly.
> 
> Impact: Configuration mistakes are masked. Users may believe they're computing metrics that were silently dropped.
> 
> Suggestion: Consider either (a) adding a `strict: bool = False` field to the hierarchical config that flips raise_on_unsupported, or (b) distinguishing between "default-enabled" metrics (can be silently skipped) and "user-enabled" metrics (where enabled was explicitly set to true and should fail loudly).

I like option (b), keeping the warning in the logs.

fme/ace/aggregator/inference/main.py
jpdunc23: Nit: Why not just keep the name `InferenceEvaluatorAggregatorConfig`?

jpdunc23 (commented):
Agent review spotted a bug in handling of the ENSO and IPO metrics. 

It would also be nice to restore the behavior of raising errors loudly for non-default metrics, while skipping default metric errors with an informative warning message.

Will this config style also replace `InferenceAggregatorConfig` in a future PR?

docs/evaluator_config.rst
mcgibbon: Yeah, that should be called out here and referenced (that config literally gets inserted into this file, in the section directly above).

fme/ace/aggregator/inference/main.py
mcgibbon: Sounds like a good idea to me. There's also an awkward `" "` that should be removed to make this one string, if it goes on one line.

fme/ace/aggregator/inference/main.py
mcgibbon: It's an artifact of the development history. Let's rename this back to InferenceEvaluatorAggregatorConfig.

mcgibbon: Re the rollout length issue, Claude should check for that and report the findings.

Generally speaking, these "appear when the rollout length is long enough" metrics have taken a lot of extra testing and handling, and been a source of a lot of AI iteration while working on these PRs. We might think about if there are simpler ways we'd accept configuring them, or API changes that would make this safety more straightforward (like formalizing the concept and re-using it across this type of metric). Not necessarily something I want to do here in this PR, but maybe worth making another PR for.

Like, from our requirements side, could these all be default-off and need to be added to the config to turn them on? Then we get errors if the run isn't long enough? That would make the handling much simpler and reduce the likelihood of incorrect configs you point out in another comment.

fme/ace/aggregator/inference/main.py
mcgibbon: I agree, but let's do both a and b. Make default-enabled metrics strict: bool = False, and make default-disabled metrics strict: bool = True.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Renamed back to `InferenceEvaluatorAggregatorConfig` in b2cb82e.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Done — combined the two f-strings into one and included the caught error message: `f"{name} metric not supported for this configuration, omitting: {e}"`.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Implemented both (a) and (b). Each metric config now has a `strict` field:

- **Default-enabled** metrics (`mean`, `step_means`, `power_spectrum`, `zonal_mean`, `time_mean`, `annual`, `enso_index`, `enso_coefficient`, `ipo_index`, `ensembles`): `strict = False` — skipped with a warning when unsupported.
- **Default-disabled** metrics (`video`, `histogram`, `seasonal`): `strict = True` — raise an error when the user explicitly enables them but requirements aren't met.

The `build_inference_evaluator_aggregator` function checks `raise_on_unsupported or metric.strict` to decide whether to raise or warn.

fme/ace/aggregator/inference/main.py
mcgibbon: Claude: Investigated the rollout length handling. The new config is **not missing** the checks — they've been moved into each metric's `build()` method as `MetricNotSupportedError`:

| Metric | Legacy threshold | New threshold | Grid check |
|---|---|---|---|
| `AnnualMetricConfig` | `> 730 days` | `<= 730 days` raises | Same (none) |
| `EnsoIndexMetricConfig` | `> 730 days` + LatLon | `<= 730 days` raises + LatLon + LatLonOps | Stricter (adds LatLonOps) |
| `EnsoCoefficientMetricConfig` | `> 1800 days` (no grid check) | `<= 1800 days` raises + LatLon | Stricter (adds LatLon check legacy lacked) |
| `IpoIndexMetricConfig` | `> 29200 days` + LatLon | `<= 29200 days` raises + LatLon | Same |

All thresholds are equivalent (`> X` for inclusion ↔ `<= X` for rejection). The new code is slightly stricter for `EnsoCoefficientMetricConfig` (now checks LatLon, which the legacy code omitted — an improvement since the metric uses lat/lon operations internally).

Test coverage in `TestMetricNotSupportedError` covers all four metrics for both short-run and wrong-grid-type cases, plus integration tests `test_hierarchical_config_skips_unsupported_on_short_run` and `test_hierarchical_config_skips_unsupported_on_healpix`.

fme/ace/aggregator/inference/test_evaluator.py
mcgibbon: Claude: Done — moved to `docs/test_default_aggregator_config.py` and `docs/test_metric_configs_documented.py`. All imports are at the top of both files.

docs/evaluator_config.rst
mcgibbon: Claude: Updated the docs text to explicitly note which metrics default to enabled/disabled, referencing the default config YAML section above. See the updated wording at lines 112–116 of evaluator_config.rst.

fme/ace/aggregator/inference/main.py
jpdunc23: Thanks! I may have been reviewing a version of this branch prior to [a9f1f42](https://github.com/ai2cm/ace/pull/1154/commits/a9f1f425b4083a91a37fd0871bc74847c9744a6c). In future I will be sure to have agents check for remote changes before finalizing review.

jpdunc23 (approved):
LGTM! Be sure to update the description as some parts are out-dated at time of review (e.g., reference to `HierarchicalInferenceEvaluatorAggregatorConfig`).

PR #1156, opened 2026-05-11
Modernizing HEALPix UNet

yikwill:
This PR is a large refactor to the HEALPix model which brings it up to speed with the DLESyM codebase and also removes a lot of dead code relating to recurrence.

Changes:
- Removes _all_ code related to recurrence (e.g., unused recurrent blocks) as this was never implemented. We will leave adding recurrence to the model for a separate project.
- Removes HEALPixRecUNet. The naming is confusing given the lack of recurrence, and its forward function tries to handle things that should be abstracted to the stepper like prognostic/diagnostic split and residual prediction. Instead, we add HEALPixUNet whose forward method is a simple pass of the encoder and decoder.
- Moves padding layers to separate `healpix_paddings.py` file. Adds `HEALPixPaddingIsolatitude` padding layer to resolve HEALPix grid imprinting issues.
- Deprecates `enable_healpixpad` option in favor of `hpx_padding_mode`.
- Adds layers `DealiasedDownsample` and `SmoothedInterpolateConv` to address checkerboarding artifacts.
- Moves all downsampling, upsampling, and conv blocks to `healpix_blocks.py`. Previously, the up/downsampling layers lived in `healpix_activations.py` for no clear reason.

- [x] Tests added


fme/ace/models/healpix/healpix_decoder.py
```diff
@@ -80,15 +79,15 @@ class UNetDecoder(nn.Module):
...
+        up_sampling_block: UpsamplingBlockConfig,
         output_layer: ConvBlockConfig,
-        recurrent_block: Optional[RecurrentBlockConfig] = None,
         n_channels: Sequence = (64, 32, 16),
         n_layers: Sequence = (1, 2, 2),
         output_channels: int = 1,
         dilations: Optional[list] = None,
         enable_nhwc: bool = False,
-        enable_healpixpad: bool = False,
+        hpx_padding_mode: str = "earth2grid",
```
mcgibbon:
```suggestion
        hpx_padding_mode: Literal["earth2grid", "karlbauer", "isolatitude"] = "earth2grid",
```

fme/ace/models/healpix/healpix_decoder.py
```diff
@@ -97,68 +96,78 @@ def __init__(
...
+                in_channels=curr_channel * 2 if n > 0 else curr_channel,
+                latent_channels=curr_channel,
+                out_channels=next_channel,
+                dilation=dilations[n],
+                n_layers=n_layers[n],
+                enable_nhwc=enable_nhwc,
+                hpx_padding_mode=hpx_padding_mode,
+                nside=face_nside,
+            )
+            conv_module = conv_cfg.build()
```
mcgibbon, line 178: Don't do it in this PR, too much going on, but generally speaking it would be nice to move "build" type arguments out of the dataclass, and instead as arguments to `.build`. Specifically, the number of input and output channels comes to mind. Anything that you wouldn't want the user to configure in a yaml file, but should instead be determined at build time from other variables in the runtime.

The fact that you have to `.replace` attributes of something we've specified in a yaml is a bit scary in this respect, those things should probably not be configurable rather than being silently ignored.

fme/ace/models/healpix/healpix_encoder.py
```diff
@@ -48,7 +45,8 @@ class UNetEncoderConfig:
     n_layers: List[int] = dataclasses.field(default_factory=lambda: [2, 2, 1])
     dilations: Optional[list] = None
     enable_nhwc: bool = False
-    enable_healpixpad: bool = False
+    hpx_padding_mode: str = "earth2grid"
```
mcgibbon: Suggestion: Similar `Literal` type hint instead of `str`.

fme/ace/models/healpix/healpix_encoder.py
```diff
@@ -48,7 +45,8 @@ class UNetEncoderConfig:
...
+    nside: Optional[int] = None
```
mcgibbon: Add new arguments to docstring, especially explaining the default for this since it's None.

fme/ace/models/healpix/healpix_encoder.py
```diff
@@ -92,48 +92,60 @@ def __init__(
             n_layers:, # of layers to use for the convolutional blocks
             dilations: list of dilations to use for the the convolutional blocks
             enable_nhwc: if channel last format should be used
-            enable_healpixpad: if healpixpad library should be used (true if installed)
+            hpx_padding_mode: HEALPix padding backend. Default ``"earth2grid"``;
```
mcgibbon, line 102: Missing nside from docstring.

fme/ace/models/healpix/healpix_encoder.py
```diff
@@ -92,48 +92,60 @@ def __init__(
...
+            )
+            modules.append(conv_cfg.build())
             old_channels = curr_channel
 
             self.encoder.append(nn.Sequential(*modules))
 
         self.encoder = nn.ModuleList(self.encoder)
 
-    def forward(self, inputs: Sequence) -> Sequence:
+    def forward(self, inputs):
```
mcgibbon: Can we keep a type hint?

fme/ace/registry/hpx.py
```diff
@@ -1,46 +1,38 @@
...
-    input_time_size: int = 0
-    output_time_size: int = 0
-    delta_time: str = "6h"
-    reset_cycle: str = "24h"
-    n_constants: int = 2
-    decoder_input_channels: int = 1
-    prognostic_variables: int = 7
     enable_nhwc: bool = False
-    enable_healpixpad: bool = False
+    hpx_padding_mode: str = "earth2grid"
```
mcgibbon: Suggestion: Use Literal type hint.

fme/ace/test_train.py
```diff
@@ -992,7 +973,7 @@ def test_train_without_inline_inference(tmp_path, very_fast_only: bool):
         timestep_days=20,
         n_time=int(366 * 3 / 20 + 1),
         inference_forward_steps=int(366 * 3 / 20 / 2 - 1) * 2,  # must be even
-        use_healpix=(nettype == "HEALPixRecUNet"),
+        use_healpix=False,
```
mcgibbon, line 1078: Is this change intentional? I think this would mean it never uses healpix data to train but I'm not sure.

fme/ace/test_train.py
```diff
@@ -419,12 +400,12 @@ def _get_test_yaml_files(
     )
 
     with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".yaml") as f_train:
-        f_train.write(yaml.dump(dataclasses.asdict(train_config)))
+        f_train.write(yaml.safe_dump(dataclasses.asdict(train_config)))
```
mcgibbon: Can we indeed safe_dump these? I thought there was a reason we didn't. In any case, please separate out things like this into their own PRs to make sure whoever looks at it understands what it's doing, I could easily have missed this and it has implications for the rest of the codebase.

If safe_dump indeed does work here, we might want to use safe_dump throughout more of our code. However, we might still decide to use `dump` to avoid authors of later changes needing to keep this safe (in particular, if we can't dump checkpoints safely, is there any point to dumping/loading anything safely?).

fme/ace/test_train.py
```diff
@@ -121,23 +120,11 @@ def _get_test_yaml_files(
     output_time_size = 1
     if derived_forcings is None:
         derived_forcings = DerivedForcingsConfig()
-    if nettype == "HEALPixRecUNet":
+    if nettype == "HEALPixUNet":
```
mcgibbon, line 171: Comment: We're breaking the ability for any existing healpix checkpoints to be loaded, which I think is fine.

mcgibbon (commented):
Just giving this a quick look with some comments/pointers.

A PR of this size isn't really something I can review for e.g. correctness, and I'd also want to do a quick agent review so it can look more thoroughly than I can at so many changes.

That would mean this code would be going in as kind of "unverified", until you show scientific results that demonstrate it's correct. This is perhaps the right path for this particular code, since the existing version wasn't "working" scientifically, but it's something to keep in mind. Once your model is indeed "working" and we have results we don't want to degrade, you'd need to separate changes out into individual PRs so a reviewer like me could understand that change and how it's tested.

mcgibbon:
Tests appear to be failing due to earth2grid missing, which I didn't expect to see outside of the "no healpix" tests. They also require CUDA which makes the CPU tests fail, but even GPU is failing. Also, make sure to install pre-commit hooks (`pre-commit install`) so they run before you commit.

fme/ace/test_train.py
yikwill: This test hardcodes `nettype = "SphericalFourierNeuralOperatorNet" so `use_healpix` would always be `False` here. Not sure if there already exists tests for SFNO on HEALPix, but we can add them later if not.

fme/ace/test_train.py
yikwill: I changed this to be consistent with the use of `safe_load` in other parts of this file. However, I don't fully understand the difference between `dump` and `safe_dump`, so I agree that this should be its own PR if `safe_dump` works. Reverting these to `dump` for now.

yikwill:
Thanks for the comments. I've addressed them and the PR is ready for re-review. I think the tests are failing because the GitHub workflow is outdated for HEALPix. Looking at the workflow file I see `uv pip install --no-build-isolation -c constraints.txt -e .[dev,healpix,graphcast]` but `healpix` is not an optional dependency in `pyproject.toml`. Matching the current `Makefile`, I think the workflow should do

`uv pip install -c constraints.txt -e .[dev,docs,graphcast]`
`uv pip install --no-build-isolation -c constraints.txt -r requirements-healpix.txt`

Not sure if this is something I can change since I'm not a maintainer.

mcgibbon:
> Thanks for the comments. I've addressed them and the PR is ready for re-review. I think the tests are failing because the GitHub workflow is outdated for HEALPix. Looking at the workflow file I see `uv pip install --no-build-isolation -c constraints.txt -e .[dev,healpix,graphcast]` but `healpix` is not an optional dependency in `pyproject.toml`. Matching the current `Makefile`, I think the workflow should do
> 
> `uv pip install -c constraints.txt -e .[dev,docs,graphcast]` `uv pip install --no-build-isolation -c constraints.txt -r requirements-healpix.txt`

Hm yes I think the issue was that extras don't let you specify a commit from a certain git repo as a requirement. I agree about your suggestion for how to update the github CI yaml for the workflow to work. It seems like perhaps we weren't actually using earth2grid, before your PR, or at least there was a fallback.

> Not sure if this is something I can change since I'm not a maintainer.

You're actually the only person who can change it, as owner of the PR (unless you've given permission to push to your repo, but it's still cleaner if you do it). Just go into the `.github` folder in your branch and make the necessary changes to the CI workflows. Then when we run your PR, it will run the updated workflows.

Pre-commit also still has [many failures](https://github.com/ai2cm/ace/actions/runs/26464505796/job/77923963025), make sure you've installed the hooks and do a `pre-commit run --all-files` locally.

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -15,112 +15,234 @@
...
     """
 
-    in_channels: int = 3
-    kernel_size: int = 1
+    block_type: Literal["MaxPool", "AvgPool", "DealiasedDownsample"]
+    pooling: int = 2
     enable_nhwc: bool = False
-    enable_healpixpad: bool = False
-    block_type: Literal["ConvGRUBlock", "ConvLSTMBlock"] = "ConvGRUBlock"
+    hpx_padding_mode: Literal["earth2grid", "karlbauer", "isolatitude"] = "earth2grid"
```
mcgibbon, line 70: Suggestion (optional): Some of these settings feel like they might be "global" settings that you're needing to duplicate on each configuration class. For example, you probably want to use the same hpx_padding_mode on all layers always. You could configure this as a higher level, and have it be passed as an argument to `.build` instead, as a way of sharing the configuration across levels/layers.

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -15,112 +15,234 @@
...
+        nside: Native face height/width for HEALPix padding.
+        in_channels: Input channels for ``DealiasedDownsample`` (set by encoder before build).
+        resample_filter: 1D filter weights for dealiased blur stages.
     """
 
-    in_channels: int = 3
-    kernel_size: int = 1
+    block_type: Literal["MaxPool", "AvgPool", "DealiasedDownsample"]
+    pooling: int = 2
     enable_nhwc: bool = False
```
mcgibbon, line 69: Suggestion (optional): Some of these settings like enable_nhwc and stride feel like they aren't meant to be changed. We hard-code the memory formats of our inputs, and you don't have any options to change/permute them, so enable_nhwc could probably be removed. `stride` feels like something determined at code-time rather than configuration time, or at best you probably want pooling == stride.

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -15,112 +15,234 @@
...
+    dilation: int = 1
+    upsample_mode: str = "nearest"
+    activation: Optional[CappedGELUConfig] = None
+    enable_nhwc: bool = False
+    hpx_padding_mode: Literal["earth2grid", "karlbauer", "isolatitude"] = "earth2grid"
+    nside: Optional[int] = None
+    nside_after: Optional[int] = None
+    align_corners: bool = False
+    scale_factor: Optional[int] = None
+    mode: Optional[str] = None
```
mcgibbon, line 162: Issue (optional): There are several configuration options that only apply to specific configurations, and are ignored otherwise. You could make it easier to reason about these groups of settings if you split this configuration into multiple classes, each with its own specific e.g. `block_type: Literal["Interpolate"]` input (that is required, with no default and only one valid value) that dacite uses to select the correct configuration during load time.

This same pattern applies to some of the other configuration classes as well.

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -489,9 +837,15 @@ def __init__(
             upscale_factor: Factor by which to upscale the number of latent channels.
             activation: Configuration for the activation function used between layers.
             enable_nhwc: Whether to enable NHWC format.
-            enable_healpixpad: Whether to enable HEALPixPadding.
+            hpx_padding_mode: HEALPix padding backend passed to wrapper.
+            nside: Native face height/width for HEALPix padding.
         """
         super().__init__()
+        hpk = _healpix_layer_kwargs(
```
mcgibbon: nit: use `kwargs` or `healpix_kwargs`, we try to avoid short names like this where possible except for short-lived iterators or indexing operators (e.g. i/j/k).

fme/ace/models/healpix/healpix_decoder.py
mcgibbon: This is still an important issue. I am specifically in the mode of reviewing this PR with few/no required changes so that these very large diffs can get merged in quickly (though it has been two weeks this PR has been open), with these issues getting fixed after.

The new (optional) comments are nice-to-haves that would make the code easier to configure, but this comment is a "I smell danger, the code may do something unexpected and undetectable" type of issue.

fme/ace/models/healpix/healpix_layers.py
```diff
@@ -16,486 +16,132 @@
...
-|4//\\5 //\\6 //\\7 //\\4|            \\4//\\5 //\\6 //\\7 //\\             |4  |5  |6  |7  |
-|//  \\//  \\//  \\//  \\|             \\/  \\//  \\//  \\//  \\            -----------------
-|| 8  |  9  |  10 |  11  |              \\8 //\\9 //\\10//\\11//            |   |   |   |   |
---------------------------               \\//  \\//  \\//  \\//             |8  |9  |10 |11 |
-                                                                            -----------------
-                                    "\\" are top and bottom, whereas
-                                    "//" are left and right borders
-
-
-Details on the HEALPix can be found at https://iopscience.iop.org/article/10.1086/427976
```
mcgibbon, line 37: Are we changing the layout we're using? If not, we should keep this in a comment somewhere (maybe here). If so, what are we moving to, and how is the layout controlled?

fme/ace/models/healpix/healpix_layers.py
```diff
@@ -16,486 +16,132 @@
...
-        """
-        NF, C, H, W = tensor.shape
-        N = int(NF / self.num_faces)
-        tensor = th.reshape(tensor, shape=(N, self.num_faces, C, H, W))
-
-        return tensor
-
-
-class HEALPixPadding(nn.Module):
+class HEALPixLayer(th.nn.Module):
```
mcgibbon, line 48: Suggestion (optional): Revert replacements of `nn.` with `th.nn` throughout this file.

fme/ace/models/healpix/healpix_unet.py
```diff
@@ -0,0 +1,137 @@
+"""
+HEALPix UNet: single forward-pass encoder–decoder stack.
+
+Adapted from the modulus-uw ``physicsnemo.models.dlwp_healpix.HEALPixUNet``.
+"""
+
+from collections.abc import Sequence
+from typing import Literal
+
+import torch as th
```
mcgibbon, line 10: It's a pre-existing issue in the healpix code so I'm not going to ask you to fix it in this PR, but I always read th as torch-harmonics, it would be nice to use this as "torch" like we do in the rest of the repo.

mcgibbon (approved):
You can fix the nit: if you like, but the Suggestion (optional)'s should be left for another PR so we can get these changes merged.

Assuming tests are passing.

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -159,239 +283,457 @@ def build(self) -> nn.Module:
...
+        """
+        super().__init__()
+
+        self.in_channels = in_channels
+        self.scale_factor = scale_factor
+        self.mode = mode
+        self.trim_size = trim_size
+        self.interp = th.nn.functional.interpolate
+
+        self.smoother_kernel = th.tensor(
```
mcgibbon: Issue: This should be assigned with a `self.register_buffer("smoother_kernel", ...)` call to avoid needing to manually send it to GPU every time this gets called (instead it would follow the `.to(device)` call on the module).

fme/ace/models/healpix/healpix_layers.py
```diff
@@ -16,486 +16,132 @@
...
-            raise NotImplementedError(
-                "HEALPixPaddingv2 is not available in this environment"
+        kernel_size = 3 if "kernel_size" not in kwargs else kwargs["kernel_size"]
+        dilation = 1 if "dilation" not in kwargs else kwargs["dilation"]
+        padding = ((kernel_size - 1) // 2) * dilation
+
+        # Define a HEALPixPadding layer if padding is necessary
+        if padding > 0:
+            # Disable native padding for conv layers
+            if layer.__bases__[0] is th.nn.modules.conv._ConvNd:
```
mcgibbon: Suggestion: Consider using `issubclass(layer, nn.modules.conv._ConvNd)` instead, which is more robust.

Full AI comment:
  1. HEALPixLayer assumes layer.__bases__[0] exists (healpix_layers.py, around the padding
  detection logic). If layer doesn't have __bases__ (e.g., it's not a class, though there's some
  protection), or if the inheritance chain is different (e.g., DealiasBlurConv2d inherits from
  nn.Module, not _ConvNd), this silently skips padding. The old code had a guard if not
  isinstance(layer, type) or not issubclass(layer, th.nn.Module). With the new code, the base-class
   check layer.__bases__[0] is th.nn.modules.conv._ConvNd is fragile — DealiasBlurConv2d extends
  nn.Module, not _ConvNd, so it won't get kwargs["padding"] = 0 applied. This is probably fine
  because DealiasBlurConv2d doesn't pass padding to nn.Conv2d internally (it uses F.conv2d
  directly), but the logic is still brittle and would silently break if someone passes a Conv
  subclass with a different inheritance chain. Consider using issubclass(layer,
  nn.modules.conv._ConvNd) instead of checking __bases__[0].

fme/ace/models/healpix/healpix_blocks.py
```diff
@@ -159,239 +283,457 @@ def build(self) -> nn.Module:
...
+            mode: Interpolation mode passed to ``F.interpolate``.
+            trim_size: Border pixels to crop after smoothing (removes edge artifacts).
+        """
+        super().__init__()
+
+        self.in_channels = in_channels
+        self.scale_factor = scale_factor
+        self.mode = mode
+        self.trim_size = trim_size
+        self.interp = th.nn.functional.interpolate
```
mcgibbon, line 611: Suggestion (optional): Unusual to assign a function to an attribute so it can be called in one place (I only detected the call in forward) and never overridden - you could instead call `th.nn.functional.interpolate` directly where you're currently calling `self.interp`.

mcgibbon:
Some additional comments from AI review above which you should consider fixing, but I think are better to do after you get tests passing and merge in the current code.

fme/ace/models/healpix/healpix_layers.py
yikwill: Not changing the layout, just moved this comment to `healpix_paddings.py`. I feel like this comment is most helpful there, but I'm happy to keep it here if you think otherwise.

fme/ace/models/healpix/healpix_blocks.py
yikwill: Fixed.

fme/ace/models/healpix/healpix_layers.py
yikwill: Fixed.

fme/ace/models/healpix/healpix_blocks.py
yikwill: Fixed.

fme/ace/models/healpix/healpix_decoder.py
yikwill: Got it. I'll address this issue in another PR along with all your (optional) comments I left open.

.github/workflows/tests.yaml
```diff
@@ -28,8 +28,8 @@ jobs:
           python -m pip install uv
           # can't use --system here because it doesn't work with --no-build-isolation and a healpix dependency
           uv venv --python 3.11
-          uv pip install -c constraints.txt -e .[dev]
-          uv pip install --no-build-isolation -c constraints.txt -e .[dev,healpix,graphcast]
+          uv pip install -c constraints.txt -e .[dev,graphcast]
```
mcgibbon, line 31: Does graphcast not require no-build-isolation?

yikwill: I'm not actually sure, was just imitating what is done in the Makefile:

```
conda run --no-capture-output -n $(ENVIRONMENT_NAME) uv pip install -c constraints.txt -e .[dev,docs,graphcast]
conda run --no-capture-output -n $(ENVIRONMENT_NAME) uv pip install --no-build-isolation -c constraints.txt -r requirements-healpix.txt
```

yikwill: Looking at it closer, I don't believe graphcast requires no-build-isolation because the deps are plain packages. From `pyproject.toml`:
```
graphcast = [
    "rtree>=1.4.1",
    "trimesh>=4.8.3",
]
```

PR #1158, opened 2026-05-11
Remove coupled `LossContributionsConfig`

jpdunc23:
Remove `LossContributionsConfig` and move its config attributes (`n_steps`, `optimize_last_step_only`, and `weight`) to `ComponentTrainingConfig` (as `n_steps`, `optimize_last_step_only`, and `loss_weight`, respectively).

Changes:
- Added `n_steps_max` and `loss_is_null` properties to `ComponentTrainingConfig`.

- [x] Tests added


fme/coupled/stepper.py
```diff
@@ -1366,15 +1367,29 @@ def _process_ensemble_output_list(
 
 
 class CoupledStepperTrainLoss:
```
jpdunc23, line 1370: Could move this to `loss.py` to reduce the length of this file, but not in this PR to avoid hiding the API changes in the diff.

fme/coupled/test_train.py
```diff
@@ -267,85 +258,6 @@ def _write_test_yaml_files(
...
-    # A weight=0 component still has a non-None n_steps_max if a value was
-    # explicitly set, and the validator surfaces the misconfiguration even
-    # though the loss contribution is null. This avoids silently accepting
-    # confused configs.
-    null_config = LossContributionsConfig(weight=0.0, n_steps=999)
-    bounds = CoupledOptionalInt(ocean=null_config.n_steps_max, atmosphere=None)
-    with pytest.raises(ValueError, match=r"ocean"):
-        _validate_loss_n_steps(
-            n_coupled_steps=1, n_inner_steps=1, component_n_steps_max=bounds
-        )
```
jpdunc23, line 346: These are duplicates of the tests in `test_train_config.py` that must've accidentally been left here.

fme/coupled/stepper.py
```diff
@@ -1417,36 +1432,83 @@ def step_is_optimized(
...
 
 
 @dataclasses.dataclass
 class ComponentTrainingConfig:
+    """Configuration for one component's training behavior within a coupled
+    stepper.
+
+    Parameters:
+        loss: The step loss function configuration (e.g. MSE, EnsembleLoss).
+        n_steps: Number of consecutive steps contributing to the loss, starting
```
mcgibbon: Suggestion: Call this the number of component steps in the optimization window, since the number of steps contributing to the loss is 1 when optimize_last_step_only=True.

Question: This is the configuration for just one component. Because of that, what is the "first" step? That component's first step after the IC? If you call this "the number of component steps in the optimization window", it would imply e.g. 22 means the second step after the first ocean step, in the case of the atmosphere.

mcgibbon (approved):
LGTM, just one minor request to reword a docstring. No issues found by AI either.

PR #1159, opened 2026-05-12
Refactor loss architecture for correct per-channel reporting

mcgibbon:
Per-channel loss reporting was broken: all channels reported identical values because loss functions reduced to scalars internally before `LossOutput` could distinguish channels (bug identified in #1111). This PR fixes the bug by restructuring loss reduction so that each loss returns pre-weighted unreduced tensors and a single cached reduction path in `LossOutput` produces both `total()` and `get_channel_losses()` consistently.

Changes:
- `fme.core.loss.LossComponent`, `StandardLoss`, `EnsembleComponentLoss`: new ABC hierarchy encoding tensor layout so each loss component knows how to reduce to `(B, C)`
- `fme.core.loss.LossOutput`: now holds `list[LossComponent]` with cached per-channel reduction; `total()` derives from per-channel values so `sum(get_channel_losses()) == total()` by construction; new `scale()` method
- `fme.core.loss.LpLoss`: returns per-channel relative norms `[(B, C)]` instead of scalar
- `fme.core.loss.AreaWeightedMSELoss`: returns `[(B, C)]` instead of scalar
- `fme.core.loss.CRPSLoss`: returns unreduced `[(B, C, H, W)]` instead of scalar
- `fme.core.loss.EnergyScoreLoss`: returns pre-weighted spectral `[(B, C, L, M)]` where `.mean(trailing)` reproduces the old scalar
- `fme.core.loss.FiniteDifferenceCRPSLoss`: returns spatial-reduced `[(B, C)]`; new `_reduce_spatial` helper
- `fme.core.loss.EnsembleLoss`, `WeightedSum`, `GlobalMeanLoss`: collect/concatenate `list[LossComponent]` from sub-losses
- `fme.core.loss.WeightedMappingLoss`: wraps raw-tensor losses (MSE, L1) into correct `LossComponent` subclass based on `channel_dim`
- `fme.core.loss.StepLoss`: uses `LossOutput.scale()` instead of accessing private attributes
- `fme.core.loss.LossConfig.build()`: `reduction` parameter removed; MSE/L1 always use `reduction="none"` internally
- `fme.core.loss._MSELoss`, `_L1Loss`: thin wrappers returning `list[LossComponent]` for use inside `WeightedSum`
- `fme.downscaling.models`, `fme.downscaling._deterministic_models`: drop `reduction` kwarg from `LossConfig.build()` calls

- [x] Tests added

Supersedes #1111


fme/core/loss.py
```diff
@@ -337,34 +387,36 @@ class EnergyScoreLoss(torch.nn.Module):
...
             )
             self.mode_weights[..., 0] = 1
-        return (
-            (get_energy_score(x_hat, y_hat) * self.mode_weights)
-            .sum(dim=(-2, -1))
-            .mean()
-        ) / self.scaling
+        assert self.n_spectral is not None
+        es = get_energy_score(x_hat, y_hat) * self.mode_weights
+        pre_weighted = es * (self.n_spectral / self.scaling)
```
mcgibbon, line 419: Issue: The weighting-reweighting here is a bit hard to follow. Can you make it more obvious, and/or explain what's going on in # inline comments? In particular it's confusing that we multiply by the sqrt of these numbers and then multiply by them again. Is it that before we summ'ed the losses and now we need to take the average? I thought both the spectral and area-based losses were already mean'ed over space rather than summ'ed.

mcgibbon: Pre-review agent: The math here converts from the old sum-based reduction to the new mean-based reduction used by `StandardLoss.reduce_to_channel()`:

- Old path: `.sum(dim=(-2,-1)).mean() / scaling` where `scaling = 2*sqrt(L*M)`
- New path: `StandardLoss` does `.mean(dim=(-2,-1))` = `sum / (L*M)`
- Factor: `L*M / scaling = L*M / (2*sqrt(L*M)) = sqrt(L*M) / 2` makes them equivalent

Added an inline comment explaining this.

fme/coupled/test_loss.py
```diff
@@ -17,7 +17,12 @@
 
 def _wrap_as_loss_output(value: torch.Tensor) -> LossOutput:
     """Wrap a scalar tensor as a LossOutput for mocking StepLoss."""
-    return LossOutput(loss=value, channel_dim=0, channel_names=["mock"])
+    from fme.core.loss import StandardLoss
```
mcgibbon: Issue: Imports should happen at the top of the file.

Arcomano1234: Nit: Not sure why but Claude still hasn't fixed this

Arcomano1234 (approved):
Left 1 comment about claude not fixing what you also spotted. The other thought I had was I feel like there is some test coverage that might not be there mainly related to the new `LossOutput.scale`, I am curious if you think there is already at test that covers this what was not obvious.

fme/core/loss.py
```diff
@@ -315,7 +361,7 @@ def __init__(self, weights: torch.Tensor, loss: torch.nn.Module):
         self.loss = loss
         self.weights = weights
 
-    def __call__(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+    def forward(self, x: torch.Tensor, y: torch.Tensor) -> Any:
```
Arcomano1234: Question: is this supposed to be Any can't it be `-> list[LossComponent]`

mcgibbon: Yeah we should use robust types where we can.

fme/coupled/test_loss.py
mcgibbon: Claude: Fixed — moved the `StandardLoss` import to the top of the file with the other `fme.core.loss` imports.

fme/core/loss.py
mcgibbon: Claude: Changed the return type to `list[LossComponent]`. Mypy passes with the update.

PR #1161, opened 2026-05-12
E3SM updated variable names

elynnwu:
Names for the latest E3SMv3 variables.


mahf708:
Shouldn't we aim for a rename/aliasing feature instead 😭 I know Jeremy will likely hate this suggestion. 

I think FME should have its own and single set of internal names (be it ERA5 or even anything else made up); with a rename/aliasing feature, the YAML config can have arbitrary mapping between disk names and internal names

I had claude cook something up here (and it works in my testing) https://github.com/ai2cm/ace/commit/4a4fddf0d0f0d4c869bbc8ba3a344b92b543a9dd 

This and a few random/optional features I am readying for near-future contrib

jpdunc23 (approved):
LGTM!

PR #1163, opened 2026-05-13
Add `IpoIndexMetricConfig` to coupled inference aggregator

jpdunc23:
Enables IPO aggregator for coupled inference.

- [ ] Tests added

### DIFF
--- a/fme/coupled/aggregator.py
+++ b/fme/coupled/aggregator.py
@@ -8,12 +8,14 @@ import torch
 import xarray as xr
 
 from fme.ace.aggregator.inference.main import (
+    APPROXIMATELY_EIGHTY_YEARS,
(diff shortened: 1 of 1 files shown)


elynnwu (approved):
[nit] not sure if we would want an additional flag to `include_nino34` for IPO, I think we always want to log IPO if nino34 is logged.

jpdunc23:
> [nit] not sure if we would want an additional flag to `include_nino34` for IPO, I think we always want to log IPO if nino34 is logged.

This is for convenience to disable the nino34 in the atmosphere component aggregator, since otherwise it would be redundantly computed for `surface_temperature`. I don't recall if the IPO is also computed for `surface_temperature`, if not we could add that to match the enso index behavior.

PR #1167, opened 2026-05-13
Remove stale tmp ckpt files

### LINKED ISSUE #1165: .tmp model weights are being propagated in beaker datasets for resumed jobs
During training if the job is terminated, the code tries to save the current state of the model. As part of the `on_terminate` and `save_checkpoint` in generic trainer, it first saves the model to a tmp file then once fully saved moves the file to `ckpt.tar`. However, if this processes fails to complete before the job is fully shutdown you are left with these tmp ckpts. It seems on beaker these tmp files will not get deleted so if a job resumes, it inherits these incomplete tmp ckpts and can quickly increase the size of the beaker datasets if the job is pre-empted a lot. See an example [here](https://beaker.org/orgs/ai2/workspaces/ace/work/01KR4MFZ07VZH4H35REGN20K8X?taskId=01KR4MFZ6D62WXKZHB9X1N27VC&jobId=01KREK7JBDHA20FCMDQ5GQYJMY). 

It would be nice if these files could be deleted as not waste potentially ~10s of GB increase in beaker datasets sizes.

elynnwu:
When a training job is hard-killed during checkpoint saving, the `.{uuid}.tmp` partial checkpoint files are left behind. On Beaker, these get captured in result datasets and propagated to resumed jobs via `shutil.copytree`, accumulating multi-GB of wasted space across preemptions. This PR removes these stale tmp files.

Changes:
- Added a `remove_stale_tmp_checkpoints` helper and called during trainer startup, after `shutil.copytree` in `ResumeResultsConfig.prepare_directory`, and downscaling training

- [x] Tests added

Resolves #1165

### DIFF
--- a/fme/core/test_cli.py
+++ b/fme/core/test_cli.py
@@ -1,6 +1,6 @@
 import os
 
-from .cli import prepare_config, prepare_directory
+from .cli import prepare_config, prepare_directory, remove_stale_tmp_checkpoints
 
 
 def test_prepare_config(tmp_path):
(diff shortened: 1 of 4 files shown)


fme/core/test_cli.py
```diff
@@ -24,3 +24,35 @@ def test_prepare_directory(tmp_path):
...
+    tmp2.write_bytes(b"data2")
+    valid_ckpt = ckpt_dir / "latest.ckpt"
+    valid_ckpt.write_bytes(b"valid")
+
+    remove_stale_tmp_checkpoints(str(ckpt_dir))
+
+    assert not tmp1.exists()
+    assert not tmp2.exists()
+    assert valid_ckpt.exists()
+
```
Arcomano1234, line 44: Very optional: If you do something like this we can get rid of the extra test / duplication from
`test_remove_stale_tmp_checkpoints_no_tmp_files`
```suggestion

     remove_stale_tmp_checkpoints(str(ckpt_dir))
     
     assert valid_ckpt.exists()
```

fme/downscaling/train.py
```diff
@@ -535,6 +537,7 @@ def _resume_from_results_dir_if_not_preempted(experiment_dir, resume_results_dir
                 f"Existing results directory {resume_results_dir} does not exist."
             )
         shutil.copytree(resume_results_dir, experiment_dir, dirs_exist_ok=True)
+        remove_stale_tmp_checkpoints(os.path.join(experiment_dir, "checkpoints"))
```
Arcomano1234, line 540: Question: Does downscaling not potentially utilize both "training_checkpoints" and "checkpoints"?

Arcomano1234 (approved):
Left a question and very optional comment, but LGTM. Thanks!

fme/core/test_cli.py
elynnwu: Updated and removed `test_remove_stale_tmp_checkpoints_no_tmp_files`

fme/downscaling/train.py
elynnwu: Based on `checkpoint_dir` property in `TrainerConfig`, no it uses `checkpoints` only

PR #1170, opened 2026-05-14
Deterministic evaluation step sampling

mcgibbon:
Training evaluation and validation passes use the same stochastic forward-step sampling as training, making `train/mean/loss` and `val/mean/loss` fluctuate epoch-to-epoch due to random step selection rather than model changes. This PR seeds the step sampler with a fixed value before each evaluation pass, making these metrics stable and comparable across epochs.

Changes:
- `fme.ace.stepper.time_length_probabilities.TimeLengthProbabilities.seed_rng` — new method to reset RNG to a deterministic state
- `fme.core.generics.train_stepper.TrainStepperABC.seed_step_sampler` — new default no-op method on the stepper ABC
- `fme.ace.stepper.single_module.TrainStepper.seed_step_sampler` and `fme.coupled.stepper.CoupledTrainStepper.seed_step_sampler` — implementations that seed the underlying `TimeLengthProbabilities` RNG
- `fme.coupled.loss.LossContributions.seed_rng` and `fme.coupled.stepper.CoupledStepperTrainLoss.seed_step_sampler` — plumbing for coupled loss samplers
- `fme.core.generics.trainer.Trainer.train_one_epoch` and `fme.core.generics.validation.run_validation_loop` — call `seed_step_sampler(seed=0)` before evaluation batches

- [x] Tests added


fme/core/generics/train_stepper.py
```diff
@@ -67,3 +67,7 @@ def set_train(self) -> None:
     @abc.abstractmethod
     def update_training_history(self, training_job: TrainingJob) -> None:
         pass
+
+    def seed_step_sampler(self, seed: int) -> None:
```
mcgibbon: Question: This is fairly low-level information about what is happening inside the generic TrainStepper, implying for example that there is a forward-step sampler (which not all TrainSteppers might do). Can you rename this to `seed_eval`, and just have that particularly be seeding the step sampler in the case of our concrete implementations?

mcgibbon: Claude: Done — renamed to `seed_eval` on the generic ABC and all call sites. Domain-specific internals (`CoupledStepperTrainLoss.seed_step_sampler`, `LossContributions.seed_rng`) keep their original names since they're specific to step sampling. See b414490.

fme/ace/stepper/test_time_length_probabilities.py
jpdunc23, line 1: Agent review identified the below issue. We also include a failing test example for `test_trainer.py`, though I'm not sure it's exactly the right test since it creates a new `TrainStepperABC` that implements the bug in a similar way to `TrainStepper` in `fme.ace`. This bug is specific to the fme.ace `TrainStepper` and `TimeLengthProbabilities` integration with multi-epoch training, so maybe a test belongs in fme.ace, e.g. a new test module like `fme/ace/train/test_stepper_trainer_integration.py`?

For fixing, I kind of like the idea of using a second RNG in `TimeLengthProbabilities` that is dedicated to validation, but perhaps the save-and-restore approach is more general.

### `seed_eval` permanently corrupts the training RNG

`seed_rng(0)` **replaces** `TimeLengthProbabilities._rng` in-place, and nothing restores it before training resumes. This makes every training epoch (after the first eval pass) use the same deterministic sequence of forward-step counts.

<details>
**Root cause** — `TimeLengthProbabilities` objects are shared by reference. For constant schedules (the common case), `TimeLengthSchedule.get_value(epoch)` returns the **same Python object** for all epochs. `seed_rng` mutates that object's `_rng`, and `initialize_rng()` is a no-op afterward (`if self._rng is None` fails). So the distributed training seed is permanently overwritten.

**Both steppers are affected:**

- **ACE stepper** (`single_module.py`): `_update_epoch` reassigns `self._n_forward_steps_sampler`, but to the *same* `TimeLengthProbabilities` instance whose `_rng` was already overwritten by `seed_rng(0)`.
- **Coupled stepper** (`coupled/stepper.py`): Loss objects are built once in `__init__` and never recreated.

**Concrete consequence** — The eval pass consumes a fixed number of samples from `RandomState(0)`, leaving the RNG in state `S`. The next training epoch starts from `S`, advances to `S'`. Then eval resets to `RandomState(0)` again, consumes the same fixed number of samples, arrives back at `S`. The epoch after starts from `S` again. **Every training epoch from the second onward gets an identical sequence of `n_forward_steps` values.** This turns what should be stochastic training diversity into a fixed pattern.

**Suggested fix**: Save and restore the RNG state around eval rather than mutating in-place:

```python
# Option A: save/restore in seed_eval + a new restore method
def seed_eval(self, seed: int) -> None:
    if self._n_forward_steps_sampler is not None:
        rng = self._n_forward_steps_sampler._rng
        self._saved_rng_state = rng.get_state() if rng is not None else None
        self._n_forward_steps_sampler.seed_rng(seed)

def restore_train_rng(self) -> None:
    if self._saved_rng_state is not None:
        self._n_forward_steps_sampler._rng.set_state(self._saved_rng_state)
        self._saved_rng_state = None
```

Or use a separate temporary RNG for eval that doesn't touch the training state at all.

**Failing test** (`fme/core/generics/test_trainer.py`) — exercises the real `Trainer` train/eval/validation cycle:

```python
def test_seed_eval_does_not_make_training_step_sampling_periodic(tmp_path: str):
    """seed_eval(seed=0) is called before each eval/validation pass to make
    evaluation deterministic.  It must not permanently replace the training RNG
    so that each training epoch draws from a distinct random sequence.

    This test uses a TrainStepper subclass that delegates to a real
    TimeLengthProbabilities sampler.  After three epochs through the real
    Trainer, the step-count sequences recorded during epochs 2 and 3 must
    differ.  If seed_rng permanently overwrites the sampler's RNG (the
    current bug), they will be identical.
    """
    from fme.ace.stepper.time_length_probabilities import (
        TimeLengthProbabilities,
        TimeLengthProbability,
    )

    class StochasticStepper(TrainStepper):
        def __init__(self):
            super().__init__()
            self._sampler = TimeLengthProbabilities(
                [TimeLengthProbability(5, 0.5), TimeLengthProbability(10, 0.5)]
            )
            self._sampler.seed_rng(12345)
            self.epoch_samples: list[list[int]] = []
            self._current: list[int] = []
            self._training = False

        def seed_eval(self, seed: int) -> None:
            self._sampler.seed_rng(seed)

        def set_train(self) -> None:
            super().set_train()
            if self._current:
                self.epoch_samples.append(self._current)
            self._current = []
            self._training = True

        def set_eval(self) -> None:
            super().set_eval()
            self._training = False

        def train_on_batch(self, batch, optimization, compute_derived_variables=False):
            sample = int(self._sampler.sample())
            if self._training:
                self._current.append(sample)
            loss = (self._modules[0].weight * 0).sum()
            optimization.accumulate_loss(loss)
            optimization.step_weights()
            if isinstance(optimization, NullOptimization):
                self.validation_batches_seen.append(batch.i)
            else:
                self.train_batches_seen.append(batch.i)
            return TrainOutput()

    max_epochs = 3
    n_train_batches = 20
    stepper = StochasticStepper()
    train_data = TrainData(n_batches=n_train_batches)
    validation_data = TrainData(n_batches=5)

    aggregator_builder = AggregatorBuilder(
        train_losses=np.zeros(max_epochs),
        validation_losses=np.zeros(max_epochs),
        inference_losses=np.zeros(max_epochs),
    )

    config = Config(
        experiment_dir=tmp_path,
        checkpoint_dir=os.path.join(tmp_path, "checkpoints"),
        max_epochs=max_epochs,
        save_checkpoint=False,
    )

    def build_optimization(modules):
        return OptimizationConfig(
            optimizer_type="Adam",
            lr=0.01,
            scheduler=SchedulerConfig(),
            enable_automatic_mixed_precision=False,
        ).build(modules, max_epochs=max_epochs)

    def validation_callback(epoch):
        validation_data.set_epoch(epoch)
        val_agg = aggregator_builder.get_validation_aggregator()
        logs = run_validation(
            train_stepper=stepper,
            validation_data=validation_data,
            aggregator=val_agg,
            record_logs=lambda logs: None,
        )
        return logs, logs["val/mean/loss"]

    trainer = Trainer(
        train_data=train_data,
        validation_data=validation_data,
        stepper=stepper,
        build_optimization=build_optimization,
        build_ema=lambda modules: EMAConfig().build(modules),
        config=config,
        aggregator_builder=aggregator_builder,
        validation_callback=validation_callback,
        do_gc_collect=False,
    )

    trainer.train()

    if stepper._current:
        stepper.epoch_samples.append(stepper._current)

    assert len(stepper.epoch_samples) == max_epochs
    assert stepper.epoch_samples[1] != stepper.epoch_samples[2], (
        "Epochs 2 and 3 produced identical step-count sequences — "
        "seed_eval permanently corrupts the training RNG, making it periodic."
    )
```
</details>

fme/coupled/stepper.py
```diff
@@ -1387,6 +1387,10 @@ def sample_n_steps(self) -> None:
         for loss_obj in self._loss_objs.values():
             loss_obj.sample_n_steps()
 
+    def seed_step_sampler(self, seed: int) -> None:
+        for loss_obj in self._loss_objs.values():
+            loss_obj.seed_rng(seed)
```
jpdunc23:
```suggestion
        for i, loss_obj in enumerate(self._loss_objs.values()):
            loss_obj.seed_rng(seed + i)
```

In some cases, it might be advantageous to set different seeds for the ocean and atmosphere, e.g. if using a 1-day timestep for both components, configured with the same `TimeLengthProbabilities`. Otherwise, ocean and atmosphere loss contributions will be locked in step during training.

fme/core/generics/train_stepper.py
```diff
@@ -67,3 +67,7 @@ def set_train(self) -> None:
     @abc.abstractmethod
     def update_training_history(self, training_job: TrainingJob) -> None:
         pass
+
+    def seed_eval(self, seed: int) -> None:
```
jpdunc23, line 72: Should this be decorated with `@abc.abstractmethod`?

jpdunc23 (commented):
Agent review found an rng issue, described in detail [in this comment](https://github.com/ai2cm/ace/pull/1170/changes#r3243691219).

Otherwise looking good.

fme/ace/stepper/test_time_length_probabilities.py
mcgibbon: I think I would prefer storing two identical samplers, one for training and one for eval, and seeding the eval sampler. The chosen sampler should be based on whether set_train or set_eval was called on the stepper.

We could add that test, though I'm a little worried about having to maintain all that mock infrastructure. In any case some kind of coverage should get added.

fme/core/generics/train_stepper.py
mcgibbon: I think yes, it should.

fme/coupled/stepper.py
mcgibbon: Yes, I think we should incorporate this update.

The same bug applies here, right? The coupled stepper must be getting polluted during training by this seeding during eval. It probably should use a similar solution. Should this also be covered by (separate) tests?

fme/coupled/test_loss.py
```diff
@@ -430,6 +435,57 @@ def test_coupled_stepper_train_loss_sample_n_steps_delegates():
...
+    loss.set_train()
+    loss._n_steps_sampler.seed_rng(42)
+    train_samples_after = []
+    for _ in range(20):
+        loss.sample_n_steps()
+        train_samples_after.append(loss._n_steps)
+    assert train_samples_before == train_samples_after
+
+
+def test_coupled_stepper_train_loss_set_train_eval_delegates():
```
mcgibbon, line 482: Do you want this test to stay @jpdunc23 ?

fme/ace/stepper/test_time_length_probabilities.py
```diff
@@ -85,3 +85,38 @@ def test_time_length_schedule_sorts_milestones():
...
+        [TimeLengthProbability(10, 0.5), TimeLengthProbability(20, 0.5)]
+    )
+    sampler.seed_rng(0)
+    first_run = [sampler.sample() for _ in range(20)]
+    sampler.seed_rng(1)
+    second_run = [sampler.sample() for _ in range(20)]
+    assert first_run != second_run
+
+
+def test_separate_instances_have_independent_rngs():
```
mcgibbon, line 112: Is this sufficient a test @jpdunc23 , given the choice of using two separate samplers for train and eval? Or do we need to also test that e.g. the right sampler is used?

I do think these tests don't fully cover the behavior, but also I'm not sure that the benefit from full coverage is worth the maintenance burden of the test, given how involved they will need to be.

fme/ace/stepper/test_time_length_probabilities.py
jpdunc23: I like the separate samplers approach.

fme/ace/stepper/test_time_length_probabilities.py
jpdunc23: Does `test_separate_instances_have_independent_rngs` actually verify that the samplers have separate RNG? Would it be better to verify that when the train and eval samplers start with the same seed, I can sample 5 times from the train sampler first, then 10 times from the eval sampler, then sample 5 more times from the train sampler and see that the 10 train sampler draws match the 10 eval sampler draws.

jpdunc23: Suggested update:

```python
  def test_separate_instances_have_independent_rngs():
      outcomes = [TimeLengthProbability(10, 0.5), TimeLengthProbability(20, 0.5)]
      train_sampler = TimeLengthProbabilities(outcomes)
      eval_sampler = TimeLengthProbabilities(list(outcomes))
      train_sampler.seed_rng(42)
      eval_sampler.seed_rng(42)
      first_5 = [train_sampler.sample() for _ in range(5)]
      [eval_sampler.sample() for _ in range(10)]  # interleaved eval draws
      next_5 = [train_sampler.sample() for _ in range(5)]
      eval_sampler.seed_rng(42)
      expected_10 = [eval_sampler.sample() for _ in range(10)]
      assert first_5 + next_5 == expected_10
```

fme/coupled/test_loss.py
jpdunc23: This seems fine to keep. Will update as needed in #1158

fme/coupled/test_loss.py
```diff
@@ -430,6 +435,57 @@ def test_coupled_stepper_train_loss_sample_n_steps_delegates():
     atmos_loss.sample_n_steps.assert_called_once()
 
 
+def test_seed_rng_does_not_corrupt_training_sampler():
```
jpdunc23, line 441: I think this test has a similar issue to the equivalent fme.ace test. Here is a suggested update:

```python
def test_seed_rng_does_not_corrupt_training_sampler():
      """seed_rng must only affect the eval sampler, leaving the training
      sampler's RNG untouched across eval/train cycles."""
      sampler = TimeLengthProbabilities(
          outcomes=[
              TimeLengthProbability(steps=1, probability=0.5),
              TimeLengthProbability(steps=4, probability=0.5),
          ]
      )
      loss = LossContributions(
          n_steps=sampler,
          weight=1.0,
          optimize_last_step_only=False,
          loss_obj=Mock(spec=StepLoss),
          time_dim=1,
          n_steps_limit=5,
      )
      assert loss._n_steps_sampler is not None
      loss._n_steps_sampler.seed_rng(42)
      loss.set_train()
      train_first_20 = []
      for _ in range(20):
          loss.sample_n_steps()
          train_first_20.append(loss._n_steps)
      # eval interlude — must not disturb the training sampler
      loss.set_eval()
      loss.seed_rng(0)
      for _ in range(10):
          loss.sample_n_steps()
      # resume training and draw 20 more
      loss.set_train()
      train_next_20 = []
      for _ in range(20):
          loss.sample_n_steps()
          train_next_20.append(loss._n_steps)
      # reference: 40 uninterrupted draws from the same seed
      loss._n_steps_sampler.seed_rng(42)
      reference_40 = []
      for _ in range(40):
          loss.sample_n_steps()
          reference_40.append(loss._n_steps)
      assert train_first_20 + train_next_20 == reference_40
```

jpdunc23 (approved):
Left a couple of blocking comments re. the new tests in `fme.ace` and `fme.coupled` to verify rng seeding independence, since I don't think the existing tests are quite verifying the independence.

Otherwise, LGTM! Don't need to look again unless you want me to.

PR #1171, opened 2026-05-14
Run all forward steps during evaluation with per-step metrics

### LINKED ISSUE #1172: Per-channel losses accumulate for wrong steps with optimize_last_step_only
## Summary

When `optimize_last_step_only=True`, per-channel losses were accumulated for all steps within the optimization window (`step < n_loss_steps`) instead of only the step that actually contributes to the optimization loss (the last step, `step == n_loss_steps - 1`).

This caused the per-channel loss breakdown to be inconsistent with the total loss reported by `optimization.get_accumulated_loss()` — the per-channel sum could exceed the total accumulated loss.

## Fix

Fixed in PR #1171. Per-channel losses now accumulate only for steps where `optimize_step` is True, aligning them with the steps that contribute to the optimization loss.

A regression test (`test_per_channel_losses_bounded_by_accumulated_loss`) verifies the per-channel total does not exceed the accumulated loss.

mcgibbon:
Stacks on #1170. Closes #1172.

During evaluation, the stepper currently only runs forward steps up to the stochastically-sampled count, so per-step loss metrics (e.g. `loss_step_40`) are only produced for the selected range and may not appear at all for longer lead times. This PR makes evaluation run all available forward steps, producing robust per-step loss metrics at every lead time, while still computing the mean loss from only the stochastically-sampled steps to stay congruent with `batch_loss`.

Changes:
- `fme.core.generics.train_stepper.TrainStepperABC.train_on_batch` — add `evaluate_all_steps` parameter (default False)
- `fme.ace.stepper.single_module.TrainStepper._accumulate_loss` — when `evaluate_all_steps=True`, loop over all data steps but only count sampled steps toward accumulated loss; fix pre-existing bug where per-channel losses accumulated for wrong steps with `optimize_last_step_only`
- `fme.coupled.loss.StepLossABC.compute_loss` — now abstract; `NullLossContributions` and `LossContributions` provide concrete implementations
- `fme.coupled.stepper.CoupledStepperTrainLoss.compute_loss` and `CoupledTrainStepper._accumulate_step_loss`/`_accumulate_loss` — coupled evaluation uses all outer steps and logs metrics for every step
- `fme.ace.aggregator.loss_metrics` — new `PerStepLossAggregator` and `PerChannelLossAggregator` sub-aggregator classes
- `fme.ace.aggregator.train.TrainAggregator` — delegates per-step and per-channel loss accumulation to sub-aggregators
- `fme.ace.aggregator.one_step.main.OneStepAggregator` — delegates per-step losses to sub-aggregator
- `fme.coupled.aggregator.TrainAggregator` — delegates per-step losses to sub-aggregator
- `fme.core.generics.trainer` and `fme.core.generics.validation` — pass `evaluate_all_steps=True` during evaluation

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/loss_metrics.py
+++ b/fme/ace/aggregator/loss_metrics.py
@@ -1,0 +1,58 @@
+import torch
+
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+
+
+class PerStepLossAggregator:
+    """Accumulates per-step loss metrics across batches and produces means."""
+
+    def __init__(self):
+        self._sums: dict[str, torch.Tensor] = {}
+        self._counts: dict[str, int] = {}
+
+    def record(self, metrics: dict[str, torch.Tensor]) -> None:
+        for key, value in metrics.items():
+            if key not in self._sums:
+                self._sums[key] = value.detach().clone()
+                self._counts[key] = 1
+            else:
+                self._sums[key] = self._sums[key] + value.detach()
+                self._counts[key] += 1
+
+    def get_logs(self, label: str) -> dict[str, float]:
+        dist = Distributed.get_instance()
+        logs: dict[str, float] = {}
+        for key in sorted(self._sums.keys()):
+            count = self._counts[key]
+            logs[f"{label}/mean/{key}"] = float(
+                dist.reduce_mean(self._sums[key] / count).cpu().numpy()
+            )
+        return logs
+
+
+class PerChannelLossAggregator:
+    """Accumulates per-channel (per-variable) loss values across batches."""
+
+    def __init__(self):
+        self._sums: dict[str, torch.Tensor] = {}
+        self._n_batches = 0
+
+    def record(self, per_channel_losses: dict[str, torch.Tensor]) -> None:
+        self._n_batches += 1
+        for var_name, value in per_channel_losses.items():
+            acc = self._sums.get(
+                var_name,
+                torch.tensor(0.0, device=get_device(), dtype=value.dtype),
+            )
+            self._sums[var_name] = acc + value
+
+    def get_logs(self, label: str) -> dict[str, float]:
+        dist = Distributed.get_instance()
+        logs: dict[str, float] = {}
+        if self._n_batches > 0:
+            for var_name, acc in self._sums.items():
+                logs[f"{label}/mean/loss/{var_name}"] = float(
+                    dist.reduce_mean(acc / self._n_batches).cpu().numpy()
+                )
+        return logs
--- a/fme/ace/aggregator/train.py
+++ b/fme/ace/aggregator/train.py
@@ -76,6 +76,8 @@ class TrainAggregator(AggregatorABC[TrainOutput]):
     def __init__(self, config: TrainAggregatorConfig, operations: GriddedOperations):
         self._n_loss_batches = 0
         self._loss = torch.tensor(0.0, device=get_device())
+        self._per_step_loss_sums: dict[str, torch.Tensor] = {}
+        self._per_step_loss_counts: dict[str, int] = {}
         self._per_channel_loss: dict[str, torch.Tensor] = {}
         self._per_channel_loss_enabled = config.per_channel_loss
         self._paired_aggregators: dict[str, Aggregator] = {}
@@ -105,6 +107,16 @@ class TrainAggregator(AggregatorABC[TrainOutput]):
     def record_batch(self, batch: TrainOutput):
         self._loss += batch.metrics["loss"]
         self._n_loss_batches += 1
+        for key, value in batch.metrics.items():
+            if key.startswith("loss_step_"):
+                if key not in self._per_step_loss_sums:
+                    self._per_step_loss_sums[key] = value.detach().clone()
+                    self._per_step_loss_counts[key] = 1
+                else:
+                    self._per_step_loss_sums[key] = (
+                        self._per_step_loss_sums[key] + value.detach()
+                    )
+                    self._per_step_loss_counts[key] += 1
         if self._per_channel_loss_enabled and batch.per_channel_losses is not None:
             for var_name, value in batch.per_channel_losses.items():
                 acc = self._per_channel_loss.get(
@@ -139,6 +151,11 @@ class TrainAggregator(AggregatorABC[TrainOutput]):
         logs[f"{label}/mean/loss"] = float(
             dist.reduce_mean(self._loss / self._n_loss_batches).cpu().numpy()
         )
+        for key in sorted(self._per_step_loss_sums.keys()):
+            count = self._per_step_loss_counts[key]
+            logs[f"{label}/mean/{key}"] = float(
+                dist.reduce_mean(self._per_step_loss_sums[key] / count).cpu().numpy()
+            )
         if self._n_loss_batches > 0 and self._per_channel_loss_enabled:
             for var_name, acc in self._per_channel_loss.items():
                 logs[f"{label}/mean/loss/{var_name}"] = float(
--- a/fme/ace/aggregator/train.py
+++ b/fme/ace/aggregator/train.py
@@ -6,6 +6,10 @@ import torch
 
 from fme.ace.aggregator.inference.data import InferenceBatchData, make_dummy_time
 from fme.ace.aggregator.inference.spectrum import PairedSphericalPowerSpectrumAggregator
+from fme.ace.aggregator.loss_metrics import (
+    PerChannelLossAggregator,
+    PerStepLossAggregator,
+)
 from fme.ace.aggregator.one_step.reduced import MeanAggregator
 from fme.ace.stepper import TrainOutput
 from fme.core.device import get_device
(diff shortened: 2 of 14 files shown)


fme/ace/aggregator/train.py
```diff
@@ -105,6 +107,16 @@ def __init__(self, config: TrainAggregatorConfig, operations: GriddedOperations)
     def record_batch(self, batch: TrainOutput):
         self._loss += batch.metrics["loss"]
         self._n_loss_batches += 1
+        for key, value in batch.metrics.items():
```
mcgibbon: This is kind of a pre-existing issue with the loss and also with the per-channel losses, but I'm not a fan of this top-level aggregator having low-level aggregation in it. Could you instead refactor out a metrics aggregator and a per-channel losses aggregator, and move those responsibilities to those objects?

fme/ace/stepper/single_module.py
```diff
@@ -1635,11 +1642,14 @@ def _accumulate_loss(
...
                 )
-            n_forward_steps = stochastic_n_forward_steps
+            n_loss_steps = stochastic_n_forward_steps
+            if not evaluate_all_steps:
+                n_forward_steps = stochastic_n_forward_steps
         per_channel_sum: dict[str, torch.Tensor] | None = None
         for step in range(n_forward_steps):
-            optimize_step = (
-                step == n_forward_steps - 1 or not self._config.optimize_last_step_only
+            counts_toward_loss = step < n_loss_steps
```
mcgibbon: nit: This logic is a little hard to follow, not your fault, but I have some specific name suggestions to make it clearer. `counts_towards_loss` could be called `within_optimization_window`. You could call `step == n_loss_steps - 1` a new temporary `last_optimization_window_step`.

Another option is a multi-tiered if-else instead of a one-liner. For example, `if within_optimization_window and not self._config.optimize_last_step_only: optimize_step = True`, with subsequent elif/else handling other cases. Up to you which one is clearer, since single-line is more concise.

fme/ace/stepper/single_module.py
```diff
@@ -1660,12 +1670,15 @@ def _accumulate_loss(
...
-                        per_channel_sum[k] = per_channel_sum[k] + per_ch[k].detach()
+                if counts_toward_loss:
+                    per_ch = step_loss.get_channel_losses()
+                    if per_channel_sum is None:
+                        per_channel_sum = {
+                            k: v.detach().clone() for k, v in per_ch.items()
+                        }
+                    else:
+                        for k in per_channel_sum:
+                            per_channel_sum[k] = per_channel_sum[k] + per_ch[k].detach()
```
mcgibbon: I believe this may be a bug in the existing code. The per-channel losses are supposed to align with the optimization loss, which we get later from `optimization.get_accumulated_loss().detach()`. That means we should only be adding per-channel losses when we `optimize_step`.

We need a test covering this, that ensures the total of the per-channel losses is not more than the `optimization.get_accumulated_loss()`.

If this is indeed the case, let's fix that issue in this PR. After that, open a new issue describing this, and update this PR's description to say it closes that issue.

fme/coupled/aggregator.py
```diff
@@ -61,18 +61,38 @@ class TrainAggregator(AggregatorABC[CoupledTrainOutput]):
...
         self._n_batches = 0
         self._loss = torch.tensor(0.0, device=get_device())
+        self._per_step_loss_sums: dict[str, torch.Tensor] = {}
+        self._per_step_loss_counts: dict[str, int] = {}
 
     @torch.no_grad()
     def record_batch(self, batch: CoupledTrainOutput):
         self._loss += batch.total_metrics["loss"]
         self._n_batches += 1
+        for component in (batch.ocean, batch.atmosphere):
```
mcgibbon, line 69: Similar comment for the ace aggregator, let's put these in their own sub-aggregator instead of directly here. That sub-aggregator could handle the `loss` itself as well.

fme/coupled/loss.py
```diff
@@ -41,6 +41,16 @@ def seed_rng(self, seed: int) -> None:
         """Seed the stochastic n_steps sampler for deterministic evaluation."""
         pass
 
+    def compute_loss(
```
mcgibbon, line 51: Rather than making a nonsensical base implementation, make this abstract.

fme/coupled/stepper.py
```diff
@@ -1670,11 +1678,18 @@ def _accumulate_step_loss(
             step=gen_step.step,
         )
         target_step_ensemble = add_ensemble_dim(target_step)
-        step_loss = self._loss(ensemble_step, target_step_ensemble)
-        if step_loss is not None:
+        if evaluate_all_steps:
+            step_loss = self._loss.compute_loss(ensemble_step, target_step_ensemble)
```
mcgibbon, line 1692: Note @jpdunc23 this usage of the internal `.compute_loss` seems a bit awkward, maybe this is something that goes away with your ongoing refactors for the step loss weighting. I could update this PR so it doesn't implement this change in the coupled code, then you could do it after those? Let me know what's best for you.

fme/ace/aggregator/train.py
mcgibbon: Claude: Done — extracted `PerStepLossAggregator` and `PerChannelLossAggregator` into `fme/ace/aggregator/loss_metrics.py`. `TrainAggregator` now delegates to these sub-aggregators. The per-channel aggregator uses `| None` typing (None = disabled) instead of a separate boolean flag.

fme/ace/stepper/single_module.py
mcgibbon: Claude: Done — renamed `counts_toward_loss` to `within_optimization_window` and extracted `last_optimization_window_step = n_loss_steps - 1`. Went with the single-liner approach since it reads cleanly with the new names.

fme/ace/stepper/single_module.py
mcgibbon: Claude: Confirmed — this was a bug. Per-channel losses now accumulate only when `optimize_step` is True, matching the steps that contribute to `optimization.get_accumulated_loss()`. Added `test_per_channel_losses_bounded_by_accumulated_loss` to verify. Opened #1172 for the issue; updated this PR description to close it.

fme/coupled/aggregator.py
mcgibbon: Claude: Done — coupled `TrainAggregator` now uses the shared `PerStepLossAggregator` from `fme/ace/aggregator/loss_metrics.py`, same as the ACE aggregators.

fme/coupled/loss.py
mcgibbon: Claude: Done — `compute_loss` is now `@abc.abstractmethod` on `StepLossABC`. Both `NullLossContributions` (returns zero) and `LossContributions` provide concrete implementations.

fme/coupled/stepper.py
mcgibbon: Claude: Noted — leaving this as-is for now. @jpdunc23 can decide whether to fold this into the ongoing step loss weighting refactors or if we should remove the coupled `evaluate_all_steps` implementation from this PR and let it land later.

fme/coupled/aggregator.py
```diff
@@ -61,18 +62,26 @@ class TrainAggregator(AggregatorABC[CoupledTrainOutput]):
...
+            step_metrics = {
+                k: v for k, v in component.metrics.items() if k.startswith("loss/")
+            }
+            self._per_step_losses.record(step_metrics)
```
mcgibbon, line 73: Question: I think this will give the same metric for atmosphere and ocean, meaning these get added. Should these losses get recorded like this, or should atmosphere and ocean have separate prefixes? Or am I wrong, and the components already contain these prefixes in their metrics?

jpdunc23: The keys include the realm prefix, see here:

https://github.com/ai2cm/ace/blob/5ed219f45722aa618af065c79303270e125cb0e1/fme/coupled/stepper.py#L1685

fme/coupled/stepper.py
jpdunc23: Yes, I agreed this is awkward, but will go away with https://github.com/ai2cm/ace/pull/1158. Fine to leave in.

fme/ace/aggregator/loss_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+        return logs
+
+
+class PerChannelLossAggregator:
+    """Accumulates per-channel (per-variable) loss values across batches."""
+
+    def __init__(self):
+        self._sums: dict[str, torch.Tensor] = {}
+        self._n_batches = 0
```
jpdunc23, line 39: Should this also include per-channel counts, like `_counts` in `PerStepLossAggregator`?

jpdunc23: In particular, the `per_channel_losses` keys could potentially change from one `record()` to another. I don't think this occurs in practice, but since `PerStepLossAggregator` tracks it, it probably makes sense to do the same here.

jpdunc23 (approved):
LGTM, but left a couple of comments, in particular asking about a potential "bug" in `PerChannelLossAggregator`.

fme/ace/aggregator/loss_metrics.py
mcgibbon: This support is added in https://github.com/ai2cm/ace/pull/1182, which explicitly adds support for some variables to be "missing" from the loss on some calls.

PR #1160, opened 2026-05-12
Add per-sample variable masking for heterogeneous data sources

mcgibbon:
Enables training and inference on heterogeneous data sources where some datasets are missing certain variables. When `allow_missing_variables: true` is set on `ModuleSelector`, missing input variables are zeroed in normalized space (equivalent to climatological mean) and missing output variables are excluded from loss computation.

The flag is named `allow_missing_variables` rather than `allow_variable_masking` to describe what the user is permitting (a data source that may omit variables) rather than the mechanism (per-variable masks threaded through the batch). The internal carrier on `BatchData` is still `data_mask`, since at that level it really is a per-variable boolean mask — the renaming is only of the user-facing config / requirements flag.

Changes:
- `BatchData`: add `data_mask: TensorMapping | None` field, propagated through all transforms
- `ModuleSelector`: add `allow_missing_variables: bool` flag
- `DataRequirements`: add `allow_missing_variables: bool` flag
- `StepArgs`: add `data_mask` parameter, preserved through `apply_input_process_func`
- `StepConfigABC`, `StepSelector`: add `allow_missing_variables` property delegation
- `SingleModuleStep.step`: apply `_apply_input_mask` to zero masked inputs in normalized space before network call
- `TrainStepper._accumulate_loss`: apply `_apply_output_mask` to zero masked predictions and targets before loss
- `Stepper.predict_generator`, `get_prediction_generator`: thread `data_mask` through to `StepArgs`
- `XarrayDataset`: skip missing variables instead of raising when `allow_missing_variables=True`
- `CollateFn`, `BatchData.from_sample_tuples`: handle heterogeneous variable sets via `_collate_with_masking`, producing per-variable boolean masks
- `DatasetConfigABC.build` chain (`XarrayDataConfig`, `ConcatDatasetConfig`, `MergeDatasetConfig`): accept and forward `allow_missing_variables`
- `get_gridded_data`, `DataLoaderConfig.get_dataset`: pass `allow_missing_variables` from requirements
- `InferenceDataset`, `get_inference_data`, `get_forcing_data`: pass `allow_missing_variables` to dataset and collation

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

PR #1174, opened 2026-05-14
Fix masked variable handling

yyexela:
This PR fixes two related bugs in the masked-variable pipeline and cleans up the loss masking implementation.

  Bug fix — `XarrayDataset._names` inconsistency (`fme/core/dataset/xarray.py`)

  When `allow_variable_masking=True`, `_names` was being silently overwritten in `__init__` to only contain the
  variables actually found on disk, diverging from the originally requested names. This caused downstream logic
  (e.g. `_group_variable_names_by_time_type`) to skip missing variables even when masking was expected to handle
  them. The fix renames the constructor-stored attribute to `_requested_names` (used only for grouping) and makes
  `_names` a computed property that always returns the union of time-dependent, time-invariant, and static-derived
   variables discovered in the dataset.

  Cleanup — remove `_apply_output_mask` (`fme/ace/stepper/single_module.py`)

  The `_apply_output_mask` helper, which zeroed out masked tensors before the loss call, is removed.
  `WeightedMappingLoss` already handles per-sample masking via `data_mask` directly, so the pre-zeroing step was
  redundant and added surface area for subtle mismatches.

  Minor improvements

  - `_collate_with_masking` in `batch_data.py` simplified to a single `dict.fromkeys` expression.
  - Docstrings for `_reduce_to_per_channel` and `BatchData` expanded for clarity.
  - Tests across `test_xarray.py`, `test_step.py`, `test_loss.py`, `test_batch_data.py`, and `test_data_loader.py` made
  more robust, descriptive, and comprehensive for masked-variable scenarios.

Changes:
- `fme.core.dataset.xarray.XarrayDataset` — fixed `_names` inconsistency when `allow_variable_masking=True`; `_names`
is now a computed property always reflecting discovered variables, with `_requested_names` used internally for
grouping
- `fme.ace.stepper.single_module._apply_output_mask`, `fme.ace.stepper.single_module.TrainStepper` — removed
`_apply_output_mask` and its call site; masking is handled by `WeightedMappingLoss` directly
- `fme.ace.data_loading.batch_data._collate_with_masking`, `fme.ace.data_loading.batch_data.BatchData` —
simplified unique-name collection; improved docstring
- `fme.core.loss._reduce_to_per_channel` — expanded docstring

- [x] Tests added


fme/core/dataset/test_xarray.py
```diff
@@ -1249,6 +1249,21 @@ def test_allow_variable_masking_skips_missing_variables(mock_monthly_netcdfs):
...
+    dataset = XarrayDataset(
+        config,
+        names_with_missing,
+        IntSchedule.from_constant(2),
+        allow_variable_masking=True,
+    )
+    assert "nonexistent_var" in dataset._requested_names
+    assert "nonexistent_var" not in dataset._names
+
+
```
yyexela, line 1266: Note: This test is used to show that the `_requested_names` variable could have more names than present in a dataset, which is populated in `_names`.

fme/core/dataset/xarray.py
```diff
@@ -750,10 +748,6 @@ def _get_files_stats(
 
         del cum_num_timesteps
 
-        ds = self._open_file(0)
-        self._get_variable_metadata(ds)
-        ds.close()
```
yyexela: This was moved to the bottom of `__init__` to ensure `_names` is updated before this is called. I think it looks cleaner this way anyways.

yyexela:
Added detailed explanations of all changes to https://github.com/ai2cm/ace/pull/1160

PR #1175, opened 2026-05-15
Fix coordinate equality device mismatch

frodre:
Running the latest main with our published HiRO ckpt revealed that `CheckpointModelConfig` with provided fine coordinates path crashed with mismatched devices for the `LatLonCoordinates.__eq__` check, which is enforced during building (prior to device coercing in the `DiffusionModel.__init__`).  

This is a specific case of a model checkpoint with an intermediate old style format that can no longer be created.  However, I felt that the equality shouldn't be what fails since the equality is a value check.


fme/downscaling/models.py
```diff
@@ -813,7 +813,7 @@ def _get_coords_backwards_compatible(
                 lon=coords_from_state["lon"].to(get_device(), copy=True),
             )
         elif fine_coordinates_path is not None:
-            return load_coords_from_path(fine_coordinates_path)
+            return load_coords_from_path(fine_coordinates_path).to(get_device())
```
frodre, line 834: To make consistent with the other coordinate loading pathway.

mcgibbon (approved):
LGTM

PR #1176, opened 2026-05-15
Fix time_mean variable filter incorrectly affecting channel_mean RMSE

mcgibbon:
The `variables` config on `TimeMeanMetricConfig` was using `VariableFilterAdapter` to strip variables from input data before the aggregator saw them, which caused the `channel_mean` RMSE (used for checkpoint selection) to be computed over only the filtered subset instead of all `channel_mean_names`. Additionally, disabling `time_mean_norm` on a weighted inference entry would only fail at runtime rather than at config validation.

Changes:
- `TimeMeanAggregator`, `TimeMeanEvaluatorAggregator`: add `log_variables` parameter that controls which per-variable metrics are reported in `get_logs` and `get_dataset`, while always computing RMSE for all variables so that `channel_mean` is correct
- `TimeMeanMetricConfig.build`: pass `variables` as `log_variables` to the aggregator instead of wrapping with `VariableFilterAdapter`
- `InlineInferenceConfig.__post_init__`: reject configs where `time_mean_norm.enabled` is `False` but `weight > 0`

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/test_time_mean.py
+++ b/fme/ace/aggregator/inference/test_time_mean.py
@@ -147,6 +147,139 @@ def test_bias_values():
     )
 
 
+def test_log_variables_does_not_affect_channel_mean():
+    torch.manual_seed(0)
+    area_weights = torch.ones(1, 1).to(get_device())
+    target_data_norm = {
+        "a": torch.ones([2, 3, 4, 4], device=get_device()),
+        "b": torch.ones([2, 3, 4, 4], device=get_device()) * 3,
+    }
+    gen_data_norm = {
+        "a": torch.ones([2, 3, 4, 4], device=get_device()) * 2.0,
+        "b": torch.ones([2, 3, 4, 4], device=get_device()) * 5,
+    }
+    batch = InferenceBatchData(
+        prediction=gen_data_norm,
+        prediction_norm=gen_data_norm,
+        target=target_data_norm,
+        target_norm=target_data_norm,
+        time=make_dummy_time(2, 3),
+        i_time_start=0,
+    )
+    agg = TimeMeanEvaluatorAggregator(
+        LatLonOperations(area_weights),
+        horizontal_dims=["lat", "lon"],
+        target="norm",
+        log_variables=frozenset(["a"]),
+    )
+    agg.record_batch(batch)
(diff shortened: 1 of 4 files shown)


mcgibbon:
Claude: Noting the behavioral changes in this PR for reviewers:

1. **`channel_mean` RMSE now always uses all variables (or `channel_mean_names`), even when `variables` filter is set.** Previously, `VariableFilterAdapter` filtered input data before the aggregator saw it, so `channel_mean` was computed only over filtered variables. Now filtering happens at output time, so `channel_mean` reflects the full variable set. Anyone who had `variables: [temperature]` on `time_mean_norm` and relied on `channel_mean` only reflecting that subset should switch to `channel_mean_names: [temperature]`.

2. **New validation rejects `weight > 0` with `time_mean_norm.enabled: false`.** Existing configs with that combination will now fail at startup with a clear error message, rather than silently proceeding without the metric needed for checkpoint selection.

3. **`get_images()` and `get_dataset()` on `TimeMeanAggregator` now also respect `log_variables`.** The `log_variables` filter is passed to the gen aggregator but not the target aggregator, which mirrors the old behavior where only the evaluator was wrapped by `VariableFilterAdapter`.

mcgibbon:
Those are all expected changes, the current main behavior is a bug.

AnnaKwa (approved):
LGTM. Question outside the scope of this PR- would we want to have an option to exclude variables from the channel mean RMSE calculation (the 'bugged' behavior fixed here)? e.g. preventing the inference checkpoint selection from being dominated by q0

mcgibbon:
> LGTM. Question outside the scope of this PR- would we want to have an option to exclude variables from the channel mean RMSE calculation (the 'bugged' behavior fixed here)? e.g. preventing the inference checkpoint selection from being dominated by q0

Perhaps, in particular we really might want to select based on some small subset of variables we really care about (and not just exclude q0).

However, the whole purpose of selecting checkpoints based on inference was that we were getting really poor training dynamics, with inference getting worse the longer we trained. In many of our newer runs this isn't the case, and especially this wouldn't be the case if we excluded q0 (which has regime shifts in the real data, leading to "regime shifts" in the skill that are really just from different somewhat-valid prediction series). Where that's the case, we should consider selecting checkpoints based on validation skill.

PR #1178, opened 2026-05-15
Oversample tropics option in downscaling training

AnnaKwa:
This PR adds an optional section `region_sampling: RegionSamplingConfig` to the downscaling training config. The `RegionSamplingConfig` class can be used to specify a lat/lon interval range and weight. When the full training data lat/lon extent is divided into `n_patches`, if this option is configured then patches with coordinate centers within the lat/lon interval will be sampled with weight `weight`. Patches outside the region are sampled with weight=1.0. `n_patches` are sampled with replacement using the assigned weights. If no sampling config is provided, there is no change from previous behavior (return all patches). 

If either `RegionSamplingConfig.lat_interval, lon_interval` are None, all lat or lon coordinates are considered to be contained in the sampling config region.

- [x] Tests added

### DIFF
--- a/fme/downscaling/data/datasets.py
+++ b/fme/downscaling/data/datasets.py
@@ -357,7 +357,7 @@ class PairedGriddedData:
         drop_partial_patches: bool = True,
         random_offset: bool = False,
         shuffle: bool = False,
-        tropical_oversampling: "TropicalOversamplingConfig | None" = None,
+        tropical_oversampling: "RegionOversamplingConfig | None" = None,
     ) -> Iterator["PairedBatchData"]:
         patched_generator = patched_batch_gen_from_paired_loader(
             self.loader,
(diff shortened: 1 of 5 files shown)


AnnaKwa:
LMK your thoughts on whether this should be able to handle multiple regions. That could be useful also emphasizing the regions with complex topography in addition to tropical basins, but would need some more complex checking about overlaps.

fme/downscaling/data/datasets.py
```diff
@@ -2,6 +2,7 @@
 
 import dataclasses
 import math
+import random
```
frodre, line 5: I know `fme.core.rand` has some seed setting capabilities, but I don't think we've ever used any of the "reproducibility" seeding in the downscaling, so it's probably okay not to worry about that here.  Just commenting for visibility.

fme/downscaling/data/datasets.py
```diff
@@ -728,4 +793,15 @@ def patched_batch_gen_from_paired_loader(
             shuffle=shuffle,
             drop_partial_patches=drop_partial_patches,
         )
+        if region_sampling is not None:
+            assert fine_patches is not None  # downscale_factor was provided
```
frodre: This could probably be a more informative error message, but I don't think it's actually possible to reach this error from within the paired loader.  Could change the comment to say it's just for typing check.

fme/downscaling/data/datasets.py
```diff
@@ -728,4 +793,15 @@ def patched_batch_gen_from_paired_loader(
...
+            coarse_lats = batch.coarse.latlon_coordinates.lat[
+                0
+            ]  # dims are [batch, lat, lon]
```
frodre:
```suggestion
            ]  # dims are [batch, lat]
```

The need to deal with the `batch` dim will get fixed when I eventually remove the use of batched coordinates, which was only relevant for the per-sample random patch locations.

fme/downscaling/data/test_datasets.py
```diff
@@ -462,3 +466,156 @@ def test_batch_data_apply_patch_already_patched_raises():
...
+            patches, coarse_lats, coarse_lons, config
+        )
+        for idx in indices:
+            counts[idx] += 1
+    total = sum(counts)
+    # Only patch 0 (lat ~ -30, lon ~ 38.6) is inside both intervals
+    assert counts[0] / total > 0.99
+
+
+def _make_paired_batch_for_tropical(
```
frodre: is "tropical" relevant to the paired batch created? Doesn't seem so.

fme/downscaling/train.py
```diff
@@ -395,6 +404,28 @@ def train(self) -> None:
 
 @dataclasses.dataclass
 class TrainerConfig:
+    """
```
frodre, line 407: Makes me think we should have a new `patching' section with it's own configuration similar to how we do for inference-type entrypoints.  That would at least allow for easier grouping and distinction for documentation and error checking instead of at the top level of the training configuration.  Not required for this PR since it would involve a deprecation.

frodre (approved):
Thanks @AnnaKwa.  I like this implementation for upweighting certain regions during training and could imagine this being extended in interesting ways.  I don't think we have to worry about multi-region for now.  Just added a few comments for thought, but no changes required from me.

PR #1182, opened 2026-05-18
Add optional mask to LossOutput for per-sample variable masking

mcgibbon:
Adds an optional `(B, C)` float mask to `LossOutput` so that when heterogeneous data sources provide different variable sets per sample, masked-out channels don't dilute the loss average. Per-channel losses are averaged only over samples where that channel is present, and fully-absent channels are excluded from the scalar `total()`.

Changes:
- `fme.core.loss.LossOutput`: accepts optional `mask` parameter; `_reduce()`, `total()`, `get_channel_losses()`, and `scale()` all respect the mask
- `fme.core.loss.ChannelLossInfo`: new dataclass pairing a per-channel loss with its active-sample count, returned by `get_channel_losses()`
- `fme.core.loss.WeightedMappingLoss.__call__` and `fme.core.loss.StepLoss.forward`: accept and forward `data_mask: TensorMapping | None`, using `Packer` to build the `(B, C)` channel mask
- `fme.ace.stepper.single_module.TrainStepper._accumulate_loss`: tracks weighted sums and counts across steps
- `fme.ace.aggregator.train.TrainAggregator`: uses count-weighted averaging instead of simple batch-count division

Note: `data_mask` is plumbed through `WeightedMappingLoss` and `StepLoss` but is not yet passed from the stepper's training loop — this PR provides the infrastructure, and a follow-up will wire the mask from the data loader.

- [x] Tests added

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -42,7 +42,7 @@ from fme.core.generics.inference import PredictFunction
 from fme.core.generics.optimization import OptimizationABC
 from fme.core.generics.train_stepper import TrainOutputABC, TrainStepperABC
 from fme.core.labels import BatchLabels
-from fme.core.loss import StepLoss, StepLossConfig
+from fme.core.loss import ChannelLossInfo, StepLoss, StepLossConfig
 from fme.core.masking import NullMasking, StaticMaskingConfig
 from fme.core.normalizer import (
     NetworkAndLossNormalizationConfig,
@@ -345,7 +345,7 @@ class TrainOutput(TrainOutputABC):
     derive_func: Callable[[TensorMapping, TensorMapping], TensorDict] = lambda x, _: (
         dict(x)
     )
-    per_channel_losses: dict[str, torch.Tensor] | None = None
+    per_channel_losses: dict[str, ChannelLossInfo] | None = None
 
     def __post_init__(self):
         for v in self.target_data.values():
@@ -1605,7 +1605,7 @@ class TrainStepper(
         target_data: BatchData,
         optimization: OptimizationABC,
         metrics: dict[str, float],
-    ) -> tuple[list[EnsembleTensorDict], dict[str, torch.Tensor] | None]:
+    ) -> tuple[list[EnsembleTensorDict], dict[str, ChannelLossInfo] | None]:
         input_data = data.get_start(self._prognostic_names, self.n_ic_timesteps)
         # output from self.predict_paired does not include initial condition
         n_forward_steps = data.time.shape[1] - self.n_ic_timesteps
@@ -1643,7 +1643,8 @@ class TrainStepper(
                     "data requirements are retrieved, so this is a bug."
                 )
             n_forward_steps = stochastic_n_forward_steps
-        per_channel_sum: dict[str, torch.Tensor] | None = None
+        weighted_sums: dict[str, torch.Tensor] | None = None
+        total_counts: dict[str, int] | None = None
         for step in range(n_forward_steps):
             optimize_step = (
                 step == n_forward_steps - 1 or not self._config.optimize_last_step_only
@@ -1668,14 +1669,28 @@ class TrainStepper(
                 step_total_loss = step_loss.total()
                 metrics[f"loss_step_{step}"] = step_total_loss.detach()
                 per_ch = step_loss.get_channel_losses()
-                if per_channel_sum is None:
-                    per_channel_sum = {k: v.detach().clone() for k, v in per_ch.items()}
+                if weighted_sums is None:
+                    weighted_sums = {
+                        k: v.loss.detach() * v.count for k, v in per_ch.items()
+                    }
+                    total_counts = {k: v.count for k, v in per_ch.items()}
                 else:
-                    for k in per_channel_sum:
-                        per_channel_sum[k] = per_channel_sum[k] + per_ch[k].detach()
+                    assert total_counts is not None
+                    for k, v in per_ch.items():
+                        weighted_sums[k] = weighted_sums[k] + v.loss.detach() * v.count
+                        total_counts[k] = total_counts[k] + v.count
             if optimize_step:
                 optimization.accumulate_loss(step_total_loss)
-        return output_list, per_channel_sum
+        per_channel_losses: dict[str, ChannelLossInfo] | None = None
+        if weighted_sums is not None and total_counts is not None:
+            per_channel_losses = {
+                k: ChannelLossInfo(
+                    loss=weighted_sums[k] / max(total_counts[k], 1),
+                    count=total_counts[k],
+                )
+                for k in weighted_sums
+            }
+        return output_list, per_channel_losses
 
     def update_training_history(self, training_job: TrainingJob) -> None:
         """
--- a/fme/core/loss.py
+++ b/fme/core/loss.py
@@ -15,6 +15,14 @@ from fme.core.packer import Packer
 from fme.core.typing_ import TensorMapping
 
 
+@dataclasses.dataclass
+class ChannelLossInfo:
+    """Per-channel loss value and the number of batch samples that contributed."""
+
+    loss: torch.Tensor
+    count: int
+
+
 class LossComponent(abc.ABC):
     """A pre-weighted loss tensor that knows how to reduce itself to ``(B, C)``.
 
@@ -75,6 +83,7 @@ class LossOutput:
         self._channel_names = channel_names
         self._mask = mask
         self._per_channel: torch.Tensor | None = None
+        self._counts: list[int] | None = None
 
     def _get_per_channel(self) -> torch.Tensor:
         """Return a ``(C,)`` tensor of per-channel losses, computed once.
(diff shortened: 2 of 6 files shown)


fme/core/loss.py
```diff
@@ -63,40 +69,89 @@ def __init__(
...
         """Scalar loss — the optimisation target."""
-        return self._get_per_channel().mean()
+        pc = self._get_per_channel()
+        if self._mask is not None:
+            active = self._mask.sum(dim=0) > 0
+            if active.any():
+                return pc[active].mean()
+        return pc.mean()
 
     def get_channel_losses(self) -> dict[str, torch.Tensor]:
```
mcgibbon: We actually need to return not just the losses now, but also the counts. This is crucial because this data is ingested downstream by an aggregator that computes the mean loss of different channels. In that aggregator, we need to track the sums and totals so that we weight each contribution equally.

Design a better API for this, possibly returning a dict from str to namedtuple or dataclass.

fme/core/loss.py
```diff
@@ -202,9 +262,20 @@ def __call__(
             )
             component_cls = StandardLoss if cdim <= 1 else EnsembleComponentLoss
             losses = [component_cls(result)]
+
+        mask = None
+        if data_mask is not None:
+            mask = _build_channel_mask(
```
mcgibbon: Is it possible to directly use the `packer` that's packing the other variables, instead of pulling out the low-level detail of thee name list and relying on coupling between the packer code and _build_channel_mask? If you need to, you can add a new method to the packer for packing this differently-shaped data, or otherwise change the API in small backwards compatible ways as you need to, but I thought it would "just work".

fme/core/loss.py
mcgibbon: Make sure to test that aggregator, also.

mcgibbon:
Pre-review agent: Note for reviewer — `data_mask` is accepted by `WeightedMappingLoss.__call__` and `StepLoss.forward` but is not yet passed from `TrainStepper._accumulate_loss` (line 1668 of `single_module.py` still calls `self._loss_obj(gen_step, target_step, step=step)` without a `data_mask` kwarg). This is intentional: the PR adds the mask infrastructure and a follow-up will wire the mask from the data loader. Flagging so there's no confusion about whether it was accidentally omitted.

fme/core/loss.py
```diff
@@ -63,37 +77,73 @@ def __init__(
...
+                self._per_channel = masked_sum / self._mask.sum(dim=0).clamp(min=1)
+                self._counts = [int(c.item()) for c in self._mask.sum(dim=0)]
             else:
                 self._per_channel = bc.mean(dim=0)
-        return self._per_channel
+                self._counts = [bc.shape[0]] * len(self._channel_names)
+        assert self._per_channel is not None and self._counts is not None
+        return self._per_channel, self._counts
 
     def total(self) -> torch.Tensor:
```
Arcomano1234, line 111: Question: Doesn't this change the behavior of loss.total() as its now a mean?

fme/ace/stepper/single_module.py
```diff
@@ -1668,14 +1669,28 @@ def _accumulate_loss(
                 step_total_loss = step_loss.total()
                 metrics[f"loss_step_{step}"] = step_total_loss.detach()
                 per_ch = step_loss.get_channel_losses()
-                if per_channel_sum is None:
-                    per_channel_sum = {k: v.detach().clone() for k, v in per_ch.items()}
+                if weighted_sums is None:
```
Arcomano1234: Claude flagged this: `_accumulate_loss` silently drops channels that appear in later steps but not the first . I don't think this is actually possible for our code to allow this, but I was curious if you think this is possible

Arcomano1234 (commented):
Left a few comments / questions that should be addressed but looks close to ready.

fme/ace/stepper/single_module.py
mcgibbon: The way we've structured the APIs right now, the step should always compute and return all channels, regardless of whether they're included in the loss. Perhaps we can add a test to at least SingleModuleStep that ensures this. If it's inexpensive computationally, we could add a safety guard here, but if it's costly we shouldn't.

fme/core/loss.py
mcgibbon: The total loss was always a mean across channels - The replaced implementation was `return self._get_per_channel().mean()`. Our expected loss didn't go up and down when we added/removed channels. I do acknowledge the docstring for get_channel_losses was incorrect and that `.total()` is perhaps confusing, but it's true if you consider it to be the total contribution of all loss terms and don't read it as the total of the losses across sample/space/channel (each of which would be untrue, not just channel).

fme/core/loss.py
```diff
@@ -15,6 +15,14 @@
 from fme.core.typing_ import TensorMapping
 
 
+@dataclasses.dataclass
+class ChannelLossInfo:
+    """Per-channel loss value and the number of batch samples that contributed."""
+
+    loss: torch.Tensor
+    count: int
```
yyexela, line 23: Nitpick (non-blocking): There's a random newline here

fme/ace/stepper/single_module.py
Arcomano1234: Yeah I agree its not worth having the safe guard but did want to point it out.

fme/core/test_loss.py
```diff
@@ -476,6 +478,185 @@ def test_finite_difference_crps_rejects_zero_levels():
...
+        normalizer=normalizer,
+    )
+    x_mapping = {name: torch.ones(4, 5, 5).to(get_device()) for name in out_names}
+    y_mapping = {name: torch.zeros(4, 5, 5).to(get_device()) for name in out_names}
+    data_mask = {
+        "var_0": torch.tensor([True, True, True, True]),
+        "var_1": torch.tensor([False, False, False, False]),
+    }
+    result = step_loss(x_mapping, y_mapping, step=0, data_mask=data_mask)
+    torch.testing.assert_close(result.total(), torch.tensor(1.0, device=get_device()))
```
yyexela, line 631: Issue: This test doesn't actually check if the mask affects the output. The issue is that `x_mapping` has the values of `var_0` and `var_1` to be 1, meaning whether or not `var_1` is masked out, the expected loss total is 1.0.

Suggestion: Update `x_mapping`:

```
      x_mapping = {
          "var_0": torch.ones(4, 5, 5).to(get_device()),
          "var_1": torch.full((4, 5, 5), 2.0).to(get_device()),
      }
```

Then, the result should be 1.0 if masked (the assert already handles this), or 2.5 if the mask doesn't work.

fme/core/test_loss.py
```diff
@@ -476,6 +478,185 @@ def test_finite_difference_crps_rejects_zero_levels():
...
+    torch.testing.assert_close(mask, expected)
+
+
+def test_packer_mask_missing_name_defaults_to_present():
+    packer = Packer(["a", "b", "c"])
+    filled = {
+        "a": torch.tensor([1.0, 0.0]),
+        "b": torch.ones(2),
+        "c": torch.tensor([0.0, 1.0]),
+    }
```
yyexela, line 505: Nitpick (non-blocking):

I think it's nicer to just have `"b": torch.tensor([1.0, 1.0]),` here, kind of odd to use torch.ones for readability.

yyexela (changes requested):
Just one test comment and two nit-picks, overall looks great! Thanks:)

fme/core/test_loss.py
mcgibbon: Agreed.

fme/core/test_loss.py
mcgibbon: Sounds reasonable to me.

mcgibbon:
Claude: addressed the outstanding comments in a05838ea1.

- @yyexela `test_loss.py:626` — `test_step_loss_forwards_data_mask` now sets `var_1` to `2.0` so the assertion only passes if masking actually drops it (would otherwise read `2.5`).
- @yyexela `test_loss.py:505` — `"b"` now uses `torch.tensor([1.0, 1.0])`.
- @mcgibbon / @Arcomano1234 `single_module.py:1672` thread — added `test_train_on_batch_per_channel_losses_include_zero_weighted_channels`, which configures the step loss with `weights={"b": 0.0}` and asserts every out channel still shows up in `per_channel_losses` with a positive count. Locking this down at the SingleModuleStep level was the lightweight option you both landed on instead of a runtime safeguard.
- @yyexela `loss.py:23` random-newline nit — left as-is; ruff-format insists on the blank line between the dataclass docstring and the first field.

The earlier API/coverage items (`ChannelLossInfo` return type, packer-built mask, aggregator weighting-by-count tests) were already in prior commits on this branch.

yyexela:
Happy with the changes:) ty!

fme/core/loss.py
Arcomano1234: Yes that is very true. To avoid user confusion in the future can we just add a sentence to the doc string for this function. Also `Scalar loss — the optimisation target` I didn't realize claude was British lol

Arcomano1234 (approved):
I think this should be ready to merge. I left one comment regarding the change to `LossComponent.total` doc string to make it more clear, but other than that it LGTM.

yyexela (approved):
I am happy w this:)

PR #1184, opened 2026-05-18
Fix `_get_variable_metadata` silently dropping variables

### LINKED ISSUE #1183: `_get_variable_metadata` silently drops variables without `units`/`long_name`
`XarrayDataset._get_variable_metadata` only populates entries for variables that have both `units` and `long_name` attributes. Variables lacking either attribute are silently omitted from `variable_metadata`, causing it to under-report the set of loaded variables.

This means `dataset.properties.variable_metadata.keys()` is not a reliable way to enumerate a dataset's variables, even though it looks like it should be.

### Fix

Add a fallback in the `else` branch so every variable in `self._names` gets an entry:

```python
else:
    result[name] = VariableMetadata(units="", long_name="")
```

jpdunc23:
`XarrayDataset._get_variable_metadata` silently omits variables that lack `units` or `long_name` attributes, causing `variable_metadata` to under-report the loaded variable set.

Changes:
- `VariableMetadata` now supports null `units` and/or `long_name`. Added helper methods `from_attrs()` (class method), `as_attrs()`, `display_name()` and `display_units()`.
- `fme.core.dataset.xarray.XarrayDataset._get_variable_metadata`: use null `VariableMetadata` attrs for variables without `units`/`long_name` attrs
- The `_variable_metadata_to_dict` util used by `ZarrWriterAdapter` and `SeparateICZarrWriterAdapter` filters out `None` values for `VariableMetadata`, preserving existing behavior.

- [x] Tests added

Resolves #1183

### DIFF
--- a/fme/ace/data_loading/test_metadata.py
+++ b/fme/ace/data_loading/test_metadata.py
@@ -108,7 +108,8 @@ def test_metadata(tmp_path, variable_metadata, n_ensemble_members):
     data = get_gridded_data(config=config, train=True, requirements=requirements)
     target_metadata = {
         name: variable_metadata[name]
-        for name in variable_metadata
         if variable_metadata[name] is not None
+        else VariableMetadata(units="", long_name="")
+        for name in variable_metadata
     }
     assert data.variable_metadata == target_metadata  # type: ignore
--- a/fme/ace/data_loading/test_metadata.py
+++ b/fme/ace/data_loading/test_metadata.py
@@ -39,12 +39,10 @@ def _save_netcdf(
         if len(dim_sizes) > 0:
             data = data.astype(np.float32)
         item_metadata = variable_metadata[name]
-        attrs: dict[str, str] = {}
         if item_metadata is not None:
-            if item_metadata.units is not None:
-                attrs["units"] = item_metadata.units
-            if item_metadata.long_name is not None:
-                attrs["long_name"] = item_metadata.long_name
+            attrs = item_metadata.as_attrs()
+        else:
+            attrs = {}
         data_vars[name] = xr.DataArray(data, dims=list(dim_sizes), attrs=attrs)
     coords = {
         dim_name: xr.DataArray(
--- a/fme/ace/data_loading/test_metadata.py
+++ b/fme/ace/data_loading/test_metadata.py
@@ -39,13 +39,12 @@ def _save_netcdf(
         if len(dim_sizes) > 0:
             data = data.astype(np.float32)
         item_metadata = variable_metadata[name]
-        if item_metadata is None:
-            attrs = {}
-        else:
-            attrs = {
-                "units": item_metadata.units,
-                "long_name": item_metadata.long_name,
-            }
+        attrs: dict[str, str] = {}
+        if item_metadata is not None:
+            if item_metadata.units is not None:
+                attrs["units"] = item_metadata.units
+            if item_metadata.long_name is not None:
+                attrs["long_name"] = item_metadata.long_name
         data_vars[name] = xr.DataArray(data, dims=list(dim_sizes), attrs=attrs)
     coords = {
         dim_name: xr.DataArray(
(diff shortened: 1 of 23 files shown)


fme/core/dataset/xarray.py
```diff
@@ -703,6 +703,8 @@ def _get_variable_metadata(self, ds):
                     units=ds[name].units,
                     long_name=ds[name].long_name,
                 )
+            else:
```
mcgibbon, line 719: This code will still silently drop units or long name if the other one is missing, in other words it is all or nothing.

Previously we could disambiguate between units=“” (dimensionless) and missing units, now we can’t. Should we use None in some way to keep this ability, like making this a dict[str, VariableMetadata | None] or by making long name and units optional?

jpdunc23: Good point, but I think the `and` meant it was silently dropped if either units or long name were missing? In any case, good idea to keep what is available, will update.

mcgibbon: Yes, these are two separate issues - the first one is pre-existing but should probably get fixed given the scope of this PR. Let's say it's Optional.

The second issue is new in this PR, and is particularly worse if you were to fix the first issue (since long_name="" is kind of a sentinel to say the data is missing, while units="" is a valid value).

fme/core/dataset/xarray.py
```diff
@@ -716,13 +716,11 @@ def _get_variable_metadata(self, ds):
         for name in self._names:
             if name in StaticDerivedData.names:
                 result[name] = StaticDerivedData.metadata[name]
-            elif hasattr(ds[name], "units") and hasattr(ds[name], "long_name"):
+            else:
                 result[name] = VariableMetadata(
-                    units=ds[name].units,
-                    long_name=ds[name].long_name,
+                    units=getattr(ds[name], "units", None),
+                    long_name=getattr(ds[name], "long_name", None),
```
mcgibbon: Suggestion (optional): ds[name].attrs.get("units") and ds[name].attrs.get("long_name")

Or even a `VarableMetadata.from_attrs(ds[name].attrs)`

fme/ace/data_loading/test_metadata.py
```diff
@@ -39,13 +39,12 @@ def _save_netcdf(
...
-        if item_metadata is None:
-            attrs = {}
-        else:
-            attrs = {
-                "units": item_metadata.units,
-                "long_name": item_metadata.long_name,
-            }
+        attrs: dict[str, str] = {}
+        if item_metadata is not None:
+            if item_metadata.units is not None:
```
mcgibbon: Suggestion (optional): This is already copy-pasted once and may be more in the future, you could use a new helper method like `item_metadata.as_attrs()` or `item_metadata.asdict()`

fme/downscaling/aggregators/no_target.py
```diff
@@ -146,12 +146,12 @@ def _plot_timeseries(
         plt.plot(times, coarse_data, color="black", label="coarse", linestyle="-.")
     plt.setp(ax.get_xticklabels(), rotation=45, ha="right")
     plt.xlabel("time")
+    caption_name, units = fine_data.name, "unknown_units"
     if variable_metadata is not None:
-        if fine_data.name in variable_metadata:
-            caption_name = variable_metadata[fine_data.name].long_name
-            units = variable_metadata[fine_data.name].units
-    else:
-        caption_name, units = fine_data.name, "unknown_units"
```
jpdunc23, line 154: This seems to be an existing bug because `_plot_timeseries` is passed the `VariableMetadata` object, not the dict of metadata, so it doesn't make sense to do `variable_metadata[fine_data.name]`. The condition `if fine_data.name in variable_metadata` would never pass.

mcgibbon (approved):
LGTM

jpdunc23:
@mcgibbon there were some remaining typing issues in the aggregators which required some further changes, so I think it would be best if you had another quick look when you are back next week.

jpdunc23:
> @mcgibbon there were some remaining typing issues in the aggregators which required some further changes, so I think it would be best if you had another quick look when you are back next week.

Whoops, forgot to turn off auto merge. Lmk if you have any concerns about the aggregator changes,

PR #1187, opened 2026-05-19
Stabilize derived-forcings train test against NaN flake on GPU

### LINKED ISSUE #1186: Flaky GPU test: test_train_and_inference_with_derived_forcings[solar-constant-as-name] NaN loss
## Summary

`fme/ace/test_train.py::test_train_and_inference_with_derived_forcings[solar-constant-as-name]` is intermittently failing on the GPU CI job with a NaN loss during training.

## Observed failure

- Workflow run: https://github.com/ai2cm/ace/actions/runs/26047509812/job/76575453461?pr=1171
- PR: #1171
- Job: `gpu`
- Commit (merge ref): `8ba9b79d0e9c53712f230b94bcd34026eab6577a`

```
FAILED fme/ace/test_train.py::test_train_and_inference_with_derived_forcings[solar-constant-as-name]
  - ValueError: Loss is NaN-valued during training.
```

Traceback originates from `fme/core/optimization.py:269` via `train_main` in `fme/ace/train/train.py:272`, called from `fme/ace/test_train.py:1046`.

Only the `solar-constant-as-name` parametrization failed; the sibling `solar-constant-as-value` case passed in the same run. The PR under test (#1171) is unrelated to training loss / optimization, which is consistent with this being a pre-existing flake rather than a regression.

## Suggested next steps

- Check whether this test has surfaced as flaky on other recent GPU runs (search `actions` logs for the same `ValueError`).
- Consider seeding more aggressively, lowering the learning rate for this test, or otherwise stabilizing the toy training so a single bad init can't produce NaNs.
- If the underlying training pipeline can legitimately diverge in one step, decide whether the test should retry / be marked flaky rather than guarded by `raise ValueError`.

mcgibbon:
The `test_train_and_inference_with_derived_forcings` test occasionally fails on GPU with `ValueError: Loss is NaN-valued during training.` (see issue #1186). The root cause is a scale mismatch between the derived insolation values fed to the model and the on-disk normalization stats for the DSWRFtoa slot: the stats are drawn from random N(0,1) noise that was used to fill the on-disk variable, but the actual values fed at train time are the output of `CM4Insolation`, which span ~0..1400 W/m² for the literal-constant case and signed random magnitudes for the load-from-disk case. With AMP enabled and Adam at lr=1e-3, the resulting out-of-distribution inputs are occasionally large enough to drive gradients to NaN on GPU.

This PR forces the solar constant to 1.0 inside the test via a small context-managed monkey-patch of `CM4Insolation.__call__`, which bounds the derived insolation to roughly [0, 1] and brings it in line with the normalizer's expected scale. It also lowers the test default learning rate from 1e-3 to 1e-4 to better mirror production training settings.

Changes:
- `fme.ace.testing.patch_cm4_solar_constant`: new context manager that temporarily replaces the solar constant passed to `CM4Insolation` with a fixed value (defaults to 1.0), reverting on exit.
- `fme.ace.test_train.test_train_and_inference_with_derived_forcings`: wrap the `train_main` / `inference_evaluator_main` calls in `patch_cm4_solar_constant(1.0)` so the derived insolation matches the synthetic-stats scale.
- `fme.ace.test_train._get_test_yaml_files`: drop the test optimizer learning rate from 1e-3 to 1e-4.

- [x] Tests added

Resolves #1186

### DIFF
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -51,6 +51,7 @@ from fme.ace.stepper.time_length_probabilities import (
 from fme.ace.testing import (
     DimSizes,
     MonthlyReferenceData,
+    patch_cm4_solar_constant,
     save_nd_netcdf,
     save_scalar_netcdf,
     save_stepper_checkpoint,
@@ -330,7 +331,7 @@ def _get_test_yaml_files(
             use_gradient_accumulation=True,
             enable_automatic_mixed_precision=True,
             optimizer_type="Adam",
-            lr=0.001,
+            lr=0.0001,
             kwargs=dict(weight_decay=0.01),
             scheduler=SchedulerConfig(
                 type="CosineAnnealingLR",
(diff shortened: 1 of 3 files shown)


brianhenn (approved):
LGTM thanks!

PR #1188, opened 2026-05-20
Revert back to torch>=2.4

elynnwu:
Since we don't actually use any new features that would require torch>2.4, we revert back to torch>=2.4.


brianhenn (approved):
LGTM -- sorry for causing the digression here :sweat_smile:

PR #1191, opened 2026-05-21
Add normalizer option to remove global mean surface temperature

mcgibbon:
Atmospheric warming/cooling signals show up as large temperature anomalies across many vertical levels at once. Normalizing each temperature field by its global mean leaves a slowly-varying global temperature offset embedded in every channel, which the network has to re-learn at every layer. This PR adds an opt-in normalization mode that strips the global-mean surface temperature anomaly from all temperature fields before normalization (and restores it after denormalization), so the network sees temperature gradients rather than absolute global temperatures.

The offset is computed per sample as `surface_temperature_norm_mean - sample.surface_temperature.mean(spatial dims)` and added to every temperature field on `normalize()`, then subtracted back on `denormalize()`. Cross-channel and horizontal gradients are preserved; only the global mean shifts. The flag is meaningful only for the network normalizer (paired normalize/denormalize); the loss normalizer always strips it because offsets computed independently from target vs. prediction inputs would not cancel cleanly.

Changes:
- `fme.core.normalizer.NormalizationConfig.remove_global_mean_surface_temperature`: new opt-in bool flag (default `False`).
- `fme.core.normalizer.StandardNormalizer`: caches the per-sample offset during `normalize()` and inverts it in `denormalize()`; raises if `surface_temperature` is missing from means or input tensors, or if `denormalize()` runs before `normalize()`.
- `fme.core.normalizer.NetworkAndLossNormalizationConfig.get_loss_normalizer`: strips the flag from the loss normalizer via the new `_disable_global_mean_surface_temperature_removal` helper.
- `fme.core.normalizer._GLOBAL_MEAN_SURFACE_TEMPERATURE_REFERENCE` / `_TEMPERATURE_FIELD_NAMES`: hard-coded reference field and target field list.
- `StandardNormalizer.from_state` reads the legacy `use_shared_temperature_offset` state-dict key as a fallback so existing checkpoints continue to load.

- [x] Tests added

### DIFF
--- a/fme/core/normalizer.py
+++ b/fme/core/normalizer.py
@@ -11,6 +11,28 @@ import xarray as xr
 from fme.core.device import move_tensordict_to_device
 from fme.core.typing_ import TensorDict, TensorMapping
 
+# Experimental: when NormalizationConfig.experimental_use_shared_temperature_offset
+# is True, every field named here gets a per-sample additive offset applied
+# before normalization (and removed after denormalization). The offset for a
+# sample is the difference between the surface_temperature normalization mean
+# and the sample's cell-wise mean surface_temperature. Hard-coded for now.
+_TEMPERATURE_FIELD_NAMES = frozenset(
+    [
+        "surface_temperature",
+        "TMP2m",
+        "TMP850",
+        "air_temperature_0",
+        "air_temperature_1",
+        "air_temperature_2",
+        "air_temperature_3",
+        "air_temperature_4",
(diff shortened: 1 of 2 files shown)


mcgibbon:
I'm still not sure the normalization is the right place for the code, as I'm wondering if I should add the global mean surface temperature back in as a separate (constant) field, and the normalizer wouldn't be the right level to do that.

mcgibbon:
Superceded by #1193.

PR #1192, opened 2026-05-26
Add channel-mean support to the ensemble aggregator

mcgibbon:
Extends the ensemble aggregator (CRPS, SSR bias, ensemble-mean RMSE) with a `target` mode and `channel_mean_names`, mirroring the RMSE-side behavior so a single channel-mean scalar is logged alongside per-variable metrics when computing on normalized data. Averaging metrics across variables is only meaningful in normalized units, so `channel_mean` is gated on `target == "norm"`. Normalized data is plumbed through the training and inference callsites.

Changes:
- `fme.ace.aggregator.one_step.ensemble._EnsembleAggregator`: added `target` and `channel_mean_names` constructor args; `record_batch` now accepts optional `target_data_norm`/`gen_data_norm` and requires them when `target="norm"`; `_get_data` emits a `channel_mean` scalar per metric when `target="norm"`.
- `fme.ace.aggregator.one_step.ensemble.SelectStepEnsembleAggregator.record_batch`: plumbs `target_data_norm`/`gen_data_norm` through to the inner aggregator; internal slicing extracted into a `_select` helper.
- `fme.ace.aggregator.one_step.ensemble.EnsembleMetricConfig` / `OneStepEnsembleMetricConfig`: added `target` field (default `"denorm"`); `EnsembleMetricConfig` auto-name now appends `_norm` when target is `"norm"`.
- `fme.ace.aggregator.one_step.build_context.EnsembleAggregator` protocol: updated to include the new optional kwargs.
- `fme.ace.aggregator.one_step.main.OneStepAggregator.record_batch` and `fme.ace.aggregator.inference.main.InferenceEvaluatorAggregator.record_batch`: compute normalized ensemble tensors and pass them to the ensemble aggregator. This now runs unconditionally whenever `n_ensemble > 1`, even when no `target="norm"` aggregator is configured.

Back-compat: existing YAMLs without a `target` field continue to produce identical names and behavior, since `target` defaults to `"denorm"` and the auto-name change only triggers for `target="norm"`.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -19,7 +19,8 @@ from fme.core.generics.aggregator import (
     InferenceLogs,
 )
 from fme.core.gridded_ops import GriddedOperations, LatLonOperations
-from fme.core.typing_ import EnsembleTensorDict, TensorDict, TensorMapping
+from fme.core.tensors import unfold_ensemble_dim
+from fme.core.typing_ import TensorDict, TensorMapping
 from fme.core.wandb import Table, WandB
 
 from ..one_step.ensemble import EnsembleMetricConfig, SelectStepEnsembleAggregator
@@ -549,11 +550,11 @@ class InferenceEvaluatorAggregator(
             unfolded_target_data, unfolded_prediction_data = (
                 data.as_ensemble_tensor_dicts(data.n_ensemble)
             )
-            unfolded_target_data_norm = EnsembleTensorDict(
-                self._normalize(unfolded_target_data)
+            unfolded_target_data_norm = unfold_ensemble_dim(
+                TensorDict(batch.target_norm), data.n_ensemble
             )
-            unfolded_prediction_data_norm = EnsembleTensorDict(
-                self._normalize(unfolded_prediction_data)
+            unfolded_prediction_data_norm = unfold_ensemble_dim(
+                TensorDict(batch.prediction_norm), data.n_ensemble
             )
             for ensemble_aggregator in self._ensemble_aggregators.values():
                 ensemble_aggregator.record_batch(
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -19,7 +19,7 @@ from fme.core.generics.aggregator import (
     InferenceLogs,
 )
 from fme.core.gridded_ops import GriddedOperations, LatLonOperations
-from fme.core.typing_ import TensorDict, TensorMapping
+from fme.core.typing_ import EnsembleTensorDict, TensorDict, TensorMapping
 from fme.core.wandb import Table, WandB
 
 from ..one_step.ensemble import EnsembleMetricConfig, SelectStepEnsembleAggregator
@@ -549,10 +549,18 @@ class InferenceEvaluatorAggregator(
             unfolded_target_data, unfolded_prediction_data = (
                 data.as_ensemble_tensor_dicts(data.n_ensemble)
             )
+            unfolded_target_data_norm = EnsembleTensorDict(
+                self._normalize(unfolded_target_data)
+            )
+            unfolded_prediction_data_norm = EnsembleTensorDict(
+                self._normalize(unfolded_prediction_data)
+            )
             for ensemble_aggregator in self._ensemble_aggregators.values():
                 ensemble_aggregator.record_batch(
                     target_data=unfolded_target_data,
                     gen_data=unfolded_prediction_data,
+                    target_data_norm=unfolded_target_data_norm,
+                    gen_data_norm=unfolded_prediction_data_norm,
                     i_time_start=self._n_timesteps_seen,
                 )
         n_times = data.time.shape[1]
--- a/fme/ace/aggregator/one_step/main.py
+++ b/fme/ace/aggregator/one_step/main.py
@@ -18,7 +18,7 @@ from fme.ace.stepper import TrainOutput
 from fme.core.dataset_info import DatasetInfo
 from fme.core.generics.aggregator import AggregatorABC
 from fme.core.tensors import fold_ensemble_dim, fold_sized_ensemble_dim
-from fme.core.typing_ import TensorMapping
+from fme.core.typing_ import EnsembleTensorDict, TensorMapping
 
 from .build_context import (
     Aggregator,
@@ -52,10 +52,12 @@ class OneStepAggregator(AggregatorABC[TrainOutput]):
     def __init__(
         self,
         deterministic_aggregator: OneStepDeterministicAggregator,
-        ensemble_aggregator: EnsembleAggregator | None = None,
+        ensemble_aggregators: dict[str, EnsembleAggregator] | None = None,
     ):
         self._deterministic_aggregator = deterministic_aggregator
-        self._ensemble_aggregator = ensemble_aggregator
+        self._ensemble_aggregators: dict[str, EnsembleAggregator] = (
+            ensemble_aggregators or {}
+        )
         self._ensemble_recorded = False
         self._per_step_losses = PerStepLossAggregator()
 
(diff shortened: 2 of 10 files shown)


mcgibbon:
Pre-review agent: notes for the reviewer (not intended for the squash commit message).

**Sibling-consistency questions** — would appreciate a call on each:

1. **Unknown-name handling in `channel_mean_names` differs across three sibling implementations.**
   - This PR (`one_step/ensemble.py:279`): silently drops unknown names, denominator excludes them.
   - `inference/time_mean.py:368-370`: raises `KeyError` on unknown names.
   - `one_step/reduced_metrics.py:90-93`: silently drops unknown names, denominator includes them.
   So `channel_mean_names=["a", "not_present"]` with only `a` present produces three different answers. A test in this PR locks in the new third variant. Happy to align with one of the existing siblings if there's a preferred semantic.

2. **No per-config `channel_mean_names` override on the new ensemble configs.** `StepMeanMetricConfig`, `OneStepMeanMetricConfig`, and `TimeMeanMetricConfig` all expose `channel_mean_names: list[str] | None` with a `(self.channel_mean_names or ctx.channel_mean_names)` fallback. `EnsembleMetricConfig` / `OneStepEnsembleMetricConfig` here read only from `ctx`. Out of scope for this PR, or worth adding for parity?

3. **`OneStepEnsembleMetricConfig.name` is not target-aware.** `OneStepMeanMetricConfig.__post_init__` produces `"mean"` / `"mean_norm"`; `EnsembleMetricConfig` was updated in this PR to suffix `_norm`. The one-step ensemble equivalent still always uses `"ensemble"`, so its wandb namespace doesn't distinguish targets. Multiple-ensemble configs are already rejected at build time (`one_step/main.py:174`), so it's cosmetic.

**Test-coverage gap worth flagging:**

4. Unit tests of `_EnsembleAggregator` are solid, but no integration test exercises `target="norm"` end-to-end. `test_inference_evaluator_aggregator_ensemble` (`inference/test_main.py:101`) uses only denorm, and `OneStepAggregator` has no test for `OneStepEnsembleMetricConfig(target="norm")`. Happy to add both if you'd like before merging.

**Smaller:**

5. The aggregator-level docstring states that `channel_mean_names=None` with `target="norm"` averages over all variables; the `EnsembleMetricConfig` docstring does not. Minor footgun for users who'd expect `None` to disable.

fme/ace/aggregator/inference/main.py
```diff
@@ -549,10 +549,18 @@ def record_batch(
             unfolded_target_data, unfolded_prediction_data = (
                 data.as_ensemble_tensor_dicts(data.n_ensemble)
             )
+            unfolded_target_data_norm = EnsembleTensorDict(
```
mcgibbon: Issue: This order of operations results in two calls to _normalize on what's really the same data. Is there any way to avoid this while keeping the APIs clean? What does it look like without changing APIs? Is that implementation misleading or confusing? If so, what minimal API change would avoid this?

mcgibbon:
> Pre-review agent: notes for the reviewer (not intended for the squash commit message).
> 
> **Sibling-consistency questions** — would appreciate a call on each:
> 
> 1. **Unknown-name handling in `channel_mean_names` differs across three sibling implementations.**
>    

Is it feasible to raise KeyError on all of these? Are there scenarios where we expect these to be missing?

> 2. **No per-config `channel_mean_names` override on the new ensemble configs.** `StepMeanMetricConfig`, `OneStepMeanMetricConfig`, and `TimeMeanMetricConfig` all expose `channel_mean_names: list[str] | None` with a `(self.channel_mean_names or ctx.channel_mean_names)` fallback. `EnsembleMetricConfig` / `OneStepEnsembleMetricConfig` here read only from `ctx`. Out of scope for this PR, or worth adding for parity?

Let's add it for parity.

> 3. **`OneStepEnsembleMetricConfig.name` is not target-aware.** `OneStepMeanMetricConfig.__post_init__` produces `"mean"` / `"mean_norm"`; `EnsembleMetricConfig` was updated in this PR to suffix `_norm`. The one-step ensemble equivalent still always uses `"ensemble"`, so its wandb namespace doesn't distinguish targets. Multiple-ensemble configs are already rejected at build time (`one_step/main.py:174`), so it's cosmetic.

This seems like an issue. We need to be able to configure both a norm and non-norm config in the same run. Can we match the behavior for OneStepMeanMetricConfig?

> **Test-coverage gap worth flagging:**
> 
> 4. Unit tests of `_EnsembleAggregator` are solid, but no integration test exercises `target="norm"` end-to-end. `test_inference_evaluator_aggregator_ensemble` (`inference/test_main.py:101`) uses only denorm, and `OneStepAggregator` has no test for `OneStepEnsembleMetricConfig(target="norm")`. Happy to add both if you'd like before merging.

We should add a test, in tandem with adding support for both norm and denorm to exist at the same time.

> **Smaller:**
> 
> 5. The aggregator-level docstring states that `channel_mean_names=None` with `target="norm"` averages over all variables; the `EnsembleMetricConfig` docstring does not. Minor footgun for users who'd expect `None` to disable.

Update any docstrings as needed for clarity.

mcgibbon:
Pre-review agent: addressed the review feedback. Summary of changes since the previous push (`28031517`):

1. **KeyError on unknown `channel_mean_names` everywhere.** Aligned all three sibling implementations to raise `KeyError` with a friendly message listing the missing names and the available ones:
   - `fme/ace/aggregator/one_step/ensemble.py` (new code in this PR).
   - `fme/ace/aggregator/one_step/reduced_metrics.py` (`AreaWeightedReducedMetric.record`).
   - `fme/ace/aggregator/inference/time_mean.py`.

2. **Per-config `channel_mean_names` on both ensemble configs.** Added `channel_mean_names: list[str] | None = None` to `EnsembleMetricConfig` and `OneStepEnsembleMetricConfig`, with the standard `(self.channel_mean_names or ctx.channel_mean_names)` fallback, matching the RMSE-side siblings.

3. **Target-aware naming + both targets coexist in one run.**
   - `OneStepEnsembleMetricConfig` now defaults its name to `\"ensemble\"` for `target=\"denorm\"` and `\"ensemble_norm\"` for `target=\"norm\"`, mirroring `OneStepMeanMetricConfig`.
   - `OneStepAggregator` now takes `ensemble_aggregators: dict[str, EnsembleAggregator]` and runs every entry; the \"Multiple ensemble metrics are not supported\" guard is removed.
   - `OneStepAggregatorConfig` now exposes `ensemble_denorm` + `ensemble_norm` fields mirroring `mean_denorm` / `mean_norm` (the previous `ensemble` field is renamed). `ensemble_norm` defaults to `enabled=False` to keep current wandb output for unmodified configs.
   - **Wandb-key namespace change**: ensemble keys are now prefixed by the aggregator name, e.g. `metrics/ensemble/crps/a` instead of `metrics/crps/a`, and `metrics/ensemble_norm/crps/a` when the norm side is enabled. This is consistent with how `mean` / `mean_norm` and `time_mean` / `time_mean_norm` already namespace their logs — dashboards that watch the old un-prefixed keys will need to be updated.

4. **Integration tests for `target=\"norm\"` end-to-end.**
   - `test_both_norm_and_denorm_ensemble_metrics_coexist` in `one_step/test_main.py`: configures both aggregators in one run, exercises the normalize callback, and checks per-variable + `channel_mean` log keys land in the right namespaces. Includes a scale-sensitive differential assertion to ensure the two sides are actually computing on different inputs.
   - `test_inference_evaluator_aggregator_ensemble` in `inference/test_main.py`: extended to use both denorm and norm `EnsembleMetricConfig` and assert per-namespace logs + differential.

5. **Docstrings.** Added class-level docstrings on `EnsembleMetricConfig` and `OneStepEnsembleMetricConfig` documenting `target`, the `channel_mean_names`-fallback behavior, and the \"None means all variables present\" default. Tightened the aggregator-level docstring to say names not in the data raise `KeyError`.

6. **Failing regression test.** Regenerated `docs/default-aggregator-config.yaml` to reflect the new `target` and `channel_mean_names` fields on `EnsembleMetricConfig`; `docs/test_default_aggregator_config.py` now passes.

Test scope run locally: `fme/ace/aggregator/` (210 passed), `docs/` (2 passed), `fme/ace/test_train.py` (18 passed), `pre-commit` clean.

fme/ace/aggregator/one_step/main.py
```diff
@@ -276,7 +301,9 @@ def build(
             loss_scaling=loss_scaling,
             channel_mean_names=channel_mean_names,
             raise_on_unsupported=False,
-            include_default_ensemble=self.ensemble.enabled,
+            include_default_ensemble=(
```
Arcomano1234: Question: Does this lead to somewhat unexpected behavior where if only `ensemble_norm.enabled=True` this also creates denorm aggregator metrics as well?

fme/ace/aggregator/inference/main.py
Arcomano1234: Yeah it would be nice not to pay this double computational penalty

Arcomano1234 (approved):
It is somewhat unfortunate that old runs with the 1 step and 20 step ensemble metrics won't be comparable as the wandb logging is going from `{label}/metric/{var}` to `{label}/ensemble/metric/{var} `but I am not sure we will ever compare to old runs. Other than that I left few questions / comments but this looks ready to go.

fme/ace/aggregator/one_step/main.py
```diff
@@ -276,7 +301,7 @@ def build(
             loss_scaling=loss_scaling,
             channel_mean_names=channel_mean_names,
             raise_on_unsupported=False,
-            include_default_ensemble=self.ensemble.enabled,
+            include_default_ensemble=False,
```
mcgibbon, line 304: We don't need this pathway for the non-legacy config - it's included in self._get_metrics() by default already.

fme/ace/aggregator/one_step/main.py
mcgibbon: Disabled this, it's already handled by the metrics= arg.

PR #1193, opened 2026-05-26
Add configurable global mean removal transform

mcgibbon:
Add an optional `global_mean_removal` config to `SingleModuleStepConfig` (and `SingleModuleStepperConfig`) that removes per-sample global means from fields before normalization and restores them after denormalization. This lets the network operate on anomalies relative to the current global mean, which can improve generalization for temperature-like fields under climate drift.

Changes:
- `fme.core.step.global_mean_removal`: new module with `SharedGlobalMeanRemoval` (single reference field offset applied to a set of fields) and `PerChannelGlobalMeanRemoval` (each field's own mean removed independently). Both support optionally appending the removed mean as extra normalized input channels.
- `fme.core.step.SingleModuleStepConfig`: new optional `global_mean_removal` field; forward transform applied before normalization, inverse transform applied after denormalization
- `fme.ace.stepper.SingleModuleStepperConfig`: passes `global_mean_removal` through to `SingleModuleStepConfig`

- [x] Tests added

### DIFF
--- a/fme/core/step/global_mean_removal.py
+++ b/fme/core/step/global_mean_removal.py
@@ -73,6 +73,27 @@ class GlobalMeanRemoval(abc.ABC):
         """
 
 
+class NoGlobalMeanRemoval(GlobalMeanRemoval):
+    """No-op implementation used when global mean removal is disabled."""
+
+    @property
+    def n_extra_input_channels(self) -> int:
+        return 0
+
+    def forward_transform(
+        self,
+        input: TensorMapping,
+        data_mask: TensorMapping | None,
+    ) -> TensorDict:
+        return dict(input)
+
+    def inverse_transform(self, output: TensorDict) -> TensorDict:
+        return output
+
+    def get_extra_channels(self) -> torch.Tensor | None:
+        return None
+
+
 class SharedGlobalMeanRemoval(GlobalMeanRemoval):
     """Remove a single reference field's global mean from a set of fields.
 
--- a/fme/core/step/global_mean_removal.py
+++ b/fme/core/step/global_mean_removal.py
@@ -180,7 +180,15 @@ class SharedGlobalMeanRemoval(GlobalMeanRemoval):
 
 
 class PerChannelGlobalMeanRemoval(GlobalMeanRemoval):
-    """Remove each field's own per-sample global mean."""
+    """Shift each field's per-sample spatial mean to its climatology.
+
+    For each field, ``forward_transform`` adds ``clim_mean - sample_mean``
+    so the field's spatial mean is equal to its climatology mean in
+    physical space; after normalization the field's spatial mean is
+    approximately zero.  This mirrors ``SharedGlobalMeanRemoval`` but
+    computes the offset per field rather than from a single reference
+    field.
+    """
 
     def __init__(
         self,
@@ -193,7 +201,7 @@ class PerChannelGlobalMeanRemoval(GlobalMeanRemoval):
         self._append_as_input = append_as_input
         self._means = means
         self._stds = stds
-        self._cached_sample_means: dict[str, torch.Tensor] | None = None
+        self._cached_shifts: dict[str, torch.Tensor] | None = None
         self._cached_extra: torch.Tensor | None = None
 
     @property
@@ -206,7 +214,7 @@ class PerChannelGlobalMeanRemoval(GlobalMeanRemoval):
         data_mask: TensorMapping | None,
     ) -> TensorDict:
         result = dict(input)
-        sample_means: dict[str, torch.Tensor] = {}
+        shifts: dict[str, torch.Tensor] = {}
         spatial_shape: tuple[int, ...] | None = None
 
         for name in self._field_names:
@@ -215,23 +223,24 @@ class PerChannelGlobalMeanRemoval(GlobalMeanRemoval):
             t = result[name]
             if spatial_shape is None:
                 spatial_shape = t.shape[1:]
-            raw_mean = t.mean(dim=tuple(range(1, t.ndim)))
+            sample_mean = t.mean(dim=tuple(range(1, t.ndim)))
+            shift = self._means[name] - sample_mean
             if data_mask is not None and name in data_mask:
                 mask = data_mask[name]
-                raw_mean = torch.where(mask, raw_mean, torch.zeros_like(raw_mean))
-            sample_means[name] = raw_mean
-            broadcast = raw_mean.view(raw_mean.shape[0], *(1,) * (t.ndim - 1))
-            result[name] = t - broadcast
+                shift = torch.where(mask, shift, torch.zeros_like(shift))
+            shifts[name] = shift
+            broadcast = shift.view(shift.shape[0], *(1,) * (t.ndim - 1))
+            result[name] = t + broadcast
 
-        self._cached_sample_means = sample_means
+        self._cached_shifts = shifts
 
         if self._append_as_input and spatial_shape is not None:
             channels: list[torch.Tensor] = []
             for name in self._field_names:
-                if name not in sample_means:
+                if name not in shifts:
                     continue
-                sm = sample_means[name]
-                normalized = (sm - self._means[name]) / self._stds[name]
+                sh = shifts[name]
+                normalized = -sh / self._stds[name]
                 ch = normalized.view(-1, 1, *(1,) * len(spatial_shape)).expand(
                     -1, 1, *spatial_shape
                 )
@@ -243,16 +252,16 @@ class PerChannelGlobalMeanRemoval(GlobalMeanRemoval):
         return result
 
     def inverse_transform(self, output: TensorDict) -> TensorDict:
-        if self._cached_sample_means is None:
+        if self._cached_shifts is None:
             raise RuntimeError("inverse_transform() called before forward_transform().")
         result = dict(output)
         for name in self._field_names:
-            if name not in result or name not in self._cached_sample_means:
+            if name not in result or name not in self._cached_shifts:
                 continue
             t = result[name]
-            sm = self._cached_sample_means[name]
-            broadcast = sm.view(sm.shape[0], *(1,) * (t.ndim - 1))
-            result[name] = t + broadcast
+            sh = self._cached_shifts[name]
+            broadcast = sh.view(sh.shape[0], *(1,) * (t.ndim - 1))
+            result[name] = t - broadcast
         return result
 
     def get_extra_channels(self) -> torch.Tensor | None:
@@ -320,18 +329,22 @@ class SharedGlobalMeanRemovalConfig:
 
 @dataclasses.dataclass
 class PerChannelGlobalMeanRemovalConfig:
-    """Remove each field's own per-sample global mean.
+    """Shift each field's per-sample spatial mean to its climatology.
 
-    Unlike ``SharedGlobalMeanRemovalConfig``, per-channel removal
-    requires each field to be present in the input so its individual
-    mean can be computed.
+    For each listed field, ``forward_transform`` adds
+    ``clim_mean - sample_mean`` so that after normalization the field's
+    spatial mean is approximately zero (avoiding large constant biases
+    on the network input).  Unlike ``SharedGlobalMeanRemovalConfig``,
+    per-channel removal requires each field to be present in the input
+    so its individual sample mean can be computed.
 
     Parameters:
         kind: Must be ``"per_channel"``.
         field_names: Names of fields to process. ``None`` means all
             input fields.  Explicit names must be input fields.
-        append_as_input: If true, each field's removed sample mean
-            (normalized) is appended as an extra network input channel.
+        append_as_input: If true, each field's normalized sample-mean
+            anomaly ``(sample_mean - clim_mean) / clim_std`` is appended
+            as an extra network input channel.
     """
 
     kind: Literal["per_channel"] = "per_channel"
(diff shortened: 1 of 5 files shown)


fme/core/step/global_mean_removal.py
```diff
@@ -0,0 +1,373 @@
...
+            broadcast = sm.view(sm.shape[0], *(1,) * (t.ndim - 1))
+            result[name] = t + broadcast
+        return result
+
+    def get_extra_channels(self) -> torch.Tensor | None:
+        return self._cached_extra
+
+
+@dataclasses.dataclass
+class GlobalMeanRemovalConfig(abc.ABC):
```
mcgibbon: Is this ABC ever used directly in a type hint, or only inherited? If it's not used as a target, the ABC is not necessary (it's considered "unused") and should be removed.

fme/core/step/single_module.py
```diff
@@ -384,8 +416,8 @@ def network_call(input_norm: TensorDict) -> TensorDict:
             output_dict.update(secondary_output_dict)
             return output_dict
 
-        return step_with_adjustments(
-            input=args.input,
+        output = step_with_adjustments(
+            input=step_input,
```
mcgibbon: Question: How does this global mean removal interact with the corrector? It's crucial that the corrector sees the inverse transformed inputs and outputs, i.e. with the global mean removal reverted.

fme/core/step/global_mean_removal.py
mcgibbon: Removed — it was only inherited, never used as a type target. The union type `GlobalMeanRemovalConfigUnion` covers the polymorphic case. The two concrete config dataclasses now stand alone. (The `GlobalMeanRemoval` ABC is still used as a type hint in `single_module.py` so that one stays.)

fme/core/step/single_module.py
mcgibbon: Good catch — the corrector (and ocean/prescribed prognostics) were seeing mean-removed values. Fixed: when `global_mean_removal` is active, `step_with_adjustments` now runs with `corrector=None, ocean=None` and no prescribed prognostics. After `inverse_transform` restores physical values, the corrector, ocean, and prescribed prognostics are applied using the original `args.input`.

mcgibbon: That's not really much of a solution - step_with_adjustments is specifically supposed to apply adjustments. Can we apply the mean removal inside step_with_adjustments instead?

mcgibbon:
The way this is implemented raises some questions about interaction with masked-input training. The way masking is implemented right now, the step requires surface temperature always be provided when this feature is on, so that it can compute the global-mean and use it to normalize the other features.

I think it's also fairly clear we would want to be able to train while independently masking e.g. surface temperature and this new global-mean surface temperature input.

On the question of information leakage, we could later add a feature to noise this removed global-mean so it doesn't contain such reliable information about the global mean surface temperature.

We have two types of masking for inputs - missing data, and "we want to train this batch without this input". For missing data, the current behavior is basically correct - we can't train on a sample without surface temperature, unless we modify the scheme for the temperature removal to be less dependent on that field. For the "we want to train this batch without this input", we could maybe think about this as a dropout instead?

fme/core/step/test_global_mean_removal.py
```diff
@@ -0,0 +1,444 @@
...
+    transform = _make_per_channel(means, stds)
+    tensors = move_tensordict_to_device(
+        {"a": torch.tensor([[[10.0, 20.0]], [[100.0, 200.0]]])}
+    )
+    result = transform.forward_transform(tensors, None)
+    # clim_mean=0, so shift per sample is -sample_mean
+    # sample 0: mean=15 → shift=-15 → [10-15, 20-15] = [-5, 5]
+    # sample 1: mean=150 → shift=-150 → [100-150, 200-150] = [-50, 50]
+    torch.testing.assert_close(
+        result["a"].cpu(), torch.tensor([[[-5.0, 5.0]], [[-50.0, 50.0]]])
```
yyexela, line 217: Nitpick: Maybe we should remove non-standard characters like → from the codebase

fme/core/step/test_step.py
```diff
@@ -1243,3 +1247,155 @@ def test_step_with_include_channel_mask_inputs_no_data_mask():
...
+    n_samples = 2
+    input_data = get_tensor_dict(step.input_names, DEFAULT_IMG_SHAPE, n_samples)
+    next_step = get_tensor_dict(
+        step.next_step_input_names, DEFAULT_IMG_SHAPE, n_samples
+    )
+    output = step.step(
+        args=StepArgs(input=input_data, next_step_input_data=next_step, labels=None),
+    )
+    for name in out_names:
+        assert output[name].shape == (n_samples, *DEFAULT_IMG_SHAPE)
```
yyexela, line 1310: Question: Should this test also check the output values? I can imagine that it would be useful to do so.

yyexela: Same applies for `test_step_per_channel_global_mean_removal`

yyexela (commented):
Overall looks great! Just some minor comments.

fme/core/step/test_global_mean_removal.py
mcgibbon: If there is a pre-commit or ruff rule to enforce this easily we should do it, unless we're already using a lot of them. Otherwise I'd leave it, it's not worth enforcing by review.

fme/core/step/test_step.py
mcgibbon: What should we test about the output values? If there's something good to cover we should add the coverage. Otherwise Claude should explain why values are already sufficiently covered, and why there is coverage linking that unit coverage to the full step (e.g. that it's already tested that the code properly gets hooked into the step and called in a way that affects output).

fme/core/step/single_module.py
mcgibbon: Claude: Done — moved into `step_with_adjustments` itself in 2b98d5b2e. It now takes a `global_mean_removal` parameter (single_module.py:511) and applies `forward_transform` before normalize and `inverse_transform` after denormalize, so the corrector, ocean, and prescribed-prognostic steps all see physical-space values. The earlier conditional-skip approach has been removed.

fme/core/step/test_global_mean_removal.py
mcgibbon: Claude: Checked the tooling angle and the cost/benefit:

- Ruff's `RUF001/002/003` only target *ambiguous* unicode — chars that look like ASCII (Cyrillic `а` vs ASCII `a`, Greek `;` vs ASCII `;`, smart quotes). That's a real correctness/security concern. `→` isn't on that list because it has no ASCII look-alike, so enabling those rules wouldn't catch it.
- The category `→` falls into is purely stylistic. It's unambiguous, renders fine in any UTF-8-aware tool, and doesn't affect program behavior. The usual arguments against it (grep-ability, encoding portability, AI-tell) are pretty weak in practice.
- The codebase already has ~30 occurrences across docstrings/comments without incident, so a one-line ban here would be inconsistent with the rest of the repo.

A local `pygrep-hooks` rule could enforce it cheaply if we want, but given the cleanup cost and the absence of a concrete problem these characters have caused, I'd skip it. Leaving as-is in this PR.

fme/core/step/test_step.py
mcgibbon: Claude: Added `test_step_global_mean_removal_affects_output` which runs the step with and without `global_mean_removal` under the same seed/inputs and asserts the outputs differ — that's the missing link proving the transform is actually invoked during the step.

Value behavior of the transforms themselves (forward/inverse round-trip, per-sample correctness, extra-channel values, climatology offset) is already covered in `test_global_mean_removal.py` — see `test_per_channel_is_per_sample`, `test_per_channel_extra_channels`, `test_shared_inverse_restores_input`, `test_per_channel_clim_offset`, etc. The `_with_extra_channels` step tests also implicitly verify channel-count plumbing (SFNO would error if the extras weren't concatenated). What was genuinely missing was a step-level assertion that the transform isn't silently a no-op when configured, which the new test now provides.

yyexela: Don't see a new commit with this change

yyexela (commented):
Issue: The new commit adds a test called `test_step_global_mean_removal_affects_output` that tests if the outputs of a full step with/without `PerChannelGlobalMeanRemovalConfig` differ. This test does not verify that `SharedGlobalMeanRemovalConfig` also produces different outputs.

Suggestion: Copy this test into two:

`test_step_per_channel_global_mean_removal_affects_output`
and
`test_step_shared_global_mean_removal_affects_output`

This then verifies that the transform is invoked during the step for both config types.

fme/core/step/test_step.py
yyexela: New commit added, added new comment

yyexela:
Question: Should `NoGlobalMeanRemoval` be tested? I think it's worth verifying that when no global mean removal is selected, forcing the use of `NoGlobalMeanRemoval`, the functionality doesn't change. Though I guess one could argue that since the rest of the unit-tests were not affected by using `NoGlobalMeanRemoval` then it's implicitly tested.

mcgibbon:
> Though I guess one could argue that since the rest of the unit-tests were not affected by using NoGlobalMeanRemoval then it's implicitly tested.

Yes, I would say this is already covered by the backwards-compatibility inference tests.

fme/core/step/test_step.py
mcgibbon: Claude: Good point. Refactored into a shared helper `_assert_global_mean_removal_affects_output` and split the test into `test_step_per_channel_global_mean_removal_affects_output` and `test_step_shared_global_mean_removal_affects_output` so both config types get symmetric coverage. Pushed in 8f1c77bf4.

yyexela:
Changes look good to me! ✅ I approve

fme/core/step/global_mean_removal.py
```diff
@@ -0,0 +1,389 @@
...
+                f"Reference field '{ref_name}' is not present in the input."
+            )
+        if data_mask is not None and ref_name in data_mask:
+            if not data_mask[ref_name].all():
+                raise ValueError(
+                    f"Reference field '{ref_name}' is masked for some samples, "
+                    "which is not supported for shared global mean removal."
+                )
+        ref = input[ref_name]
+        sample_mean = ref.mean(dim=tuple(range(1, ref.ndim)))
```
Arcomano1234, line 153: Question: When we do the stats calculation for variables we use the area averaged mean correct? I was curious if you think this makes a difference / should we use aree-weighted just to match everything else we do

fme/core/step/global_mean_removal.py
```diff
@@ -0,0 +1,389 @@
...
+    """
+
+    kind: Literal["shared"] = "shared"
+    reference_field: str = "surface_temperature"
+    field_names: list[str] = dataclasses.field(
+        default_factory=lambda: list(DEFAULT_TEMPERATURE_FIELD_NAMES)
+    )
+    append_as_input: bool = False
+
+    def get_n_extra_input_channels(self, in_names: list[str]) -> int:
```
Arcomano1234: Nit: `in_names: list[str]` is an unused argument to this function I suggest to remove it.

Arcomano1234 (commented):
Left a question that should be addressed / answered and a nit that I think should also be fixed

fme/core/step/global_mean_removal.py
mcgibbon: I did think about doing it that way, but I don't think it will make a difference or be worth the added complexity.

Arcomano1234: Yeah I don't think it should have profound effect but I think we should document it somewhere as I think most if not all of our "global means" in ACE are area-weighted. So if in the doc strings we say something like "subtract the raw global mean" that should be good enough

fme/core/step/global_mean_removal.py
mcgibbon: Claude: Good catch — and going further, this method was only ever called from tests; production code reads `n_extra_input_channels` from the built transform object. Removed it from both configs and updated the one test (`test_per_channel_config_none_field_names_means_all`) to go through the build path. Pushed in e296a2d96.

fme/core/step/global_mean_removal.py
mcgibbon: How about "cellwise" global mean?

mcgibbon: I think our convention is generally to call the area-weighted global mean the area-weighted or weighted global mean, and I think cellwise global mean would be more verbose than needed in the variable/configuration names, but I think we could use it in the class names and docstrings.

Arcomano1234: Yeah we don't need to be that verbose for the variable names / function names just want to make sure its explicitly documented in the docstrings so we have a reference and so its obvious to external users.

Arcomano1234 (approved):
I still think we should document it some where that we are using "cellwise" means for the subtraction but other than that this looks good to go. That documentation should be added but I don't need to re-review this, so I am approving the PR.

fme/core/step/global_mean_removal.py
mcgibbon: Claude: Added cellwise-vs-area-weighted note in docstrings on `GlobalMeanRemoval` (the primary explanation, with rationale: simpler, network compensates during end-to-end training), and propagated short references to the two concrete impls (`SharedGlobalMeanRemoval`, `PerChannelGlobalMeanRemoval`) and both Config dataclasses so the distinction is obvious to external users at the user-facing entry points. Pushed in 786b07460.

PR #1195, opened 2026-05-26
Run multiple inline inference configs concurrently through one batched forward pass

mcgibbon:
> [!IMPORTANT]
> **Superseded by a reviewable 3-PR chain** (rebased onto current `main`,
> which has since diverged from this branch):
> - #1290 — additive sample-dim `cat`/`split` helpers (+ `stepper_state` propagation)
> - #1291 — behavior-preserving inference-callback refactor
> - #1292 — concurrent inline inference + a concurrent-vs-sequential equivalence test
>
> The chain reproduces this PR's content reconciled with current `main` (the
> `InferenceSummary.loss` and LR-tuning-epoch refactors landed on `main` after
> this PR was opened) and adds the equivalence test this PR lacked. Review the
> chain instead; this PR can be closed once it merges.

---

Multiple inline inference configurations currently run sequentially in the training callback, which is the dominant cost when several inferences are configured per training run. This PR routes inferences that share a compatible per-window step count through a single batched forward pass per window, so adding inferences scales like a wider batch instead of like more epochs. It also deduplicates the two near-identical `get_inference_callback` implementations between ACE and coupled training so the new dispatch path applies to both.

The coordination is a small lazy layer in front of the stepper: `BatchedPredictor.submit(ic, forcing)` collects pending calls and returns a `PredictionPromise`; the first `promise.resolve()` partitions the pending calls by forcing-window shape, runs one `stepper.predict_paired` per distinct shape on the concatenated batch, and splits each result back into per-caller slices. A `LazyLooper` wraps each per-task loader so the runner can submit a window from every active task and then trigger one flush, dropping exhausted tasks from subsequent rounds for heterogeneous run lengths.

Changes:
- `fme.core.labels.BatchLabels.cat` / `BatchLabels.semantically_equal` — new sample-dim concat that takes the sorted union of names (deduped) and conforms each input via `conform_to_encoding`, plus a semantic equality helper that compares after conforming both sides to a common encoding. This is what lets two inferences with different conditional-label sets share a batched forward pass.
- `fme.ace.data_loading.batch_data.BatchData.cat` / `BatchData.split` (and matching helpers on `PrognosticState`, `PairedData`) — sample-dim concat and inverse split with validation of `n_timesteps`, `n_ensemble`, variable names, and labels/data_mask presence. Mismatched label encodings across inputs are now supported: labels are unioned via `BatchLabels.cat` so the result carries every label seen in any input.
- `fme.coupled.data_loading.batch_data` — `CoupledBatchData`, `CoupledPrognosticState`, `CoupledPairedData` `cat` / `split` that delegate to their ocean/atmosphere components.
- `fme.ace.stepper.single_module` — the two `ic.labels != forcing.labels` sanity guards now use semantic equality, since the model consumes `forcing.labels` exclusively. This lets the new state (which may carry the unioned encoding from a prior concurrent round) coexist with a per-entry forcing encoding.
- `fme.core.generics.inference.BatchedPredictor`, `PredictionPromise`, `LazyLooper`, `ConcurrentInferenceEntry`, `run_concurrent_inference` — generic coordinator parameterized by concat/split callables, with the lock-step submit/resolve driver. A `batch_key_of_forcing` callable partitions each flush so only windows of the same shape are concatenated: the common all-equal-shape case runs a single forward pass, but when grouped runs of different lengths reach a short final window in the same round another run still has a full window, those windows go through separate passes instead of failing `cat` on mismatched `n_timesteps`. The keys are the forcing window's `n_timesteps` (ACE) and the ocean+atmosphere `n_timesteps` (coupled).
- `fme.core.generics.trainer.InferenceTask`, `build_inference_callback` — shared inline-inference callback that groups tasks by `concurrent_group`, sends each group through the shared predictor when one is supplied, and falls back to the existing sequential `inference_one_epoch` path otherwise.
- `fme.ace.train.train.get_inference_callback` and `fme.coupled.train.train.get_inference_callback` — rewritten to build `InferenceTask`s with per-trainer aggregator-factory closures and an appropriate `BatchedPredictor` factory. ACE groups on `(forward_steps_in_memory, n_ensemble_per_ic)`; coupled groups on `coupled_steps_in_memory`. Label encodings are no longer part of the grouping key because cat unions them. Runs of different total length can share a group: the predictor's per-window-shape partitioning keeps the short final window of a shorter run from colliding with a full window of a longer one.

Behavior change worth flagging: coupled training now raises (matching ACE) when a `weight>0` inference entry's logs lack `time_mean_norm/rmse/channel_mean`, instead of silently dropping that entry from the weighted-error sum. The corresponding coupled test was updated to assert the new behavior.

- [x] Tests added


mcgibbon:
Merged the window-shape partitioning fix (#1279) into this branch — it stops grouped inline-inference runs of different lengths from crashing on `Cannot cat BatchData with different n_timesteps` when a shorter run hits its short final window in the same round a longer run still has a full window.

Validating it on a real training run: rs1 (the concurrent-inference side of a speed A/B), which previously crashed before epoch 1 on this exact error, is relaunched from this branch's merge at `59ed0bf3c` on the `experiment/2026-06-05-aimip-like` branch → [beaker 01KV6P5M](https://beaker.org/ex/01KV6P5MG100PTXNV436HD40AY) (`train-4deg-daily-v1-era5-only-rs1-59ed`).

mcgibbon:
Split into a reviewable stacked chain against current `main`: #1290 (cat/split helpers) → #1291 (callback refactor) → #1292 (concurrent inference + equivalence test). The chain reconciles this branch with `main`'s newer `InferenceSummary.loss` / LR-tuning-epoch refactors and adds the concurrent-vs-sequential equivalence test. Recommend reviewing/merging the chain and closing this PR once it lands.

PR #1196, opened 2026-05-26
[WIP] Feature/var masking train stepper

yyexela:
Working on masking to train stepper

-----

Changes:

- [ ] Tests added

### DIFF
--- a/fme/core/step/single_module.py
+++ b/fme/core/step/single_module.py
@@ -362,12 +362,15 @@ class SingleModuleStep(StepABC):
         """
 
         def network_call(input_norm: TensorDict) -> TensorDict:
-            if args.data_mask is not None:
-                input_norm = _apply_input_mask(input_norm, args.data_mask)
+            effective_mask = (
+                args.channel_mask if args.channel_mask is not None else args.data_mask
+            )
+            if effective_mask is not None:
+                input_norm = _apply_input_mask(input_norm, effective_mask)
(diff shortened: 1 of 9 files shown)


fme/core/step/single_module.py
```diff
@@ -514,10 +522,20 @@ def step_with_adjustments(
     if residual_prediction:
         output_norm = add_names(input_norm, output_norm, prognostic_names)
     output = normalizer.denormalize(output_norm)
+    # Build a NaN-free view of input for the corrector and ocean model.
+    # When variable masking augmentation fills IC channels with NaN, those NaNs
+    # survive normalization's fill (which only applies to input_norm) and would
+    # propagate through area-weighted means in physics corrections, producing NaN
+    # outputs and zero gradients.  Replacing with the normalizer's denormalized
+    # estimate (climatological mean for masked variables) keeps corrections valid.
```
mcgibbon, line 530: I disagree with this take - the corrector is not valid if we're giving it climatological means instead of actual values. Like, we can't apply a corrector if the values it needs to do the correction are missing. This avoids a NaN or crash but I think a crash is the right call in this scenario.

If input data is missing for the correction, the way the code is right now I think it should raise an exception and halt. Perhaps we could add configuration to only apply certain corrections if values are available, and make the corrector mask-aware.

However for just masking inputs randomly for the network, we could consider making use of the input timestep data despite that input being masked from the network itself. This could work in a dropout-style approach for masking.

PR #1199, opened 2026-05-27
Fix channel_mean broken by variable filtering in step_means metrics

mcgibbon:
`StepMeanMetricConfig` wrapped its aggregator with `VariableFilterAdapter`, which filtered variables out of `record_batch` input. This starved the inner `MeanAggregator` of variables needed for `channel_mean` computation, raising `KeyError` when `channel_mean_names` referenced variables not in the filtered subset (e.g. when inference `step_means` configs specify a `variables` list that is a subset of the loss variable names).

Replace input-level filtering with output-level filtering: add a `report_variables` parameter to `MeanAggregator` and `_EnsembleAggregator` that computes metrics over all variables but only includes the requested subset in `get_logs`/`get_dataset` output. Aggregate entries like `channel_mean` are always included.

Changes:
- `MeanAggregator`: add `report_variables` param, filter per-variable entries in `_get_data()`
- `_EnsembleAggregator`: same pattern
- `StepMeanMetricConfig.build()`: pass `variables` as `report_variables` instead of wrapping with `maybe_filter`
- Remove unused `maybe_filter` import from `reduced.py`

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/one_step/reduced.py
+++ b/fme/ace/aggregator/one_step/reduced.py
@@ -10,11 +10,7 @@ from fme.core.distributed import Distributed
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.typing_ import TensorMapping
 
-from ..inference.build_context import (
-    MetricBuildContext,
-    MetricNotSupportedError,
-    maybe_filter,
-)
+from ..inference.build_context import MetricBuildContext, MetricNotSupportedError
 from ..inference.data import InferenceBatchData, MetricBuildResult, SubAggregator
 from .build_context import OneStepBuildContext, OneStepMetricBuildResult
 from .reduced_metrics import AreaWeightedReducedMetric, ReducedMetric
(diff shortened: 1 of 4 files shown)


fme/ace/aggregator/one_step/reduced.py
```diff
@@ -133,16 +137,25 @@ def _get_data(self):
...
         for name, metric in self._variable_metrics.items():
             metric_results = metric.get()
+            all_variable_names.update(metric_results.keys())
             for key in metric_results:
                 data[f"{name}/{key}"] = metric_results[key] / self._n_batches
             if self._target == "norm":
                 data[f"{name}/channel_mean"] = (
                     metric.get_channel_mean() / self._n_batches
                 )
+        if self._report_variables is not None:
```
Arcomano1234, line 152: Comment very optional: This block of code is duplicated in both here and `ensemble.py`. Having a help function that does this might be useful in the future if the rules for excluded are changed, that way we don't need to remember to change both files.

Arcomano1234 (approved):
Left one very optional comment regarding code duplication but other than that this is ready to go

fme/ace/aggregator/one_step/reduced.py
mcgibbon: I'll just merge, I don't expect the logic to change and if it does the other code in both objects is also going to change. Likely not worth adding the coupling/import between the files.

PR #1200, opened 2026-05-28
Coupled IC script update

elynnwu:
Found a few bugs causing the IC file to be incompatible, this PR fixes the issues.


scripts/data_process/create_coupled_ic.py
```diff
@@ -218,6 +219,78 @@ def _load_and_merge(
...
+    return None
+
+
+def _fill_nan_in_valid_mask(ds: xr.Dataset) -> xr.Dataset:
+    """Replace NaNs with 0 in cells the model treats as valid (mask == 1).
+
+    At inference the model only fills cells where mask == 0 (with 0); any NaN
+    sitting in a mask == 1 cell is left untouched and then propagates through
+    the convolutional network, turning the entire field into NaN. Sea-ice
+    variables such as UI/VI are NaN wherever there is no ice on the IC date,
```
jpdunc23: The base uncoupled dataset doesn't include the right NaNs for UI/VI, but the coupled ocean dataset does. Should we use the merge dataset approach to avoid putting the wrong version of UI/VI in the initial condition file? 

I thought this was already possible in this script by providing `coupled_ocean_zarr`, but maybe not?

jpdunc23 (commented):
Rather than recreate the processing that is done in `create_coupled_datasets`, I think we should refactor this script to avoid dependence on `create_coupled_datasets` altogether. We just need to be able to merge datasets, configured and handled in an analogous way to `MergeDatasetConfig` in `fme`, with similar priority ordering for familiar configuration.

scripts/data_process/create_coupled_ic.py
elynnwu: Good catch, updated the config instead

jpdunc23 (approved):
LGTM, thanks! I still think we should eventually refactor this script to remove dependence on create_coupled_datasets, but no need to do that now.

PR #1203, opened 2026-05-28
Refactor: extract LossSchedule from TrainStepper

mcgibbon:
Extract loss-scheduling logic from `TrainStepper` into a dedicated `LossSchedule` class, aligning the ace stepper with the abstraction pattern already used in the coupled stepper.

Changes:
- `fme.ace.stepper.loss_schedule.LossSchedule`: new class encapsulating stochastic n_forward_steps sampling, epoch-based schedule management, and the `optimize_last_step_only` policy
- `fme.ace.stepper.single_module.TrainStepper`: delegates to `LossSchedule` instead of managing sampler state inline; extracts `_accumulate_step_loss` and `_finalize_per_channel_losses` helpers; removes dead `input_data` parameter from `_accumulate_loss`

- [x] Tests added


fme/ace/stepper/test_loss_schedule.py
jpdunc23, line 1: A couple of potential testing gaps identified in agent review. Up to you whether these are worth covering:

* Epoch cache hit — no test calls `init_for_epoch` with the same epoch twice to verify the early return preserves the existing sampler.
* sample in eval mode —` test_train_eval_samplers_are_independent` calls `_eval_sampler.sample()` directly on the internal, bypassing `sample()` + `set_eval()`. No test verifies that `schedule.sample()` dispatches to the eval sampler after `set_eval()`.

jpdunc23 (approved):
LGTM. Left a comment about coverage in test_loss_schedule.py, up to you whether it is worth addressing here.

fme/ace/stepper/test_loss_schedule.py
mcgibbon: > Epoch cache hit — no test calls init_for_epoch with the same epoch twice to verify the early return preserves the existing sampler.

I'd pass on this test, we don't particularly care that the sampler is preserved only that it's set to the right value (which is already covered, even if not in this niche case).

> sample in eval mode — test_train_eval_samplers_are_independent calls _eval_sampler.sample() directly on the internal, bypassing sample() + set_eval(). No test verifies that schedule.sample() dispatches to the eval sampler after set_eval()

Let's test that.

PR #1205, opened 2026-05-28
Deduplicate inline inference callback

mcgibbon:
The two `get_inference_callback` implementations in `fme.ace.train.train` and `fme.coupled.train.train` had drifted into near-identical copies of the same per-entry loop (build aggregator, run `inference_one_epoch`, accumulate weighted error). This PR extracts the shared logic into a generic `build_inference_callback` helper in `fme.core.generics.trainer` parameterized by an `InferenceTask` dataclass, so each trainer only owns the trainer-specific aggregator construction.

Changes:
- `fme.core.generics.trainer.InferenceTask`, `fme.core.generics.trainer.build_inference_callback`: new dataclass + helper that runs each task sequentially via `inference_one_epoch` and computes the weighted checkpoint-selection error
- `fme.ace.train.train.get_inference_callback`, `fme.coupled.train.train.get_inference_callback`: now build `InferenceTask` lists and delegate to the shared helper; per-trainer aggregator construction lives in new `_make_ace_aggregator_factory` / `_make_coupled_aggregator_factory` closures
- `fme.coupled.train.train.get_inference_callback`: behavior change — now raises (matching ACE) when a `weight>0` entry's logs lack `time_mean_norm/rmse/channel_mean`, instead of silently dropping that entry from the combined error
- `fme.coupled.train.train.get_inference_callback`: secondary behavior change — when every active entry has `weight=0`, the returned checkpoint-selection error is now `None` instead of `0.0`. Previously, coupled training with all-zero weights would trigger "best inference checkpoint" saving every epoch (since `0.0 <= inf`); now it skips that, matching ACE's pre-existing behavior.
- Tests: patch targets updated to `fme.core.generics.trainer.inference_one_epoch`; coupled `test_missing_metric_returns_none_error` replaced with `test_weighted_entry_missing_metric_raises`

- [x] Tests added


mcgibbon:
Pre-review agent: walked the audit passes; PR is in good shape. Summary for the reviewer:

**Silent-failure pass** — clean. The two new silent-paths are intentional and called out in the description: weight>0 + missing metric now raises (both ACE and coupled), and all-weight-0 epochs now return `None` instead of `0.0` (coupled only — matches ACE's pre-existing behavior).

**Behavior-change pass** — ACE's `get_inference_callback` body is identical to `main` modulo the helper split (verified line-by-line). Coupled has exactly the two behavior changes listed in the description; no incidental drift.

**Differential-test pass** — added one missing test in 4424b2ea6: `test_all_zero_weight_returns_none_error` in the coupled suite. The "secondary behavior change" from the description was otherwise unasserted. Existing weighted-error tests use distinct numeric values, so reductions can't silently match.

**Sibling/mirror coverage** — both call sites of `inference_one_epoch` (ACE + coupled) are updated. No other callers.

**Config back-compat** — N/A; no loaded configs or checkpoint state touched.

**AI-agent artifact pass** — closures `_make_ace_aggregator_factory` / `_make_coupled_aggregator_factory` and inner `factory()` lack return-type annotations; `tasks: list[InferenceTask]` is unparameterized. Both are minor.

Areas worth a closer look from a human reviewer:
- `InferenceTask.weight: float = 0.0` default means a forgotten `weight=` would silently contribute nothing. Both internal callers pass it explicitly, so currently safe; flagging since the default is the "quiet" branch.

PR #1206, opened 2026-05-28
Universal sed for job runner

elynnwu:
Mac specific sed command.

### DIFF
--- a/job_runner/create_coupled_train_config.sh
+++ b/job_runner/create_coupled_train_config.sh
@@ -9,7 +9,8 @@ CONFIG_PATH=${4}
 
 beaker dataset fetch ${EXISTING_RESULTS_ATMOS_DATASET} --prefix config.yaml
 mv ./config.yaml ./atmos-config.yaml
-sed -i '' 's/statsdata/atmos_stats/g' ./atmos-config.yaml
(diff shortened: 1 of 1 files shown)


job_runner/create_coupled_train_config.sh
```diff
@@ -9,11 +9,11 @@ CONFIG_PATH=${4}
 
 beaker dataset fetch ${EXISTING_RESULTS_ATMOS_DATASET} --prefix config.yaml
 mv ./config.yaml ./atmos-config.yaml
-sed -i 's/statsdata/atmos_stats/g' ./atmos-config.yaml
+sed -i '' 's/statsdata/atmos_stats/g' ./atmos-config.yaml
```
jpdunc23: This doesn't work with GNU `sed` unfortunately.

Can you please try:

```suggestion
sed -i.bak 's/statsdata/atmos_stats/g' ./atmos-config.yaml
rm atmos-config.yaml.bak
```

jpdunc23: Another option is to use `perl`, assuming it is installed by default on macOS:

```
perl -pi -e 's/statsdata/atmos_stats/g' ./atmos-config.yaml
```

elynnwu: This worked: sed -i.bak

PR #1208, opened 2026-06-01
Add per-variable filters to power-spectrum + histogram aggregators

mcgibbon:
Adds three small per-variable filter knobs to the inference aggregators so callers can keep cheap scalar metrics cohort-wide while restricting the expensive per-variable plot/percentile output to a small reference set. Motivated by the CMIP6 daily training-run plan whose 11 inline-inference entries would otherwise emit ~hundreds of spectrum-pair PNGs + tail-percentile keys per epoch.

Changes:
- `fme.ace.aggregator.inference.spectrum.PowerSpectrumMetricConfig`:
  - `report_directional_bias: bool = True` — when `False`, drop the per-variable `positive_norm_bias` and `negative_norm_bias` scalars. `mean_abs_norm_bias` (their absolute sum) stays as the canonical directional summary.
  - `plot_variables: list[str] | None = None` — when set, restrict the per-variable spectrum-pair PNG output to these variable names. Scalar metrics are still emitted for every variable that passed the existing `variables` filter — `plot_variables` is the finer-grained knob.
- `fme.ace.aggregator.inference.histogram.HistogramMetricConfig`:
  - `percentile_variables: list[str] | None = None` — when set, only these variables get the 99.9999th-percentile (and any other configured percentile) scalar keys emitted. Histogram plots are unchanged. Filter is applied in `HistogramAggregator.get_logs` at the wandb-key level rather than threading the allowlist into the core `ComparedDynamicHistograms` class — keeps the core histogram code unchanged.

All three default to current behaviour, so existing configs are unaffected.

- [x] Tests added (`test_paired_spherical_power_spectrum_directional_bias_flag`, `test_paired_spherical_power_spectrum_plot_variables_filter`, new `test_histogram.py` with three cases — `None` default, explicit allowlist, empty allowlist)


fme/ace/aggregator/inference/histogram.py
```diff
@@ -11,22 +11,46 @@
...
+        enabled: master toggle for the metric.
+        strict: raise if the metric can't be built.
+        percentile_variables: when set, only these variables get the
+            99.9999th-percentile (and any other configured percentile)
+            scalar metrics emitted. The histogram plot is still emitted
+            for every variable that passed ``variables``. Defaults to
+            None (emit percentile keys for every variable that passed
+            ``variables`` — current behaviour). Use to restrict the noisy
+            tail-percentile keys to a small list (e.g. precipitation
+            only) while keeping the histogram plot cohort-wide.
```
AnnaKwa, line 29: If `variables` is set and `percentile_variables` contains key(s) not in `variables` I think currently this will just silently not record the extra variables. Can you check for this case and raise an error on build?

AnnaKwa (changes requested):
Mostly LGTM but could you make the histogram aggregator raise an error instead of failing silently if the finer-grained `percentile_variables` set includes names not in a specified `variables` set?

AnnaKwa (approved):

PR #1211, opened 2026-06-01
Deduplicate inference and validation callbacks

### LINKED ISSUE #537: Lower maintenance burden of overlapping `fme.{ace,coupled}` config params
With https://github.com/ai2cm/full-model/pull/1447, coupled training depends on the `OneStepAggregator` and `InferenceEvaluatorAggregator` from `fme.ace` and uses the same `InferenceEvaluatorAggregatorConfig` params. @mcgibbon [suggested](https://github.com/ai2cm/full-model/pull/1447#discussion_r1911151420) we use a shared data container to store these params.

mcgibbon:
Closes #537.

The two `get_inference_callback` implementations and the two `get_validation_callback` implementations in `fme.ace.train.train` / `fme.coupled.train.train` had drifted into four near-identical per-entry loops. This change extracts both into shared generic helpers in `fme.core.generics.trainer` parameterized by `InferenceTask` / `ValidationTask` dataclasses, so each trainer only owns its trainer-specific aggregator construction.

Behavior changes (all in coupled inference):
- coupled `get_inference_callback`: now raises (matching ACE) when a `weight>0` entry's logs lack `time_mean_norm/rmse/channel_mean`, instead of silently dropping that entry from the combined error
- coupled `get_inference_callback`: when every active entry has `weight=0`, the returned checkpoint-selection error is now `None` instead of `0.0`. Previously coupled training with all-zero weights would trigger "best inference checkpoint" saving every epoch (since `0.0 <= inf`); now it skips that, matching ACE's pre-existing behavior
- `build_inference_callback`: now raises if two entries' logs share a key, matching the pre-existing `build_validation_callback` behavior

`get_validation_callback` is pure refactor — both ACE and coupled bodies were already identical modulo aggregator construction.

Changes:
- `fme.core.generics.trainer.InferenceTask`, `fme.core.generics.trainer.build_inference_callback`: new dataclass + helper for inline inference runs
- `fme.core.generics.trainer.ValidationTask`, `fme.core.generics.trainer.build_validation_callback`: new dataclass + helper for per-epoch validation runs
- `fme.ace.train.train.get_inference_callback`, `fme.coupled.train.train.get_inference_callback`: build `InferenceTask` lists and delegate to the shared helper; per-trainer aggregator construction lives in new `_make_*_aggregator_factory` closures
- `fme.ace.train.train.get_validation_callback`, `fme.coupled.train.train.get_validation_callback`: build `ValidationTask` lists and delegate; per-trainer aggregator construction lives in new `_make_*_validation_aggregator_factory` closures
- Tests: helper behavior consolidated into `fme.core.generics.test_trainer.TestBuildValidationCallback` and `TestBuildInferenceCallback`, including previously-untested raise paths (missing-metric for `weight>0`, overlapping log keys, all-zero-weight returning `None`). ACE/coupled `TestGetValidationCallback` trimmed to a single wiring smoke test. `TestGetInferenceCallback` patch targets retargeted to `fme.core.generics.trainer.inference_one_epoch`; coupled `test_missing_metric_returns_none_error` replaced with `test_weighted_entry_missing_metric_raises`; new coupled `test_all_zero_weight_returns_none_error` locks in the behavior change


fme/core/generics/trainer.py
```diff
@@ -792,6 +855,73 @@ def inference_one_epoch(
...
+    """
+
+    name: str
+    data: InferenceDataABC[PS, FD]
+    aggregator_factory: Callable[[], InferenceAggregatorABC[PS, SD]]
+    epoch_set: frozenset[int]
+    weight: float = 0.0
+
+
+def build_inference_callback(
```
jpdunc23, line 879: Preexisting issue identified by agent review, optional to fix here:

`build_inference_callback lacks` the log-key overlap check that build_validation_callback has (`all_logs.update(logs)` with no `overlap = all_logs.keys() & logs.keys()` guard).

fme/core/generics/test_trainer.py
```diff
@@ -1598,3 +1600,118 @@ def test_finetune_ema_checkpoint_loads_ema_state(tmp_path: str):
 
     # decay is from stage2 config, not the checkpoint
     assert float(stage2_ema_state["decay"]) == pytest.approx(stage2_decay)
+
+
+class TestBuildValidationCallback:
```
jpdunc23, line 1607: Optional: Consider adding `TestBuildInferenceCallback` mirroring `TestBuildValidationCallback`.

> Same structure as `TestBuildValidationCallback` — a `_make_task` helper that builds an `InferenceTask` with a mock factory, and a `_call` helper that patches `inference_one_epoch` and invokes the callback. Test cases would cover:
> 
> - Epoch gating: not in `inference_epochs` returns `({}, None)`; in `inference_epochs` but no task has it in `epoch_set` also returns `({}, None)`
> - Weighted error accumulation: single entry, multiple entries, weighted sum
> - Weight-zero entries: logs kept but excluded from error; all-zero returns `None`
> - Missing metric: raises for weight > 0, skipped for weight = 0
> - Factory called fresh each invocation (like the existing validation test)
> - Per-task `epoch_set` filtering: only matching tasks run on a given epoch
> 
> Essentially the same cases the validation suite covers, plus the inference-specific epoch-set filtering. The overlap-check gap would also become visible — you'd either add a test that expects it to raise (and discover it doesn't), or consciously document the omission.

jpdunc23 (approved):
LGTM. Have some optional comments; I think adding `TestBuildInferenceCallback` would be good, but up to you.

I think we can finally close https://github.com/ai2cm/ace/issues/537 with this PR, so please include it in the description to auto-resolve on merge.

fme/core/generics/trainer.py
mcgibbon: Yeah that would be good to add.

PR #1215, opened 2026-06-02
Fix DenoisingMoEPredictor.generate_on_batch loss reduction

AnnaKwa:
PR #1159 changed self.loss to return `list[LossComponent]`. DiffusionModel.generate_on_batch was updated to unpack that list and call .mean(), but DenoisingMoEPredictor was not updated. Downstream calls of `generate_on_batch` from the MoE predictor would error.


mcgibbon (approved):
It would be nice to add a test that covers this case.

PR #1218, opened 2026-06-03
Add corrector regularization to training

mcgibbon:
Add an optional training-time penalty on the magnitude of corrector adjustments, enabling L1/L2 regularization that discourages the model from relying on the corrector. Supported for both single-module and coupled training; when configured, the specified loss is computed between each corrector's per-variable correction (post-corrector minus pre-corrector) and a zero baseline in the loss-normalized space, then accumulated into the training loss.

Changes:
- `fme.core.step.step.StepResult`: new dataclass replacing the bare `TensorDict` return type of `StepABC.step`; carries the denormalized output plus optional per-variable corrections. All `StepABC` implementations and consumers updated.
- `fme.core.step.single_module.step_with_adjustments`: records `corrections = post − pre` when a corrector runs; `MultiCallStep` forwards corrections from its wrapped step.
- `fme.core.loss.CorrectorRegularizationConfig`: new `LossConfig` + `weight` wrapper, with validation rejecting `EnsembleLoss`/`NaN`/`global_mean_type`.
- `fme.ace.stepper.single_module.TrainStepperConfig.corrector_regularization`: optional field; built via `Stepper.build_corrector_regularization_loss` and applied per optimized step in `TrainStepper._accumulate_loss`. Per-step (\`corrector_regularization_step_{i}\`) and epoch-aggregated (\`corrector_regularization\`) metrics.
- `fme.coupled.stepper.ComponentTrainingConfig.corrector_regularization`: per-realm version. \`CoupledStepperTrainLoss.compute_corrector_regularization\` wired into \`CoupledTrainStepper._accumulate_step_loss\`; emits \`loss/{realm}_corrector_regularization_step_{i}\`.
- \`fme.ace\` re-exports \`LossConfig\` and \`CorrectorRegularizationConfig\` so nested-dataclass symbol checks see them.

- [x] Tests added

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -109,7 +109,7 @@ from fme.core.dataset.utils import FillNaNsConfig
 from fme.core.dataset.xarray import OverwriteConfig, XarrayDataConfig
 from fme.core.generics.lr_tuning import LRTuningConfig
 from fme.core.gridded_ops import GriddedOperations
-from fme.core.loss import StepLossConfig
+from fme.core.loss import CorrectorRegularizationConfig, LossConfig, StepLossConfig
 from fme.core.normalizer import NormalizationConfig
 from fme.core.ocean import OceanConfig, SlabOceanConfig
 from fme.core.optimization import CheckpointConfig
--- a/fme/ace/step/fcn3.py
+++ b/fme/ace/step/fcn3.py
@@ -24,7 +24,7 @@ from fme.core.packer import Packer
 from fme.core.registry import CorrectorSelector
 from fme.core.step.args import StepArgs
 from fme.core.step.single_module import step_with_adjustments
-from fme.core.step.step import StepABC, StepConfigABC, StepSelector
+from fme.core.step.step import StepABC, StepConfigABC, StepResult, StepSelector
 from fme.core.typing_ import TensorDict, TensorMapping
 
 DEFAULT_TIMESTEP = datetime.timedelta(hours=6)
@@ -439,7 +439,7 @@ class FCN3Step(StepABC):
         self,
         args: StepArgs,
         wrapper: Callable[[nn.Module], nn.Module] = lambda x: x,
-    ) -> TensorDict:
+    ) -> StepResult:
         """
         Step the model forward one timestep given input data.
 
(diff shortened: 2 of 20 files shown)


jpdunc23:
What loss / weight do you suggest I try with this feature? Thinking about launching two experiments, one with an L1 penalty and another with an L2 penalty, but lmk if you have a particular config in mind.

mcgibbon:
> What loss / weight do you suggest I try with this feature? Thinking about launching two experiments, one with an L1 penalty and another with an L2 penalty, but lmk if you have a particular config in mind.

Yeah that matches what I was thinking.

The normalization is the same as for the main loss term, which helps reason about weighting. L2 is probably the better match for the behavior you want, and also allows the model to keep slightly-negative zero-values with little penalty, so I'd focus on that if you want to do a weight sweep. 1.0 is a reasonable starting point for weighting, which means the model puts equal importance to obeying the corrector's goals and to having skill. Even a smaller value like 0.05-0.2 should significantly reduce the divergence between pre-and post-corrector values, because you'll only get divergence to the extent that this helps the model predict better. If you got large divergence even with 0.2 weight, it would imply the divergence is fairly significantly helping the model's skill, which I don't expect.

Maybe try 0.2, 1.0, and 5.0 on L2 as a first pass if you want to sweep, or 1.0 on L2 if you're limiting to one config?

jpdunc23:
@mcgibbon please merge https://github.com/ai2cm/ace/pull/1223 when you can (fixed on my exper branch after [failing job](https://beaker.org/orgs/ai2/workspaces/ace/work/01KT7VZ79P00EVYTZGN8X7B4H7?taskId=01KT7VZ7FPRH4TTEP7TR05QVYP&jobId=01KT7VZ7N1B1192DZWGN5ZNEW8))

jpdunc23:
Experiment here: https://wandb.ai/ai2cm/ace-samudra-cm4/runs/xzmbufek

mcgibbon:
> Experiment here: https://wandb.ai/ai2cm/ace-samudra-cm4/runs/xzmbufek

Looks like so far it's improving the thetao evolution as much as the precorrector optimization. To be seen if his continues, or if the regularization approach bottoms out earlier on how much improvement it gives. 
image

jpdunc23:
Could you please add a feature to exclude the penalty for specific variables? Excluding sea ice in pre-corrector optim removed the small positive sea ice biases in mid latitudes, so I think it should do the same here.

Here's an example from my PR of configuring this using the `NameAndPrefixMatcher` class:

https://github.com/ai2cm/ace/blob/046a3f8a9a0813f6e0f20789d0343e2e918a1753/fme/ace/stepper/single_module.py#L1462-L1515

PR #1219, opened 2026-06-03
Serialized mixture of experts checkpoint

AnnaKwa:
This PR adds the ability to serialize a denoising mixture of experts predictor for inference so that it can just be loaded by providing a single `mixture_of_experts_path` in the config, instead of having to specify individual model paths and the range of noise each is intended to be routed for.

The changes are separated into individual commits for easier reviewing.

Changes:
- renaming functionality (used to handle the cases when ACE output field names differ from the input field names used to train the model) are consolidated into the `DiffusionModelConfig.build`. Previously, renaming was in two different places: the `CheckpointModelConfig` did the channel renaming in the config before calling `build` and `DiffusionModelConfig.build` would do some inverse lookups to make sure the normalizer names matched the renamed set. This update makes the renaming cleaner and ensures that the serialized MoE checkpoint works if renaming is done (before it would not because the renamed normalizer names in the serialized MoEPredictor would no longer need renaming).
- Add a `DenoisingMoEPredictor.save` method so that this predictor can be serialized. Add a `DenoisingMoECheckpointConfig` class that only needs a path arg. 
- Add `DenoisingMoECheckpointConfig` as an option to use when providing a checkpoint config in the entrypoints (inference, predict, evaluator). Add a script that takes a `DenoisingMoECheckpointConfig`, builds, and saves the predictor while allowing for this to be done on a local environment that may lack the apex and type options specified in the checkpoints. 

I verified that a bundled predictor loaded using `DenoisingMoECheckpointConfig` works for prediction ([beaker link](https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KT569NV57WHAF7S84CWNPNTH?taskId=01KT569P0SZSF9CDWECWDGRRPX&jobId=01KT569P3ZAFGX0S1659CNG5CZ)).

- [x] Tests added

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -247,7 +247,6 @@ class DiffusionModelConfig:
         # runtime names exposed by the built DiffusionModel. All rename
         # application happens here: normalizer keys, in_names, out_names. The
         # DiffusionModel itself just stores what it is given.
-        rename = rename or None
         rename_map = rename or {}
         in_names = [rename_map.get(n, n) for n in self.in_names]
         out_names = [rename_map.get(n, n) for n in self.out_names]
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -242,10 +242,16 @@ class DiffusionModelConfig:
         rename: dict[str, str] | None = None,
         static_inputs: StaticInputs | None = None,
     ) -> "DiffusionModel":
-        invert_rename = {v: k for k, v in (rename or {}).items()}
-        orig_in_names = [invert_rename.get(name, name) for name in self.in_names]
-        orig_out_names = [invert_rename.get(name, name) for name in self.out_names]
-        normalizer = self.normalization.build(orig_in_names, orig_out_names, rename)
+        # self.in_names / self.out_names are the ORIGINAL (training-time) names
+        # keying the normalization means/stds. `rename` translates them to the
+        # runtime names exposed by the built DiffusionModel. All rename
+        # application happens here: normalizer keys, in_names, out_names. The
+        # DiffusionModel itself just stores what it is given.
+        rename = rename or None
+        rename_map = rename or {}
+        in_names = [rename_map.get(n, n) for n in self.in_names]
+        out_names = [rename_map.get(n, n) for n in self.out_names]
+        normalizer = self.normalization.build(self.in_names, self.out_names, rename)
         loss = self.loss.build(gridded_operations=None)
         # We always use standard score normalization, so sigma_data is
         # always 1.0. See below for standard score normalization:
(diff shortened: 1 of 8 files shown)


fme/downscaling/models.py
```diff
@@ -242,18 +242,24 @@ def build(
...
-        invert_rename = {v: k for k, v in (rename or {}).items()}
-        orig_in_names = [invert_rename.get(name, name) for name in self.in_names]
-        orig_out_names = [invert_rename.get(name, name) for name in self.out_names]
-        normalizer = self.normalization.build(orig_in_names, orig_out_names, rename)
+        # self.in_names / self.out_names are the ORIGINAL (training-time) names
+        # keying the normalization means/stds. `rename` translates them to the
+        # runtime names exposed by the built DiffusionModel. All rename
+        # application happens here: normalizer keys, in_names, out_names. The
+        # DiffusionModel itself just stores what it is given.
+        rename = rename or None
```
frodre: already specified as this in the `build` arguments

fme/downscaling/models.py
```diff
@@ -242,18 +242,24 @@ def build(
         rename: dict[str, str] | None = None,
         static_inputs: StaticInputs | None = None,
     ) -> "DiffusionModel":
-        invert_rename = {v: k for k, v in (rename or {}).items()}
-        orig_in_names = [invert_rename.get(name, name) for name in self.in_names]
-        orig_out_names = [invert_rename.get(name, name) for name in self.out_names]
-        normalizer = self.normalization.build(orig_in_names, orig_out_names, rename)
+        # self.in_names / self.out_names are the ORIGINAL (training-time) names
```
frodre, line 245: Makes me think we could remove the internals touchpoint by just making the renaming live in the data layer rather than the built model layer.  Not necessary for this PR

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -170,11 +176,16 @@ def __init__(
         sigma_ranges: list[tuple[float, float]],
         num_diffusion_generation_steps: int,
         churn: float,
+        expert_renames: list[dict[str, str] | None] | None = None,
     ) -> None:
         if not experts:
             raise ValueError("experts must be non-empty.")
         if len(experts) != len(sigma_ranges):
             raise ValueError("experts and sigma_ranges must have the same length.")
+        if expert_renames is None:
```
frodre: Is there a reason we need this expansion rather than leaving as `None` and doing something like `if expert_renames is not None and len(...) != len(...)?  It would also be nice if we could just have an option to specify a single rename applied to both experts.

fme/downscaling/predictors/serial_denoising.py
```diff
@@ -277,3 +289,94 @@ def generate_on_batch(
...
+            experts=experts,
+            sigma_ranges=sigma_ranges,
+            num_diffusion_generation_steps=int(state["num_diffusion_generation_steps"]),
+            churn=float(state["churn"]),
+            expert_renames=list(expert_renames),
+        )
+
+
+@dataclasses.dataclass
+class DenoisingMoECheckpointConfig:
```
frodre: Consider renaming to `DenoisingMoEBundledConfig`. Otherwise, I'd have a hard time knowing which is which.

frodre (approved):
LGTM!  I would consider renaming the new config name and adding a convenience renaming where we can apply the same renaming to both sub models, since that's likely to be our most common use.

PR #1220, opened 2026-06-03
Mark dhconv timing test as serial

mcgibbon:
This PR marks a timing-sensitive test as serial, to avoid it flaking due to noisy xdist neighbors.

Changes:
- No public API changes.

- [ ] Tests added


jpdunc23 (approved):
Thanks!

PR #1221, opened 2026-06-03
Fix histogram ckpt selection and percentile frac_of_target reporting

AnnaKwa:
This fixes two bugs related to the percentile fraction of target metrics:

1) 'best histogram checkpoint' selection in downscaling was switched to use this metric but was still picking the checkpoint by minimizing the value. This fixes the selection to minimize abs(1.0 - percentile_frac_of_target).
2) `ComparedDynamicTailsHistograms.get_wandb` had numerator/denominator flipped (is ok in ComparedDynamicHistograms`)
3) Moves the calculation of the best checkpoint selection metrics (always expected to be minimized) into the generation aggregator. They are accessed by the trainer by called the new method `GenerationAggregator.get_checkpoint_selection_metrics`
 

- [x] Tests added

### DIFF
--- a/fme/core/histogram.py
+++ b/fme/core/histogram.py
@@ -597,7 +597,7 @@ class ComparedDynamicTailsHistograms(ComparedDynamicHistograms):
                     )
         for key, value in target_metrics.items():
             return_metrics[f"prediction_frac_of_target/{key}"] = (
-                value / return_metrics[key]
+                return_metrics[key] / value
             )
         return return_metrics
 
--- a/fme/downscaling/aggregators/generation.py
+++ b/fme/downscaling/aggregators/generation.py
@@ -312,9 +312,9 @@ class GenerationAggregator:
         variable_metadata: Mapping[str, VariableMetadata] | None = None,
         include_positional_comparisons: bool = True,
         histogram_ckpt_selection_metric: str = (
-            "prediction_frac_of_target/99.9999th-percentile"
+            "histogram/prediction_frac_of_target/99.9999th-percentile"
         ),
-        best_ckpt_selection_metric: str = "relative_crps_bicubic",
+        best_ckpt_selection_metric: str = "metrics/relative_crps_bicubic",
     ) -> None:
         self._agg = Aggregator(
             dims,
--- a/fme/downscaling/aggregators/generation.py
+++ b/fme/downscaling/aggregators/generation.py
@@ -1,3 +1,4 @@
+import re
 from collections.abc import Collection, Mapping
 from typing import Any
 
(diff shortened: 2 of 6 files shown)


fme/core/histogram.py
```diff
@@ -597,7 +597,7 @@ def get_wandb(
                     )
         for key, value in target_metrics.items():
             return_metrics[f"prediction_frac_of_target/{key}"] = (
-                value / return_metrics[key]
+                return_metrics[key] / value
```
mcgibbon, line 600: nit: naming this `target_value` would make it harder to write a bug like the one fixed.

fme/downscaling/train.py
```diff
@@ -347,6 +343,12 @@ def save_best_checkpoint(self, valid_metrics: dict[str, float]) -> None:
         else:
             raise ValueError("Best checkpoint path is not set")
 
+    def _get_best_histogram_tail_metric(self, valid_metrics: dict[str, float]) -> float:
+        metric = valid_metrics[self._best_histogram_tail_name]
+        if "frac_of_target" in self._best_histogram_tail_name:
+            return abs(1.0 - metric)
```
mcgibbon: Issue: This is some fairly low-level metric-specific code to put in the already-quite-large Trainer. I had to read it for a fair bit of time to convince myself it's right, because I don't have much context about the values it's operating on from this level.

Suggestion: Is there any way to have the validation metrics already contain this value? If you don't want to report it to wandb, you could add a method `validation_aggregator.get_loss()` that returns a scalar we want to minimize for checkpoint selection (similar for inference).

fme/downscaling/aggregators/generation.py
```diff
@@ -52,6 +53,50 @@ def _get_map_caption(key: str, data: torch.Tensor) -> str:
     return caption
 
 
+def _get_complement_percentile_prefix(prefix):
```
AnnaKwa, line 56: These helper functions are copied over from their previous location in train.py.

fme/downscaling/aggregators/test_aggregators.py
```diff
@@ -421,3 +421,25 @@ def test_upsample_tensor():
     t = torch.tensor([[1, 2], [3, 4]])
     expected = torch.tensor([[1, 1, 2, 2], [1, 1, 2, 2], [3, 3, 4, 4], [3, 3, 4, 4]])
     assert torch.equal(expected, upsample_tensor(t, 2))
+
+
+@pytest.mark.parametrize(
```
AnnaKwa, line 430: This test is copied over from where it used to live in test_train.py.

fme/downscaling/aggregators/generation.py
```diff
@@ -362,3 +415,23 @@ def get_dataset(self) -> xr.Dataset:
             if hasattr(agg, "get_dataset"):
                 ds = ds.merge(agg.get_dataset())
         return ds
+
+    def get_checkpoint_selection_metrics(self) -> dict[str, float]:
```
mcgibbon, line 419: This is still tightly coupled to the Trainer in that they both need to know the selection metric string, but it's fine. I'm going to make a follow-on PR refactoring the ACE aggregators, and I'll make a minor update to this which removes the coupling when I do that.

mcgibbon (approved):

PR #1222, opened 2026-06-03
Feature/uncorrected inference metrics [don't merge]

jpdunc23:

Changes:

- [ ] Tests added


fme/ace/step/fcn3.py
```diff
@@ -448,7 +448,8 @@ def step(
             wrapper: Wrapper to apply over each nn.Module before calling.
 
         Returns:
-            The denormalized output data at the next time step.
+            The output at the next timestep and the pre-correction values of any
```
mcgibbon, line 451: Let's be generic here and just say it's the output including next-timestep data, so this can't go out of date with further updates to StepOutput.

fme/core/step/step.py
```diff
@@ -15,6 +15,28 @@
...
+            adjustments (corrector, ocean, prescribed prognostics) are applied.
+        uncorrected: The pre-correction values of exactly the variables the
+            corrector modified. Empty if no corrector ran or none were modified.
+            This is a sparse subset of ``output`` used to evaluate how much the
+            stepper relies on the corrector. Note ocean and prescribed-prognostic
+            adjustments run after the corrector and are not reflected here.
+    """
+
+    output: TensorDict
+    uncorrected: TensorDict
```
mcgibbon, line 33: As long as you put this as `corrections: TensorDict | None` (or the singular `correction`) in your re-implementation this should be fine.

The `| None` is important to support steps that don't define this.

mcgibbon (commented):
We had a slack discussion and converged on reporting metrics on the corrections instead of the pre-corrected state, since it's hard to derive the mean effect of the corrector and the variability of that effect based on metrics like RMSE applied to pre-corrector fields. Also talked about having a Loss like object that takes in StepOutput, and being able to swap it between computing regularization on the corrections, and applying the loss to de-corrected state (would be nice to be able to configure both options).

PR #1225, opened 2026-06-04
Pin global dry-air mass to the IC via threaded StepperState

### LINKED ISSUE #424: Dry-air mass non-conservation in ACE2 models
Documenting here that we're observing dry-air mass non-conservation in some ACE2 inference despite configuring these models with `stepper.corrector.conserve_dry_air = True`. Pictured below is an example from a SHiELD-AMIP 1° model's 82-year inference. There is a drift of about 12 Pa in 82 years. The drift rate is small, some runs drift positive and some negative, and the rate seems fairly consistent over time for a given model. Currently I don't think we know quite where this is coming from. 

W B Chart 8_14_2024, 3_07_19 PM
 
@spencerkclark also saw this result in ACE-SOM model inference (below).

Screenshot 2024-07-29 at 3 27 11 PM

It also looks like [some drift is evident in the ERA5 ACE2 inference](https://wandb.ai/ai2cm/ace2/reports/inference-mean-weighted_mean_gen-surface_pressure_due_to_dry_air-24-08-14-15-34-05---Vmlldzo5MDM2NjQw) -- @oliverwm1?

Something perhaps relevant happens [when running "dataset inference" using the two SHiELD-AMIP 1° ensemble members as the target (IC2) and prediction (IC1), respectively](https://wandb.ai/ai2cm/ace/runs/bb1icrnr?nw=69ww1pbj5ws), and also including the training ensemble member (IC1) as the monthly reference data to show up on the annual mean plots. Normally there would be no point in including IC1 as the monthly reference data in this kind of dataset inference run, because it will line up exactly with the prediction as they are both just IC1. And that is true for every variable except dry air mass, which [has a strange looking plot](https://wandb.ai/ai2cm/ace/reports/inference-annual-surface_pressure_due_to_dry_air-24-08-14-15-14-06---Vmlldzo5MDM2NTI4). Since the monthly reference data is computed on a different commit than the dataset inference run, it suggest that something is not the same in the dry-air mass computations for these two workflows despite having the same input data.

mcgibbon:
Long inference rollouts can drift in global dry-air mass even with ``conserve_dry_air`` enabled, because the previous per-step conservation against the prior step's input accumulates small numerical errors. This PR changes ``conserve_dry_air`` to a strictly stronger constraint: pin the global-mean dry-air mass to its value at the initial condition, threaded across prediction windows via a new per-sample ``StepperState`` carried on ``BatchData``.

The carrier is generic — components own their own sub-state. Today only the corrector uses it (``CorrectorState.global_dry_air_mass``); the stepper is opaque to its contents.

### Behavior change for existing saved checkpoints

This changes the semantics of the existing ``conserve_dry_air`` flag rather than adding a new one, so checkpoints trained or run with ``conserve_dry_air=True`` will produce different output than before:

- **Before:** on each step, gen's dry-air mass was matched to the prior step's ``input_data`` dry-air mass. The target drifted with the rollout, so accumulated round-off let the global mean wander.
- **After:** the corrector seeds ``CorrectorState.global_dry_air_mass`` from the IC's area-weighted dry-air pressure on its first call, and on every subsequent call pins gen's dry-air mass to that fixed reference. The reference is threaded across inference windows via ``BatchData.stepper_state``, so multi-window rollouts target the same IC value end-to-end.

No config migration is needed — the flag name is unchanged — but anyone reproducing prior numerics from a saved checkpoint should expect bit-for-bit differences in any rollout long enough for the old per-step drift to be visible.

### Changes

- ``fme.core.stepper_state.StepperState`` and ``fme.core.corrector.state.CorrectorState`` — new dataclasses for per-sample state that flows across step calls.
- ``fme.ace.data_loading.batch_data.BatchData.stepper_state`` — new optional field, threaded through ``to_device`` / ``to_cpu`` / ``pin_memory`` / ``broadcast_ensemble`` / ``scatter_spatial`` / ``subset_names`` / ``select_time_slice`` / ``prepend`` / ``remove_initial_condition`` / ``compute_derived_variables``, with shape validation in ``__post_init__``. The data loader never sets it; only ``Stepper.predict`` does.
- ``fme.core.corrector.registry.CorrectorABC.__call__`` — now takes ``corrector_state`` and returns ``(gen_data, corrector_state)``. Concrete correctors (``AtmosphereCorrector``, ``OceanCorrector``, ``IceCorrector``) updated.
- ``fme.core.step.args.StepArgs.stepper_state`` and ``fme.core.step.single_module.step_with_adjustments`` — thread state into and out of the per-step path; ``step_with_adjustments`` extracts/injects the corrector's slice opaquely.
- ``StepABC.step``, ``Stepper.step``, ``Stepper.predict_generator``, ``Stepper.get_prediction_generator``, ``Stepper.predict``, ``Stepper.predict_paired``, and ``process_prediction_generator_list`` — return/forward the new tuple; ``predict`` attaches the terminal state to the returned ``PrognosticState`` so it propagates from one inference window to the next via ``Looper``. Same threading is used in training; each batch seeds and discards its own state.
- ``fme.coupled.stepper`` — unpacks the new tuple at the consumer site (no behavior change; coupled stepper doesn't use stepper_state today).
- ``fme.core.corrector.atmosphere`` — ``conserve_dry_air`` reworked as described above. The previous ``_force_conserve_dry_air`` helper (with its "match the prior step's input" semantics) is removed; gen's surface pressure is now solved from the pinned dry-air pressure target on every call.

- [x] Tests added
  - ``BatchData.stepper_state`` pass-through tests in ``fme/ace/data_loading/test_batch_data.py``
  - ``StepArgs.stepper_state`` propagation in ``fme/core/step/test_args.py``
  - Cross-``predict`` propagation integration test (``test_predict_threads_stepper_state_across_calls``) using a recording corrector
  - ``conserve_dry_air`` unit tests in ``fme/core/corrector/test_atmosphere.py``: seeding from the IC, persistence across calls (drifted input does not re-seed), and the missing-vertical-coordinate guard
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated (n/a)

Closes https://github.com/ai2cm/ace/issues/424


fme/core/corrector/atmosphere.py
```diff
@@ -131,6 +142,7 @@ class AtmosphereCorrectorConfig(CorrectorConfigABC):
     ) = None
     force_positive_names: list[str] = dataclasses.field(default_factory=list)
     total_energy_budget_correction: EnergyBudgetConfig | None = None
+    pin_global_dry_air_mass: bool = False
```
oliverwm1: my feeling is we want this to always happen if `conserve_dry_air` is set to True. Not convinced this should be user configurable.

oliverwm1 (commented):
related: https://github.com/ai2cm/ace/issues/424

Did you consider adding statefulness to the stepper or step? Instead of carrying this around in BatchData or as a separate arg to the predict method.

fme/core/corrector/atmosphere.py
mcgibbon: I'm good with always-on.

mcgibbon:
> Did you consider adding statefulness to the stepper or step? Instead of carrying this around in BatchData or as a separate arg to the predict method.

I considered it, and would really like to avoid it. Stateful functions are much harder to reason about and make it easier to introduce bugs. While the stepper and step do have state, it's purely training and configuration type state of what's basically a stateless function, and that's helped simplify things so far when developing the code.

There's nothing to prevent a single Stepper from being used for many ICs in series (as we do in training) or in an interleaved way (as we do in candidate-lr-tuning inference, and did back when we supported multi-batch inference), and having state in the stepper would require more APIs to manage resetting that state, or for us to re-create the stepper for each IC.

The BatchData is what's responsible for holding the model's state, and there are other benefits to keeping it as a complete model state descriptor. For example, if we ever wanted to enable restart checkpoints, this change would seamlessly work (the stepper state would be stored in the IC checkpoint). It would also let us call one stepper up to a certain point in a simulation, and then call a differently-configured stepper with the same internal state, if that's ever important (e.g. to see if a perturbation to how hidden state is treated would cause or prevent an instability on a certain step). If we want to record information about the internal state in aggregators, that also seamlessly works (makes less sense for a constant dry air mass, but more sense for other potential hidden-state types). And it would let us propagate internal state for arbitrarily long rollouts if we ever implement pooling-style reinforcement learning approaches that we've discussed.

All in all, this feels like the "right" way to go.

mcgibbon:
> related: #424

Updated to mark as being closed by this PR, since it will once I remove the configurability.

spencerkclark (approved):
A lot of places to thread this through, but overall this looks good to me and is thoroughly tested!

> For example, if we ever wanted to enable restart checkpoints

I believe we do have something like this in the form of [segmented inference](https://github.com/ai2cm/ace/blob/4abc471862f7bfe2c0e87fa06c9633f81f2ae9e9/fme/ace/inference/inference.py#L389-L425).  

I think until we include this state in restart files and reinitialization, segmented runs will produce different results than continuous runs, since upon reinitialization the dry air mass will be recomputed from the intermediate state and may differ infinitesimally due to floating point error.  I do not think that is necessarily critical for this PR, but something to keep in mind.

oliverwm1:
Will this PR change rollout behavior for existing checkpoints? If so, this should be highlighted clearly in the commit title/message and PR description.

mcgibbon:
> Will this PR change rollout behavior for existing checkpoints? If so, this should be highlighted clearly in the commit title/message and PR description.

Whoops, it wasn't originally and then I missed updating the PR description when we changed the behavior. Should I amend the commit to main and force-push, or leave it?

PR #1226, opened 2026-06-04
Extract NameAndPrefixMatcher to shared `fme.core` module

jpdunc23:
Move the prefix-aware name/prefix matching logic out of StaticSpatialMasking into a new shared `NameAndPrefixMatcher` in `fme/core/name_and_prefix_matcher.py`, exposing a single `match(name) -> bool` method. The matching semantics are preserved exactly: a bare name matches the 2D variable and its 3D levels, a trailing-underscore prefix matches all levels, and an explicit `name_<level>` matches exactly.

Spatial masking now constructs and uses `NameAndPrefixMatcher` in place of its private `_build_regex`. Pure refactor, no behavior change.

Changes:
- `StaticSpatialMasking` init arg `exclude_names_and_prefixes: list[str]` removed in favor of `exclude: NameAndPrefixMatcher = NameAndPrefixMatcher()`. The `NameAndPrefixMatcher` object is built from the configured `exclude_names_and_prefixes` in `StaticSpatialMaskingConfig.build()`.

- [x] Tests added

### DIFF
--- a/fme/core/name_and_prefix_matcher.py
+++ b/fme/core/name_and_prefix_matcher.py
@@ -29,6 +29,8 @@ class NameAndPrefixMatcher:
             return r"|".join(regex)
         return None
 
-    def matches(self, name: str) -> bool:
+    def match(self, name: str) -> bool:
         """Return whether ``name`` matches any configured name or prefix."""
-        return bool(self._regex and re.match(self._regex, name))
+        if self._regex is None:
+            return False
+        return bool(re.match(self._regex, name))
--- a/fme/core/name_and_prefix_matcher.py
+++ b/fme/core/name_and_prefix_matcher.py
@@ -1,0 +1,34 @@
+import re
+
+
+class NameAndPrefixMatcher:
+    """Match variable names against a list of names and prefixes.
+
+    The matching convention is:
+        - A bare name (e.g. ``thetao``) matches the 2D variable ``thetao`` and
+          all of its 3D levels ``thetao_<level>``.
+        - A trailing-underscore prefix (e.g. ``thetao_``) matches all
+          ``thetao_<level>``.
+        - An explicit ``name_<level>`` (e.g. ``thetao_0``) matches exactly.
+    """
+
+    def __init__(self, names_and_prefixes: list[str] | None = None):
+        self._regex = self._build_regex(names_and_prefixes)
+
+    def _build_regex(self, names_and_prefixes: list[str] | None) -> str | None:
+        if names_and_prefixes:
+            regex = []
+            for name in names_and_prefixes:
+                if name.endswith("_"):
+                    regex.append(rf"^{name}\d+$")
+                elif not re.match(r".+_\d+$", name):
+                    regex.append(f"^{name}$")
+                    regex.append(rf"^{name}_\d+$")
+                else:
+                    regex.append(rf"^{name}$")
+            return r"|".join(regex)
+        return None
+
+    def matches(self, name: str) -> bool:
+        """Return whether ``name`` matches any configured name or prefix."""
+        return bool(self._regex and re.match(self._regex, name))
--- a/fme/core/spatial_masking.py
+++ b/fme/core/spatial_masking.py
@@ -72,6 +72,7 @@ class StaticSpatialMaskingConfig:
         Build StaticSpatialMasking.
 
         """
+        exclude = NameAndPrefixMatcher(self.exclude_names_and_prefixes)
         if isinstance(self.fill_value, float):
             return StaticSpatialMasking(
                 mask_value=self.mask_value,
@@ -79,7 +80,7 @@ class StaticSpatialMaskingConfig:
                     lambda: torch.as_tensor(self.fill_value)
                 ),
                 mask=mask,
-                exclude_names_and_prefixes=self.exclude_names_and_prefixes,
+                exclude=exclude,
             )
         if means is None:
             raise ValueError(
@@ -90,7 +91,7 @@ class StaticSpatialMaskingConfig:
             mask_value=self.mask_value,
             fill_value=means,
             mask=mask,
-            exclude_names_and_prefixes=self.exclude_names_and_prefixes,
+            exclude=exclude,
         )
 
 
@@ -100,7 +101,7 @@ class StaticSpatialMasking:
         mask_value: int,
         fill_value: float | TensorMapping,
         mask: HasGetSpatialMask,
-        exclude_names_and_prefixes: list[str] | None = None,
+        exclude: NameAndPrefixMatcher = NameAndPrefixMatcher(),
     ):
         if isinstance(fill_value, float):
             fill_mapping: TensorMapping = collections.defaultdict(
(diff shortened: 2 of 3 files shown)


fme/core/name_and_prefix_matcher.py
```diff
@@ -0,0 +1,34 @@
...
+                    regex.append(f"^{name}$")
+                    regex.append(rf"^{name}_\d+$")
+                else:
+                    regex.append(rf"^{name}$")
+            return r"|".join(regex)
+        return None
+
+    def matches(self, name: str) -> bool:
+        """Return whether ``name`` matches any configured name or prefix."""
+        return bool(self._regex and re.match(self._regex, name))
```
mcgibbon: None-style checks make it clearer what this is checking.
```suggestion
        return bool((self._regex is not None) and re.match(self._regex, name))
```

mcgibbon: nit: an explicit `if self._regex is None: return False` would make this logic a bit clearer, I'm having a hard time parsing it.

fme/core/name_and_prefix_matcher.py
```diff
@@ -0,0 +1,34 @@
...
+                    regex.append(rf"^{name}\d+$")
+                elif not re.match(r".+_\d+$", name):
+                    regex.append(f"^{name}$")
+                    regex.append(rf"^{name}_\d+$")
+                else:
+                    regex.append(rf"^{name}$")
+            return r"|".join(regex)
+        return None
+
+    def matches(self, name: str) -> bool:
```
mcgibbon: nit: using `def match(` would keep the naming consistent with what's in `re`, which is slightly more expected.

fme/core/spatial_masking.py
mcgibbon: Suggestion (very optional): If you took a `exclude: NameAndPrefixMatcher = NameAndPrefixMatcher()` then `exclude` could be duck-typed with a re.match object which is kind of interesting and maybe useful. In general I've noticed "build an object during the init of an object that has a builder-config class" is an antipattern, with "build the object the builder method and pass it as arg to init" being the better version.

fme/core/spatial_masking.py
```diff
@@ -111,25 +111,10 @@ def __init__(
...
-                elif not re.match(r".+_\d+$", name):
-                    regex.append(f"^{name}$")
-                    regex.append(rf"^{name}_\d+$")
-                else:
-                    regex.append(rf"^{name}$")
-            return r"|".join(regex)
-        return None
+        self._exclude_matcher = NameAndPrefixMatcher(exclude_names_and_prefixes)
 
     def _masks(self, name: str):
```
mcgibbon: nit: type hint return value as bool

fme/core/name_and_prefix_matcher.py
```diff
@@ -0,0 +1,34 @@
+import re
+
+
+class NameAndPrefixMatcher:
```
mcgibbon, line 4: Suggestion (optional): this object could be replaced by a function that returns a `re.match` object. All it takes from here is having it return a regex that does not match anything, when names_and_prefixes is None, such as `(?=a)(?!a)` which requires a character to be a and not a.

fme/core/spatial_masking.py
jpdunc23: > `exclude: NameAndPrefixMatcher = NameAndPrefixMatcher()`

I like this.

fme/core/name_and_prefix_matcher.py
jpdunc23: Good point, but I lean toward keeping the class. The benefit of the class lies in the documentation it provides at the consumer (`StaticSpatialMasking`). The `NameAndPrefixMatcher` type hint communicates the matching convention right `exclude` is used, whereas with a `re.Pattern` type hint you'd have to trace back to `build()` to see how the regex was created.

fme/core/name_and_prefix_matcher.py
jpdunc23: Updated

fme/core/name_and_prefix_matcher.py
jpdunc23: Updated

fme/core/spatial_masking.py
jpdunc23: Updated

fme/core/spatial_masking.py
jpdunc23: Updated

fme/core/spatial_masking.py
```diff
@@ -111,25 +112,10 @@ def __init__(
...
-            return r"|".join(regex)
-        return None
-
-    def _masks(self, name: str):
-        exclude = self._exclude_regex and re.match(self._exclude_regex, name)
-        return not exclude
+        self._exclude = exclude
+
+    def _masks(self, name: str) -> bool:
+        return not self._exclude.match(name)
```
mcgibbon, line 118: nit: it would simplify the call logic slightly if this were replaced with a `self._get_mask_tensor_for(name)` that returns None when `self._exclude.match(name)`.

```
            if not self._masks(name):
                continue
            mask = self._mask.get_mask_tensor_for(name)
            if mask is None:
                continue
```
becomes
```
            mask = self._get_mask_tensor_for(name)
            if mask is None:
                continue
```

What follows is more significant and I don't want you to block merge on even optionally, but academically I wanted to write it down because I thought of it.

This also raises the question, could this exclude get bundled into `self._mask` instead? It could for example if you made a `ExcludeMaskFilter` class that implements `HasGetSpatialMask` and wraps what you're currently using for masking. Then SSM doesn't need to know about exclude logic at all.

fme/core/name_and_prefix_matcher.py
mcgibbon: OK, you can keep it as-is. IMO the benefit you're saying is that it's easy to trace to the concrete implementation of what we always pass in practice, and that can be enough. Below is mostly academic.

The reason you might consider a more general regex is if you think the type hint is gratuitous context for its scope. Specifically, the question would be, should StaticSpatialMasking's exclude be restricted specifically to restricting on matching all levels or specific levels of a given name? Does the logic of SSM depend on this, or does SSM just care that certain names get excluded from masking? For example, say we run a combined atmosphere-ocean model and prefix names with `atmos_` or `ocean_`, do we want to disallow excluding all `atmos_<name>_<level>` variables or should it work for that case without editing? How about an exclude that disallows all level 0-3 variables? The flip side of the question is, when we're reading SSM, do we need to understand any of these details to understand SSM? If the type hint is `re.Pattern` or a protocol class `HasMatchMethod`, a reader knows they don't need to, and can more quickly understand the local scope.

But yeah all this is weighed against the fact that if we're always passing this one type it's nice to know how this object is used.

mcgibbon (approved):
LGTM. Had some thoughts I needed to write down but it's good to merge as-is, just the first part of the one nit to consider.

fme/core/spatial_masking.py
jpdunc23: Won't block merge for now, but good to keep in mind as a future improvement.

PR #1227, opened 2026-06-04
Refactor: aggregators expose loss-to-minimize via container attributes

### LINKED ISSUE #1136: Checkpoint-selection metric depends on brittle `"inference/time_mean_norm/rmse/channel_mean"` string contract
`build_trainer` selects the "best inference" checkpoint from a single hardcoded log key:

```python
error = logs.get("inference/time_mean_norm/rmse/channel_mean")
```

That string is assembled across three files — the `"inference"` label is glued on by `inference_one_epoch` in `fme/core/generics/trainer.py`, the `"time_mean_norm"` segment is the dict key chosen in `InferenceEvaluatorAggregatorConfig.build` (`fme/ace/aggregator/inference/main.py`), and `"rmse/channel_mean"` is emitted only when `target=="norm"` inside `TimeMeanEvaluatorAggregator.get_logs` (`fme/ace/aggregator/inference/time_mean.py`) — with no shared constant, no runtime assertion, and no test asserting the round trip. If any of those segments is renamed, made optional, or customized via the user-configurable aggregator, `logs.get(...)` silently returns `None`, the trainer treats the epoch as "no inference ran," and `best_inference_ckpt` stops updating with no warning while wandb continues to display full inference panels. 

A minimal fix is to centralize the key (e.g. as a method on `InferenceEvaluatorAggregatorConfig`), have the callback call that instead of string-keying, raise when the key is missing, and add one integration test that asserts the key the callback expects is actually present in the aggregator's output.

(NB: agent generated text)

mcgibbon:
Decouples trainer code from hardcoded log-dict keys (e.g. ``{name}/mean/loss``, ``{name}/time_mean_norm/rmse/channel_mean``, the downscaling histogram-tail strings) by having each aggregator expose its loss-to-minimize directly. Checkpoint selection no longer breaks when an aggregator renames or removes one of those keys.

Changes:
- `fme.core.generics.aggregator.AggregatorABC` and `InferenceAggregatorABC`: add abstract `get_loss() -> float | None`
- `fme.ace.aggregator.{train.TrainAggregator, one_step.deterministic.OneStepDeterministicAggregator, one_step.main.OneStepAggregator, inference.main.{InferenceEvaluatorAggregator, InferenceAggregator}}` and `fme.coupled.aggregator.{TrainAggregator, OneStepAggregator, InferenceEvaluatorAggregator, InferenceAggregator}`: implement `get_loss`
- `fme.ace.aggregator.inference.main.InferenceEvaluatorAggregator` and `fme.coupled.aggregator.InferenceEvaluatorAggregator`: cache `get_summary_logs()` so `get_loss()` doesn't recompute; coupled copies sub-aggregator dicts before mutating to keep their caches intact
- `fme.core.generics.trainer.{build_validation_callback, build_inference_callback}`: call `aggregator.get_loss()` instead of looking up specific keys
- `fme.ace.train.train` and `fme.coupled.train.train` `validate_stepper`: same
- `fme.downscaling.aggregators.generation.GenerationAggregator`: replace `get_checkpoint_selection_metrics()` with explicit `get_validation_loss()` and `get_histogram_tail_metric()` methods
- `fme.downscaling.train.Trainer`: call those two methods directly; drop now-unused `_best_valid_loss_name` / `_best_histogram_tail_name` strings

- [x] Tests added

Closes #1136


fme/ace/aggregator/inference/main.py
```diff
@@ -611,11 +612,16 @@ def record_initial_condition(
         return logs
 
     def get_summary_logs(self) -> InferenceLog:
-        logs: InferenceLog = {}
-        for name, aggregator in self._summary_aggregators.items():
-            logging.info(f"Getting summary logs for {name} aggregator")
-            logs.update(aggregator.get_logs(label=name))
-        return logs
+        if self._summary_logs_cache is None:
```
mcgibbon: Make sure to invalidate this cache by setting it to None when recording a batch, and also make sure that invalidation is unit tested.

fme/ace/aggregator/inference/test_main.py
```diff
@@ -81,7 +81,7 @@ def test_inference_evaluator_aggregator_channel_mean_names(
 
     agg.record_batch(paired_data)
 
-    summary_logs = agg.get_summary_logs()
+    summary_logs = agg.get_summary().logs
```
mcgibbon: To reduce changes, let's add a helper `get_summary_logs` method that returns `self.get_summary().logs`.

fme/ace/aggregator/one_step/deterministic.py
```diff
@@ -72,26 +76,30 @@ def record_batch(
         if len(batch.gen_data) == 0:
             raise ValueError("No data in gen_data")
 
+        loss = batch.metrics.get("loss", np.nan)
+        self._loss = self._loss + loss
+        self._n_loss_batches += 1
```
mcgibbon: For now, we shouldn't need to add this logic. Previously we just got the loss from the logs under a specific name (I think "val/mean/rmse" or something similar?). While maybe we should do this cleanup later to remove the assumption that a certain aggregator exists, we should keep the behavior as-is for this PR (which is already large).

fme/ace/aggregator/one_step/test_main.py
```diff
@@ -174,7 +174,7 @@ def test_explicit_metrics_build():
             normalize=lambda x: x,
         ),
     )
-    logs = agg.get_logs(label="test")
+    logs = agg.get_summary(label="test").logs
```
mcgibbon: Similarly, add a helper `get_logs` method that returns `self.get_summary(label).logs`, to reduce the size of the diff.

fme/ace/aggregator/train.py
```diff
@@ -130,27 +130,20 @@ def record_batch(self, batch: TrainOutput):
...
-    def get_logs(self, label: str) -> dict[str, torch.Tensor]:
-        """
-        Returns logs as can be reported to WandB.
-
-        Args:
-            label: Label to prepend to all log keys.
-        """
-        logs = {}
+    def get_summary(self, label: str) -> AggregatorSummary:
+        dist = Distributed.get_instance()
```
mcgibbon: If these lines didn't need to be moved/changed from main, put them back how they were to reduce the diff.

fme/core/generics/test_trainer.py
```diff
@@ -1635,83 +1648,107 @@ def _call(tasks, run_validation_side_effect, epoch=1):
...
-                {"a/mean/loss": 0.5},
-                {"b/mean/loss": 999.0},
+                AggregatorSummary(logs={"a/some_metric": 0.5}, loss=0.5),
+                AggregatorSummary(logs={"b/some_metric": 999.0}, loss=999.0),
             ],
         )
         assert loss == pytest.approx(0.5)
-        assert "a/mean/loss" in logs
-        assert "b/mean/loss" in logs
+        assert "a/some_metric" in logs
```
mcgibbon: Can we keep the testing metric as mean/loss to reduce the size of this diff?

fme/ace/train/test_train_config.py
```diff
@@ -397,72 +419,52 @@ def test_epoch_not_in_inference_epochs_returns_empty(self):
...
-            [
-                {"a/time_mean_norm/rmse/channel_mean": 0.3},
-                {"b/time_mean_norm/rmse/channel_mean": 999.0},
-            ],
-        )
+        logs, error = self._call(entries)
         assert error == pytest.approx(0.3)
-        assert "a/time_mean_norm/rmse/channel_mean" in logs
-        assert "b/time_mean_norm/rmse/channel_mean" in logs
+        assert "a/some_metric" in logs
```
mcgibbon: Could we revert this to stay as time_mean_norm/rmse/channel_mean to reduce the diff? Actually, did we need to modify the side-effect being passed?

fme/downscaling/aggregators/generation.py
AnnaKwa, line 315: Since these are no longer provided when the trainer initializes the aggregator, should they just be hard coded attributes?

AnnaKwa (approved):
LGTM, one minor comment about the hard coded metric names.

fme/downscaling/aggregators/generation.py
mcgibbon: On one hand maybe - the benefit would be it's easier to completely change how the selection metric is selected if we removed this unused functionality. But on the other hand, I don't think we'll do that, and they do have the benefit of making it clearer how the checkpoints are selected. I think I'll leave it on this PR.

PR #1230, opened 2026-06-05
Add clip_latent_global_means option to conditional SFNO

mcgibbon:
Adds a `clip_latent_global_means` option to conditional SFNO that bounds the per-channel spatial mean of the post-encoder latent at inference to the range observed during the most recent training epoch. During training the model tracks a per-channel min/max envelope of that mean; during eval, the latent is shifted by `clamp(mean) - mean` so the mean falls within the envelope (a no-op when it already does). The envelope is reset at the start of each training epoch via a new `set_epoch` hook on `TrainStepperABC`. Defaults to False so existing models are unaffected and the existing regression checkpoint still matches.

Changes:
- `fme.core.generics.train_stepper.TrainStepperABC.set_epoch` — new no-op default hook for per-epoch in-module state.
- `fme.core.generics.trainer.Trainer.train_one_epoch` — invokes `stepper.set_epoch` on fresh-epoch boundaries (skipped on mid-epoch resume so partial-epoch state continues to accumulate).
- `fme.ace.stepper.single_module.Stepper.set_epoch` / `TrainStepper.set_epoch` and `fme.coupled.stepper.CoupledStepper.set_epoch` / `CoupledTrainStepper.set_epoch` — walk submodules and call `request_latent_global_mean_envelope_reset` where present, so model components opt in without coupling the stepper to their internals.
- `fme.core.models.conditional_sfno.sfnonet.SFNONetConfig.clip_latent_global_means` and `SphericalFourierNeuralOperatorNet` — buffers `_gm_min` / `_gm_max`, lazy reset flag, distributed reduce of per-batch min/max during training, eval-mode clip-residual shift, and `request_latent_global_mean_envelope_reset`.
- `fme.ace.registry.stochastic_sfno.NoiseConditionedSFNOBuilder.clip_latent_global_means` — exposes the option in the public builder.
- Tests in `fme.core.models.conditional_sfno.test_sfnonet` covering envelope initialization, training-mode tracking, eval-mode shift with a tight envelope, parity with `clip_latent_global_means=False` when the envelope is uninitialized, lazy reset, and a `parallel`-marked test that the envelope is synchronized across data-parallel ranks.

- [x] Tests added


fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -721,6 +755,28 @@ def forward(self, x: torch.Tensor, context: Context):
...
+                    batch_max = global_means.detach().amax(dim=0, keepdim=True)
+                    dist = Distributed.get_instance()
+                    dist.reduce_min(batch_min)
+                    dist.reduce_max(batch_max)
+                    self._gm_min.copy_(torch.minimum(self._gm_min, batch_min))
+                    self._gm_max.copy_(torch.maximum(self._gm_max, batch_max))
+            elif torch.isfinite(self._gm_max).all():
+                clipped = torch.clamp(global_means, min=self._gm_min, max=self._gm_max)
+                # Shift x by the clip residual so its per-channel spatial
+                # mean falls within the envelope observed during training.
```
Arcomano1234: Nit: this is in the weeds a little but `observed during training` isn't exactly true. The clipped values are just during the training of the particular epoch that corresponds to the ckpt being used right?

fme/core/models/conditional_sfno/test_sfnonet.py
```diff
@@ -356,3 +356,141 @@ def test_filter_preserves_global_mean_allows_grad():
...
+        noise=None,
+        embedding_pos=None,
+    )
+    with torch.no_grad():
+        out_clip = model_clip(x, ctx)
+        out_no_clip = model_no_clip(x, ctx)
+    torch.testing.assert_close(out_clip, out_no_clip)
+
+
+def test_clip_latent_global_means_eval_shifts_when_outside_envelope():
```
Arcomano1234, line 455: Nit: This test is a little misleading. All it does is checks if out is finite, which I assume would be true even if the clipping did not work as intended. I suggest making sure things are clipped as expected or remove the test as I think there is quite of bit in terms of overlap between this test and the others added.

fme/core/models/conditional_sfno/sfnonet.py
```diff
@@ -721,6 +755,28 @@ def forward(self, x: torch.Tensor, context: Context):
...
+            global_means = x.mean(dim=(-2, -1), keepdim=True)
+            if self.training:
+                with torch.no_grad():
+                    if self._gm_reset_pending:
+                        self._gm_min.fill_(float("inf"))
+                        self._gm_max.fill_(float("-inf"))
+                        self._gm_reset_pending = False
+                    batch_min = global_means.detach().amin(dim=0, keepdim=True)
+                    batch_max = global_means.detach().amax(dim=0, keepdim=True)
+                    dist = Distributed.get_instance()
```
Arcomano1234, line 768: Question: I am not sure how extensive the testing is for dist stuff, but considering this features relies heavily on getting the global min and max from multiple devices should we add a parallel test for this clipped feature?

fme/core/generics/train_stepper.py
```diff
@@ -65,6 +65,15 @@ def set_eval(self) -> None:
     def set_train(self) -> None:
         pass
 
+    def set_epoch(self, epoch: int) -> None:
```
Arcomano1234, line 68: Nit: claude asked about this should `epoch` be underscored (`_epoch`) as this is an unused parameter right now. On that note, should we even have this `set_epoch` if its only used to restart per-epoch stuff couldn't we call it something different that correctly says that it resets parameters at the start of a new epoch

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Technically it's correct, but I agree it's clearer to specify "during last-epoch training".

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Ah yeah, agreed.

fme/core/generics/train_stepper.py
mcgibbon: We could make it `increment_epoch`, though I'd be a little worried about the possibility it gets double-called. Maybe `increment_epoch(new_epoch)` would be a better fit, along with a validation check that this is in fact the next epoch integer.

Also, this method should be abstract, then the epoch arg will be used.

fme/core/models/conditional_sfno/sfnonet.py
Arcomano1234: I guess I was confused at first before reading the code because I was worried about large gradients at the beginning of training but then realized it was isolated to a single epoch

fme/core/generics/train_stepper.py
Arcomano1234: Yeah I was thinking if this abstracted or had future use its fine to keep as is.

Arcomano1234: Actually Elynn's PR #1260 uses this hook so lets not change it

fme/core/models/conditional_sfno/sfnonet.py
mcgibbon: Let's clarify it.

Arcomano1234 (approved):
Thanks for addressing the comments, this looks good to me.

PR #1232, opened 2026-06-05
Stabilize synthetic stats fixtures used by training tests

### LINKED ISSUE #1143: `fme/coupled/test_train.py::test_train_and_inference[3-True]` flakiness

mcgibbon:
Fixes flakiness in \`fme/coupled/test_train.py::test_train_and_inference[3-True]\` (and the GPU-skipped \`test_train_and_inference_with_derived_forcings\`) by removing a hidden source of pathological inputs in the synthetic stats fixtures.

The integration tests built per-variable normalization stats by drawing \`np.random.randn()\` for both means and stds. Stds occasionally landed near zero or negative, which made normalized inputs blow up. The CRPS + \`NoiseConditionedSFNO\` training case amplified that through multi-step rollouts and produced inference outputs in the \`1e+3\` to \`1e+17\` range and occasional NaN values. Variation between runs came from \`list(set(...))\` of variable names depending on \`PYTHONHASHSEED\`, so the same test would pull a pathological std assignment on some CI invocations but not others. Verified locally on CPU: five runs of the CRPS case with the old behavior produced \`PRESsfc\` magnitudes ranging from \`1e+0\` to \`3e+3\` (one run with std drawn at \`-1.2\`); after the fix all five runs stayed at \`~1e+0\` with no NaN/Inf.

Changes:
- \`fme.ace.testing.fv3gfs_data.get_scalar_dataset\` / \`save_scalar_netcdf\`: replace the \`randn()\` default with \`center + uniform(-0.05, 0.05)\` so per-variable noise stays bounded.
- \`fme.ace.testing.save_stats_netcdfs\`: new helper that writes a paired mean/std file with the std centered at 1.0, so callers cannot forget to keep stds bounded away from zero.
- Route \`StatsData\`, \`fme.coupled.data_loading.test_data_loader.create_coupled_data_on_disk\`, and \`fme.ace.test_train._setup\` through the new helper. Drop the now-redundant \`stats_std_fill_value\` parameter from \`_setup\`.

- [x] Tests added (existing integration tests continue to pass; the fix is on the test fixture itself)

Resolves #1143

### DIFF
--- a/fme/ace/test_train.py
+++ b/fme/ace/test_train.py
@@ -54,7 +54,7 @@ from fme.ace.testing import (
     MonthlyReferenceData,
     patch_cm4_solar_constant,
     save_nd_netcdf,
-    save_scalar_netcdf,
+    save_stats_netcdfs,
     save_stepper_checkpoint,
 )
 from fme.ace.train.train import build_trainer, prepare_directory
@@ -488,7 +488,6 @@ def _setup(
     derived_forcings=None,
     use_schedule: bool = False,
     validate_using_ema: bool = False,
-    stats_std_fill_value: float | None = None,
     multi_validation: bool = False,
     use_variable_masking: bool = False,
 ):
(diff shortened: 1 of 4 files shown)


AnnaKwa (approved):
Thanks for tracking this down!

PR #1234, opened 2026-06-06
Validate expert grid compatibility in DenoisingMoEPredictor.__init__

frodre:
First in a 5-PR stack adding support for longitude domains that cross the 0/360 prime meridian in downscaling. This standalone hardening PR moves expert grid-compatibility validation into the predictor constructor so every construction path is protected, not just the config-build path: only the primary expert's coordinates are used for input prep and output coords, so an expert built on a mismatched grid would otherwise silently downscale onto the wrong grid.

Changes:
- `fme.downscaling.predictors.serial_denoising`: move `_validate_experts_compatible` from `DenoisingMoEConfig.build` into `DenoisingMoEPredictor.__init__`, so it holds for `build`, `from_state`, and future callers (e.g. `with_rolled_lon`).
- `fme.downscaling.test_models`: add `test_denoising_moe_predictor_rejects_mismatched_expert_grids`, constructing the predictor directly with mismatched-grid experts and asserting it raises.
- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

Base: `main`

### Stack

| PR | Head → Base | Title |
|----|-------------|-------|
| [#1234](https://github.com/ai2cm/ace/pull/1234) | `refactor/moe-validate-experts-init` → `main` | Validate expert grid compatibility in `DenoisingMoEPredictor.__init__` |
| [#1235](https://github.com/ai2cm/ace/pull/1235) | `feature/lon-roll-primitives` → PR1 | Add longitude roll primitives |
| [#1236](https://github.com/ai2cm/ace/pull/1236) | `feature/lon-roll-data-layer` → PR2 | Roll seam-crossing longitudes in the data layer |
| [#1237](https://github.com/ai2cm/ace/pull/1237) | `feature/lon-roll-model` → PR3 | Add with_rolled_lon to models |
| [#1238](https://github.com/ai2cm/ace/pull/1238) | `feature/lon-roll-integration` → PR4 | Roll the model in inference/predict/evaluator |

PR #1235, opened 2026-06-06
Add longitude roll primitives for prime-meridian-crossing domains

frodre:
PR 2 of 5 in the prime-meridian longitude stack. Adds the pure coordinate/data rolling utilities needed to re-express a global grid in a seam-crossing domain's convention. These have no production callers yet — later PRs wire them into the data and model layers — so they are reviewable in isolation with full unit coverage. The interval-based roll only triggers when an interval actually crosses the seam (`start < 0` or `stop > 360`), so in-range intervals are a no-op and non-global grids are left untouched.

Primitives overview (PR #1235)

These primitives are always used as a pair: find_roll_anchor (or find_roll_anchor_from_interval) computes the roll amount once; callers pass it to all subsequent roll_lon_coords and roll_lon_data so coordinates and field tensors shift by the same amount.

Two downstream pathways use them:
- Dataset load — rolls each loaded grid into the user's configured lon_extent convention (PR #1236)
- Model setup — rolls the model's fine grid to match the incoming coarse batch's convention (PR #1237)

Changes:
- `fme.downscaling.data.utils`: add `ClosedInterval.finite_values`, `_requires_lon_roll`, `coords_require_lon_roll`, `find_roll_anchor`, `find_roll_anchor_from_interval`, `roll_lon_coords`, `roll_lon_data`, and private helpers `_validate_rollable_lon` and `_validate_monotonic_lon`.
  - `roll_lon_coords` (1-D coordinate tensor) and `roll_lon_data` (N-D field tensor) form a parallel pair: both apply the same roll amount, but `roll_lon_coords` also remaps values to keep the result monotonically increasing, while `roll_lon_data` is a pure cyclic shift. Callers pre-compute the roll amount once via `find_roll_anchor` and pass it to both.
  - `roll_latlon_coords` is not included here; it operates on a `LatLonCoordinates` struct rather than a raw tensor and belongs in the PR that first uses it.
- `fme.downscaling.data` (`__init__`): export the new roll helpers.
- `fme.downscaling.data.test_utils`: unit tests for roll amounts, seam-crossing conventions, round-trip invertibility, non-global/non-uniform rejection, and invalid input validation.
- [x] Tests added

Base: `refactor/moe-validate-experts-init` (PR 1)

### Stack

| PR | Head → Base | Title |
|----|-------------|-------|
| [#1234](https://github.com/ai2cm/ace/pull/1234) | `refactor/moe-validate-experts-init` → `main` | Validate expert grid compatibility in `DenoisingMoEPredictor.__init__` |
| [#1235](https://github.com/ai2cm/ace/pull/1235) | `feature/lon-roll-primitives` → PR1 | Add longitude roll primitives |
| [#1236](https://github.com/ai2cm/ace/pull/1236) | `feature/lon-roll-data-layer` → PR2 | Roll seam-crossing longitudes in the data layer |
| [#1237](https://github.com/ai2cm/ace/pull/1237) | `feature/lon-roll-model` → PR3 | Add with_rolled_lon to models |
| [#1238](https://github.com/ai2cm/ace/pull/1238) | `feature/lon-roll-integration` → PR4 | Roll the model in inference/predict/evaluator |

### DIFF
--- a/fme/downscaling/data/utils.py
+++ b/fme/downscaling/data/utils.py
@@ -157,6 +157,10 @@ def find_roll_anchor(lon_coords: torch.Tensor, anchor: float) -> int:
     anchor is taken mod 360 so negative or >360 values are accepted. The count is taken
     mod n so that a full-wrap (all coords below anchor) reduces to 0, a no-op.
 
+    Callers pre-compute this once and pass it to both :func:`roll_lon_coords` and
+    :func:`roll_lon_data`, keeping coordinates and field tensors aligned without
+    repeating the computation.
+
     Assumes lon_coords are monotonically increasing and cyclic.
     """
     _validate_monotonic_lon(lon_coords)
--- a/fme/downscaling/data/utils.py
+++ b/fme/downscaling/data/utils.py
@@ -70,6 +70,178 @@ class ClosedInterval:
         """
         return coords[self.slice_from(coords)]
 
+    @property
+    def finite_values(self) -> tuple[float, float]:
+        """
+        Return longitude constrained values for this interval,
+        handling infinite endpoints as 0 and 360, respectively.
+        """
+        start = self.start if self.start != -float("inf") else 0.0
+        stop = self.stop if self.stop != float("inf") else 360.0
+        return start, stop
+
+
+def _validate_rollable_lon(lon_coords: torch.Tensor) -> None:
+    """Raise if lon_coords cannot be rolled across the prime meridian.
+
+    Rolling adds 360 to the coordinates that wrap around the end of the array,
+    which yields a monotonic, contiguous result only if the grid extends the whole
+    globe: it must be uniformly spaced and the gap from the last point back to
+    the first must equal that spacing (i.e. the grid reaches the 360° wrap
+    point). This is the precondition the wrap arithmetic assumes.
+
+    The wrap-gap test ``lon[0] + 360 - lon[-1] == spacing`` is convention
+    independent: for any global grid ``lon[-1] - lon[0] == 360 - spacing``,
+    whether expressed as [0, 360), [-180, 180), or any offset.
+    """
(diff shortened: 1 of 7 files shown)


fme/downscaling/data/utils.py
```diff
@@ -70,6 +70,180 @@ def subset_of(self, coords: torch.Tensor) -> torch.Tensor:
         """
         return coords[self.slice_from(coords)]
 
+    @property
```
AnnaKwa, line 73: Can this functionality go into `fme.core` so that in the future ACE code can use it to? IIRC it also cannot handle saving windows across the prime meridian?

fme/downscaling/data/utils.py
```diff
@@ -70,6 +70,180 @@ def subset_of(self, coords: torch.Tensor) -> torch.Tensor:
...
+        return lon_coords
+    _validate_rollable_lon(lon_coords)
+    n = len(lon_coords)
+    rolled = torch.roll(lon_coords, -roll_amount).clone()
+    rolled[n - roll_amount :] += 360.0
+    period_offset = lon_start - (lon_start % 360.0)
+    return rolled + period_offset
+
+
+def roll_lon_data(
```
AnnaKwa: Nit: the name sounds like it does the action 'roll longitude data' which could be confused with rolling the lon coords. Something like `roll_data_along_lon` would sound more distinct.

AnnaKwa (approved):
LGTM, a couple comments.

fme/downscaling/data/utils.py
frodre: Yes, I think that will be the plan once we move to integrate this into the dataset writing for ACE.  `ClosedInterval` would also have to be moved, so I think this should go in a separate PR instead of partially moving things and then cleaning up later.

PR #1237, opened 2026-06-06
Add with_rolled_lon to downscaling models

frodre:
PR 4 of 5 in the prime-meridian longitude stack (PRs 1–3 now merged to main). Lets a model re-express its grid in a seam-crossing coarse domain's longitude convention while sharing the trained network weights, so a single checkpoint can generate over a domain expressed west of 0 or east of 360.

Changes:
- `fme.downscaling.models.DiffusionModel.with_rolled_lon`: rebuild the model through its constructor with `full_fine_coords` and `static_inputs` rolled to match the coarse grid, anchored on the western coarse-cell edge so the fine grid stays aligned to whole coarse cells; returns `self` when no roll is needed. Inference-only (rebuilding re-wraps the module under torch distributed).
- `fme.downscaling.predictors.serial_denoising.DenoisingMoEPredictor.with_rolled_lon`: roll every expert (preserving the shared-grid invariant) and rebuild so the sigma dispatcher is reconstructed from the rolled experts.
- `fme.downscaling.data` exports `roll_lon_coords` for the model layer.
- `fme.downscaling.test_models`: tests for no-roll passthrough, coord shifting with shared weights (including value-level checks that coords and static data roll together, and that a double roll is a no-op), and coarse-cell alignment for a seam-crossing domain. MoE rolling tests live in `test_serial_denoising` next to the existing grid-validation test.
- Test cleanup: shared `cell_centered_coordinate` helper in `test_utils` replaces per-file midpoint-coordinate constructions (`test_models`, `test_config`); removed a test and helper in `test_models`/`test_serial_denoising` duplicated from #1234.
- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

Base: `main` (PRs 1–3 of the stack merged)

### Stack

| PR | Head → Base | Title | Status |
|----|-------------|-------|--------|
| [#1234](https://github.com/ai2cm/ace/pull/1234) | `refactor/moe-validate-experts-init` → `main` | Validate expert grid compatibility in `DenoisingMoEPredictor.__init__` | merged |
| [#1235](https://github.com/ai2cm/ace/pull/1235) | `feature/lon-roll-primitives` → `main` | Add longitude roll primitives | merged |
| [#1236](https://github.com/ai2cm/ace/pull/1236) | `feature/lon-roll-data-layer` → `main` | Roll seam-crossing longitudes in the data layer | merged |
[...]

PR #1236, opened 2026-06-06
Roll seam-crossing longitudes in the downscaling data layer

frodre:
PR 3 of 5 in the prime-meridian longitude stack. Applies the roll primitives (PR 2) in the data layer so a longitude interval that crosses the 0/360 seam can be subset instead of raising `NotImplementedError`. In-range intervals resolve to a zero roll and behave exactly as before.

Changes:
- `fme.downscaling.data.datasets.HorizontalSubsetDataset`: roll data and coordinates into the requested interval's convention rather than raising on wraparound.
- `fme.downscaling.data.config`: extract `_build_aligned_subset_pair`, which rolls coarse and fine lon coords into the extent's convention (`_roll_lons_to_extent_convention`) before `adjust_fine_coord_range`, so fine/coarse subselection stays aligned across the seam.
- `fme.downscaling.data.static.StaticInputs.roll`: roll static fields and their lon coordinates to match.
- `fme.downscaling.data.test_config`, `fme.downscaling.data.test_datasets`, `fme.downscaling.data.test_static`: tests for seam-crossing subsetting (negative and >360 conventions), fine/coarse scale-factor preservation across the seam (even and odd downscale factors), end-to-end paired loader with a seam-crossing extent, and `StaticInputs.roll`.

Note: surfacing the coarse grid convention on `GriddedData`/`PairedGriddedData` (`coarse_latlon_coords`) was deferred to the integration PR after review discussion.

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

Base: `feature/lon-roll-primitives` (PR 2)

### Stack

| PR | Head → Base | Title |
|----|-------------|-------|
| [#1234](https://github.com/ai2cm/ace/pull/1234) | `refactor/moe-validate-experts-init` → `main` | Validate expert grid compatibility in `DenoisingMoEPredictor.__init__` |
| [#1235](https://github.com/ai2cm/ace/pull/1235) | `feature/lon-roll-primitives` → PR1 | Add longitude roll primitives |
| [#1236](https://github.com/ai2cm/ace/pull/1236) | `feature/lon-roll-data-layer` → PR2 | Roll seam-crossing longitudes in the data layer |
| [#1237](https://github.com/ai2cm/ace/pull/1237) | `feature/lon-roll-model` → PR3 | Add with_rolled_lon to models |
[...]

PR #1238, opened 2026-06-06
Roll the model to the coarse grid in inference, predict, and evaluator

frodre:
Final PR (5 of 5) in the prime-meridian longitude stack — turns the feature on end to end. **Stack fully merged; this targets `main` with a clean 10-file diff** (see the table below for the merged history). Each generation entry point reads its coarse domain's longitude grid and rolls the model via `with_rolled_lon` (from #1237) so the generated fine grid lands in the coarse domain's convention. The roll is applied unconditionally and is a no-op for in-range domains (the seam-crossing check lives inside `with_rolled_lon`), so a single code path covers both the crossing and in-range cases.

Changes:
- `fme.downscaling.inference.inference.Downscaler`: build the model once from the coarse coords (rolling up front) instead of lazily from the first batch. Note that `input_shape` is now derived from `output.data.coarse_extent_latlon_coords` rather than the per-batch `horizontal_shape` — these describe the same extent (see `_get_generation_model` docstring).
- `fme.downscaling.predict` (`Downscaler`, `EventDownscaler`): replace the `generation_model` property with `_get_generation_model`, which rolls the model before optional patch wrapping.
- `fme.downscaling.evaluator.EvaluatorConfig`: roll in both the default and event build paths.
- Entry points call `with_rolled_lon` unconditionally and rely on its internal no-op rather than gating on `coords_require_lon_roll`, keeping a single source of truth for the roll decision.
- Data layer: `GriddedData`, `PairedGriddedData`, and `SliceWorkItemGriddedData` now carry `coarse_extent_latlon_coords` (the subset coarse coords for the generated region), populated in their builders; `BatchItemDatasetAdapter` exposes `latlon_coordinates`. This is the longitude grid each entry point passes to `with_rolled_lon`.
- Tests: `test_predict`/`test_evaluator` cover the roll end to end on real seam-crossing data (no mocks) — `test_predictor_runs_seam_crossing` (parametrized over `Downscaler` and `EventDownscaler`) and `test_evaluator_rolls_for_seam_crossing`. The in-range no-op is covered by the model-level `with_rolled_lon` unit tests and the existing end-to-end runs.

- [x] Tests added

### Stack (all merged)

| PR | Head → Base | Title | Status |
|----|-------------|-------|--------|
| [#1234](https://github.com/ai2cm/ace/pull/1234) | `refactor/moe-validate-experts-init` → `main` | Validate expert grid compatibility in `DenoisingMoEPredictor.__init__` | ✅ merged |
| [#1235](https://github.com/ai2cm/ace/pull/1235) | `feature/lon-roll-primitives` → `main` | Add longitude roll primitives | ✅ merged |
| [#1236](https://github.com/ai2cm/ace/pull/1236) | `feature/lon-roll-data-layer` → `main` | Roll seam-crossing longitudes in the data layer | ✅ merged |
| [#1237](https://github.com/ai2cm/ace/pull/1237) | `feature/lon-roll-model` → `main` | Add `with_rolled_lon` to models | ✅ merged |
| **#1238 (this PR)** | `feature/lon-roll-integration` → `main` | Roll the model in inference/predict/evaluator | 🔵 open |


frodre:
Test runs of entrypoints:
`predict.py`: [wandb](https://wandb.ai/ai2cm/andrep-downscaling/runs/fvteehtn)
`evaluator.py`: [wandb](https://wandb.ai/ai2cm/andrep-downscaling/runs/joiveabj)
`inference.py`: [wandb](https://wandb.ai/ai2cm/andrep-downscaling/runs/08bbxm92)

fme/downscaling/inference/inference.py
```diff
@@ -59,19 +59,23 @@ def _get_generation_model(
         self,
         input_shape: tuple[int, int],
         output: DownscalingOutput,
+        coarse_lon: torch.Tensor,
```
AnnaKwa: Can this be derived from `output.data.coarse_extent_latlon_coords` instead of being passed in?

fme/downscaling/test_evaluator.py
```diff
@@ -92,6 +99,95 @@ def get_trainer_model_config():
...
+            lon_extent=ClosedInterval(-90.0, 90.0),
+        ),
+        logging=LoggingConfig(),
+    )
+
+    built = config._build_default_evaluator()
+
+    # A real roll moved the model's fine grid into the seam-crossing convention.
+    assert float(built.model.full_fine_coords.lon.min()) < 0.0
+
```
AnnaKwa, line 217: Is it possible to check that the target data being compared in the evaluator is also rolled consistently with the model output?

AnnaKwa (changes requested):
Mostly LGTM, could you have the evaluator check that the target data is also rolled with the model?

fme/downscaling/inference/inference.py
frodre: I collapsed the generate model entry to only take in the output since that can be used to derive everything, more similar to the other entrypoint adjustments

frodre:
Pre-review agent: @AnnaKwa your `CHANGES_REQUESTED` review (on `843720209`) asked that the evaluator confirm the target data is also rolled with the model. That's addressed in the two most recent commits:

- `c83fa464a` — extends `test_evaluator_rolls_for_seam_crossing` to pull a batch and assert the target's fine coords also land in the seam-crossing convention.
- `747f889c0` — full end-to-end run under the evaluator, asserting the model's per-batch fine grid (`get_fine_coords_for_batch`) is `allclose` to the target's fine coords, so predictions and targets are compared in the same longitude convention (the model roll and the data-layer roll are independent paths, and this confirms they agree).

The delta since your review is `843720209..747f889c0`, so a re-review can focus there rather than the whole PR.

fme/downscaling/test_evaluator.py
frodre: Updated to an end-to-end test of seam rolling, since that hadn't been done and evaluator has the most coverage (target and gen fields on coare/fine grid)

AnnaKwa (approved):
Thanks for the changes, LGTM! Comment that does not need review: how long does the new end-to-end test take? Should it be skipped for `--very-fast` tests?

frodre:
It only takes about 1 second to run, so I think we are fine leaving it in the fast tests.

PR #1240, opened 2026-06-08
Allow for variable renaming in coupled modeling 

Arcomano1234:
Currently, we are attempting to mix and match an atmospheric model trained on ERA5 and an ocean model trained on UFS replay dataset. The problem is in ERA5 wind stresses are called `eastward_surface_stress` while in CM4 and UFS replay they are called `eastward_surface_wind_stress`. Given that we will not be using the atmospheric variables to train a new ACE version for the foreseeable future we need a way to allow variable renaming. Similar to how it was done for HiRO-ACE in the downscaling model, this PR adds an optional stepper arg `atmosphere_output_rename` that will remap the user specified atmospheric variables to be renamed when the ocean model is grabbing forcing variables. Can also be used during inference and standalone eval. 

Changes:
- add `atmosphere_output_rename` to coupled stepper that can rename atmospheric forcing variables for the ocean model

- [x] Tests added


oliverwm1:
I think this would be a reasonable feature to support. Although in my experience there are always a lot of tricky edge cases. Previously we had support for renaming variables within data loading, but we removed this due to the complexity of supported merged datasets: https://github.com/ai2cm/full-model/pull/1511. Keeping it confined to the coupled stepper for now at least limits surface area.

PR #1241, opened 2026-06-08
Allow training on a broader variable pool than inference uses

mcgibbon:
Extracts the "training supports a broader variable pool than inference" piece from `feature/combine-infill-prediction` so it can land independently of the multi-task training work, which will follow in a separate PR.

Changes:
- `fme.core.step.step.StepConfigABC.all_training_names`: new abstract property; returning `None` (the default added on each existing subclass) preserves the old `input_names + output_names` behavior, while returning an explicit list tells the data loader to fetch the broader training set.
- `fme.ace.stepper.single_module.StepperConfig.all_names`: returns `step.all_training_names` when set, falls back to the previous union otherwise.
- `fme.core.step.infill_prediction.InfillPredictionStepConfig` / `InfillPredictionStep` / `InferenceSchemeConfig`: first consumer. Training pulls all `all_names` variables, missing ones are zero-filled, and per-channel mask indicators are appended so the network knows which inputs are real. At inference the step routes via `inference_scheme` and behaves like `SingleModuleStep`. Registered as `"infill_prediction"` and re-exported from `fme.ace`.

- [x] Tests added (`fme/core/step/test_infill_prediction.py`)

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -118,10 +118,12 @@ from fme.core.registry.module import ModuleSelector
 from fme.core.scheduler import SchedulerConfig, SequentialSchedulerConfig
 from fme.core.spatial_masking import StaticSpatialMaskingConfig
 from fme.core.step import (
+    InfillPredictionStepConfig,
     MultiCallStepConfig,
     SeparateRadiationStepConfig,
     SingleModuleStepConfig,
 )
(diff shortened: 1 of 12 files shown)


mcgibbon:
Tried splitting this from https://github.com/ai2cm/ace/pull/1224 but it doesn't make much sense on its own. Closing.

PR #1244, opened 2026-06-09
thread global mean removal state through forward/inverse explicitly

mcgibbon:
Stacked on top of #1245.

`GlobalMeanRemoval` used to cache per-step state (offsets / shifts and the normalized extras dict) on the instance between `forward_transform` and `inverse_transform` / `extras_normalized`. Correctness then depended on call order — interleaved or skipped calls silently produced wrong outputs.

This PR makes the transform stateless: `forward_transform` returns an opaque `GlobalMeanRemovalState` that callers thread through to `inverse_transform` and `extras_normalized`. The instance carries no per-step state; the `RuntimeError("called before forward_transform")` guard is no longer needed because the contract is enforced by the type signature.

Changes:
- `fme.core.step.global_mean_removal`: add `GlobalMeanRemovalState` dataclass; `forward_transform` returns `(TensorDict, state)`; `inverse_transform` and `extras_normalized` take `state`. Remove the per-instance cached attributes.
- `fme.core.step.single_module.step_with_adjustments`: capture state from `forward_transform` and pass it back to `extras_normalized` and `inverse_transform`.

- [x] Tests added — interleaved forward/forward/inverse/inverse round-trips for both shared and per-channel variants verify state-threading is order-independent.

### DIFF
--- a/fme/core/step/global_mean_removal.py
+++ b/fme/core/step/global_mean_removal.py
@@ -45,26 +45,45 @@ def extra_channel_source_field(name: str) -> str | None:
     return None
 
 
+@dataclasses.dataclass
+class GlobalMeanRemovalState:
+    """Opaque token produced by ``forward_transform`` and consumed by the
+    same ``GlobalMeanRemoval`` instance via ``inverse_transform`` and
+    ``extras_normalized``.
+
+    Callers should thread this value from ``forward_transform`` through
+    to those methods, but not read its fields directly — layout is an
+    implementation detail.
+    """
+
+    shifts: dict[str, torch.Tensor]
+    extras: TensorDict
+
+
 class GlobalMeanRemoval(abc.ABC):
     """Removes global means from fields before normalization and restores
     them after denormalization.
 
     ``forward_transform`` shifts input fields toward their climatological
-    means.  ``inverse_transform`` reverses that shift on output fields.
+    means and returns a ``GlobalMeanRemovalState`` that must be passed
+    back to ``inverse_transform`` to reverse the shift on output fields.
+    Threading the state explicitly (rather than caching it on the
+    instance) makes the transform stateless and order-independent.
+
+    Optional synthetic input channels (when ``append_as_input=True``) are
+    produced by ``forward_transform`` already in *normalized* space and
+    exposed via ``extras_normalized(state)`` as a name -> tensor mapping
+    keyed by ``extra_channel_names``.  Callers append the sentinel names
+    to their input packer's name list and merge the dict into the
+    normalized-input dict, so the extras flow through packing and the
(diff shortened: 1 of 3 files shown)


fme/core/step/global_mean_removal.py
```diff
@@ -210,24 +233,26 @@ def forward_transform(
             t = result[name]
             broadcast = offset.view(offset.shape[0], *(1,) * (t.ndim - 1))
             result[name] = t + broadcast
-        return result
 
-    def inverse_transform(self, output: TensorDict) -> TensorDict:
-        if self._cached_offset is None:
-            raise RuntimeError("inverse_transform() called before forward_transform().")
```
Arcomano1234, line 229: Question: It seems we loss this capability to catch this error. I did not review the original PR in great detail but I assume that was just a pretty unlikely safe guard to hit correct?

fme/core/step/single_module.py
```diff
@@ -544,24 +545,25 @@ def step_with_adjustments(
...
-        network_input: TensorMapping = global_mean_removal.forward_transform(
+        network_input, gmr_state = global_mean_removal.forward_transform(
             input, data_mask
         )
     else:
-        network_input = input
+        network_input = dict(input)
     input_norm = normalizer.normalize(network_input)
-    if global_mean_removal is not None:
+    if global_mean_removal is not None and gmr_state is not None:
```
Arcomano1234: Question: Isn't this check of `gmr_state is not None` redundant as  `forward_transform` returns a `gmr_state` that is not None? If so I suggest we remove this part of the check.

Arcomano1234 (approved):
This should be ready to merge. The comments/ questions I left do not seem too important to hold up this PR from being merge, but if you agree with them it might be good to implement the suggestions.

fme/core/step/global_mean_removal.py
mcgibbon: With the new design, this is no longer required, and thus shouldn't produce an error. It was disallowed before because internal state got cached during forward transform and then used during inverse transform. Now that state is returned and then passed explicitly. The corresponding check would be something like `assert state is not None`.

fme/core/step/single_module.py
mcgibbon: Mypy may require this for type checking, because it can't tell this without some more serious parsing of the code logic. However, is there a way to avoid this by restructuring the code? E.g. if we put the `normalizer.normalize` call inside the `if global_mean_removal is not None` and `else` blocks, I think we don't need this `gmr_state` check.

Arcomano1234: Sorry I missed this reply but its not worth refactoring just to remove an check like this.

PR #1245, opened 2026-06-09
treat global mean removal extras as ordinary input channels

mcgibbon:
The synthetic input channels produced by `GlobalMeanRemoval` used to be concatenated separately in `network_call`, which meant the channel-mask machinery did not produce mask channels for them — when both `include_channel_mask_inputs` and `append_as_input` were enabled, the extras silently lacked their accompanying mask channels.

This PR makes each `GlobalMeanRemoval` expose the synthetic channels as named tensors in normalized space: `extra_channel_names` returns sentinel names (prefixed `__gmr_extra__`) and `extras_normalized()` returns `{name: tensor}` after `forward_transform`. The stepper extends its `in_packer` with the sentinel names, `step_with_adjustments` merges the extras into the normalized input dict, and `network_call` no longer mentions GMR. The channel-mask path produces mask channels for the extras: each sentinel inherits its source field's mask (resolved via the `__gmr_extra__` prefix), so a masked source field gets a 0-valued extra paired with a 0-valued mask channel rather than a contradictory 1.

The cached `forward_transform` / `inverse_transform` contract (state is stored on the instance) is unchanged in this PR. The follow-up PR #1244 makes that state explicit / order-independent.

**Backwards compatibility:** Checkpoints trained with **both** `include_channel_mask_inputs=True` and a GMR with `append_as_input=True` will no longer load — the network input channel count goes from `2*n_in + n_extra` to `2*(n_in + n_extra)` (each GMR extra now gets its own mask channel), and the channel ordering changes from `[in | mask(in) | extras]` to `[in | extras | mask(in) | mask(extras)]`. We are accepting this break because we haven't trained many models with this combination. Checkpoints that use only one of the two features are unaffected.

Changes:
- `fme.core.step.global_mean_removal`: add `extra_channel_names` and `extras_normalized()`; drop `get_extra_channels` (replaced by the named dict API). Expose `extra_channel_source_field` so callers can map a sentinel name back to its source field.
- `fme.core.step.single_module.SingleModuleStep.__init__`: extend the input packer with the GMR's `extra_channel_names` so packing and the channel-mask routines treat extras as ordinary input channels.
- `fme.core.step.single_module._build_channel_mask_dict`: resolve `__gmr_extra__<field>` sentinel names back to `<field>` for the `data_mask` lookup so the extra and its mask channel stay consistent on masked samples.
- `fme.core.step.single_module.step_with_adjustments`: merge normalized GMR extras into the input dict before `network_call`.

- [x] Tests added — combined `include_channel_mask_inputs` + GMR extras path, plus a unit test that the GMR extra mask channel inherits the source field's mask.

### DIFF
--- a/fme/core/step/single_module.py
+++ b/fme/core/step/single_module.py
@@ -271,20 +271,24 @@ class SingleModuleStep(StepABC):
             init_weights: Function to initialize the weights of the module.
         """
         super().__init__()
-        n_in_channels = len(config.in_names)
-        if config.include_channel_mask_inputs:
-            n_in_channels *= 2
         if config.global_mean_removal is not None:
             self._global_mean_removal: GlobalMeanRemoval = (
                 config.global_mean_removal.build(
                     normalizer=normalizer, in_names=config.in_names
                 )
             )
-            n_in_channels += self._global_mean_removal.n_extra_input_channels
         else:
             self._global_mean_removal = NoGlobalMeanRemoval()
+        # Synthetic GMR channels are packed alongside real inputs so they
+        # flow through the packer and channel-mask machinery uniformly.
+        packed_in_names = (
+            list(config.in_names) + self._global_mean_removal.extra_channel_names
+        )
+        n_in_channels = len(packed_in_names)
+        if config.include_channel_mask_inputs:
+            n_in_channels *= 2
         n_out_channels = len(config.out_names)
-        self.in_packer = Packer(config.in_names)
+        self.in_packer = Packer(packed_in_names)
         self.out_packer = Packer(config.out_names)
         self._normalizer = normalizer
         if config.ocean is not None:
@@ -379,15 +383,12 @@ class SingleModuleStep(StepABC):
             input_tensor = self.in_packer.pack(input_norm, axis=self.CHANNEL_DIM)
             if self._config.include_channel_mask_inputs:
                 mask_dict = _build_channel_mask_dict(
-                    self.in_names, args.data_mask, input_tensor
+                    self.in_packer.names, args.data_mask, input_tensor
                 )
                 mask_tensor = self.in_packer.pack(mask_dict, axis=self.CHANNEL_DIM)
                 input_tensor = torch.cat(
                     [input_tensor, mask_tensor], dim=self.CHANNEL_DIM
                 )
-            extra = self._global_mean_removal.get_extra_channels()
-            if extra is not None:
-                input_tensor = torch.cat([input_tensor, extra], dim=self.CHANNEL_DIM)
             output_tensor = self.module.wrap_module(wrapper)(
                 input_tensor,
                 labels=args.labels,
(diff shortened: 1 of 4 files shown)


fme/core/step/single_module.py
```diff
@@ -379,15 +383,12 @@ def network_call(input_norm: TensorDict) -> TensorDict:
             input_tensor = self.in_packer.pack(input_norm, axis=self.CHANNEL_DIM)
             if self._config.include_channel_mask_inputs:
                 mask_dict = _build_channel_mask_dict(
-                    self.in_names, args.data_mask, input_tensor
+                    self.in_packer.names, args.data_mask, input_tensor
```
Arcomano1234, line 387: Nit / Question. I believe `_build_channel_mask_dict` misses variables with `__gmr_extra__ ` type patterns and the model sees a channel masked out (e.g., __gmr_extra__a will be all zeroes but the mask channel for __gmr_extra__a is all ones not zeroes like it should be). Claude had this to say: 

This PR fixes the old bug where GMR extras had no mask channels at all (by passing `in_packer.names` into `_build_channel_mask_dict`). However, `_build_channel_mask_dict` looks up masks only by exact key in `data_mask`, which uses real field names (`"a"`), not sentinel names (`"__gmr_extra__a"`).
For a masked sample on field `"a"` with per-channel GMR + `append_as_input=True`:
- `"a"`: input zeroed via `_apply_input_mask`, mask channel = **0.0** ✓
- `"__gmr_extra__a"`: extra value = **0.0** (GMR `forward_transform` handles masking correctly), mask channel = **1.0** (fallback here) ✗
So the network gets contradictory signals on masked samples: the extra value says "no anomaly" but the mask says "present." If mask=0 for masked channels is core to this methodology, this is a blocker for the combined `include_channel_mask_inputs` + `append_as_input` setting this PR targets.
**Suggested fix:** in `_build_channel_mask_dict`, for names with the `__gmr_extra__` prefix, look up the suffix field in `data_mask` (e.g. `"__gmr_extra__a"` → `"a"`). Add a step-level test asserting the GMR extra mask is 0 when the source field is masked.

Arcomano1234 (commented):
Left a big comment / question regarding if this PR actually fixes the bug that was previously present. If I recall correctly on how we are doing the masking I do think that there is a bug. I just want to make sure this is either resolved or the way its currently implemented does not actually effect how we typically do the masking

fme/core/step/single_module.py
mcgibbon: Claude: This was addressed in the follow-up commit 992f59f7 ("share source field's mask with GMR extra channels"). `_build_channel_mask_dict` now strips the `__gmr_extra__` prefix and looks up the source field in `data_mask`, so `__gmr_extra__a`'s mask channel inherits `a`'s mask — 0 when the source is masked, matching the 0-valued extra produced by `forward_transform`. New unit test `test_build_channel_mask_dict_gmr_extra_inherits_source_mask` in `fme/core/step/test_step.py` locks this in.

Arcomano1234 (approved):
Thanks for fixing the bug. This LGTM!

PR #1246, opened 2026-06-09
Feature: Variable Masking

yyexela:
Adds training-time input channel dropout (variable masking) to `SingleModuleStep`.

The motivation is to study how randomly withholding input channels during training affects model robustness and generalization — particularly whether a model trained with masking can perform inference when some inputs are unavailable.

The masking is sampled in the ensemble-aware training layer and applied as a concrete, late input mask inside the Step. This keeps the Step ensemble-agnostic (no `n_ensemble` is threaded into `StepArgs`/`Step`, resolving the PR #1246 review comments at `args.py:43` and `single_module.py:1066`). Synthetic dropout is kept separate from the real `data_mask` (which marks genuinely-absent variables consumed by preprocessing and loss masking): predictions made from dropped inputs are still scored by the loss; only truly-absent target variables are excluded.

Changes:
- `fme.core.var_masking.UniformMaskingConfig` / `PerVariableMaskingConfig`: two masking strategies — uniform count-based and per-channel Bernoulli, selectable via a `kind` discriminator. `sample_mask` returns a plain `[batch, n_channels]` mask (no ensemble awareness).
- `fme.core.step.SingleModuleStepConfig.input_dropout`: new optional field; when set, randomly zeros a subset of packed input channels per sample during training (no-op at eval time). Added only to `SingleModuleStepConfig`; sibling configs (`FCN3StepConfig`, `SecondaryModuleStepConfig`, `MultiCallStepConfig`, `SeparateRadiationStepConfig`) are untouched. The config is serialized but inert at inference.
- `fme.core.step.args.StepArgs.input_dropout_mask`: synthetic training-only input presence mask, keyed by the Step's packed input channel names; preserved through input processing and applied late inside `network_call` (after GMR/normalization), separate from `data_mask`.
- `fme.core.step.StepABC.make_input_dropout_mask` (mode-guarded sampling hook) and `has_input_dropout` (non-random introspection); delegated through `MultiCallStep` and `fme.ace.stepper.Stepper`. `SingleModuleStep` samples over `in_packer.names` (incl. GMR extra channels, which remain independently maskable).
- `fme.ace.stepper.TrainStepper._accumulate_loss`: samples the mask on the pre-broadcast batch via the hook, repeat-interleaves it to the folded ensemble batch (ensemble members of a base sample share one mask), and passes it through `predict_generator`.
- `fme.coupled.CoupledTrainStepper`: raises a clear error if any component step configures `input_dropout` (unsupported for coupled training — fail loud rather than silently no-op).
- `fme.core.distributed.Distributed.broadcast_spatial`: new primitive; the sampling hook broadcasts the mask across the spatial/model-parallel group so all tiles of a sample agree (identity for non-distributed and data-parallel backends).

**Example configs:**

Bernoulli per-variable masking at 5% rate:
```yaml
stepper:
  input_dropout:
    kind: per_variable
    rate: 0.05
```

Uniform masking of up to 3 channels:
```yaml
stepper:
  input_dropout:
    kind: uniform
    max_masked_vars: 3
```

**Behavior notes:**
- Dropout is now **per-window** (sampled once per rollout, held across forward steps), matching inference reality and the real absent-variable `data_mask` — was per-step.
- `UniformMaskingConfig` minimum masked count is now **0** (`min_vars` dropped, `max_vars` renamed `max_masked_vars`); a window may have no channels dropped. This is a training-only config with no inference back-compat concern, so no alias/deprecation shims were added.
- Channel-mask indicator channels (`include_channel_mask_inputs`) are not maskable; an indicator reflects its data channel's AND-combined `real & synthetic` presence.
- Synthetic dropout no longer uses `data_mask`; only real missing-variable masks exclude targets from the loss.

- [x] Tests added


fme/core/var_masking.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Masks a uniformly sampled count of channels per sample.
+
+    Parameters:
+        min_vars: Minimum number of channels to mask. "min" resolves to 0.
+        max_vars: Maximum number of channels to mask. "max" resolves to the
+            total number of input channels.
+    """
+
+    min_vars: int | str = "min"
```
mcgibbon: I think 1 is a more reasonable default. Also, no need for `"min"` when we have `0`.
```suggestion
    min_vars: int = 1
```

fme/core/var_masking.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Training-time input channel dropout config. Exactly one of ``uniform`` or
+    ``per_variable`` must be set.
+
+    Parameters:
+        uniform: Uniform count-based masking config.
+        per_variable: Independent per-channel Bernoulli masking config.
+    """
+
+    uniform: UniformMaskingConfig | None = None
```
mcgibbon: Issue: this configuration typing doesn't imply only one type is allowed.

Suggestion: Put a `kind: Literal["uniform"] = "uniform"` attribute on UniformMaskingConfig, and then use it directly as the type. Similarly put a `kind: Literal["per_variable"] = "per_variable"` on the PerVariableMaskingConfig.

This will have a secondary benefit that if for some reason we want to remove support for one of these masking types, it will only invalidate checkpoints that use that type, instead of all checkpoints trained with that code.

fme/core/step/test_step.py
```diff
@@ -1524,3 +1529,266 @@ def test_step_shared_global_mean_removal_raises_on_masked_reference():
...
+    Outputs with rate=1.0 must differ from rate=0.0 with the same weights,
+    confirming that zeroing both input channels and their mask indicators
+    has an effect.
+    """
+    step_full = _make_input_dropout_step(
+        VariableMaskingConfig(per_variable=PerVariableMaskingConfig(rate=1.0))
+    )
+    step_none = _make_input_dropout_step(
+        VariableMaskingConfig(per_variable=PerVariableMaskingConfig(rate=0.0))
+    )
```
mcgibbon, line 1639: Claude: dead code — `step_full` and `step_none` are reassigned at lines 1658-1659 from `_make_with_mask(...)` before being used. Drop these two assignments.

fme/core/step/test_step.py
```diff
@@ -1524,3 +1529,266 @@ def test_step_shared_global_mean_removal_raises_on_masked_reference():
...
+    out_none, _ = step_none.step(
+        args=StepArgs(input=input_data, next_step_input_data=next_step, labels=None)
+    )
+    differs = any(
+        not torch.allclose(out_full[name], out_none[name]) for name in out_names
+    )
+    assert (
+        differs
+    ), "rate=1.0 should zero inputs and mask indicators, producing different outputs"
```
mcgibbon: Claude: this test (and `test_input_dropout_with_channel_mask_inputs_zeroes_gmr_indicators` below) does not actually isolate the mask-indicator zeroing branch. With `data_mask=None`, `_build_channel_mask_dict` returns all-ones regardless, so the input-channel zeroing alone is sufficient to make rate=1.0 differ from rate=0.0 — the assertion would still pass if the `mask_tensor = mask_tensor * channel_mask...` line were deleted. Either pass a non-trivial `data_mask`, inspect the packed network input directly, or trim the docstring to not overclaim.

mcgibbon:
Claude: PR description is misleading — it credits this PR with the GMR stateless refactor and `PerChannelGlobalMeanRemoval` additions, but those land via the stacked base PR #1244 (`refactor/global-mean-removal-stateless`). This PR only adds variable masking, a one-line docstring on `n_extra_input_channels`, and a `gmr_state is not None` assert tightening in `step_with_adjustments`. A reviewer reading the description will look for GMR code that is not here. Also worth noting that `input_dropout` is added only to `SingleModuleStepConfig`; siblings (`FCN3StepConfig`, `SecondaryModuleStepConfig`, `MultiCallStepConfig`, `SeparateRadiationStepConfig`) are intentionally untouched — a one-line note would save the reviewer asking.

mcgibbon:
Re the PR description, note I'll merge that branch into main before this PR gets merged, and then this PR will get rebased from main and merged. So the PR description should only be for the part on this branch.

fme/core/var_masking.py
yyexela: Done — changed to `min_vars: int = 1` and removed the `"min"` string handling entirely (updated `__post_init__` and `_sample_uniform`).

fme/core/var_masking.py
yyexela: Done — added `kind: Literal["uniform"] = "uniform"` to `UniformMaskingConfig` and `kind: Literal["per_variable"] = "per_variable"` to `PerVariableMaskingConfig`. Moved `sample_mask` onto each sub-config and replaced the `VariableMaskingConfig` wrapper dataclass with a type alias `UniformMaskingConfig | PerVariableMaskingConfig`. Call sites now use the sub-configs directly (e.g. `PerVariableMaskingConfig(rate=0.5)`), and dacite discriminates correctly in strict mode.

fme/core/step/test_step.py
yyexela: Done — removed the two dead assignments.

fme/core/step/test_step.py
yyexela: Done — trimmed the docstrings to not overclaim. Both tests now just assert that rate=1.0 produces different outputs from rate=0.0 (without claiming to isolate the mask-indicator zeroing path). Also renamed `test_input_dropout_with_channel_mask_inputs_zeroes_gmr_indicators` → `test_input_dropout_with_channel_mask_inputs_and_gmr`.

yyexela:
Updated the PR description to only describe what's on this branch: variable masking support in `SingleModuleStep` (the GMR refactor and `PerChannelGlobalMeanRemoval` additions belong to the base PR #1244). Also added a note that `input_dropout` is intentionally added only to `SingleModuleStepConfig`.

mcgibbon:
Will do a full code review once I have the base PRs reviewed and merged, and then this PR merges any of those updates in.

fme/core/var_masking.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Parameters:
+        min_vars: Minimum number of channels to mask. Defaults to 1.
+        max_vars: Maximum number of channels to mask. "max" resolves to the
+            total number of input channels.
+    """
+
+    kind: Literal["uniform"] = "uniform"
+    min_vars: int = 1
+    max_vars: int | str = "max"
```
mcgibbon: Issue (required): When I looked at this code, I immediately filled in that `min_vars` would mean the minimum number of variables we're working with, not the minimum number of masked variables. Also, I kind of expect that we should always be including the option of not masking a variable, meaning min_vars doesn't need to be configurable (and should always be set to zero, not 1).

Suggestion (optional): rename `max_vars` to `max_masked_vars`, and remove `min_vars` (set it to 0 always).

fme/ace/stepper/single_module.py
```diff
@@ -1063,6 +1063,7 @@ def get_prediction_generator(
         forcing_data: BatchData,
         n_forward_steps: int,
         optimizer: OptimizationABC,
+        n_ensemble: int | None = None,
```
mcgibbon: Question: `n_ensemble` is already an attribute on both initial_condition and forcing_data, why does it need to be passed separately? Can we revert the changes to this file?

fme/core/step/args.py
```diff
@@ -35,12 +40,14 @@ def __init__(
         labels: BatchLabels | None = None,
         data_mask: TensorMapping | None = None,
         stepper_state: StepperState | None = None,
+        n_ensemble: int = 1,
```
mcgibbon: Issue: We've been intentionally hiding the ensemble-ness of the input from the Step object up to this point. I hadn't realized that this feature is going to require breaking that contract.

Option 1: Find a way to avoid passing n_ensemble to a step, for example by refactoring the data channel dropout concern into a different location that already has access to the BatchData containing the n_ensemble (hard), or by deciding we're just going to always dropout the channels uniformly across all batch members (easy, but requires scientific re-evaluation at least for your "best" case).

Option 2: If we are going to pass this data, it should be done by passing TensorMapping that contain an ensemble dimension, instead of adding a low-level contract that the batch dimension has a certain striding pattern and this integer indicates its interpretation. This is a fundamental refactor to all of our Step concepts, and I don't love it because it's training-specific - even when we need this in training, it's not something we need during post-training inference.

If using uniform-across-batch dropout isn't feasible, I'd explore whether we can more fundamentally refactor out this step-specific training concern into a layer that knows about ensembles before refactoring all Step TensorMappings to have an ensemble dimension.

mcgibbon: If we go uniform-across-batch, make sure that we coordinate using dist cross-gpu so that the behavior doesn't change depending on GPU decomposition, which requires a @pytest.mark.parallel test.

yyexela:
Another PR doing this merged: https://github.com/ai2cm/ace/pull/1329

PR #1247, opened 2026-06-09
Add spectral_ratio to SpectralConvS2 for bottlenecked spectral filters

mcgibbon:
Adds a `spectral_ratio` knob to the linear spectral filter so the SHT and the per-mode complex weight inside `SpectralConvS2` can operate on a reduced latent channel width while the surrounding network keeps its full `embed_dim`. When `spectral_ratio < 1`, a real `Conv1x1` down-projection is applied before `forward_transform` and an up-projection after `inverse_transform`. Because channel-wise linear maps commute with the spatial SHT (`F(Px) = PF(x)`), this is mathematically equivalent to constraining each per-mode `C × C` complex weight to factor as `Q W'_l P` with shared `P, Q` across modes -- a mode-shared low-rank bottleneck, distinct from the existing per-mode `spectral_lora_rank` factorization. For `embed_dim=512`, `spectral_ratio=0.5` shrinks the per-mode weight tensor ~4× and halves the SHT channel count. Default `spectral_ratio=1.0` is a no-op; no extra parameters are added and existing checkpoints load unchanged.

Changes:
- `fme.core.models.conditional_sfno.s2convolutions.SpectralConvS2`: new `spectral_ratio` arg; `pre_proj` / `post_proj` `Conv1x1`s when `< 1`; internal weight and LoRA tensors scale with `spectral_channels = round(embed_dim * spectral_ratio)`.
- `fme.core.models.conditional_sfno.sfnonet`: thread `spectral_ratio` through `SpectralFilterLayer`, `FourierNeuralOperatorBlock`, `SFNONetConfig`, and `SphericalFourierNeuralOperatorNet`.
- `fme.ace.registry.stochastic_sfno.NoiseConditionedSFNOBuilder`: expose `spectral_ratio` in the YAML-loadable config.
- Config-time validation lives in a single shared helper `validate_spectral_ratio` in `s2convolutions.py`, invoked from `SpectralConvS2.__init__`, `SFNONetConfig.__post_init__`, and `NoiseConditionedSFNOBuilder.__post_init__`. The rules: `spectral_ratio` must be in (0, 1]; `< 1` is rejected with non-`linear` filter types, with `filter_preserves_global_mean` (the l=0 mode is sandwiched between the projections), with `local_blocks` (DISCO blocks have no spectral filter to bottleneck), and when `round(embed_dim * spectral_ratio)` is not divisible by `filter_num_groups`. Each call site still validates its own inputs so its API stays self-validating, while the rules are defined once.
- Tests: shape reduction, forward pass, manual `Q W'_l P` equivalence, default-unchanged backwards compat, validation errors, `filter_residual` interaction, and end-to-end SFNO build/forward/backward at `spectral_ratio = 0.5`.
- Regenerated `NoiseConditionedSFNO_module_config.yaml` cached fixture to include the new key (isolated commit per repo convention).

Notes for review:
- With `filter_residual=True` (or mismatched in/out grids), the SHT round-trip residual that feeds the block's inner skip also passes through `pre_proj`/`post_proj`, so the skip path is bottlenecked as well. This is forced by shapes (the spectral tensor is `spectral_channels` wide at that point) and avoids a second full-width SHT round trip; documented in the config docstrings and covered by a test.
- The legacy modulus SFNO (`fme/ace/models/modulus/s2convolutions.py`, used by the plain `SFNO` registry type) has its own `SpectralConvS2` and is intentionally left unchanged.

- [x] Tests added

## Test plan
- [x] `python -m pytest fme/core/models/conditional_sfno/ fme/ace/registry/ fme/core/registry/` (all pass)
- [x] `pre-commit run --files <changed files>` (ruff, ruff-format, mypy all pass)


Arcomano1234 (commented):
Looks good just have one comment / nit / question. The logic for checking `spectral_ratio`, `filter_type`, `filter_preserves_global_mean`, `local_blocks`, etc are repeated in 3 different locations. I didn't check the logic flow closely but my gut feeling is that some of these checks at lower levels in `fme/core/models/conditional_sfno/s2convolutions.py` would be get caught at the registry or sfnonet.py level and are thus duplicated. If this is true I recommend reducing the redundant checks. If this is not true I then recommend just making a helper function like `_validate_spectral_ratio(embed_dim, ratio, num_groups, ...)` that can be shared among the 3 different files. This would improve readability.

mcgibbon:
> Looks good just have one comment / nit / question. The logic for checking `spectral_ratio`, `filter_type`, `filter_preserves_global_mean`, `local_blocks`, etc are repeated in 3 different locations. I didn't check the logic flow closely but my gut feeling is that some of these checks at lower levels in `fme/core/models/conditional_sfno/s2convolutions.py` would be get caught at the registry or sfnonet.py level and are thus duplicated. If this is true I recommend reducing the redundant checks. If this is not true I then recommend just making a helper function like `_validate_spectral_ratio(embed_dim, ratio, num_groups, ...)` that can be shared among the 3 different files. This would improve readability.

Agreed about deduplicating the code. I do think we should keep the redundant checks even if they won't in practice get called by our critical path, since we have no guarantees APIs won't be used differently in the future. But we can refactor out to a shared helper.

mcgibbon:
Refactored the duplicated validation into a shared `validate_spectral_ratio` helper in `s2convolutions.py` (06d637d). It's invoked from all three sites (`SpectralConvS2.__init__`, `SFNONetConfig.__post_init__`, `NoiseConditionedSFNOBuilder.__post_init__`), so the rules are defined once but each API still validates its own inputs. Kept the per-site calls intentionally since we can't guarantee the lower-level APIs always get reached through the registry/config path.

Arcomano1234 (approved):
LGTM!

PR #1248, opened 2026-06-09
Add Swin Transformer backbone architecture

yyexela:
This PR adds the Swin Transformer backbone architecture to ACE as a drop-in replacement for SFNO via the existing `ModuleSelector` registry.

## New features

- **`fme.core.models.swin_transformer.SwinTransformerNet`** — 2D Swin U-Net backbone adapted from ArchesWeather's 3D design; supports U-Net encoder/decoder, shifted-window attention, column-wise interaction, and optional AdaLN/CLN label conditioning.
- **`fme.core.models.miles_credit.boundary_padding.TensorPadding`** — shared `mirror` and `earth` (cyclic longitude + mirror latitude) boundary padding utility used by the Swin backbone.
- **`fme.ace.registry.swin_transformer.SwinTransformerBuilder` / `NoiseConditionedSwinTransformerBuilder`** — registers Swin Transformer into the `ModuleSelector` registry. `SwinTransformer` and `NoiseConditionedSwinTransformer` added to `CONDITIONAL_BUILDERS` in `fme/core/registry/module.py`.
- **`HorizontalCoordinates.lat_1d`** — new property on `HorizontalCoordinates` (returns `None` by default, overridden on `LatLonCoordinates`) used by the Swin builder for cos-lat CPB scaling.

## Checklist

- [x] Tests added


fme/ace/models/miles_credit/crossformer.py
mcgibbon, line 1: The name of this folder makes me feel like there should be a LICENSE file or header somewhere, should there?

fme/ace/stepper/single_module.py
```diff
@@ -1657,7 +1657,7 @@ def _accumulate_loss(
             n_forward_steps,
             optimization,
             labels=input_ensemble_data.labels,
-            data_mask=input_ensemble_data.data_mask,
+            data_mask=forcing_ensemble_data.data_mask,
```
mcgibbon: Question: Is this valid for all cases? I feel like we likely sometimes (usually?) need input data and sometimes need forcing data.

mcgibbon: Suggestion: If these changes are unrelated to this particular PR (I would think it is, it's not backbone-specific) then revert them so we can discuss them in the context of the PR that needs these changes.

mcgibbon:
Having claude do a pre-review to make this easier.

fme/core/models/swin_transformer/boundary_padding.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Initialize the TensorPadding class with the specified mode and padding sizes.
+
+        Args:
+            mode (str): The padding mode, either 'mirror' or 'earth'.
+            pad_lat (list[int]): Padding sizes [top, bottom] for lat (N-S) dimension.
+            pad_lon (list[int]): Padding sizes [left, right] for lon (W-E) dimension.
+            **kwargs: Additional keyword arguments (ignored).
+        """
+        self.mode = mode
```
mcgibbon, line 63: Claude: An unrecognized `mode` makes `pad()`/`unpad()` fall through both branches and silently return `None`, so a YAML typo surfaces as a confusing `NoneType` error downstream. Suggest validating in `__init__`:
```suggestion
        if mode not in ("mirror", "earth"):
            raise ValueError(f"Unknown padding mode {mode!r}")
        self.mode = mode
```

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,280 @@
...
+    num_heads: list[int] = dataclasses.field(default_factory=lambda: [3, 6, 6, 3])
+    window_size: list[int] = dataclasses.field(default_factory=lambda: [4, 8])
+    mlp_ratio: float = 4.0
+    drop_path_rate: float = 0.2
+    use_skip: bool = True
+    mlp_layer: str = "mlp"
+    embed_dim_scalar: int = 0
+    embed_dim_labels: int = 0
+    cpb_hidden_dim: int = 64
+    padding_conf: dict | None = None
```
mcgibbon: Claude: `padding_conf` is a raw dict forwarded as `TensorPadding(**padding_conf)`: typo'd keys are silently swallowed by `**kwargs`, and a missing `"activate"` key is a bare `KeyError` at build time. A small typed config validated in `__post_init__` would match repo convention. Applies equally to `NoiseConditionedSwinTransformerBuilder` and `CrossFormerConfig`.

fme/ace/step/camulator.py
```diff
@@ -0,0 +1,552 @@
...
+        default_factory=lambda: AtmosphereCorrectorConfig()
+    )
+    next_step_forcing_names: list[str] = dataclasses.field(default_factory=list)
+    prescribed_prognostic_names: list[str] = dataclasses.field(default_factory=list)
+    residual_prediction: bool = False
+    atmosphere_level_start: int = 0
+    global_mean_removal: GlobalMeanRemovalConfigUnion | None = None
+
+    def __post_init__(self):
+        if self.builder.frames != 1:
```
mcgibbon: Claude: This rejects `frames != 1` but not mixed patch sizes: `patch_height=1, patch_width=2` builds a net that crashes at forward time (CrossFormer gates `cube_embedding` on `pw>1 and ph>1 or frames>1` in `__init__` but `pw>1 and ph>1` in `forward`). Consider requiring patch dims to be both 1 or both >1.

fme/ace/step/camulator.py
```diff
@@ -0,0 +1,552 @@
...
+import datetime
+import logging
+from collections.abc import Callable
+from typing import Any, Literal
+
+import dacite
+import torch
+from torch import nn
+
+from fme.ace.models.miles_credit.crossformer import CrossFormer  # type: ignore
```
mcgibbon: Claude: This `# type: ignore` is unnecessary — mypy passes without it.

fme/ace/models/miles_credit/test_crossformer.py
```diff
@@ -0,0 +1,200 @@
...
+        embed_dim_labels=0,
+        embed_dim_noise=noise_embed_dim,
+        embed_dim_pos=0,
+    )
+    model = make_crossformer(img_shape, context_config=context_config)
+    batch_size = 2
+    n_channels = 2 * 2 + 2 + 1
+    x = torch.randn(batch_size, n_channels, 1, *img_shape)
+    # noise shape: (batch, embed_dim_noise, H, W)
+    from fme.core.models.conditional_sfno.layers import Context
```
mcgibbon: Claude: `Context`/`ConditionalLayerNorm` are imported inside four test bodies (here, L121, L130, L163) — lift to module top.

fme/ace/models/miles_credit/boundary_padding.py
```diff
@@ -0,0 +1 @@
+from fme.core.models.boundary_padding import TensorPadding  # noqa: F401
```
mcgibbon: Claude: Nothing in the repo imports this re-export — delete unless it's deliberately kept for diff parity with the exp branch.

fme/core/models/swin_transformer/swin_transformer.py
```diff
@@ -0,0 +1,280 @@
+"""A 2D Swin U-Net backbone for ACE.
+
+This is a 2D adaptation of ArchesWeather's 3D Swin U-Net to ACE's
+``(B, C, H, W)`` interface, where all vertical levels are stacked into the
+channel dimension. See ``swin_transformer.md`` for the design notes.
```
mcgibbon: Claude: `swin_transformer.md` doesn't exist in the repo (also referenced in `swin_layers.py:4`) — commit the design notes or drop the references.

fme/ace/step/camulator.py
```diff
@@ -0,0 +1,552 @@
...
+from fme.core.ocean import Ocean, OceanConfig
+from fme.core.packer import Packer
+from fme.core.registry import CorrectorSelector
+from fme.core.step.args import StepArgs
+from fme.core.step.global_mean_removal import (
+    GlobalMeanRemoval,
+    GlobalMeanRemovalConfigUnion,
+    NoGlobalMeanRemoval,
+)
+from fme.core.step.single_module import _apply_input_mask, step_with_adjustments
```
mcgibbon: Claude: Cross-module import of private `_apply_input_mask` — consider making it public in `fme.core.step.single_module`, since it's now part of that module's interface.

fme/ace/step/camulator.py
```diff
@@ -0,0 +1,552 @@
...
+        for name in self.atmosphere_diagnostic_names:
+            for i in range(level_start, level_start + self.atmosphere_levels):
+                atmosphere_out_names.append(f"{name}_{i}")
+        self.atmosphere_input_names = atmosphere_in_names
+        self.atmosphere_output_names = atmosphere_out_names
+        self.surface_input_names = self.surface_prognostic_names
+        self.surface_output_names = (
+            self.surface_prognostic_names + self.surface_diagnostic_names
+        )
+        # surface before atmosphere to match swin YAML channel ordering
```
mcgibbon: Claude: "to match swin YAML channel ordering" is confusing in a CrossFormer step — which config convention is this matching?

fme/core/registry/module.py
```diff
@@ -47,6 +47,22 @@ def build(
         """
         ...
 
+    def build_for_selector(
```
mcgibbon: Claude: Only the new Swin builders honor `conditional` via this hook; `NoiseConditionedSFNO`/`LocalNet` still enable label conditioning from `dataset_info.all_labels` regardless of the flag, so `conditional` now means different things per builder. If intentional, worth a note in the PR description or a follow-up issue.

fme/core/coordinates.py
```diff
@@ -586,6 +586,10 @@ def meshgrid(self) -> tuple[torch.Tensor, torch.Tensor]:
     def shape(self) -> tuple[int, ...]:
         pass
 
+    @property
+    def lat_1d(self) -> torch.Tensor | None:
```
mcgibbon: Claude: A concrete `None` default on the ABC means HEALPix (and any future coordinate type) silently opts out of cos(lat) CPB scaling. An `@abstractmethod` with explicit per-subclass returns would make that choice visible. (Same abstract-or-final concern applies to `ModuleConfig.build_for_selector`, which is concrete-with-overrides.)

mcgibbon (commented):
Claude: The `single_module.py` data_mask fix + its two tests are independently mergeable — consider splitting into a small PR for a fast merge. (Verified: both new tests fail without the two-line fix and pass with it.) More broadly, the Swin and CrossFormer halves share only `TensorPadding`, if two smaller PRs are preferred.

fme/ace/step/camulator.py
mcgibbon:
```suggestion
from fme.ace.models.miles_credit.crossformer import CrossFormer
```

fme/ace/step/camulator.py
```diff
@@ -0,0 +1,552 @@
...
+            embed_dim_pos=0,
+            n_labels=0,
+            label_embed_dim=0,
+            inverse_sht=None,
+        )
+
+
+@StepSelector.register("CrossFormer")
+@dataclasses.dataclass
+class CrossFormerStepConfig(StepConfigABC):
```
mcgibbon: We should talk about what leads us to need separate step classes for CrossFormer and Camulator, instead of just new network backbones. I'm often adding new features to SingleModuleStep, so these will quickly become out-of-date. If there are small changes we can make to SingleModuleStep so that it supports these model versions, or ways to factor out some of these varying concerns so they can be dependency-injected more easily, we should consider them.

Maybe Claude can take a look and give a summary for me (I'll ask).

fme/core/models/swin_transformer/boundary_padding.py
mcgibbon: Suggestion: Use a Literal["mirror", "earth"] type hint on `mode` in the init to also get static type checking consistency here.

fme/core/registry/module.py
```diff
@@ -47,6 +47,22 @@ def build(
...
+        self,
+        n_in_channels: int,
+        n_out_channels: int,
+        dataset_info: DatasetInfo,
+        conditional: bool,
```
mcgibbon: Issue: `conditional` is unused.

fme/core/registry/module.py
```diff
@@ -191,10 +209,11 @@ def build(
...
         else:
             label_encoding = None
-        module = self._instance.build(
+        module = self._instance.build_for_selector(
             n_in_channels=n_in_channels,
             n_out_channels=n_out_channels,
             dataset_info=dataset_info,
+            conditional=self.conditional,
         )
         return Module(module, label_encoding)
```
mcgibbon, line 201: I do think there's something we might want to clean up here, but for your PR I'd like you to maintain existing behavior.

Conditional models are ones that _support_ label encodings. A conditional model still gets called non-conditionally if it doesn't get passed label encodings. There's no need to pass conditional along to the Step implementation - a single step should handle both conditional and non-conditional calls based on whether it's passed labels.

Of course, it "handling" that call can mean "raise an exception" if the state of the labels doesn't match the network configuration as defined on init.

We can consider a refactor like this as a separate question, but for now please keep it as-is - conditional determines whether the Module wrapper passes along labels, and the step separately decides whether it is or isn't going to raise if passed labels (we could technically make zero-labels if you want when passed no labels in a label-supporting model, but raising might be safer - please do whatever SingleModuleStep does).

fme/core/coordinates.py
mcgibbon: Please update the code so that the swin transformer raises an exception when the coords are passed as None, it seems like we don't want to support the healpix grid for this model.

If we do want this model to work in a no-scaling mode, then please add a configuration option to the transformer that disables that scaling and also the coords-is-not-none requirement (so it would run but not have a silent change in behavior).

mcgibbon (commented):
You can tell which comments are mine because they don't start with "Claude:".

fme/ace/models/miles_credit/crossformer.py
yyexela: Claude: Added `fme/ace/models/miles_credit/LICENSE` with the Apache 2.0 license text provided.

fme/ace/stepper/single_module.py
yyexela: ~Claude: Reverted to `input_ensemble_data.data_mask`. The change was not needed for this PR; the correct choice (IC vs forcing mask) deserves its own discussion with a targeted test.~

fme/core/models/swin_transformer/boundary_padding.py
yyexela: Claude: Done — added `if mode not in ("mirror", "earth"): raise ValueError(...)` in `TensorPadding.__init__`, and also added a `Literal["mirror", "earth"]` type annotation per the follow-up suggestion.

fme/ace/registry/swin_transformer.py
yyexela: Claude: Done — introduced `TensorPaddingConfig` dataclass in `boundary_padding.py` with `__post_init__` validation, and replaced `padding_conf: dict | None` with `padding_conf: TensorPaddingConfig | None` across all four builder/config classes (`SwinTransformerBuilder`, `NoiseConditionedSwinTransformerBuilder`, `CrossFormerConfig`, `NoiseConditionedCrossFormerConfig`). Each class coerces YAML-sourced dicts to `TensorPaddingConfig` in `__post_init__`, so typo'd keys raise `TypeError` and invalid modes raise `ValueError` at config-load time.

fme/ace/step/camulator.py
yyexela: Claude: Done — added a check in `CrossFormerStepConfig.__post_init__` that raises `ValueError` when `patch_height` and `patch_width` are mixed (one is 1, the other > 1). The crash path you identified: when mixed, `__init__` sets `first_dim = dim[0]` (expecting cube_embedding to project), but `cube_embedding` isn't created and `forward` takes the `squeeze(2)` path, so the first `CrossEmbedLayer` gets `input_channels` instead of `dim[0]`.

fme/ace/step/camulator.py
yyexela: Claude: Removed.

fme/ace/models/miles_credit/test_crossformer.py
yyexela: Claude: Done — `Context` and `ConditionalLayerNorm` are now imported at module top alongside `ContextConfig`.

fme/ace/models/miles_credit/boundary_padding.py
yyexela: Claude: Deleted.

fme/core/models/swin_transformer/swin_transformer.py
yyexela: Claude: Removed the `swin_transformer.md` reference from both `swin_transformer.py` and `swin_layers.py`.

fme/ace/step/camulator.py
yyexela: Claude: Renamed `_apply_input_mask` to `apply_input_mask` in `fme/core/step/single_module.py` and updated all call sites (`single_module.py`, `test_step.py`, `test_global_mean_removal.py`, `camulator.py`).

fme/ace/step/camulator.py
yyexela: Claude: Removed — the variable names make the ordering clear without the comment.

fme/core/registry/module.py
yyexela: Claude: Tracked in issue #1258. The `conditional` flag behavior is left as-is in this PR per the guidance at the follow-up comment below.

fme/core/coordinates.py
yyexela: Claude: Updated both Swin builders to raise `ValueError` when `lat_1d` is `None` (either from `MissingDatasetInfo` or a non-lat-lon coordinate type). Reasoning documented in issue #1259.

fme/core/models/swin_transformer/boundary_padding.py
yyexela: Claude: Done — added `Literal["mirror", "earth"]` to both `TensorPaddingConfig.mode` and `TensorPadding.__init__(mode: ...)`.

fme/core/registry/module.py
yyexela: Claude: Acknowledged. The base-class default ignores `conditional` intentionally (it just calls `build()`); the parameter is part of the interface that subclasses override. Leaving as-is per the guidance in the follow-up comment.

fme/core/registry/module.py
yyexela: Claude: Understood — leaving the `conditional` semantics and the Module wrapper behavior unchanged in this PR.

fme/core/coordinates.py
yyexela: Claude: Done — both `SwinTransformerBuilder._build` and `NoiseConditionedSwinTransformerBuilder._build` now raise `ValueError` when `lat_1d` is `None`. Updated the two tests that previously asserted the no-lat-coords case builds successfully to instead assert the `ValueError` is raised.

fme/ace/step/camulator.py
yyexela: Claude: The main differences that drove a separate `CrossFormerStep` rather than plugging into `SingleModuleStep` are: (1) CrossFormer uses a different channel layout — atmosphere channels are the fast (inner) dimension rather than the outer one, so `atmosphere_name_level` ordering differs from `SingleModuleStepConfig`; (2) the input tensor has an extra time dimension of 1 that must be squeezed out after the forward pass; (3) CrossFormer has its own atmospheric-level concept (`n_atmo_groups`, `n_atmo_channels`) that doesn't map to the flat channel list `SingleModuleStep` expects. That said, I agree these are friction points that will grow over time. One path worth exploring: make the packer ordering and the squeeze configurable in `SingleModuleStep`, so CrossFormer can be registered as a backbone there rather than requiring its own step class. Happy to prototype that refactor if you'd like.

fme/core/registry/module.py
yyexela: Will work on this further

fme/ace/step/camulator.py
yyexela: Will look at this further

mcgibbon: It seems from the meeting like we'll drop the crossformer, so perhaps this is moot.

fme/core/coordinates.py
yyexela: Claude: Added `use_cpb_scaling: bool = True` to both `SwinTransformerBuilder` and `NoiseConditionedSwinTransformerBuilder`. When set to `False`, the builders skip the latitude coordinate requirement and pass `lat_coords=None` to `SwinTransformerNet`, which routes through the existing `lat_mean is None` branch already present in `WindowAttention2D.forward()` — plain log-spaced CPB offsets (Swin V2 style) with no silent behavior change. The error message when `use_cpb_scaling=True` and coords are missing now includes a hint to set `use_cpb_scaling=False`. Two new tests cover the no-coords build path for both builders.

fme/ace/step/camulator.py
yyexela: Yes, deleted it in a new commit.

fme/core/coordinates.py
yyexela: Claude:

Closing #1259 via changes on this branch.

The issue flagged that after the `use_cpb_scaling` option was introduced, the `lat_1d is None` guard was dropped from both Swin builders. This meant that if `use_cpb_scaling=True` (the default) but the coordinate type returns `None` for `lat_1d` (e.g. any subclass that forgets to override it, like `HEALPixCoordinates`), scaling would be silently skipped — the opposite of what the user requested.

**Fix:** restored the `lat_1d is None` check in both `SwinTransformerBuilder.build` and `NoiseConditionedSwinTransformerBuilder.build`. When `use_cpb_scaling=True` and `lat_1d` returns `None`, a clear `ValueError` is raised directing the user to set `use_cpb_scaling=False`. Two new tests cover this path using `HEALPixCoordinates` (which inherits the `None` default from the base class).

yyexela: Basically, added `use_cpb_scaling: bool = True` config option, allowing swin to work without lat coords. Added a check to make sure if `use_cpb_scaling` is `True` but `lat_1d` is `None` we fail.

fme/core/registry/module.py
yyexela: Added follow up issue https://github.com/ai2cm/ace/issues/1258.

fme/core/registry/module.py
yyexela: Added follow up issue https://github.com/ai2cm/ace/issues/1258.

fme/ace/stepper/single_module.py
yyexela: Codex: Yes, I think this is valid for training generally. The mask passed to
  `predict_generator` is applied to the full network input, which is built from both
  state and forcing:

  ```python
  input_data = {**state, **input_forcing}
  ```

  `input_ensemble_data.data_mask` is incomplete because `input_ensemble_data` is
  built only from `self._prognostic_names`. For `in_names=["a", "b"]` and
  `out_names=["a"]`, it only has `a`, so the mask for input-only forcing `b` is
  dropped.

  `forcing_ensemble_data.data_mask` comes from the full batch after ensemble
  broadcasting, so it includes both prognostic and forcing-only masks with the
  right shape. Extra mask entries are harmless because masking only applies to
  variables present in the input.

  The failing coverage is
  `test_train_on_batch_masked_forcing_repeats_mask_across_ensemble`.

  Agreed this is not backbone-specific. I’m fine splitting it into a separate PR if
  you’d prefer, but it fixes the failing masked-forcing ensemble test here.

fme/core/registry/module.py
yyexela: Ok yes, in #1258 the issue is explained, claude suggests fixing old builders (non-swin, sfno builders) to match what swin is doing.

yyexela:
Ready for re-review @mcgibbon !

fme/core/models/swin_transformer/boundary_padding.py
```diff
@@ -0,0 +1,170 @@
+import dataclasses
```
Arcomano1234, line 23: Comment: We should be pretty careful if we are just basically copying camulators `boundary_padding.py` with out giving them credit

fme/core/registry/module.py
mcgibbon: Ah I realize I was confusing.

Please modify your PR to remove this method, using the pattern used by the existing builder instead. We can modify the pattern of all as a follow up, but don’t leave it inconsistent in this PR, and don’t change the existing pattern.

mcgibbon:
See my comment re the conditional treatment. Also, PR title and description will need to be updated.

fme/core/models/swin_transformer/boundary_padding.py
yyexela: Added! Good catch.

fme/core/registry/module.py
yyexela: Updated to have the swin match what's already in the code.

yyexela:
Ready for re-review @mcgibbon and @Arcomano1234 

Updated swin to match SFNO's builder, added license to `boundary_padding.py`.
Updated PR description and title.

fme/core/models/swin_transformer/boundary_padding.py
Arcomano1234: Thanks!

fme/ace/stepper/test_single_module.py
```diff
@@ -2596,6 +2596,60 @@ def test_train_on_batch_masked_variable_has_zero_loss_count():
...
+    }
+    config = _get_stepper_config(["a"], ["a", "b"], module_name="RepeatChannel")
+    stepper = _get_train_stepper(config, loss=StepLossConfig(type="MSE"))
+    stepped = stepper.train_on_batch(data=data_with_ic, optimization=NullOptimization())
+    assert stepped.per_channel_losses is not None
+    assert stepped.per_channel_losses["b"].count == 0
+    assert stepped.per_channel_losses["a"].count == n_samples
+
+
+def test_train_on_batch_masked_forcing_repeats_mask_across_ensemble():
```
mcgibbon: Issue: This test (and the other in this file) is unrelated to adding the swin transformer. Please separate it out into its own PR, along with the `single_module.py` changes.

mcgibbon (commented):
Going to post AI review comments after this, which I'll look through (and reply to as relevant) before saying it's ready to iterate.

fme/core/registry/module.py
mcgibbon: Great, let's do that in a follow on PR.

fme/core/models/conditional_sfno/layers.py
```diff
@@ -293,11 +289,11 @@ def forward(
                     list(x.shape[:-2]) + [1, 1], device=x.device, dtype=x.dtype
                 )
 
-            if self.W_scale_labels is not None:
+            if self.W_scale_labels is not None and context.labels is not None:
```
mcgibbon: Claude: Removing the `labels must be provided` guard is a behavior change to the existing SFNO path, not just Swin: a net built with label weights now silently skips label conditioning when `labels=None` instead of raising. I believe it's the right semantics (it fixes the latent crash for an unconditional selector on a labeled dataset, and `Module.__call__` still fail-fasts with "Labels are required for conditional models" on the registry path) — but no test in `conditional_sfno` locks in the new None-labels behavior. Suggest adding one, since direct callers of the net lose the fail-fast.

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,292 @@
...
+        self,
+        n_in_channels: int,
+        n_out_channels: int,
+        dataset_info: DatasetInfo,
+    ) -> nn.Module:
+        return self._build(
+            n_in_channels=n_in_channels,
+            n_out_channels=n_out_channels,
+            dataset_info=dataset_info,
+            enable_label_conditioning=len(dataset_info.all_labels) > 0
```
mcgibbon: Claude: With `build_for_selector` gone, `build()` is the only caller of `_build()` and always passes `enable_label_conditioning = len(all_labels) > 0 or embed_dim_labels > 0`, which makes the `if not enable_label_conditioning` branches unreachable (when both are zero, `embed_dim_labels` is already 0). Inlining `_build` into `build` (here and in `NoiseConditionedSwinTransformerBuilder`) would complete the match with the SFNO builders, which have no such indirection.

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,292 @@
...
+        dataset_info: DatasetInfo,
+        enable_label_conditioning: bool,
+    ) -> nn.Module:
+        n_labels = len(dataset_info.all_labels)
+        embed_dim_labels = self.embed_dim_labels
+        if not enable_label_conditioning:
+            embed_dim_labels = 0
+        elif n_labels > 0 and embed_dim_labels == 0:
+            # The registry feeds one-hot labels of dimension n_labels.
+            embed_dim_labels = n_labels
```
mcgibbon: Claude: Edge case: if a user sets `embed_dim_labels > 0` that differs from `n_labels`, the registry feeds one-hot labels of dimension `n_labels` straight into `Linear(embed_dim_labels, ...)` (there's no `Linear(n_labels, embed_dim)` mapping here, unlike the noise-conditioned variant), so it fails with a cryptic shape error at forward time. A guard raising when `embed_dim_labels not in (0, n_labels)` would surface the config error at build time. Optional.

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,292 @@
...
+        drop_path_rate: Maximum stochastic-depth rate.
+        use_skip: Whether to concatenate the layer-1 skip into the decoder.
+        mlp_layer: ``"mlp"`` or ``"swiglu"``.
+        noise_embed_dim: Dimension of the Gaussian noise field injected through
+            each block's ``ConditionalLayerNorm``.
+        label_embed_dim: Dimension of the learned label embedding space.
+            When > 0, a shared ``Linear(n_labels, label_embed_dim)`` layer maps
+            one-hot labels before downstream CLN conditioning.
+            When 0 (default), one-hot labels are used directly.
+            Label conditioning is enabled only when the selector is conditional,
```
mcgibbon: Claude: Stale docstring — "Label conditioning is enabled only when the selector is conditional" describes the removed `build_for_selector` behavior. Conditioning is now enabled from dataset labels / `label_embed_dim` alone.

fme/ace/registry/test_swin_transformer.py
```diff
@@ -0,0 +1,307 @@
...
+    ).to(fme.get_device())
+    x = torch.randn(2, n_in, *IMG_SHAPE, device=fme.get_device())
+    labels = BatchLabels.new_from_set(all_labels, n_samples=2, device=fme.get_device())
+    out = module(x, labels=labels)
+    assert out.shape == (2, n_out, *IMG_SHAPE)
+    net = getattr(module.torch_module, "module")
+    assert net.embed_dim_labels == len(all_labels)
+
+
+def test_swin_transformer_unconditional_ignores_dataset_labels():
```
mcgibbon: Claude: This test name (and `test_nc_swin_transformer_unconditional_ignores_dataset_labels` at L197) now contradicts the assertion — they assert `embed_dim_labels == len(all_labels)`, i.e. dataset labels are *not* ignored. Suggest renaming to something like `test_swin_transformer_unconditional_builds_label_weights_from_dataset_labels`.

fme/core/step/single_module.py
```diff
@@ -446,7 +446,7 @@ def load_state(self, state: dict[str, Any]) -> None:
             self.secondary_decoder.load_module_state(state["secondary_decoder"])
 
 
-def _apply_input_mask(input_norm: TensorDict, data_mask: TensorMapping) -> TensorDict:
+def apply_input_mask(input_norm: TensorDict, data_mask: TensorMapping) -> TensorDict:
```
mcgibbon: Claude: With the camulator/CrossFormer step dropped from the PR, nothing outside this module and its tests uses `apply_input_mask` anymore, so the public rename lost its motivation. Suggest reverting to `_apply_input_mask` (repo convention: private functions get a `_` prefix) and removing the "Minor" rename bullet from the PR description.

mcgibbon (commented):
Claude: Re-review of `ee47fe4f0...abf1f626a` (the three commits since the last review). The main ask — dropping `build_for_selector` and matching the SFNO builder pattern — is properly addressed: `ModuleSelector.build` is back to plain `build()`, and the Swin builders resolve `embed_dim_labels` from `dataset_info.all_labels` the same way `NoiseConditionedSFNO` does. The boundary-padding LICENSE/attribution and PR description updates also look good (upstream miles-credit is Apache-2.0, same as this repo, so no compatibility issue). All touched test suites pass locally (swin, boundary padding, conditional_sfno, step, stepper mask tests). Remaining inline comments below are minor; no blocking issues from these commits.

fme/ace/registry/swin_transformer.py
mcgibbon: Chore: Agreed.

fme/ace/registry/swin_transformer.py
mcgibbon: I think this is a bug? The whole purpose of embed_dim_labels is the linear projection. Minimal change is to just delete that option, but if you want parity with the SFNO in terms of conditioning, you should update the code to use the same code pattern for projection.

fme/ace/registry/swin_transformer.py
mcgibbon: Chore.

fme/ace/registry/test_swin_transformer.py
mcgibbon: Chore.

fme/ace/stepper/single_module.py
mcgibbon: Let's please split it, and the test (which is new and not related to the purpose of this PR) into a separate PR. I don't have the context in this PR to know why the change is needed or why the test exists (e.g. why doesn't it test the IC gets mask-repeated?), these will be easier to discuss in their own context.

fme/core/models/conditional_sfno/layers.py
mcgibbon: Suggestion: Revert the change, unless there's a clear need for it.

I included this raise as a feature. Even if the check is redundant (i.e. this will already fail), it's safer to avoid unintended configurations (e.g. forgetting to specify labels during a standalone inference run) if we require labels for conditional models.

mcgibbon:
Ready for your iteration, thanks!

fme/core/models/conditional_sfno/layers.py
yyexela: Claude: Reverted — guard restored and the two `and context.labels is not None` conditions removed.

fme/ace/registry/swin_transformer.py
yyexela: Claude: Done — `_build` inlined into `build` for both `SwinTransformerBuilder` and `NoiseConditionedSwinTransformerBuilder`.

fme/ace/registry/swin_transformer.py
yyexela: Claude: Removed `embed_dim_labels` from `SwinTransformerBuilder` entirely. The builder now always passes `embed_dim_labels=n_labels` to the context config. `NoiseConditionedSwinTransformerBuilder` already had `label_embed_dim` with the proper projection path, matching `NoiseConditionedSFNO`.

fme/ace/registry/swin_transformer.py
yyexela: Claude: Removed the stale sentence from the `label_embed_dim` docstring in `NoiseConditionedSwinTransformerBuilder`.

fme/ace/registry/test_swin_transformer.py
yyexela: Claude: Renamed both tests to `test_swin_transformer_unconditional_builds_label_weights_from_dataset_labels` and `test_nc_swin_transformer_unconditional_builds_label_weights_from_dataset_labels`. Also removed the forward call from the NC variant — with the CLN guard restored, calling the noise-conditioned model without labels raises, so the test now just checks the `embed_dim_labels` attribute.

fme/core/step/single_module.py
yyexela: Claude: Reverted — `apply_input_mask` renamed back to `_apply_input_mask` in `single_module.py`, with the call site and the three test references updated to match.

fme/ace/stepper/single_module.py
yyexela: Separated this out into another PR:
https://github.com/ai2cm/ace/pull/1262

fme/ace/stepper/test_single_module.py
yyexela: Added separate PR, moved the test out: https://github.com/ai2cm/ace/pull/1262

yyexela:
Thanks @mcgibbon !

Responded to the follow-on comments. I created a separate PR https://github.com/ai2cm/ace/pull/1262 addressing the bug found earlier. I thought this would block the PR but it doesn't. The new PR #1262 is for a bug discovered in the process of making this PR.

This PR is ready for another review!

Arcomano1234:
Ill wait for Jeremy's assessment but this seems to be shaping up well. Can you fix the PR description I think there is quite of bit of stale things in there still.

yyexela:
Updated title/description to what's currently in the PR!

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,246 @@
...
+            window_size=(self.window_size[0], self.window_size[1]),
+            mlp_ratio=self.mlp_ratio,
+            drop_path_rate=self.drop_path_rate,
+            use_skip=self.use_skip,
+            context_config=context_config,
+            mlp_layer=self.mlp_layer,
+            cpb_hidden_dim=self.cpb_hidden_dim,
+            lat_coords=lat_coords,
+            padding_conf=dataclasses.asdict(self.padding_conf)
+            if self.padding_conf is not None
```
mcgibbon: nit: since none-handling for each other arg is handled before this assignment, you could move this one out too.

fme/core/models/miles_credit/LICENSE
mcgibbon, line 1: nit (more optional than usual): Since at this point this is really only for one file, you could put a license header in boundary_padding.py instead (doesn't need the full icense, just a specification that it uses an apache 2.0 license and its copyright/author info). Just less likely to get separated from the code if anything moves.

fme/ace/registry/swin_transformer.py
```diff
@@ -0,0 +1,246 @@
...
+                    "scaling, but the dataset provides none. Non-lat-lon grids such as "
+                    "HEALPix are not supported. Set use_cpb_scaling=False to disable "
+                    "this requirement."
+                ) from None
+            if lat_coords is None:
+                raise ValueError(
+                    "SwinTransformer requires 1D latitude coordinates for cos-lat CPB "
+                    "scaling, but this coordinate type returns None for lat_1d. "
+                    "Set use_cpb_scaling=False to disable this requirement."
+                )
```
mcgibbon: nit: the dataset info should either raise MissingDatasetInfo when the field is missing, or return None, but not both. So only one of these should be necessary (I'd guess the second one).

mcgibbon (commented):
Looking good - I was able to find some nits, but I don't need to re-review. Thanks!

fme/core/models/miles_credit/LICENSE
yyexela: @Arcomano1234 what do you prefer? I'm happy w/ it either way

fme/ace/registry/swin_transformer.py
yyexela: Moved out `padding_conf` none-handling!

fme/ace/registry/swin_transformer.py
yyexela: Claude:

The fix needs to live in `DatasetInfo` so `swin_transformer.py` never touches `MissingDatasetInfo` at all - just calls `dataset_info.lat_1d` and gets `None` when unavailable. That requires adding `torch` to `dataset_info.py`.

Alexey:

I think this is the only way to get the behavior you prefer, otherwise you'd have a block like:

```
try:
    lat_coords = dataset_info.horizontal_coordinates.lat_1d
except MissingDatasetInfo:
    lat_coords = None
if lat_coords is None:
    ...
```

The proposed fix then makes it 

```
lat_coords = dataset_info.horizontal_coordinates.lat_1d
if lat_coords is None:
    ...
```

but then adds a torch dependency to `dataset_info.py` just for the data type `torch.Tensor`, which otherwise doesn't need to be there. I'm ok with this but would like to know how you feel about it!

yyexela: @mcgibbon what do you recommend? Do you have a reason `lat_coords` would ever be `None`?

fme/core/models/miles_credit/LICENSE
yyexela: Troy says to pull out, will do this

yyexela: Pulled out!

fme/core/coordinates.py
```diff
@@ -586,6 +586,10 @@ def meshgrid(self) -> tuple[torch.Tensor, torch.Tensor]:
...
+        return None
+
```
yyexela, line 599: Make abstract, have healpix return none

yyexela (commented):
TODO

fme/core/coordinates.py
yyexela: Remove, make sure lat_1d never returns None, always raises error

yyexela: Make property abstract, raise missingdatasetinfo on healpix class

yyexela: Ready for a look-over @mcgibbon ! Ty!

mcgibbon: Issue: we said this will never return None. Update the return type hint, and raise an informative MissingDatasetInfo from the healpix class instead (describing that because it uses 12 tiles, it doesn't have a clean 1-dimensional representation for latitude).

yyexela: Sorry, fixed that in a new commit!

fme/core/coordinates.py
```diff
@@ -11,6 +11,7 @@
 
 from fme.core import metrics
 from fme.core.constants import EARTH_RADIUS, GRAVITY
+from fme.core.dataset_info import MissingDatasetInfo
```
mcgibbon: I think if you move MissingDatasetInfo into its own file that doesn't really import anything, it will fix the circular import you're getting.

yyexela: Tests pass now, ty!

mcgibbon (approved):
LGTM

fme/ace/registry/swin_transformer.py
yyexela: (resolved below, lat_coords is none in healpix)

fme/core/models/swin_transformer/boundary_padding.py
```diff
@@ -0,0 +1,177 @@
+# Adapted from
+# https://github.com/NCAR/miles-credit/blob/main/credit/boundary_padding.py
+# CREDIT (Community Research Earth Digital Intelligence Twin) is an open
+# software platform by the NSF NCAR MILES group for training and deploying
+# AI atmospheric prediction models.
+# Reference: Schreck et al., npj Clim Atmos Sci 8, 239 (2025).
+#   https://doi.org/10.1038/s41612-025-01125-6
```
Arcomano1234, line 7: Question: Do we need to copy the Apache License 2.0 license into the header for this file similar to how its done in `swin_layers.py` and `swin_transformer.py`?

Arcomano1234 (approved):
Approved with one more licensing question. If this is needed just add the license and feel free to merge.

PR #1249, opened 2026-06-09
Fall back to static input coords when loading checkpoints without full_fine_coords

frodre:
## Description

Checkpoints saved during the window where coordinates were stored per-field in `StaticInputs` (but not yet as top-level `full_fine_coords`) previously could not be loaded without providing `fine_coordinates_path`. The publicly released HiRO v1 checkpoint on HuggingFace is in this in-between format, so loading it required a config workaround even though the coordinates are present in the checkpoint's static inputs.

`CheckpointModelConfig.build()` now loads static inputs first and falls back to their coordinates when resolving `full_fine_coords`. The fallback order is:

1. `full_fine_coords` in checkpoint state
2. `fine_coordinates_path` from config
3. Static input coordinates (new)

Existing behavior is unchanged, including configs that currently work around the issue via `fine_coordinates_path` — a mismatch between path coords and static input coords is still caught by the existing validation in `DiffusionModelConfig.build`.

## Testing

Added `test_checkpoint_model_build_with_legacy_static_input_coords`, which simulates the in-between checkpoint format (no `full_fine_coords` key, legacy per-field coords in static inputs) and verifies loading succeeds. Confirmed the test failed before the fix.

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -818,6 +818,7 @@ class CheckpointModelConfig:
     def _get_coords_backwards_compatible(
         coords_from_state: dict | None,
         fine_coordinates_path: str | None,
+        static_inputs: StaticInputs | None,
     ) -> LatLonCoordinates:
         if coords_from_state and fine_coordinates_path:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


frodre:
```
In [1]: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:  from huggingface_hub import hf_hub_download
:
:  from fme.downscaling.models import CheckpointModelConfig
:
:  ckpt_path = hf_hub_download(repo_id="allenai/HiRO-ACE", filename="HiRO.ckpt")
:
:  # With the fix on this branch, no fine_coordinates_path workaround is needed —
:  # coordinates are recovered from the checkpoint's static inputs.
:  model = CheckpointModelConfig(checkpoint_path=ckpt_path).build()
:
:  print("in_names: ", model.in_names)
:  print("out_names:", model.out_names)
:  print("coarse_shape:", model.coarse_shape, "downscale_factor:", model.downscale_factor)
:  print("fine grid:", model.full_fine_coords.lat.shape, model.full_fine_coords.lon.shape)
:--

Could not import pad from earth2grid.healpix.
HiRO.ckpt: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 921M/921M [00:11<00:00, 81.1MB/s]
/home/frodre/repos/ace/.worktrees/ace-fix-ckpt-load/fme/downscaling/models.py:202: UserWarning: p_mean and p_std are deprecated. Use training_noise_distribution field instead.
  warnings.warn(
in_names:  ['PRATEsfc', 'eastward_wind_at_ten_meters', 'northward_wind_at_ten_meters']
out_names: ['PRATEsfc']
coarse_shape: (16, 16) downscale_factor: 32
fine grid: torch.Size([5760]) torch.Size([11520])
```

oliverwm1 (approved):

PR #1251, opened 2026-06-10
Add eval-only global mean relaxation to SingleModuleStep

mcgibbon:
Adds an eval-only Newtonian relaxation of the global mean of selected output variables toward configured target values, applied between any global-mean-removal inverse transform and the corrector so the corrector operates on the relaxed field. Eval/train state on the step is made explicit via new `train()`/`eval()` methods on `StepABC`, matching the `torch.nn.Module` API.

Changes:
- `fme.core.step.global_mean_relaxation`: new `GlobalMeanRelaxationConfig`, `GlobalMeanRelaxationVariableConfig`, and `GlobalMeanRelaxation`; per variable, subtracts `(area_weighted_mean(x) - target) / timescale_steps`. `target` may be a float or `"mean"` (resolved to the network normalizer's mean at build time).
- `fme.core.step.single_module.SingleModuleStepConfig`: new optional `global_mean_relaxation` field, validated against `out_names`. Wired into `step_with_adjustments` via a new kwarg, gated on the step's training flag so it is inert during training.
- `fme.core.step.step.StepABC`: new `train(mode=True)`/`eval()` methods and a `_training` flag, mirroring `torch.nn.Module`. `fme.ace.stepper.single_module.SingleModuleStepper.set_eval`/`set_train` now delegate to the step's methods instead of iterating modules directly.

- [x] Tests added

### DIFF
--- a/fme/core/step/global_mean_relaxation.py
+++ b/fme/core/step/global_mean_relaxation.py
@@ -1,0 +1,149 @@
+import dataclasses
+from typing import Literal
+
+import torch
+
+from fme.core.gridded_ops import GriddedOperations
+from fme.core.normalizer import StandardNormalizer
+from fme.core.typing_ import TensorDict, TensorMapping
+
+USE_NORMALIZATION_MEAN: Literal["mean"] = "mean"
+
+
+@dataclasses.dataclass
+class GlobalMeanRelaxationVariableConfig:
+    """
+    Per-variable target and timescale for global-mean Newtonian relaxation.
+
+    Parameters:
+        target: Target global mean. Either a float in the variable's
+            physical units, or the literal string ``"mean"`` to use the
+            variable's normalization mean as the target.
+        timescale_steps: Relaxation timescale in steps. Each step shifts the
+            field by ``-(area_weighted_mean(x) - target) / timescale_steps``,
+            i.e. an exponential approach to ``target`` with e-folding time of
+            ``timescale_steps`` steps. Must be positive.
+    """
+
+    target: float | Literal["mean"]
+    timescale_steps: float
+
+    def __post_init__(self):
+        if self.timescale_steps <= 0:
+            raise ValueError(
+                f"timescale_steps must be positive, got {self.timescale_steps}."
+            )
+        if isinstance(self.target, str) and self.target != USE_NORMALIZATION_MEAN:
+            raise ValueError(
+                f"target must be a number or '{USE_NORMALIZATION_MEAN}', got "
+                f"{self.target!r}."
+            )
+
+
+@dataclasses.dataclass
+class GlobalMeanRelaxationConfig:
+    """
+    Newtonian relaxation of the global mean of selected output variables
+    toward configured target values.
+
+    The relaxation is applied during ``.step`` in eval mode only, after
+    the network call (and any global-mean-removal inverse transform) and
+    before the corrector. For each named variable, the step subtracts a
+    uniform offset of ``(area_weighted_mean(x) - target) / timescale_steps``
+    from the field.
+
+    Names in ``variables`` must appear in ``out_names`` of the enclosing
+    step config. Variables whose ``target`` is ``"mean"`` must have a
+    normalization mean defined in the network normalizer.
+
+    Parameters:
+        variables: Mapping from output variable name to its per-variable
+            relaxation configuration.
+    """
+
+    variables: dict[str, GlobalMeanRelaxationVariableConfig]
+
+    def __post_init__(self):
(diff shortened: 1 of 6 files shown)


fme/core/step/global_mean_relaxation.py
```diff
@@ -0,0 +1,149 @@
+import dataclasses
+from typing import Literal
+
+import torch
+
+from fme.core.gridded_ops import GriddedOperations
+from fme.core.normalizer import StandardNormalizer
+from fme.core.typing_ import TensorDict, TensorMapping
+
+USE_NORMALIZATION_MEAN: Literal["mean"] = "mean"
```
yyexela, line 10: Question: Is this the right name to use? The name `USE_NORMALIZATION_MEAN` seems like it'd be a bool more than a fixed literal string.

fme/core/step/step.py
```diff
@@ -235,11 +235,33 @@ def from_state(cls, state: Mapping[str, Any]) -> Self:
 class StepABC(abc.ABC):
     SelfType = TypeVar("SelfType", bound="StepABC")
 
+    def __init__(self) -> None:
+        # Mirrors ``torch.nn.Module.training`` so that step-level eval/train
+        # state is observable without reaching into the underlying modules.
+        self._training: bool = True
```
yyexela, line 241: Claude: This new base-class `__init__` sets `_training`, but `MockStep` in
  `test_step_registry.py` doesn't call `super().__init__()`, so it won't have
  that attribute. Calling `.train()` or `.eval()` on a `MockStep` will raise
  `AttributeError`. Fix: add `super().__init__()` as the first line of
  `MockStep.__init__`.

yyexela: Alexey: If this is right, also add a test for this

fme/core/step/global_mean_relaxation.py
```diff
@@ -0,0 +1,149 @@
...
+class GlobalMeanRelaxationConfig:
+    """
+    Newtonian relaxation of the global mean of selected output variables
+    toward configured target values.
+
+    The relaxation is applied during ``.step`` in eval mode only, after
+    the network call (and any global-mean-removal inverse transform) and
+    before the corrector. For each named variable, the step subtracts a
+    uniform offset of ``(area_weighted_mean(x) - target) / timescale_steps``
+    from the field.
```
yyexela, line 53: Issue: Missing integration test for global_mean_removal + global_mean_relaxation together

Suggestion: Make sure the ordering is tested since that's a big part of how the PR says the data should flow.

fme/core/step/global_mean_relaxation.py
```diff
@@ -0,0 +1,149 @@
...
+        self,
+        targets: dict[str, float],
+        timescales: dict[str, float],
+        gridded_operations: GriddedOperations,
+    ):
+        if set(targets) != set(timescales):
+            raise ValueError(
+                "targets and timescales must have the same keys: "
+                f"targets={list(targets)}, timescales={list(timescales)}."
+            )
```
yyexela, line 130: Issue: This guard is never tested.

Suggestion: Looks like `class GlobalMeanRelaxation` only ever gets made in `GlobalMeanRelaxationConfig.build()`, so this guard would never trigger since in that case `targets` == `timescales`. Either test this guard or make `class GlobalMeanRelaxation` private (ie. `class _GlobalMeanRelaxation`)

fme/core/step/single_module.py
```diff
@@ -401,6 +421,16 @@ def network_call(input_norm: TensorDict) -> TensorDict:
...
 
+        # Newtonian relaxation toward configured target global means is an
+        # eval-time-only adjustment; gating on the step's training flag
+        # keeps it from contributing to the loss during training.
+        if self._global_mean_relaxation is not None and not self._training:
+            global_mean_relaxation: GlobalMeanRelaxation | None = (
+                self._global_mean_relaxation
+            )
+        else:
+            global_mean_relaxation = None
```
yyexela, line 432: Nitpick: This can be simplified to:


```
global_mean_relaxation = self._global_mean_relaxation if not self._training else None
```

yyexela: Pretty ugly otherwise

fme/core/step/global_mean_relaxation.py
yyexela: Suggestion: Use `NORMALIZATION_MEAN_TARGET` instead

yyexela (commented):
Looks good overall, just some comments/questions/suggestions to clarify.

PR #1252, opened 2026-06-10
Trim slow test parametrizations and oversized test data

mcgibbon:
Several of the slowest tests in the suite spend their time on data volume or parametrized cases that add no coverage. This PR trims them while preserving the behaviors each test verifies, cutting several minutes of CI time without changing any production code.

Changes:
- `fme/ace/aggregator/inference/ipo/test_ipo_index.py::TestPairedIPOIndexAggregator`: monkeypatch `MIN_YEARS_FOR_FILTERED_TPI` to 5 years (and use a 2-year filter cutoff) so the long/short rollout behaviors of `get_logs` are exercised with ~10x less monthly data. Local timings: `test_get_logs_returns_expected_keys` 13.1s -> 1.7s, `test_get_logs_empty_for_short_rollout` 5.0s -> 0.3s (39s and 12s in CI before).
- `fme/ace/test_train.py::test_train_and_inference`: fold `validate_using_ema=True` into the `SFNO-crps-schedule` case and delete the otherwise-identical `SFNO-crps-schedule-ema` case, removing one full train+inference cycle (~27s locally, ~1 min in CI). Non-EMA validation remains covered by the `SFNO-val-maps-multi`, `SFNO-masking`, `Samudra`, and `HEALPix` cases.
- `fme/ace/test_train.py::test_train_and_inference_with_derived_forcings`: drop the `solar-constant-as-value` parametrization (~12s locally), keeping the more complex as-name path that loads the solar constant from disk. Value-vs-name resolution is already unit-tested by `fme/ace/stepper/insolation/test_insolation.py` (`test_update_requirements`, `test_insolation_compute`) and `fme/ace/stepper/test_derived_forcings.py::test_forcing_deriver`.
- `fme/ace/models/graphcast/test_graphcast.py::test_graphcast_normalization`: reduce the activation parametrization from 5 to `["SiLU", "GELU"]` (~4s per case locally, ~7s in CI); the test only asserts output shape and absence of NaN/Inf, which the activation choice does not affect.

- [ ] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/ipo/test_ipo_index.py
+++ b/fme/ace/aggregator/inference/ipo/test_ipo_index.py
@@ -9,8 +9,8 @@ from fme import get_device
 from fme.ace.aggregator.inference.data import InferenceBatchData
 
 from ..utils import LatLonRegion
+from . import ipo_index
 from .ipo_index import (
-    MIN_YEARS_FOR_FILTERED_TPI,
     PairedIPOIndexAggregator,
     _IPORegionalAccumulator,
     low_pass_filter,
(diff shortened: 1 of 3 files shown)


mcgibbon:
Claude: I verified that the removed parametrizations are covered by existing tests, in support of removing them. CI cost of the removed cases was ~53s of test runtime (timings from the last green run on main).

**`solar-constant-as-value` (11.1s)** — covered at three levels:
- `fme/ace/stepper/insolation/test_insolation.py::test_update_requirements`: 4 cases over `ValueConfig`/`NameConfig`, verifying as-name adds the solar-constant variable to data requirements and as-value doesn't (the key behavioral difference).
- `test_insolation.py::test_insolation_compute[solar-constant-as-value]`: runs the computation against its own `.pt` regression baseline.
- `fme/ace/stepper/test_derived_forcings.py::test_forcing_deriver[solar-constant-as-value]`: the `ForcingDeriver` wiring.
- In addition, `fme/ace/inference/test_evaluator.py::test_evaluator_with_derived_forcings[solar-constant-as-value]` still runs a full inference integration test with `ValueConfig(1360.0)`, including the stepper-checkpoint round-trip of the config. What this PR removes is only the duplicate at the training entry point; integration-level as-value coverage remains.

**`SFNO-crps-schedule-ema` (21.8s)** — not removed but folded: the surviving `SFNO-crps-schedule` case now runs with `validate_using_ema=True`, so the flag's end-to-end wiring is still exercised. The EMA mechanism has 11 unit tests in `fme/core/test_ema.py` (updates, state transfer, resume, finetuning), and `fme/core/generics/test_validation.py::test_run_validation_loop_with_ema`/`_without_ema` directly test the EMA-swap-during-validation behavior. One nuance: there is no longer a training e2e of non-EMA validation combined specifically with crps+schedule, but non-EMA validation runs in the other four cases (Samudra, HEALPix, masking, val-maps-multi).

**GraphCast `Mish`/`ReLU`/`Tanh` (19.8s total)** — no dedicated unit tests, and none needed: activation resolves via `getattr(nn, activation)` at `fme/ace/models/graphcast/main.py:84`, so there is no per-activation code in this repo. The removed cases re-ran torch's own activation modules through an identical path, with assertions (shape/NaN/Inf) the activation choice cannot affect. The surviving SiLU (the default) + GELU cases keep the lookup mechanism covered.

oliverwm1 (approved):
LGTM thanks!

PR #1254, opened 2026-06-10
Convert slow-test gating from runtime fixtures to pytest markers

mcgibbon:
Slow-test gating used `skip_slow`/`very_fast_only` fixtures with `pytest.skip(...)` boilerplate at the top of ~78 test bodies, deciding at runtime what the `--fast` and `--very-fast` modes should run. This PR converts that gating to `slow` and `medium_duration` markers deselected at collection time, making the tiers discoverable via `pytest --markers` and removing the per-test boilerplate. The CLI flags, Makefile targets, and CI are unchanged; `medium_duration` means "not very fast" (runs under `--fast` but not `--very-fast`), while `slow` runs only in full mode.

Changes:
- `conftest.py`: register `slow` and `medium_duration` markers, deselect them in `pytest_collection_modifyitems` based on `--fast`/`--very-fast`; remove the now-unused `skip_slow`/`very_fast_only` fixtures (`enforce_timeout` reads the options directly)
- test files across `fme/` and `scripts/`: replace fixture-plus-skip boilerplate with `@pytest.mark.slow`/`@pytest.mark.medium_duration`; parametrizations that were gated conditionally (e.g. `fme/ace/data_loading/test_data_loader.py`) use `pytest.param(..., marks=...)` on the specific cases

Semantics verified by running the full suite with `--very-fast` on this branch vs main: identical executed-test counts (3077), and main's 91 skips become exactly 26 skips + 65 deselections on this branch (the 65 tier-gated tests). The only failures in both runs were the known-flaky graphcast tests.

- [ ] Tests added


oliverwm1 (commented):
Nice, agreed the mark is much clearer. 

Unlike "fast" and "slow", "medium" doesn't necessarily suggest a duration so `pytest.mark.medium` on a test without other context is a bit unclear. Could do "medium_slow" or "slowish" or something else?

mcgibbon:
Let's do `medium_duration`.

oliverwm1 (approved):

PR #1257, opened 2026-06-10
Add ensemble-mean annual RMSE metric to annual aggregator

mcgibbon:
The annual aggregator reports R2 of the predicted ensemble-mean annual evolution, but R2 requires reference data and is undefined without it. An RMSE against the target ensemble-mean annual evolution gives a complementary skill measure that requires no reference data, so it is computed and logged by default alongside the existing R2 metrics.

Changes:
- `fme.ace.aggregator.inference.annual.PairedGlobalMeanAnnualAggregator`: reports `rmse/<var>`, the RMSE of the predicted ensemble-mean annual evolution against the target ensemble-mean annual evolution
- `fme.ace.aggregator.inference.annual.get_rmse`: helper computing RMSE over years between a DataArray and a reference, ignoring NaN gap years introduced by reindexing
- `fme.ace.aggregator.inference.annual.get_r2`: now also ignores NaN gap years, so R2 and RMSE handle the same reindexed inputs consistently

- [x] Tests added


fme/ace/aggregator/inference/annual.py
```diff
@@ -316,6 +321,14 @@ def get_r2(da: xr.DataArray, reference: xr.DataArray) -> float:
     return float(1 - SS_pred / SS_ref)
 
 
+def get_rmse(da: xr.DataArray, reference: xr.DataArray) -> float:
+    """Compute the RMSE of the data compared to the reference over years,
+    ignoring NaN values (e.g. gap years filled in by reindexing).
+    """
+    ref_data = reference.sel(year=da.year)
+    return float(np.sqrt(np.nanmean((da - ref_data).values ** 2)))
```
yyexela, line 332: Issue: Code has `np.nanmean` to ignore NaN values, but this functionality is not tested in the current added tests.

Suggestion: Add a check to `test_paired_annual_aggregator_with_nans` checking `assert not np.isnan(logs["test/rmse/a"])`

fme/ace/aggregator/inference/annual.py
```diff
@@ -316,6 +321,14 @@ def get_r2(da: xr.DataArray, reference: xr.DataArray) -> float:
     return float(1 - SS_pred / SS_ref)
```
yyexela, line 324: Question: `get_rmse` added in this pr is nan-safe, but `get_r2` is not. Is there a good reason to have these two handle NaN's differently?

fme/ace/aggregator/inference/annual.py
```diff
@@ -113,6 +113,11 @@ def get_logs(self, label: str) -> dict[str, Any]:
             if gen.sizes["year"] > 1:
                 target_ensemble_mean = target[name].mean("sample")
                 gen_ensemble_mean = gen[name].mean("sample")
+                # RMSE of the predicted ensemble-mean annual evolution against
+                # the target ensemble-mean annual evolution
+                metrics[f"rmse/{name}"] = get_rmse(
+                    gen_ensemble_mean, target_ensemble_mean
+                )
```
yyexela: Nitpick (optional): The comments here are not necessary, code is pretty readable already

yyexela (commented):
Looks good overall, minimall changes, just some issues/suggestions to iterate on.

fme/ace/aggregator/inference/annual.py
mcgibbon: Claude: Good catch — the nanmean path was not actually exercised. The `with_nans` test only masks a spatial point, which the masked area-weighted mean removes, so the annual series had no NaNs. Added `assert not np.isnan(logs["test/rmse/a"])` there, and added `test_get_rmse_ignores_nan_gap_years`, which feeds a year series with a real NaN gap year and checks the hand-computed RMSE. Fixed in 9ada21b.

fme/ace/aggregator/inference/annual.py
mcgibbon: Claude: No good reason — they receive the same gap-year-reindexed inputs (annual.py:228), so the inconsistency was a latent bug in `get_r2`. Made `get_r2` NaN-safe as well, restricting the computation to years where both series are defined, and added `test_get_r2_ignores_nan_gap_years`. Fixed in 9ada21b.

fme/ace/aggregator/inference/annual.py
mcgibbon: Claude: Removed the comment above the `get_rmse` call in 9ada21b.

fme/ace/aggregator/inference/annual.py
yyexela: Nitpick (non-blocking): Tests have separate `nan` checks for `da` and `reference`, but tests just check if `da` with `nan` runs `get_rmse` and `get_r2`.

Suggestion: Have `nan` in `reference` in those tests too.

yyexela (approved):
Nitpick (non-blocking): Tests have separate `nan` checks for `da` and `reference`, but tests just check if `da` with `nan` runs `get_rmse` and `get_r2`.

Suggestion: Have `nan` in `reference` in those tests too.

Note: This is a pretty minor issue and code is correct anyways, I don't imagine those two functions changing and needing the additional check, but worth bringing up anyways. Looks great otherwise.

PR #1260, opened 2026-06-10
Disable corrector for the first N epochs

elynnwu:
Adds a `corrector_disabled_epochs` option to corrector configs that skips the corrector for **train-mode** steps during the first N training epochs. It is always applied in eval mode (validation and inference), so only training is affected.

Some runs get frozen precipitation stuck at exactly zero (e.g. [job](https://wandb.ai/ai2cm/ace-eamv3/runs/tu6kkzxd/overview)): the force-positive clamp zeroes the sparse frozen precip channel whenever an early update makes the raw prediction negative, and gradients through the clamp are zero there, so it never recovers. Disabling the corrector for the first epoch lets the loss see the raw prediction so gradients can pull the channel positive before the clamp returns.

`corrector_disabled_epochs` is a field on `CorrectorConfigABC`; its `@final get_corrector` wraps the built corrector in a new `EpochScheduledCorrector` (which skips an arbitrary wrapped corrector in train mode during the disabled epochs) when the value is > 0. Concrete configs just implement `_get_corrector` and are unaware of scheduling. The flag is valid only on the corrector config — `CorrectorSelector` rejects it and delegates — so there's one place to set it and no double-wrapping. The disabled flag is persisted in checkpoint state so mid-epoch resume keeps the interrupted epoch's behavior. Defaults to 0, so existing configs and checkpoints are unaffected.

### Config

```yaml
corrector:
  conserve_dry_air: true
  force_positive_names:
  - frozen_precipitation_rate
  corrector_disabled_epochs: 1   # inside `config:` for selector-based correctors
```

### Changes

- `CorrectorConfigABC`: `corrector_disabled_epochs` field + `@final get_corrector` wrapping via new abstract `_get_corrector`; `CorrectorABC` gains default no-op train/set_epoch/get_state/load_state; new `EpochScheduledCorrector` wrapper.
- `CorrectorSelector` rejects `corrector_disabled_epochs` (must be set on the wrapped config).
- Atmosphere/Ocean/Ice configs implement `_get_corrector`; step layer + Stepper forward set_epoch/train/get_state/load_state to the corrector.


elynnwu:
These two jobs are identical except the new job disabled corrector for the first epoch: [wandb](https://wandb.ai/ai2cm/ace-eamv3?nw=3bnepcraecc). The new job does not have the same zero frozen precipitation issue. Note that I was able to reproduce the zero frozen precip behavior in 3 experiments, all with n384 (one with equal loss, two with ACE2 loss weighting on Beaker and Perlmutter).

fme/core/step/single_module.py
```diff
@@ -402,12 +417,16 @@ def network_call(input_norm: TensorDict) -> TensorDict:
             output_dict.update(secondary_output_dict)
             return output_dict
 
+        corrector: CorrectorABC | None = self._corrector
+        if self._corrector_disabled and self.module.torch_module.training:
```
mcgibbon: Suggestion (optional): If it's convenient, could you pull in the .set_train/.set_eval additions from https://github.com/ai2cm/ace/pull/1251 and use those? I could refactor this later, but it will be easy to miss updating this line. I went that route to avoid going multiple levels deep into the state of a module and be explicit instead (Law of Demeter).

fme/core/step/single_module.py
```diff
@@ -87,9 +91,15 @@ class SingleModuleStepConfig(StepConfigABC):
     residual_prediction: bool = False
     include_channel_mask_inputs: bool = False
     global_mean_removal: GlobalMeanRemovalConfigUnion | None = None
+    corrector_disabled_epochs: int = 0
 
     def __post_init__(self):
         self.crps_training = None  # unused, kept for backwards compatibility
+        if self.corrector_disabled_epochs < 0:
```
mcgibbon: Two related questions, not sure if this points to a better solution but it should be brought up:

1. The corrector also exists on SeparateRadiationStep, SecondaryModuleStep, and FCN3Step. Should we update those to also have this feature?
2. It might be easier to port this feature over to new Steps if this were contained within the corrector. That is to say, having set_epoch and set_train methods on the corrector, and having the corrector own state that gets saved/loaded with the checkpoint (in this case, `corrector_disabled`, but you could store `epoch` directly if you wanted instead).

fme/core/step/single_module.py
elynnwu: Added in [180c343](https://github.com/ai2cm/ace/pull/1260/commits/180c343814f9e0f59e3c77cb35142372be0b3bb3)

fme/core/step/single_module.py
elynnwu: For this PR I’d prefer to keep the fix scoped to the affected `SingleModuleStep` path, but I agree the better generalization is a follow-up that makes this corrector-owned so new Step implementations inherit it automatically. Opened an issue [here](https://github.com/ai2cm/ace/issues/1261) to track.

mcgibbon: Keep in mind this change will require backwards compatibility checkpoint handling code if we merge the current version and refactor to a different version too much later.

mcgibbon (approved):
LGTM, but consider the corrector refactor sooner rather than later. I can also act on it if you like, a Claude agent pointed to your issue and this PR should be able to handle it.

fme/core/step/single_module.py
elynnwu: Okay in this case let's do it in this PR, I'll update

PR #1263, opened 2026-06-11
Fix hard coded land fraction name in coupled stepper

### LINKED ISSUE #1264: `CoupledOceanFractionConfig` assumes configured `land_fraction_name` and `sea_ice_fraction_name` are in `OCEAN_FIELD_NAME_PREFIXES` values
By design, `CoupledOceanFractionConfig` assumes that the user-specified names `land_fraction_name` and `sea_ice_fraction_name` are included in the lists `OCEAN_FIELD_NAME_PREFIXES["land_fraction"]` (for `land_fraction_name`) and either `OCEAN_FIELD_NAME_PREFIXES["sea_ice_fraction"]` or `OCEAN_FIELD_NAME_PREFIXES["ocean_sea_ice_fraction"]` (for `sea_ice_fraction_name`). 

While intentional, this is confusing and restrictive for outside users with data names that have not yet been added to the appropriate `OCEAN_FIELD_NAME_PREFIXES` value list.

elynnwu:
Fix coupled ocean-fraction prediction when the configured atmosphere land-fraction field is not named `land_fraction`, for example `LANDFRAC`.

The coupled stepper already uses `stepper.ocean_fraction_prediction.land_fraction_name` to retrieve the land-fraction tensor from atmosphere forcing data. However, it then passed that tensor into `OceanData` under the configured data name. Since `OceanData` expects canonical semantic names, later `ocean_data.land_fraction` lookup could fail with `KeyError: 'land_fraction'`.

This change makes `CoupledOceanFractionConfig.build_ocean_data` act as the adapter between configured data names and `OceanData` names: it reads land fraction using the configured atmosphere field name, then passes it to `OceanData` as canonical `land_fraction`. It also validates during `__post_init__` that the configured ocean sea-ice field is registered in `OCEAN_FIELD_NAME_PREFIXES` as either `sea_ice_fraction` or `ocean_sea_ice_fraction`, so invalid sea-ice names fail early with a clearer error.

Changes:
- `fme.coupled.stepper.CoupledOceanFractionConfig`: document the naming contract for land fraction and sea ice fraction, and validate the configured sea-ice fraction name in `__post_init__`.
- `fme.coupled.stepper.CoupledOceanFractionConfig.build_ocean_data`: canonicalize the land-fraction tensor as `land_fraction` before constructing `OceanData`, while preserving `OceanData`’s registered sea-ice semantics.
- `fme.coupled.test_stepper`: add coverage for a non-default land-fraction name such as `LANDFRAC`, and for early validation of unregistered sea-ice fraction names.

- [x] Tests added

Resolved #1264

### DIFF
--- a/fme/coupled/stepper.py
+++ b/fme/coupled/stepper.py
@@ -152,17 +152,20 @@ class CoupledOceanFractionConfig:
         # fill nans with 0s
         sea_ice_frac = torch.nan_to_num(forcings_from_ocean[sea_ice_frac_name])
         land_frac = atmos_forcing_data[land_frac_name]
-        ocean_field_name_prefixes = {
-            k: list(v) for k, v in OCEAN_FIELD_NAME_PREFIXES.items()
-        }
-        ocean_field_name_prefixes["land_fraction"] = [land_frac_name]
-        if sea_ice_frac_name == "sea_ice_fraction":
-            ocean_field_name_prefixes["sea_ice_fraction"] = [sea_ice_frac_name]
+        if sea_ice_frac_name in OCEAN_FIELD_NAME_PREFIXES["sea_ice_fraction"]:
+            canonical_sea_ice_frac_name = "sea_ice_fraction"
+        elif sea_ice_frac_name in OCEAN_FIELD_NAME_PREFIXES["ocean_sea_ice_fraction"]:
+            canonical_sea_ice_frac_name = "ocean_sea_ice_fraction"
         else:
-            ocean_field_name_prefixes["ocean_sea_ice_fraction"] = [sea_ice_frac_name]
+            raise ValueError(
+                f"CoupledOceanFractionConfig expected {sea_ice_frac_name} to be "
+                "registered in OCEAN_FIELD_NAME_PREFIXES as a sea ice fraction."
+            )
         return OceanData(
-            {land_frac_name: land_frac, sea_ice_frac_name: sea_ice_frac},
-            ocean_field_name_prefixes=ocean_field_name_prefixes,
+            {
+                "land_fraction": land_frac,
+                canonical_sea_ice_frac_name: sea_ice_frac,
+            }
         )
 
 
--- a/fme/coupled/stepper.py
+++ b/fme/coupled/stepper.py
@@ -42,7 +42,7 @@ from fme.core.generics.optimization import OptimizationABC
 from fme.core.generics.train_stepper import TrainOutputABC, TrainStepperABC
 from fme.core.loss import StepLoss, StepLossConfig
 from fme.core.ocean import OceanConfig
-from fme.core.ocean_data import OceanData
+from fme.core.ocean_data import OCEAN_FIELD_NAME_PREFIXES, OceanData
 from fme.core.optimization import NullOptimization
 from fme.core.tensors import add_ensemble_dim, unfold_ensemble_dim
 from fme.core.timing import GlobalTimer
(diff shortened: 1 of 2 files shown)


fme/coupled/stepper.py
```diff
@@ -152,7 +152,18 @@ def build_ocean_data(
...
+        }
+        ocean_field_name_prefixes["land_fraction"] = [land_frac_name]
+        if sea_ice_frac_name == "sea_ice_fraction":
+            ocean_field_name_prefixes["sea_ice_fraction"] = [sea_ice_frac_name]
+        else:
+            ocean_field_name_prefixes["ocean_sea_ice_fraction"] = [sea_ice_frac_name]
+        return OceanData(
+            {land_frac_name: land_frac, sea_ice_frac_name: sea_ice_frac},
+            ocean_field_name_prefixes=ocean_field_name_prefixes,
+        )
```
jpdunc23, line 182: A simpler bandaid fix which only requires knowledge of the canonical name for "land_fraction":

```suggestion
        return OceanData({"land_fraction": land_frac, sea_ice_frac_name: sea_ice_frac})
```

You could also add explicit validation that the configured `sea_ice_frac_name` is in `OCEAN_FIELD_NAME_PREFIXES` to `CoupledOceanFractionConfig.__post_init__`.

Personally, I prefer to just add `LANDFRAC` to `OCEAN_FIELD_NAME_PREFIXES["land_fraction"]` (and any similar changes to `OCEAN_FIELD_NAME_PREFIXES["sea_ice_fraction"]` and `OCEAN_FIELD_NAME_PREFIXES["ocean_sea_ice_fraction"]`) to support E3SM naming conventions; this is what I assumed we'd do when I wrote this method.

fme/coupled/stepper.py
```diff
@@ -152,7 +152,18 @@ def build_ocean_data(
...
         land_frac = atmos_forcing_data[land_frac_name]
-        return OceanData({land_frac_name: land_frac, sea_ice_frac_name: sea_ice_frac})
+        ocean_field_name_prefixes = {
+            k: list(v) for k, v in OCEAN_FIELD_NAME_PREFIXES.items()
+        }
+        ocean_field_name_prefixes["land_fraction"] = [land_frac_name]
+        if sea_ice_frac_name == "sea_ice_fraction":
+            ocean_field_name_prefixes["sea_ice_fraction"] = [sea_ice_frac_name]
+        else:
+            ocean_field_name_prefixes["ocean_sea_ice_fraction"] = [sea_ice_frac_name]
```
jpdunc23: I'm concerned about the fallback of treating any non-"sea_ice_fraction" name as "ocean_sea_ice_fraction". Previously, names other than "sea_ice_fraction" and "ocean_sea_ice_fraction" would fail with `KeyError`. We're also reimplementing some of `OceanData` internals here which has its own sea ice name resolution logic.

jpdunc23 (commented):
Thanks for the bug fix.

My preference is to simply update `OCEAN_FIELD_NAME_PREFIXES` as needed to support E3SM naming conventions, rather than try to support more general naming conventions now. It would also be good to document the assumptions that this config class makes, namely that the configured names must be registered in `OCEAN_FIELD_NAME_PREFIXES`.

Regardless, I understand that this assumption is not clear and is somewhat restrictive. I added https://github.com/ai2cm/ace/issues/1264 to track.

fme/coupled/stepper.py
elynnwu: I updated this to use the simpler adapter approach: `CoupledOceanFractionConfig` still uses the configured `land_fraction_name` to retrieve the tensor from atmosphere forcing data, but then passes it to `OceanData` under the canonical `land_fraction` key.

For completeness, I'll also add `LANDFRAC` to `OCEAN_FIELD_NAME_PREFIXES["land_fraction"]` in the other E3SM config specific PR. I just think it's really confusing that the config clearly specifies land fraction name but `OceanData` doesn't know about it.

fme/coupled/stepper.py
elynnwu: Good point, I agree. Removed the “anything other than sea_ice_fraction is ocean_sea_ice_fraction” behavior.

fme/coupled/stepper.py
```diff
@@ -152,7 +152,21 @@ def build_ocean_data(
...
-        return OceanData({land_frac_name: land_frac, sea_ice_frac_name: sea_ice_frac})
+        if sea_ice_frac_name in OCEAN_FIELD_NAME_PREFIXES["sea_ice_fraction"]:
+            canonical_sea_ice_frac_name = "sea_ice_fraction"
+        elif sea_ice_frac_name in OCEAN_FIELD_NAME_PREFIXES["ocean_sea_ice_fraction"]:
+            canonical_sea_ice_frac_name = "ocean_sea_ice_fraction"
+        else:
+            raise ValueError(
+                f"CoupledOceanFractionConfig expected {sea_ice_frac_name} to be "
+                "registered in OCEAN_FIELD_NAME_PREFIXES as a sea ice fraction."
+            )
```
jpdunc23: Could we please do this in `__post_init__` to get some earlier feedback?

jpdunc23 (approved):
Thank you for the changes. Approved, but please move the `sea_ice_frac_name` validation to `__post_init__`.

fme/coupled/stepper.py
elynnwu: Done

PR #1265, opened 2026-06-12
Add gradient clipping to optimizer

yyexela:
Adds `max_grad_norm` gradient clipping to `Optimization` and `OptimizationConfig`. Useful for stabilizing training when gradients occasionally spike (e.g. early training, high LR, or long rollouts).

## How it works

- Clipping is applied as a global norm clip immediately before each optimizer step.
- With AMP enabled, `GradScaler.unscale_()` is called before clipping so gradients are in real (not scaled) space — this is the standard PyTorch-recommended pattern.
- With `use_gradient_accumulation`, clipping is applied to the full N-step accumulated gradient (i.e. what the optimizer sees), not per sub-step. Users should ensure their per-step loss is appropriately scaled if mixing accumulation with clipping.
- After each step, `Optimization._last_grad_norm` holds the pre-clip global norm, available for logging.

## Changes

- `Optimization`: new `max_grad_norm: float | None` parameter; `_clip_gradients()` method called from `step_weights()`; `_last_grad_norm` attribute stores the pre-clip norm
- `OptimizationConfig`: new `max_grad_norm: float | None = None` field, wired through to `Optimization`

## Tests

- `test_gradient_clipping` — verifies clipping fires when norm exceeds threshold and is a no-op when `max_grad_norm=None`
- `test_gradient_clipping_with_amp` — verifies `gscaler.unscale_()` is called before clipping (CUDA-skip-guarded)
- `test_optimization_config_passes_max_grad_norm` — verifies config wiring

- [x] Tests added


mcgibbon (approved):
LGTM

PR #1266, opened 2026-06-12
Move `corrector_disabled_epochs` scheduling into corrector

### LINKED ISSUE #1261: Move `corrector_disabled_epochs` scheduling into corrector layer
## Generalize epoch-dependent corrector enablement across Steps

PR #1260 adds `corrector_disabled_epochs` for `SingleModuleStep`, allowing the corrector to be disabled during the first N training epochs while still applying during eval/validation/inference.

A review comment raised that this behavior may be better owned by the corrector itself rather than by individual `StepABC` implementations. Right now, the logic is specific to `SingleModuleStep`, but other steps also use correctors, including:

- `SeparateRadiationStep`
- `SecondaryModuleStep`
- `FCN3Step`
- `InfillPredictionStep` from #1224

If the scheduling logic stays inside each Step, every new corrector-using Step needs to remember to copy/support the same epoch/train/eval behavior.

## Proposal

Move epoch/train/eval-aware corrector enablement into the corrector layer, or introduce a small corrector wrapper/schedule object that can:

- receive `set_epoch(epoch)`
- receive train/eval mode changes
- decide whether to apply or skip the wrapped corrector
- persist any needed state for mid-epoch checkpoint resume

Then each Step can continue calling the corrector normally, and the behavior is shared consistently.

elynnwu:
These changes move `corrector_disabled_epochs` out of `SingleModuleStep` and into the shared corrector configuration layer. Correctors are now wrapped by `EpochScheduledCorrector`, which can disable any corrector during the first N training epochs while still applying it during eval/validation/inference.
This makes the behavior available to atmosphere, ocean, ice, and future correctors without copying logic into each Step. Step train/eval, epoch, and checkpoint state now forward through the corrector lifecycle so mid-epoch resume preserves the scheduler state.

Resolves #1261

### DIFF
--- a/fme/ace/stepper/test_single_module.py
+++ b/fme/ace/stepper/test_single_module.py
@@ -7,7 +7,7 @@ import unittest
 import unittest.mock
 from collections import namedtuple
 from collections.abc import Iterable, Mapping
-from typing import Literal
+from typing import Any, Literal
 from unittest.mock import patch
 
 import cftime
@@ -1293,6 +1293,18 @@ class _RecordingCorrector(CorrectorABC):
     def __init__(self):
         self.call_count = 0
         self.seen_states: list[CorrectorState | None] = []
+
+    def train(self, mode: bool = True) -> "_RecordingCorrector":
+        return self
+
+    def set_epoch(self, epoch: int) -> None:
+        pass
+
+    def get_state(self) -> dict[str, Any]:
+        return {}
+
+    def load_state(self, state: dict[str, Any]) -> None:
+        pass
 
     def __call__(
         self,
--- a/fme/ace/stepper/test_single_module.py
+++ b/fme/ace/stepper/test_single_module.py
@@ -2139,8 +2139,10 @@ def _get_force_positive_stepper(corrector_disabled_epochs: int) -> Stepper:
     return _get_stepper(
         ["a"],
         ["a"],
-        corrector=AtmosphereCorrectorConfig(force_positive_names=["a"]),
-        corrector_disabled_epochs=corrector_disabled_epochs,
+        corrector=AtmosphereCorrectorConfig(
+            force_positive_names=["a"],
+            corrector_disabled_epochs=corrector_disabled_epochs,
+        ),
     )
 
 
--- a/fme/core/corrector/registry.py
+++ b/fme/core/corrector/registry.py
@@ -12,15 +12,6 @@ from fme.core.typing_ import TensorDict, TensorMapping
 
 @dataclasses.dataclass
 class CorrectorConfigABC(abc.ABC):
-    corrector_disabled_epochs: int = dataclasses.field(default=0, kw_only=True)
-
-    def __post_init__(self):
-        if self.corrector_disabled_epochs < 0:
-            raise ValueError(
-                "corrector_disabled_epochs must be non-negative, got "
-                f"{self.corrector_disabled_epochs}"
-            )
-
     @classmethod
     @final
     def from_state(cls, state: Mapping[str, Any]) -> Self:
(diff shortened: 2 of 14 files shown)


fme/core/corrector/registry.py
```diff
@@ -25,14 +36,48 @@ def remove_deprecated_keys(cls, state: Mapping[str, Any]) -> dict[str, Any]:
...
+
+    def eval(self) -> "CorrectorABC":
+        """Set the corrector to evaluation mode."""
+        return self.train(False)
+
+    def set_epoch(self, epoch: int) -> None:
+        """Called by the stepper at the start of each training epoch."""
+        pass
+
+    def get_state(self) -> dict[str, Any]:
```
mcgibbon, line 82: Suggestion: Make all methods on this ABC `@abstract` (only implemented on the children) or `@final` (only implementable on the base). Likely they should all be abstract. It will help to not have to scan back and forth between this ABC and the subclasses to understand which methods are implemented where.

fme/core/step/radiation.py
```diff
@@ -418,3 +422,5 @@ def load_state(self, state: dict[str, Any]) -> None:
         """
         self.module.load_state(state["module"])
         self.radiation_module.load_state(state["radiation_module"])
+        if "corrector" in state:
+            self._load_corrector_state(state["corrector"])
```
mcgibbon: Question: Would you want to call `_load_corrector_state({})` if the saved corrector had no state? For example to raise if somehow we had a has-state corrector and were trying to load a no-corrector-state checkpoint?

fme/core/corrector/registry.py
elynnwu: Updated

fme/core/step/radiation.py
elynnwu: Yes, udpated

fme/ace/stepper/test_single_module.py
```diff
@@ -1294,6 +1294,18 @@ def __init__(self):
         self.call_count = 0
         self.seen_states: list[CorrectorState | None] = []
 
+    def train(self, mode: bool = True) -> "_RecordingCorrector":
```
mcgibbon, line 1297: Question: Should there be tests exercising these new methods? I didn't notice earlier but I think an AI reviewer would call it out.

elynnwu: Added

fme/core/corrector/registry.py
```diff
@@ -9,7 +10,17 @@
 from fme.core.typing_ import TensorDict, TensorMapping
 
 
+@dataclasses.dataclass
 class CorrectorConfigABC(abc.ABC):
+    corrector_disabled_epochs: int = dataclasses.field(default=0, kw_only=True)
```
mcgibbon: Chore: Remove this property from the ABC, right? Is it required for all CorrectorConfig?

mcgibbon (approved):
LGTM, one possible nit/chore to look at.

fme/core/step/step.py
```diff
@@ -255,13 +255,30 @@ def train(self, mode: bool = True) -> "StepABC":
...
             module.train(mode)
+        self._set_corrector_train_mode(mode)
         return self
 
     @final
     def eval(self) -> "StepABC":
         """Set the step (and all submodules) to evaluation mode."""
         return self.train(False)
 
+    def _set_corrector_train_mode(self, mode: bool) -> None:
```
mcgibbon: getattr is a red flag, and I'm realizing from it that I didn't notice this is on the ABC. The corrector is specific to the subclass Step, shouldn't this code go there? That means `train` should be abstract instead of final - you could make a case for that one making an exception and allowing a default, but if you do that, at least use the super-call version to do the base part instead of copy-pasting.

mcgibbon (changes requested):

fme/core/corrector/registry.py
elynnwu: Removed

Issue #1267, opened 2026-06-12
ACE Aggregator Flaky Test?


frodre:
Failing in #1236, but seems unrelated

```
=================================== FAILURES ===================================
_____________________________ test_logs_regression _____________________________
[gw0] linux -- Python 3.11.15 /home/runner/work/ace/ace/.venv/bin/python3

    def test_logs_regression():
        torch.manual_seed(0)
        n_sample = 10
        n_time = 24
        nx = 90
        ny = 45
        ds_info = get_ds_info(nx, ny)
        initial_time = get_zero_time(shape=[n_sample, 0], dims=["sample", "time"])
    
        agg = build_inference_evaluator_aggregator(
            metrics=[
                MeanMetricConfig(target="denorm"),
                MeanMetricConfig(target="norm"),
                StepMeanMetricConfig(step=20, target="denorm"),
                StepMeanMetricConfig(step=20, target="norm"),
                StepMeanMetricConfig(step=4, name="one_day_mean", target="denorm"),
                StepMeanMetricConfig(step=4, name="one_day_mean_norm", target="norm"),
                PowerSpectrumMetricConfig(),
                ZonalMeanMetricConfig(zonal_mean_max_size=LOG_ZONAL_MEAN_IMAGES),
                TimeMeanMetricConfig(target="denorm"),
                TimeMeanMetricConfig(target="norm"),
            ],
            dataset_info=ds_info,
            n_ic_steps=1,
            n_forward_steps=n_time - 1,
            initial_time=initial_time,
            normalize=lambda x: dict(x),
            save_diagnostics=False,
        )
        time = xr.DataArray(np.zeros((n_sample, n_time)), dims=["sample", "time"])
    
        logs = agg.record_batch(
            data=PairedData.new_on_device(
                prediction={"a": torch.randn(n_sample, n_time, ny, nx).to(get_device())},
                reference={"a": torch.randn(n_sample, n_time, ny, nx).to(get_device())},
                time=time,
                labels=None,
            ),
        )
        assert len(logs) == n_time
        expected_step_keys = [
            "mean/forecast_step",
            "mean/weighted_mean_gen/a",
            "mean/weighted_mean_target/a",
            "mean/weighted_rmse/a",
            "mean/weighted_std_gen/a",
            "mean/weighted_bias/a",
            "mean/weighted_grad_mag_percent_diff/a",
            "mean_norm/forecast_step",
            "mean_norm/weighted_mean_gen/a",
[...]

PR #1268, opened 2026-06-12
Fix test_logs_regression tolerance for matplotlib rendering changes

yyexela:
Fixes a CI break caused by the matplotlib 3.11.0 release on 2026-06-12.

## Root cause

matplotlib 3.11.0 renders colormapped images with ±1 uint8 pixel differences relative to 3.10.x. The regression test compared rendered `Image` tensors with `atol=1e-3`, which is effectively zero tolerance for uint8 data, so any pixel shift by 1 fails.

## Changes

Instead of regenerating the baseline (which would break again on the next matplotlib release), the comparison now uses a uint8-appropriate tolerance:
- **Image tensors**: `atol=2, rtol=0` — tolerates ±1 pixel rendering differences across matplotlib versions
- **Float metrics**: unchanged at `atol=1e-3, rtol=1e-3`

This makes the test robust to minor rendering-stack changes without requiring baseline regeneration on every matplotlib release.

## Checklist

- [x] Tests added


yyexela:
Addresses https://github.com/ai2cm/ace/issues/1267

frodre (approved):
LGTM, thanks

PR #1270, opened 2026-06-12
Fix pin_memory crash for inference persistence_names variables

mcgibbon:
Inline inference runs configured with `persistence_names` (e.g. holding `global_mean_co2` constant at its initial-condition value) crash in the DataLoader pin-memory thread at the first inference epoch. `InferenceDataset.__getitem__` returns persisted variables as `expand_as` views with stride 0; pin_memory allocates a pinned tensor with the same strides and copies into it, which fails for overlapping memory with `RuntimeError: unsupported operation: more than one element of the written-to tensor refers to a single memory location.` This PR makes the expanded tensor contiguous before it is returned.

Changes:
- `fme.ace.data_loading.inference.InferenceDataset.__getitem__`: call `.contiguous()` on the `expand_as` view for persisted variables so the result can be memory-pinned
- `test_inference_persistence_names_tensors_support_pin_memory`: regression test performing the same layout-preserving copy as the pin-memory thread

- [x] Tests added

### DIFF
--- a/fme/ace/data_loading/test_data_loader.py
+++ b/fme/ace/data_loading/test_data_loader.py
@@ -983,7 +983,8 @@ def test_inference_persistence_names_tensors_support_pin_memory(tmp_path):
     )
     foo = dataset[0].data["foo"]
     # same layout-preserving copy that torch.utils.data pin_memory performs
-    torch.empty_strided(foo.size(), foo.stride()).copy_(foo)
+    result = torch.empty_strided(foo.size(), foo.stride()).copy_(foo)
+    torch.testing.assert_close(result, foo)
 
 
 def test_inference_dataset_label_override_without_dataset_labels(tmp_path):
--- a/fme/ace/data_loading/test_data_loader.py
+++ b/fme/ace/data_loading/test_data_loader.py
@@ -957,6 +957,35 @@ def test_inference_persistence_names(tmp_path):
     assert not torch.all(first_item["bar"] == second_item["bar"])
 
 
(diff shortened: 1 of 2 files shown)


fme/ace/data_loading/test_data_loader.py
```diff
@@ -957,6 +957,35 @@ def test_inference_persistence_names(tmp_path):
...
+        n_timesteps=3,
+    )
+    dataset = InferenceDataset(
+        config,
+        9,
+        requirements=window_requirements,
+    )
+    foo = dataset[0].data["foo"]
+    # same layout-preserving copy that torch.utils.data pin_memory performs
+    torch.empty_strided(foo.size(), foo.stride()).copy_(foo)
```
yyexela: Nitpick (optional): I think this is good, but it would be a _bit_ better if the test didn't end without actually checking something

Suggestion: Have this at the end if this test instead:

```
result = torch.empty_strided(foo.size(), foo.stride()).copy_(foo)
torch.testing.assert_close(result, foo)
```

yyexela (commented):
Looks great otherwise! Small nitpick that you can decide what to do with.

fme/ace/data_loading/test_data_loader.py
mcgibbon: Updated

mcgibbon: Good call, done in c4040b5f4 — the test now asserts the copied result matches the source.

yyexela (approved):
Approved:)

PR #1274, opened 2026-06-12
Update plot_events script

AnnaKwa:
- Add option to cache the downloaded dataset instead of saving to temp directory. Currently it defaults to caching so that it's less tedious to call the script, but if desired I can change the default to not cache.
- Fix bug where the coarse data would select the 00 hour timestep if the event filename only had YYYYMMDD in the name. Now the script checks the config in the beaker dataset to get the exact event timestamp.
- Add TMP2m to the coarse variables to read
- Remove coarse PRESsfc relabeling to PRMSL since we now use PRMSL for both inputs and outputs.


scripts/downscaling/utils.py
```diff
@@ -6,20 +6,51 @@
...
+        cmd += ["--prefix", prefix]
+    subprocess.run(cmd, check=True)
+    return target_dir
 
 
 def find_event_files(directory: str) -> dict[str, Path]:
-    """Find netCDF files matching the event naming pattern, keyed by event name."""
+    """Find netCDF files matching the event naming pattern, keyed by filename
+    stem (event name including date, so multiple dates of the same event are
+    kept)."""
```
AnnaKwa, line 63: Before, if there were files with the same name and different dates, only one was plotted.

scripts/downscaling/plot_events.py
```diff
@@ -183,11 +188,6 @@ def plot_event(ds, var_name, samples=None, sel=None, n_cols=5, **plot_kwargs):
     if len(samples) == 0:
         samples = [0]
 
-    if var_name == "PRMSL":
-        # fill PRMSL_coarse with nans
-        ds_["PRMSL_coarse"].values[:] = np.nan
-        # For colorbar range, use only target and predicted (coarse is hidden)
-        arr = ds_[["PRMSL_target", "PRMSL_predicted"]].to_array()
     else:
```
frodre: Does this `else` block need to be adjusted? It looks like it pairs with the `if len(samples)...` from above

frodre: Does this `else` block need to be adjusted? It looks like it pairs with the `if len(samples)...` from above

scripts/downscaling/utils.py
```diff
@@ -6,20 +6,51 @@
...
-def fetch_beaker_dataset(dataset_id: str, target_dir: str) -> None:
-    """Fetch a beaker dataset to the specified directory."""
-    subprocess.run(
-        ["beaker", "dataset", "fetch", dataset_id, "--output", target_dir],
-        check=True,
-    )
+def fetch_beaker_dataset(
+    dataset_id: str,
+    target_dir: str,
+    prefix: str | None = None,
```
frodre, line 15: Prefix seems like a straightforward addition, but I didn't see it actually used anywhere.

scripts/downscaling/utils.py
```diff
@@ -6,20 +6,51 @@
...
+    cache_dir: str | None = "~/Downloads/beaker_cache",
```
frodre: I think persistent cache would make more sense as opt-in not opt-out.

scripts/downscaling/utils.py
```diff
@@ -6,20 +6,51 @@
...
+            When a cached copy exists the download is skipped and files
+            are served from the cache. *target_dir* is ignored when a
+            cache hit occurs.
+
+    Returns:
+        The directory containing the fetched dataset files.
+    """
+    if cache_dir is not None:
+        cached = Path(cache_dir).expanduser() / dataset_id
+        if cached.is_dir() and any(cached.iterdir()):
```
frodre: One thing for persistent caches flagged by Claude: a partial beaker fetch that failed would still pass this check. You could add a sentinel file after the subprocess completes successfully.

frodre (changes requested):
I think there's one leftover error from the adjusted PRMSL conditional, and I think I would prefer an opt in for the caching instead of defaulting to storage (I don't feel strongly).  My reasoning being that the datasets are usually small and quick to download.  At the very least if not making opt-in, I would suggest the default cache be tied to something that gets wiped with the machine reset like /tmp/beaker, but would still be persistent across script usage.

scripts/downscaling/utils.py
AnnaKwa: It gets used in L43

scripts/downscaling/utils.py
AnnaKwa: I changed the default to opt in but added a check for an env var that can be set if someone (me) uses this often and doesn't want to add the cli arg each time.

scripts/downscaling/utils.py
frodre: Sorry, I meant that it doesn't seem to be wired into any of the entrypoint usage, so it's effectively blocked off.

frodre (approved):
LGTM, thanks for the cache adjustments.

PR #1278, opened 2026-06-15
Fix standalone-inference ensemble double-broadcast

mcgibbon:
## Summary

Standalone inference (`run_inference_from_config`) crashes with `DataShapesNotUniform` when `n_ensemble_per_ic > 1`, because the forcing data is expanded across ensemble members twice.

## The bug

`get_initial_condition(n_ensemble=N)` tiles the initial condition to `N` members. `get_forcing_data` then reads those already-tiled times and builds one forcing window per member — but labels the forcing `n_ensemble=1`. Because of that stale label, `predict_paired` broadcasts the forcing a *second* time, inflating it from `N` to `N**2` members while the initial condition stays at `N`. The mismatched sample counts then fail in the packer.

The inline-training evaluator path is unaffected: it broadcasts the initial condition *after* building the forcing loader, so its forcing genuinely has `n_ensemble=1` and is broadcast exactly once.

## Impact

No reported numerical results are affected — the run fails fast rather than silently miscomputing. The second broadcast is also pure redundancy: it *copies* existing forcing windows, so the inflated members are exact duplicates. Even without the crash it would cost ~N× the inference compute for identical results.

## Fix

Thread the initial condition's `n_ensemble` through `get_forcing_data` -> `get_inference_data` -> `InferenceDataset` -> `BatchData.from_sample_tuples`, so the forcing reports the correct `n_ensemble` and `predict_paired` no longer re-broadcasts it. The parameter defaults to `1`, so the evaluator, training, and coupled callers are unchanged.

Changes:
- `fme.ace.data_loading.getters.get_forcing_data` — reads the initial condition's `n_ensemble` and passes it to `get_inference_data`.
- `fme.ace.data_loading.getters.get_inference_data` / `fme.ace.data_loading.inference.InferenceDataset` — new `n_ensemble` parameter (default `1`), used to stamp each forcing batch.
- `fme.ace.data_loading.batch_data.BatchData.from_sample_tuples` — new `n_ensemble` parameter (default `1`), forwarded to the constructed `BatchData`.
- `test_inference_entrypoint` — parametrized over `n_ensemble_per_ic` in `[1, 2]`; the `2` case reproduces the crash without the fix.
- `test_get_forcing_data_broadcasts_ensemble_flag` — focused regression: forcing built from an ensemble-tiled IC must carry the IC's `n_ensemble`.

## How it was found

Surfaced while setting up the blowup-investigation probes, which worked around it by building forcing from a single-sample initial condition. No production run is on record hitting it — standalone inference had not been exercised with `n_ensemble_per_ic > 1`.

- [x] Tests added

### DIFF
--- a/fme/ace/data_loading/batch_data.py
+++ b/fme/ace/data_loading/batch_data.py
@@ -399,6 +399,7 @@ class BatchData:
         horizontal_dims: list[str] | None = None,
         label_encoding: LabelEncoding | None = None,
         allow_missing_variables: bool = False,
+        n_ensemble: int = 1,
     ) -> "BatchData":
         (
             sample_data,
@@ -429,6 +430,7 @@ class BatchData:
             labels=labels,
             horizontal_dims=horizontal_dims,
             epoch=sample_epochs[0],
+            n_ensemble=n_ensemble,
             data_mask=data_mask,
         )
 
--- a/fme/ace/data_loading/getters.py
+++ b/fme/ace/data_loading/getters.py
@@ -171,6 +171,7 @@ def get_inference_data(
     label_override: list[str] | None = None,
     surface_temperature_name: str | None = None,
     ocean_fraction_name: str | None = None,
+    n_ensemble: int = 1,
     _force_forkserver: bool = False,
 ) -> InferenceGriddedData:
     """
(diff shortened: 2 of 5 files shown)


Arcomano1234 (approved):
LGTM. Thanks for fixing this bug. 

I did go back into the previous PRs related to ensemble inference and figured out what happened. Previously, we did not want to thread `n_ensemble` into the getters but I just realized that inference vs evaluator are pretty different on how they build forcings. So I think it is basically not possible to do this without having getters know the `n_ensembles` unless we change `inference.py` to better mirror how we do the evaluator.

mcgibbon:
> LGTM. Thanks for fixing this bug.
> 
> I did go back into the previous PRs related to ensemble inference and figured out what happened. Previously, we did not want to thread `n_ensemble` into the getters but I just realized that inference vs evaluator are pretty different on how they build forcings. So I think it is basically not possible to do this without having getters know the `n_ensembles` unless we change `inference.py` to better mirror how we do the evaluator.

Good catch @Arcomano1234 , I made this alternative PR which does it the way you suggested, much smaller: https://github.com/ai2cm/ace/pull/1285

mcgibbon:
Superceded by #1285

PR #1279, opened 2026-06-15
Fix concurrent inference crash on mismatched window shapes

mcgibbon:
Concurrent inline-inference runs of different lengths share one `BatchedPredictor`. A shorter run (e.g. `10year`, 3652 steps = 50×73 + 2) reaches its short final window in the same round a longer run (e.g. `long_46year`, 16794 steps) still has a full window. Those windows have different `n_timesteps` (3 vs 74), so concatenating them into one forward pass raised `ValueError: Cannot cat BatchData with different n_timesteps: 3 != 74`, crashing training before epoch 1 finished. This satisfies the requirement that concurrent inline inference run correctly for heterogeneous-length tasks that share a `concurrent_group`.

The fix partitions each `BatchedPredictor` flush into groups whose forcing windows share a batch key, running one forward pass per group and reassembling results per slot. The common all-equal-shape case still runs a single forward pass, so concurrency is preserved except in the rounds that genuinely mix window shapes (where the mismatched window simply gets its own pass that round).

Changes:
- `fme.core.generics.inference.BatchedPredictor`: new `batch_key_of_forcing` callable (default keys everything the same, preserving prior single-pass behavior); `_ensure_flushed` now partitions pending submissions by key and runs one forward pass per distinct window shape.
- `fme.ace.train.train.get_inference_callback` / `fme.coupled.train.train`: wire `batch_key_of_forcing` to the forcing window's `n_timesteps` (ACE) and ocean+atmosphere `n_timesteps` (coupled).
- `fme.core.generics.test_batched_predictor`: failing-first regression test reproducing the short-final-window vs full-window collision, plus a unit test of the partition logic.

- [x] Tests added

### DIFF
--- a/fme/ace/train/train.py
+++ b/fme/ace/train/train.py
@@ -223,6 +223,7 @@ def get_inference_callback(
             split_output=lambda sd, sizes: sd.split(sizes),
             split_state=lambda ps, sizes: ps.split(sizes),
             sample_size_of_state=lambda ic: ic.as_batch_data().time.shape[0],
+            batch_key_of_forcing=lambda forcing: forcing.n_timesteps,
         )
 
     return build_inference_callback(
--- a/fme/core/generics/inference.py
+++ b/fme/core/generics/inference.py
@@ -1,5 +1,5 @@
 import logging
-from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Callable, Hashable, Iterator, Sequence
 from typing import Any, Generic, Protocol, TypeVar
 
 from fme.core.generics.aggregator import InferenceAggregatorABC, InferenceLogs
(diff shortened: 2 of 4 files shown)


mcgibbon:
Claude: pre-review (independent, read cold against base `feature/concurrent-inline-inference`).

**Audit passes**
- ✅ Silent-failure: new `batch_key_of_forcing` default (`lambda _forcing: None`) collapses all submissions to one group, preserving prior single-pass behavior; no path silently drops a submission. Other `cat` invariants (`n_ensemble`, `epoch`, variable names, labels/data_mask presence) are already pinned within a `concurrent_group` (`(forward_steps_in_memory, n_ensemble_per_ic)`), so `n_timesteps` is the only axis that legitimately varies within a group — the right thing to key on.
- ✅ Behavior-change: only `_ensure_flushed` changes. Default callers group under key `None` → identical merged forward pass + split as before; the `len(slot_ids) == 1` fast path matches the old single-submission path. Results reassembled by absolute slot_id, so group iteration order is irrelevant to correctness.
- ✅ Differential-test: confirmed by no-oping the partition (forcing one group) — both new tests fail with the real `ValueError: Cannot cat ... n_timesteps: 3 != 74`; restored. Tests genuinely distinguish the fix.
- ✅ Sibling/mirror coverage: only two `BatchedPredictor` construction sites (ACE, coupled); both wired. No diffusion/downscaling sites. Coupled keys on both `ocean_data.n_timesteps` and `atmosphere_data.n_timesteps` (tuple), correctly distinguishing a short final window in either component.
- ✅ Type/lint: `pre-commit` (ruff, ruff-format, mypy) clean on all four files. `forcing.n_timesteps` and `CoupledBatchData.ocean_data/.atmosphere_data.n_timesteps` all exist on the real types passed to `submit`.
- ✅ Tests green: `test_batched_predictor.py` (18) and `test_inference_callback.py` (5) pass CPU-only.
- ✅ Description accuracy: matches the diff (function locations, two added tests + helper changes).

**What I changed**
- Dropped the opaque `rs1` reference from the regression-test docstring (author-context token, meaningless to a reviewer; the rest already explains the scenario). Pushed as d6631bb.

No open questions. The callback-layer test file does not add a mismatched-window case, but that scenario belongs at the `BatchedPredictor`/`run_concurrent_inference` layer where it is now covered — not a gap.

**Verdict: ready for review.** Small, well-scoped fix with a failing-first regression test that reproduces the exact crash; correct keying axis given the existing `concurrent_group` invariants.

mcgibbon:
This fix is merged into the `experiment/2026-06-05-aimip-like` branch at `59ed0bf3c` and the rs1 concurrent-inference run is relaunched from it to verify the crash is cleared: [beaker 01KV6P5M](https://beaker.org/ex/01KV6P5MG100PTXNV436HD40AY) (`train-4deg-daily-v1-era5-only-rs1-59ed`). The `-612f` run on the prior commit failed before epoch 1 with `Cannot cat BatchData with different n_timesteps: 3 != 74`; this relaunch should train past that point.

PR #1280, opened 2026-06-15
Add CRPS to the annual aggregator, with CRPS/RMSE config toggles

mcgibbon:
The annual aggregator reports R2 and (as of #1257) ensemble-mean RMSE of the predicted annual-mean evolution. This adds CRPS, a probabilistic skill measure of the same annual evolution that, like RMSE, needs no external reference data, and makes both CRPS and RMSE independently toggleable. Both are reported by default. This is a follow-on to #1257 and targets its branch.

Changes:
- `fme.ace.aggregator.inference.annual.PairedGlobalMeanAnnualAggregator`: reports `crps/<var>`, the fair CRPS of the predicted ensemble annual-mean evolution against the target ensemble-mean evolution, alongside the existing `rmse/<var>`; new `report_crps`/`report_rmse` constructor args (default `True`) gate each metric
- `fme.ace.aggregator.inference.annual.get_crps`: helper computing fair CRPS over years between an ensemble `DataArray` and a per-year reference, reusing `fme.core.ensemble.get_crps` and ignoring NaN gap years from reindexing the way `get_rmse`/`get_r2` do
- `fme.ace.aggregator.inference.AnnualMetricConfig`: `report_crps` and `report_rmse` fields (both default `True`), validated in `__post_init__` to reject disabling both (the only reference-free skill measures); new fields default to `True`, so checkpoint/inference configs without them are unchanged

- [x] Tests added


mcgibbon:
Claude: independent pre-review (no authoring context). Reviewed the diff against the PR base `feature/annual-mean-rmse` — only `annual.py` (+ `test_annual.py`), +117/-4.

**Audit passes**
- ✅ Silent-failure: toggles default `True` (explicit); `AnnualMetricConfig.__post_init__` raises if both are `False`, so the only failure mode fails loudly at config-load time. NaN gap years are dropped via `np.nanmean`, exactly mirroring `get_rmse`/`get_r2`. The `n_ens==1` case degrades correctly to MAE (`internal_term=0`), not a silent miss.
- ✅ Behavior-change on existing paths: the existing `rmse/<var>` emission is now gated behind `report_rmse` (default `True`), so existing configs/checkpoints are unchanged; R2 and the annual-evolution plot are untouched. The sole non-test caller (`AnnualMetricConfig.build`) is updated to pass the new args. No other call sites of `PairedGlobalMeanAnnualAggregator` exist.
- ✅ Differential tests: `test_paired_annual_aggregator_metric_toggles` keys presence/absence to each flag (a no-op'd toggle fails a parametrization); `test_get_crps_ignores_nan_gap_years` pins the fair-CRPS math with a hand-computed value over multi-sample/multi-year/NaN-gap fixtures, so correctness is anchored independently of the implementation.
- ✅ AI-artifact scan: comments explain the *why* (CRPS framing, the `[n_batch, n_ensemble, ...]` dim convention), not the diff; PR description matches the code (reuse of `fme.core.ensemble.get_crps`, NaN handling, defaults, the both-False guard). No invented references.
- ✅ Split heuristic: one coherent change, no split warranted.
- (Record passes 4–7 N/A — this is a code-only PR in the `fme` package.)

**Checks run**
- `FME_FORCE_CPU=1 pytest test_annual.py test_evaluator.py -q` → 52 passed (warnings pre-existing/unrelated).
- `pre-commit run` (ruff, ruff-format, mypy) on both changed files → all passed.

**Changes I made:** none — no findings required a fix.

**Note (confirmed intentional, not blocking):** CRPS scores the generated ensemble against the *target ensemble mean* as the verifying observation, discarding the target's own spread. This mirrors how the existing RMSE compares against that same ensemble mean and is stated explicitly in the PR description and code comment, so I'm treating it as deliberate.

### Stepping back
The PR does exactly what it says: adds a reference-free probabilistic skill measure (fair CRPS) to the annual aggregator alongside RMSE, with both independently toggleable and defaulting on, and refuses a config that would leave no skill measure. It is small, self-contained, well-tested, and easy to review.

fme/ace/aggregator/inference/annual.py
```diff
@@ -393,6 +423,17 @@ class AnnualMetricConfig:
     reference_data: str | None = None
     enabled: bool = True
     strict: bool = False
+    report_crps: bool = True
+    report_rmse: bool = True
+
+    def __post_init__(self):
```
mcgibbon: Remove this post-init, it also stores the raw non-scalar values so we could in theory want to disable both of these.

mcgibbon:
Claude: Baseline beaker run exercising the new annual CRPS metric (4-degree daily ERA5, 1 epoch, jupiter+titan high priority): https://beaker.org/ex/01KV6P6DVQ12MJM9JPVXDHEE3P

It runs experiment branch `experiment/2026-06-15-annual-aggregator-crps-baseline` (commit 5fcb8e9), which integrates this PR with the AIMIP-like baseline config; the `long_46year` inference blocks set `report_crps: true`/`report_rmse: true` to verify `annual/.../crps/<var>` end-to-end before this PR is presented for review.

mcgibbon:
Wandb run https://wandb.ai/ai2cm/ace/runs/9obos9if

fme/ace/aggregator/inference/annual.py
```diff
@@ -113,9 +118,15 @@ def get_logs(self, label: str) -> dict[str, Any]:
...
-                    gen_ensemble_mean, target_ensemble_mean
-                )
+                if self._report_rmse:
+                    metrics[f"rmse/{name}"] = get_rmse(
+                        gen_ensemble_mean, target_ensemble_mean
+                    )
+                if self._report_crps:
+                    # CRPS treats the generated ensemble as the forecast and the
+                    # target ensemble mean as the verifying observation, mirroring
+                    # how RMSE compares against that same ensemble mean.
```
yyexela: Nitpick (non-blocking): I think this comment is unnecessary, it feels like it just adds bloat since code is already clean/descriptive

Suggestion: Remove comment, code is already descriptive

fme/ace/aggregator/inference/test_annual.py
```diff
@@ -230,6 +277,29 @@ def test_get_r2_ignores_nan_gap_years():
...
+    )
+    # fair CRPS per year (2 members) is mean_i|x_i - y| - 0.5 |x_0 - x_1|:
+    #   2000: members [1, 2], y=1 -> (0 + 1)/2 - 0.5 * 1 = 0.0
+    #   2002: members [3, 4], y=2 -> (1 + 2)/2 - 0.5 * 1 = 1.0
+    #   2003: members [5, 6], y=2 -> (3 + 4)/2 - 0.5 * 1 = 3.0
+    # mean over the non-NaN years 2000/2002/2003: (0 + 1 + 3)/3 = 4/3
+    expected = 4.0 / 3.0
+    result = get_crps(da, reference)
+    assert not np.isnan(result)
+    np.testing.assert_allclose(result, expected, rtol=1e-6)
```
yyexela, line 291: Question: Should 'reference' also be able to handle 'NaN'? I think a test should be added with 'nan' in 'reference' handling how you expect it to deal with it.

Suggestion: Based on the tests for `get_rmse` I think this test should have `nan` in `reference` too.

fme/ace/aggregator/inference/test_annual.py
```diff
@@ -97,6 +112,37 @@ def test_paired_annual_aggregator(tmpdir):
...
+def test_paired_annual_aggregator_metric_toggles(tmpdir, report_crps, report_rmse):
+    agg = _record_paired_annual_aggregator(
+        tmpdir, report_crps=report_crps, report_rmse=report_rmse
+    )
+    logs = agg.get_logs(label="test")
+    assert ("test/crps/a" in logs) == report_crps
+    assert ("test/rmse/a" in logs) == report_rmse
+    # the annual-evolution plot and R2 are reported regardless of the toggles
+    assert "test/a" in logs
+    assert "test/r2/a_target" in logs
```
yyexela, line 134: Question: Should`report_crps` and `report_rmse` have the option to both be disabled? Based on an earlier comment in the PR, it seems like yes.

Suggestion: Update `test_paired_annual_aggregator_metric_toggles` to handle both being False as a test. Remove earlier `__post_init__` as earlier comment mentioned.

fme/ace/aggregator/inference/test_annual.py
```diff
@@ -97,6 +112,37 @@ def test_paired_annual_aggregator(tmpdir):
     )
     np.testing.assert_allclose(logs["test/rmse/a"], expected_rmse, rtol=1e-5)
 
+    # the reported CRPS should be the fair CRPS of the predicted ensemble annual
+    # evolution against the target ensemble-mean annual evolution
+    expected_crps = get_crps(ds["a"].sel(source="prediction"), target_ensemble_mean)
+    assert not np.isnan(logs["test/crps/a"])
```
yyexela, line 117: Nitpick (non-blocking): Another test handles `assert not np.isnan(...)`.

Suggestion: Remove this line

yyexela:
Overall looks good, just some comments/ suggestions!

fme/ace/aggregator/inference/annual.py
mcgibbon: Agreed.

fme/ace/aggregator/inference/test_annual.py
mcgibbon: Let's not do this, in the case that test is failing it's better we get an assertion error on this test than a strange error (non-failure) on assert_allclose.

fme/ace/aggregator/inference/test_annual.py
mcgibbon: Agreed.

fme/ace/aggregator/inference/test_annual.py
mcgibbon: Agreed.

mcgibbon:
Addressed the review in 5925b9976:

- Removed `AnnualMetricConfig.__post_init__` so `report_crps` and `report_rmse` can both be disabled.
- Dropped the redundant CRPS comment in `get_logs`.
- Added the `(False, False)` case to `test_paired_annual_aggregator_metric_toggles` and removed the now-obsolete `test_annual_metric_config_requires_a_metric` (and its unused import).
- `test_get_crps_ignores_nan_gap_years` now also places a NaN in the `reference` to cover NaN handling on both the prediction and the reference.

Kept the `assert not np.isnan(logs["test/crps/a"])` in `test_paired_annual_aggregator` so a NaN surfaces as a clear assertion failure rather than a silent `assert_allclose` mismatch.

yyexela (approved):
Looks great! Thank you!

PR #1281, opened 2026-06-15
Add ensemble_metrics option to TrainAggregatorConfig

mcgibbon:
Ports the `ensemble_metrics` option for `TrainAggregatorConfig` from the `experiment/2026-06-05-aimip-like` branch back to `main`. When enabled, the training aggregator computes CRPS, SSR bias, and ensemble-mean RMSE over the training batches, so we can compare train vs. val SSR bias and diagnose whether ensemble overconfidence is caused by overfitting. Disabled by default, so existing configs are unaffected; enable with `train_aggregator.ensemble_metrics: true` in the training YAML.

Changes:
- `fme.ace.aggregator.train.TrainAggregatorConfig`: new `ensemble_metrics` flag (default `False`).
- `fme.ace.aggregator.train.TrainAggregator`: builds an `_EnsembleAggregator` (with `log_mean_maps=False`) when the flag is set, records each batch's native ensemble-dim target/gen data (i.e. the `EnsembleTensorDict`s as-is, with the ensemble dimension intact, rather than the folded data the paired aggregators receive), and merges its `ensemble/*` logs into the summary under `{label}/ensemble/...`.
- `fme.ace.aggregator.test_train`: tests that the metrics appear when enabled and are absent by default.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/train.py
+++ b/fme/ace/aggregator/train.py
@@ -10,6 +10,7 @@ from fme.ace.aggregator.loss_metrics import (
     PerChannelLossAggregator,
     PerStepLossAggregator,
 )
+from fme.ace.aggregator.one_step.ensemble import _EnsembleAggregator
 from fme.ace.aggregator.one_step.reduced import MeanAggregator
 from fme.ace.stepper import TrainOutput
 from fme.core.device import get_device
@@ -31,11 +32,15 @@ class TrainAggregatorConfig:
         weighted_rmse: Whether to compute the weighted RMSE.
         per_channel_loss: Whether to accumulate and report per-variable (per-channel)
             loss in get_logs (e.g. train/mean/loss/<var_name>).
+        ensemble_metrics: Whether to compute ensemble metrics (CRPS, SSR bias, and
+            ensemble-mean RMSE) over the training batches. Disabled by default;
+            requires the batch to carry an ensemble dimension.
     """
 
     spherical_power_spectrum: bool = True
     weighted_rmse: bool = True
     per_channel_loss: bool = True
+    ensemble_metrics: bool = False
 
 
 class Aggregator(Protocol):
@@ -106,6 +111,12 @@ class TrainAggregator(AggregatorABC[TrainOutput]):
                 include_bias=False,
                 include_grad_mag_percent_diff=False,
             )
+        self._ensemble_aggregator: _EnsembleAggregator | None = None
+        if config.ensemble_metrics:
+            self._ensemble_aggregator = _EnsembleAggregator(
+                gridded_operations=operations,
+                log_mean_maps=False,
+            )
 
     @torch.no_grad()
     def record_batch(self, batch: TrainOutput):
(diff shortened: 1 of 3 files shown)


mcgibbon:
Claude: Independent pre-review (read cold, no authoring context). This is a code PR, so passes 1-3, 8, 9 apply.

**Audit passes**
- 1 Silent-failure: ✅ The new `ensemble_metrics` flag is default-off and quietly no-ops when disabled — that is the documented, intended default, not a swallowed error. ⚠️ One edge case (open question below): with the flag on but `gen` ensemble dim = 1, `_EnsembleAggregator` produces meaningless values (SSR bias -> -1.0) and emits a torch `var(): degrees of freedom is <= 0` UserWarning rather than failing loudly. Pre-existing `_EnsembleAggregator` behavior; documented via "requires the batch to carry an ensemble dimension".
- 2 Behavior-change on existing paths: ✅ No existing caller is affected. The only modified path (`record_batch`/`get_summary`) gates all new work behind `self._ensemble_aggregator is not None`; the paired aggregators still receive the same folded data as before.
- 3 Differential tests: ✅ Verified by probe — stubbing out the `_EnsembleAggregator` construction makes `test_aggregator_ensemble_metrics` fail (keys absent) while the disabled-by-default test still passes. Fixtures use a real ensemble (n_ensemble=2, target=1), so the distinguishing behavior is visible. (Probe reverted; worktree left clean.)
- 8 AI-artifact scan: ✅ after fix. The description said the aggregator records "(unfolded)" data; it actually passes the native `EnsembleTensorDict`s as-is (the ensemble dim is intact, not folded like the paired aggregators get). Reworded the description for accuracy. No invented links, no in-body imports, imports at module top.
- 9 Split: ✅ Single coherent change, 2 files, one reason — no split.

**What I changed**
- PR description only: clarified the "(unfolded)" wording. No code changes were needed.

**Verification**
- Tests: `fme/ace/aggregator/test_train.py` 9 passed (incl. `test_aggregator_ensemble_metrics`, `test_aggregator_ensemble_metrics_disabled_by_default`), `FME_FORCE_CPU=1`.
- Lint: `ruff check` clean, `ruff format --check` clean.
- mypy: the two errors reported in `train.py` (lines 74, 155) both pre-exist on `origin/main` (the loose `dict[str, float]`/Protocol return annotations). No new mypy errors introduced.
- Port fidelity: matches the `experiment/2026-06-05-aimip-like` source; the direct `_EnsembleAggregator` import is the original author's deliberate design (the public `get_one_step_ensemble_aggregator` factory wraps it in a step-selecting adapter the train path does not want).

**Open questions (non-blocking)**
1. n_ensemble=1 with the flag on yields nonsense SSR bias + a torch warning rather than a loud error — acceptable for an opt-in diagnostic, but worth a one-line guard if you want loud failure here.
2. `train.py` reaches past the public factory into the private `_EnsembleAggregator`. Faithful to the source and justified (no step-selection wanted), so I left it; flagging only for visibility.

**Stepping back**
The PR does exactly what it claims: adds a default-off training-aggregator flag that surfaces CRPS / SSR-bias / ensemble-mean RMSE for train vs. val comparison. Small, well-scoped, faithfully ported, well-tested, easy to review.

Verdict: ready for review.

fme/ace/aggregator/train.py
```diff
@@ -121,6 +132,12 @@ def record_batch(self, batch: TrainOutput):
         ):
             self._per_channel_losses.record(batch.per_channel_losses)
 
+        if self._ensemble_aggregator is not None:
```
elynnwu: Add `n_ensemble > 1` also?

elynnwu (approved):

fme/ace/aggregator/train.py
mcgibbon: Yeah we should check `batch.n_ensemble.`

PR #1282, opened 2026-06-16
Fix broadcast_ensemble time-coordinate ordering for distinct-time samples

mcgibbon:
`BatchData.broadcast_ensemble` (and `PairedData.broadcast_ensemble`) expand the data, labels, and data_mask with **block** ordering (`repeat_interleave`: sample `s` lands at positions `[s*n_ensemble, (s+1)*n_ensemble)`), but built the `time` coordinate with `xr.concat([time] * n_ensemble)`, which **tiles** the samples (`[s0, s1, ..., s0, s1, ...]`). When samples carry distinct times — e.g. an inference task whose initial conditions start on different dates with `n_ensemble_per_ic > 1` — data and time then disagree on sample order. This surfaced as a hard crash in concurrent inline inference: `ValueError: Forcing data must have the same time coordinate as the batch data.` raised in `compute_derived_variables` (the `weather_2024` / `weather_1994` tasks: 8 ICs on 8 start dates, `n_ensemble_per_ic=8`).

The bug is pre-existing on `main`; concurrent inline inference is the path that first exercised broadcasting a multi-IC, distinct-time forcing window against an already-ensembled initial condition.

Changes:
- `fme.ace.data_loading.batch_data.BatchData.broadcast_ensemble` and `PairedData.broadcast_ensemble`: repeat-interleave the `time` coordinate per sample (`time.isel(sample=np.repeat(np.arange(n_samples), n_ensemble))`) so it matches the block ordering of the data/labels/data_mask.
- `fme.ace.data_loading.test_batch_data`: add failing-first regression tests (`test_broadcast_ensemble_aligns_distinct_sample_times`, `test_paired_data_broadcast_ensemble_aligns_distinct_sample_times`) that encode each sample's identity in both its data and time values and assert alignment after broadcasting; correct the existing `test_broadcast_ensemble`, which asserted the old tiled time order and only passed because its labels/times were uniform across samples.

- [x] Tests added

Resolves the second crash of the Wave-18 `rs1` concurrent-inference relaunch ([beaker, full URL](https://beaker.org/ex/01KV6P5MG100PTXNV436HD40AY)), after the window-shape fix #1279.

**Verified** on the `-05b8` relaunch — this fix cherry-picked onto the `experiment/2026-06-05-aimip-like` branch (commit `05b8a88c`):
https://beaker.org/ex/01KV76JQR4AEMXQTFNWXRTVS1Q (wandb run https://wandb.ai/ai2cm/ace/runs/icr4tr3s). The run completed the full first inline-inference callback — including the weather group's window that crashed the prior `-59ed` run — with no time-coordinate error, saved its epoch-1 checkpoint (`best_inference_error` 0.100), and proceeded into epoch 2.


mcgibbon:
Claude: Independent pre-review (read cold against `main`, tested in isolated worktree+venv, CPU). I did not write this PR.

**Audit passes**
- ✅ Silent-failure: No new defaults/fallbacks/optionals. The fix is a pure index-construction change; the existing `n_ensemble != 1` guard still raises loudly. `np.repeat(np.arange(n_samples), n_ensemble)` always produces a fully-specified index — no quiet default-taking.
- ✅ Behavior-change: `broadcast_ensemble` output ordering for `data`/`labels`/`data_mask` is unchanged. Only the `time` coordinate ordering changes, from tiled (`[s0,s1,...,s0,s1,...]`) to block-interleaved (`[s0,s0,s1,s1,...]`) — which is the bug being fixed and now matches `repeat_interleave_batch_dim` + `torch.repeat_interleave(dim=0)`. Verified `n_ensemble=1` stays a no-op: `np.repeat(arange(n),1)` is the identity isel, so the fast path is preserved. No downstream consumer relies on the old tiled time order (checked all callers in `inference.py`, `evaluator.py`, `single_module.py`, `coupled/stepper.py`, `train_config.py`).
- ✅ Differential-test: Reverted the source fix in my worktree → `test_broadcast_ensemble[2,3]` (the corrected existing test), `test_broadcast_ensemble_aligns_distinct_sample_times[2,3]`, and `test_paired_data_..._aligns_distinct_sample_times[2,3]` all go RED; the `n_ensemble=1` cases stay green (correct, since the fix is a no-op there). Restored → all 9 green. Fixtures are non-degenerate: 3 samples with distinct per-sample times/data (`1000*s`), so tiled≠interleaved is actually observable. Good catch that the old `test_broadcast_ensemble` only passed due to uniform-across-sample times.
- ✅ AI-agent artifacts: Comments explain the ordering rationale (legitimate, not diff-narration). Error string `"Forcing data must have the same time coordinate as the batch data."` cited in the description matches the real `ValueError` raised in `BatchData.compute_derived_variables` (batch_data.py:452). No invented symbols, no `Any`, no in-body imports, identifiers are clean. `ruff check` passes on both files.
- ✅ Scope/split: Single focused commit, 2 files, +85/-9 — matches the PR description's reported counts exactly. Coherent and easy to review.
- ✅ Records passes: N/A — this PR touches no `.md` records.

**Both methods handled consistently:** ✅ `BatchData.broadcast_ensemble` and `PairedData.broadcast_ensemble` get the identical fix, each with its own regression test.

**Other suites:** `test_single_module.py` + `aggregator/inference/test_main.py` (the broadcast_ensemble callers with test coverage) → 124 passed, 1 skipped.

**Changed/pushed:** Nothing — no issues found.

**Step back:** The PR does exactly what it claims. The root cause (tile-vs-block disagreement between `time` and `data`/`labels`/`data_mask`) is correctly diagnosed, the chosen approach (repeat-interleave the time index to mirror the data) is the right fix rather than a plausible-but-incidental one, both affected methods are handled symmetrically, and the n_ensemble=1 no-op + existing labels/data_mask ordering are preserved. Tests are genuinely failing-first. Small and self-contained.

fme/ace/data_loading/test_batch_data.py
```diff
@@ -568,6 +569,69 @@ def test_broadcast_ensemble(n_ensemble):
             )
 
 
+@pytest.mark.parametrize("n_ensemble", [2, 3])
+def test_broadcast_ensemble_aligns_distinct_sample_times(n_ensemble):
+    """Regression for the concurrent-inference rs1 crash (-59ed).
```
mcgibbon: If we're going to reference a specific run, please use a complete beaker url, as this will be hard to track when we don't know what it's talking about.

mcgibbon:
Claude: For the record, how the block-data/tiled-time mismatch this PR fixes came to be (the first two PRs predate the ace split and live in `ai2cm/full-model`):

- [**full-model#2204**](https://github.com/ai2cm/full-model/pull/2204) (Oct 2025, "Add ensemble properties to `BatchData`") introduced `broadcast_ensemble`. From the start it built the two halves with different orderings: **data** via `add_ensemble_dim` + `fold_sized_ensemble_dim` (a `(sample, ensemble)` reshape → block order `[s0,s0,s1,s1]`), but **time** via `xr.concat([time] * n_ensemble)` (tiled `[s0,s1,s0,s1]`). The mismatch was baked in here, just latent.
- [**full-model#2338**](https://github.com/ai2cm/full-model/pull/2338) (Dec 2025, "Condition stochastic SFNO on labels") refactored the method: **data** → `repeat_interleave_batch_dim` (same block order, no behavior change) and **labels** → `torch.repeat_interleave` (brought labels into block order). The **time** line was left tiled — so labels were aligned to the data while time stayed the straggler.
- This PR ([**#1282**](https://github.com/ai2cm/ace/pull/1282)) repeat-interleaves the time coordinate to match.

It stayed latent because it only bites when samples carry distinct per-sample times *and* the time coordinate is checked downstream (`compute_derived_variables`). Concurrent inline inference was the first path to broadcast a multi-IC, distinct-start-date forcing window (the `weather_2024`/`weather_1994` tasks, `n_ensemble_per_ic=8`) against an already-ensembled IC — which is where it surfaced.

mcgibbon:
Note the bug also can only surface when there are enough ICs that each GPU is getting more than one of them.

fme/ace/data_loading/test_batch_data.py
brianhenn: (nit): Not sure what the first line means here

brianhenn (approved):
LGTM thanks

fme/ace/data_loading/test_batch_data.py
mcgibbon: Done in 27e94dd5d — the docstring now names the crash plainly and links the full beaker run URL (https://beaker.org/ex/01KV6P5MG100PTXNV436HD40AY) instead of the `-59ed` shorthand.

mcgibbon: Reworded in 27e94dd5d — the opening line now reads "Regression for a concurrent inline-inference crash" with a short description of when it first appeared and the beaker link, dropping the `rs1 (-59ed)` shorthand.

mcgibbon: I figured this is OK/helpful because it's in test code, I would avoid referencing specific beaker runs in production code.

PR #1283, opened 2026-06-16
Replace step return tuple with `StepOutput` dataclass carrying pre-correction values

### LINKED ISSUE #1271: Replace step return tuple with a StepOutput dataclass carrying pre-correction values
## What to build

Replace the `tuple[TensorDict, StepperState | None]` return of the step abstraction with a `StepOutput` dataclass, and plumb pre-correction values up through the stepper so prediction `BatchData` optionally carries them. Pure plumbing: no user-visible behavior change.

This is PR 1 of a 3-PR split of #1218 and #1222, both of which predate the `StepperState` refactor on main. Branch from current main HEAD. Follow-up issues cover correction inference metrics and corrector training features.

Design decisions (settled in design review):

- Named `StepOutput`, following the repo's `*Output` convention (`TrainOutput`, `LossOutput`); `*Result` is reserved for benchmarks.
- Fields: `output` (denormalized post-adjustment data), `stepper_state` (unchanged passthrough semantics), and `uncorrected`.
- `uncorrected` carries pre-correction *values* (not deltas) of exactly the corrector-modified variables; empty dict when no corrector ran, so consumers need no None checks. Deltas are always derivable as `output − uncorrected`; values are not recoverable from deltas, and the training-features follow-up needs the values.
- `StepperState` is *not* used to carry uncorrected data: its contract is recurrent per-sample state threaded across predict windows, while uncorrected values are a per-window, per-timestep diagnostic series.
- Scope: the step package and corrector tap, the stepper's step/predict generator plumbing, and an optional `uncorrected` field on `BatchData` (following the `data_mask`/`stepper_state` optional-field precedent), populated only on prediction `BatchData` returned by `predict`. No `return_uncorrected` argument or overloads. `PairedData` exposes the uncorrected prediction for the metrics follow-up. If the `BatchData` layer bloats this PR, it may slip to the metrics PR.
- Uncorrected tensors are detached unconditionally here; attach/detach control arrives in the training-features follow-up with its first consumer.
- The output-process function (e.g. spatial masking) applies per-variable and name-preserving, so it is applied to the uncorrected subset identically.
- The rollout always feeds corrected output forward as state.
- `fme.coupled` touches only what is needed to unwrap `StepOutput.output`; all coupled feature support is deferred to a later PR.

## Acceptance criteria

- [ ] Every step implementation returns `StepOutput`, with `uncorrected` populated exactly when a corrector modifies variables and an empty dict otherwise; `stepper_state` semantics unchanged. (Prior art: the step package's existing unit tests; updated versions exist in both source PRs.)
- [ ] `predict` returns prediction `BatchData` with `uncorrected` set; rollout state is the corrected output; uncorrected tensors are detached. (Prior art: existing stepper unit tests.)
- [ ] The existing distributed/spatial-parallel step test matrix passes (baselines generated single-rank, verified under torchrun, per repo convention).
- [ ] No user-visible behavior change; existing checkpoints load unchanged.

## Blocked by

None - can start immediately.

## Notes

Source PRs serve as reference implementations only; neither merges as-is (#1222 is explicitly marked don't-merge): #1222 for the corrector tap and step-return change, #1218 for its alternative `StepResult` framing. Where they disagree, the decisions above win.

jpdunc23:
ACE steppers can apply correctors that adjust the network's raw output each step, but today the corrected and uncorrected outputs are conflated the moment `step` returns, so nothing downstream can interrogate the correction itself. This PR replaces the step abstraction's `tuple[TensorDict, StepperState | None]` return with a `StepOutput` dataclass that additionally carries the pre-correction values of the corrector-modified variables. It is pure plumbing within the step/stepper layer: no user-visible behavior change, and existing checkpoints load unchanged.

`StepOutput.uncorrected` holds the pre-correction *values* (not deltas) of exactly the variables a corrector modified, captured at the corrector boundary by tensor-identity detection of out-of-place edits and detached unconditionally; it is an empty dict when no corrector ran, so consumers need no None checks. The correction is derivable as `output - uncorrected`. Ocean and prescribed-prognostic adjustments run after the corrector and are excluded; the rollout always feeds the corrected output forward as state; `fme.coupled` only unwraps `StepOutput.output`. The corrector ABC is intentionally left unchanged (its existing `corrector_state` tuple is untouched).

This PR is now scoped to the step/stepper layer only. Per review feedback, the consumer-facing plumbing that surfaces `uncorrected` on prediction `BatchData`/`PairedData` has been moved to the follow-on aggregator PR (#1284), where it can be reviewed together with the metrics that consume it.

Changes:
- `fme.core.step.step.StepOutput`: new dataclass (`output`, `stepper_state`, `uncorrected`) replacing the step return tuple; `StepABC.step` and all implementations (`SingleModuleStep`, `SecondaryModuleStep`, `SeparateRadiationStep`, `FCN3Step`) return it.
- `fme.core.corrector.utils.captured_before`: new helper returning the pre-correction values of variables modified out-of-place.
- `fme.core.step.single_module.step_with_adjustments`: taps the corrector boundary to build the detached `uncorrected` shadow and returns `StepOutput`.
- `fme.core.step.multi_call` / `_multi_call`: `MultiCallStep` composes the wrapped step's shadow; the `MultiCall` helper returns an empty shadow.
- `fme.ace.stepper.single_module.Stepper.step`: applies the name-preserving output process func to the shadow and returns `StepOutput`; `predict_generator` yields `StepOutput` and feeds the corrected output forward; `_stack_step_outputs`/`process_prediction_generator_list` consume `StepOutput` (stacking only `output` and threading the terminal `stepper_state`).
- `fme.coupled.stepper.CoupledStepper`: unwraps `StepOutput.output`.

- [x] Tests added

Resolves #1271


fme/ace/data_loading/batch_data.py
```diff
@@ -131,6 +138,7 @@ class BatchData:
     n_ensemble: int = 1
     data_mask: TensorMapping | None = None
     stepper_state: StepperState | None = None
+    uncorrected: TensorMapping | None = None
```
mcgibbon: Issue: This attribute is very public, but it pertains to the low-level details of a specific Step implementation. Can we avoid tightly coupling BatchData to that Step using an encapsulation/hiding strategy like we did for StepperState (which also contains information specific to certain Step implementations, but only makes that information available to those implementations)?

mcgibbon: Issue: BatchData.broadcast_ensemble silently drops uncorrected

If we add this as a container instead of this specific low-level instance, then future private data we add could be covered by the same code/tests that we add at this stage to cover a e.g. a StepperMetrics container being added.

fme/ace/data_loading/batch_data.py
```diff
@@ -292,6 +315,7 @@ def new_on_cpu(
         n_ensemble: int = 1,
         data_mask: TensorMapping | None = None,
         stepper_state: StepperState | None = None,
+        uncorrected: TensorMapping | None = None,
```
mcgibbon: Issue: uncorrected got added to new methods for BatchData but not for PairedData, would be nice to keep them consistent.

mcgibbon (commented):
It looks like many BatchData methods like comptue_derived_variables drop these uncorrected values. If it's intentional it should get tested, if not it should get tested and fixed.

I would suggest taking this PR and moving all changes to BatchData/PairedData into the follow-on aggregator PR. The StepOutput changes look pretty mergeable as-is, though I'll do a full agent review and check of that review once you've done the move.

jpdunc23:
> 🤖 _Posted by Claude Code on James's behalf._

Thanks @mcgibbon. Done — I've moved all `BatchData`/`PairedData` changes out of this PR. It's now scoped to the step/stepper layer only:

- `BatchData.uncorrected` / `PairedData.uncorrected` and their threading through the movement methods are gone from this branch (no diff to `fme/ace/data_loading/batch_data.py` vs `main`).
- `Stepper.predict` no longer stacks the per-step shadows onto the returned `BatchData`; `_stack_step_outputs` stacks only `output` and threads the terminal `stepper_state`. The `uncorrected` shadow still flows through `StepOutput` at the step/stepper boundary, but stops there.
- The consumer-facing plumbing (and therefore the `compute_derived_variables`-drops-`uncorrected` question) now lives in the follow-on aggregator PR #1284, where the metrics that consume it can be reviewed and tested together.

The description has been updated to reflect the narrowed scope. Ready for the full agent review whenever you are.

fme/core/corrector/utils.py
```diff
@@ -9,3 +9,20 @@ def force_positive(data: TensorMapping, names: list[str]) -> TensorDict:
     for name in names:
         out[name] = torch.clamp(data[name], min=0.0)
     return out
+
+
+def captured_before(original: TensorMapping, corrected: TensorMapping) -> TensorDict:
```
mcgibbon, line 14: Issue: I'm wary of relying on the memory mutation behavior of the step code (which isn't constrained by a test to have this property) to determine... 

Issue: it's not very clear what this function does based on its name.

fme/core/corrector/utils.py
```diff
@@ -9,3 +9,20 @@ def force_positive(data: TensorMapping, names: list[str]) -> TensorDict:
...
+    """Return the pre-correction values of variables a corrector modified.
```
mcgibbon, line 15: Issue: the relationship between "a corrector" and these inputs should be concretely defined, if we are going to define the output based on that relationship.

fme/core/step/single_module.py
```diff
@@ -603,13 +606,23 @@ def step_with_adjustments(
...
+        pre_correction = output
         output, corrector_state = corrector(
             input, output, next_step_input_data, corrector_state
         )
+        # Tap the corrector boundary: capture the pre-correction values of
+        # exactly the variables it modified (ocean and prescribed-prognostic
+        # adjustments run after and are intentionally excluded). Detached
+        # unconditionally here: unused on the train path, so this avoids
+        # retaining the autograd graph. The correction is output - uncorrected.
+        uncorrected = {
```
mcgibbon, line 623: Clarifying: This dict contains all output variables, correct?

fme/core/step/step.py
```diff
@@ -16,6 +16,30 @@
...
+            so consumers need no None checks. This is a sparse "shadow" of
+            ``output`` used to evaluate how much the stepper relies on the
+            corrector; the correction is derivable as ``output - uncorrected``.
+            Note ocean and prescribed-prognostic adjustments run after the
+            corrector and are not reflected here.
+    """
+
+    output: TensorDict
+    stepper_state: StepperState | None
+    uncorrected: TensorDict = dataclasses.field(default_factory=dict)
```
mcgibbon, line 40: Issue: The corrector doesn't strictly apply corrections (despite the name), it also does things like completely derive the tendency of water vapor due to advection from-scratch, ignoring the NN output. We likely don't want to optimize these unused weights. I'm actually not sure whether the current code would or wouldn't optimize them, because of the memory-update-based logic for selection.

If we supplied values of corrections as direct metrics, we'd have a direct pathway to control whether a corrector action is classified as a "correction" or not, by not including non-correction effects (like derived overwrites, or residual relationships) in the correction metric.

Claude was particularly concerned about this correction, which would allegedly conflict with the residual prediction of hfds and should probably not be included in pre-corrector loss or corrector regularization:
```
fme/core/corrector/ocean.py:260-261
if method == "residual_prediction":
    out[hfds_name] = net_flux * ocean_fraction + gen_hfds
```

mcgibbon:
I would suggest re-framing from uncorrected values to the value of the correction being applied, and only storing corrections for variables that get corrected. This change would resolve several issues in this and the next PR.

jpdunc23:
Superseded by #1328

PR #1285, opened 2026-06-16
Fix standalone-inference ensemble double-broadcast by mirroring the evaluator (alt to #1278)

mcgibbon:
Standalone inference (`run_inference_from_config`) crashes with `DataShapesNotUniform` when `n_ensemble_per_ic > 1`, because the forcing data is broadcast across ensemble members twice. This PR fixes it without teaching any data-loading API about `n_ensemble`: it defers the ensemble broadcast until after the forcing loader is built, exactly as the inline-training evaluator already does, so the forcing is broadcast exactly once. This is an alternative to #1278, which instead threads `n_ensemble` through four data-loading APIs. See the comment below for the detailed root cause and trade-off vs. #1278.

Changes:
- `fme.ace.inference.inference.run_inference_from_config` — stop tiling the initial condition before building forcing; call `get_initial_condition` with the default `n_ensemble=1` so `get_forcing_data` builds forcing with `n_ensemble=1`, then broadcast `data._initial_condition` to `n_ensemble_per_ic` after `get_forcing_data` returns. `get_initial_condition`'s `n_ensemble` parameter is now used only by the coupled caller.
- `test_inference_entrypoint` — parametrized over `n_ensemble_per_ic` in `[1, 2]`; the `2` case reproduces the crash without the fix.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/inference.py
+++ b/fme/ace/inference/inference.py
@@ -309,7 +309,6 @@ def run_inference_from_config(config: InferenceConfig):
             config.initial_condition.get_dataset(),
             stepper_config.prognostic_names,
             labels=config.labels,
-            n_ensemble=config.n_ensemble_per_ic,
         )
         stepper = config.load_stepper()
         stepper.set_eval()
@@ -323,6 +322,18 @@ def run_inference_from_config(config: InferenceConfig):
             ocean_fraction_name=stepper.ocean_fraction_name,
             label_override=config.labels,
         )
+        # Broadcast the initial condition across ensemble members only after the
+        # forcing loader is built, mirroring the evaluator path. The forcing then
+        # has one window per initial condition (n_ensemble=1) and predict_paired
+        # broadcasts it exactly once to match the ensemble-broadcast initial
+        # condition. Broadcasting before get_forcing_data would tile the forcing
+        # start times too, and predict_paired would broadcast the already-tiled
+        # forcing a second time (the standalone double-broadcast bug).
+        if config.n_ensemble_per_ic > 1:
+            ic = data.initial_condition.as_batch_data()
+            data._initial_condition = PrognosticState(
+                ic.broadcast_ensemble(config.n_ensemble_per_ic)
+            )
 
         if not config.allow_incompatible_dataset:
             try:
--- a/fme/ace/inference/test_inference.py
+++ b/fme/ace/inference/test_inference.py
@@ -110,7 +110,8 @@ def save_stepper(
     torch.save({"stepper": stepper.get_state()}, path)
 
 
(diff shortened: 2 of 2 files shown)


mcgibbon:
Moving the detailed analysis here so the PR description stays in the repo template format.

## The bug

`get_initial_condition(n_ensemble=N)` tiles the initial condition to `N` members *before* the forcing loader is built. `get_forcing_data` derives its forcing start times from the initial condition's times, so the already-tiled IC produces one forcing window per tiled sample — but the forcing batches still report `n_ensemble=1`. Because of that stale flag, `predict_paired` broadcasts the forcing a *second* time, inflating it from `N` to `N**2` members while the initial condition stays at `N`. The mismatched sample counts then fail in the packer.

The inline-training evaluator path is unaffected: it broadcasts the initial condition *after* building the forcing loader, so its forcing genuinely has `n_ensemble=1` and is broadcast exactly once.

## Fix (vs. #1278)

#1278 threads the initial condition's `n_ensemble` through `get_forcing_data` -> `get_inference_data` -> `InferenceDataset` -> `BatchData.from_sample_tuples` so the forcing reports the correct `n_ensemble`.

Per @Arcomano1234's review note on #1278 — that this can't be done without the getters knowing `n_ensemble` *unless* `inference.py` is changed to mirror the evaluator — this PR takes that path instead. It defers the ensemble broadcast until after the forcing loader is built, exactly as the evaluator does:

- `get_initial_condition` is called with the default `n_ensemble=1`, so `get_forcing_data` sees an untiled IC and builds forcing with `n_ensemble=1` (one window per IC).
- After `get_forcing_data` returns, the IC on the returned `InferenceGriddedData` is broadcast to `n_ensemble_per_ic`, so `predict_paired` broadcasts the forcing exactly once to match.

No data-loading API needs to know `n_ensemble`, and the two inference paths converge rather than diverge. The writer's `initial_condition_times` follow automatically because `InferenceGriddedData.initial_time` is derived from the initial condition (no separate time re-tiling needed, unlike the evaluator).

## Trade-off vs. #1278

The cost is moving the ensemble broadcast out of `get_initial_condition` and into `run_inference_from_config` (the caller now owns the broadcast). That is already the established pattern in the evaluator, so it makes the two paths converge — the opposite of #1278, which adds an `n_ensemble` parameter to four data-loading APIs that exists solely to paper over the ordering difference.

## How it was found

Surfaced while setting up the blowup-investigation probes, which worked around it by building forcing from a single-sample initial condition. No production run is on record hitting it — standalone inference had not been exercised with `n_ensemble_per_ic > 1`.

Arcomano1234 (approved):
Approved, I am in favor of this PR instead of #1278

PR #1292, opened 2026-06-17
Add concurrent inline inference + equivalence test (3/3 of #1195 split)

mcgibbon:
**PR 3 of 3** splitting #1195. Targets PR 2
(`feature/cinf-2-refactor-inference-callback`); retarget to `main` after PR 1
and PR 2 merge. **Supersedes #1195.**

Adds the concurrent inline-inference feature on top of the cat/split helpers
(PR 1) and the per-task callback dispatch (PR 2), plus the
concurrent-vs-sequential equivalence test that #1195 was missing.

Concurrent inference means that each inference configuration queues its work, and then stepper calls themselves happen as a single call across the grouped inference configurations, on a single concatenated BatchData. The result then gets split for each inference task to collect and pass to its aggregator.

Changes:
- `fme.core.generics.inference` — `BatchedPredictor`, `PredictionPromise`, `LazyLooper`, `ConcurrentInferenceEntry`, `run_concurrent_inference`. `batch_key_of_forcing` keeps windows of different shapes (e.g. a short final window beside a full window in the same round) in separate forward passes (ace#1279).
- `fme.core.generics.trainer` — `InferenceTask.concurrent_group`, the grouping branch in `build_inference_callback` (tasks sharing a non-None group run together; singleton/`None` groups fall back to the sequential helper), and `_run_inference_tasks_concurrent`.
- `fme.ace.train.train` / `fme.coupled.train.train` — build the `BatchedPredictor` and set `concurrent_group`.
- `fme.ace.stepper.single_module` — compare IC vs forcing labels with `semantically_equal` so concurrently-cat'd labels (possibly differing in encoding) still pass.
- `fme.ace.train.test_concurrent_inference_equivalence` — runs one real stepper through both the sequential and concurrent paths and asserts equal predictions, derived/diagnostic fields, time coordinates, and aggregator metric logs; covers the mismatched-window-shape (ace#1279) and ensemble broadcast (ace#1282) paths that diverged only in production. (Cross-window `stepper_state` propagation is covered by the cat/split round-trip tests in PR 1.)

Reconciliation note: this chain rebases #1195 onto current `main`, which since #1195 was opened replaced the string-keyed inference-error lookup with `InferenceSummary.loss` and added LR-tuning epoch threading. The original PR's `error_metric_template` and unused `aggregator_factory(epoch)` changes are dropped as obsolete; checkpoint-selection error now comes from `summary.loss`.

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt — n/a

### DIFF
--- a/fme/core/generics/trainer.py
+++ b/fme/core/generics/trainer.py
@@ -58,7 +58,7 @@ import signal
 import sys
 import time
 import uuid
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Hashable, Mapping, Sequence
 from typing import Any, ClassVar, Generic, Protocol, TypeVar
 
 import torch
@@ -73,7 +73,12 @@ from fme.core.generics.aggregator import (
     InferenceSummary,
 )
 from fme.core.generics.data import GriddedDataABC, InferenceDataABC
-from fme.core.generics.inference import run_inference
+from fme.core.generics.inference import (
+    BatchedPredictor,
+    ConcurrentInferenceEntry,
+    run_concurrent_inference,
+    run_inference,
+)
 from fme.core.generics.lr_tuning import (
     LRTuningConfig,
     ValidateStepper,
@@ -879,6 +884,10 @@ class InferenceTask(Generic[PS, FD, SD]):
         epoch_set: Epochs on which this task should run.
         weight: Contribution weight for the combined checkpoint-selection error.
             Zero means the task runs but does not contribute to the metric.
+        concurrent_group: Tasks sharing this key are eligible to run through a
+            single batched forward pass per step. ``None`` disables concurrent
+            batching for this task. Ignored when no ``build_batched_predictor``
+            is supplied to ``build_inference_callback``.
     """
 
     name: str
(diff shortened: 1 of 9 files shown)


mcgibbon:
In the 4degree-daily case this only gives a 3% consistent speed up on all runs, however in sequential mode the inference _frequently_ suffers a large spike in inference times, which the concurrent mode did not suffer during my benchmark tests nor during the full training runs. This in practice gives more like a 17% speed-up for inference in the dual-46year configuration. As inference is a large fraction of those runs, even when executing every 5 epochs, total training time improves by about 10%.

Combining that with the expectation that this could be more impactful for configurations with more than 2 similar-length inferences, and that the code maintains the existing sequential pathway, making it easy to delete the parallel version in the future if it causes issues, and meaning changes are regression tested against the simpler sequential implementation, I do think we should merge this feature.

fme/core/generics/trainer.py
```diff
@@ -958,6 +995,38 @@ def _run_inference_task_sequential(
     )
 
 
+def _run_inference_tasks_concurrent(
```
mcgibbon, line 998: This is the main entrypoint for the new concurrent inference, it mirrors `_run_inference_tasks_sequential` in purpose.

fme/core/generics/trainer.py
```diff
@@ -958,6 +995,38 @@ def _run_inference_task_sequential(
...
+        ConcurrentInferenceEntry(
+            name=t.name,
+            data=t.data,
+            aggregator=aggregators[t.name],
+        )
+        for t in tasks
+    ]
+    stepper.set_eval()
+    with torch.no_grad(), GlobalTimer():
+        run_concurrent_inference(predictor, entries)
```
mcgibbon, line 1015: A key feature is that all inference has to happen before any aggregators can be called.

PR #1295, opened 2026-06-17
Add concurrent inline inference + sequential-equivalence test (cinf 3/3)

mcgibbon:
Splits the stuck concurrent-inline-inference PR #1195 into a reviewable 3-PR chain. This is **PR 3 of 3** (base: the PR 2 branch `cinf/pr2-callback-refactor`). Adds the concurrent inline-inference feature on top of the sequential callback refactor, **plus** the concurrent-vs-sequential equivalence test that #1195 was missing. Together PRs 1+2+3 reproduce #1195's content (reconciled with `main`) plus that test.

Changes:
- `fme.core.generics.inference` — `BatchedPredictor`, `PredictionPromise`, `ConcurrentInferenceEntry`, and `run_concurrent_inference`, including the window-shape partitioning (`batch_key_of_forcing`) that is the #1279 mismatched-window concurrent-cat fix.
- `fme.core.generics.trainer` — `concurrent_group` on `InferenceTask`, `_run_inference_tasks_concurrent`, and the grouping branch in `build_inference_callback` (group by `concurrent_group`; singleton / `None` groups fall back to the sequential helper), plus the `build_batched_predictor` parameter. Both paths return per-task `InferenceSummary`, so concurrent and sequential are interchangeable.
- `fme.ace.train.train` / `fme.coupled.train.train` — wire `build_predictor` and set `concurrent_group` (`(forward_steps_in_memory, n_ensemble_per_ic)` for ACE, `(coupled_steps_in_memory,)` for coupled).
- `fme.ace.stepper.single_module` — compare IC-vs-forcing labels with semantic equality so concurrently-cat'd labels (which may differ in encoding) still pass.

- [x] Tests added:
  - `fme/core/generics/test_batched_predictor.py` and the concurrent-grouping tests in `fme/core/generics/test_inference_callback.py`.
  - **New equivalence test** `fme/core/generics/test_inference_concurrent_equivalence.py`: builds one small real stepper and runs the same tasks through both `run_inference` (sequential) and `run_concurrent_inference` (concurrent), asserting equal predicted fields, derived variables, and aggregator metric logs (`torch.testing.assert_close`, tight rtol/atol). Covers the two production-divergent paths — a heterogeneous-length pair (a short task reaching its final window while a longer task submits a full window in the same round → #1279 partitioning, verified by a forward-pass count showing the tasks really shared a pass) and an ensemble task (`n_ensemble_per_ic > 1`) on distinct start dates (→ #1282 broadcast_ensemble time-coordinate ordering). These divergences were previously caught only empirically on beaker, because production configs are effectively fully concurrent and never run the sequential reference path.
- [ ] If dependencies changed, "deps only" image rebuilt — n/a.

---
**Stacked chain** (replaces #1195; review bottom-up):
1. `cat`/`split` helpers → `main`
2. callback refactor → PR 1 branch
3. **This PR** — concurrent inference + equivalence test → PR 2 branch

Supersedes #1195.

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -315,6 +315,14 @@ class SingleModuleStepperConfig:
         )
 
 
+def _labels_semantically_equal(a: BatchLabels | None, b: BatchLabels | None) -> bool:
+    if a is None and b is None:
+        return True
+    if a is None or b is None:
+        return False
+    return a.semantically_equal(b)
+
+
 def _prepend_timesteps(
     data: EnsembleTensorDict, timesteps: TensorMapping, time_dim: int = 2
 ) -> EnsembleTensorDict:
@@ -1083,7 +1091,7 @@ class Stepper:
             Generator yielding (output, stepper_state) tuples at each timestep.
         """
         ic_batch_data = initial_condition.as_batch_data()
-        if ic_batch_data.labels != forcing_data.labels:
+        if not _labels_semantically_equal(ic_batch_data.labels, forcing_data.labels):
             raise ValueError(
                 "Initial condition and forcing data must have the same labels, "
                 f"got {ic_batch_data.labels} and {forcing_data.labels}."
@@ -1660,7 +1668,7 @@ class TrainStepper(
         input_data = data.get_start(self._prognostic_names, self.n_ic_timesteps)
         n_ensemble = self._config.n_ensemble
         input_batch_data = input_data.as_batch_data()
-        if input_batch_data.labels != data.labels:
+        if not _labels_semantically_equal(input_batch_data.labels, data.labels):
             raise ValueError(
                 "Initial condition and forcing data must have the same labels, "
                 f"got {input_batch_data.labels} and {data.labels}."
--- a/fme/ace/train/test_train_config.py
+++ b/fme/ace/train/test_train_config.py
@@ -361,11 +361,13 @@ class TestGetValidationCallback:
 
 class TestGetInferenceCallback:
     @staticmethod
-    def _make_entry(name, weight=1.0):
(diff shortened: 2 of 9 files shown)


mcgibbon:
Claude: independent cold pre-review run (fresh agent, no authoring context).

- Silent-failure ✅ — exhausted-loader drop, `batch_key_of_forcing` shape partitioning, singleton-group and `build_batched_predictor=None` fallbacks all behave loudly/intentionally.
- Behavior-change ✅ — the `_labels_semantically_equal` swap conforms both sides to the union encoding (the correct relaxation for concurrently-cat'd labels, not a weakening).
- Correctness ✅ — outputs routed back by stable slot id; split sizes taken from per-IC sample counts.
- Differential-test — the heterogeneous-length equivalence test is rigorous: the con=4 / seq=5 forward-pass counts genuinely prove the tasks shared a pass and then partitioned by shape; reversing the split makes it fail.

Fixed in response: the distinct-start-date scenario previously used a **single** concurrent entry, so the predictor took its single-slot fast path and `cat`/`split` never ran (and `n_ensemble` stayed 1) — it could not catch a #1282-style time-coordinate misordering. Reworked it to batch **two** tasks (4 samples each, distinct per-sample start dates) through one group and assert the forward-pass count drops (4→2), so it now genuinely exercises the concurrent `cat`/`split`. Also removed an unused name + unused recorded IC and corrected the overclaiming docstrings.

Verdict: ready for review.

mcgibbon:
Closing as a duplicate. A parallel session had already opened the equivalent 3-PR split of #1195 first — #1290 / #1291 / #1292 (non-draft, and more complete: it threads stepper_state through cat/split). Deferring to that chain per the author's direction.

PR #1296, opened 2026-06-17
Report 0 for prescribed (forced) cells in ssr_bias instead of -1

mcgibbon:
The ensemble spread-skill ratio bias (`ssr_bias`) was `-1` over prescribed regions such as SST over ocean. There every ensemble member is identical (zero spread) and equals the target (zero ensemble-mean error), so the ratio is a genuine `0/0` that `spread / skill - 1` collapses to the `-1` floor. Because such variables are prescribed over most of the surface, this dominated the per-variable mean map and the scalar global average, making the metric uninformative about calibration.

The fix reports `0` (perfectly calibrated -- zero spread is right for zero error) at these prescribed `0/0` cells. They are identified from the accumulated totals at reduction time: `total_variance == 0` (exact) and `total_unbiased_mse` below a small fraction (`1e-6`) of the field's largest unbiased MSE. The relative threshold is needed because averaging identical members leaves a tiny rounding residue in the unbiased MSE, so it is not exactly zero. Genuine zero-spread-with-error underdispersion still reports `-1`.

Changes:
- `fme.ace.aggregator.one_step.ensemble.SSRBiasMetric.get`: report `0` for prescribed (`0/0`) cells via the totals-based threshold above, keeping `-1` for genuine zero-spread/nonzero-skill cells.

- [x] Tests added


mcgibbon (commented):
A few fundamental points of confusion or issues with this approach to solving the issue:
- It seems this will give invalid results if a field is sometimes prescribed and sometimes not
- Isn't the SSR bias when the predictions equal the target just zero? I get that it's nan/undefined to take 0/0, but can't we by convention say that the SSR bias is 0 for cells that have total variance and total unbiased MSE which are both zero? Doesn't that solve this without the above issue?

mcgibbon:
Thanks — both points are fair. I've restructured (latest commit `d16a4e9f`):

**On the convention (your 2nd point): agreed, switched to `ssr_bias = 0`.** Reporting 0 for the prescribed `0/0` cell (zero spread is exactly right for zero error → perfectly calibrated) is cleaner than NaN and lets me delete the NaN handling and the `_area_weighted_mean_skipna` exclusion entirely. Prescribed cells now contribute the benign `0` to the scalar mean instead of the `-1` floor.

**One catch on detecting it from the totals.** `(total_variance == 0) & (total_unbiased_mse == 0)` doesn't actually fire at prescribed cells. `total_variance` is exactly 0, but the unbiased MSE is *not* — averaging the bit-identical members leaves a rounding residue, so `total_unbiased_mse ≈ 1.4e-15`, not 0 (verified locally). With that residue the cell takes the `skill > 0` branch and lands back at `spread/skill - 1 = -1`. That residue is exactly why detection is on the raw values (`gen == target`, before the lossy mean) rather than the post-mean totals.

**On "sometimes prescribed" (your 1st point):** a cell is flagged only if every recorded batch had `gen == target` there. A sometimes-prescribed cell is *not* flagged and falls through to the normal computation, where its prescribed batches contribute 0 to both the pooled variance and the pooled unbiased-MSE, so the result is the SSR over the batches where it was actually forecast — not an invalid number. Your totals-based idea would behave identically for such a cell (its `total_variance` is > 0, so the 0/0 convention wouldn't apply either). So the only real difference between the two routes was how the always-prescribed `0/0` cell is handled, and the residue above is what blocks the pure-totals version of it. Happy to switch to a tolerance-based totals check instead if you'd prefer that to the raw-value detection.

mcgibbon:
Yes — done in `e2f2561` (dropped the per-cell `gen == target` accumulation entirely; detection is now purely from the totals).

A cell is prescribed when:

```python
mse_floor = 1e-6 * skill.square().max()        # 1e-6 of the field's largest unbiased MSE
prescribed = (total_variance == 0) & (total_unbiased_mse <= mse_floor)
```

Why a *relative* threshold is safe: the residue is `~ε²·value²` (ε = float32 epsilon ≈ 1.2e-7) — that's the ~1e-15 we measured — whereas any genuine forecast cell's unbiased MSE is within a few orders of magnitude of the field's worst cell. So the residue sits 6–10 orders below the floor while real skill sits far above it, with no unit dependence (an absolute tolerance would have to know the variable's units). The `total_variance == 0` gate is exact (variance of bit-identical members is exactly 0), so only zero-spread cells are ever in play, and among those it's residue-vs-real, a huge gap.

One caveat worth noting: this normalizes against the field's max unbiased MSE, so a variable that is prescribed over the *entire* grid (no forecast cells to set the scale) would fall back to `-1` rather than `0`. That doesn't affect the motivating case (surface temperature is prescribed only over ocean, so land sets the scale), but it's the one thing the previous raw-value detection handled that this doesn't. Happy to keep it as-is or add an absolute floor for that degenerate case if you'd rather — let me know.

fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -151,9 +157,30 @@ def get(self) -> torch.Tensor:
...
-        # When skill is zero (clamped or genuinely perfect), SSR is undefined.
-        # Use -1 as the convention (equivalent to zero spread).
-        return torch.where(skill > 0, spread / skill - 1, torch.tensor(-1.0))
+        # When skill is zero but spread is not, SSR is undefined; -1 is the
+        # convention there (it is also the exact value of spread / skill - 1 in
+        # the genuine zero-spread, nonzero-skill underdispersive limit).
+        ssr_bias = torch.where(
+            skill > 0, spread / skill - 1, torch.full_like(spread, -1.0)
+        )
+        # Prescribed/forced cells (every ensemble member equals the target, e.g.
```
mcgibbon: This is way too much text, please compact or remove.

Arcomano1234 (approved):
Approved

PR #1297, opened 2026-06-17
Fail loud when a NaN reaches the network input

mcgibbon:
Masked-data handling currently fails silently: if an input/forcing variable contains a NaN that is neither covered by `data_mask` nor filled by `fill_nans_on_normalize`, the NaN propagates through the network and surfaces only as a NaN loss, with no indication of which variable caused it or that masking was the issue. This is exactly the failure mode fixed in #1262 (ensemble training reading `data_mask` from the wrong, prognostic-only source so masked forcing variables were never zeroed). The existing `Optimization._validate_loss` only catches the symptom — much later, training-only, and with an unhelpful "Loss is NaN-valued" message.

This PR adds a guard at the network-input boundary in `SingleModuleStep` that raises a `ValueError` naming the offending variable(s) the moment a NaN survives masking and normalization, so a mis-wired or missing mask is caught immediately and located rather than debugged backwards from a NaN loss.

The guard is gated on `torch.is_grad_enabled()` so the `no_grad` inference/eval rollout hot path pays no added device-sync cost; the cheap packed-tensor `.any()` check runs first, and the per-variable scan that builds the message only runs on the error path.

Changes:
- `fme.core.step.single_module.SingleModuleStep.step`: raise (via new `_raise_input_nan_error`) when a NaN survives `_apply_input_mask` to the packed network input, under `torch.is_grad_enabled()`.
- `fme.core.step.single_module._raise_input_nan_error`: helper that names the NaN-bearing input variable(s) in the error message.
- Tests: unmasked NaN forcing now raises in `train_on_batch`; `step()` raises on unmasked NaN input and stays silent on a properly masked NaN input.

Relationship to #1262: this guard is complementary. With #1262's fix, a properly masked forcing NaN is zeroed and the guard stays quiet; without a correct mask (the #1262 bug), training now raises a clear, located error instead of silently producing NaN loss. Note that #1262 adds `test_train_on_batch_unmasked_nan_forcing_produces_nan_loss_in_ensemble`, which asserts the old silent-NaN-loss behavior; once both land, that test should be updated to assert the raised error instead.

- [x] Tests added

### DIFF
--- a/fme/core/step/single_module.py
+++ b/fme/core/step/single_module.py
@@ -383,6 +383,14 @@ class SingleModuleStep(StepABC):
             if args.data_mask is not None:
                 input_norm = _apply_input_mask(input_norm, args.data_mask)
             input_tensor = self.in_packer.pack(input_norm, axis=self.CHANNEL_DIM)
+            # Fail loud if a NaN survives to the network input. This only happens
+            # when an input variable contains NaN that was neither covered by
+            # data_mask nor filled by fill_nans_on_normalize, which otherwise
+            # propagates silently to a NaN loss with no clue which variable (or
+            # that masking was the cause) -- see PR #1262. Gated on grad-enabled
+            # so the no_grad inference/eval rollout hot path pays no sync cost.
+            if torch.is_grad_enabled() and torch.isnan(input_tensor).any():
+                _raise_input_nan_error(input_norm)
             if self._config.include_channel_mask_inputs:
                 mask_dict = _build_channel_mask_dict(
                     self.in_packer.names, args.data_mask, input_tensor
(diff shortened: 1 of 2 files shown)


fme/core/step/single_module.py
```diff
@@ -383,6 +383,14 @@ def network_call(input_norm: TensorDict) -> TensorDict:
...
                 input_norm = _apply_input_mask(input_norm, args.data_mask)
             input_tensor = self.in_packer.pack(input_norm, axis=self.CHANNEL_DIM)
+            # Fail loud if a NaN survives to the network input. This only happens
+            # when an input variable contains NaN that was neither covered by
+            # data_mask nor filled by fill_nans_on_normalize, which otherwise
+            # propagates silently to a NaN loss with no clue which variable (or
+            # that masking was the cause) -- see PR #1262. Gated on grad-enabled
+            # so the no_grad inference/eval rollout hot path pays no sync cost.
+            if torch.is_grad_enabled() and torch.isnan(input_tensor).any():
+                _raise_input_nan_error(input_norm)
```
yyexela: Nitpick (non-blocking): I think 6 lines of comments for this 2-line readable code is excessive.

Suggestion: Condense to a 1-line comment or remove entirely.

yyexela (approved):
Code is good, one nitpick about long comment, also ensure that the test `test_train_on_batch_unmasked_nan_forcing_produces_nan_loss_in_ensemble` on main after merging #1262 is updated according to this PR.

First point isn't a code change, second point will throw an error in tests. So, if tests pass then this PR is good to go, don't need to re-review.

yyexela:
Note: This PR doesn't add an inference guard, but it's unlikely to have any issues during inference if training ran without errors. This mostly just raises errors sooner during training rather than later.

mcgibbon:
> Note: This PR doesn't add an inference guard, but it's unlikely to have any issues during inference if training ran without errors. This mostly just raises errors sooner during training rather than later.

Yes, if anything I would say propagating nans during inference is a feature not a bug (I'd rather get output showing it NaN'd at a point in time than have the application crash).

PR #1298, opened 2026-06-18
Integrating sea ice into coupled data loaders

William-gregory:
This is a first step towards integrating a stand-alone sea ice component into `fme.coupled`. This PR makes changes to:

- `fme.coupled.data_loading`
- `fme.coupled.dataset_info`
- `fme.coupled.requirements`

These changes both add a new `ice` component as well as make each `atmosphere`, `ocean` and `ice` component optional `None`. Therefore, if, for example, no `atmosphere` component is given, then the code assumes you are coupling an `ocean` model to an `ice` model. Similarly, if no `ice` component is given, then `atmosphere-ocean` coupling is assumed. 

Had to make some additional changes outside of `fme.coupled.data_loading` to resolve some errors that pre-commit was throwing up around assertions. These include changes to `fme.coupled.stepper`, `fme.coupled.aggregator`, `fme.coupled.inference.evaluator`, `fme.coupled.inference.test_evaluator`, `fme.coupled.inference.test_inference`, `fme.coupled.inference.loop`, `fme.coupled.inference.data_writer`, `fme.coupled.train`,  `fme.coupled.test_stepper`,

### DIFF
--- a/fme/coupled/aggregator.py
+++ b/fme/coupled/aggregator.py
@@ -124,6 +124,8 @@ class OneStepAggregator(AggregatorABC[CoupledTrainOutput]):
         self._loss_atmos = torch.tensor(0.0, device=get_device())
         self._n_batches = 0
         config = OneStepAggregatorConfig()
+        assert dataset_info.ocean is not None
+        assert dataset_info.atmosphere is not None
         self._aggregators = {
             "ocean": config.build(
                 dataset_info=dataset_info.ocean,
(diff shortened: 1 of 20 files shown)


jpdunc23 (commented):
Thanks for doing this. What you've done here makes sense, but I'm hoping we can reduce many of the `if` existence checks.

What I notice is that almost all of the changes involve 1) expanding the existing atmosphere and ocean logic to handle `None` typing and 2) the addition of a similar third block for sea ice.

For 2), this makes me wonder if we can first refactor the ocean / atmosphere handling to be use generic "coupled component data" handling via generic helper utils that don't need to know or care about what specific component we're dealing with. This might reduce or eliminate much of the code repetition and make it more straightforward to add additional components (e.g., biogeochemical) in the future.

For 1), I think we should avoid adding to `CoupledDatasetConfig` and instead add a new `CoupledIceOceanDatasetConfig` which has only `ocean` and `ice`, so `CoupledDataLoaderConfig.dataset` typing gets expanded to `dataset: CoupledConcatDatasetConfig | CoupledDatasetConfig | CoupledIceOceanDatasetConfig`.

PR #1299, opened 2026-06-19
  Support segmented inference on remote experiment directories

brianhenn:
Segmented inference (`--segments`) to a remote `experiment_dir` (e.g. `gs://`) failed at the
  segment boundary: each segment carries its initial condition forward via a `restart.nc`, but while
  the write is remote-safe, the read side wasn't — `InitialConditionConfig.get_dataset` opened netCDF
  only from local paths, and `run_segmented_inference`'s resume check used `os.path.exists` (always
  False for `gs://`). This routes remote netCDF reads through a local temp copy so segments chain and
  already-completed segments are skipped on resubmit, with zarr and local paths unchanged.

  Changes:
  - `fme.core.cloud.open_dataset_via_inter_filesystem_copy` and `fme.core.cloud.exists` — read/exists
    counterparts to the existing `to_netcdf_via_inter_filesystem_copy` (copy to a local temp, then
    `xr.open_dataset(...).load()`; fsspec-aware existence check)
  - `fme.ace.inference.InitialConditionConfig.get_dataset` routes remote, non-zarr opens through the
    copy helper; `fme.ace.inference.run_segmented_inference` uses `cloud.exists` for the resume-skip

  - [x] Tests added


spencerkclark (approved):
Looks good, thanks!  I hit this in another context, but it didn't end up being critical for my use-case.  It will be great to have it fixed.

PR #1300, opened 2026-06-20
Block n_ensemble_per_ic>1 in segmented runs

brianhenn:
Segmented inference reused the same `n_ensemble_per_ic` for every segment, but a segment's `restart.nc` already holds the broadcasted ensemble as its `sample` dimension. `broadcast_ensemble` guards only on the `n_ensemble` attribute (1 for a freshly loaded restart), not the actual sample count, so if `n_ensemble_per_ic>1` then segments after the first re-broadcast the ensemble (`n_ensemble` → `n_ensemble**2` samples) and OOM the GPU. Rather than silently de-broadcasting on restart segments — which quietly reinterprets the user's config — this PR rejects the combination up front: `n_ensemble_per_ic>` is unsupported with segmented inference, and should use a single non-segmented inference.

Changes:
- `fme.ace.inference.run_segmented_inference` — raise `ValueError` when `n_ensemble_per_ic > 1`, before any segment runs
- `fme.ace.inference.test_segmented` — test asserting `run_segmented_inference` raises `ValueError` for `n_ensemble_per_ic > 1`

- [x] Tests added

### DIFF
--- a/fme/ace/inference/inference.py
+++ b/fme/ace/inference/inference.py
@@ -412,6 +412,14 @@ def run_segmented_inference(config: InferenceConfig, segments: int):
         multiple folders, each corresponding to one of the segments and labeled by
         the segment number.
     """
+    if config.n_ensemble_per_ic > 1:
+        raise ValueError(
+            "Ensemble inference (n_ensemble_per_ic > 1) is not supported with "
+            "segmented inference. A segment's restart already carries the "
+            "broadcasted ensemble as its sample dimension, so later segments "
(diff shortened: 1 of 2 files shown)


Arcomano1234 (approved):
LGTM. In the future (assume we find a use case for large IC ensemble members + segmented config),  a potential avenue to support this feature would be having an optional `n_ensemble` dimension in the `restart.nc` so that `n_ensemble_per_ic` could get properly propagated across segments.

PR #1302, opened 2026-06-22
Refactor step to take an explicit [batch, ensemble] dimension pair

mcgibbon:
The `step`/stepper now receive data with an explicit `[batch, ensemble, *spatial]` leading dimension pair (via `BatchData.ensemble_data`) instead of a folded `[batch*ensemble, *spatial]` dimension paired with an `n_ensemble: int` that re-interprets it. This makes the data self-describing (`n_ensemble = shape[1]`) — no integer side-channel — which is the prerequisite for input channel dropout that is shared across ensemble members but independent across the batch (a follow-up, not implemented here).

The ensemble dimension is folded back into the batch at the entry of `step_with_adjustments` (each ensemble member is an independent sample), so the per-sample body — normalization, global mean removal, the network call, the corrector, the ocean — runs on the folded batch exactly as before, and the output is unfolded back. The change is therefore behavior-preserving (the step body is byte-identical to `main` on the folded batch).

Changes:
- `fme.ace.data_loading.BatchData.ensemble_data` (existing) returns `[batch, ensemble, time, *spatial]`; `n_ensemble` length 1 is valid even before `broadcast_ensemble`. No other accessors are added — `data_mask` (constant across ensemble) and `stepper_state` stay in their folded layout.
- `Stepper.predict_generator` consumes `BatchData.ensemble_data` (no `n_ensemble` argument); `data_mask`/`stepper_state` are passed in their folded form. `Stepper.TIME_DIM` stays 1 (the coupled stepper indexes its folded component data through it); predict_generator uses a local `time_dim = TIME_DIM + 1`.
- `step_with_adjustments` folds the explicit ensemble dimension into the batch at entry and unfolds the output; the per-op step internals are unchanged from `main`.
- `process_prediction_generator_list` stacks the explicit yields and folds back into `BatchData`'s folded storage (deriving `n_ensemble` from the shape).
- `fme.coupled.stepper` drives its components over a flat batch; it folds the components' explicit yields at the boundary and treats the component initial conditions as a flat (`n_ensemble=1`) view matching the freshly built forcings.

Verification:
- `test_step_ensemble_members_are_independent`: each ensemble member's output equals an independent single-member step.
- Spatial-parallel `test_step_regression` passes under `H=2` and `W=2` with regenerated `[batch, ensemble]` baselines (values reproduce the pre-refactor outputs modulo the size-1 ensemble dim).
- ace step/stepper, coupled, inference, and aggregator suites pass.

- [x] Tests added


fme/ace/data_loading/batch_data.py
```diff
@@ -213,13 +213,34 @@ def n_timesteps(self) -> int:
...
-            The tensor dict with an explicit ensemble dimension.
+        Derived from the folded ``data`` and ``n_ensemble`` (an ensemble length
+        of 1 is valid even before ``broadcast_ensemble`` is called). Consumers
+        that need the ensemble dimension should read this rather than ``data``
+        plus an ``n_ensemble`` argument.
         """
         return unfold_ensemble_dim(TensorDict(self.data), n_ensemble=self.n_ensemble)
 
+    @property
+    def ensemble_data_mask(self) -> TensorMapping | None:
```
mcgibbon: Why would we need this? If data is missing from a batch member, it will be missing from all ensemble instances. Can't we just broadcast later?

fme/ace/data_loading/batch_data.py
```diff
@@ -213,13 +213,34 @@ def n_timesteps(self) -> int:
...
+        """The per-variable ``data_mask`` with an explicit ``[batch, ensemble]``
+        layout, or ``None``. Companion to :attr:`ensemble_data`.
+        """
+        if self.data_mask is None:
+            return None
+        return unfold_ensemble_dim(dict(self.data_mask), n_ensemble=self.n_ensemble)
+
+    @property
+    def ensemble_stepper_state(self) -> "StepperState | None":
```
mcgibbon: Question: why do we need this? Shouldn't the stepper state be able to always know internally what its ensemble count is?

fme/ace/stepper/single_module.py
```diff
@@ -1108,26 +1113,33 @@ def _input_only_names(self) -> list[str]:
 
     def predict_generator(
         self,
-        ic_dict: TensorMapping,
-        forcing_dict: TensorMapping,
+        ic_data: TensorMapping,
```
mcgibbon: unnecessary name change, please revert. This change is big enough that we should avoid changelog lines where possible.

fme/ace/stepper/single_module.py
```diff
@@ -1108,26 +1113,33 @@ def _input_only_names(self) -> list[str]:
...
         stepper_state: StepperState | None = None,
     ) -> Generator[tuple[TensorDict, StepperState | None], None, None]:
-        state = {k: ic_dict[k].squeeze(self.TIME_DIM) for k in ic_dict}
+        # All tensors carry an explicit [batch, ensemble, time, *spatial] layout
+        # (e.g. BatchData.ensemble_data / ensemble_data_mask / ensemble_stepper_
+        # state). The ensemble dimension is never folded into the batch outside
+        # the nn.Module call, which folds transiently inside the Step; the
+        # generator yields the same explicit layout. The time dimension is thus
+        # one past TIME_DIM (the ensemble dim is inserted before it).
+        time_dim = self.TIME_DIM + 1
```
mcgibbon, line 1127: is it reasonable to instead update self.TIME_DIM to one higher? If so we don't need this lengthy comment.

fme/ace/data_loading/batch_data.py
mcgibbon: Removed `ensemble_data_mask`. You're right — the data_mask is constant across ensemble members, so it no longer gets a `[batch, ensemble]` view. The step now folds the `[batch, ensemble]` data into the batch at entry and applies the per-sample (folded `[batch*ensemble]`) data_mask on the folded batch, exactly as on main.

fme/ace/data_loading/batch_data.py
mcgibbon: Removed `ensemble_stepper_state`. The step now folds the ensemble into the batch at entry, so the corrector runs on the folded batch and the stepper_state stays in its folded `[batch*ensemble, ...]` layout throughout — no unfolding, and the CorrectorState/StepperState fold-helpers are gone too.

fme/ace/stepper/single_module.py
mcgibbon: Reverted — back to `ic_dict`/`forcing_dict`.

fme/ace/stepper/single_module.py
mcgibbon: I tried bumping `self.TIME_DIM` to 2, but the coupled stepper indexes its *folded* component data via `self.atmosphere.TIME_DIM`/`self.ocean.TIME_DIM` (time at dim 1) — bumping it broke 39 coupled tests. So I kept `Stepper.TIME_DIM = 1` and use a local `time_dim = self.TIME_DIM + 1` in predict_generator (the ensemble dim sits before time in `ensemble_data`), and shortened the comment to one line.

mcgibbon:
Closing without merging. This refactor was a prerequisite only for ensemble-scoped input masking. Alexey's VarMasking6 runs show batch-uniform masking (same_mask_per_batch=true) is as good or better than independent per-row masking (ties at a 10-var budget, clearly better at 20 vars on skill, CRPS, and PRATEsfc calibration), so ensemble-scoped masking — the only variant needing this refactor — is not worth building. Decision documented in research report 2026-07-01-var-masking-ensemble-scoping (investigation 2026-07-01-ensemble-scoped-input-masking-vs-batch-uniform). Keeping the simple batch-level masking on feature/var_masking_simple.

PR #1303, opened 2026-06-22
Add per-sample spectral whitening option to the energy-score loss

mcgibbon:
Small-scale (high-`l`) spectral power converges very slowly in SFNO training: the energy score lives in a red spectrum, so high-`l` modes carry tiny amplitude, contribute little to the score, and their `dhconv` filter weights see proportionally small, low-SNR gradients that the energy score's uniform per-`(l,m)` `mode_weights` do not correct.

This PR adds an opt-in per-sample **spectral whitening** mode to `EnergyScoreLoss`. It reweights the per-`(l,m)` energy score by a power of the inverse per-degree RMS amplitude of each target sample, flattening the target's angular power spectrum so high-`l` modes are no longer starved and their gradient SNR rises. The reweight is computed from the **detached** target coefficients (no new gradient path) and is **magnitude-preserving** (rescaled per `(sample, channel)`, so the overall energy-score magnitude — and the meaning of `energy_score_weight` — is unchanged; only the per-scale balance shifts). A white-spectrum target yields a uniform factor (no-op), and the default (whitening off) is bit-for-bit backward compatible.

Whitening strength is the exponent γ ∈ (0, 1] (per-degree factor `(1/amp_l)**γ`), interpolating between no reweighting (γ→0) and full flattening (γ=1). Enabling whitening (`kind='per_sample'`) defaults to γ=0.5, the validated stable knee; full flattening (γ=1) over-upweights noise-dominated low-amplitude degrees and destabilizes residual rollouts, so it is opt-in. Options are grouped in a `SpectralWhiteningConfig` (`kind`, `eps_frac`, `exponent`); `kind='none'` (default) disables whitening, and `eps_frac`/`exponent` left unset resolve to `0.02`/`0.5` when enabled (setting either with `kind='none'` raises). In a loss config it is a nested block:

```yaml
loss:
  type: EnsembleLoss
  kwargs:
    crps_weight: 0.9
    energy_score_weight: 0.1
    energy_score_whitening:
      kind: per_sample
      exponent: 0.5
```

Changes:
- `fme.core.loss.SpectralWhitening` / `SpectralWhiteningConfig` — new operator and config for per-sample whitening. The config (validated in `__post_init__`) `build()`s the operator, or `None` when `kind='none'`; the operator owns the reweight (per-`(sample, channel, l)`, detached, floored at `eps_frac` of the per-sample mean degree amplitude, raised to `exponent`, magnitude-preserving).
- `fme.core.loss.EnergyScoreLoss` / `EnsembleLoss` — take the built `whitening: SpectralWhitening | None` operator (not the config) and apply its factor after `mode_weights`.
- `fme.core.loss.LossConfig.build` — constructs `SpectralWhiteningConfig` from the opaque loss `kwargs` dict, builds the operator, and passes it down.
- `fme.core.test_loss` — no-op default, white-target invariance, small-scale boost, magnitude preservation, exponent range, symmetric `per_sample`-required validation of both knobs, resolved `per_sample` defaults, partial-whitening interpolation, and config wiring via dict and dataclass.

- [x] Tests added


fme/core/loss.py
```diff
@@ -484,13 +484,73 @@ class EnergyScoreLoss(torch.nn.Module):
     .. [1] https://sites.stat.washington.edu/people/raftery/Research/PDF/Gneiting2007jasa.pdf
     """
 
-    def __init__(self, sht: Callable[[torch.Tensor], torch.Tensor]):
+    def __init__(
+        self,
+        sht: Callable[[torch.Tensor], torch.Tensor],
+        spectral_whitening: str = "none",
```
mcgibbon: Make this a Literal[(options)] instead of str

fme/core/loss.py
```diff
@@ -595,6 +657,8 @@ def __init__(
         finite_difference_crps_weight: float = 0.0,
         finite_difference_crps_levels: int = 1,
         almost_fair_crps_alpha: float = 1.0,
+        energy_score_whitening: str = "none",
```
mcgibbon: Similar type comment.

mcgibbon:
Full analysis and figures: [whitening-γ-selection report](https://github.com/ai2cm/explore2/blob/main/jeremym/2026-07-08-whitening-gamma-selection/report.md)

### `whitening_exponent` (γ) knob added + validation

The latest commit adds the `whitening_exponent` γ knob (`(1/amp_l)**γ`, γ ∈ (0, 1], default 1 = prior full-whitening behavior). Motivation: full whitening (γ=1) maximizes the small-scale gradient boost but over-upweights noise-dominated low-amplitude high-`l` degrees, which destabilizes residual-prediction rollouts (stratospheric-temperature-led blowup). Partial whitening keeps most of the spectral win while staying stable.

**γ selected: 0.5.** A two-seed sweep of γ ∈ {0, 0.2, 0.5, 0.8} on the 4°/daily v2 ERA5-only residual recipe (matched epoch 120, seed-pair noise sizing) found γ=0.5 the unique arm clearing all four adoption conditions: (a) large smallest-scale spectral win well outside seed noise (h500 −51% / >7σ, TMP2m −86%; the win saturates by γ=0.5, γ=0.8 adds only −5 pts), (b) 10-yr rollout health within seed noise of baseline, (c) 1-step val CRPS within seed noise, (d) no regression on the 46-yr baseline surface. γ=0.8 fails (d) with a reproducible two-seed 46-yr stratospheric regression; γ=0.2 under-delivers (a) and destabilizes one seed.

Sweep runs (γ, seed):

| γ | seed 0 | seed 1 |
|---|---|---|
| 0 (baseline) | wandb `oshj5u79` | [beaker 01KWW0J0KWPGDXT2X5PNXY2SBG](https://beaker.org/ex/01KWW0J0KWPGDXT2X5PNXY2SBG) / [wandb eyfz92l8](https://wandb.ai/ai2cm/ace/runs/eyfz92l8) |
| 0.2 | [beaker 01KWW0EPJZKKE25P6ZSWHEPS7S](https://beaker.org/ex/01KWW0EPJZKKE25P6ZSWHEPS7S) / [wandb bwbg9r2k](https://wandb.ai/ai2cm/ace/runs/bwbg9r2k) | [beaker 01KWW0FBKYYNA7R9ET1Z6T1JF4](https://beaker.org/ex/01KWW0FBKYYNA7R9ET1Z6T1JF4) / [wandb jjprdcdt](https://wandb.ai/ai2cm/ace/runs/jjprdcdt) |
| **0.5 (selected)** | [beaker 01KWHS1DHTR858DC51720J1Q5C](https://beaker.org/ex/01KWHS1DHTR858DC51720J1Q5C) / [wandb erug2fqs](https://wandb.ai/ai2cm/ace/runs/erug2fqs) | [beaker 01KWW0HBG0509A1C6Q86YQA803](https://beaker.org/ex/01KWW0HBG0509A1C6Q86YQA803) / [wandb 4dltgowi](https://wandb.ai/ai2cm/ace/runs/4dltgowi) |
| 0.8 | [beaker 01KWW0G0SN0383BM8FSD3SR940](https://beaker.org/ex/01KWW0G0SN0383BM8FSD3SR940) / [wandb 8vgzlp54](https://wandb.ai/ai2cm/ace/runs/8vgzlp54) | [beaker 01KWW0GNXKVWRBKGY3DMAY12VA](https://beaker.org/ex/01KWW0GNXKVWRBKGY3DMAY12VA) / [wandb xen5ofv9](https://wandb.ai/ai2cm/ace/runs/xen5ofv9) |

Caveat carried by the verdict: the γ arms stop at epoch 120 while the baseline (extended to 259) drifts past ~ep 150, so γ=0.5 is verified clean *through 120* but not yet proven not to merely delay the same intrinsic residual-recipe drift; a decisive hardening run (extend the two γ=0.5 seeds past ep 150) is tracked separately.

---
Note: this commit also fixes a pre-existing `pre-commit` (mypy) CI failure — the `test_energy_score_invalid_whitening_raises` test passes a deliberately-invalid literal, now marked `# type: ignore[arg-type]`.

fme/core/loss.py
```diff
@@ -484,13 +484,89 @@ class EnergyScoreLoss(torch.nn.Module):
...
+        is ``1 / amp_l``, floored at ``whitening_eps_frac`` of the per-sample mean
+        degree-amplitude to bound the boost where the target has near-zero power.
+        ``whitening_exponent`` gamma < 1 applies partial whitening
+        (``(1/amp_l)**gamma``), interpolating between no reweighting at 0 and full
+        flattening at 1; it tames the upweighting of noise-dominated low-amplitude
+        degrees (e.g. high-l residual-prediction targets).
+        The factor is then rescaled per (sample, channel) so the amplitude-
+        weighted total weight matches the unwhitened weighting, leaving the
+        overall energy-score magnitude (and the meaning of ``energy_score_weight``)
+        approximately unchanged.
```
mcgibbon: Issue: This is far too verbose. Cite something if we need this detail, or be precise and concise.

fme/core/loss.py
```diff
@@ -595,6 +673,9 @@ def __init__(
         finite_difference_crps_weight: float = 0.0,
         finite_difference_crps_levels: int = 1,
         almost_fair_crps_alpha: float = 1.0,
+        energy_score_whitening: Literal["none", "per_sample"] = "none",
+        energy_score_whitening_eps_frac: float = 0.02,
+        energy_score_whitening_exponent: float = 1.0,
```
mcgibbon: Could we refactor this into a new configuration dataclass (default value none -> don't apply) that produces a built object that keeps these floats and the Literal private to that operation?

mcgibbon: Claude: Done in d6e020d1b. Added a `SpectralWhiteningConfig` dataclass (`kind` / `eps_frac` / `exponent`, validated in `__post_init__`) whose default `kind='none'` disables whitening; its `build()` returns a `SpectralWhitening` operator (which owns the floats and the reweight computation) or `None`. `EnergyScoreLoss` and `EnsembleLoss` now take one config instead of the loose floats, so the params/Literal stay private to the operation. Mirrors the `global_mean_removal` config→operator pattern. Since the loss `kwargs` mapping is opaque to dacite, `LossConfig.build` constructs the config from its YAML dict.

fme/core/loss.py
mcgibbon: Claude: Trimmed. The long derivation is gone; the whitening docstring now states what the factor is and its key properties (detached, magnitude-preserving, exponent γ, white-target no-op) in a few lines, with the step-by-step mechanics kept as inline code comments. It also moved onto the new `SpectralWhitening` operator (d6e020d1b).

fme/core/loss.py
```diff
@@ -460,6 +460,91 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> list[LossComponent]:
...
+
+    ``kind='none'`` (the default) disables it. ``exponent`` gamma in (0, 1] sets
+    the whitening strength (gamma=1 fully flattens the target spectrum; smaller
+    values whiten partially, taming the upweighting of noise-dominated
+    low-amplitude degrees). ``eps_frac`` floors the per-degree boost.
+    """
+
+    kind: Literal["none", "per_sample"] = "none"
+    eps_frac: float = 0.02
+    exponent: float = 1.0
```
mcgibbon: Because it's not obvious, include documentation of the meaning of these arguments.

fme/core/loss.py
```diff
@@ -484,9 +569,14 @@ class EnergyScoreLoss(torch.nn.Module):
...
 
-    def __init__(self, sht: Callable[[torch.Tensor], torch.Tensor]):
+    def __init__(
+        self,
+        sht: Callable[[torch.Tensor], torch.Tensor],
+        whitening: SpectralWhiteningConfig | None = None,
+    ):
         super().__init__()
         self.sht = sht
+        self._whitening = (whitening or SpectralWhiteningConfig()).build()
```
mcgibbon: Isn't this always non-None? But you're checking for None below? Choose one.

fme/core/loss.py
mcgibbon: Claude: Done in 801a719d2. Added an Args block to `SpectralWhiteningConfig` documenting each argument — `kind` (none disables / per_sample enables), `eps_frac` (the per-degree amplitude floor, as a fraction of the per-sample mean, bounding the boost at near-zero-power degrees), and `exponent` (whitening strength γ, weight `(1/amp_l)**γ`, γ=1 full / γ<1 partial).

fme/core/loss.py
mcgibbon: Claude: Good catch on the readability — the `(whitening or SpectralWhiteningConfig())` config is always non-None, but `.build()` returns `SpectralWhitening | None` (None when `kind='none'`), so `self._whitening` genuinely can be None and the check in forward is deliberate: it skips the whole reweight when whitening is disabled (rather than building a no-op operator that still runs the SHT/factor math every step). I kept the None path and made the intent explicit — annotated `self._whitening: SpectralWhitening | None` with a comment saying build() returns None when disabled (801a719d2).

mcgibbon:
Claude: Applied self-review follow-ups (pushed as `7df72cf9`), updating the earlier γ-knob note above:

- **Default γ flipped 1 → 0.5.** Enabling `kind='per_sample'` now defaults to the sweep-selected stable knee (0.5), not full whitening — shipping the opt-in defaulted to the value the sweep found destabilizing was a footgun. `eps_frac`/`exponent` are now `None`-sentinel and resolve to the validated defaults only when enabled; setting either with `kind='none'` raises symmetrically (previously only `exponent` did).
- **Build moved to the config layer.** `LossConfig.build` now builds the `SpectralWhitening` operator and passes it down; `EnergyScoreLoss`/`EnsembleLoss` take the built operator instead of the config, so the nn.Modules no longer build their own sub-config.
- **`factor` cleanup.** Drops the target's singleton ensemble dim up front (`y_hat[:, 0]`) instead of a trailing conditional squeeze, and computes `|y_hat|` once.

New tests cover the `eps_frac`-with-none guard and the resolved `per_sample` defaults. Full loss suite (72) + mypy/ruff pass locally.

mcgibbon:
Report for reviewer: https://github.com/ai2cm/reports/pull/2/changes#diff-a2d53a810a58b87bed5ab372f8611c88b4041df7d758ef0afb29b6112ca3c0e5

mcgibbon:
Claude: Empirical motivation and validation, moved out of the description so it reads as a clean commit message (detail kept here):

**Motivation — gradient probe.** On a trained checkpoint, a per-`l` gradient probe showed an ~8× monotone decline of high-`l` gradient magnitude across degrees, confirming the high-`l` `dhconv` filter weights are gradient-starved under the red-spectrum energy score.

**Validation.** On 4°/daily residual runs, base whitening delivered a large small-scale spectral-fidelity gain at neutral CRPS. A two-seed γ ∈ {0, 0.2, 0.5, 0.8} sweep selected **γ=0.5** as the stable knee — the default this PR ships for `kind='per_sample'`. Full run table, adoption criteria, and the whitening-γ-selection report are in the sweep comment above.

fme/core/loss.py
```diff
@@ -460,6 +460,128 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> list[LossComponent]:
...
+    (sample, channel) to preserve the amplitude-weighted total, so the overall
+    energy-score magnitude (and the meaning of ``energy_score_weight``) is
+    unchanged. A white-spectrum target yields a uniform factor (no-op).
+    """
+
+    def __init__(self, eps_frac: float = 0.02, exponent: float = 1.0):
+        self.eps_frac = eps_frac
+        self.exponent = exponent
+
+    def factor(self, y_hat: torch.Tensor) -> torch.Tensor:
```
Arcomano1234, line 483: Optional: Very minor nit but there are a lot of constant values that can be cached as they get recomputed every time `factor` is called. E.g., `l_idx`, `m_idx`, `valid`, and `redundancy`. I know we some times caches these and it might be worth it.

Arcomano1234 (approved):
Looks good, I didn't give a thorough review of the math in `SpectralWhitening` but claude thought it looked good and the results suggest that. I did leave an optional comment about caching a lot of the variables that get recalculated when `factor` is called.

PR #1304, opened 2026-06-22
Refactor step to take [batch, ensemble] dims (on var_masking_simple)

mcgibbon:
Same refactor as #1302, based on `feature/var_masking_simple` so the explicit `[batch, ensemble]` step contract is integrated with the input-channel-dropout work on that branch. **Base: `feature/var_masking_simple`** (not `main`). Mirrors the review changes made on #1302.

The `step`/stepper receive data with an explicit `[batch, ensemble, *spatial]` leading dimension pair (via `BatchData.ensemble_data`) — self-describing, no `n_ensemble` integer side-channel. `step_with_adjustments` folds the ensemble back into the batch at entry (each member is an independent sample), so the per-sample body — incl. the input-dropout masking in `network_call` — runs on the folded batch exactly as before, and the output is unfolded. Behavior-preserving (the step body is byte-identical to the base on the folded batch).

On top of #1302's changes, integrated with var_masking's input dropout:
- The input-dropout mask is built per base sample and repeat-interleaved across the ensemble into the folded `[batch*ensemble]` layout (shared across ensemble members of a base sample, independent across the batch — matching `broadcast_ensemble`'s block ordering); `network_call` applies it on the folded batch. No `[batch, ensemble]` mask is materialized.
- `StepArgs.input_dropout_mask` flows through `predict_generator` in its folded layout.

(The ensemble-shared/batch-independent dropout behavior is unchanged; this refactor just gives the step the explicit `[batch, ensemble]` data contract.)

Verification: ace step / stepper / tensors / var_masking unit suites pass (incl. the input-dropout and ensemble tests); data_loading + inference + aggregator + coupled pass; spatial-parallel `test_step_regression` passes under `H=2` and `W=2`.

- [x] Tests added


mcgibbon:
Closing without merging. This refactor was a prerequisite only for ensemble-scoped input masking. Alexey's VarMasking6 runs show batch-uniform masking (same_mask_per_batch=true) is as good or better than independent per-row masking (ties at a 10-var budget, clearly better at 20 vars on skill, CRPS, and PRATEsfc calibration), so ensemble-scoped masking — the only variant needing this refactor — is not worth building. Decision documented in research report 2026-07-01-var-masking-ensemble-scoping (investigation 2026-07-01-ensemble-scoped-input-masking-vs-batch-uniform). Keeping the simple batch-level masking on feature/var_masking_simple.

PR #1305, opened 2026-06-22
Frozen precip corrector via clipping

### LINKED ISSUE #1269: Total frozen precipitation is currently unconstrained
We now use `total_frozen_precipitation` as a diagnosed variable in several models / datasets. Currently, this variable is completely unconstrained and does not depend on surface_total_precipitation. This can lead to ACE predictions where `total_frozen_precipitation` exceeds the predicted `surface_total_precipitation` in a time step (see examples below). 

I propose an additional step in the `advection_and_precipitation` corrector that does a grid point by grid point clipping of the `total_frozen_precipitation` so that it does not exceed the predicted `surface_total_precipitation`. 

Example of max  `total_frozen_precipitation` exceedance over `PRATEsfc` from a 10-day forecast from ACE2S-ERA5 with `total_frozen_precipitation`. We see several grid points with 0.5mm/6hr+ exceedance which is about two orders of magnitude greater than what occurs in ERA5.
Image

Potential proposals for fixes include:

1. Predicting `fraction_of_total_precip_thats_frozen` where `total_frozen_precipitation` = `fraction_of_total_precip_thats_frozen`  * `surface_total_precipitation`
2. Corrector based approach that clips `total_frozen_precipitation` to `surface_total_precipitation`
3. Have `total_frozen_precipitation` and `total_liquid_precipitation` be independently predicted and make `total_precipitation` a derived variable where `surface_total_precipitation` =  `total_frozen_precipitation` + `total_liquid_precipitation`

Arcomano1234:
ACE can produce frozen precipitation (`total_frozen_precipitation_rate`) exceeding the total precipitation rate (`PRATEsfc`), which is unphysical since frozen precipitation is a component of total precipitation. This PR clips frozen precipitation to the total precipitation rate (downstream of the correctors that can change `PRATEsfc`), but upstream of the energy correctors which utilizes `total_frozen_precipitation_rate`. The clip is off by default so that previously trained checkpoints are unaffected.

Here is the impact of the correctors show as the fraction of PRATEsfc that is frozen precipitation 
image
For more analysis see this [notebook](https://github.com/ai2cm/explore2/blob/main/troya/2026-06-12-frozen-precip-analysis/2026-06-12-frozen-precip.ipynb).

Changes:
- `fme.core.corrector.atmosphere.AtmosphereCorrectorConfig`: add a `clip_frozen_precipitation` option (default `False`, preserving behavior of existing checkpoints).
- `fme.core.corrector.atmosphere.MoistureBudgetCorrection`: when enabled, after closing the moisture budget, clip `total_frozen_precipitation_rate` to the corrected `PRATEsfc` when frozen precipitation is predicted. This runs before any total-energy-budget correction, since frozen precipitation contributes to the surface energy flux via the latent heat of freezing.
- `fme.core.atmosphere_data.AtmosphereData`: add a `set_frozen_precipitation_rate` setter to support the clip.

- [x] Tests added

Resolves #1269


mcgibbon:
With the corrector we're strongly under-predicting non/zero-one values and over-predicting the edge values. I think for this task we particularly want to propagate gradients through the clip. It would be nice to see what that does to the result, before we use it (but I don't think blocks review/merge of the feature).

Arcomano1234:
> With the corrector we're strongly under-predicting non/zero-one values and over-predicting the edge values. I think for this task we particularly want to propagate gradients through the clip. It would be nice to see what that does to the result, before we use it (but I don't think blocks review/merge of the feature).

Considering I need to re-run the baseline training I will re-run this training with the propagate gradients through the clip feature one to see how much this makes a difference.

fme/core/corrector/atmosphere.py
```diff
@@ -161,7 +168,11 @@ def __call__(
             timestep_seconds=self.timestep_seconds,
             terms_to_modify=self.terms_to_modify,
         )
-        return corrected, corrector_state
+        # Clip frozen precipitation against the (possibly corrected) total
+        # precipitation rate. Done here, after the budget correction, so the
+        # ceiling is the final precipitation rate.
+        frozen_clip = _clip_frozen_precipitation({**gen_data, **corrected})
```
brianhenn: I don't have strong opinions either way, but as implemented this changes the behavior of existing steppers (probably not that many of them out there that predict frozen precipitation still). An alternative would be to have this off by default (and maybe then explicitly turn it on in the baseline config). Not sure if the extra config arg is really worth it or not.

Arcomano1234: Yeah that was what I previous had but I can't imagine anyone would want this off even for old check points. Right now only ACE-SHiELD+ is a public ckpt with frozen precip. 

@spencerkclark is it possible for you to run this feature with your ACE-SHiELD+ ckpt. If you can just run a 10 day forecast saving PRATEsfc and total_frozen_precip with and without this feature, I can run it through my notebook. I suspect that this feature will actually help your `fraction_of_precip_that_is_frozen` without even needing to retrain but I also want to confirm it. What I did just realize is this will affect the model behavior as this happens before the energy correctors (and thus can affect the model evolution) so we should confirm this doesn't break your ckpt. Can you also run a 1000 year simulation with this (you shouldn't have to save any output)?

spencerkclark: Thanks, yes, I agree this is important to take a look at.  I'll run the experiments you suggest and maybe a couple others.

Arcomano1234:
> With the corrector we're strongly under-predicting non/zero-one values and over-predicting the edge values. I think for this task we particularly want to propagate gradients through the clip. It would be nice to see what that does to the result, before we use it (but I don't think blocks review/merge of the feature).

The propagation of gradients through the clipping (`keep_gradient_through_clamps: true)` did not seem to work well for the new baseline ACE2S-ERA5 ([wandb](https://wandb.ai/ai2cm/ace/runs/8lg5mg56)). After 10ish epochs the run exploded. I suggest we do not try to use this feature at this time.

fme/core/corrector/test_atmosphere.py
```diff
@@ -276,6 +277,156 @@ def test_force_conserve_moisture(
...
+    torch.manual_seed(0)
+    tensor_shape = (1, 5, 5)
+    input_data, gen_data, forcing_data, vertical_coord = _get_corrector_test_input(
+        tensor_shape
+    )
+    gen_data = {
+        **gen_data,
+        "total_frozen_precipitation_rate": 1e-4 + 1e-4 * torch.rand(size=tensor_shape),
+    }
+    assert torch.any(gen_data["total_frozen_precipitation_rate"] > gen_data["PRATEsfc"])
```
spencerkclark: nit: it's unlikely this would ever fail, but should we make this test deterministic?

spencerkclark (approved):
Thanks @Arcomano1234 for making this optional and default off.  The updates look good to me!  I will follow up on Slack regarding the results of the sensitivity experiments with ACE2S-SHiELD+.

PR #1307, opened 2026-06-23
Add 2-degree (F45) ERA5 dataset support

mcgibbon:
Adds support for generating a 2-degree (F45) ERA5 dataset, needed for the multi-resolution SHT latent-stepping emulator (1°↔2° prototype): both the native-2° baseline and the 2° training data. The 2° product mirrors the existing 1-degree 8-layer / daily / 1940–2025 ERA5 dataset so configs port across resolutions, and is produced the same way (conservative xESMF regrid from native 0.25°) as the 1° and 4° datasets so the native-2° baseline is comparable.

The F45 grid is 90×180 Gauss–Legendre (`leggauss(90)`), which is exactly the grid the model's SHT lands on when truncating 1°→2° (the ACE registries set `data_grid="legendre-gauss"`), so the data grid is consistent with the model's internal resolution change.

Changes:
- `scripts/era5/pipeline/xr-beam-pipeline.py`: register `F45` (N=45) in `GAUSSIAN_GRID_N`; update `--output_grid` help. Regridding (conservative xESMF) and 8-layer vertical default are unchanged.
- `scripts/era5/Makefile`: add `era5_dataflow_two_degree` (`output_grid=F45`, 1940–2025). Kept out of the `era5_dataflow` aggregate so it doesn't relaunch the already-produced 1°/4° runs.
- `scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml`: new config — chunking/sharding 90×180, factor-4 daily coarsen, stats 1990–2019, same ~70-variable set as the 1° config.
- `scripts/data_process/Makefile`: add `era5_2deg_stats_beaker_dataset`.

No Dataflow or training jobs have been launched; running the pipeline is gated on review/go-ahead.

- [ ] Tests added (none — no test infrastructure exists for the ERA5 ingest pipeline; F45 grid math verified to produce 90×180 Gauss–Legendre at 2° spacing)
- [ ] If dependencies changed, "deps only" image rebuilt (n/a — no dependency changes)

### DIFF
--- a/scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml
@@ -1,0 +1,114 @@
+runs:
+  2026-06-23-era5-2deg-8layer-1940-2025: ""  # no real data source, config only for computing stats
+data_output_directory: gs://vcm-ml-intermediate
+stats:
+  output_directory: gs://vcm-ml-intermediate/2026-06-23-era5-2deg-8layer-stats-1990-2019
+  beaker_dataset: 2026-06-23-era5-2deg-8layer-stats-1990-2019
+  start_date: "1990-01-01"
+  end_date: "2019-12-31"
+  data_type: ERA5
+time_coarsen:
+  chunking:
+    time: 1
+    latitude: 90
+    longitude: 180
+  sharding:
+    time: 360
+    latitude: 90
+    longitude: 180
+  factor: 4
+  data_output_directory: gs://vcm-ml-intermediate/2026-06-23-era5-2deg-8layer-daily-1940-2025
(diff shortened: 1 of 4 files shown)


scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml
```diff
@@ -0,0 +1,114 @@
...
+  start_date: "1990-01-01"
+  end_date: "2019-12-31"
+  data_type: ERA5
+time_coarsen:
+  chunking:
+    time: 1
+    latitude: 90
+    longitude: 180
+  sharding:
+    time: 360
```
spencerkclark: Suggestion: could probably bump this up by a factor of four to `1440` to keep the shard sizes (in MB) similar between the 1° and 2° time-coarsened datasets:
```suggestion
    time: 1440
```

spencerkclark (approved):
Looks good to me.

PR #1310, opened 2026-06-24
Inline +4K SST perturbation-response evaluation during training

mcgibbon:
Make the +4 K uniform-SST warming response visible **during training** by
running an unperturbed baseline and a perturbed climate together as separate
batch members of a single inline rollout at evaluated checkpoints, differencing
their time-means, and logging the structural diagnostics the land-amplification
goal is judged on. This is a diagnostic eval only — it never contributes to
checkpoint selection — and involves no external (c96-SHiELD) reference; it
reports the model's own response structure.

This is a new inline-inference type (peer to inline inference), not a change to
the inference evaluator: the response needs a 0 K baseline climatology, so its
top-level operation differs (run + difference two climates). It reuses the
existing `InferenceTask`/`InferenceCallback` plumbing with `weight=0`.

Changes:
- `fme.ace.aggregator.inference.perturbation_response` — new `PerturbationResponseAggregator` + `PerturbationResponseAggregatorConfig`: accumulates a per-group time-mean over baseline/perturbed batch members (one-hot group encoding, group 0 = baseline; >1 perturbation raises `NotImplementedError`), differences each perturbation against the baseline, and logs near-surface land/ocean warming ratio by latitude band, free-troposphere/surface warming ratio over tropical ocean, and the global-mean column warming profile. NaN-guarded ratios; cross-rank reduction via `reduce_sum`.
- `fme.ace.data_loading.perturbation_pair.build_perturbation_pair_data` — turns an unperturbed inference IC + forcing loader into a paired baseline/perturbed rollout. Perturbs every forcing window (ocean-masked) and the prognostic IC when surface temperature is prognostic, with a configurable IC-perturbation scope (whole-field vs ocean-masked). Returns the `[2*n_local, 2]` one-hot group encoding.
- `fme.ace.train.train_config` — `LabeledPerturbation`, `PerturbationResponseInferenceConfig`, the `TrainConfig.perturbation_inference` field, name/label validation, epoch sets, and `TrainBuilders.get_perturbation_inference_data`.
- `fme.ace.train.train` — `get_inference_callback` also builds weight-0 perturbation-response tasks; `_make_perturbation_response_task` does the pairing; `build_trainer` wiring.
- `fme.ace.stepper.single_module.TrainStepper` — expose `surface_temperature_name`/`ocean_fraction_name` (delegating to the inner stepper).

- [x] Tests added (aggregator unit tests, paired-data unit tests, end-to-end training smoke test asserting the response log keys + diagnostics appear and that the eval does not affect checkpoint selection)
- [ ] If dependencies changed, "deps only" image rebuilt — n/a

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -12,6 +12,10 @@ from fme.ace.aggregator.inference.main import (
     LegacyFlagInferenceEvaluatorAggregatorConfig,
     StepMeanEntry,
 )
+from fme.ace.aggregator.inference.perturbation_response import (
+    LatitudeBand,
+    PerturbationResponseAggregatorConfig,
+)
 from fme.ace.aggregator.inference.reduced import MeanMetricConfig
 from fme.ace.aggregator.inference.seasonal import SeasonalMetricConfig
 from fme.ace.aggregator.inference.spectrum import PowerSpectrumMetricConfig
@@ -134,8 +138,10 @@ from .train.train_config import (
     EMAConfig,
     InlineInferenceConfig,
     InlineValidationConfig,
+    LabeledPerturbation,
     LoggingConfig,
     OptimizationConfig,
+    PerturbationResponseInferenceConfig,
     TrainConfig,
 )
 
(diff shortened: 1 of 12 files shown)


mcgibbon:
We're deciding not to merge this and instead to start doing our training and evaluation on shield, where it's just any other kind of out-of-sample target inference.

PR #1311, opened 2026-06-24
Add inference trend-map aggregator

mcgibbon:
Long climate inference runs can be too large to hold in memory, but a linear
trend is a sufficient-statistic computation: the least-squares slope of a
field against time depends on the data only through a handful of additive
running sums. This adds an inference-time aggregator that exploits that to
produce per-grid-cell **trend maps** in a single streaming pass, with memory
independent of the number of timesteps.

The aggregator accumulates, per grid cell, the regression sufficient
statistics (sums of `t`, `t**2`, `y`, `t*y`, and the count `n`) and finalizes
the slope `(n·Σt·y − Σt·Σy) / (n·Σt² − (Σt)²)` only when logs/diagnostics are
requested. It produces predicted and target trend maps, a difference map, and
area-weighted RMSE; slope units are `<variable units> / year`. The additive
sufficient statistics are reduced across data-parallel ranks (`reduce_sum`,
emit on root only), matching the existing `enso_coefficient` / `annual`
distributed patterns. Time is referenced to a fixed epoch and accumulated in
float64; the slope is shift-invariant so the epoch only affects conditioning.

It is configured via `TrendMetricConfig` with a configurable `variables` list
(`None` = all), wired into `InferenceEvaluatorAggregatorConfig` as the `trend`
metric, and is **disabled by default**.

Changes:
- `fme.ace.aggregator.inference.trend` (new): `TrendEvaluatorAggregator` and
  `TrendMetricConfig`, plus the `_years_since_reference` time helper.
- `fme.ace.aggregator.inference.main`: add `trend` to the `MetricConfig`
  union, to `InferenceEvaluatorAggregatorConfig` (default-disabled field), and
  to the enabled-metric list.
- `fme.ace.aggregator.inference.__init__`: export `TrendMetricConfig`.
- Tests: known-slope recovery, streaming-equals-single-batch, distributed
  reduction (`mock_distributed`), variable filtering, and default-disabled /
  enable-on-config wiring.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/trend.py
+++ b/fme/ace/aggregator/inference/trend.py
@@ -18,6 +18,10 @@ from ..plotting import plot_paneled_data
 from .build_context import MetricBuildContext, MetricNotSupportedError, maybe_filter
 from .data import InferenceBatchData, MetricBuildResult, SubAggregator
 
+# A fixed Julian year (365.25 days) is used for the slope's per-year units
+# regardless of the data calendar; this introduces a small (~1.5%) scaling
+# error for non-365.25-day calendars (e.g. CMIP6 360-day), which is fine for
+# a diagnostic trend.
 SECONDS_PER_YEAR = 365.25 * 24 * 60 * 60
 # A fixed reference epoch is used (rather than each sample's start time) so the
 # running sums stay consistent across batches and distributed ranks without
@@ -60,11 +64,11 @@ class TrendEvaluatorAggregator:
     """
 
     _image_captions = {
-        "trend_maps": (
+        "maps": (
             "{name} linear trend; (left) target and (right) generated "
             "[{units} / year]"
         ),
-        "trend_difference_map": (
+        "difference_map": (
             "{name} linear trend difference (generated - target) [{units} / year]"
         ),
     }
@@ -207,18 +211,18 @@ class TrendEvaluatorAggregator:
         for name in gen_trends:
             target_map = target_trends[name]
             gen_map = gen_trends[name]
-            images[f"trend_maps/{name}"] = plot_paneled_data(
+            images[f"maps/{name}"] = plot_paneled_data(
                 [[target_map.cpu().numpy(), gen_map.cpu().numpy()]],
                 diverging=True,
-                caption=self._caption("trend_maps", name),
+                caption=self._caption("maps", name),
             )
             difference = gen_map - target_map
-            images[f"trend_difference_map/{name}"] = plot_paneled_data(
+            images[f"difference_map/{name}"] = plot_paneled_data(
                 [[difference.cpu().numpy()]],
                 diverging=True,
-                caption=self._caption("trend_difference_map", name),
+                caption=self._caption("difference_map", name),
             )
-            metrics[f"rmse/{name}"] = float(
+            metrics[f"weighted_rmse/{name}"] = float(
                 self._ops.area_weighted_rmse(
                     predicted=gen_map.to(torch.float32),
                     truth=target_map.to(torch.float32),
--- a/fme/ace/aggregator/inference/trend.py
+++ b/fme/ace/aggregator/inference/trend.py
@@ -1,0 +1,293 @@
+import dataclasses
+from collections.abc import Mapping
+from typing import Any
+
+import cftime
+import numpy as np
+import torch
+import xarray as xr
+
+from fme.core.dataset.data_typing import VariableMetadata
+from fme.core.device import get_device
+from fme.core.distributed import Distributed
+from fme.core.gridded_ops import GriddedOperations
+from fme.core.typing_ import TensorDict, TensorMapping
+from fme.core.wandb import Image
+
+from ..plotting import plot_paneled_data
+from .build_context import MetricBuildContext, MetricNotSupportedError, maybe_filter
+from .data import InferenceBatchData, MetricBuildResult, SubAggregator
+
+SECONDS_PER_YEAR = 365.25 * 24 * 60 * 60
+# A fixed reference epoch is used (rather than each sample's start time) so the
+# running sums stay consistent across batches and distributed ranks without
+# communication. The regression slope is shift-invariant, so the epoch does not
+# affect the result; it is chosen near the climate-data era only to keep the
+# magnitude of the time regressor small, which improves the floating-point
+# conditioning of the (float64) running sums.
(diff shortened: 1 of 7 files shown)


fme/ace/aggregator/inference/trend.py
```diff
@@ -0,0 +1,310 @@
...
+from fme.core.distributed import Distributed
+from fme.core.gridded_ops import GriddedOperations
+from fme.core.typing_ import TensorDict, TensorMapping
+from fme.core.wandb import Image
+
+from ..plotting import plot_paneled_data
+from .build_context import MetricBuildContext, MetricNotSupportedError, maybe_filter
+from .data import InferenceBatchData, MetricBuildResult, SubAggregator
+
+SECONDS_PER_YEAR = 365.25 * 24 * 60 * 60
```
mcgibbon, line 25: I think this is reasonable, but note the trend "rate" is hard-coded and not specific to the calendar used. Will introduce small errors for e.g. 360-day calendars in CMIP6 but that's alright.

fme/ace/aggregator/inference/trend.py
```diff
@@ -0,0 +1,310 @@
...
+    def get_logs(self, label: str) -> dict[str, float | Image]:
+        target_trends, gen_trends = self._get_trends()
+        if target_trends is None or gen_trends is None:
+            return {}  # only the root process returns logs
+        images: dict[str, Image] = {}
+        metrics: dict[str, float] = {}
+        for name in gen_trends:
+            target_map = target_trends[name]
+            gen_map = gen_trends[name]
+            images[f"trend_maps/{name}"] = plot_paneled_data(
```
mcgibbon: Question: why use "trend_" on these metrics but just "rmse" below? If "trends/" is already going to be prefixed, use plain names for all three.

fme/ace/aggregator/inference/trend.py
```diff
@@ -0,0 +1,310 @@
...
+                diverging=True,
+                caption=self._caption("trend_maps", name),
+            )
+            difference = gen_map - target_map
+            images[f"trend_difference_map/{name}"] = plot_paneled_data(
+                [[difference.cpu().numpy()]],
+                diverging=True,
+                caption=self._caption("trend_difference_map", name),
+            )
+            metrics[f"rmse/{name}"] = float(
```
mcgibbon: nit: name weighted_rmse to be clear this is area-weighted.

fme/ace/aggregator/inference/trend.py
mcgibbon: Added a comment at the constant noting the per-year units use a fixed 365.25-day Julian year regardless of the data calendar, so the ~1.5% scaling offset on 360-day (e.g. CMIP6) calendars is expected/accepted. (257e945b3)

fme/ace/aggregator/inference/trend.py
mcgibbon: Done — dropped the redundant `trend_` prefix. Keys are now `maps/{var}`, `difference_map/{var}`, and `weighted_rmse/{var}`, all under the `trend/` label. (257e945b3)

fme/ace/aggregator/inference/trend.py
mcgibbon: Renamed to `weighted_rmse` to make the area-weighting explicit. (257e945b3)

mcgibbon:
Eval run exercising this aggregator is live: **https://wandb.ai/ai2cm/ace/runs/3rv821wf**

It loads the `j8r0z322` 4°-daily ERA5 checkpoint and runs inference with `max_epochs: 0` + `evaluate_before_training: true` (so evaluation runs at epoch 0 with no training), with `trend` enabled on `[TMP2m, surface_temperature, PRATEsfc, h500]` over the existing 10-year × 8-IC daily inference. The `trend/{maps,difference_map,weighted_rmse}/<var>` panels will populate once the inference window completes.

Beaker: https://beaker.org/ex/01KVXWQ6EVN7890VD15S6ZC2A2 · config on branch `exp/2026-06-24-trend-eval-4deg-daily`.

mcgibbon:
Run finished cleanly (exit 0) and the trend metrics logged end-to-end. The summary has, for each configured variable, `10year/trend/maps/<var>` and `10year/trend/difference_map/<var>` (target-vs-generated panels) plus a scalar `10year/trend/weighted_rmse/<var>`:

| var | weighted_rmse (per year) |
|---|---|
| TMP2m | 0.052 |
| surface_temperature | 0.050 |
| h500 | 1.76 |
| PRATEsfc | 8.9e-7 |

So the aggregator runs correctly in a real (distributed) inference, not just the unit tests. https://wandb.ai/ai2cm/ace/runs/3rv821wf

elynnwu (commented):
I can't remember where we're with spatial parallelism, but this aggregator likely doesn't work with that. But I think many aggregators don't work with spatial parallelism so possibly no action here.

fme/ace/aggregator/inference/trend.py
```diff
@@ -0,0 +1,314 @@
...
+                caption=self._caption("maps", name),
+            )
+            difference = gen_map - target_map
+            images[f"difference_map/{name}"] = plot_paneled_data(
+                [[difference.cpu().numpy()]],
+                diverging=True,
+                caption=self._caption("difference_map", name),
+            )
+            metrics[f"weighted_rmse/{name}"] = float(
+                self._ops.area_weighted_rmse(
```
AnnaKwa, line 226: Does the gridded ops area weighting handle ocean fields that are masked over land? i.e. Maybe @elynnwu or @jpdunc23 is also familiar with how this works.

AnnaKwa (commented):
LGTM, I have a question about the gridded ops for ocean variables.

fme/ace/aggregator/inference/trend.py
elynnwu: Yes it's taken care of `_spatial_mask_area_weights`

mcgibbon:
> I can't remember where we're with spatial parallelism, but this aggregator likely doesn't work with that. But I think many aggregators don't work with spatial parallelism so possibly no action here.

Yeah I would suggest we do nothing here. Ideally when we do implement that it will be in a way that kind of works seamlessly, like having a `dist.plot_imshow` method or something.

AnnaKwa (approved):

PR #1312, opened 2026-06-24
Add explicit review passes to the pre-review skill

mcgibbon:
Adds the standing "required audit passes" list to the `pre-review` skill.

Changes:
- `.claude/skills/pre-review/SKILL.md` — add the required-audit-passes list; add the Builder pattern pass (§9) with the rules, exceptions, and pre-existing-violation carve-out.

- [ ] Tests added (n/a — docs/process only)
- [ ] If dependencies changed, "deps only" image rebuilt (n/a)

### DIFF
--- a/.claude/skills/pre-review/SKILL.md
+++ b/.claude/skills/pre-review/SKILL.md
@@ -24,3 +24,91 @@ Communicate with the user about whether:
  - The PR is in good shape, perhaps you have some comments to make on the PR or updates to its title/description but the code is ready for the reviewer.
 
 When leaving comments on a PR, always start your comment with "Pre-review agent: ".
+
+## Required audit passes
+
+Before reporting the PR ready, walk the diff and explicitly perform each of these passes. Report findings even if none — silence on a pass means it wasn't done. Address findings inline where small; surface them as comments or to the user where they need a decision.
+
+### 1. Silent-failure pass
+
+For every new branch, default, fallback, optional field, union member, or aggregation introduced by this PR, name what happens when its input is missing, empty, `None`, `0.0`, a default-valued weight, or a key not present in the prior schema. If the answer is "the code takes the default branch quietly," flag it. Specifically check:
+ - New dacite-loaded unions: can every previously-valid YAML still parse and produce equivalent behavior? Are deprecation warnings present where a config form is being phased out?
+ - New aggregations / weighted sums: behavior at empty input, mismatched keys across calls, zero weight, all-zero weights.
+ - "Unsupported" / "not applicable" code paths: do they raise loudly when the user explicitly asked for the unsupported behavior, or only when it was a silent default?
+
+The user's reviewers catch this class of issue more reliably than the author's AI agent does. Treat it as the highest-priority pass.
+
+### 2. Behavior-change pass on existing code paths
+
+For every function, method, or code path that was *modified* (not newly added) by this PR, decide whether the change alters what existing callers observe. AI-generated edits regularly drift the semantics of code they "clean up" — particularly:
+ - Reduction changes (sum ↔ mean, mean ↔ weighted mean, mean over channels vs samples).
+ - Default values of existing arguments.
+ - Which inputs cause a raise vs. a return vs. a log.
+ - Fields silently dropped from intermediate containers (`BatchData`-like objects, dicts, dataclasses) that downstream code relied on.
+ - Log lines removed, renamed, or moved (downstream dashboards/parsers may depend on the string).
+ - Return types narrowed or widened in ways that change downstream branching.
+
+For each behavior change found: if intended, it should be stated in the PR description (and a docstring updated if the public contract shifted); if unintended, revert it. If unclear, seek clarification from the author. Do not let "incidental cleanup" through.
+
+Keep an eye out also for unintended effects of intended behavior changes, especially if the change is in a widely-used utility function or a core code path.
+
+### 3. Differential-test pass
+
+For every test added or modified to cover new behavior, confirm the assertion would actually fail if the new code path were no-oped out. Flag any test whose inputs make masked/unmasked, weighted/unweighted, filtered/unfiltered, or seeded/unseeded outputs numerically identical (e.g., all-ones inputs for a masking test, identical values across channels for a per-channel test, single-element collections for a multi-element behavior). Strengthen inputs in place where the fix is small.
+
+### 4. Sibling / mirror coverage pass
+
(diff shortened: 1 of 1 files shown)


.claude/skills/pre-review/SKILL.md
```diff
@@ -24,3 +24,91 @@ Communicate with the user about whether:
  - The PR is in good shape, perhaps you have some comments to make on the PR or updates to its title/description but the code is ready for the reviewer.
 
 When leaving comments on a PR, always start your comment with "Pre-review agent: ".
+
+## Required audit passes
```
mcgibbon, line 28: I've been using most of these for a while, they're originally drafted after an audit of reviews I got of PRs I had made largely with Claude. Since using this to pre-review PRs, I've gotten far fewer comments/change requests on PRs.

.claude/skills/pre-review/SKILL.md
```diff
@@ -24,3 +24,98 @@ Communicate with the user about whether:
...
+
+Quick scan for patterns that AI-generated code drops in and human reviewers consistently flag:
+ - `Any` return types where a concrete type is available.
+ - `isinstance` + `type: ignore` pairs (prefer refactoring the type).
+ - Imports inside test function bodies (lift to module top).
+ - Abbreviations in identifiers (`cb` → `callback`).
+ - Stale log lines or comments left over from a refactor.
+ - Unnecessary removal of comments or docstrings unrelated to the change.
+
+### 9. Builder pattern pass
```
mcgibbon, line 92: I've been thinking about this set of rules for a long while, but have only just added it to my pre-review skill. I would really like to go through our repo and update our existing code to follow these two rules - they're ones that I've frequently seen make the code harder to edit, understand, and use, and whenever I refactor the code to follow these rules it's much easier to test, re-use, refactor, and understand.

Especially if these can be enforced by an AI review pass, i.e. not even take additional context for PR writers and still leave the possibility for a user to override the findings, I think this is an easy win. I'd mainly like this merged to simplify writing future PRs that refactor our code to follow these guidelines.

I'll be making some example refactor PRs to get reviewed/merged as "stress tests" for the rules, before we consider merging this. I could split this section to another PR if for whatever reason we want the other rules/audit passes to get merged faster.

mcgibbon:
Replying to [comment](https://github.com/ai2cm/ace/pull/1327#issuecomment-4862709046) from @frodre , which is mainly about the builder pattern pass on this PR. Moving the discussion here since it implies changes in this PR, not 1327.

> I'm mainly coming at this from whether or not we need to have a rule and larger refactor around this style of usage. Frozen configurations seems like a genuine win for bad use prevention. As I said, I'm not convinced we need a repo wide rule around this since I don't think there's been any egregious violation of using config features.

FWIW I already was not thinking of this pre-review pass as a "rule" we need to enforce among ourselves, mainly I was thinking it would be helpful information to get a failure marker from a pre-review agent and have to think about it. I think we should assess this guideline in the context of each piece of code it would touch, and use that experience to see how closely we should follow the guideline.

In part based on your feedback, I think what we should actually do is weaken this from a "rule" pass to a "check if it would improve the code" pass for that part. As in, downgrade the part about not passing the Config directly from a rule to a suggestion. I can state that an exception is carved for a Config being passed to the one implementation it builds when the overhead of having a Params layer isn't worth the benefit, but that the pre-reviewer should assess whether adding the Params layer would improve code quality and suggest it if so.

My intuition is that in each case we look at, there will be similar benefits, and that in each case it will turn out we should apply this pattern.

> If we have an essential/clear win use case, then I'd be more onboard with this benefit.

I'd say my main motivation here is "whenever I look at applying this rule it makes a lot of sense and improves the code in ways I didn't expect going into it", the "use cases" below are somewhat weaker motivation.

In the case of the [Trainer refactor PR](https://github.com/ai2cm/ace/pull/1333), I really liked that passing a Params instead of a Config meant that the generic code can rely on a basic frozen dataclass, instead of asking for a TrainConfigProtocol that has multiple implementations.

When we went from having SingleModuleStepper to having Stepper and the SingleModuleStep, and refactored SingleModuleStepperConfig to StepperConfig, that refactor would have been much easier if the stepper implementations took in a Params object instead of the explicit config class. That would mean each config could be written to produce the Params during the interim, instead of having to translate between the two configs.

I don't know if this will happen, but one use case that's been on the back of my mind is being able to more easily support "versions" of our configuration, like having a very basic Config class with few options that relies on defaults, or being able to run training on published configuration files because we maintain each Config class and it builds our infrastructure through the Params. It's more difficult to do that if we have to translate Config classes between each other and reason about how the config-translation interacts with the `.build` method itself in the final config.

> I don't know that we should make a firm rule about how the config attributes/properties/methods are being used as long as it's done in an easily traceable, understandable, and consistent way.

Yeah I think this wasn't very clear, but I would rather this be a firm (and well-defined, not vague) rule for a pre-review agent to flag, but not something we are bound to as code authors.

frodre:
> In part based on your feedback, I think what we should actually do is weaken this from a "rule" pass to a "check if it would improve the code" pass for that part. 

Thanks for all the response with good clarification and examples.  I'm onboard with adding guidance but still being judicious about whether or not the change makes sense in different situations.  I totally agree that having less nebulous/complicated objects to create, especially for tests is a clear win.  And the potential use case sounds interesting.  I was worried this was going to be a new paradigm change everything sort of update, but adding in this style guidelines for agents gives us a chance to feel it out as we vibe to the next phase.  I'll take a look at the actual skill adjustments after lunch.

.claude/skills/pre-review/SKILL.md
```diff
@@ -24,3 +24,99 @@ Communicate with the user about whether:
...
+
+Exceptions:
+ - A config may be passed to the one implementation class its `build()` constructs, and that impl may read the config's fields. If the config is a **leaf**, i.e. none of its fields are themselves dataclasses, your work is done on this point. `list[Config]`, `dict[str, Config]`, and `Config | None` all make a config non-leaf. If the config is non-leaf, consider whether the code is improved by packing the scalars the impl needs into a separate, non-dacite-loaded leaf dataclass passed as an argument. Would this better separate concerns and make the impl easier to test? If so, recommend doing so, perhaps strongly if it's a significant improvement, but it's not a rule-violation. Is the configuration clearly an extension of the class itself, and adding the params doesn't make the code clearer and just adds boilerplate? Then it's fine to pass it to the impl.
```
mcgibbon, line 100: Designed this carve-out based on discussion, @frodre lmk what you think.

frodre (approved):
LGTM.  I'll switch over to using this as I've mostly just been using the /review as my pre-review step.  Having it analyze your previous reviews (since you have a good # of data points) was a good idea!

PR #1313, opened 2026-06-24
Refactor corrector configs to the Builder pattern

mcgibbon:
This PR refactors the three correctors so the corrector no longer reads config fields at all. Each correction step (force-positive, dry-air conservation, moisture/energy budget closure, sea-ice-fraction, surface-energy-flux, ocean-heat-content, ice budget) becomes a small callable `Correction` object that bundles its own scalar params plus the operators it needs (e.g. an area-weighted-mean operator, a vertical coordinate) and applies one step. Each `*CorrectorConfig._build()` packages the configured params into an ordered list of these objects; the corrector is a thin `CorrectionSequence` that simply applies them in order. So `yaml` answers "how is this configurable" (the config dataclasses) while the `Correction` objects answer "what operation does the corrector apply".

The YAML config dataclasses are **unchanged** — `_build` only repackages already-parsed config params into the function-objects. Behavior-preserving (all corrector unit tests pass unchanged in intent). Intended as the template-setting sibling for the rest of the burn-down.

Changes:
- `fme.core.corrector.registry`: add the `Correction` protocol (the `(input, gen, forcing, state) -> (gen, state)` step signature) and a `CorrectionSequence` corrector base whose `__call__` applies an ordered list of `Correction`s.
- `fme.core.corrector.utils`: add the `ForcePositive` correction object (wraps the existing `force_positive`).
- `fme.core.corrector.atmosphere`: add `ConserveDryAir`, `ZeroGlobalMeanMoistureAdvection`, `MoistureBudgetCorrection`, `TotalEnergyBudgetCorrection`; `AtmosphereCorrector` is now a thin `CorrectionSequence`; `_build` assembles the ordered list.
- `fme.core.corrector.ocean`: add `SeaIceFractionCorrection` (wraps `SeaIceFractionConfig`), `SurfaceEnergyFluxCorrection`, `OceanHeatContentCorrection`; `OceanCorrector` is now a thin `CorrectionSequence`; `_build` assembles the ordered list.
- `fme.core.corrector.ice`: add `IceBudgetCorrection` (wraps `IceBudgetCorrectionConfig`); `IceCorrector` is now a thin `CorrectionSequence`.
- `test_atmosphere`, `test_ocean`: route direct impl construction through `config._build(...)` so the tests exercise the real build path.

- [x] Behavior-preserving refactor; existing 50 corrector tests pass unchanged (no new tests added)

### DIFF
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -1,6 +1,6 @@
 import dataclasses
 import datetime
-from collections.abc import Mapping
+from collections.abc import Callable, Mapping
 from typing import Any, Literal, Protocol
 
 import torch
@@ -62,6 +62,12 @@ class SeaIceFractionConfig:
         for name in self.zero_where_ice_free_names:
             out[name] = gen_data[name] * (out[self.sea_ice_fraction_name] > 0.0)
         return out
+
+
+# Operation the OceanCorrector applies to enforce sea-ice-fraction constraints;
+# provided by SeaIceFractionConfig.__call__. Passed as a callable (not the config)
+# so the corrector relies only on the operation, not on reading config fields.
+SeaIceFractionCorrection = Callable[[TensorMapping, TensorMapping], TensorDict]
 
 
 @dataclasses.dataclass
@@ -191,7 +197,11 @@ class OceanCorrectorConfig(CorrectorConfigABC):
         )
         return OceanCorrector(
             params=params,
-            sea_ice_fraction_correction=self.sea_ice_fraction_correction,
+            sea_ice_fraction_correction=(
+                None
+                if self.sea_ice_fraction_correction is None
+                else self.sea_ice_fraction_correction.__call__
+            ),
             gridded_operations=gridded_operations,
             vertical_coordinate=vertical_coordinate,
             timestep=timestep,
@@ -202,7 +212,7 @@ class OceanCorrector(CorrectorABC):
     def __init__(
         self,
         params: OceanCorrectorParams,
-        sea_ice_fraction_correction: SeaIceFractionConfig | None,
+        sea_ice_fraction_correction: SeaIceFractionCorrection | None,
         gridded_operations: GriddedOperations,
         vertical_coordinate: HasOceanDepthIntegral | None,
         timestep: datetime.timedelta,
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -1,6 +1,6 @@
 import dataclasses
 import datetime
-from collections.abc import Callable, Mapping
+from collections.abc import Mapping
 from typing import Any, Literal, Protocol
 
 import torch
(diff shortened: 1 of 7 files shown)


fme/core/corrector/ocean.py
```diff
@@ -145,23 +161,54 @@ def _get_corrector(
...
+            timestep=timestep,
+        )
+
 
 class OceanCorrector(CorrectorABC):
     def __init__(
         self,
-        config: OceanCorrectorConfig,
+        params: OceanCorrectorParams,
+        sea_ice_fraction_correction: SeaIceFractionConfig | None,
```
mcgibbon: This kind of begs the question of why the other corrections don't also get passed by having Callables passed as argument instead of having the code in the corrector and the options on params, but that refactor is out-of-scope for this PR.

jpdunc23: I recall there being some emphasis in the past on having the config classes contain their associated ops as methods, but agreed it would be good to have this be more consistent.

jpdunc23: Would it be better to pass `SeaIceFractionConfig.__call__` via corrector params?

jpdunc23 (commented):
I think it is a good idea to avoid the many opaque / potentially brittle `self._config.*` reads in the corrector.

Disallowing "a non-leaf config reaching its implementation class" is probably a good rule-of-thumb, but IMO this PR reshuffles things without improving the situation: We're simply repeating all of the leaf corrector config attributes on new non-config dataclass like `*CorrectorParams`, and this doesn't change what I see as the more fundamental problem here of having the `*Corrector` implementation know too much about the leaf configs.

I think a better approach is to follow the `SeaIceFractionConfig` pattern where the leaf config class has a method (`__call__` or something else) that accesses its own config attributes so that the corrector implementation doesn't have to worry about those low-level details. An object like `*CorrectorParams` (perhaps `*CorrectorCalls` is a better name for what I'm proposing) can then simply be a collection of these methods (including, perhaps, "no-op" methods for cases where the leaf config is null), and all the `*Corrector` needs to know is in what order to call the `*CorrectorCalls` items.

fme/core/corrector/ocean.py
mcgibbon: I think it would be better to pass the `.__call__` than the config, it did confuse me to see a Config being passed and later realize it's a Callable.

I'm not sure about passing it on the params though - params feels like it should be a container of primitive could-be-yaml-configurable types (even if the yaml is massaged into this type). Without that boundary, you have to also ask whether the gridded_operations should be on the params, for example. It's kind of nice for the explicit arguments to describe all of the custom types/complex objects/callables this type relies on, and have `params` sweep under the rug all the types that don't have code related to it, and are just configurables. That way yaml answers "how is this object configurable" while the other arguments answer "what operations/operators and complex data types does this code rely on"

mcgibbon: Maybe we can refactor to pass the Callable, and make a follow-on PR right now based from this branch to refactor the other corrections to behave more like the SeaIceFractionConfig and also pass Callables.

jpdunc23 (approved):
LGTM, very clean. 

One note from agent review:

> `IceCorrector` now copies `gen_data` (`dict(gen_data)`) before invoking `budget_correction`, whereas the old code passed the raw mapping. This only matters if `IceBudgetCorrectionConfig.__call__` mutated its input in place — which it doesn't (it returns a fresh dict). The change brings ice in line with how atmosphere/ocean already behaved, so it's a harmless consistency improvement, not a regression. Not worth changing.

Good change, just worth noting.

PR #1314, opened 2026-06-24
Move data_writer sub-writer construction into the config builders

mcgibbon:
In `fme/ace/inference/data_writer/`, the config objects and the runtime writer classes were doing parts of each other's jobs. `DataWriterConfig.build` handed `DataWriter` its raw sub-configs, and `DataWriter.__init__` then built its own child writers from them. Separately, the inference config classes reached directly into `data_writer`'s internal fields to validate time coarsening.

This PR moves writer construction into the config builders and gives the validation a real method. The writer classes now just receive a finished list of writers, and the config that defines the YAML schema stays separate from the runtime implementation. The change is behavior-preserving.

Changes:
- `fme.ace.inference.data_writer.main.DataWriterConfig`: `build` / `build_paired` now build the full list of sub-writers themselves (the prediction writer, the monthly writer, the per-file writers, and the time-coarsening wrapper) and pass that list in. `DataWriter` / `PairedDataWriter.__init__` take a ready-made `writers` list instead of raw sub-configs plus enable flags, so they no longer translate config into writers in their own `__init__`.
- `fme.ace.inference.data_writer.main.DataWriterConfig.validate_time_coarsen` (and a small `FileWriterConfig.validate_time_coarsen` helper it calls): a single method that validates the top-level and per-file time coarsening against the inference schedule.
- `fme.ace.inference.inference.InferenceConfig` / `fme.ace.inference.evaluator.InferenceEvaluatorConfig`: `__post_init__` now calls `data_writer.validate_time_coarsen(...)` instead of reaching into `data_writer.time_coarsen` and iterating `data_writer.files`.
- Tests: build writers through `DataWriterConfig.build` / `build_paired` rather than the old writer constructor signatures.

Note on HEALPix: for HEALPix (`face`) coordinates the non-paired `DataWriter` still ends up with an empty list of writers, exactly as before. The only difference is that `path` and metadata are now always stored, which only affects the HEALPix `DataWriter.write` path. That path is not used in production (both inference entrypoints use `build_paired`).

Follow-up: #1327 takes the next step of giving `FileWriter` a dedicated runtime-params object instead of its config. That part is split out so it can be discussed on its own. The similar validation on the coupled ocean/atmosphere config is also left for later; this PR keeps `time_coarsen` a public field so nothing there breaks.

- [x] Tests added (existing data_writer tests migrated to the public builder API)

### DIFF
--- a/fme/ace/inference/data_writer/file_writer.py
+++ b/fme/ace/inference/data_writer/file_writer.py
@@ -269,6 +269,11 @@ class FileWriterConfig:
             for base_filename in base_filenames
         ]
 
+    def validate_time_coarsen(self, forward_steps_in_memory: int, n_forward_steps: int):
+        """Validate this writer's time coarsening against the inference schedule."""
+        if self.time_coarsen is not None:
+            self.time_coarsen.validate(forward_steps_in_memory, n_forward_steps)
+
     def build_paired(
         self,
         experiment_dir: str,
--- a/fme/ace/inference/data_writer/main.py
+++ b/fme/ace/inference/data_writer/main.py
@@ -75,6 +75,13 @@ class DataWriterConfig:
             filenames.extend(file.filenames)
         return filenames
 
+    def validate_time_coarsen(self, forward_steps_in_memory: int, n_forward_steps: int):
+        """Validate time coarsening (top-level and per-file) against the schedule."""
+        if self.time_coarsen is not None:
+            self.time_coarsen.validate(forward_steps_in_memory, n_forward_steps)
+        for file_config in self.files or []:
+            file_config.validate_time_coarsen(forward_steps_in_memory, n_forward_steps)
+
     def build_paired(
         self,
         experiment_dir: str,
@@ -85,19 +92,50 @@ class DataWriterConfig:
         coords: Mapping[str, np.ndarray],
         dataset_metadata: DatasetMetadata,
     ) -> "PairedDataWriter":
+        writers: list[PairedSubwriter] = []
+        if self.save_prediction_files:
+            raw_writer: PairedSubwriter = PairedRawDataWriter(
+                path=experiment_dir,
+                initial_condition_times=initial_condition_times,
+                save_names=self.names,
+                variable_metadata=variable_metadata,
+                coords=coords,
+                dataset_metadata=dataset_metadata,
+            )
+            if self.time_coarsen is not None:
+                raw_writer = self.time_coarsen.build_paired(raw_writer)
+            writers.append(raw_writer)
+        if self.save_monthly_files:
+            writers.append(
+                PairedMonthlyDataWriter(
+                    path=experiment_dir,
+                    initial_condition_times=initial_condition_times,
+                    n_timesteps=n_timesteps,
+                    timestep=timestep,
+                    save_names=self.names,
+                    variable_metadata=variable_metadata,
+                    coords=coords,
+                    dataset_metadata=dataset_metadata,
+                )
+            )
+        for writer_config in self.files or []:
+            writers.append(
+                writer_config.build_paired(
+                    experiment_dir=experiment_dir,
+                    initial_condition_times=initial_condition_times,
+                    n_timesteps=n_timesteps,
+                    timestep=timestep,
+                    variable_metadata=variable_metadata,
+                    coords=coords,
+                    dataset_metadata=dataset_metadata,
+                )
+            )
         return PairedDataWriter(
+            writers=writers,
             path=experiment_dir,
-            initial_condition_times=initial_condition_times,
-            n_timesteps=n_timesteps,
-            timestep=timestep,
             variable_metadata=variable_metadata,
             coords=coords,
-            enable_prediction_netcdfs=self.save_prediction_files,
-            enable_monthly_netcdfs=self.save_monthly_files,
-            save_names=self.names,
-            time_coarsen=self.time_coarsen,
             dataset_metadata=dataset_metadata,
-            files=self.files,
         )
 
     def build(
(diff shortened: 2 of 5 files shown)


fme/ace/inference/data_writer/file_writer.py
```diff
@@ -175,6 +175,33 @@ def _time_subselector(sample_ds: xr.Dataset) -> xr.Dataset:
     return combined_data
 
 
+@dataclasses.dataclass
+class FileWriterParams:
```
frodre: This should be a private class, right?

frodre: I'm not convinced we need a new parameter handoff class to tighten usage declaration over just using the configuration in the build.  While builds might have access to other fields they don't need, I don't necessarily think that's a large issue, given it's tightly coupled to code in this module.  We could explore whether or not we need to restructure the top-level configuration and sub-configuration types.

fme/ace/inference/data_writer/file_writer.py
```diff
@@ -269,6 +296,21 @@ def filenames(self) -> list[str]:
...
+    def validate_time_coarsen(self, forward_steps_in_memory: int, n_forward_steps: int):
+        """Validate this writer's time coarsening against the inference schedule."""
+        if self.time_coarsen is not None:
+            self.time_coarsen.validate(forward_steps_in_memory, n_forward_steps)
+
+    def _build_writer_params(self) -> FileWriterParams:
+        return FileWriterParams(
+            label=self.label,
+            names=self.names,
+            subselect_horizontal=bool(self.lat_extent or self.lon_extent),
```
frodre: We could have `subselect_horizontal` be a property and it would achieve the same thing, I think.

fme/ace/inference/data_writer/main.py
```diff
@@ -85,19 +92,50 @@ def build_paired(
         coords: Mapping[str, np.ndarray],
         dataset_metadata: DatasetMetadata,
     ) -> "PairedDataWriter":
+        writers: list[PairedSubwriter] = []
+        if self.save_prediction_files:
+            raw_writer: PairedSubwriter = PairedRawDataWriter(
```
frodre, line 97: There's a lot of duplication in these build calls, but that wasn't introduced by this PR.  Just noticing.

fme/ace/inference/data_writer/main.py
```diff
@@ -110,109 +148,80 @@ def build(
...
-            files=self.files,
         )
 
 
 class PairedDataWriter(WriterABC[PrognosticState, PairedData]):
     def __init__(
         self,
+        writers: list[PairedSubwriter],
         path: str,
-        initial_condition_times: npt.NDArray[cftime.datetime],
```
frodre, line 133: This was a nice cleanup!

fme/ace/inference/evaluator.py
```diff
@@ -241,18 +241,10 @@ class InferenceEvaluatorConfig:
...
-                self.n_forward_steps,
-            )
-        if self.data_writer.files is not None:
-            for file_config in self.data_writer.files:
-                if file_config.time_coarsen is not None:
-                    file_config.time_coarsen.validate(
-                        self.forward_steps_in_memory,
-                        self.n_forward_steps,
-                    )
+        self.data_writer.validate_time_coarsen(
```
frodre, line 244: Another nice cleanup from this refactor.

frodre (commented):
Thanks @mcgibbon.  This is a nice consistency update that cleans up a fair amount of usage.  Overall, I think the consolidation into builds is worthwhile.  I'm not convinced of the need for a `FileWriterParams` over the configuration usage.  I think there are simpler fixes to the issue of the runtime `subselect_horizontal` and I don't know if I agree with 

> dataclass attributes configured from yaml with dacite are considered private, and can only be used in methods on that dataclass

If a built component stores a configuration instead of locally re-storing all the information that could be useful, it should be able to look at that configuration.  I do think the items that are publicly available from configurations should be shallow/imutable features of the config, maybe?

Other than that, I did find the agent generated PR description here used  language or artifacts I'm not familiar with (leaf, builder audit, rulesets etc).  If it's based on a document it would be helpful if you had the agent link to descriptions of the task with source of truth for definitions included.

fme/ace/inference/data_writer/file_writer.py
```diff
@@ -175,6 +175,33 @@ def _time_subselector(sample_ds: xr.Dataset) -> xr.Dataset:
     return combined_data
 
 
+@dataclasses.dataclass
+class FileWriterParams:
+    """Plain runtime parameters for :class:`FileWriter`.
```
mcgibbon: Issue: we don't use these sphinx-style notations `:class:` and `:meth:` in our docs.

mcgibbon:
> If a built component stores a configuration instead of locally re-storing all the information that could be useful, it should be able to look at that configuration. I do think the items that are publicly available from configurations should be shallow/imutable features of the config, maybe?

Going into this, I'll say I think I should separate this part of the rule that leads to Params classes into its own PR so we can treat the question separately. But I'll argue why this change makes sense. Also I like the idea of saying these params classes should be frozen/immutable.

----- 

Traditionally the view I've held is that the config class and the class it's building are really just one class, and should be tightly coupled. The way this PR goes decouples them, so that the built class no longer depends on the config class, avoiding a circular dependency.

Having seen the change for this specific case, I do think it's better. `FileWriter` depends on `subselect_horizontal=bool(self.lat_extent or self.lon_extent)`, and does not directly use `lat_extent` or `lon_extent`. When the class takes in the full config, you can't tell that's the case. Generally speaking the pattern of "use config in the course of building the objects that go into the class init" is not uncommon.

It simplifies the init logic in cases where inits currently translate config values into internally-stored values, moving that to the builder layer instead. It avoids this kind of default pattern we fall into where we always pass the Config to the init, without thinking about what the object's actual API is.

Having the config build exactly the frozen Params the class needs and pass it to init is simple. The class no longer does configuration-to-attribute parsing in its init, and we know the Params won't be mutated. And we no longer have a circular dependency - it gives us the freedom to define multiple configs for the same object if we want, or to move our configs to different files if we want.

An issue is when passing Config, it's difficult to cleanly define what behavior is or isn't allowed. It's easy to say you can only pass a single-level frozen dataclass of basic/configurable types. It's harder to say you're allowed to access the basic types on the config but not any of its methods except for property methods - there are cases where we use `config.__call__()` for example. And should we allow property methods in that case? Or should the object's init handle that conversion? Do we allow methods to be called but not builder methods? What constitutes a builder method?

By contrast the choice here makes it unambiguous how the code should be written, which seems to give better AI results.

Now, one thing I'll say is I would only suggest this kind of rule or refactor because we're using AI-based coding now. This way of writing it is significantly safer and more explicit, making it harder to write bad code. I expect it to make AI-based code easier to review, and harder for it to hide bad patterns. It also requires more boilerplate, enough that I wouldn't typically ask people to write these Params classes, but I'm not worried about AI writing these.

In the long run this is also kind of interesting because it opens up the path to have multiple ways to configure and build our objects. We could define a heavily simplified config as an alternative way to build the same object hierarchy as our full configs. This is only possible if the config and Params are decoupled.

mcgibbon:
Re the terminology, I'll get Claude to clean up the description to be less jargony, but it's referring to this PR, particularly the builder pattern audit pass: https://github.com/ai2cm/ace/pull/1312/changes#diff-369e44873a784cb13ea8f4757a6aec8dfe080798bcc7d50029ebc0ac1da26e9cR92

mcgibbon:
Per the discussion above, I've split the `FileWriterParams` part out into its own follow-on PR (#1327, based on this branch) so the question of whether those runtime-params objects are worth it can be treated separately. This PR is now just the writer-construction move and the `validate_time_coarsen` cleanup. I also dropped the sphinx-style `:class:`/`:meth:` notation in the docstrings I touched.

frodre (approved):
LGTM, Thanks for splitting off the parameter question

PR #1316, opened 2026-06-24
Add `ZonallyPeriodicBilinearUpsample` for Samudra

elynnwu:
The Samudra decoder upsamples with `torch.nn.Upsample(mode="bilinear")`, which interpolates the longitude (width) boundary against a replicated edge column and leaves a discontinuity at the lon=0 seam on the zonally-periodic ocean grid. This PR ports `ZonallyPeriodicBilinearUpsample` from [m2lines/Samudra](https://github.com/m2lines/Samudra/blob/ab554631973ced3c567c1ef65ef2f84c222458d7/src/samudra/models/modules/blocks.py), which circularly pads one column on each longitude edge before interpolating and crops the padding back off so the seam is interpolated against its true periodic neighbor. It is exposed via a new `zonally_periodic_upsample` flag that defaults to `False`, preserving the behavior of existing checkpoints.

Changes:
- `fme.ace.models.ocean.m2lines.layers.ZonallyPeriodicBilinearUpsample`: new upsampling module enforcing longitude periodicity; latitude axis left unpadded, output shape matches `BilinearUpsample`
- `fme.ace.models.ocean.m2lines.samudra.Samudra`: new `zonally_periodic_upsample` arg (default `False`); selects the upsample class for the decoder and extends the shape-cropping `isinstance` check to cover both upsample types
- `fme.ace.registry.m2lines.SamudraBuilder`: plumb `zonally_periodic_upsample` config field through to the model

- [x] Tests added


elynnwu:
Turning this on just in inference gives a more smooth seam at `lon=0`:
image

elynnwu:
Without any re-training and simply enables it during eval, it cuts the largest SST bias by ~half:
image

fme/ace/models/ocean/m2lines/test_samudra.py
```diff
@@ -11,6 +15,65 @@
 from fme.ace.models.ocean.m2lines.samudra import Samudra
 
 
+def test_zonally_periodic_upsample_matches_bilinear_shape():
```
jpdunc23: Maybe parametrize this test to check a few different shapes? In particular, including 180x360 would be good.

jpdunc23 (approved):
LGTM, just one optional suggestion to expand one of the tests.

PR #1317, opened 2026-06-25
Add `optimize_single_component_per_batch` flag to `CoupledTrainStepperConfig`

jpdunc23:
Adds an opt-in `optimize_single_component_per_batch` flag to `CoupledTrainStepperConfig`. When enabled, each `train_on_batch` call optimizes exactly one of {ocean, atmosphere}: a per-batch component choice owned by `CoupledStepperTrainLoss` (`component_optimization_choice()`) selects one realm (fair 50/50 among realms whose loss is non-null for that batch, the sole non-null realm otherwise), and the non-selected realm contributes zero loss and runs under `torch.no_grad()`. The selected realm keeps its existing `n_steps` / `optimize_last_step_only` / `loss_weight` schedule — the choice only decides eligibility. The choice is re-made once per batch right after `sample_n_steps()`. It stays active during validation too (drawn from a separate, deterministically-seeded eval RNG), so the validation loss is single-component-per-batch and stays comparable in scale to the training batch loss; per-step diagnostic metrics for both realms are still produced. Default off reproduces current behavior exactly, and training-config changes are not persisted in inference checkpoints, so checkpoint loading is unaffected.

Changes:
- `CoupledStepperTrainLoss.__call__` now routes its optimized-step check through the
  `step_is_optimized(realm, step)` wrapper (previously called the per-realm schedule
  directly); `compute_loss` deliberately still calls the schedule directly to keep
  per-step diagnostic metrics for the non-selected realm (in both train and eval)
- `CoupledStepperTrainLoss` gains `component_optimization_choice()` plus selection state:
  a train RNG `_component_choice_rng` (lazily seeded via `_init_component_choice_rng()` from
  the distributed seed with a unique offset) and an eval RNG `_eval_component_choice_rng`
  (reseeded inside the loss's `seed_rng()`, renamed from `seed_step_sampler` and driven by
  the validation loop's `seed_eval(0)`); the `step_is_optimized` short-circuit applies
  in both train and eval (mirrors `ComponentLossSchedule`'s train/eval n_steps sampler split)
- `CoupledStepperTrainLoss.__init__` gains an `optimize_single_component_per_batch` parameter
- `CoupledTrainStepperConfig` gains `optimize_single_component_per_batch: bool = False`,
  plumbed through `_build_loss` and validated in `__post_init__`: with the flag on, a
  statically-null realm (`loss_weight == 0.0`, or an `n_steps` schedule whose max yields no
  optimized step) raises `ValueError`, since the choice would have no observable effect

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+        ocean_loss: StepLoss,
+        atmosphere_loss: StepLoss,
+        ocean_schedule: ComponentLossSchedule,
+        atmosphere_schedule: ComponentLossSchedule,
+        single_component_optim: bool = False,  # NEW — enable the per-batch coin
+    ):
+        ...
+        # NEW coin state:
+        self._single_component_optim: bool = single_component_optim
+        self._coin_rng: np.random.RandomState | None = None  # lazy, like TimeLengthProbabilities
```
jpdunc23: Can we maybe just call this `_single_component_optim_rng`? The "coin" terminology is somewhat random.

I also think we could probably have a separate, more general object to own the RNG, perhaps defined in `fme.core` and reused by `TimeLengthProbabilities`; otherwise, we'll just be repeating a lot of the `TimeLengthProbabilities` API here. Something like:

```python
class RandomChoice:
    def __init__(
        self, 
        outcomes,  # 1-D array-like or int, same as np.random.RandomState.choice's a argument
        probabilities, # 1-D array-like, same as np.random.RandomState.choice's p argument
    ):
        ...

    def initialize_rng(self):
        ...  # same as TimeLengthProbabilities implementation

    def seed_rng(self):
        ...  # same as TimeLengthProbabilities implementation

    def sample(self) -> int:
        ...  # same as TimeLengthProbabilities implementation
```

I think it makes sense to include that change as part of the PR since this is maybe the first time we need rng similar to what is in `TimeLengthProbabilities`, or we could add `RandomChoice` and the `TimeLengthProbabilities` refactor in a separate PR first.

@mcgibbon please let me know what you think.

jpdunc23:
@mcgibbon requesting your review of the plan prior to implementation. You can view the plan with nice markdown rendering here: https://github.com/ai2cm/ace/blob/a0a4f16a6ddc74fb9990dc47492b4dd8bccf1859/pr-plan.md

pr-plan.md
mcgibbon: Re _coin_rng yes, it implies choice from 2 values and it's possible we'd add more components in the future. Maybe `_which_component_rng`? Or even just `_rng`, does it use another one?

I would advise against a RandomChoice class. I don't get what benefit it provides that isn't already provided by `np.random.RandomState`, all it seems to do is restrict the class to only having one of its sampling methods and moving the "choice" verb from the method to the class name.

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+---
+
+## Open Questions
+
+- When `use_single_component_optim=True` but one component is statically null (zero
+  `loss_weight`, or `n_steps` whose max is 0), the coin degrades to always selecting the
+  sole non-null realm — i.e. the flag has no observable effect. Should `__post_init__`
+  (a) raise a `ValueError` (the flag is meaningless here), (b) emit a warning and proceed,
```
mcgibbon: It should raise a ValueError.

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+  or (c) silently allow it as graceful degradation? The plan currently assumes (c); (b)
+  seems most useful for catching misconfigured experiments. Note this is distinct from the
+  always-rejected case where *both* realms are null.
+- Flag name: `use_single_component_optim` (proposed) vs `optimize_single_component_per_batch`?
```
mcgibbon: `optimize_single_component_per_batch` is much clearer, I prefer it. It's ambiguous what "single component optim" would be, e.g. arguably FTO and FTA are single component optimization. This resolves that issue.

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+  Method name: `flip_optimization_coin()` vs an alternative. Soft proposals only — `Config`
```
mcgibbon: Use a different name for this method.

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+  suffix and `_`-private conventions hold either way.
+- Coin policy is fixed at fair 50/50 among non-null realms; loss-weight-proportional
```
mcgibbon: We should use a different name than "coin". I do think this should give equal weighting to components, for now - no need for configurability past what we already have in the component loss configs.

pr-plan.md
jpdunc23: >  Or even just _rng, does it use another one?

No, though `seed_step_sampler` does call `ComponentLossSchedule.seed_rng`, so I think it is best to disambiguate using `_which_component_rng`

> I would advise against a RandomChoice class.

OK, there will be some repetition but perhaps less than I feared given that I don't think we actually need `seed_rng` here.

jpdunc23: Let's call this `_component_choice_rng`.

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+        atmosphere_schedule: ComponentLossSchedule,
+        single_component_optim: bool = False,  # NEW — enable the per-batch coin
+    ):
+        ...
+        # NEW coin state:
+        self._single_component_optim: bool = single_component_optim
+        self._coin_rng: np.random.RandomState | None = None  # lazy, like TimeLengthProbabilities
+        self._selected_realm: Literal["ocean", "atmosphere"] | None = None  # None == both eligible
+
+    def _initialize_coin_rng(self) -> None:  # NEW — private; lazy-init, mirrors TimeLengthProbabilities.initialize_rng
```
jpdunc23: Call this `_init_component_choice_rng`

pr-plan.md
```diff
@@ -0,0 +1,228 @@
...
+        # if self._coin_rng is None:
+        #     self._coin_rng = np.random.RandomState(
+        #         Distributed.get_instance().get_seed() + <UNIQUE_OFFSET>  # != 684
+        #     )  # identical across ranks -> identical coin decisions, reproducible from run seed
+        ...
+
+    def flip_optimization_coin(self) -> None:  # NEW — re-draw selected realm; no-op when disabled
```
jpdunc23: Call this `component_optimization_choice`

pr-plan.md
```diff
@@ -0,0 +1,228 @@
+# Add `use_single_component_optim` flag to `CoupledTrainStepperConfig`
```
jpdunc23: Replace `use_single_component_optim` with `optimize_single_component_per_batch` throughout.

pr-plan.md
jpdunc23: Suggesting `component_optimization_choice()` to match the numpy `RandomState.choice()` convention.

pr-plan.md
jpdunc23: Agreed

pr-plan.md
jpdunc23: Revisiting the seed_rng point — we do need a separate eval-seeded RNG after all. The reported loss is `optimization.get_accumulated_loss()`, gated by `step_is_optimized`; if the choice is disabled in eval, validation accumulates both realms while a training batch accumulates one, so val loss ≈ 2× the (mixture) train loss and the curves are not comparable. Fix: keep the choice active in eval with a separate eval RNG reseeded in `seed_rng` (renamed from `seed_step_sampler`), mirroring the `_n_steps_sampler`/`_eval_n_steps_sampler` split. Per-step metrics for both realms still flow through the ungated `compute_loss`. Naming stays `_component_choice_rng` (+ `_eval_component_choice_rng`). Details in the plan — leaving this open for your take on the added eval-RNG machinery.

(written by agent)

pr-plan.md
jpdunc23: Done — renamed to `optimize_single_component_per_batch` throughout the plan.

pr-plan.md
jpdunc23: Done — `_init_component_choice_rng()`.

pr-plan.md
jpdunc23: Done — `component_optimization_choice()`.

pr-plan.md
jpdunc23: Adopted — `__post_init__` raises `ValueError` when the flag is on and a realm is statically null (`loss_weight == 0.0`, or an `n_steps` max yielding no optimized step).

pr-plan.md
jpdunc23: Adopted `optimize_single_component_per_batch`.

pr-plan.md
jpdunc23: Done — renamed to `component_optimization_choice()`.

pr-plan.md
jpdunc23: Done — dropped "coin" terminology throughout; kept equal-weighted 50/50 among non-null realms, no configurability.

fme/coupled/stepper.py
```diff
@@ -1434,15 +1453,81 @@ def sample_n_steps(self) -> None:
...
 
-    def seed_step_sampler(self, seed: int) -> None:
+    def seed_rng(self, seed: int) -> None:
+        """Reseed the eval-only RNGs deterministically.
+
+        Renamed from ``seed_step_sampler``: besides the per-schedule eval
+        n_steps samplers it now also reseeds the eval component-choice RNG, so
+        each validation pass replays the same single-component selections
+        regardless of the training RNG position. Training RNGs are distributed-
+        seeded lazily and are intentionally left untouched.
```
jpdunc23: Remove or revise this comment; in general, comments should relate to current state of code, not previous states.

fme/coupled/stepper.py
```diff
@@ -1398,6 +1399,13 @@ def _process_ensemble_output_list(
     return ocean_gen_data, atmos_gen_data
 
 
+# Offset added to the distributed/eval seed for the per-batch component-choice
+# RNG. Must differ from TimeLengthProbabilities' reserved 684 and from the
+# small per-schedule offsets (0, 1) used by seed_rng, so the choice RNG never
+# shares a seed with an n_steps sampler.
+_COMPONENT_CHOICE_RNG_OFFSET = 90210
```
jpdunc23:
```suggestion
_COMPONENT_CHOICE_RNG_OFFSET = 685
```

fme/coupled/stepper.py
```diff
@@ -1434,15 +1453,81 @@ def sample_n_steps(self) -> None:
...
+        Renamed from ``seed_step_sampler``: besides the per-schedule eval
+        n_steps samplers it now also reseeds the eval component-choice RNG, so
+        each validation pass replays the same single-component selections
+        regardless of the training RNG position. Training RNGs are distributed-
+        seeded lazily and are intentionally left untouched.
+        """
         for i, schedule in enumerate(self._schedules.values()):
             schedule.seed_rng(seed + i)
+        self._eval_component_choice_rng = np.random.RandomState(
+            seed + _COMPONENT_CHOICE_RNG_OFFSET
```
jpdunc23, line 1479: Ensures `_eval_component_choice_rng` has a distinct seed from the schedules.

fme/coupled/stepper.py
```diff
@@ -1434,15 +1453,81 @@ def sample_n_steps(self) -> None:
...
+        )
+
+    def _init_component_choice_rng(self) -> None:
+        """Lazily seed the training component-choice RNG from the distributed
+        seed, mirroring ``TimeLengthProbabilities.initialize_rng``. Identical
+        across ranks (so every rank makes the same selection) and reproducible
+        from the run seed.
```
mcgibbon: Note: I'd hold back on declaring reproducibility, I think this will still give different results especially when pre-empted. Weight initialization is reproducible because it happens at the start, and dataset shuffling is reproducible because it's seeded per-epoch and when resuming we go to the correct index in that seeded epoch. However for batch-by-batch selections like this and the random-number-of-forward-steps, we would need to determine a per-epoch sequence of values reproducibly from an epoch seed, and then follow that sequence for each batch index of the epoch. Since GPUs are already not reproducible I don't think we need to worry about doing this, but the docstring should be updated.

Seeding is still necessary so that across distributed GPU ranks, the same choice is made. We should consider adding a distributed reduction check to ensure the same seed gets set on all ranks, but that's out of scope for this PR, and reading the code I think it's implemented correctly.

I do think you could limit this object to a single `self._rng`.

fme/coupled/stepper.py
```diff
@@ -1434,15 +1453,81 @@ def sample_n_steps(self) -> None:
         for schedule in self._schedules.values():
             schedule.sample_n_steps()
 
-    def seed_step_sampler(self, seed: int) -> None:
+    def seed_rng(self, seed: int) -> None:
+        """Reseed the eval-only RNGs deterministically.
```
mcgibbon, line 1469: Question: why would we seed just the eval rng and not the train rng?

If the concern is we want the sampler to be seeded the same during each validation, it should be sufficient to seed a single `self._rng` when validation starts.

fme/coupled/stepper.py
```diff
@@ -1434,15 +1453,81 @@ def sample_n_steps(self) -> None:
...
+        if self._component_choice_rng is None:
+            self._component_choice_rng = np.random.RandomState(
+                Distributed.get_instance().get_seed() + _COMPONENT_CHOICE_RNG_OFFSET
+            )
+
+    def component_optimization_choice(self) -> None:
+        """Re-draw which single realm is optimized this batch (no-op when the
+        flag is disabled).
+
+        Must be called *after* ``sample_n_steps()`` so the per-batch null check
```
jpdunc23: Since the order matters, maybe this should just be combined with `sample_n_steps()` into a single method like `sample_from_rng()`.

pr-plan.md
mcgibbon, line 1: This file will need to be deleted before merge.

mcgibbon (commented):
Over-arching comment is it seems like this could be significantly simplified by using a single `self._rng` object. Can you see if that's possible?

mcgibbon:
From Claude discussion about replacing with one rng:

  Collapsing to a single self._rng — agreed, and it's a bigger simplification than it first looks. The second RNG existed only to make eval selections reproducible across validation passes; but the
  property eval actually needs is single-component scale-comparability, which is about how many realms are gated, not which one — so a single free-running RNG shared by train and eval preserves
  comparability and stays rank-synchronized (all ranks draw once per batch in the same order). Once you accept that (consistent with dropping the reproducibility claim), you can drop not just
  _eval_component_choice_rng but also:
  - the seed_step_sampler→seed_rng rename (it was renamed because it now reseeds the eval choice RNG — drop that and it's back to being just the n_steps-sampler reseeder),
  - the train/eval branch inside component_optimization_choice(), and
  - the _is_training field on the loss (it's only read to pick between the two RNGs; the schedules track their own train/eval state).

  The only thing lost is exact reproducibility of which realm is picked per validation batch, which doesn't affect the aggregate val loss (still ~50/50 over the set) — i.e. precisely the property
  you're saying isn't worth keeping. Net: one RNG, lazily seeded once from the distributed seed, drawn from in both modes.

pr-plan.md
jpdunc23: Yes

fme/coupled/stepper.py
jpdunc23: > Question: why would we seed just the eval rng and not the train rng?

Eval rng is re-seeded deterministically before each validation loop; this is exactly what we already do for random steps seeding. 

> If the concern is we want the sampler to be seeded the same during each validation, it should be sufficient to seed a single self._rng when validation starts.

This would corrupt the training seed, which, like in the `TimeLengthProbabilities` case, is seeded once at the beginning of training and left as is. We had that bug in ace training and I recall reviewing a PR where you fixed it.

Yes, we don't have perfect reproducibility of training due to preemption, but the eval seed being set deterministic should mean that validation always does the same thing for each batch regardless of preemption.

fme/coupled/stepper.py
jpdunc23: > I do think you could limit this object to a single self._rng

I don't think that will work, and will lead to the same bug we previously had in `fme.ace` training.

pr-plan.md
jpdunc23: Naming and the no-`RandomChoice` decision both landed in the implementation (`_component_choice_rng` / `_eval_component_choice_rng`; the choice method is now privatized as `_component_optimization_choice`, with `sample_from_rng` the sole public entry point). The eval-RNG question I raised here is now being worked out concretely against the code — see the `stepper.py` threads on `seed_rng` / `_eval_component_choice_rng`. Closing this plan-level thread in favor of those; `pr-plan.md` will be deleted before merge.

fme/coupled/stepper.py
jpdunc23: Done — dropped the “Renamed from `seed_step_sampler`” note; the `seed_rng` docstring now describes only current behavior (reseeds the per-schedule eval n_steps samplers and the eval component-choice RNG each validation pass; training RNGs are left untouched).

fme/coupled/stepper.py
jpdunc23: Applied — `_COMPONENT_CHOICE_RNG_OFFSET = 685`, one past `TimeLengthProbabilities`' reserved 684 and distinct from the per-schedule `(0, 1)` offsets; updated the explanatory comment to match.

fme/coupled/stepper.py
jpdunc23: The distinct seed is now guaranteed by `_COMPONENT_CHOICE_RNG_OFFSET = 685` vs the per-schedule `(0, 1)` offsets and `TimeLengthProbabilities`' 684; the comment above the constant documents this. Resolving.

fme/coupled/stepper.py
jpdunc23: Done — folded the two draws into a single public `sample_from_rng()` (n_steps, then the component choice, in order) and privatized `sample_n_steps` / `component_optimization_choice` as `_sample_n_steps` / `_component_optimization_choice`, so external callers can't transpose them. `CoupledTrainStepper.train_on_batch` now calls `sample_from_rng()`; tests updated accordingly.

fme/coupled/stepper.py
jpdunc23: Reproducibility wording: agreed — toned down the `_init_component_choice_rng` docstring. It now says the training RNG is seeded identically across ranks (so every rank makes the same per-batch choice) but no longer claims per-batch *training* reproducibility across preemption/resume, matching the random-n_steps sampler.

On limiting it to a single `self._rng`: I think it has to stay two RNGs. If we reseed one shared RNG when validation starts, training resumes from that same RNG state after *every* validation pass, so the per-batch training selection sequence repeats identically each epoch (batch 0 always ocean, batch 1 always atmosphere, …) — it collapses the per-epoch randomness rather than just perturbing it. Not reseeding at all would instead make the *validation* selection non-reproducible (the val-loss curve picks up selection noise that varies run-to-run and under preemption). One RNG can't be both free-running for training and deterministically reseeded for eval — which is exactly why `_n_steps_sampler` / `_eval_n_steps_sampler` are split today. Left the cross-rank seed-equality check out of scope as you suggested.

fme/coupled/stepper.py
jpdunc23: Directly: we deterministically reseed the eval RNG before each validation pass but not the train RNG because they have opposite requirements — validation should replay the same selections every pass (comparable val-loss curve), training should draw fresh each epoch.

A single `self._rng` reseeded when validation starts breaks the training side: every validation resets the shared RNG to the same seed, so training resumes from the same state each epoch and its per-batch selection sequence repeats identically epoch-to-epoch. That's the concrete failure — and it's sharper than how I put it earlier: it isn't that the seed is “corrupted,” it's that the training sequence stops *varying* across epochs. The free-running train RNG + deterministically-reseeded eval RNG mirrors `_n_steps_sampler` / `_eval_n_steps_sampler`. Happy to resolve unless you'd like the cross-rank seed-equality check pulled in (currently out of scope).

fme/coupled/stepper.py
```diff
@@ -1398,6 +1399,13 @@ def _process_ensemble_output_list(
     return ocean_gen_data, atmos_gen_data
 
 
+# Offset added to the distributed/eval seed for the per-batch component-choice
+# RNG. Chosen as 685 -- one past TimeLengthProbabilities' reserved 684, and
+# distinct from the small per-schedule offsets (0, 1) used by seed_rng, so the
+# choice RNG never shares a seed with an n_steps sampler.
+_COMPONENT_CHOICE_RNG_OFFSET = 685
```
mcgibbon, line 1406: Suggestion for a different PR: Add a `Distributed.get_independent_seed()` routine that will return `self._seed + 684 + self._independent_seed_index` and increment the index, so we can't accidentally use two identical "independent" seeds.

fme/coupled/stepper.py
```diff
@@ -1422,6 +1431,16 @@ def __init__(
...
             "atmosphere": atmosphere_schedule.loss_weight,
         }
+        # Per-batch single-component selection state. Two RNGs mirror
+        # ComponentLossSchedule's _n_steps_sampler / _eval_n_steps_sampler split
+        # so the validation loss stays single-component (comparable in scale to
+        # the training batch loss) yet reproducible. ``_selected_realm is None``
+        # means both realms are eligible (the disabled / no-selection case).
+        self._optimize_single_component_per_batch = optimize_single_component_per_batch
+        self._component_choice_rng: np.random.RandomState | None = None
+        self._eval_component_choice_rng: np.random.RandomState | None = None
```
mcgibbon, line 1441: For a different PR, maybe we can make a single concept that manages having an eval and train version of a thing that can be seeded, since we're using it both here and in the step sampler and it's a fair bit of logic.

mcgibbon (approved):
LGTM, thank you!

Two potential follow-on suggestions, but please don't implement them in this PR.

fme/coupled/stepper.py
jpdunc23: Good idea.

fme/coupled/stepper.py
jpdunc23: This is sort of why I proposed `RandomChoice`, but agreed that a more general object not necessarily tied to the `choice` API makes sense for a follow up to reduce repetition / improve reusability.

PR #1318, opened 2026-06-25
Exclude missing variables from channel_mean

AnnaKwa:
If `allow_missing_variables=True` during training, the channel mean metrics are all NaN because that flag sets the missing variables target values to NaN. This PR updates the channel mean to exclude fields with all-nan targets so that it is finite for training with missing variables.

I opted to check the targets for all-nans since it was a lot cleaner than threading missing variable names through the dataset info through to the aggregator. If we want to be strict about only excluding the fields if allow_missing_variables is True then `XarrayDataset._missing_names` would have to be threaded through to the aggregator.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/time_mean.py
+++ b/fme/ace/aggregator/inference/time_mean.py
@@ -384,8 +384,17 @@ class TimeMeanEvaluatorAggregator:
             channel_mean_names = [
                 name for name in channel_mean_names if name not in all_nan_target_names
             ]
-            values_to_average = [rmse_all_channels[name] for name in channel_mean_names]
-            logs.update({metric_name: sum(values_to_average) / len(values_to_average)})
+            if channel_mean_names:
+                values_to_average = [
+                    rmse_all_channels[name] for name in channel_mean_names
+                ]
+                logs.update(
+                    {metric_name: sum(values_to_average) / len(values_to_average)}
+                )
+            else:
+                raise ValueError(
+                    "All target variables are NaN; cannot compute channel mean."
+                )
 
         if len(label) != 0:
             return {f"{label}/{key}": logs[key] for key in logs}
--- a/fme/ace/aggregator/inference/time_mean.py
+++ b/fme/ace/aggregator/inference/time_mean.py
@@ -342,8 +342,14 @@ class TimeMeanEvaluatorAggregator:
         preds = self._get_target_gen_pairs()
         bias_map_key = "bias_map"
         rmse_all_channels = {}
+        # Variables whose target is entirely NaN have no valid data (e.g. filled
+        # by allow_missing_variables); their RMSE is NaN and they are excluded
+        # from the channel mean below.
+        all_nan_target_names: set[str] = set()
         for pred in preds:
             rmse_all_channels[pred.name] = pred.rmse()
+            if torch.isnan(pred.target).all():
+                all_nan_target_names.add(pred.name)
             should_log = self._log_variables is None or pred.name in self._log_variables
             if should_log:
                 bias_image = plot_paneled_data(
@@ -363,7 +369,7 @@ class TimeMeanEvaluatorAggregator:
         if self._target == "norm":
             metric_name = "rmse/channel_mean"
             if self._channel_mean_names is None:
-                values_to_average = list(rmse_all_channels.values())
+                channel_mean_names = list(rmse_all_channels)
             else:
                 missing = [
                     n for n in self._channel_mean_names if n not in rmse_all_channels
@@ -374,9 +380,11 @@ class TimeMeanEvaluatorAggregator:
                         f"recorded data: {missing}. Available: "
                         f"{sorted(rmse_all_channels)}."
                     )
-                values_to_average = [
-                    rmse_all_channels[name] for name in self._channel_mean_names
-                ]
+                channel_mean_names = list(self._channel_mean_names)
+            channel_mean_names = [
+                name for name in channel_mean_names if name not in all_nan_target_names
+            ]
+            values_to_average = [rmse_all_channels[name] for name in channel_mean_names]
             logs.update({metric_name: sum(values_to_average) / len(values_to_average)})
 
         if len(label) != 0:
--- a/fme/ace/aggregator/one_step/ensemble.py
+++ b/fme/ace/aggregator/one_step/ensemble.py
@@ -325,17 +325,17 @@ class _EnsembleAggregator:
                     names = list(self._channel_mean_names)
                 # Exclude variables whose target was entirely NaN (e.g. filled
                 # by allow_missing_variables) from the channel mean.
-                excluded = self._all_nan_target_names or set()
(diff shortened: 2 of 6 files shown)


fme/ace/aggregator/inference/time_mean.py
```diff
@@ -374,9 +380,11 @@ def get_logs(self, label: str) -> dict[str, float | torch.Tensor | Image]:
...
                         f"{sorted(rmse_all_channels)}."
                     )
-                values_to_average = [
-                    rmse_all_channels[name] for name in self._channel_mean_names
-                ]
+                channel_mean_names = list(self._channel_mean_names)
+            channel_mean_names = [
+                name for name in channel_mean_names if name not in all_nan_target_names
+            ]
+            values_to_average = [rmse_all_channels[name] for name in channel_mean_names]
```
yyexela: Nitpick (non-blocking): Here, if all channels are nan, then a division by zero would occur.

Suggestion: It would be nice to have a guard to raise an error saying that no channels here have non-nan values (everything is nan) to raise an error sooner (or, as Claude suggested just return NaN) and be more informative than having it fail on division by zero.

fme/ace/aggregator/one_step/reduced_metrics.py
```diff
@@ -94,8 +98,19 @@ def record(self, target: TensorMapping, gen: TensorMapping):
...
+            }
+        # Exclude variables whose target is entirely NaN (e.g. filled by
+        # allow_missing_variables) from the channel mean.
+        included = [
+            name
+            for name in channel_mean_names
+            if name not in self._all_nan_target_names
+        ]
+        for name in included:
+            self._channel_mean += batch_avgs[name] / len(included)
```
yyexela: Nitpick (minor): `fme/ace/aggregator/one_step/ensemble.py` uses `excluded`, but this file uses `included`.

Suggestion: Consider picking either `included` or `excluded` for the logic here, and make it more descriptive like `all_nan_channels` so that it's clear what the variable represents and to keep it consistent across files.

fme/ace/aggregator/one_step/reduced_metrics.py
```diff
@@ -68,6 +68,10 @@ def __init__(
         self._channel_mean: torch.Tensor | None = None
         self._device = device
         self._channel_mean_names = channel_mean_names
+        # Variables whose target is entirely NaN (e.g. filled by
+        # allow_missing_variables); detected once on the first batch since
+        # missingness is constant, then excluded from the channel mean.
```
yyexela, line 73: Nitpick (non-blocking): This PR adds lots of comments, I think many of them are unnecessary, but they don't directly hurt the code.

Suggestion: Consider removing or shortening the comments added.

fme/ace/aggregator/one_step/reduced_metrics.py
yyexela, line 93: Issue: If no metrics are non-nan, then `_channel_mean` is shown as `0.0`.

Suggestion: If all variables are NaN then I think we should have this metric return NaN as well, not `0.0`. Open to other ideas.

yyexela (changes requested):
Looks good overall, just a couple things worth flagging. Overall, Claude raised this issue:

When every channel-mean variable has an all-NaN target (empty included set), the three files behave differently: time_mean.py raises ZeroDivisionError,   reduced_metrics.py silently returns 0.0, and ensemble.py omits the key. The 0.0 is the worst — it looks like a real metric. This is only reachable via degenerate config   (e.g. a whole component missing in coupled runs), so it's not a normal path, but the divergence is worth unifying. Suggest all three return NaN when the set is empty,   matching the existing "no valid data → NaN" semantics in get/get_channel_mean.

Arcomano1234:
> When every channel-mean variable has an all-NaN target (empty included set), the three files behave differently: time_mean.py raises ZeroDivisionError, reduced_metrics.py silently returns 0.0, and ensemble.py omits the key. The 0.0 is the worst — it looks like a real metric. This is only reachable via degenerate config (e.g. a whole component missing in coupled runs), so it's not a normal path, but the divergence is worth unifying. Suggest all three return NaN when the set is empty, matching the existing "no valid data → NaN" semantics in get/get_channel_mean.

Yeah my claude flagged that as well but I am pretty sure its a nonissue as the ZeroDivisionError should not ever be hit because we would never have a situation where every variable is excluded (and if we did that by mistake I would rather have the code error out or have a check some where else)

fme/ace/aggregator/inference/time_mean.py
Arcomano1234: If we do want to save guard this I suggest we mirror how its done in `ensemble.py` where we have `if name:` so there is no dividing by zero. But also I do not think we will ever hit that here.

AnnaKwa: Added check to raise ValueError if not `channel_mean_names`

fme/ace/aggregator/one_step/reduced_metrics.py
AnnaKwa: renamed to `non_nan_targets`

AnnaKwa:
> Looks good overall, just a couple things worth flagging. Overall, Claude raised this issue:
> 
> When every channel-mean variable has an all-NaN target (empty included set), the three files behave differently: time_mean.py raises ZeroDivisionError, reduced_metrics.py silently returns 0.0, and ensemble.py omits the key. The 0.0 is the worst — it looks like a real metric. This is only reachable via degenerate config (e.g. a whole component missing in coupled runs), so it's not a normal path, but the divergence is worth unifying. Suggest all three return NaN when the set is empty, matching the existing "no valid data → NaN" semantics in get/get_channel_mean.

I added checks for the case where the non-NaN name list is empty, but raise an error instead of returning a value. As in Troy's comment I think it is preferable to stop the job if this situation arose somehow (I guess you could set `allow_missing_variables=True` and point it at a dataset that has a different naming convention?)

fme/ace/aggregator/one_step/ensemble.py
```diff
@@ -315,6 +323,10 @@ def _get_data(self):
                             f"{sorted(all_keys)}."
                         )
                     names = list(self._channel_mean_names)
+                # Exclude variables whose target was entirely NaN (e.g. filled
+                # by allow_missing_variables) from the channel mean.
+                excluded = self._all_nan_target_names or set()
+                names = [name for name in names if name not in excluded]
```
yyexela: Thanks for the updates! Looks like `excluded` didn't get renamed to `nan_targets` or `non_nan_targets` like in `fme/ace/aggregator/one_step/reduced_metrics.py`. Otherwise PR looks good.

yyexela (changes requested):

fme/ace/aggregator/one_step/ensemble.py
AnnaKwa: Renamed

yyexela (approved):
Looks good!

PR #1319, opened 2026-06-26
Refactor correctors to ordered sequences of Correction objects

mcgibbon:
Follow-on to #1313 (stacked on its branch). #1313 addressed the review by passing `SeaIceFractionConfig.__call__` to `OceanCorrector` as a Callable instead of the config object. This PR generalizes that across all correctors, per the review discussion: rather than a `*CorrectorParams` scalar bag that the corrector reads, each correction step becomes a small callable object that bundles its own params plus the operators it needs and applies one step. The corrector holds an ordered sequence of these and just applies them in turn, so it no longer reads any config/params fields itself (addresses the review point that the corrector knows too much about the leaf configs).

The YAML config dataclasses are **unchanged** — `_build` packages the already-parsed config params into these function-objects. No config-schema change.

Changes:
- `fme.core.corrector.registry`: add `Correction` protocol (the `(input, gen, forcing, state) -> (gen, state)` step signature) and a `CorrectionSequence` corrector base whose `__call__` applies an ordered list of `Correction`s.
- `fme.core.corrector.utils`: add `ForcePositive` correction object (wraps the existing `force_positive`).
- `fme.core.corrector.atmosphere`: add `ConserveDryAir`, `ZeroGlobalMeanMoistureAdvection`, `MoistureBudgetCorrection`, `TotalEnergyBudgetCorrection`; `AtmosphereCorrector` is now a thin `CorrectionSequence`; `_build` assembles the ordered list (drops `AtmosphereCorrectorParams`).
- `fme.core.corrector.ocean`: add `SeaIceFractionCorrection` (wraps `SeaIceFractionConfig`), `SurfaceEnergyFluxCorrection`, `OceanHeatContentCorrection`; `OceanCorrector` is now a thin `CorrectionSequence`; `_build` assembles the ordered list (drops `OceanCorrectorParams` and the interim `SeaIceFractionCorrection` Callable alias from #1313).
- `fme.core.corrector.ice`: add `IceBudgetCorrection` (wraps `IceBudgetCorrectionConfig`); `IceCorrector` is now a thin `CorrectionSequence`.

- [x] Behavior-preserving refactor; existing 50 corrector tests pass unchanged (no new tests added)

### DIFF
--- a/fme/core/corrector/atmosphere.py
+++ b/fme/core/corrector/atmosphere.py
@@ -12,13 +12,23 @@ from fme.core.atmosphere_data import (
     compute_layer_thickness,
 )
 from fme.core.constants import GRAVITY, SPECIFIC_HEAT_OF_DRY_AIR_CONST_VOLUME
-from fme.core.corrector.registry import CorrectorABC, CorrectorConfigABC
+from fme.core.corrector.registry import (
+    Correction,
+    CorrectionSequence,
+    CorrectorConfigABC,
+)
 from fme.core.corrector.state import CorrectorState
-from fme.core.corrector.utils import force_positive
+from fme.core.corrector.utils import ForcePositive
 from fme.core.dataset_info import DatasetInfo
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.registry.corrector import CorrectorSelector
 from fme.core.typing_ import TensorDict, TensorMapping
+
+
+class AreaWeightedMean(Protocol):
+    def __call__(
+        self, data: torch.Tensor, keepdim: bool, name: str | None = None
+    ) -> torch.Tensor: ...
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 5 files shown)


mcgibbon:
Pre-review agent: Verified behavior-preserving locally — 50/50 corrector tests pass, `ruff check` + `ruff format --check` clean. Checked against the pre-refactor behavior: correction ordering assembled in each `_build`, the `gen_data` copy semantics in `CorrectionSequence.__call__`, `corrector_state` threading (including `ConserveDryAir` seeding from the IC), the call-time `ValueError` for a missing vertical coordinate (still raised on call, not at build), and the device-dependent precision logic that moved into `_build`. No code outside the package references the removed `*CorrectorParams` classes or the interim `SeaIceFractionCorrection` Callable alias.

mcgibbon:
Pre-review agent (optional, non-blocking): `AreaWeightedMean(Protocol)` is now defined identically in both `atmosphere.py` and `ocean.py`. The duplication predates this PR (it's only moved to the top of each file here), but since both files' `Correction` objects depend on it and this PR already introduces a shared `Correction` protocol in `registry.py`, it might be worth hoisting `AreaWeightedMean` into `registry.py` while you're in here. Up to you.

mcgibbon:
Merged into #1313 (fast-forwarded its branch `refactor/corrector-configs-builder-pattern` to include these commits), so this stacked PR is now empty. Closing; the Correction-objects refactor will land as part of #1313.

PR #1320, opened 2026-06-26
Add exclude_top_levels option to energy budget corrector

mcgibbon:
The `constant_temperature` total-energy-budget correction applies a vertically- and horizontally-uniform air-temperature increment to enforce total-energy conservation. It currently adds that increment to *every* vertical level, including the thin model top level. In the 4°/daily ERA5-only baseline this introduces a positive bias in the topmost air temperature level. This PR adds an option to exclude the top N levels from receiving the increment, so the bias can be tested/removed without changing the conserved global magnitude.

Changes:
- `fme.core.corrector.atmosphere.EnergyBudgetConfig`: add `exclude_top_levels: int = 0` field (with a non-negative `__post_init__` check); it only restricts which levels receive the per-level temperature addition, the global magnitude of the correction is still computed across all levels.
- `fme.core.corrector.atmosphere._force_conserve_total_energy`: accept `exclude_top_levels` and skip the topmost N `air_temperature` levels (level 0 is the model top) when applying the increment; raise if it exceeds the number of levels. Threaded through `AtmosphereCorrector.__call__`.
- `fme.core.corrector.test_atmosphere`: excluded top level receives no increment while remaining levels get the same per-level magnitude as the no-exclusion case; negative-value and too-many-levels error cases.

- [x] Tests added

### DIFF
--- a/fme/core/corrector/atmosphere.py
+++ b/fme/core/corrector/atmosphere.py
@@ -33,10 +33,23 @@ class EnergyBudgetConfig:
             to the energy flux into the atmosphere when conserving total energy.
             This can be useful for correcting errors in energy budget in target data.
             The same additional heating is imposed at all time steps and grid cells.
+        exclude_top_levels: Number of topmost vertical levels (level 0, then 0-1,
+            etc.) to exclude from receiving the temperature correction. The global
+            magnitude of the correction is still computed across all levels; the
+            excluded top levels simply do not receive the per-level temperature
+            addition. Defaults to 0 (apply the correction to all levels).
     """
 
     method: Literal["constant_temperature"]
     constant_unaccounted_heating: float = 0.0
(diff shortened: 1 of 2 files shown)


mcgibbon:
Validating this change on the 4°/daily v2 ERA5-only baseline with the energy corrector and `exclude_top_levels: 1` (excludes the model top level, level 0):

- beaker: https://beaker.org/ex/01KW2D1V2F4FY85R3N7M0GFWEV
- wandb: https://wandb.ai/ai2cm/ace/runs/bre6554p

Tests whether excluding the top level removes the positive `air_temperature_0` bias the all-levels corrector introduced, without reintroducing energy-budget drift.

mcgibbon:
Closing as impact was not positive and not particularly promising. Won't merge.

PR #1321, opened 2026-06-26
Add `CorrectorOutput` with `CorrectorDiagnostics` containing `delta = corrected - network_output`

jpdunc23:
Introduces the value objects and helper that let a corrector report, per step, which variables it adjusted and by how much, without a downstream caller having to infer it. `CorrectorDiagnostics` holds a single `delta` mapping where `delta[name] = corrected[name] − network_output[name]` for each variable a corrector declares it touched; the raw pre-correction network value stays recoverable as `output − delta`. `CorrectorOutput` bundles the corrected output, its diagnostics, and the passthrough corrector state, giving the corrector a named return value that can grow new per-step payloads without breaking call sites. `build_corrector_diagnostics` computes the delta from an explicit declared touched-name set rather than by diffing tensors by object identity.

This is a pure addition: nothing returns or consumes these types yet, so corrected outputs, rollout state, losses, logged metrics, and saved files are unchanged, and existing checkpoints load unchanged. A follow-up change has the corrector return `CorrectorOutput` and removes the current object-identity capture; this PR lands the foundation in isolation so it can be reviewed on its own.

Changes:
- `fme.core.corrector.output.CorrectorDiagnostics` — new value object with a single `delta` `TensorMapping`, defaulting to empty.
- `fme.core.corrector.output.CorrectorOutput` — new value object bundling `corrected`, `diagnostics`, and a passthrough `corrector_state`.
- `fme.core.corrector.utils.build_corrector_diagnostics` — new helper computing `delta = corrected − input_snapshot` over an explicit touched-name set (cumulative for a name touched more than once).

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,122 @@
...
+    delta[name] = corrected[name] − network_output[name] for each variable the
+    corrector declares it touched. Empty when no corrector ran or none modified
+    anything, so consumers need no None checks. The network's raw pre-correction
+    value is recoverable as output − delta wherever the corrected output is known.
+
+    Shaped as a value object (not a bare mapping) so later work can add a category
+    label and metric fields without re-opening the corrector return type.
+    """
+
+    delta: TensorMapping = dataclasses.field(default_factory=dict)
```
mcgibbon: Note: Our plan also included having something like a `prescribed` field that is deemed to be replacing a field, not applying a delta, and would e.g. be treated differently in corrector-based losses.

pr-plan.md
```diff
@@ -0,0 +1,122 @@
...
+from fme.core.corrector.state import CorrectorState
+from fme.core.typing_ import TensorDict, TensorMapping
+
+
+@dataclasses.dataclass
+class CorrectorDiagnostics:
+    """Per-step diagnostic payload describing a corrector's effect.
+
+    delta[name] = corrected[name] − network_output[name] for each variable the
+    corrector declares it touched. Empty when no corrector ran or none modified
```
mcgibbon: Note: "replaced" values like the advected total column humidity in the case of columnwise moisture budget closure would not be included in "delta", at least not by default.

pr-plan.md
```diff
@@ -0,0 +1,122 @@
...
+    object-identity ``captured_before`` heuristic (removed when the corrector
+    starts returning ``CorrectorOutput``). For a name modified by more than one
+    enabled correction, ``delta`` is the cumulative ``corrected − input_snapshot``,
+    so ``input_snapshot = corrected − delta`` stays exact with no per-operation
+    bookkeeping.
+    """
+    delta = {
+        name: corrected[name] - input_snapshot[name] for name in touched_names
+    }
+    return CorrectorDiagnostics(delta=delta)
```
mcgibbon: Issue: this doesn't allow us to control what corrector impacts are deemed to be deltas and which ones aren't, as we discussed.

mcgibbon:
I don't know if you were planning on handling these in downstream PRs, but at least `build_corrector_diagnostics` seems like something to avoid as it can cause issues with the overall design in later PRs if ever used.

pr-plan.md
jpdunc23: A follow-on PR will handle labeling for default behavior. For various reasons the "delta" vs "prescribed" distinction is not well-defined (e.g., the hfds "prescribed" correction is both replacement and delta) but every correction can be described cleanly as `delta = corrected - network_output`. So the current plan I have in mind is to simply include labels which signal default intentions for downstream uses. 

Let's please save the debate for the follow on PR.

pr-plan.md
jpdunc23: `delta = corrected - network_output` is a clean and well-defined definition for every network output, regardless of whether it is replaced. The "replaced" vs "delta" distinction adds unnecessary confusion to the output structure that is not necessary at this stage.

In a follow-on PR, intended default downstream usage can be added when it is needed; that's out of scope here.

pr-plan.md
jpdunc23: All corrections are `delta = corrected - network_output`. Downstream handling for diagnostic and optim purposes is intentionally out of scope for this PR.

jpdunc23:
> I don't know if you were planning on handling these in downstream PRs, but at least build_corrector_diagnostics seems like something to avoid as it can cause issues with the overall design in later PRs if ever used.

`build_corrector_diagnostics` won't cause issues with design in later PRs. All corrected fields can be cleanly described as `delta = corrected - network_output`. Later PRs will tackle the handling of default optimization behavior, but that is out-of-scope for the time being.

jpdunc23:
Sharing the reasoning for keeping it, since the real question is whether it boxes in the later design — I don't think it does.

`build_corrector_diagnostics` is deliberately the most general primitive available: a declared diff, `delta[name] = corrected[name] − network_output[name]`, over the set of names a corrector says it touched. It computes a *value*, not a *policy*. Every later feature on the roadmap — a `prescribed`/`delta` label, per-field loss treatment, an opt-out for replaced fields — is additive on `CorrectorDiagnostics` (a new field on the value object) and leaves this helper's contract untouched. So if anything it's the foundation those labels sit on, not something they'd fight.

What *would* box us in is the opposite: deciding at corrector time which fields are "real" deltas vs replacements and storing only that. Because `StepOutput.output` is always present, a single `delta` keeps every view recoverable downstream — raw network value `= output − delta`, scale `= output / (output − delta)` — so no later policy is foreclosed. More detail on the inline threads.

pr-plan.md
jpdunc23: I'd argue this PR *preserves* that control rather than removing it. Because the corrected value (`StepOutput.output`) is always present alongside `delta`, every per-field view stays recoverable downstream:
- correction magnitude → `delta`
- raw network value → `output − delta`
- multiplicative scale → `output / (output − delta)`

So the correction loss/metric in the next PR can decide, per field, whether to treat an impact as a delta, a replacement, or neither — driven by a label or config, where the intent actually lives. The control isn't lost; it moves from a corrector-time storage split to a consumer-time policy, which is strictly more flexible since the corrector no longer has to pre-commit each field to a bucket. Storing one uniform `delta` is what keeps all of those options open.

pr-plan.md
jpdunc23: Agreed the prescribed-vs-delta concept needs to exist — the plan keeps it, just as a *label* on `CorrectorDiagnostics` in the follow-on PR rather than a second mapping here. Two reasons the label is the better home:

1. It's additive — a new field on this value object, no change to the corrector return type or this helper.
2. The bucket isn't always clean at the data layer: as you note, hfds is genuinely both a replacement and a delta, so a corrector can't authoritatively pre-sort fields. A label set where the loss/consumer knows its intent can.

And nothing is dropped in the meantime: for a fully-prescribed field, `delta = prescribed − network_output` captures the change exactly, with the prescribed value recoverable as `output` and the raw network value as `output − delta`. The loss in the next PR has everything it needs to treat prescribed fields differently.

pr-plan.md
jpdunc23: Quick clarification on the mechanic, since I think this case is actually covered: the touched-name set is the union of every variable the enabled options *write*, which includes replaced fields like the advected total column humidity under moisture-budget closure. So those fields *are* in `delta` (`= prescribed_value − network_value`), with the raw network value recoverable as `output − delta` — nothing replaced is excluded. Whether a consumer later chooses to *treat* a replaced field differently (e.g. drop it from a correction-magnitude loss) is the labeling policy from the other thread, in the follow-on PR.

pr-plan.md
mcgibbon: > So the current plan I have in mind is to simply include labels which signal default intentions for downstream uses.
>
> Let's please save the debate for the follow on PR.

OK, we can discuss this in the context of the PR that implements these features. At least we're in agreement a class like this with `delta` defined should be added, and I think the other details are best discussed in that PR's context.

> So the current plan I have in mind is to simply include labels which signal default intentions for downstream uses.

We had a long discussion where this was your plan going into the conversation, I posed my concerns with the plan, and we agreed on a different plan. That plan was basically to use multiple TensorMapping instead of labels, which would a single field to exist in multiple label buckets and be easier to work with than low-level one-hot-encodings of labels. Going back to the original plan I had concerns with will likely mean re-hashing that discussion, but it should be easier to have in the context of the next PR.

I think we need to talk more about HFDS specifically even separate from these PRs, because I'm kind of confused by some of it. Like the idea of corrector regularization and pre-corrector regularization, to my mind, are that we want to make it so that if the corrector were dropped, the network would give reasonable outputs, and that if we then use the corrector it's just being used for minor corrections. But that really doesn't seem to fit for HFDS.

I wonder if we might need a separate concept, outside of the corrector, for these "outputs that are derived from network outputs", and doing that could completely remove the need to have "prescribed" terms for the corrector (since they would live in the derived-output concept instead).

pr-plan.md
mcgibbon: The issue isn't that `delta` is poorly defined, it's that the purpose is supposed to be to enable regularization strategies, but it doesn't fit that purpose. 

For HDFS, with residual_prediction `delta = ocean_fraction · net_flux`. What's the purpose of regularizing this? When using `prescribed`, `delta = ocean_fraction · (net_flux − gen_hfds)` becomes a regularization term that encourages gen_hfds to become equal to net_flux. Is net_flux what the channel is being optimized on? If not, this is also an issue.

It's difficult to review this `delta` in a context where it's not used, can we move it to the next PR?

pr-plan.md
jpdunc23: > We had a long discussion where this was your plan going into the conversation, I posed my concerns with the plan, and we agreed on a different plan. That plan was basically to use multiple TensorMapping instead of labels, which would a single field to exist in multiple label buckets and be easier to work with than low-level one-hot-encodings of labels. Going back to the original plan I had concerns with will likely mean re-hashing that discussion, but it should be easier to have in the context of the next PR.

Actually trying to implement the plan we discussed last Thursday led to too much friction, thus the change in tack.

mcgibbon: > Actually trying to implement the plan we discussed last Thursday led to too much friction, thus the change in tack.

Hm, having just looked at some of the corrector code I think I can imagine what you mean. For example, some values get derived only after certain corrections are applied.

These scientific issues are something we're going to have to address, but maybe let's move forward with the PR.

mcgibbon:
We should move forward with a PR based on this plan, I sign off: https://github.com/ai2cm/ace/pull/1321#discussion_r3483426815

pr-plan.md
jpdunc23: Per https://github.com/ai2cm/ace/pull/1321#discussion_r3483426815 we're in agreement that `delta` stays in this PR.

pr-plan.md
jpdunc23: > the idea of corrector regularization and pre-corrector regularization, to my mind, are that we want to make it so that if the corrector were dropped, the network would give reasonable outputs, and that if we then use the corrector it's just being used for minor corrections

I agree with this

> But that really doesn't seem to fit for HFDS

I don't agree. For the "prescribed" method of HFDS correction (which I now realize is a terrible name), we still have a hybrid of ML + surface forcing. 

- Where sea ice fraction is 0 (fully ocean) the output is the "derived" surface flux from the forcing; this always has small irreducible error in practice because our AM4 fluxes are output on a grid that is not fully consistent with the OM4 surface flux target HFDS.
- Where sea ice fraction is 1 (fully ice) the output is fully ML.

But the network still predicts something for HFDS over the entire ocean including open ocean grid cells that are entirely replaced by the derived flux from the AM4 flux forcing. Optimizing the error of that output  (or penalizing the magnitude of the delta) regardless encourages HFDS to stay close to the target (which, as we saw, improved the interannual variability of the 3D sea water temperature), but "prescribing" the open ocean flux *forces* Samudra to use the AM4 reference flux forcing when correcting the ocean heat content.

> I wonder if we might need a separate concept, outside of the corrector, for these "outputs that are derived from network outputs"

Don't we already have this concept via `derive_func`?

jpdunc23:
Plan as approved:

# Add `CorrectorOutput` with `CorrectorDiagnostics` containing `delta = corrected - network_output`

<details>

Adds two value objects to the corrector package — `CorrectorDiagnostics` (a single
`delta` mapping) and `CorrectorOutput` — plus a `build_corrector_diagnostics`
helper that computes `delta[name] = corrected[name] − input_snapshot[name]` over an
explicit declared name set. Pure addition: nothing returns or consumes these yet,
so behavior is unchanged.

---

## `fme/core/corrector/output.py` (new)

```python
import dataclasses

from fme.core.corrector.state import CorrectorState


@dataclasses.dataclass
class CorrectorDiagnostics:
    """Per-step diagnostic payload describing a corrector's effect.

    delta[name] = corrected[name] − network_output[name] for each variable the
    corrector declares it touched. Empty when no corrector ran or none modified
    anything, so consumers need no None checks. The network's raw pre-correction
    value is recoverable as output − delta wherever the corrected output is known.

    Shaped as a value object (not a bare mapping) so later work can add a category
    label and metric fields without re-opening the corrector return type.
    """

    delta: TensorMapping = dataclasses.field(default_factory=dict)


@dataclasses.dataclass
class CorrectorOutput:
    """Result of applying a corrector to a single step's generated data.

    corrected is the adjusted output (semantically identical to today's corrected
    gen_data). corrector_state is threaded through unchanged. diagnostics carries
    the per-step delta.
    """

    corrected: TensorDict
    diagnostics: CorrectorDiagnostics = dataclasses.field(
        default_factory=CorrectorDiagnostics
    )
    corrector_state: CorrectorState | None = None
```

## `fme/core/corrector/utils.py` (modified)

```python
from collections.abc import Iterable

from fme.core.corrector.output import CorrectorDiagnostics

# Existing contents kept unchanged: force_positive(data, names) and the
# ForcePositive Correction dataclass.

def build_corrector_diagnostics(  # NEW
    input_snapshot: TensorMapping,
    corrected: TensorMapping,
    touched_names: Iterable[str],
) -> CorrectorDiagnostics:
    """Build the diagnostics for a corrector from an explicit touched-name set.

    delta[name] = corrected[name] − input_snapshot[name] for name in touched_names.
    This is a declared diff over a known name set — it replaces the role of the
    object-identity ``captured_before`` heuristic (removed when the corrector
    starts returning ``CorrectorOutput``). For a name modified by more than one
    enabled correction, ``delta`` is the cumulative ``corrected − input_snapshot``,
    so ``input_snapshot = corrected − delta`` stays exact with no per-operation
    bookkeeping.
    """
    delta = {
        name: corrected[name] - input_snapshot[name] for name in touched_names
    }
    return CorrectorDiagnostics(delta=delta)
```

The helper takes `touched_names` explicitly so it is unit-testable in isolation.
A follow-up change supplies that set as the union of the names each enabled
`Correction` declares it writes (a corrector is a `CorrectionSequence` of
composable `Correction` objects), and detaches the resulting `delta` at the
step boundary. None of that is in this PR — this PR only adds the value objects
and the helper.

---

## Tests

## `fme/core/corrector/test_output.py` (new)

```python
def test_corrector_diagnostics_delta_defaults_empty():
    # GOAL: a default-constructed CorrectorDiagnostics has an empty delta mapping,
    # so non-corrector / no-op steps need no None checks.
    ...

def test_corrector_output_defaults():
    # GOAL: CorrectorOutput(corrected=...) defaults to empty diagnostics and
    # corrector_state=None; corrected is carried through verbatim.
    ...
```

## `fme/core/corrector/test_utils.py` (new)

```python
# Hand-computable tensors so delta values are asserted exactly. Prior-art test
# style: fme/core/corrector/test_ocean.py (DEVICE = get_device(), small tensors).

def test_build_corrector_diagnostics_delta():
    # GOAL: delta contains exactly touched_names and delta[name] equals
    # corrected[name] − input_snapshot[name], asserted exactly. Names present in
    # the inputs but absent from touched_names do NOT appear in delta.
    ...

def test_build_corrector_diagnostics_cumulative():
    # GOAL: when corrected reflects two corrections applied to one field
    # (snapshot x → x + a + b), delta[name] == a + b, and
    # corrected − delta == input_snapshot exactly (recoverability).
    ...

def test_build_corrector_diagnostics_empty():
    # GOAL: empty touched_names → empty delta (the no-op / no-corrector case).
    ...
```

</details>

fme/core/corrector/output.py
```diff
@@ -0,0 +1,92 @@
+"""Value objects describing the per-step result of a corrector.
+
+These are pure additions in PR1 of the corrector-diagnostics work: they define
+the *shape* of a corrector's diagnostic payload without yet being returned by
+any corrector or carried on ``StepOutput`` (that wiring is issue 02). They live
+in the corrector package so ``fme.core`` does not import from ``fme.ace``.
+"""
```
mcgibbon: This comment refers to the process of the code being added, and should probably be removed from the file header. We don't usually add headers like this.

fme/core/corrector/output.py
```diff
@@ -0,0 +1,92 @@
...
+from fme.core.typing_ import TensorDict, TensorMapping
+
+
+@dataclasses.dataclass
+class CorrectorDiagnostics:
+    """The diagnostic effect of a corrector on a single step.
+
+    Parameters:
+        delta: ``corrected[name] - network_output[name]`` for each variable the
+            corrector declares it touched. The network's raw pre-correction
```
mcgibbon, line 14: While I continue to have reservations about the difference between an output derived from network outputs and a corrected network output, I don't expect to fix the issue in this PR chain. Hopefully we can punt on it and handle it after.

fme/core/corrector/output.py
```diff
@@ -0,0 +1,92 @@
...
+            value is recoverable as ``corrected[name] - delta[name]``. Defaults
+            to empty, which is the value carried when no corrector ran or none
+            modified anything.
+
+    This is a value object rather than a bare mapping so the follow-on PR can add
```
mcgibbon: Issue: avoid referring to the PR process itself. Be more concise.

@jpdunc23 can you massage each of the docstrings into a form you're happy with that doesn't have excess text? I've been using something like this in my own pre-review skill:

> We need to use a writing style that is explicit,
concise, and easy to understand. We should not use paragraphs a few sentences
would suffice.
>
> Does the PR restate information with unnecessary summaries before or after
that information? Does it state things that will be obvious in the context
where the text is being used? Does it state things that are obvious based on
other information in the text? These are all opportunities to tighten
the writing.

fme/core/corrector/output.py
jpdunc23: Agreed, let's remove this / move inline. Also, this violates the principle of "comments should be self-contained" by referring to "PR1" and "issue 02" which is going to just be stale context if this gets merged.

fme/core/corrector/output.py
jpdunc23: OK, though I want to note again that `hfds` does not fall under the category of "an output derived from network outputs"; It could be considered "an output in part derived from network inputs".

`tendency_of_total_water_path_due_to_advection` falls into the category of "an output derived from network outputs", as does the `IceBudgetCorrectionConfig.__call__` corrections -- these are full replacements of the network outputs. I remember when @William-gregory was developing FloeNet we had this tension of "derived variables that should be included in the optimization" and we considered adding the ability to "optimize derived outputs". Ultimately, we ended up using the corrector for simplicity (and, I guess, the `tendency_of_total_water_path_due_to_advection` precedent). 

So why does ACE2 include `tendency_of_total_water_path_due_to_advection` as an output name (optimized) rather than as a derived variable (non-optimized)? The answer is that with `advection_*` corrections enabled, optimizing `twp_total_tendency - gen.evaporation_rate - gen.precipitation_rate` provides valuable gradient signal about the advection error that is more than the sum of the errors for the individual outputs involved.

However, having `tendency_of_total_water_path_due_to_advection` as a network output and then throwing it away is computationally wasteful. Allowing for "pre-corrector" optimization of `tendency_of_total_water_path_due_to_advection` or "corrector regularization" of the `delta` amortizes this wastefulness by including potentially useful gradient signal from the network output for `tendency_of_total_water_path_due_to_advection`.

jpdunc23: A bit more detail on each point, for the record:

**Why the advection signal is "more than the sum of the individual errors."** The advection loss penalizes one specific *linear combination* of errors that are already penalized individually — `e_adv ≈ e_TWP/Δt − (e_E − e_P)`, i.e. the column moisture-budget residual. Adding it puts extra weight on that single direction in the joint error space: two states with identical per-field L2 errors in `specific_total_water_*`, `surface_pressure`, `evaporation_rate`, and `precipitation_rate` can still differ in `e_adv` (errors that conspire to violate budget closure vs. errors that cancel). The per-field losses can't distinguish those two states; the advection term can, so it pushes the network toward budget-consistent *error structure*. The `surface_pressure × specific_total_water` nonlinearity in the vertical integral is a second, smaller source of non-redundancy, but the linear-combination argument holds even in the fully linear case.

**The `IceBudgetCorrectionConfig` corrections are the same budget identity, just rearranged.** `tendency_of_total_water_path_due_to_advection` solves the moisture budget for a *flux* given the *state* change minus the other tendencies; `IceBudgetCorrectionConfig.__call__` solves for the *state* — `out[key] = x_in[key] + Δt·(source + sink + transport)` — given the previous state plus the tendencies. Same identity, opposite unknown. One nuance worth noting: the ice case isn't a clean single-channel overwrite the way advection is, because `constrain_budgets` clamps/redistributes the source/sink/transport channels first, so those touched tendency channels are only *partially* replaced (gradient still flows through them in the unclamped regions).

**On the wastefulness.** The sharp version of "throwing it away": the network's *raw* `tendency_of_total_water_path_due_to_advection` output receives exactly zero gradient from the advection loss term, since it's discarded before the loss is computed — so a dedicated advection head is dead weight for this particular supervision. That's exactly what a `delta`-based "corrector regularization" would recover: `delta = corrected − network_output` (this PR) is the handle, and a loss pulling `network_output` toward `corrected` routes gradient back into the otherwise-dead raw channel — for both advection and the ice mass/concentration channels.

fme/core/corrector/output.py
jpdunc23: Yes, sorry I missed this; I need to update my "implementation" skill so this doesn't happen.

fme/core/corrector/output.py
mcgibbon: > I remember when @William-gregory was developing FloeNet we had this tension of "derived variables that should be included in the optimization" and we considered adding the ability to "optimize derived outputs". Ultimately, we ended up using the corrector for simplicity (and, I guess, the tendency_of_total_water_path_due_to_advection precedent).

Yes, I was proposing this same system back then as now. We went with the corrector so that Will wouldn't have to make large code changes, despite the resulting code being less clear. We're dealing with that issue now.

The proposal is that we gain the ability to separately define the list of network output channels (that the output packer uses), and the list of optimized output channels, and add a layer which can derive optimized output channels from network output channels. There's some added complication though, because some corrections kind of need to happen before derivations happen, e.g. we only compute the TWP advection after clamping moisture terms and removing the global-mean bias.

> OK, though I want to note again that hfds does not fall under the category of "an output derived from network outputs"; It could be considered "an output in part derived from network inputs".

The way I look at it, the network produces "hfds_flux_residual" as a true output which is not directly optimized, and "hfds" is a derived output equal to `hfds_flux_residual + (flux-derived values)`.

Framing it this way raises the question of why don't we add values to the target dataset for hfds_flux_residual, derived from the residual you'd actually get if you predicted the other terms accurately? It also provides a pathway to explicitly set that target to 0, if you want to encourage the term to be small. I think this is a much more natural way to predict hfds, which I don't think cleanly falls into "we should optimize the outputs of the network before they're corrected".

Maybe this is complicated by something like it always being 0 in ocean-isolated training, but needing it to be nonzero during coupled training, which would raise a question of whether we should be optimizing the atmosphere to produce unbiased estimates of hfds more directly instead of going through the ocean having a correction term.

This also re-frames the question into, should hfds be derived in-corrector or should it be a post-corrector derived variable (or both, by making it in-corrector but futzing with the relative variable weights of hfds and hfds_flux_residual). That clarifies exactly the way in which hfds is different from "encourage corrections to be small". I find it makes it much easier to think about what we're doing and come up with good ways to improve prediction of hfds when we say "hfds" is a derived output equal to `hfds_flux_residual + (flux-derived values)`.

I want to put a break here because hfds is a very different situation from the ones that don't have a "residual" term (like twp tendency due to advection). Let's call them fully derived.

For fully derived outputs, I agree we shouldn't be wastefully predicting a channel for them. We've left it because in practice there aren't gradients for these outputs, the last-layer weights slowly decay to 0 due to the adam weight decay and it doesn't affect training.

If we care, the solution is we should be removing that output channel, not propagating gradients to improve an output we will never use. That's a true waste of computational power - the gradients then do affect the network and try to train it to produce something we won't use. From a multi-task learning perspective, I think it's an interesting research question to see if adding this as an additional predicted channel that gets optimized improves some model metric, but I don't have an issue in mind this would solve and without one, it seems like a waste of computational power. If you have a specific metric doing poorly to improve with this idea it could be worth pursuing, but I think you're currently pursuing the lower-hanging fruit for those metrics and finding good success.

BTW the post-optim derived variables generally speaking are ones for which we directly optimize each of the outputs already, and it's a diagnostic variable so we don't need to worry about its autoregressive impact on the model. One potential way to do this would be to pull all derived outputs into the Step, but to say that by default derived outputs have zero loss weight, but it might be worth keeping these types of derived variables separated.

fme/core/corrector/output.py
```diff
@@ -0,0 +1,68 @@
...
+    delta: TensorMapping = dataclasses.field(default_factory=dict)
+
+
+@dataclasses.dataclass
+class CorrectorOutput:
+    """The full result of applying a corrector to one step's generated data.
+
+    Parameters:
+        corrected: The adjusted generated data.
+        diagnostics: The corrector's declared correction deltas.
```
mcgibbon: Issue: This line shouldn't re-document the internals of CorrectorDiagnostics, which risks going out-of-date if that object changes.
```suggestion
        diagnostics: The corrector's diagnostic outputs.
```

fme/core/corrector/output.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Produces ``delta[name] = corrected[name] - input_snapshot[name]`` for each
+    ``name`` in ``touched_names``.
+
+    When more than one correction touches a field, ``corrected`` is the
+    corrector's exit value and ``input_snapshot`` its entry value, so the stored
+    delta is their cumulative net effect and ``input_snapshot == corrected -
+    delta`` stays exact.
+
+    The returned tensors are *not* detached from the autograd graph here.
```
mcgibbon: Question: So when only one correction touches a field, is that field excluded from these dicts? I would expect not.

However this is also just duplicate information, the args already state it. We can remove the paragraph.


```suggestion
    The returned tensors are *not* detached from the autograd graph here.
```

mcgibbon (approved):
Please incorporate the two suggested changes, otherwise LGTM.

PR #1322, opened 2026-06-26
Add keep_gradient_through_clamps STE for corrector clamps

mcgibbon:
The hard `torch.clamp` operations in the correctors run inside the training autograd path, and `clamp` has zero gradient in its saturated tail — so a cell the network pushes out of range gets no learning signal to come back; the projection silently absorbs the error. This adds a straight-through estimator (STE) that keeps the exact projected (clamped) value in the forward pass while letting gradient flow as identity in the backward pass, gated behind a new default-off `keep_gradient_through_clamps` flag on each corrector. Because forward values are unchanged when the flag is on, it is cleanly A/B-able against a baseline; only the gradient path differs, and default-off preserves existing behavior.

Validated by a same-seed A/B on the selected 4° residual baseline: with the flag on, training is stable to convergence (best_val_loss neutral) and the predicted near-zero precipitation distribution moves toward target (the dry-cell over-prediction is roughly halved across all four climate evals).

Changes:
- `fme.core.corrector.utils.replace_value_keep_gradient`: new STE helper (`x + (new_value - x).detach()` — forward is `new_value`, backward is identity).
- `fme.core.corrector.utils.force_positive` / `ForcePositive`: optional `keep_gradient` that applies the precip clamp via the STE.
- `fme.core.corrector.ocean.SeaIceFractionConfig` / `SeaIceFractionCorrection`: optional `keep_gradient` applying the STE to the 0-1 sea-ice-fraction clamp and the negative-ocean-fraction rebalance.
- `fme.core.corrector.atmosphere.AtmosphereCorrectorConfig` / `ocean.OceanCorrectorConfig`: new default-off `keep_gradient_through_clamps` field wiring the above.

- [x] Tests added

### DIFF
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -17,7 +17,7 @@ from fme.core.corrector.registry import (
     CorrectorConfigABC,
 )
 from fme.core.corrector.state import CorrectorState
-from fme.core.corrector.utils import ForcePositive
+from fme.core.corrector.utils import ForcePositive, replace_value_keep_gradient
 from fme.core.dataset_info import DatasetInfo
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.ocean_data import HasOceanDepthIntegral, OceanData
@@ -55,12 +55,17 @@ class SeaIceFractionConfig:
     remove_negative_ocean_fraction: bool = True
 
     def __call__(
-        self, gen_data: TensorMapping, input_data: TensorMapping
+        self,
+        gen_data: TensorMapping,
+        input_data: TensorMapping,
+        keep_gradient: bool = False,
     ) -> TensorDict:
         out = {**gen_data}
-        out[self.sea_ice_fraction_name] = torch.clamp(
-            out[self.sea_ice_fraction_name], min=0.0, max=1.0
-        )
+        sif = out[self.sea_ice_fraction_name]
+        clamped_sif = torch.clamp(sif, min=0.0, max=1.0)
+        if keep_gradient:
+            clamped_sif = replace_value_keep_gradient(sif, clamped_sif)
+        out[self.sea_ice_fraction_name] = clamped_sif
         if self.remove_negative_ocean_fraction:
             negative_ocean_fraction = (
                 1
@@ -68,7 +73,12 @@ class SeaIceFractionConfig:
                 - input_data[self.land_fraction_name]
             )
             negative_ocean_fraction = negative_ocean_fraction.clip(max=0)
-            out[self.sea_ice_fraction_name] += negative_ocean_fraction
+            rebalanced_sif = out[self.sea_ice_fraction_name] + negative_ocean_fraction
+            if keep_gradient:
+                rebalanced_sif = replace_value_keep_gradient(
+                    out[self.sea_ice_fraction_name], rebalanced_sif
+                )
+            out[self.sea_ice_fraction_name] = rebalanced_sif
         for name in self.zero_where_ice_free_names:
             out[name] = gen_data[name] * (out[self.sea_ice_fraction_name] > 0.0)
         return out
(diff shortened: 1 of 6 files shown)


fme/core/corrector/utils.py
```diff
@@ -6,11 +6,38 @@
...
+    instead of a zero gradient.
+
+    Forward: ``x + (new_value - x) = new_value`` (the exact projected value is
+    preserved). Backward: the detached term contributes zero, so the gradient is
+    the identity ``d(out)/dx = 1``.
+    """
+    return x + (new_value - x).detach()
+
+
+def force_positive(
```
jpdunc23: Nit: I think `force_positive` can be made private now?
```suggestion
def _force_positive(
```

fme/core/corrector/ocean.py
```diff
@@ -204,6 +221,11 @@ class OceanCorrectorConfig(CorrectorConfigABC):
     sea_ice_fraction_correction: SeaIceFractionConfig | None = None
     surface_energy_flux_correction: SurfaceEnergyFluxCorrectionConfig | None = None
     ocean_heat_content_correction: OceanHeatContentBudgetConfig | None = None
+    keep_gradient_through_clamps: bool = False
+    """If True, apply the corrector's hard clamps (``force_positive`` and the
+    ``sea_ice_fraction_correction`` bound/rebalance) with a straight-through
+    estimator: the forward value is still clamped, but gradient flows as if the
+    clamp had not happened, so out-of-range cells still get a learning signal."""
```
jpdunc23: This seems like an odd way to document `keep_gradient_through_clamps`. Would this appear in the docs?

Preexisting issue, but it seems the [`OceanCorrectorConfig` docs](https://ai2-climate-emulator.readthedocs.io/en/latest/api.html#fme.ace.OceanCorrectorConfig) are lacking. 

I can work on more complete documentation in a separate PR, but would you be willing to add a short, low-effort description for each parameter to this PR? At the level of `ocean_heat_content_correction: Optional configuration for an ocean heat content correction.`

fme/core/corrector/test_atmosphere.py
```diff
@@ -289,6 +289,32 @@ def test_force_positive():
...
+    x_plain = torch.tensor([-2.0, 1.0], requires_grad=True)
+    force_positive({"foo": x_plain}, ["foo"])["foo"].sum().backward()
+    torch.testing.assert_close(x_plain.grad, torch.tensor([0.0, 1.0]))
+
+    x_ste = torch.tensor([-2.0, 1.0], requires_grad=True)
+    fixed = force_positive({"foo": x_ste}, ["foo"], keep_gradient=True)
+    # forward value is still clamped
+    torch.testing.assert_close(fixed["foo"], torch.tensor([0.0, 1.0]))
+    fixed["foo"].sum().backward()
+    torch.testing.assert_close(x_ste.grad, torch.tensor([1.0, 1.0]))
```
jpdunc23: Maybe these should be in a dedicated `test_utils.py` module?

jpdunc23 (commented):
Looking good, I just have a couple comments I hope you can address re. the `OceanCorrectorConfig` docs and adding `test_utils.py`.

fme/core/corrector/test_atmosphere.py
mcgibbon: Agreed, we can move the test for force_positive when we do that. This is a pre-existing issue but it makes sense to address in this PR.

fme/core/corrector/utils.py
mcgibbon: Done — renamed to `_force_positive` and updated the one internal call site in `ForcePositive` (and the docstring references in the configs now point at the public `force_positive_names` field). Pushed in d87039b94.

fme/core/corrector/ocean.py
mcgibbon: Good call — replaced the attribute-docstring with a proper `Parameters:` block on `OceanCorrectorConfig` giving a short description for each field (`force_positive_names`, `sea_ice_fraction_correction`, `surface_energy_flux_correction`, `ocean_heat_content_correction`, `keep_gradient_through_clamps`), matching the `AtmosphereCorrectorConfig` style. Happy to leave the more complete docs sweep to your separate PR. Pushed in d87039b94.

fme/core/corrector/test_atmosphere.py
mcgibbon: Done — moved `test_force_positive` along with the new `test_replace_value_keep_gradient` and `test_force_positive_keep_gradient_passes_gradient_on_clamped_cells` into a new `fme/core/corrector/test_utils.py`. Pushed in d87039b94.

jpdunc23 (approved):
LGTM!

PR #1323, opened 2026-06-26
Add near-zero-fraction inference metric

mcgibbon:
Adds a default-off inference metric reporting, per variable, the area-weighted fraction of cells whose value is `<= eps` for the prediction, plus its difference from the target. `eps` is required to be **non-negative** (enforced in `__post_init__` when enabled): for precipitation (`PRATEsfc`) with a small `eps` this is the fraction of near-zero (dry / drizzle) cells — a scalar the existing RMSE/bias/histogram metrics do not surface directly, useful for diagnosing a model's dry-cell / drizzle behaviour against the target distribution. `eps = 0` is permitted, and is meaningful for corrector-clamped fields (e.g. precipitation clamped to a physical floor of zero), where it counts exactly-dry cells.

This is the metric used in the keep_gradient_through_clamps precipitation validation (companion to #1322); opening as a draft to decide whether the near-zero-fraction diagnostic is worth keeping in `main` as a general-purpose metric.

Changes:
- `fme.ace.aggregator.inference.near_zero_fraction.NearZeroFractionMetricConfig` / `NearZeroFractionAggregator`: new default-off metric. Configurable `variables` (required when `enabled` — raises in `__post_init__` otherwise), `eps` (must be `>= 0` when `enabled`), `per_variable_eps` (per-variable override of the global `eps`, each `>= 0`), and `name` (default `near_zero_fraction`). Log keys are `<name>/gen/<var>` and `<name>/gen_minus_target/<var>` (the standalone target fraction is computed internally for the difference but not logged).
- `include_maps` (default off): when set, additionally accumulates the per-cell time-mean at-or-below-`eps` fraction and logs 2D maps — a side-by-side generated/target comparison (`<name>/gen_target_map/<var>`) and the generated-minus-target error map (`<name>/error_map/<var>`), mirroring the time-mean aggregator's map convention. When no target is available only the generated map (`<name>/gen_map/<var>`) is logged. The maps are also exposed via `get_dataset` as `gen_map-<var>` / `target_map-<var>` / `error_map-<var>` DataArrays.
- `fme.ace.aggregator.inference.main`: register `near_zero_fraction` on `InferenceEvaluatorAggregatorConfig` (opt-in).

- [x] Tests added


fme/ace/aggregator/inference/dry_fraction.py
```diff
@@ -0,0 +1,118 @@
...
+from fme.core.typing_ import TensorMapping
+
+from .build_context import MetricBuildContext, maybe_filter
+from .data import InferenceBatchData, MetricBuildResult, SubAggregator
+
+AreaWeightedMean = Callable[..., torch.Tensor]
+
+
+@dataclasses.dataclass
+class DryFractionMetricConfig:
```
mcgibbon: Call this ZeroFractionMetricConfig, or not DryFraction. Update all documentation you've added accordingly.

fme/ace/aggregator/inference/dry_fraction.py
```diff
@@ -0,0 +1,118 @@
...
+    ``<= threshold``) for the prediction, the target, and their difference. For
+    precipitation (``PRATEsfc``) with the default threshold of 0 this is the
+    dry-cell fraction, which existing metrics (RMSE/bias/histogram) do not
+    surface directly — useful for diagnosing a model's dry-cell / drizzle
+    behaviour against the target distribution. Disabled by default.
+
+    Parameters:
+        variables: when set, only compute the metric for these variables
+            (typically ``["PRATEsfc"]``).
+        threshold: a cell counts as "dry" when its value is ``<= threshold``.
```
mcgibbon: Add per_variable_threshold, dict[str, float] = (field defaulting to dict) to be able to specify the threshold per-variable.

fme/ace/aggregator/inference/dry_fraction.py
```diff
@@ -0,0 +1,118 @@
...
+            Defaults to 0.0, i.e. exactly-zero for non-negative precipitation.
+        name: log prefix and wandb key prefix.
+        enabled: master toggle for the metric.
+        strict: raise if the metric can't be built.
+    """
+
+    variables: list[str] | None = None
```
mcgibbon: Default the list of variables to an empty list, the user should specify the variables they want. Raise an exception if enabled=True but no variables are given, in the post-init.

fme/ace/aggregator/inference/dry_fraction.py
```diff
@@ -0,0 +1,118 @@
...
+    threshold: float = 0.0
+    name: str = "dry_fraction"
```
mcgibbon:
```suggestion
    name: str = "zero_threshold_fraction"
```

mcgibbon:
Thanks — addressed in the latest commit:

- **Rename to `ZeroFraction`**: `dry_fraction.py` → `zero_fraction.py`, `DryFractionMetricConfig`/`DryFractionAggregator` → `ZeroFractionMetricConfig`/`ZeroFractionAggregator`, config field `dry_fraction` → `zero_fraction`, and docstrings updated. `name` default is now `"zero_threshold_fraction"`.
- **`per_variable_threshold: dict[str, float]`**: added (defaults to `{}`); overrides the global `threshold` for the named variables, with a test covering the override + fallback.
- **`variables` required when enabled**: default is now an empty list, and `__post_init__` raises if `enabled=True` with no variables given (test added).
- While renaming I also dropped the redundant metric-name component the aggregator was prepending to its log keys — the framework already prefixes with the metric name (`label=`), which was producing a doubled `…/dry_fraction/dry_fraction/…`. Keys are now `<metric>/gen/<var>` etc.

All aggregator + inference-main tests pass; pre-commit (ruff/mypy) clean.

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,141 @@
...
+            listed in ``per_variable_threshold``.
+        per_variable_threshold: optional per-variable thresholds overriding
+            ``threshold`` for the named variables.
+        name: log prefix and wandb key prefix.
+        enabled: master toggle for the metric.
+        strict: raise if the metric can't be built.
+    """
+
+    variables: list[str] = dataclasses.field(default_factory=list)
+    threshold: float = 0.0
```
mcgibbon: Question for reviewer: should we call this `eps`?

fme/ace/aggregator/inference/zero_fraction.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @torch.no_grad()
+    def get_logs(self, label: str) -> dict[str, float]:
+        gen_means = self._reduced_means(self._gen_sum, self._gen_count)
+        target_means = self._reduced_means(self._target_sum, self._target_count)
+        logs: dict[str, float] = {}
+        for name, value in gen_means.items():
+            logs[f"gen/{name}"] = value
+        for name, value in target_means.items():
+            logs[f"target/{name}"] = value
```
mcgibbon: Let's remove the target value, just keep gen/ and gen_minus_target/.

mcgibbon:
We should add 2D maps to be able to compare against AIMIP panel C11. https://arxiv.org/pdf/2605.06944

jpdunc23:
> We should add 2D maps to be able to compare against AIMIP panel C11. https://arxiv.org/pdf/2605.06944

I intend to include 2D maps of all correction `delta` (including "force-positive" corrections) as part of a follow-up to https://github.com/ai2cm/ace/pull/1321. I think this would meet your requirement?

fme/ace/aggregator/inference/main.py
```diff
@@ -203,6 +205,12 @@ class InferenceEvaluatorAggregatorConfig:
         ipo_index: Interdecadal Pacific Oscillation index metrics.
         trend: Per-grid-cell linear trend (slope vs. time) map metrics.
             Disabled by default.
+        zero_fraction: Area-weighted fraction of cells at or below a
```
jpdunc23: Given this is not limited to thresholds of 0, `zero_fraction` seems a poor choice of name. Also, why limit to `< threshold` an not more general `< threshold` or `> threshold`. With `> threshold` allowed then this metric would be also be useful for `sea_ice_fraction > 1` and for logging extremes.

jpdunc23 (commented):
I can see the more general usefulness of an "above or below a configurable threshold" metric, but in particular for "force-positive" or other clamping corrections I think this would be somewhat redundant with planned aggregator metrics for corrector deltas as a follow-up to https://github.com/ai2cm/ace/pull/1321

mcgibbon:
> > We should add 2D maps to be able to compare against AIMIP panel C11. https://arxiv.org/pdf/2605.06944
> 
> I intend to include 2D maps of all correction `delta` (including "force-positive" corrections) as part of a follow-up to #1321. I think this would meet your requirement?

In part, but force-positive isn't the only correction that happens to some fields, and the threshold can be configured to something slightly greater than zero as well.

fme/ace/aggregator/inference/main.py
mcgibbon: > Given this is not limited to thresholds of 0, zero_fraction seems a poor choice of name.

Agreed, let's rename to NearZeroFraction and near_zero_fraction. Let's also enforce that the threshold must be > 0.

> Also, why limit to < threshold an not more general < threshold or > threshold.

Because this would significantly complicate the configuration, and I don't have a present use case for that metric. I'd avoid the YAGNI since it's clearer how to write that feature when you're in the context where you're using it. Sea ice fraction > 1 is an interesting one, but as you say it probably lives better in the corrector aggregator.

docs/default-aggregator-config.yaml
```diff
@@ -105,5 +105,13 @@ aggregator:
     name: trend
     enabled: false
     strict: false
+  zero_fraction:
+    variables: []
+    threshold: 0.0
+    per_variable_threshold: {}
+    name: zero_threshold_fraction
```
mcgibbon: Rename to near_zero_fraction.

docs/default-aggregator-config.yaml
```diff
@@ -105,5 +105,13 @@ aggregator:
     name: trend
     enabled: false
     strict: false
+  zero_fraction:
+    variables: []
+    threshold: 0.0
+    per_variable_threshold: {}
```
mcgibbon: Let's keep threshold as-is, no need to rename to `eps`.

fme/ace/aggregator/inference/main.py
mcgibbon: Addressed in fc7bd2e1c:

- Renamed `ZeroFraction` → `NearZeroFraction` throughout: `zero_fraction.py` → `near_zero_fraction.py`, `ZeroFractionMetricConfig`/`ZeroFractionAggregator` → `NearZeroFraction*`, the config field `zero_fraction` → `near_zero_fraction`, and the `name` default → `near_zero_fraction`. Docs (rst + the default-aggregator config regression yaml) updated to match.
- `__post_init__` now enforces `threshold > 0` (and that every `per_variable_threshold` value is `> 0`) when `enabled`, with a test. The metric is explicitly for *near*-zero cells; an exact-zero indicator is left as a degenerate case for a dedicated metric, and the generalized `< or > threshold` form is deliberately omitted (YAGNI) per the discussion above.
- Kept `threshold` as the field name (no `eps` rename).

pre-commit (ruff/mypy) clean; aggregator + inference-main + default-config tests pass.

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def __post_init__(self):
+        if not self.enabled:
+            return
+        if not self.variables:
+            raise ValueError(
+                "NearZeroFractionMetricConfig is enabled but no variables were "
+                "given; specify the variables to compute the metric for."
+            )
+        if self.threshold <= 0:
```
jpdunc23: So the default value `threshold: float = 0.0` leads to a `ValueError`? Is this intentional?

fme/ace/aggregator/inference/near_zero_fraction.py
jpdunc23: Yes, I think `eps` is more clear for the one-sided `<` nature of this aggregator, whereas `threshold` is more general and two-sided.

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+
+    Parameters:
+        variables: variables to compute the metric for (e.g. ``["PRATEsfc"]``).
+            Must be non-empty when ``enabled``.
+        threshold: a cell counts toward the fraction when its value is
+            ``<= threshold``. Must be ``> 0`` when ``enabled``. Applied to any
+            variable not listed in ``per_variable_threshold``.
+        per_variable_threshold: optional per-variable thresholds overriding
+            ``threshold`` for the named variables. Each must be ``> 0`` when
+            ``enabled``.
```
jpdunc23: How many fields do you ever expect to need this for? If only a small number as I expect (e.g., `PRATEsfc`, `total_frozen_precipitation_rate`, `sea_ice_fraction`) then it will simplify things a lot to remove `variables` and `threshold` and just use `per_variable_threshold` (prefer `per_variable_eps`) as the sole way to configure this. 

How do you think the value of `eps` should be chosen? Is it more of a "numerical noise" threshold or supposed to be physically meaningful?

fme/ace/aggregator/inference/near_zero_fraction.py
jpdunc23: I see from the tests that this is intentional, but this is a bit awkward IMO.

What about `eps: float | None = None` with a "threshold must be set when enabled" message.

Or a sensible default like `eps: float = 1e-6`?

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+    """Accumulates the area-weighted fraction of cells with value ``<= threshold``.
+
+    For each variable it reports the predicted fraction and, when a target is
+    available, its difference from the target (gen - target), averaged over all
+    sample/time entries seen. The initial-condition timestep of the first batch
+    (``i_time_start == 0``) is excluded, since it is prescribed rather than
+    predicted (as in :class:`TimeMeanAggregator`). The metric name is supplied
+    by the caller as the ``label`` prefix in :meth:`get_logs`.
+
+    When ``include_maps`` is set it additionally accumulates the per-cell
```
jpdunc23, line 107: Do you have an example of a run with `include_maps = True`? I'm curious to see what the `PRATEsfc` maps look like.

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+            horizontal_dims=ctx.horizontal_coordinates.dims,
+        )
+        return MetricBuildResult(aggregator=maybe_filter(agg, self.variables))
+
+
+class NearZeroFractionAggregator:
+    """Accumulates the area-weighted fraction of cells with value ``<= threshold``.
+
+    For each variable it reports the predicted fraction and, when a target is
+    available, its difference from the target (gen - target), averaged over all
```
jpdunc23: "its difference from the target fraction"

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+        time_slice = slice(1, None) if ignore_initial else slice(None)
+        for name, tensor in data.items():
+            # Drop the initial-condition timestep of the first batch: it is
+            # ground-truth-initialized, not predicted, so counting it would
+            # pull the generated fraction toward the target (mirrors
+            # TimeMeanAggregator's ignore_initial handling).
+            tensor = tensor[:, time_slice]
+            # NaN <= threshold is False, so NaN cells count as "not below"
+            # rather than propagating NaN; fine for fields like PRATEsfc that
+            # are never NaN, but note it for NaN-filled (missing) variables.
```
jpdunc23: Is this comment needed? It seems to imply the handling is incomplete for fields with NaNs (e.g., `sea_ice_fraction`), though I think the usual `GriddedOperations` masked area-weighted mean should properly handle ocean fields. Is that somehow not the case?

fme/ace/aggregator/inference/near_zero_fraction.py
```diff
@@ -0,0 +1,303 @@
...
+                self._target_map_sum,
+                self._target_map_count,
+                ignore_initial,
+            )
+
+    def _reduced_means(
+        self, sums: dict[str, torch.Tensor], counts: dict[str, int]
+    ) -> dict[str, float]:
+        means: dict[str, float] = {}
+        for name, total in sums.items():
```
jpdunc23: For consistency with `reduced.py` / `time_mean.py`, sort the keys:

```suggestion
        for name in sorted(sums):
```

I don't think this is needed for correctness of the distributed reduction, but seems worth preserving / keeping consistent.

jpdunc23 (approved):
I'm strongly in favor of renaming `threshold` to `eps`. This commits to the "near-zero" specificity of this feature, and the implicit use-case of evaluating drizzling bias. If a more general `threshold` aggregator is needed we can add it separately rather than try to glom onto this aggregator.

I also think the default `threshold: float = 0.0` is awkward and suggest you revise to `threshold: float | None = None`. Better yet, simplify somewhat by removing `variables` and `threshold` and keep only `per_variable_threshold` which can default to an empty dict and throw an error if enabled but empty.

These are somewhat aesthetic / quality-of-life concerns, so I won't block merge or need to re-review (but happy to if you want). Approved.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: I do not

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: I actually don't see why 0.0 isn't valid. I think Claude interpreted me saying the values should be "strictly positive" where I was being loose with words and meant "non-negative".

Let's update to `eps: float = 0.0`, and also rename to `per_variable_eps`.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: I think it will more commonly be used as a per-variable threshold, but I wanted it to be easy to set this up with a copy-paste of the list of force-non-negative fields in the corrector without having to look up physical thresholds for each field. So at least right now, I think it's important to have a variables: list[str] configuration available.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Done in b0a22e8cf — `eps` is now allowed to be `0` (guard is `>= 0`, not `> 0`). Kept it a plain `float = 0.0` rather than `float | None`: `0` is a meaningful default here (the exact-dry-cell fraction for corrector-clamped fields), and `enabled` already gates whether the metric runs, so there is no ambiguous "set but unused" state that `None` would need to disambiguate.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Renamed `threshold` → `eps` (and `per_variable_threshold` → `per_variable_eps`) throughout in b0a22e8cf, committing to the one-sided `<= eps` near-zero semantics.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Renamed `per_variable_threshold` → `per_variable_eps` in b0a22e8cf. Per Jeremy's note above, keeping the `variables` list alongside it so the common case — copy-pasting the corrector's force-non-negative field list, all sharing one `eps` — does not require looking up a per-field value for each. On choosing `eps`: it is a near-zero / numerical-floor threshold (drizzle cutoff, or a corrector's clamp floor), not a physically-derived constant, so `0` is the natural default for clamped fields.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Done in b0a22e8cf.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Tightened the comment in b0a22e8cf — dropped the "incomplete for missing variables" phrasing. The behavior is intentional: `NaN <= eps` is `False`, so NaN cells count as not-below rather than propagating NaN; this metric targets never-NaN fields like `PRATEsfc`.

fme/ace/aggregator/inference/near_zero_fraction.py
mcgibbon: Claude: Done in b0a22e8cf — both `_reduced_means` and `_reduced_maps` now iterate `sorted(sums)`.

PR #1325, opened 2026-06-26
Fix SeparateICZarrWriterAdapter mislabeling time calendar as julian

### LINKED ISSUE #1324: SeparateICZarrWriterAdapter mislabels the time coordinate calendar as "julian" for non-julian (e.g. noleap) data
## Summary
When the zarr data writer is used with `separate_ensemble_members: true` (`SeparateICZarrWriterAdapter`), the output `time` coordinate is always tagged `calendar = "julian"`, regardless of the calendar of the input data. The numeric time values are encoded correctly using the *true* data calendar, so the file is internally inconsistent: values say noleap, attribute says julian. Any reader that honors the CF `calendar` attribute (e.g. `xr.open_zarr` with default decoding) then misinterprets every timestamp. For noleap data this introduces a growing offset (≈ one day per 4 years since the 1970 epoch); for piControl runs that start near year 1 the offset is more than a year.

This is distinct from #1315 (which is about *which* IC timestamp is assigned to *which* output). Here the assigned values are right but the calendar label is wrong.

## Expected vs actual
- **Expected:** `output.time` decodes to the same dates as the source data; the stored `calendar` attribute equals the source calendar (e.g. `noleap`).
- **Actual:** `time` values are encoded with the source calendar but the attribute is hardcoded to the default `"julian"`, so decoding with the stated attribute yields wrong dates.

## Mechanism
`fme/ace/inference/data_writer/zarr.py`, `SeparateICZarrWriterAdapter._initialize_writers`: the `time` values are produced with `cftime.date2num(..., calendar=first_batch_time.dt.calendar)` (correct), stored as numeric int64, then handed to `ZarrWriter` **without** a `time_calendar` argument. `ZarrWriter`/`_initialize_zarr` defaults `time_calendar="julian"` (`fme/core/writer.py:229`, used at lines 134-135) and writes that as the coordinate's `calendar` attribute. Because the stored `time` values are numeric (not `cftime.datetime`) by the time they reach `_initialize_zarr`, they are written as-is (noleap-encoded) while the attribute is set to the default julian.

The non-separate `ZarrWriterAdapter` is unaffected because it passes `time_calendar=None` and stores datetimes as nondim coords (`init_time`/`valid_time`) with explicit calendar attributes; its dim coord `time` is lead-time (a timedelta). The netCDF writers (`raw.py`, `monthly.py`) infer the calendar from the ICs and use it for both label and values, so they are self-consistent.

## Reproduction
Config: `configs/experiments/2026-06-16-ace2s-land-feedback-inference/frameworkA-cm4-rs0.yaml` (branch `exp/ace2s-land-feedback-inference`); IC and forcing both `gs://.../2025-03-21-CM4-piControl-atmosphere-land-1deg-8layer-200yr.zarr` (noleap), `separate_ensemble_members: true`.

Minimal standalone repro driving `SeparateICZarrWriterAdapter` with a noleap time axis:
```
stored 'time' calendar attribute : julian          <- WRONG
decoded with STATED attr         : 0002-05-08 ...   <- offset >1 yr
decoded with TRUE forcing cal    : 0001-01-01 ...   <- correct
```
`xr.open_zarr(..., decode_times=False)` shows the stored values are the noleap encoding; default decoding (honoring the `julian` attribute) shifts every date by >1 year. Switching the source calendar to `julian` makes the output accidentally correct, confirming the bug only bites non-julian calendars.

## Suggested fix
Pass the data calendar through to the writer in `_initialize_writers`, e.g. `ZarrWriter(..., time_calendar=first_batch_time.dt.calendar)`. (Values are already encoded with matching units, so only the attribute needs correcting.) Audit other `ZarrWriter` construction sites for the same `time_calendar` default footgun, and consider making `time_calendar` required when a datetime `time` coord is supplied rather than silently defaulting to `"julian"`.

## Test
Add a regression test next to `fme/ace/inference/data_writer/test_zarr.py`: write through `SeparateICZarrWriterAdapter` (and `ZarrWriterAdapter`) with a `noleap` time axis and assert the stored `time.attrs["calendar"] == "noleap"` and that decoded dates round-trip to the source. A julian-only test would pass today, which is why this slipped through.

## Impact / severity
- **Silent**: no error; values are physically correct, so only content-vs-label verification exposes it.
- Corrupts date-based analysis/verification of any non-julian (noleap, 360_day, proleptic_gregorian) zarr inference output written with `separate_ensemble_members: true`.
- **Recoverable**: existing affected files can be relabeled in place (overwrite the `time` `calendar` attribute with the true calendar; values are unchanged) — no need to rerun inference.

brianhenn:
The zarr data writer used with `separate_ensemble_members: true` (`SeparateICZarrWriterAdapter`) encoded the `time` coordinate values using the source calendar but constructed the `ZarrWriter` without a `time_calendar` argument, so the coordinate was always tagged with the default `"julian"` calendar. For non-julian data (e.g. the noleap CM4 piControl dataset) this silently mislabeled every timestamp, so any reader that honors the CF `calendar` attribute misinterprets the dates (a >1-year offset for piControl data starting near year 1). This PR passes the source calendar through to the writer so the label matches the already-correct encoded values.

Changes:
- `fme.ace.inference.data_writer.zarr.SeparateICZarrWriterAdapter._initialize_writers`: pass `time_calendar=first_batch_time.dt.calendar` when constructing each `ZarrWriter`.
- `fme.ace.inference.data_writer.test_zarr.test_separate_ic_writer_preserves_time_calendar`: new regression test, parametrized over julian/proleptic_gregorian/noleap, asserting the stored `time` calendar attribute matches the source and decoded times round-trip. Fails on noleap/proleptic_gregorian before the fix.

- [x] Tests added

Resolves #1324

### DIFF
--- a/fme/ace/inference/data_writer/test_zarr.py
+++ b/fme/ace/inference/data_writer/test_zarr.py
@@ -127,6 +127,66 @@ def test_zarr_adapter_can_overwrite(tmpdir, writer_cls):
     adapter.append_batch(data, time)
 
 
+@pytest.mark.parametrize("calendar", ["julian", "proleptic_gregorian", "noleap"])
+def test_separate_ic_writer_preserves_time_calendar(tmpdir, calendar):
+    """A downstream reader must decode the written ``time`` to the source dates.
+
+    Regression test: previously ``SeparateICZarrWriterAdapter`` encoded the time
+    values with the source calendar but did not pass ``time_calendar`` to the
+    ``ZarrWriter``, so the coordinate was tagged with the default ``"julian"``
+    calendar. For non-julian data the values then no longer matched the label, so
+    decoding shifted every timestamp. The shift comes from leap days accumulated
+    between the 1970 encoding epoch and the data dates, so it appears even when the
+    coordinate itself does not span a leap day (here ~12 days for 2020).
+    """
+    n_timesteps = 2
+    data = {"foo": torch.zeros((1, n_timesteps, 2, 2))}
+    timestep = datetime.timedelta(days=1)
+    initial_condition_times = get_initial_condition_times(
+        (2019, 12, 31, 0, 0, 0), calendar, n_initial_conditions=1
+    )
+    # Daily batch times so the timestep-derived lead times reproduce them exactly.
+    batch_time = xr.DataArray(
+        xr.date_range(
+            "2020-01-01",
+            freq="1D",
+            periods=n_timesteps,
+            calendar=calendar,
+            use_cftime=True,
+        ).values[None, :],
+        dims=("sample", "time"),
+    )
+    adapter = SeparateICZarrWriterAdapter(
+        path=str(tmpdir / "test.zarr"),
+        dims=("time", "lat", "lon"),
+        data_coords=ensure_numpy_coords(
+            {
+                "lat": xr.DataArray([0, 1], dims=["lat"]),
+                "lon": xr.DataArray([0, 1], dims=["lon"]),
+            }
+        ),
+        timestep=timestep,
+        n_timesteps=n_timesteps,
+        initial_condition_times=initial_condition_times,
+    )
+    adapter.append_batch(data, batch_time)
+    adapter.finalize()
+
+    output = str(tmpdir / "test_ic0000.zarr")
+    # A downstream reader uses xarray's default decoding, which honors the stored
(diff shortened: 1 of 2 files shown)


Arcomano1234 (approved):
LGTM

PR #1326, opened 2026-06-26
HEALPix UNet code cleanup

yikwill:
Several HEALPix settings were duplicated across YAML configs, mutated at build time, or lived on monolithic dataclasses where variant-specific fields were silently ignored. This PR makes configs easier to load from YAML, reason about, and extend. It refactors HEALPix UNet configuration and build path to address feedback from [a previous HEALPix PR](https://github.com/ai2cm/ace/pull/1156), namely

> ...generally speaking it would be nice to move "build" type arguments out of the dataclass, and instead as arguments to `.build`. Specifically, the number of input and output channels comes to mind. Anything that you wouldn't want the user to configure in a yaml file, but should instead be determined at build time from other variables in the runtime.
> 
> The fact that you have to `.replace` attributes of something we've specified in a yaml is a bit scary in this respect, those things should probably not be configurable rather than being silently ignored.

Additionally, block configs are split into dacite discriminated unions and runtime/build-time concerns are separated.

Changes:
Build-time vs config-time separation
- Channel counts (`input_channels` / `output_channels`) have been removed from `UNetEncoderConfig`, `UNetDecoderConfig`, and conv/downsample/upsample block configs. Insteady, they are now passed explicitly to `.build()` from `HEALPixUNet` / `HEALPixUNetBuilder`.
- Config objects are no longer mutated (e.g. via `.replace`) when constructing the model.

Shared build context
- Added `HEALPixBuildContext` and `HEALPixLayerBuildContext` for model-level settings (`hpx_padding_mode`, `nside` per level). They are configured once on `HEALPixUNet` / `HEALPixUNetBuilder`, then the encoder, decoder, and blocks receive per-level views via `ctx.layer(level)`.

Discriminated union block configs
- Monolithic block configs have been split into one dataclass per `block_type`, each with a `Literal[...]` discriminator for dacite union resolution. For example, the union alias `DownSamplingBlockConfig` has variants `MaxPoolDownsamplingBlockConfig`, `AvgPoolDownsamplingBlockConfig`, and `DealiasedDownsampleBlockConfig`. Each variant exposes only fields relevant to that block type.

Other
- `enable_nhwc` option removed. Memory format of inputs is already hard coded.
- Replace `th` calls with `torch` for clarity.

- [ ] Tests added

### DIFF
--- a/fme/ace/models/healpix/healpix_activations.py
+++ b/fme/ace/models/healpix/healpix_activations.py
@@ -19,7 +19,7 @@
 
 import dataclasses
 
-import torch as th
+import torch
 import torch.nn as nn
 
 
@@ -30,11 +30,9 @@ class CappedGELUConfig:
 
     Parameters:
         cap_value: Cap value for the GELU function, default is 10.
-        enable_nhwc: Flag to enable NHWC data format, default is False.
     """
 
     cap_value: int = 10
-    enable_nhwc: bool = False
 
     def build(self) -> nn.Module:
         """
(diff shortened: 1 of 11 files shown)


mcgibbon (approved):
LGTM

PR #1327, opened 2026-06-29
Give FileWriter a runtime-params object instead of its config

mcgibbon:
Follow-on to #1314. That PR moves writer construction into the config builders; this one takes the next, more debatable step: instead of handing the runtime `FileWriter` its full `FileWriterConfig`, it builds a small object holding only the values the writer actually uses at runtime and passes that in.

The motivation is to decouple the runtime writer from the config that builds it. Today `FileWriter` is given the whole `FileWriterConfig` and reaches into a handful of its fields; for example, whether to subselect the horizontal domain is really `bool(lat_extent or lon_extent)`, but from the writer's signature you can't tell which config fields it depends on. With a dedicated params object the writer's inputs are explicit, the config does the config-to-attribute translation once when it builds the writer, and the writer no longer depends on the config class.

This is split out from #1314 on purpose so the question of whether these params objects are worth the extra boilerplate can be discussed on its own. It is behavior-preserving.

Changes:
- `fme.ace.inference.data_writer.file_writer.FileWriterParams`: a new frozen dataclass holding only the values `FileWriter` uses at runtime (`label`, `names`, `subselect_horizontal`, `lat_slice`, `lon_slice`, `time_selection`). Frozen because these are read-only runtime inputs the writer never mutates.
- `fme.ace.inference.data_writer.file_writer.FileWriterConfig._build_writer_params`: builds the params object; the config's `build` now passes it to `FileWriter` instead of passing `self`.
- `fme.ace.inference.data_writer.file_writer.FileWriter`: takes `FileWriterParams` and reads its values, rather than taking and reading `FileWriterConfig`.
- Tests: build the `FileWriter` via `FileWriterConfig._build_writer_params()`.

Open questions from review (see #1314 discussion) to settle here:
- Whether `FileWriterParams` should be private.
- Whether `subselect_horizontal` is better expressed as a property on the config.

- [x] Tests added (existing `FileWriter` test builds through the new params helper)

### DIFF
--- a/docs/default-aggregator-config.yaml
+++ b/docs/default-aggregator-config.yaml
@@ -105,5 +105,13 @@ aggregator:
     name: trend
     enabled: false
     strict: false
+  near_zero_fraction:
+    variables: []
+    eps: 0.0
+    per_variable_eps: {}
+    name: near_zero_fraction
+    include_maps: false
+    enabled: false
+    strict: true
   monthly_reference_data: null
   time_mean_reference_data: null
--- a/docs/evaluator_config.rst
+++ b/docs/evaluator_config.rst
@@ -157,6 +157,9 @@ in the default aggregator configuration above.
 .. autoclass:: fme.ace.EnsembleMetricConfig
    :noindex:
 
+.. autoclass:: fme.ace.NearZeroFractionMetricConfig
+   :noindex:
+
 .. autoclass:: fme.ace.StepperOverrideConfig
    :show-inheritance:
    :noindex:
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -12,6 +12,7 @@ from fme.ace.aggregator.inference.main import (
     LegacyFlagInferenceEvaluatorAggregatorConfig,
     StepMeanEntry,
 )
+from fme.ace.aggregator.inference.near_zero_fraction import NearZeroFractionMetricConfig
 from fme.ace.aggregator.inference.reduced import MeanMetricConfig
 from fme.ace.aggregator.inference.seasonal import SeasonalMetricConfig
 from fme.ace.aggregator.inference.spectrum import PowerSpectrumMetricConfig
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -126,6 +126,12 @@ from fme.core.step import (
 )
 from fme.core.step.multi_call import MultiCallConfig
 from fme.core.typing_ import Slice
+from fme.core.var_masking import (
+    BernoulliMaskingConfig,
+    MaskingGroupConfig,
+    UniformMaskingConfig,
+    VariableMaskingConfig,
+)
 
 from . import step
 from .inference.inference import get_initial_condition
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -34,6 +34,7 @@ from .enso.dynamic_index import EnsoIndexMetricConfig
 from .enso.enso_coefficient import EnsoCoefficientMetricConfig
 from .histogram import HistogramMetricConfig
 from .ipo.ipo_index import MIN_YEARS_FOR_FILTERED_TPI, IpoIndexMetricConfig
+from .near_zero_fraction import NearZeroFractionMetricConfig
 from .reduced import MeanMetricConfig, SingleTargetMeanAggregator
 from .seasonal import SeasonalMetricConfig
 from .spectrum import PowerSpectrumMetricConfig, SphericalPowerSpectrumAggregator
@@ -65,6 +66,7 @@ MetricConfig = (
     | EnsembleMetricConfig
     | IpoIndexMetricConfig
     | TrendMetricConfig
+    | NearZeroFractionMetricConfig
 )
 
 
@@ -203,6 +205,12 @@ class InferenceEvaluatorAggregatorConfig:
         ipo_index: Interdecadal Pacific Oscillation index metrics.
         trend: Per-grid-cell linear trend (slope vs. time) map metrics.
             Disabled by default.
+        near_zero_fraction: Area-weighted fraction of cells at or below a
+            small non-negative ``eps`` per variable, reported for prediction and
+            its difference from the target. Optionally (``include_maps``) also
+            logs side-by-side generated/target maps of the per-cell
+            at-or-below-``eps`` fraction and the error map. Disabled by
+            default.
         monthly_reference_data: Path to monthly reference data to compare against.
         time_mean_reference_data: Path to reference time means to compare against.
     """
@@ -252,6 +260,9 @@ class InferenceEvaluatorAggregatorConfig:
         default_factory=IpoIndexMetricConfig
     )
     trend: TrendMetricConfig = dataclasses.field(default_factory=TrendMetricConfig)
+    near_zero_fraction: NearZeroFractionMetricConfig = dataclasses.field(
+        default_factory=NearZeroFractionMetricConfig
+    )
     monthly_reference_data: str | None = None
     time_mean_reference_data: str | None = None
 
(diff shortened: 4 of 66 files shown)


frodre:
Repost of your previous discussion: https://github.com/ai2cm/ace/pull/1314#issuecomment-4836289617

> If a built component stores a configuration instead of locally re-storing all the information that could be useful, it should be able to look at that configuration. I do think the items that are publicly available from configurations should be shallow/imutable features of the config, maybe?

Going into this, I'll say I think I should separate this part of the rule that leads to Params classes into its own PR so we can treat the question separately. But I'll argue why this change makes sense. Also I like the idea of saying these params classes should be frozen/immutable.

----- 

Traditionally the view I've held is that the config class and the class it's building are really just one class, and should be tightly coupled. The way this PR goes decouples them, so that the built class no longer depends on the config class, avoiding a circular dependency.

Having seen the change for this specific case, I do think it's better. `FileWriter` depends on `subselect_horizontal=bool(self.lat_extent or self.lon_extent)`, and does not directly use `lat_extent` or `lon_extent`. When the class takes in the full config, you can't tell that's the case. Generally speaking the pattern of "use config in the course of building the objects that go into the class init" is not uncommon.

It simplifies the init logic in cases where inits currently translate config values into internally-stored values, moving that to the builder layer instead. It avoids this kind of default pattern we fall into where we always pass the Config to the init, without thinking about what the object's actual API is.

Having the config build exactly the frozen Params the class needs and pass it to init is simple. The class no longer does configuration-to-attribute parsing in its init, and we know the Params won't be mutated. And we no longer have a circular dependency - it gives us the freedom to define multiple configs for the same object if we want, or to move our configs to different files if we want.

An issue is when passing Config, it's difficult to cleanly define what behavior is or isn't allowed. It's easy to say you can only pass a single-level frozen dataclass of basic/configurable types. It's harder to say you're allowed to access the basic types on the config but not any of its methods except for property methods - there are cases where we use `config.__call__()` for example. And should we allow property methods in that case? Or should the object's init handle that conversion? Do we allow methods to be called but not builder methods? What constitutes a builder method?

By contrast the choice here makes it unambiguous how the code should be written, which seems to give better AI results.

Now, one thing I'll say is I would only suggest this kind of rule or refactor because we're using AI-based coding now. This way of writing it is significantly safer and more explicit, making it harder to write bad code. I expect it to make AI-based code easier to review, and harder for it to hide bad patterns. It also requires more boilerplate, enough that I wouldn't typically ask people to write these Params classes, but I'm not worried about AI writing these.

In the long run this is also kind of interesting because it opens up the path to have multiple ways to configure and build our objects. We could define a heavily simplified config as an alternative way to build the same object hierarchy as our full configs. This is only possible if the config and Params are decoupled.

frodre (commented):
I think this may need a rebase and force push.  Some of the changes are the same as what took place in #1314

frodre:
I'm mainly coming at this from whether or not we need to have a *rule* and larger refactor around this style of usage.  Frozen configurations seems like a genuine win for bad use prevention.  As I said, I'm not convinced we need a repo wide rule around this since I don't think there's been any egregious violation of using config features.  I'm not necessarily opposed to the FileParams implementation here, but I want to push on what's actually occurring with this change and how well it achieves certain points.

Here are some responses to the points you've covered:

> Traditionally the view I've held is that the config class and the class it's building are really just one class, and should be tightly coupled.

I've also viewed them as tightly coupled.  Translate information from our external configuration into the runtime objects.  

> Having seen the change for this specific case, I do think it's better. FileWriter depends on subselect_horizontal=bool(self.lat_extent or self.lon_extent), and does not directly use lat_extent or lon_extent. When the class takes in the full config, you can't tell that's the case. 

Agree that the exact intentions for each config param is obfuscated by handing along the whole configuration, but that was a decision from the idea that it's a nice simplification tradeoff when using more than a few parameters from the configuration.  Passing along a parameters object would do the same obfuscation but it does highlight the `subselect_horizontal` usage and exact used params.  I'd argue you could highlight it similarly in a configuration object.  However, the tight coupling of the config, using that config, and seeing how it's used within build maintains pretty easy local traceability IMO.  Maybe this writer construction is the most complicated build case? Perhaps it means that the configuration class needs some design adjustments.

> It simplifies the init logic in cases where inits currently translate config values into internally-stored values, moving that to the builder layer instead. It avoids this kind of default pattern we fall into where we always pass the Config to the init, without thinking about what the object's actual API is.

Agree, it does have a more strict enforcement of what we can hand off to the init via the params.  Again, could probably be done with better config class design instead of additional layer of boilerplate.  

> Having the config build exactly the frozen Params the class needs and pass it to init is simple. The class no longer does configuration-to-attribute parsing in its init, and we know the Params won't be mutated. 

It's just moved to param class parsing for init, which is functionally the same except there's now a layer where you're _encouraged_ to derive parameters from any intermediate params instead of doing so in the init.  Params could become just as complicated, but the idea is that these are from scratch so the review enforcement would be simpler.  We could similarly simplify current configuration patterns with rules as they come up (more difficult and long term path).

> And we no longer have a circular dependency - it gives us the freedom to define multiple configs for the same object if we want, or to move our configs to different files if we want.

I'm not sure if the circular dependency is an issue given that they are a part  of the "same class" (tightly coupled) in the same file.  Also, without an explict use case in mind, having multiple configuration flavors for the same thing sounds like it would reduce clarity.  We've tended to lean for having only one pathway to achieve a certain outcome when possible.

> An issue is when passing Config, it's difficult to cleanly define what behavior is or isn't allowed. It's easy to say you can only pass a single-level frozen dataclass of basic/configurable types. It's harder to say you're allowed to access the basic types on the config but not any of its methods except for property methods - there are cases where we use config.__call__() for example. And should we allow property methods in that case? Or should the object's init handle that conversion? Do we allow methods to be called but not builder methods? What constitutes a builder method?

I don't know that we should make a firm rule about how the config attributes/properties/methods are being used as long as it's done in an easily traceable, understandable, and consistent way.  I think the defined builder responsibilities is a good move in that direction.  There's nothing stopping someone from designing a more complicated parameter class, the same way some of our configurations have become more complicated over time.  It's really about the rules/guidelines we set, which are enforced in review.  Those could be done more strictly for configurations, it's just that they are already in a messier state compared to the fresh parameter class in this PR.

> By contrast the choice here makes it unambiguous how the code should be written, which seems to give better AI results.

At the expense of another layer for us to consider.  It is a clear usage, but I don't know if an AI has any issues with the current setup other than there being inconsistency in our current config usage in different parts of the codebase.  It's a hard  thing to measure.

> Now, one thing I'll say is I would only suggest this kind of rule or refactor because we're using AI-based coding now. This way of writing it is significantly safer and more explicit, making it harder to write bad code. I expect it to make AI-based code easier to review, and harder for it to hide bad patterns. It also requires more boilerplate, enough that I wouldn't typically ask people to write these Params classes, but I'm not worried about AI writing these.

I can see this.  Code is now cheap, especially easily verifiable and definable outcomes.  This is part of a larger discussion about how much we are willing to hand off the ease of our understanding and convenience in light of not actually doing most of the code exploration and building ourselves.

> In the long run this is also kind of interesting because it opens up the path to have multiple ways to configure and build our objects. We could define a heavily simplified config as an alternative way to build the same object hierarchy as our full configs. This is only possible if the config and Params are decoupled.

If we have an essential/clear win use case, then I'd be more onboard with this benefit.

mcgibbon:
Thanks, very helpful feedback! I posted my comment reply in #1312, I think we should move discussion about the "rule" itself there. Given that, can you review this particular PR in a context of "does this refactor improve the code or make it worse", with zero weight on it following a particular rule? It wasn't clear to me whether you felt this PR itself is a positive change, though perhaps you were just holding off until I merge main (which I'm about to do).

frodre (approved):
Thanks for updating.  LGTM

PR #1328, opened 2026-06-30
Carry corrector correction deltas on a new StepOutput

jpdunc23:
Two layered parts. First, the corrector contract reopens: `CorrectorABC.__call__` returns a `CorrectorOutput` (corrected data, a `CorrectorDiagnostics` carrying the per-variable correction `delta`, and the passed-through corrector state) instead of a `(corrected, corrector_state)` tuple. Each `Correction.__call__` now returns **only the fields it modified** (today they return the full `gen_data`); the `CorrectionSequence` dict-updates `gen_data` with that subset, accumulates the modified-name set from the returned keys, and emits `delta = corrected − snapshot` over that set. Because the returned dict is exactly what gets applied, the modified-name set is load-bearing and cannot silently drift from the write. `CorrectorOutput.modified_names` exposes those names (the `delta` keys).

Second, a new `StepOutput` (`output`, `stepper_state`, `corrector_diagnostics`) carries the diagnostics through the step/stepper layer up to `Stepper.predict`, which discards them and returns a corrected-only `BatchData` exactly as before. `step_with_adjustments` detaches the deltas and asserts the post-corrector adjustment names are disjoint from the corrector's modified names. This part is behavior-preserving: corrected output, rollout state, loss, metrics, and saved files are identical to `main`, and existing checkpoints load unchanged.

Changes:
- `CorrectorABC.__call__` returns `CorrectorOutput` instead of a tuple; `CorrectorOutput` gains a `modified_names` property. Each `Correction.__call__` returns only the fields it modified (its tuple return type is unchanged); `CorrectionSequence` unions the returned keys into the modified-name set.
- `StepOutput` (new) replaces the `tuple[TensorDict, StepperState | None]` return of `StepABC.step` and every step implementation.
- `Stepper.step` and the predict generators return/yield `StepOutput`; `process_prediction_generator_list` consumes it; `Stepper.predict` is unchanged externally.

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+Physics corrections write through `AtmosphereData`/`OceanData` stackers, so the
+written gen_data keys (`PRESsfc` vs `PS`, present air-temperature/thetao level
+names, `sst`/`hfds` presence) are resolved from the data, not from config fields
+— and `_build` is not given the variable naming. `touched_names` therefore takes
+`gen_data` and resolves at sequence entry, where the data is in hand; the result
+is still static across a run's steps. Config-only corrections (`ForcePositive`,
+`SeaIceFractionCorrection`, `IceBudgetCorrection`) ignore the argument.
+Alternative considered: thread the resolved names into each correction at build
+time — rejected because `_build` has no variable list and `DatasetInfo` carries
+none.
```
mcgibbon: I don't understand this prose.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+def test_predict_returns_corrected_only_batchdata(...):
+    # GOAL: Stepper.predict exposes no diagnostics field; rollout state is the corrected
+    #       output; result is byte-for-byte identical to main for a corrector-equipped config.
+```
+
+(FCN3's step seam is exercised via its existing step test alongside the others.)
+
+---
+
+## Design note — why `touched_names` takes `gen_data`
```
mcgibbon: touched_names hasn't been defined in this PR plan document, it also might help to describe why it exists, or what alternatives to its existence were considered and why this plan won out in the end.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+
+    def __call__(self, input_data, gen_data, forcing_data, corrector_state) -> CorrectorOutput:  # CHANGED
+        # Snapshot ONLY the declared touched names from incoming gen_data (hold
+        # references; corrections apply out-of-place so entry tensors are not
+        # mutated). Apply corrections unchanged, then build populated diagnostics.
+        touched = self.touched_names(gen_data)
+        snapshot = {name: gen_data[name] for name in touched}
+        gen_data = dict(gen_data)
+        for correction in self._corrections:
+            gen_data, corrector_state = correction(input_data, gen_data, forcing_data, corrector_state)
```
mcgibbon: `touched_names` feels like API that risks going out-of-date, since it lives separate from the code actually applying corrections. If it automatically defines the names based on the data, it's not clear why it would be defined per-corrector. And it's a little unclear what "touching" means.

Could we instead have the correction return a dictionary which then gets accumulated and also added to the gen_data field? Normally I'd ask for that to be `deltas`, but to reduce the amount of change we're making here, I'd suggest the returned dict be the `new_gen_data`, which you can use to dict-update the gen_data and accumulate a `touched` set. The correctors must only put values they've updated into this dict. This would remove a fairly confusing new method from the API.

Actually I guess an even smaller change on this is to say, the gen_data returned by the correction must only include corrected fields - with that change, you don't need this new method.

Previously I was concerned with being able to separate between different types of corrector updates at this level, but I think now we should handle that separately, afterwards. For now we will at least be able to specify in the corrector loss what fields to exclude from the pre-corrector/regularization set.

jpdunc23: OK, I agree that individual `CorrectorABC` implementations can already specify what fields were "modified" (let's use this in place of "touched") for free if they only include those fields in their return. Then, the modified fields can be combined with non-modified fields by the caller (maybe the top-level `Corrector` objects) while keeping track of the names that were modified.

pr-plan.md
jpdunc23: Per https://github.com/ai2cm/ace/pull/1328#discussion_r3499716941 we'll remove `touched_names`.

pr-plan.md
jpdunc23: This is just about resolving between "standard" names and data names, but should be irrelevant given https://github.com/ai2cm/ace/pull/1328#discussion_r3499716941.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+value objects/helper already exist in `fme/core/corrector/output.py` (no consumer
+yet); this PR wires them in.
+
+---
+
+## `fme/core/corrector/registry.py` (modified)
+
+```python
+class Correction(Protocol):
+    def touched_names(self, gen_data: TensorMapping) -> set[str]:  # NEW
```
jpdunc23: We won't need this or other `touched_names` methods; Instead, `Correction.__call__` will be self-documenting by returning a `TensorDict` that only includes the modified fields.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+        # Snapshot ONLY the declared touched names from incoming gen_data (hold
+        # references; corrections apply out-of-place so entry tensors are not
+        # mutated). Apply corrections unchanged, then build populated diagnostics.
+        touched = self.touched_names(gen_data)
+        snapshot = {name: gen_data[name] for name in touched}
+        gen_data = dict(gen_data)
+        for correction in self._corrections:
+            gen_data, corrector_state = correction(input_data, gen_data, forcing_data, corrector_state)
+        corrected = dict(gen_data)
+        return CorrectorOutput(
```
jpdunc23: `CorrectorOutput` may benefit from an `@property` called `names` that returns the keys of `corrected`.

pr-plan.md
jpdunc23: Does `build_corrector_diagnostics` still need its `touched_names` argument?

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+def step_with_adjustments(...) -> StepOutput:  # CHANGED — was tuple[TensorDict, StepperState | None]
+    ...
+    diagnostics = CorrectorDiagnostics()  # NEW — empty when no corrector
+    if corrector is not None:
+        corrector_state = stepper_state.corrector_state if stepper_state is not None else None
+        result = corrector(input, output, next_step_input_data, corrector_state)  # CHANGED — CorrectorOutput
+        output = result.corrected
+        # Detach diagnostic tensors here, once, unconditionally (step-boundary
+        # detach keeps the corrector autograd-agnostic; PR5's attach/detach
+        # control is a one-line change at this same seam).
```
jpdunc23: Fine for the plan, but clean up this comment in the implementation, e.g. remove reference to future PR.

pr-plan.md
jpdunc23: Adopted. The plan now drops `touched_names` entirely: each `Correction.__call__` returns **only the fields it modified**, `CorrectionSequence` dict-updates `gen_data` with that subset and unions the returned keys into the modified-name set, and `delta` is built over that set. Because the returned dict is exactly what gets applied, the modified-name set is load-bearing and can't drift from the write — a dropped field would change the output and fail the behavior-preservation test. Terminology is now "modified" throughout. See the rewritten `registry.py`/`utils.py`/`atmosphere.py`/`ocean.py`/`ice.py` sections and the new "why corrections return only their modified fields" design note.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Right — `touched_names` is gone; `Correction.__call__` is self-documenting by returning only the modified fields. On the second question: `build_corrector_diagnostics` keeps its explicit name-set argument (now `modified_names`), fed the set `CorrectionSequence` accumulates from the returned keys. Keeping the signature avoids re-touching the already-merged value-objects helper, and its snapshot-at-entry + modified-set + final-corrected inputs match exactly.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: `touched_names` is removed from the plan, so the undefined-term issue is moot. The new "why corrections return only their modified fields" design note explains the chosen mechanism and records the rejected alternative (a per-correction `touched_names` declaration separate from the write, which can drift and needs a backstop drift test).

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: That prose described the old `touched_names`-takes-`gen_data` resolution and is removed along with the mechanism. The replacement design note states plainly that each correction resolves its written keys against `gen_data` via the same stacker its write goes through, at the moment it returns — no separate declaration.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Added as `CorrectorOutput.modified_names`, returning `self.diagnostics.delta.keys()` (the corrector-modified variables). Note `corrected` holds the full output `gen_data`, so its keys are all variable names rather than the modified set — `modified_names` is the diagnostically meaningful one and is what the disjointness guard uses.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Done in the plan: the detach comment no longer references a future PR. Flagging for implementation too — the inline comment should stay free of forward PR references.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+        # contain ONLY the fields this correction modified — not the full gen_data.
+        # The caller dict-updates gen_data with it and unions its keys into the
+        # modified-name set. (Today each correction returns the full gen_data.)
+
+
+class CorrectorABC(abc.ABC):
+    @abc.abstractmethod
+    def __call__(self, ...) -> CorrectorOutput:  # CHANGED — was tuple[TensorDict, CorrectorState | None]
+        ...
+    # no touched_names — the modified-name set is derived from what corrections return
```
jpdunc23: Don't refer to defunct design such as `touched_names`. No action required here, just be sure not to refer to `touched_names` in the implementation except when referring to the `build_corrector_diagnostics` argument.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+    def step(self, args, wrapper=...) -> StepOutput:  # CHANGED
+        ...
+```
+
+## `fme/core/step/multi_call.py` (modified)
+
+Wraps another step; not in the issue's "four step implementations" list but
+must change for the contract to hold. The secondary `_multi_call.step` output is
+merged into the wrapped step's output as today; its diagnostics are discarded —
+only the wrapped step's diagnostics/state are carried.
```
jpdunc23: @mcgibbon does this sound right to you?

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+        if self._corrector_disabled and self._training:
+            # applied nothing -> corrected passthrough, empty diagnostics
+            return CorrectorOutput(corrected=dict(gen_data), corrector_state=corrector_state)
+        return self._wrapped(input_data, gen_data, forcing_data, corrector_state)
+```
+
+The individual `Correction.__call__`s keep their `tuple[TensorDict,
+CorrectorState | None]` return type — only their *content* contract changes
+(return the modified subset, not the full `gen_data`). Only `CorrectorABC.__call__`
+changes its return type, to `CorrectorOutput`.
```
jpdunc23: Will the proposed `test_modified_returns_match_actual_changes` guard against regressions on the new content contract? How can we prevent it from getting stale?

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+            ...,
+        )
+        ...
+```
+
+---
+
+## Tests
+
+## `fme/core/corrector/test_atmosphere.py` (modified) · `test_ocean.py` (modified)
```
jpdunc23: Need to add new `test_ice.py` as well.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+
+def test_corrector_delta_matches_modified_returns(...):
+    # GOAL: diagnostics.delta keys == union of the corrections' returned keys
+    #       (== CorrectorOutput.modified_names); delta[name] == corrected[name] - input[name] exactly.
+    # Prefer a deterministic constant-offset/scale scenario so delta is hand-computable.
+
+def test_corrector_delta_empty_when_nothing_modified(...):
+    # GOAL: no enabled option that modifies a field -> empty delta.
+
+def test_modified_returns_match_actual_changes(...):  # THE BACKSTOP
```
jpdunc23: It will be somewhat difficult to comprehensively guarantee that no corrector returns unmodified outputs (e.g., if a regression introduces a bug where some auxiliary field in `gen_data` is included in the corrector's output), especially in a way that won't easily go stale.

I agree this test should work with "a representative input" rather than try to be comprehensive. To guard against regressions, we should rely on inline documentation / code comments so that maintainers and agents understand and avoid changing the `Corrector` return contract.

pr-plan.md
jpdunc23: `test_modified_returns_match_actual_changes` will hopefully provide some defense, but we shouldn't try to make it a comprehensive guard. Instead, make sure that inline docs for `Corrector` clearly state the expected output contract.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+
+## `fme/core/corrector/registry.py` (modified)
+
+```python
+class Correction(Protocol):
+    def __call__(self, ...) -> tuple[TensorDict, CorrectorState | None]: ...
+        # CONTRACT CHANGE (not a signature change): the returned TensorDict must
+        # contain ONLY the fields this correction modified — not the full gen_data.
+        # The caller dict-updates gen_data with it and unions its keys into the
+        # modified-name set. (Today each correction returns the full gen_data.)
```
jpdunc23: Be sure to include clear documentation of the expected `__call__` output contract, both here in the `__call__` inline documentation (`Returns:` annotation stating that the `TensorDict` only includes fields modified by the correction) and in the `Correction` class docs (something along the lines of "`__call__` implementations are expected to return only corrector-modified fields").

pr-plan.md
mcgibbon: Reporting the "original" step's corrector diagnostics seems right to me as a baseline. You might want to have someone like Spencer or Brian look at this when it's PR'd at least to make sure they're aware of it.

pr-plan.md
mcgibbon: I think what we have is pretty good assurance, especially with AI review. The test should ensure at least something that isn't corrected isn't propagated which guards against the most likely failure case. If a corrector isn't copying everything de facto, I think it's unlikely an author would copy specific fields they aren't actually correcting.

The one potentially confusing thing is, we should probably be clear that it should propagate fields that the class _does correct_ but happens not to have corrected for this step because the values have no error to correct. I think this will happen naturally with the way we write our code, but it might be worth specifying in a docstring.

pr-plan.md
```diff
@@ -0,0 +1,434 @@
...
+def step_with_adjustments(...) -> StepOutput:  # CHANGED — was tuple[TensorDict, StepperState | None]
+    ...
+    diagnostics = CorrectorDiagnostics()  # NEW — empty when no corrector
+    if corrector is not None:
+        corrector_state = stepper_state.corrector_state if stepper_state is not None else None
+        result = corrector(input, output, next_step_input_data, corrector_state)  # CHANGED — CorrectorOutput
+        output = result.corrected
+        # Detach diagnostic tensors here, once, unconditionally (step-boundary
+        # detach keeps the corrector autograd-agnostic).
+        diagnostics = CorrectorDiagnostics(
```
mcgibbon: Suggestion: helper method `diagnostics = result.diagnostics.detach()` of type CorrectorDiagnostics.

mcgibbon (commented):
Posted comments, but you can proceed with the plan.

pr-plan.md
jpdunc23: I will incorporate this or something similar into the plan for a future PR where some of the deltas won't be detached, depending on the optimization.

pr-plan.md
jpdunc23: @spencerkclark or @brianhenn, does the multi-call output of `MultiCallStep` potentially have additional corrector diagnostics we should incorporate into the `StepOutput`? 

If so, let me know if you'd like me to handle that in this PR or if we can do it in a follow-on.

pr-plan.md
jpdunc23: > it should propagate fields that the class _does correct_ but happens not to have corrected for this step because the values have no error to correct

I think this is equivalent to "correction delta is 0 everywhere". Seems unlikely (or is it?), but I agree it should be included in the output.

pr-plan.md
jpdunc23: Will do. The plan now requires the contract be documented where it can't drift from the code: a `Correction` class docstring ("implementations return only corrector-modified fields") and an explicit `Returns:` annotation on each concrete `__call__`. Added as an acceptance criterion too.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Agreed — `test_modified_returns_*` is a backstop, not a comprehensive guard. The plan now leans on the inline `Correction` docs (class docstring + `Returns:` annotation) as the primary defense against drift, and the backstop test is explicitly scoped narrow.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Reconciled. The backstop test is narrowed to guard the most-likely regression — that a field the corrector isn't responsible for is never propagated (i.e. it isn't blindly returning the full `gen_data`) — rather than asserting returned-keys == numerically-changed. Per your and @mcgibbon's point, a correction legitimately returns a field it's responsible for even when that field's delta is zero this step (e.g. `ForcePositive` on already-non-negative data); `delta[name] == corrected[name] − input[name]` still holds exactly (zero for such a field). The contract is documented inline as the primary guard.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Added a `test_ice.py` section: behavior preservation for `IceCorrector` plus a check that `IceBudgetCorrection` returns only its processed prognostic keys and three budget-term names (empty subset when `corrected_variables is None`).

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Kept the detach inline for this PR and noted that a `CorrectorDiagnostics.detach()` convenience helper is deferred to the later PR that makes attach/detach conditional on the optimization (where some deltas must stay attached).

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23: Reworded to drop the defunct `touched_names` reference. The implementation should likewise avoid the term except as the `build_corrector_diagnostics` argument name.

— 🤖 *Drafted by Claude (Opus 4.8, 1M context) via Claude Code, posted on @jpdunc23's behalf.*

pr-plan.md
jpdunc23, line 1: Delete before merge

fme/ace/stepper/single_module.py
```diff
@@ -504,10 +505,11 @@ def process_prediction_generator_list(
 
     Attaches the terminal stepper_state (from the last entry in
     ``output_list``) to the returned BatchData so it can propagate to the
-    next ``Stepper.predict`` call.
+    next ``Stepper.predict`` call. The per-step corrector diagnostics carried by
+    each ``StepOutput`` are not consumed here.
```
jpdunc23: Revert

fme/core/corrector/ice.py
```diff
@@ -133,17 +133,39 @@ def _rebalance(s, k, t, mask, mass, sign=1):
...
+        sic_vars = {"siconc", "sea_ice_fraction", "ocean_sea_ice_fraction"}
+        processing_order: list[str] = []
+        if "simass" in self.corrected_variables:
+            processing_order.append("simass")
+        for var in sic_vars:
+            if var in self.corrected_variables:
+                processing_order.append(var)
+        if "sisnmass" in self.corrected_variables:
+            processing_order.append("sisnmass")
+        return processing_order
```
jpdunc23: Why was this factored out?

fme/core/corrector/ocean.py
```diff
@@ -450,13 +473,16 @@ def _force_conserve_ocean_heat_content(
     heat_content_correction_ratio = (
         global_input_ocean_heat_content + expected_change_ocean_heat_content
     ) / global_gen_ocean_heat_content
-    # apply same temperature correction to all vertical layers
+    # apply same temperature correction to all vertical layers; each written
+    # level (and the SST) is placed in the output as it is corrected, so the
+    # return contains only the modified fields.
```
jpdunc23: Revert

fme/core/corrector/registry.py
```diff
@@ -76,6 +77,15 @@ class Correction(Protocol):
...
     maintain state passes ``corrector_state`` through unchanged.
+
+    Each ``__call__`` returns a ``TensorDict`` containing **only the fields this
+    correction modified** -- not the full ``gen_data``. The caller dict-updates
+    ``gen_data`` with the returned subset and takes its keys as the set of
+    variables the correction is responsible for writing. Because the returned
+    dict is exactly what gets applied, the returned keys are the single source of
+    truth for what changed and cannot drift from the write. This contract is
+    documented here and reiterated in a ``Returns`` annotation on each concrete
+    ``__call__``; that inline documentation is the primary guard against drift.
```
jpdunc23:
```suggestion
    truth for what changed and cannot drift from the write.
```

fme/core/corrector/test_atmosphere.py
```diff
@@ -537,3 +548,83 @@ def test_conserve_dry_air_requires_vertical_coordinate():
...
+def _build_full_atmosphere_corrector(tensor_shape):
+    """Build an atmosphere corrector with every field-modifying option active,
+    along with matching input/gen/forcing data."""
+    config = AtmosphereCorrectorConfig(
+        conserve_dry_air=True,
+        zero_global_mean_moisture_advection=True,
+        moisture_budget_correction="advection_and_precipitation",
+        force_positive_names=["PRATEsfc"],
+        total_energy_budget_correction=EnergyBudgetConfig("constant_temperature", 1.0),
+    )
```
jpdunc23, line 562: Possibly beneficial staleness check: Ensure that the init args we expect are the only ones defined for `AtmosphereCorrectorConfig`. If not (i.e., WIP changes added a new corrector and this test needs updating) then raise an error.

Similarly for the `OceanCorrectorConfig` and `IceCorrectorConfig` tests.

fme/ace/stepper/test_single_module.py
```diff
@@ -1298,6 +1301,51 @@ def test_predict_threads_stepper_state_across_calls():
...
+        self._offset = offset
+
+    def __call__(
+        self,
+        input_data: TensorMapping,
+        gen_data: TensorMapping,
+        forcing_data: TensorMapping,
+        corrector_state: CorrectorState | None,
+    ) -> tuple[dict, CorrectorState | None]:
+        return {self._name: gen_data[self._name] + self._offset}, corrector_state
```
jpdunc23: This is identical to `_ConstantOffsetCorrection` in `fme/ace/stepper/test_single_module.py`; Can we just import `_ConstantOffsetCorrection` from there? Make public: `ConstantOffsetCorrection`.

fme/core/step/single_module.py
```diff
@@ -603,15 +606,35 @@ def step_with_adjustments(
...
     if corrector is not None:
         corrector_state: CorrectorState | None = (
             stepper_state.corrector_state if stepper_state is not None else None
         )
-        output, corrector_state = corrector(
-            input, output, next_step_input_data, corrector_state
+        result = corrector(input, output, next_step_input_data, corrector_state)
+        output = result.corrected
+        # Detach the diagnostic tensors once, at the step boundary, so the
+        # corrector itself stays autograd-agnostic.
```
jpdunc23: This comment is not super clear; Remove or simplify: "Detach the corrector diagnostic tensors"

fme/core/ocean_data.py
```diff
@@ -105,6 +105,10 @@ def _get(self, name):
                 return self._get_prefix(prefix)
         raise KeyError(name)
 
+    def get_all_vertical_level_names(self, standard_name: str) -> list[str]:
+        """Return names of all vertical levels for a given standard name."""
+        return self._stacker.get_all_level_names(standard_name, self.data)
+
```
jpdunc23: This doesn't seem to be used. We can remove if we don't need it.

fme/ace/stepper/single_module.py
jpdunc23: Reverted — dropped the added sentence; the docstring ends at the `Stepper.predict` line again.

fme/core/corrector/ice.py
jpdunc23: Inlined it back into `__call__`. It had a single call site, so the separate method wasn't earning the indirection — the ordering logic now sits where it's used, alongside the modified-subset accumulation.

fme/core/corrector/ocean.py
jpdunc23: Reverted to the original one-line comment.

fme/core/corrector/registry.py
jpdunc23: Trimmed the docstring to end at “…cannot drift from the write.” as suggested.

fme/core/corrector/test_atmosphere.py
jpdunc23: Added `test_atmosphere_corrector_config_fields_are_exercised` (and ocean/ice equivalents) that assert the config's dataclass fields equal an expected set, so adding a new corrector option fails the test until the delta/modified-return tests cover it. It already surfaced the inherited `corrector_disabled_epochs` field, now listed explicitly.

fme/ace/stepper/test_single_module.py
jpdunc23: Made it public as `ConstantOffsetCorrection` in `test_registry.py` and imported it here, dropping the duplicate.

fme/core/step/single_module.py
jpdunc23: Simplified to “Detach the corrector diagnostic tensors.”

fme/core/ocean_data.py
jpdunc23: Removed it — it was unused; only the `AtmosphereData` version is called (by the total-energy correction).

fme/ace/stepper/test_single_module.py
```diff
@@ -1298,6 +1302,34 @@ def test_predict_threads_stepper_state_across_calls():
     torch.testing.assert_close(pres_after_2, pres_after_1 + float(n_steps))
 
 
+def test_predict_seam_yields_stepoutput_and_discards_diagnostics():
+    stepper = _get_stepper(["a"], ["a"])
+    stepper._step_obj._corrector = CorrectionSequence(  # type: ignore[attr-defined]
```
mcgibbon: Nit: better to `assert isinstance(stepper._step_obj, <class>)` than to type: ignore.

fme/ace/stepper/test_single_module.py
```diff
@@ -1298,6 +1302,34 @@ def test_predict_threads_stepper_state_across_calls():
     torch.testing.assert_close(pres_after_2, pres_after_1 + float(n_steps))
 
 
+def test_predict_seam_yields_stepoutput_and_discards_diagnostics():
```
mcgibbon: nit: Is it necessary to test that the diagnostics are discarded? We'll be refactoring this in the next PR, right?

fme/core/atmosphere_data.py
```diff
@@ -110,6 +112,16 @@ def _set(self, name, value):
 
     def _set_prefix(self, prefix, value):
         self.data[prefix] = value
+        self._modified_keys.add(prefix)
+
+    @property
+    def modified_data(self) -> TensorDict:
```
mcgibbon, line 118: Comment: The changes to this object make me go hmmm but I don't have a better suggestion or even a clear formulation of why it does in this PR. Just wanted to point out this change as something to be particularly aware of. Like, _set_prefix could be a side-effect of a number of things, and I _think_ that will correspond to what we want here, but it's hard to verify and to be sure we won't add a bug in the future.

AtmosphereData was originally made as a way to easily chain multiple equations together to derive something without needing to manually reason about their connections. Maybe now that we have AI coding we could use function calls instead.

No action here though.

fme/core/corrector/registry.py
```diff
@@ -159,13 +175,27 @@ def __call__(
...
+            changed, corrector_state = correction(
                 input_data, gen_data, forcing_data, corrector_state
             )
-        return dict(gen_data), corrector_state
+            # ``changed`` holds only the fields this correction modified; its
+            # keys are the set of variables it is responsible for writing.
+            gen_data.update(changed)
+            modified |= changed.keys()
+        corrected = dict(gen_data)
+        return CorrectorOutput(
```
mcgibbon, line 194: Question: Should the CorrectorOutput arguments be required, to enforce that they should all get set by the corrector (even if we choose to set them as empty from certain correctors)? Kind of out of scope of this PR but something to consider.

fme/core/corrector/registry.py
```diff
@@ -159,13 +175,27 @@ def __call__(
...
+            corrected=corrected,
+            diagnostics=build_corrector_diagnostics(snapshot, corrected, modified),
```
mcgibbon, line 196: Comment: In the future I think we'll want the sub-correctors to have more explicit control over diagnostics, but this fits the current feature set.

fme/core/corrector/test_atmosphere.py
```diff
@@ -480,7 +489,8 @@ def test_conserve_dry_air_seeds_from_ic_input():
     # Sanity: gen and ic differ (otherwise the test is trivial).
     assert not torch.allclose(ic_dry_air, gen_dry_air_before)
 
-    corrected, state = corrector(ic, gen, {}, None)
+    result = corrector(ic, gen, {}, None)
```
mcgibbon, line 492: Question: Should any of these tests gain new assertions about the newly added `result.diagnostics` attribute?

fme/ace/stepper/test_single_module.py
jpdunc23: Does the proposed assertion avoid the mypy error here? If so, remove the ignore. 

Otherwise, if we leave the ignore would the test naturally fail if some refactor removes the private attribute we're assigning to? If not, the ignore is too brittle and we should consider refactoring the test.

fme/ace/stepper/test_single_module.py
jpdunc23: Agreed, we probably don't need this.

fme/core/atmosphere_data.py
jpdunc23: I agree with these concerns.

However, this bit of extra "modified names" tracking here helped avoid more invasive refactoring of the atmosphere corrections by giving some natural API surface that makes transparent the manner in which they're already using AtmosphereData to help with the corrections.

One potential improvement for a future PR would standardize this by using `set_*` throughout and maintaining a single AtmosphereData (or OceanData) across all the corrections, with `modified_data` giving the final result.

pr-plan.md
mcgibbon: > Seems unlikely (or is it?)

I think we definitely have force-positive fields (basically all the humidity levels for example) that I expect much of the time are never negative/zero, but it's still good to record them and to get correct statistics in the rare cases where they are negative (maybe early in training).

fme/ace/stepper/test_single_module.py
mcgibbon: Generally speaking asserts fix this kind of mypy error, i.e. it casts/specializes the type. We use this pattern commonly in other tests, where we know we've happened to structure a generic class with specific implementations of their container objects.

fme/core/corrector/registry.py
jpdunc23: > Should the CorrectorOutput arguments be required

Do you mean we should remove the defaults for the `diagnostics` and `corrector_state` attrs of `CorrectorOutput`?

fme/core/corrector/test_atmosphere.py
jpdunc23: Good question, let's look into it.

fme/core/atmosphere_data.py
mcgibbon: Yeah I think this is the most natural way to do it with AtmosphereData, it's just making me up the priority on something like a complete refactor of this code. There's no reason AtmosphereData can't be replaced with a series of stateless pure functions that represent atmospheric equations, the problem is it's tedious to stitch them together - but if an AI is doing the tedious part, it's a lot cleaner to do it that way. At least that's my intuition, we'll see if/when we try to turn that idea into a PR.

fme/ace/stepper/test_single_module.py
jpdunc23: Done — replaced the `# type: ignore[attr-defined]` with `assert isinstance(stepper._step_obj, SingleModuleStep)` before the corrector assignment (the same specialize-via-assert pattern already used elsewhere in this file). mypy is satisfied without the ignore, and the assert now fails loudly if the private `_step_obj` structure is refactored away.

fme/ace/stepper/test_single_module.py
jpdunc23: Agreed — dropped the predict-side discard assertions and renamed the test to `test_predict_generator_yields_detached_stepoutput`, keeping only the generator-seam checks (yields `StepOutput`, delta populated and detached). `predict` itself is unchanged from main and the follow-on PR adds the diagnostics field to the returned data immediately, so a negative "no diagnostics" assertion here would be short-lived.

fme/core/corrector/test_atmosphere.py
jpdunc23: Looked into it — the `diagnostics.delta` contract already has dedicated coverage added in this PR: `test_atmosphere_corrector_delta_matches_modified_returns` (delta keys == modified names, `delta[name] == corrected[name] - input[name]`, and the exact modified-name set), `test_atmosphere_corrector_unmodified_fields_pass_through` (the narrow backstop), and `test_atmosphere_corrector_empty_delta_when_nothing_modified`. These dry-air tests are focused on seeding/persistence of the dry-air reference, so I'd keep them on that rather than duplicate the delta assertions. Happy to add an explicit delta check to one of them if you'd prefer belt-and-suspenders.

fme/core/corrector/registry.py
mcgibbon: I mean should we remove, but yes, if we want these to always be given we should require they're always given. is there a scenario where we want to not give them, where it makes sense to fall back on a default instead of explicitly saying "for this code we're going to pass an empty dict here"?

fme/ace/test_train.py
```diff
@@ -408,7 +408,7 @@ def _get_test_yaml_files(
                             ),
                         ),
                         ocean=OceanConfig(
-                            surface_temperature_name=in_variable_names[0],
+                            surface_temperature_name="surface_temperature",
```
mcgibbon, line 411: Chore: revert unrelated change (unless it is in fact related/necessary)

```suggestion
                            surface_temperature_name=in_variable_names[0],
```

mcgibbon (approved):
LGTM

fme/ace/test_train.py
jpdunc23: This was needed because `in_variables_names[0]` is `"PRESsfc"` here, which gets modified by the dry air mass corrector. This then raises an error with the new `step_with_adjustments` check that no names in `CorrectorDiagnostics.delta` are modified by post-corrector adjustments (only `Ocean` at this point).

PR #1329, opened 2026-06-30
Feature: Variable Masking (simplified, batch-broadcast)

yyexela:
Adds training-time input channel dropout (variable masking) to `SingleModuleStep`. This is a simplified reimplementation of #1246 against `main`.

The motivation is to study how randomly withholding input channels during training affects model robustness and generalization — particularly whether a model trained with masking can perform inference when some inputs are unavailable.

The mask is sampled once per forward `step` and held constant across that step's `multi_call` sub-calls; each step draws independently. It is **broadcast across the whole batch** (shape `[1, n_channels]`), so `n_ensemble` is never threaded into the Step and every sample and every ensemble member receives the same mask as a free consequence. Synthetic dropout is kept separate from the real `data_mask` (which marks genuinely-absent variables consumed by preprocessing and loss masking): predictions made from dropped inputs are still scored by the loss; only truly-absent target variables are excluded.

Changes:
- `fme.core.var_masking`: refactored into a small generator hierarchy. A masking **scheme** (`MaskingConfig = BernoulliMaskingConfig | UniformMaskingConfig`) governs a disjoint pool of channels and, each step, answers which of its own channels to drop:
  - `BernoulliMaskingConfig(rate)` — one `Bernoulli(rate)` is drawn for the whole pool; when it fires every channel in the pool is dropped together, otherwise none are, so the marginal drop probability of the pool equals `rate` exactly. A single-variable pool reproduces per-variable masking. Validates `rate ∈ [0, 1]`.
  - `UniformMaskingConfig(max_masked_vars)` — a count `k` is drawn uniformly from `[0, min(max_masked_vars, n)]` (min hard-coded to 0, `n` = pool size) and `k` of the pool's channels are dropped, so a draw may drop nothing. Rejects a bool `max_masked_vars` (bool is an `int` subclass) and negatives.
  - The union has **disjoint required fields** (`rate` vs `max_masked_vars`), so dacite discriminates it under `strict=True`, matching the `SchedulerConfig` precedent — no `kind` tag.
- `fme.core.var_masking.MaskingGroupConfig`: holds `variables` (list of channel names sharing one draw) and `masking` (a `MaskingConfig`). A variable may appear in at most one group; non-empty `variables` validated in `__post_init__`.
- `fme.core.var_masking.VariableMaskingConfig`: a **single** config (no top-level `kind` discriminator) with two fields:
  - `default` (a `MaskingConfig`, defaults to `UniformMaskingConfig(0)` = mask nothing) — the scheme applied to all **ungrouped** channels.
  - `override_groups` (list of `MaskingGroupConfig`) — each group's channels are governed **solely** by that group's own scheme and excluded from the `default` pool.
  Because the pools are disjoint, the schemes are fully decoupled: a group's `rate` is its exact per-group drop probability regardless of `default`. `build(names)` binds the config to the packed channel list and raises loudly naming any grouped variable absent from `names` (a typo fails at build, not silently masks nothing). `VariableMasking.sample_mask(device)` returns a `[1, n_channels]` presence mask (`True` = present); the mask is the union of every generator's dropped names. Draws are taken from a **private per-rank CPU `torch.Generator`** owned by `VariableMasking` (seeded lazily with `distributed_seed + rank`), not the global torch RNG, so masking is decoupled from the global stream and never forces a device sync. The `MaskingConfig` dataclasses are `frozen=True`. Exported from `fme.ace`: `BernoulliMaskingConfig`, `UniformMaskingConfig`, `MaskingGroupConfig`, `VariableMaskingConfig`.
- `fme.core.step.SingleModuleStepConfig.input_dropout`: new optional field; when set, randomly zeros a subset of packed input channels during training (no-op at eval time). Added only to `SingleModuleStepConfig`; sibling configs (`FCN3StepConfig`, `SecondaryModuleStepConfig`, `MultiCallStepConfig`, `SeparateRadiationStepConfig`) are untouched. The config is serialized but inert at inference.
- `SingleModuleStep` builds the runtime `VariableMasking` **once** at construction from `in_packer.names` (incl. GMR extra channels, which remain independently maskable); `build` performs the loud name validation. The mask is sampled at the top of each forward `step` and applied late inside `network_call` (after GMR/normalization), keyed by the Step's packed input channel names, and is separate from `data_mask`. It is constant across the step's `multi_call` sub-calls but resampled on the next step. `_draw_input_dropout_mask` returns `None` (no dropout) when `input_dropout` is unconfigured or the module is in eval mode, so inference and validation batches stay inert.
- `fme.coupled.CoupledTrainStepper`: input dropout on a coupled component step now trains without error — each forward step samples its own mask in train mode, so coupled training needs no special coordination.
- `fme.core.distributed.Distributed.broadcast_spatial`: new primitive; the Step broadcasts the sampled mask across the spatial/model-parallel group so all tiles of a sample agree (identity for non-distributed and data-parallel backends). Bool tensors round-trip through `uint8` since NCCL rejects bool collectives; the non-bool path clones its input so the collective never mutates the caller's tensor in place.

**Example config** (within a `single_module` step config):

Uniform masking of up to 3 ungrouped channels plus a Bernoulli-masked group:
```yaml
input_dropout:
  default:
    max_masked_vars: 3
  override_groups:
    - variables: [air_temperature]
      masking:
        rate: 0.1
```

Uniform masking only:
```yaml
input_dropout:
  default:
    max_masked_vars: 3
```

**Behavior notes:**
- Dropout is **per-step** (sampled once at the top of each forward `step`, held across that step's `multi_call` sub-calls, resampled on the next step).
- The mask is broadcast across the whole batch, so every sample and every ensemble member sees the same mask.
- In data-parallel training, the per-rank RNG seed (`distributed_seed + rank`) makes each data-parallel rank draw an independent mask (distinct local batches drop distinct channels); `broadcast_spatial` is the identity there and only re-syncs spatial co-ranks under model parallelism.
- `UniformMaskingConfig` minimum masked count is **0** (hard-coded, not a config option); a draw may drop no channels.
- Each masking scheme governs a **disjoint** pool of channels: grouped channels are masked solely by their group's scheme and excluded from the `default` pool, so a group's configured `rate` is its exact per-group drop probability, independent of `default`.
- GMR extra channels (`__gmr_extra__<field>`) fall in the `default` pool like any other ungrouped packed channel, but are group-masked only if a group lists the sentinel name itself, so an extra and its source field are masked independently.
- Channel-mask indicator channels (`include_channel_mask_inputs`) are not maskable; an indicator reflects its data channel's AND-combined `real & synthetic` presence.
- Synthetic dropout does not use `data_mask`; only real missing-variable masks exclude targets from the loss.

- [x] Tests added


yyexela:
Note:

A GMR sentinel channel (`__gmr_extra__<source>`) inherits its source field's `data_mask`: when the source is masked, `forward_transform` already zeros the sentinel, so the two must agree. Synthetic `input_dropout` works differently — it draws an independent mask for every packed channel, so a source and its sentinel can be dropped separately (intended, to expose the network to more train-time combinations). For the channel-mask inputs, each sentinel's presence bit is its source's `data_mask` AND its own dropout draw, so the bit always matches whether that channel's value was zeroed.

yyexela:
Two minor notes considered and intentionally left as-is:
- **RNG divergence across spatial tiles:** not an issue — every rank runs identical, data-independent RNG draw counts, so spatial ranks advance their RNG by the same amount and `broadcast_spatial` only reconciles the sampled values; no desync is introduced.
- **Empty-batch guard:** `next(iter(input_batch_data.data.values()))` in `_accumulate_loss` assumes non-empty data, which is always true in practice, so no guard was added.

fme/ace/stepper/single_module.py
```diff
@@ -1664,6 +1673,9 @@ def _accumulate_loss(
                 "Initial condition and forcing data must have the same labels, "
                 f"got {input_batch_data.labels} and {data.labels}."
             )
+        input_dropout_mask = self._stepper.make_input_dropout_mask(
```
mcgibbon: You can remove a lot of unnecessary API/changelog by removing the public methods related to making the dropout mask, and just making the dropout mask privately inside the Step object. I think the only reason this PR was written this way is because it was refactored from a previous state where we needed to handle the mask at this level in the Stepper - that's no longer the case.

fme/coupled/stepper.py
```diff
@@ -1902,6 +1902,17 @@ def __init__(
         self._stepper = stepper
         self._config = config
         self._loss = self._config._build_loss(stepper, config.n_coupled_steps)
+        for realm, component in (
+            ("atmosphere", stepper.atmosphere),
+            ("ocean", stepper.ocean),
+        ):
+            if component.has_input_dropout():
```
mcgibbon: Is this necessary? It feels like a lot to add an entire public method for, and it doesn't even give us config-vaidation-time safety. Without thinking too hard about it, I would expect it to be fine to enable input dropout in coupled runs, though I don't expect us to actually use that configuration.

mcgibbon:
For now, let's just have each forward step independently sample its own input dropout mask, applied only when the module is in train mode. We may revisit holding the mask constant across a rollout later. This doesn't impact your one-forward-step-training runs.

Once the mask has no lifetime beyond a single step() call, there's nothing to cache and no rollout boundary to coordinate, so the entire arming mechanism and its cross-class API disappear:

  - Drop StepABC.new_rollout() (and its MultiCallStep forward). Its only purpose was to mark the "resample now, not again until next rollout" boundary — with per-step sampling there is no boundary.
  - Drop the SingleModuleStep cache (_input_dropout_pending, _input_dropout_mask, _sample_input_dropout_mask, new_rollout) and the _accumulate_loss arming call. Just draw the mask at the top of step() and apply it late in network_call as you do now.
  - Drop has_input_dropout() everywhere and the coupled-training guard (per my other comment).

The eval-mode check already inside _draw_input_dropout_mask gives us "train mode only" for free, so inference and validation batches stay inert with no plumbing.

That should leave the change as roughly: var_masking.py, the input_dropout config field, the mask draw + late application + _build_channel_mask_dict change in SingleModuleStep, and the broadcast_spatial set — which is worth keeping for spatial-parallel correctness. No new methods on StepABC, MultiCallStep, the stepper wrapper, TrainStepper, or fme/coupled.

One note: MultiCallStep will draw independent masks on the two steps. That's fine — no need to make them consistent; training is perfectly valid without it.

Also note this will mean we don't do masking during validation. I think that's fine, it matches the behavior we'd get from a torch Dropout layer. It's also more informative of our model skill, giving us better information for interpreting our inference results.

Arcomano1234:
> Once the mask has no lifetime beyond a single step() call, there's nothing to cache and no rollout boundary to coordinate, so the entire arming mechanism and its cross-class API disappear:

Agreed. Claude also found a potential bug regarding this caching mechanism that could leak the cached dropout mask can leak into the inference predict path breaking things as those variables would always be masked during inference / eval.

yyexela:
Re: residual training

Current impl `y = x + f(m·x)` keeps net target as **true tendency `x_next − x` always**, masked or not — masking only degrades input features `f` uses, never flips output semantics. Baseline `x` is additive outside `f`, so it drops from the gradient (`∂x/∂θ=0`) — no shortcut/leakage, net genuinely learns tendency from degraded features. This mirrors non-residual (predict absolute from `m·x`) exactly: one coherent target, mask hits inputs only.

fme/core/var_masking.py
```diff
@@ -0,0 +1,156 @@
...
+    Parameters:
+        variables: Names of the input channels in this group. Every name must
+            match a packed input channel (validated by
+            ``VariableMaskingConfig.validate_names``); a typo raises rather than
+            silently masking nothing.
+        rate: Bernoulli masking rate in ``[0, 1]`` shared by the whole group.
+    """
+
+    variables: list[str]
+    rate: float
```
mcgibbon: I'd kind of like to drop this version.

fme/core/var_masking.py
```diff
@@ -0,0 +1,156 @@
...
+            count is sampled uniformly from
+            ``[0, min(max_masked_vars, n_uniform)]`` where ``n_uniform`` is the
+            number of ungrouped channels, so a draw may mask no variables.
+        variable_masking_groups: Optional list of variable groups, each with a
+            shared Bernoulli masking rate. Grouped channels are excluded from
+            the uniform pool. A variable may appear in at most one group.
+    """
+
+    max_masked_vars: int = 0
+    variable_masking_groups: list[MaskingGroupConfig] | None = None
```
mcgibbon: Since there's a lot of logic on this class, it would be nice to define it as a frozen dataclass.

mcgibbon: Issue: There are two kinds of masking here, handled at two different levels. This class is responsible for performing a type of masking but also for containing a container that performs a type of masking. It would be ideal to keep the same level of abstraction (low-level masking logic) at the same level of the hierarchy.

Issue: I know this might not seem important but I think it's plausible that at some point we'll want to mask vertical-layered variables as all-or-nothing, and the current config would be a bit challenging to support this and other variations we might want to try.

Suggestion (optional): Refactor this as follows.

Define these config levels:

VariableMaskingConfig:
default: MaskingConfig
override_groups: list[MaskingGroupConfig]
has a .build(self) -> VariableMasking (new implementation class)

MaskingGroupConfig:
variables: list[str]
masking: MaskingConfig

MaskingConfig = BernoulliMaskingConfig | UniformMaskingConfig
has def build(self, names: list[str]) -> MaskingGenerator, implemented on both of these Config classes

MaskingGenerator has .sample() -> list[str],  a list of masked fields. Likely implementation classes BernoulliMaskingGenerator and UniformMaskingGenerator

BernoulliMaskingConfig:
rate: float

UniformMaskingConfig:
max_masked_vars: int (required, as for this we're using this to indicate the type of the config under MaskingConfig)

VariableMasking is initialized with a `list[MaskingGenerator]`, and in its sample_mask method returns the mask that corresponds to the union of its generator outputs.

-----

The alternative to all of this would be to keep the specialization or go even further down it, but just refactor the level where low level logic is happening so it's consistent between the two masking types used here. Also my design here is somewhat quickly drawn up so feel free to improve it if necessary.

mcgibbon: If we go the specialization route I think we should just have `str` and not `list[str]` on the bernoulli masking config, since we're in practice just using it for CO2. Also calling it GroupMaskingConfig is a bit confusing.

fme/ace/stepper/single_module.py
yyexela: Removed unnecessary API/changelog

fme/coupled/stepper.py
yyexela: Removed by not having unique mask per rollout trajectory. Masks sampled each step.

fme/core/var_masking.py
yyexela: Refactored given below comment.

fme/core/var_masking.py
yyexela: Refactored without specialization, using suggested method.

mcgibbon (approved):
LGTM

PR #1330, opened 2026-06-30
Seedable random state threaded through StepperState for reproducible stochastic inference

mcgibbon:
`NoiseConditionedSFNO` draws its conditioning noise from the global torch RNG (via `fme.core.rand.randn`), so stochastic standalone/evaluator inference is not reproducible, and results depend on how the global RNG is consumed between predict windows (i.e. on `forward_steps_in_memory`). This PR threads a seedable random source through the existing `StepperState` channel so a seeded rollout produces a deterministic noise sequence, independent of `forward_steps_in_memory`, and exposes a `seed` config for both inference entrypoints.

The same generator object advances in place as it is consumed and rides `PrognosticState.stepper_state` from one step (and one `predict` call) to the next, so chunking the rollout into windows does not change the noise sequence. Draws are made on the CPU and moved to the device (mirroring `use_cpu_randn`), so a seed reproduces results across devices. `NoiseConditionedSFNO` itself is unchanged — its existing `randn` call transparently picks up the active generator.

Changes:
- `fme.core.rand.use_generator` / `randn` / `randn_like`: a context-managed active CPU generator that `randn`/`randn_like` draw from when set (no-op when `None`, preserving default behavior).
- `fme.core.random_state.RandomState`: wraps a CPU `torch.Generator` (`from_seed`); device/ensemble helpers are no-ops preserving the single advancing generator (it has no per-sample dim).
- `fme.core.stepper_state.StepperState`: new `random_state` field, threaded through the device/ensemble helpers.
- `fme.ace.stepper.single_module.Stepper.step`: activate the generator around the inner step and re-attach the advanced `RandomState` to the returned `StepperState` (surviving steps that rebuild it to seed corrector state).
- `fme.ace.data_loading.batch_data.PrognosticState.with_random_state`: attach a seed to an initial condition.
- `fme.ace.inference.inference.InferenceConfig.seed` and `fme.ace.inference.evaluator.InferenceEvaluatorConfig.seed`: optional seed wired into the rollout's initial condition.

Tests:
- `fme/ace/stepper/test_random_state.py` (integration, minimal `NoiseConditionedSFNO` with non-zero noise dim): a seeded rollout is identical whether run in one window or split into smaller windows (`forward_steps_in_memory` invariant); two same-seed rollouts match while different seeds (and the unseeded default) differ; a zero-noise-dim model is deterministic without a seed (confirming the noise is what the seed controls).
- `fme/ace/inference/test_inference.py::test_inference_entrypoint_seed_reproducible`: end-to-end through the standalone inference entrypoint (`main`) with a noise-conditioned checkpoint — same `seed` gives bitwise-identical output, different `seed` differs, and the unseeded default is non-reproducible.
- `fme/core/test_rand.py`: `use_generator` reproducibility, advancement, seed sensitivity, `randn_like` parity, `None` passthrough, and nesting restore.
- `fme/core/test_random_state.py`, `fme/ace/data_loading/test_batch_data.py`, and `fme/ace/stepper/test_single_module.py`: `RandomState`/`StepperState` plumbing, `with_random_state`, and `random_state` surviving a corrector `StepperState` rebuild.

Notes for the reviewer:
- The integration and entrypoint tests fill the (zero-initialized) `ConditionalLayerNorm` noise-conditioning weights so an untrained model's noise actually affects the output; otherwise different seeds would be indistinguishable.
- The seed drives a dedicated threaded generator rather than reseeding the global RNG, which is what makes reproducibility independent of `forward_steps_in_memory` and of unrelated global-RNG use.
- The stepper integration tests use a corrector that does not seed state, so the `StepperState`-rebuild survival path is covered separately by `test_predict_threads_random_state_alongside_corrector_state`.
- Restart files do not persist generator state, so cross-process resume reproducibility is out of scope here.
- The generator is one-per-process: if inference were ever run data-parallel across ranks, every rank would seed an identical generator and draw an identical noise sequence for its shard of initial conditions. The inference entrypoints are single-process today, so this only matters as a caveat for future distributed inference.

Verified on a real trained model: ran short (20-step) standalone inference of a stochastic 4°/daily `NoiseConditionedSFNO` (noise_embed_dim=32, isotropic) on beaker — twice on GPU and twice on CPU, fixed `seed`. Each same-device pair reproduced **bitwise-identically** across all 61 predicted fields; CPU vs GPU diverge chaotically (floating-point differences in the deterministic compute, not the noise — noise is drawn on CPU and moved to device, so it is identical across devices). GPU reproducibility did not require `use_deterministic_algorithms`.

- [x] Tests added

### DIFF
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -1058,9 +1058,6 @@ class Stepper:
         random_state = (
             args.stepper_state.random_state if args.stepper_state is not None else None
         )
-        # The inner step preserves random_state on the StepperState it returns
-        # (advancing the generator in place as it consumes it), so it threads
-        # unchanged into the next step and predict call.
         with use_generator(None if random_state is None else random_state.generator):
             output, stepper_state = self._step_obj.step(args=args, wrapper=wrapper)
         return self._output_process_func(output), stepper_state
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -47,6 +47,7 @@ from fme.core.normalizer import (
 )
 from fme.core.ocean import OceanConfig
 from fme.core.optimization import NullOptimization
+from fme.core.rand import use_generator
 from fme.core.registry import CorrectorSelector, ModuleSelector
 from fme.core.spatial_masking import NullSpatialMasking, StaticSpatialMaskingConfig
 from fme.core.step.args import StepArgs
@@ -1054,7 +1055,20 @@ class Stepper:
             the per-sample state to thread into the next call (or ``None``).
         """
         args = args.apply_input_process_func(self._input_process_func)
-        output, stepper_state = self._step_obj.step(args=args, wrapper=wrapper)
+        random_state = (
+            args.stepper_state.random_state if args.stepper_state is not None else None
+        )
+        with use_generator(None if random_state is None else random_state.generator):
+            output, stepper_state = self._step_obj.step(args=args, wrapper=wrapper)
+        if random_state is not None:
+            # The step may rebuild StepperState (e.g. when the corrector seeds
+            # its own state), dropping the random_state. Re-attach the same
+            # generator - which advanced in place as the step consumed it - so
+            # it threads unchanged into the next step and predict call.
+            stepper_state = dataclasses.replace(
+                stepper_state if stepper_state is not None else StepperState(),
+                random_state=random_state,
+            )
         return self._output_process_func(output), stepper_state
 
     def get_prediction_generator(
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -1058,17 +1058,11 @@ class Stepper:
         random_state = (
             args.stepper_state.random_state if args.stepper_state is not None else None
         )
+        # The inner step preserves random_state on the StepperState it returns
+        # (advancing the generator in place as it consumes it), so it threads
+        # unchanged into the next step and predict call.
         with use_generator(None if random_state is None else random_state.generator):
             output, stepper_state = self._step_obj.step(args=args, wrapper=wrapper)
-        if random_state is not None:
-            # The step may rebuild StepperState (e.g. when the corrector seeds
-            # its own state), dropping the random_state. Re-attach the same
-            # generator - which advanced in place as the step consumed it - so
-            # it threads unchanged into the next step and predict call.
-            stepper_state = dataclasses.replace(
-                stepper_state if stepper_state is not None else StepperState(),
-                random_state=random_state,
-            )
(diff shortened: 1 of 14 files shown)


fme/ace/stepper/single_module.py
```diff
@@ -1054,7 +1055,20 @@ def step(
...
+            output, stepper_state = self._step_obj.step(args=args, wrapper=wrapper)
+        if random_state is not None:
+            # The step may rebuild StepperState (e.g. when the corrector seeds
+            # its own state), dropping the random_state. Re-attach the same
+            # generator - which advanced in place as the step consumed it - so
+            # it threads unchanged into the next step and predict call.
+            stepper_state = dataclasses.replace(
+                stepper_state if stepper_state is not None else StepperState(),
+                random_state=random_state,
+            )
```
spencerkclark: Question: is this something that could be addressed at a lower level, e.g. where the `StepperState` is rebuilt following an update to the `CorrectorState`?

Naively it seems like it could be safer to keep the `StepperState` coherent across `self._step_obj.step` calls, rather than need to potentially fix it up here.

spencerkclark (approved):
Thanks very much for this @mcgibbon.  I had one minor question, but otherwise it looks good and well-tested.  It's nice that the `StepperState` infrastructure helped set the stage for this.  I verified that this branch led to reproducible results in a "real-world" setting ([WandB](https://wandb.ai/ai2cm/ace-shield/workspace?nw=23zv8usm8zpj)).

At this point it might help to create an issue documenting the reasons we know inference runs may not be reproducible across restarts, since this can be mysterious to external users—I guess it ultimately comes down to there not currently being a mechanism to serialize and reinitialize the `StepperState`.

mcgibbon:
> At this point it might help to create an issue documenting the reasons we know inference runs may not be reproducible across restarts, since this can be mysterious to external users—I guess it ultimately comes down to there not currently being a mechanism to serialize and reinitialize the StepperState.

Maybe before we document it we should first see if we can make them reproducible without too much work?

fme/ace/stepper/single_module.py
mcgibbon: Claude: Good call — done in 001c1c34c. The lower-level `step_with_adjustments` was rebuilding `StepperState` from scratch (`StepperState(corrector_state=...)`), which dropped `random_state`; that's what the caller-side fix-up in `Stepper.step` was compensating for. Changed the corrector branch to `dataclasses.replace(stepper_state, corrector_state=corrector_state)` (guarding the `None` case) so the inner step keeps `StepperState` coherent and preserves any other field, and removed the `Stepper.step` fix-up entirely. All the seedable-random-state tests (including `test_predict_threads_random_state_alongside_corrector_state` and the inference-entrypoint reproducibility test) still pass.

PR #991, opened 2026-03-19
Update time coarsening configurations for 4-degree datasets and era5

mcgibbon:
Add time coarsening configurations for daily-resolution datasets across CM4, SHiELD, and ERA5 sources, and make the stats/upload scripts idempotent so pipeline steps can be safely re-run.

Changes:
- Add `time_coarsen` sections to 4-degree CM4 configs (1pctCO2, piControl, random-CO2-ensemble) and SHiELD random-CO2-ensemble config
- Add `time_coarsen` section to `era5-1deg-8layer-1940-2022.yaml` with `total_frozen_precipitation_rate` as a window name
- Add new `era5-4deg-8layer-1940-2025.yaml` config for 4-degree ERA5 time coarsening and stats
- `scripts/data_process/get_stats.py`: skip stats generation when output files already exist
- `scripts/data_process/combine_stats.py`: skip combining when combined stats already exist
- `scripts/data_process/upload_stats.py`: skip Beaker upload when dataset already exists; add logging

- [ ] Tests added

PR #1331, opened 2026-06-30
Add configuration for processing 4-degree AMIP constant CO2 data

spencerkclark:
This PR adds a configuration for processing 4-degree AMIP constant CO2 data, which I produced and uploaded today.  It includes a configuration for a daily coarsened dataset.

An argo job is in progress here: `compute-fme-dataset-ensemble-j2kmq`.

### DIFF
--- a/scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
+++ b/scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
@@ -1,0 +1,182 @@
+runs:
+  AMIP-constant-CO2: gs://vcm-ml-raw-flexible-retention/2025-12-16-C96-SHiELD-AMIP-constant-CO2/regridded-zarrs/gaussian_grid_45_by_90/AMIP-constant-CO2
+data_output_directory: gs://vcm-ml-intermediate/2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset
+stats:
+  output_directory: gs://vcm-ml-intermediate/2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset-stats
+  beaker_dataset: 2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset-stats
+  start_date: "1979-01-01T06:00:00"
+  end_date: "2021-01-01T00:00:00"  # Only use whole years for computing the stats
+  data_type: FV3GFS
+dataset_computation:
+  reference_vertical_coordinate_file: gs://vcm-ml-raw-flexible-retention/2024-03-10-C96-SHiELD-FME-reference/vertical-coordinate-file/fv_core.res.nc
+  vertical_coarsening_indices:
+    - [0, 11]
+    - [11, 21]
+    - [21, 30]
+    - [30, 39]
+    - [39, 49]
+    - [49, 58]
+    - [58, 67]
+    - [67, 79]
+  renaming:
+    specific_humidity_at_two_meters: Q2m
+    air_temperature_at_two_meters: TMP2m
+    eastward_wind_at_ten_meters: UGRD10m
+    northward_wind_at_ten_meters: VGRD10m
+  variable_sources:
+    fluxes_2d.zarr:
+      - PRATEsfc
+      - LHTFLsfc
+      - SHTFLsfc
+      - DLWRFsfc
+      - DSWRFsfc
+      - DSWRFtoa
+      - ULWRFsfc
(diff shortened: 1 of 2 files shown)


scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
```diff
@@ -0,0 +1,182 @@
...
+    - air_temperature_7
+    - eastward_wind_0
+    - eastward_wind_1
+    - eastward_wind_2
+    - eastward_wind_3
+    - eastward_wind_4
+    - eastward_wind_5
+    - eastward_wind_6
+    - eastward_wind_7
+    - global_mean_co2
```
mcgibbon: I don't think it actually matters since CO2 is so slowly varying, but generally speaking concentrations that affect fluxes over a window are probably best to supply as time-window averages. This will be more relevant when we have aerosol forcings. Might require handling in the model config that produces the raw datasets.

Kind of makes me wonder if it's best to evolve daily averages for all fields, but we'll see for CMIP6 where that's the natural form of the data.

scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
```diff
@@ -0,0 +1,182 @@
+runs:
+  AMIP-constant-CO2: gs://vcm-ml-raw-flexible-retention/2025-12-16-C96-SHiELD-AMIP-constant-CO2/regridded-zarrs/gaussian_grid_45_by_90/AMIP-constant-CO2
+data_output_directory: gs://vcm-ml-intermediate/2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset
+stats:
+  output_directory: gs://vcm-ml-intermediate/2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset-stats
+  beaker_dataset: 2026-06-30-vertically-resolved-c96-4deg-shield-amip-constant-co2-dataset-stats
+  start_date: "1979-01-01T06:00:00"
+  end_date: "2021-01-01T00:00:00"  # Only use whole years for computing the stats
```
mcgibbon: I thought our stats files normally indicated the time period they're computed over in the file path, which is pretty helpful for understanding what it is. For example, thinking about this really makes me wonder if this choice affects the results of training with pre-1979 data.

mcgibbon (approved):
LGTM, argo job completed successfully. Make sure to copy the data (and I would suggest also the stats) to weka. Perhaps we should automate that part of the process - I somehow often mess it up.

scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
spencerkclark: Thanks, good catch. Indeed I have made this point in the past, https://github.com/ai2cm/ace/pull/991#discussion_r2962472799, but copied this time coarsening config from [the SHiELD AMIP ensemble](https://github.com/ai2cm/ace/blob/349aa906c7f18206c7662359ad6ccb5d6b056e12/scripts/data_process/configs/shield-amip-ensemble-c96-1deg-8layer.yaml#L95-L185). In this case it really does not matter since CO2 is constant, but I updated the config and relaunched the job so that the metadata is consistent.

In the SHiELD and AM4 random CO2 runs, the new CM4 piControl and 1pctCO2 runs, and in fact [this AMIP constant CO2 run](https://github.com/ai2cm/ace2s-shield-plus-paper/blob/b271da1cb96a5e88da176a6569dcc35a481f6e6b/SHiELD-simulations/SHiELD/config/AMIP-production-C96-constant-CO2.yaml#L176-L187), we save out the CO2 as an interval average for this reason (and use it as a next step forcing in ACE).

scripts/data_process/configs/shield-amip-constant-co2-c96-4deg-8layer.yaml
spencerkclark: We have not been that consistent, e.g. [we do not do so in the original AMIP dataset](https://github.com/ai2cm/ace/blob/349aa906c7f18206c7662359ad6ccb5d6b056e12/scripts/data_process/configs/shield-amip-ensemble-c96-1deg-8layer.yaml#L5-L12) among others.  I can see why it might be helpful at a glance, however, so I updated the path.

To be honest the stats for this particular dataset have been somewhat of an afterthought. If we have a plan of using them for anything it would be helpful to know, and it might inform the period we compute them over.  This run starts in 1979, so by default I have just computed the stats over the full set of whole years from the run.

PR #1150, opened 2026-05-11
Add typed metric configs for OneStepAggregatorConfig

mcgibbon:
Applies the typed metric config pattern (from the InferenceEvaluatorAggregator refactoring in #1123, #1126) to the validation `OneStepAggregator`. This makes validation metrics explicitly configurable via YAML and extensible without adding new boolean flags.

Changes:
- `fme.ace.aggregator.one_step.deterministic.OneStepDeterministicAggregator`: accept pre-built sub-aggregators via `Mapping[str, Aggregator]` instead of constructing them internally
- `fme.ace.aggregator.one_step.build_context`: new module with `OneStepBuildContext`, `OneStepMetricBuildResult`, `MetricNotSupportedError`
- `fme.ace.aggregator.one_step.reduced.OneStepMeanMetricConfig`, `OneStepSnapshotMetricConfig`, `OneStepMapMetricConfig`, `OneStepSpectrumMetricConfig`, `OneStepEnsembleMetricConfig`: typed metric config classes with `get_name()`/`build()` methods
- `fme.ace.aggregator.one_step.main.OneStepAggregatorConfig`: hierarchical config with named fields (`mean_denorm`, `mean_norm`, `power_spectrum`, `snapshot`, `mean_map`, `ensemble`) with `enabled`/`strict` flags, matching the `InferenceEvaluatorAggregatorConfig` pattern
- `fme.ace.aggregator.one_step.main.LegacyFlagOneStepAggregatorConfig`: preserves old boolean-flag interface with deprecation warning
- `fme.ace.train.train_config.TrainConfig`, `fme.ace.train.train.AggregatorBuilder`: accept `OneStepAggregatorConfig | LegacyFlagOneStepAggregatorConfig`
- `fme.ace.inference.evaluator`: rename scoped `aggregator` variable to `val_aggregator` to fix pre-existing mypy type narrowing issue

**Note on inference metric defaults:** `HistogramMetricConfig`, `VideoMetricConfig`, and `SeasonalMetricConfig` changed their class-level `enabled` default from `True` to `False`. This is a cleanup — `InferenceEvaluatorAggregatorConfig` already overrode these to `enabled=False` via `default_factory` lambdas, so the net behavior of the inference pipeline is unchanged. Code that instantiates these config classes directly (outside of `InferenceEvaluatorAggregatorConfig`) will now get `enabled=False` by default.

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated

PR #1333, opened 2026-07-01
Refactor train entrypoints to build-then-execute

mcgibbon:
Applies the Builder pattern to the `fme/ace/train` and `fme/coupled/train` entrypoints (see the 2026-06-24 builder-pattern audit): the shared `Trainer` no longer reads a configuration object's fields, and the entrypoints become thin build-then-execute shells. The `Trainer` is shared by ace and coupled, so both entrypoints are migrated together per the mirror rule.

Changes:
- `fme.core.generics.trainer`: `Trainer` takes a plain `TrainerParams` leaf dataclass holding only the scalar fields it actually reads, instead of a config; `TrainConfigProtocol` is removed. Each `TrainConfig` grows `get_trainer_params()`.
- `fme.ace.train.train_config` / `fme.coupled.train.train_config`: `build_trainer` becomes a method on `TrainConfig` and constructs its builders internally; the aggregator factories become `InlineValidationConfig`/`InlineInferenceConfig.build_aggregator_factory` methods; `AggregatorBuilder`/`CoupledAggregatorBuilder` move here. `TrainBuilders` is removed — its config-shaped methods fold into `TrainConfig` as private `_get_*` data/stepper/optimization builders; the callback builders, which read only a couple of config scalars but several runtime objects, become module-private `_get_validation_callback`/`_get_validate_stepper_callback`/`_get_inference_callback` functions with explicit parameters. The dead `atmosphere_timestep`/`ocean_timestep` properties on the coupled side are dropped.
- `fme.ace.train.train` / `fme.coupled.train.train`: thinned to process side-effects (IO, logging, distributed) plus `config.build_trainer()`; config is immutable at the entrypoint (`main` rebinds `resume_results` via `dataclasses.replace` instead of mutating it). `run_train`/`run_train_from_config` collapse into a single `run_train(config)`.

Typing-only (no runtime change): validation-entry tuples are typed with `GriddedData` rather than `GriddedDataABC`, matching what the data builders produce; the `build_aggregator_factory` methods return `Callable[[], OneStepAggregator]` / `Callable[[], InferenceEvaluatorAggregator]` rather than `Callable[[], Any]`.

Behavior-preserving.


fme/coupled/train/train.py
```diff
@@ -8,280 +8,8 @@
...
-from fme.coupled.train.train_config import (
-    InlineInferenceConfig,
-    InlineValidationConfig,
-    TrainBuilders,
-    TrainConfig,
-)
-from fme.coupled.typing_ import CoupledTensorMapping
-
-
-def get_validation_callback(
```
mcgibbon, line 40: Can you separate this commit into one that purely moves these functions to train_config.py near where they'll be refactored onto the config (but makes no modification to the code), and a separate commit that actually does that refactor? I'd like it to be clear what changed in these functions, if anything.

Similar comment for the ace code, do it to both in the same commit(s).

mcgibbon: Claude: Done. Split into two commits (applied to both ace and coupled in the same commits):

- `c61ece5c3` **pure move** — relocates `build_trainer`, the `get_*_callback` helpers, the `_make_*_aggregator_factory` helpers, and `AggregatorBuilder`/`CoupledAggregatorBuilder` from `train.py` into `train_config.py` verbatim (re-exported from `train.py` so importers/tests are untouched). Verified the moved blocks are byte-identical to the originals (`diff` is empty), so this commit changes no logic.
- `06b7396a7` **refactor** — turns those free functions into `TrainConfig` methods, moves each aggregator factory onto the entry config that owns its fields, thins the entrypoints to side-effects + `config.build_trainer(builders)`, and makes the config immutable via `dataclasses.replace`. This is where the real changes are visible.

Also addressed pre-review nit (1): the callback-wiring tests now call the methods on a real `TrainConfig` (via the existing `_make_train_config*` helpers) instead of an unbound method with a mock `self`.

(Force-pushed the rewritten history.)

fme/ace/train/train.py
```diff
@@ -381,7 +93,7 @@ def run_train(builders: TrainBuilders, config: TrainConfig):
         logging.info(
             f"Resuming training from results in {config.resume_results.existing_dir}"
         )
-    trainer = build_trainer(builders, config)
+    trainer = config.build_trainer(builders)
```
mcgibbon: Can we remove the `builders` argument by constructing TrainBuilders privately inside the config? It seems like TrainBuilders is really private to that context. Correct me if I'm wrong.

fme/ace/train/train_config.py
```diff
@@ -73,6 +90,26 @@ def __post_init__(self):
     def using_labels(self) -> bool:
         return self.loader.using_labels
 
+    def build_aggregator_factory(
```
mcgibbon, line 95: Claude: Moved verbatim from the free `_make_ace_validation_aggregator_factory`. The only change is `entry_config.` -> `self.` (it is now a method on the entry config). No behavioral change.

fme/ace/train/train_config.py
```diff
@@ -152,6 +189,31 @@ def get_inference_data(
             )
         return data
 
+    def build_aggregator_factory(
```
mcgibbon, line 194: Claude: Moved verbatim from the free `_make_ace_aggregator_factory`. Only change is `entry_config.` -> `self.`. No behavioral change.

fme/ace/train/train_config.py
```diff
@@ -404,6 +466,240 @@ def get_inference_epochs(self) -> list[int]:
...
+            checkpoint_save_epochs=self.checkpoint_save_epochs,
+            ema_checkpoint_save_epochs=self.ema_checkpoint_save_epochs,
+            evaluate_before_training=self.evaluate_before_training,
+            save_best_inference_epoch_checkpoints=(
+                self.save_best_inference_epoch_checkpoints
+            ),
+            lr_tuning=self.lr_tuning,
+        )
+
+    def get_validation_callback(
```
mcgibbon: Claude: Beyond becoming a method: the aggregator factory is now built via `entry_config.build_aggregator_factory(...)` instead of the free `_make_ace_validation_aggregator_factory(entry_config=entry_config, ...)`, and `save_per_epoch_diagnostics`/`output_dir` are read from `self` rather than passed as parameters. No other change.

fme/ace/train/train_config.py
```diff
@@ -404,6 +466,240 @@ def get_inference_epochs(self) -> list[int]:
...
+                    save_per_epoch_diagnostics=self.save_per_epoch_diagnostics,
+                    output_dir=self.output_dir,
+                ),
+                weight=entry_config.weight,
+            )
+            for entry_config, data, name in validation_entries
+        ]
+        return build_validation_callback(tasks=tasks, stepper=stepper)
+
+    def get_validate_stepper_callback(
```
mcgibbon: Claude: No behavioral change. Only the docstring/comment was re-wrapped and `validate_using_ema` is now read from `self` (via a local) instead of a parameter. The aggregator build and weighted-loss logic are identical.

fme/ace/train/train_config.py
```diff
@@ -404,6 +466,240 @@ def get_inference_epochs(self) -> list[int]:
...
+                )
+                if entry_config.weight > 0:
+                    summary = aggregator.get_summary(label=name)
+                    if summary.loss is not None:
+                        weighted_loss += entry_config.weight * summary.loss
+            return weighted_loss
+
+        return validate_stepper
+
+    def get_inference_callback(
```
mcgibbon: Claude: Beyond becoming a method: the aggregator factory is now `entry_config.build_aggregator_factory(...)` instead of the free `_make_ace_aggregator_factory(entry_config=entry_config, ...)`; `output_dir`/`save_per_epoch_diagnostics` come from `self`. No other change.

fme/ace/train/train_config.py
```diff
@@ -404,6 +466,240 @@ def get_inference_epochs(self) -> list[int]:
...
+                )
+            )
+
+        return build_inference_callback(
+            tasks=tasks,
+            inference_epochs=inference_epochs,
+            stepper=stepper,
+        )
+
+    def build_trainer(self, builders: "TrainBuilders") -> Trainer:
```
mcgibbon: Claude: One real change beyond method-ification: `dataset_info.update_variable_metadata(variable_metadata)` is now computed once into `updated_dataset_info` and reused for the aggregator builder and the validation/validate-stepper callbacks, instead of being recomputed inline at each of the three call sites. `update_variable_metadata` returns a fresh object with no side effects, so behavior is unchanged. The `get_*_callback` calls are now `self.` with the args that moved onto `self` dropped; the remaining line changes are cosmetic wraps.

fme/coupled/train/train_config.py
```diff
@@ -151,6 +173,36 @@ def __post_init__(self):
             self.aggregator.log_global_mean_time_series = False
             self.aggregator.log_global_mean_norm_time_series = False
 
+    def build_aggregator_factory(
```
mcgibbon, line 175: Claude: Moved verbatim from the free `_make_coupled_aggregator_factory`. Only change is `entry_config.` -> `self.`. No behavioral change.

fme/coupled/train/train_config.py
```diff
@@ -350,6 +402,223 @@ def get_inference_epochs(self) -> list[int]:
             return []
         return sorted(set().union(*epoch_sets))
 
+    def get_validation_callback(
```
mcgibbon: Claude: Structural change beyond method-ification: the free `_make_coupled_validation_aggregator_factory(...)` was inlined into a nested `make_factory(name)` closure here. Because the coupled validation aggregator does not depend on any entry-config field, it was not moved onto `InlineValidationConfig` (this is the ace/coupled asymmetry called out in pre-review nit 2). The `OneStepAggregator(...)` construction itself is identical; `save_per_epoch_diagnostics`/`output_dir` now come from `self`.

fme/coupled/train/train_config.py
```diff
@@ -350,6 +402,223 @@ def get_inference_epochs(self) -> list[int]:
...
+                name=name,
+                data=data,
+                aggregator_factory=make_factory(name),
+                weight=entry_config.weight,
+            )
+            for entry_config, data, name in validation_entries
+        ]
+        return build_validation_callback(tasks=tasks, stepper=stepper)
+
+    def get_validate_stepper_callback(
```
mcgibbon: Claude: No behavioral change. Comment re-wrap plus `validate_using_ema` read from `self` via a local. Aggregator build and weighted-loss logic identical.

fme/coupled/train/train_config.py
```diff
@@ -350,6 +402,223 @@ def get_inference_epochs(self) -> list[int]:
...
+                )
+                if entry_config.weight > 0:
+                    summary = aggregator.get_summary(label=name)
+                    if summary.loss is not None:
+                        weighted_loss += entry_config.weight * summary.loss
+            return weighted_loss
+
+        return validate_stepper
+
+    def get_inference_callback(
```
mcgibbon: Claude: Beyond becoming a method: aggregator factory is now `entry_config.build_aggregator_factory(...)` instead of the free `_make_coupled_aggregator_factory(...)`; `output_dir`/`save_per_epoch_diagnostics` from `self`. `dataset_info` is still passed through. No other change.

fme/coupled/train/train_config.py
```diff
@@ -350,6 +402,223 @@ def get_inference_epochs(self) -> list[int]:
...
+                )
+            )
+
+        return build_inference_callback(
+            tasks=tasks,
+            inference_epochs=inference_epochs,
+            stepper=stepper,
+        )
+
+    def build_trainer(self, builders: "TrainBuilders") -> Trainer:
```
mcgibbon: Claude: No behavioral change beyond method-ification. The `get_*_callback` calls are now `self.` with the moved-to-`self` args dropped, plus a couple of cosmetic line wraps. (Unlike the ace side, `dataset_info` was already computed once here, so there is no hoisting change.)

fme/ace/train/train.py
mcgibbon: Claude: You are right — `TrainBuilders` is only ever `TrainBuilders(config)` (grepped: never subclassed, never injected with anything other than the config), so it is private to this context. Fixed in 31c3ab834: `build_trainer(self)` now constructs `TrainBuilders(self)` internally. That made the `builders` parameter on `run_train` dead, so it is dropped too, which collapses `run_train`/`run_train_from_config` into a single `run_train(config)`. Mirrored on the coupled side; tests updated; behavior-preserving (fast suite + integration train tests pass).

fme/ace/train/train.py
```diff
@@ -381,7 +87,7 @@ def run_train(builders: TrainBuilders, config: TrainConfig):
         logging.info(
             f"Resuming training from results in {config.resume_results.existing_dir}"
         )
-    trainer = build_trainer(builders, config)
+    trainer = config.build_trainer()
```
mcgibbon, line 90: nit: rename to `config.build()`

jpdunc23: I would maybe agree if the config class were called `TrainerConfig` and not `TrainConfig`, but with `TrainConfig` I prefer `build_trainer()`. `build_trainer()` also aligns with the convention of other config `build_*` methods like `build_aggregator_factory()`. Even if it where `TrainerConfig`, I still prefer the descriptiveness of `build_*()` naming since `config.build_trainer()` is much more informative than `config.build()`.

Either way, I think we should aim for consistency.

fme/ace/train/train.py
jpdunc23, line 1: With this file now gutted, a natural follow-on PR would move the remaining contents to `__main__.py` and delete it. The copyright notice at the top of the file is also no longer applicable (which may already have been true long ago when we added `fme/core/generics/trainer.py`).

fme/coupled/train/train_config.py
```diff
@@ -151,6 +172,141 @@ def __post_init__(self):
...
+    tasks: list[ValidationTask] = [
+        ValidationTask(
+            name=name,
+            data=data,
+            aggregator_factory=make_factory(name),
+            weight=entry_config.weight,
+        )
+        for entry_config, data, name in validation_entries
+    ]
+    return build_validation_callback(tasks=tasks, stepper=stepper)
```
jpdunc23, line 234: Why is the implementation here different than in `fme.ace`? Should the `fme.coupled` `InlineValidationConfig` get a `build_aggregator_factory()` method? But it looks as though the `fme.coupled` `InlineValidationConfig` doesn't include `aggregator`, and https://github.com/ai2cm/ace/pull/1150 hard-coded `config = OneStepAggregatorConfig()` in the `fme.coupled` `OneStepAggregator.__init__`. If I understand correctly, it seems coupled inline validation entries do not currently have per-entry aggregator configurability.

Since you're already refactoring things, it seems straightforward in this PR to bring `fme.coupled` inline validation configurability up to speed with `fme.ace`.

jpdunc23: More generally, mirroring these things consistently across `fme.coupled` and `fme.ace` is clearly still a pain point that agents are not avoiding without hand-holding.

I wonder if now is a good time for a follow-on PR to move toward generic configuration of the inline validation and inference aggregators would save us some headaches. There may be some pain (e.g., `fme.coupled` `InlineInferenceConfig` is missing `n_ensemble_per_ic`; what to do about `n_coupled_steps` vs `n_forward_steps`?), but I don't think "mirroring" is a sustainable model.

jpdunc23 (commented):
Thanks, I agree this cleanup / reorg is a step in the right direction.

fme/ace/train/train.py
mcgibbon: I agree, though yes let's do it as a separate branch and PR.

fme/coupled/train/train_config.py
mcgibbon: Generally agreed. I think that we should look at these possibilities after this PR is merged. Claude, can you reply to this thread with any potential issues/challenges you already see with that follow-on refactor? Don't actually plan it out, just look for potential pain points we should have in mind.

fme/ace/train/train.py
mcgibbon: OK, I'm good with that. Let's keep build_trainer.

fme/coupled/train/train_config.py
mcgibbon: Claude: Potential pain points for a generic-inline-aggregator-config follow-on, without planning it:

1. **Coupled `OneStepAggregator` is a composite, not a single aggregator.** It builds one per-component `OneStepAggregator` for ocean and one for atmosphere (`fme/coupled/aggregator.py`), and #1150 hard-coded `config = OneStepAggregatorConfig()` inside `__init__`. So exposing validation-aggregator config on the coupled `InlineValidationConfig` is not a 1:1 mirror of `fme.ace` — you first have to decide the *shape*: one shared `OneStepAggregatorConfig` applied to both components, or a per-component (ocean/atmosphere) pair. It also threads component-split params that ace does not (`ocean_channel_mean_names`/`atmosphere_channel_mean_names`, and `loss_scaling.ocean`/`.atmosphere`).

2. **Two-timescale horizon.** ace inference has a single `n_forward_steps`; coupled `n_coupled_steps` fans out to two horizons — `n_timesteps_ocean = n_coupled_steps + ocean.n_ic_timesteps` and `n_timesteps_atmosphere = n_coupled_steps * n_inner_steps + atmosphere.n_ic_timesteps`. The two `InferenceEvaluatorAggregatorConfig.build` signatures differ accordingly (coupled takes both timestep counts + `initial_time` + separate ocean/atmosphere normalizers; ace takes one of each). Any shared config has to abstract "horizon" and "normalizer" over the single-vs-dual-component structure — that abstraction boundary is exactly where naive mirroring breaks.

3. **`n_ensemble_per_ic` has no coupled equivalent.** Before adding it to coupled `InlineInferenceConfig`, its semantics for a two-component stepper need deciding — not just a field copy.

4. **The validate-stepper (LR-tuning) callback constructs `OneStepAggregator(...)` inline**, separate from the validation factory. If the validation aggregator becomes configurable, the config must reach both that factory *and* this second construction site, or LR-tuning validation silently diverges from the logged validation aggregator.

5. **LegacyFlag config variants.** ace’s `InlineValidationConfig.aggregator`/`InlineInferenceConfig.aggregator` are unions with `LegacyFlag*` compat types; coupled has none. Parity means explicitly deciding whether coupled needs the legacy-flag path (likely not, since coupled is newer) rather than mirroring it by default.

Net: the recurring "mirror" failures come from #1 and #2 — coupled’s two-component/two-timescale structure means the ace and coupled aggregators are genuinely different classes, so real parity is a shared abstraction over that structure, not a field-by-field copy.

mcgibbon: > Since you're already refactoring things, it seems straightforward in this PR to bring fme.coupled inline validation configurability up to speed with fme.ace.

Handled in https://github.com/ai2cm/ace/pull/1346/changes, please review and I'll merge it separately. I wanted to keep this PR no-change to the config and behavior, because the changes are already extensive.

jpdunc23 (commented):
Nit: The description says "The work is split into cleanly separable commits, each passing tests on its own", which becomes stale once we squash-and-merge as a single commit on main. Maybe worth noting as a PR comment, rather than in the PR description feeding into the final squash-and-merge commit message. There is some additional agent-iteration cruft in the description, e.g. "(review response)".

jpdunc23 (approved):
LGTM

PR #1335, opened 2026-07-02
Carry corrector deltas onto prediction data and add a step-diagnostics netCDF writer

jpdunc23:
Correctors declare their per-step effect as a `delta` mapping on `StepOutput` (#1328), but that payload is discarded at the `Stepper.predict` boundary, so no inference output can show how much the model leans on its corrector. This PR carries the delta series out of `predict` onto the prediction data and proves the carriage end-to-end with an opt-in netCDF writer. The pre-correction network output is recoverable as `output − delta`, so no separate raw-output series is carried. Aggregator metrics and normalizer wiring are deliberately not included here.

Changes:
- **`StepDiagnostics`** (new, `fme/core/step/`): opaque container modeled on `StepperState`, holding the stacked forward-step-aligned delta series (an empty series is valid). Attached as an optional `step_diagnostics` field on `BatchData` and `PairedData` and forwarded through the device/ensemble structure-preserving ops. The time-windowing methods, `subset_names`, and `scatter_spatial` raise if asked to carry it — slicing the time axis would silently misalign the series, and the other two are input-side ops that never see a diagnostics-bearing batch.
- **`StepOutput.stack_diagnostics`** (new classmethod): stacks the per-step deltas and returns the `StepDiagnostics` itself (or `None` when no step carried a delta), so `predict` never sees corrector internals. Attached at `predict`'s final reconstruction; coupled runs are follow-up work (threading `CoupledStepper`'s component-step predictions is a separate PR), so `CoupledStepper` leaves the field unset.
- **Masking at the step seam**: the new `CorrectorDiagnostics.apply_output_masking` masks the delta with the same NaN-fill output masker as the prediction, so off-mask cells report NaN rather than a spurious correction. The output post-process surface is renamed to masking-specific types: a `SpatialMasking` alias returned by `build_output_spatial_masker`, an `output_masking` stepper parameter, and deletion of the now-unused `NullPostProcessFn`/`PostProcessFnType`.
- **Writer**: with `save_step_diagnostics: true` on `DataWriterConfig` (wired in both `build()` and `build_paired()`), inference writes `autoregressive_step_diagnostics.nc` via a new `StepDiagnosticsWriter` that consumes an `xarray.Dataset` from `StepDiagnostics.to_dataset(time)`. It contains the per-step correction (already in physical units) for exactly the corrector-modified variables, honoring `save_names` and time coarsening. When no corrector modified anything, no diagnostics are appended (the file holds no diagnostic variables). The flag is off by default and the carriage has no other consumer, so default inference output is unchanged. End-to-end coverage of the writer through the real inference loop is deferred alongside the coupled follow-up; the seam is covered pairwise (stepper → `BatchData` carriage, and `PairedData` → writer via the `DataWriterConfig` builders).

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+    def new_on_cpu(cls, ..., step_diagnostics: StepDiagnostics | None = None) -> "BatchData": ...
+    @classmethod
+    def new_on_device(cls, ..., step_diagnostics: StepDiagnostics | None = None) -> "BatchData": ...
+
+    def __post_init__(self):  # CHANGED — also validate step_diagnostics.sample_dim_size()
+        ...
+
+    def subset_names(self: SelfType, names: Collection[str]) -> SelfType:
+        # CHANGED — pass step_diagnostics through UNCHANGED (delta is its own
+        # sparse name set; not filtered to `names`)
```
jpdunc23: Maybe this should also RAISE like the time-windowing methods? Or is this used while `step_diagnostics` are attached?

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+class DataWriterConfig:
+    save_correction_files: bool = False  # NEW — write autoregressive_corrections.nc
+
+    def build_paired(self, ...) -> "PairedDataWriter":
+        # CHANGED — when save_correction_files, build a single-source
+        # RawDataWriter(label="autoregressive_corrections", save_names=self.names, ...),
+        # wrapped with self.time_coarsen.build(...) (single-source, NOT
+        # build_paired — there is no target series), and pass it to
+        # PairedDataWriter as correction_writer.
+        ...
```
jpdunc23: `DataWriterConfig.build()` also needs to be modified to support the new `save_correction_files` attribute.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+
+## `fme/core/corrector/output.py` (modified)
+
+```python
+@dataclasses.dataclass
+class CorrectorDiagnostics:
+    delta: TensorMapping = dataclasses.field(default_factory=dict)
+
+    def apply_output_process_func(  # NEW — map the stepper's output masker over delta
+        self, func: Callable[[TensorMapping], TensorDict]
```
mcgibbon: Question: How can we guarantee that an arbitrary `func` which post-processes output values is valid to apply on diagnostic values? For example, there's currently nothing restricting `func` from computing derived output values, and equations applied might not be valid for deltas.

One potential fix is to specialize `output_process_func` according to the restrictions you're placing on it with this usage, and the current way we use this feature. For example, if we truly only use this to mask outputs, you could rename this method `apply_output_masking(`.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+## `fme/ace/data_loading/step_diagnostics.py` (new)
+
+Opaque per-sample diagnostics container, modeled on `StepperState`. Holds the
+stacked per-step correction series, aligned with the prediction `data` along
+`(sample, time, ...)`. When attached it always carries a non-empty `delta`
+(`Stepper.predict` attaches `None` instead of an empty container), so
+`step_diagnostics is None` is the single "no correction" gate. It deliberately
+has **no** time-slice or prepend-time op — the series is attached only after
+time-windowing is finished, and the time-touching `BatchData` methods guard
+against carrying it (below).
```
mcgibbon: This is a lot of focus on `delta` always being non-empty. Can we seamlessly support the delta being empty? Otherwise there's no way to distinguish between "we're applying corrections but explicitly none of them are delta-type corrections", and "there's no corrector". Granted this is moot with the current planned code state where delta is always present for all corrections, but if there's no reason not to support delta being empty (even for a hypothetical case that a corrector has been explicitly configured with all corrections turned off), I think we should support it now.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+
+```python
+@dataclasses.dataclass
+class StepDiagnostics:
+    delta: TensorMapping  # stacked per-step CorrectorDiagnostics.delta, (sample, time, ...)
+
+    def to_device(self) -> "StepDiagnostics": ...
+    def to_cpu(self) -> "StepDiagnostics": ...
+    def pin_memory(self) -> "StepDiagnostics": ...
+    def scatter_spatial(self, global_img_shape: tuple[int, int]) -> "StepDiagnostics": ...
```
mcgibbon: Describe the behavior of scatter_spatial. I trust there is a use/need for it (and you're not just including this because it is included on similar objects), but I can't imagine what it is in this moment.

pr-plan.md
mcgibbon: Do we need this? subset_names is generally used on batch data inputs. Dropping the names feels like the natural outcome.

I'd suggest either only keeping names that would be present, or raising an exception on subset_names if StepDiagnostics is present (since the behavior is kind of undefined or at least, we don't think we want to allow this to happen), and require the StepDiagnostics to be cleared before it's passed in to the next call (I'd suggest with a `BatchData.pop_step_diagnostics()` which you use to retrieve the values instead of direct access).

When enabling Step-registered aggregators, I'll also be considering options that involve not attaching the StepDiagnostics to BatchData at all, as I think this is going to be necessary. I don't suggest doing that now, because it will look quite artificial and awkward.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+        ...
+
+    def subset_names(self: SelfType, names: Collection[str]) -> SelfType:
+        # CHANGED — pass step_diagnostics through UNCHANGED (delta is its own
+        # sparse name set; not filtered to `names`)
+        ...
+
+    # CHANGED — time-touching methods RAISE when step_diagnostics is not None.
+    # delta is time-indexed and these would silently misalign it; no current
+    # caller windows a diagnostics-bearing batch, so a call here is a bug:
```
mcgibbon: Yeah that seems reasonable, maybe do it for subset_names too.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+        ...
+
+    def predict(self, ...) -> tuple[BatchData, PrognosticState]:
+        # CHANGED — collect each StepOutput.corrector_diagnostics.delta from the
+        # prediction generator, stack forward-step-aligned via
+        # stack_list_of_tensor_dicts(..., time_dim=1), and attach at the closing
+        # BatchData.new_on_device(...) reconstruction — i.e. AFTER the
+        # prepend → compute_derived → remove_initial_condition dance, where data
+        # and series are both n_forward_steps long. Attach None when the stacked
+        # series has no keys (no corrector, or nothing modified this rollout).
```
mcgibbon: This is not good. Create a StepOutput method or class method to encapsulate this concern, and use it instead of `stack_list_of_tensor_dicts`. We don't want the Stepper.predict to depend on implementation details of StepOutput, that's tight coupling which would make it impossible to later allow these details to change depending on the Step chosen.

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+        ...
+```
+
+`CoupledStepper` is unchanged: it leaves `step_diagnostics` at its `None`
+default (no correction file for coupled runs).
```
mcgibbon: Why not?

pr-plan.md
```diff
@@ -0,0 +1,278 @@
...
+
+## `fme/ace/inference/data_writer/main.py` (modified)
+
+```python
+@dataclasses.dataclass
+class DataWriterConfig:
+    save_correction_files: bool = False  # NEW — write autoregressive_corrections.nc
```
mcgibbon: Please abstract away any low-level implementation details of the StepDiagnostics when you write this. That means, name this `save_step_diagnostics`, and possibly define new API on `StepDiagnostics` that returns data in the form you need it to handle writing netCDF files. This DataWriter should not need to know low level implementation details of StepDiagnostics.

If we need to be able to save separate netCDF files with different metric types (for example to be able to store deltas and other metrics with the variable names just being the data variable names and not `{name}_delta`, then write the API in a way that allows the StepDiagnostics to define multiple such files for writing by this writer.

One potential form of this would be to have the StepDiagnostics class itself provide references to this DataWriterConfig class and its DataWriter implementation. That would allow the inference files to stay independent of this dependence. If you do that, just make sure the configuration still gets `__post_init__` validated correctly, by explicitly initializing this DataWriterConfig (since the inference config will not be allowed to reference it, and would instead call it `dict[str, Any]`.

Propagate this naming, it's a step diagnostics writer, not strictly a correction writer (even if that's what's in the step diagnostics right now).

pr-plan.md
jpdunc23: Sounds good. I initially had named this `apply_output_masking` and claude objected for some reason, but let's go back to that. I agree it's better to be explicit here.

pr-plan.md
jpdunc23: > "we're applying corrections but explicitly none of them are delta-type corrections"

As you say this is moot, since `delta = corrected - network_output` is not a "type" of correction but applies for any correction.

> if there's no reason not to support delta being empty (even for a hypothetical case that a corrector has been explicitly configured with all corrections turned off), I think we should support it now

Not sure, we should look into this.

pr-plan.md
jpdunc23: I think this is just mirroring the existing `BatchData.scatter_spatial`, so when it is called it will call `StepDiagnostics.scatter_spatial` so that the `delta`s are also sliced to the local spatial chunk.

pr-plan.md
jpdunc23: Agreed: raise an error if StepDiagnostics are non-None.

pr-plan.md
jpdunc23: Agreed, a method on `StepOutput` sounds cleaner.

pr-plan.md
jpdunc23: I've been planning the `fme.coupled` port for a later PR. That said, maybe the `fme.coupled` changes are more minor than I'm thinking and can be included here. We should consider this.

pr-plan.md
jpdunc23: > Please abstract away any low-level implementation details of the `StepDiagnostics` when you write this. That means, name this `save_step_diagnostics`, and possibly define new API on `StepDiagnostics` that returns data in the form you need it to handle writing netCDF files.

`save_step_diagnostics` sounds good to me, though if we expand the set of things being carried on `StepDiagnostics` then we would probably want a more granular config that allows for saving a subset of the diagnostics. Easy refactor in a later PR.

> One potential form of this would be to have the `StepDiagnostics` class itself provide references to this `DataWriterConfig` class and its `DataWriter` implementation. That would allow the inference files to stay independent of this dependence

Not sure I understand this suggestion. That said, with the `save_step_diagnostics` boolean I don't think we currently need something quite so complicated at this point; the `StepDiagnostics` data writer helper can simply return `delta` (maybe together with the associated `time`) so `DataWriter` won't need to reach so far into the `StepDiagnostics`.

pr-plan.md
mcgibbon: We should worry that perhaps the func being applied is doing more than just output masking. If that's the case, Claude should flag this instead of blindly following orders.

pr-plan.md
mcgibbon: Well, the one kind of weird case that makes this non-moot is that you can technically configure a corrector that does nothing, and it should be perfectly valid to do so, but this would break it.

pr-plan.md
mcgibbon: Can we raise instead when StepDiagnostics is present like you do in the other routines? I don't see why we should ever scatter StepDiagnostics. It's not something we read centrally and then scatter to ranks. I think I'm later going to have to remove StepDiagnostics off of BatchData anyways, and I'd like to avoid vestigial unused code like this method if/when that happens.

pr-plan.md
mcgibbon: I'd recommend the writer helper going as far as returning an xarray.Dataset object (which is kind of just a convenient container for a data dict combined with a time axis). Don't worry about the more involved suggestion, I'll tackle it when I let these types depend on Step.

pr-plan.md
jpdunc23: Resolved by making `subset_names` raise when `step_diagnostics` is set (see the `BatchData` guard list in pr-plan.md). Checked call sites: `subset_names` is only used on input-side data (forcing subsetting in `predict`, inference persistence data, and internally by `get_start`/`get_end`), all before diagnostics are attached.

pr-plan.md
jpdunc23: Done — `save_step_diagnostics` (renamed per the writer-naming thread) is now wired in both `build_paired()` and `build()`, so forcing-only inference runs also get the file.

pr-plan.md
jpdunc23: Renamed to `apply_output_masking`, taking the masking op specifically rather than an arbitrary func. Verified the stepper's output process func is only ever the NaN-fill spatial masker (`build_output_spatial_masker`) or the `NullPostProcessFn` no-op today; the method's name, parameter type, and inline contract now pin that restriction, and the NaN-fill dependency (why masking a difference is valid at all) is documented at the method.

pr-plan.md
jpdunc23: Plan updated: an empty `delta` is now valid on `StepDiagnostics` — nothing asserts non-emptiness and every op is a safe no-op — so a corrector configured with all corrections off is fine. `Stepper.predict` still attaches `None` when nothing was modified, so 'no file when nothing corrected' semantics are unchanged; distinguishing 'corrector ran but modified nothing' from 'no corrector' is deferred until a consumer needs the distinction.

pr-plan.md
jpdunc23: Agreed — dropped `scatter_spatial` from `StepDiagnostics`, and `BatchData.scatter_spatial` now raises when diagnostics are present, joining the other guarded methods.

pr-plan.md
jpdunc23: `subset_names` now raises when `step_diagnostics` is set. Holding off on `pop_step_diagnostics()` until a caller actually needs to window or re-subset a diagnostics-bearing batch — today the only consumer reads the field at the writer dispatch.

pr-plan.md
jpdunc23: Done — `subset_names` (and `scatter_spatial`) joined the raise-guard list.

pr-plan.md
jpdunc23: Added `StepOutput.stack_corrector_deltas(outputs)` in `fme/core/step/output.py`; `Stepper.predict` calls that instead of `stack_list_of_tensor_dicts`, so the stacking stays an implementation detail of `StepOutput`.

pr-plan.md
jpdunc23: Scoped this: the coupled generator's `ComponentStepPrediction` drops diagnostics at the component-step seam, so carrying deltas through `CoupledStepper` means threading `ComponentStepPrediction`, per-realm stacking, and `CoupledBatchData` — a real chunk of work rather than a minor add. Keeping it out of this PR and stating it in the plan as explicit follow-up work (a coupled-carriage PR) instead of silently unsupported.

pr-plan.md
jpdunc23: Adopted throughout: the flag is `save_step_diagnostics`, the writer is a `StepDiagnosticsWriter` writing `autoregressive_step_diagnostics.nc`, and per your recommendation it consumes an `xarray.Dataset` from a new `StepDiagnostics.to_dataset(time)` — the container's only data-export API — so the writer knows nothing about `StepDiagnostics` internals.

pr-plan.md
```diff
@@ -0,0 +1,352 @@
...
+
+## `fme/core/corrector/output.py` (modified)
+
+```python
+@dataclasses.dataclass
+class CorrectorDiagnostics:
+    delta: TensorMapping = dataclasses.field(default_factory=dict)
+
+    def apply_output_masking(  # NEW — map the stepper's output spatial masker over delta
+        self, masking: PostProcessFnABC
```
mcgibbon: Please propagate the renaming through all PostProcessFn-related objects and methods. We want to be sure it's obvious if this is incorrect, e.g. by there being a masking_function = dataset_info.compute_derived_variables (if that were to be an issue, my understanding is you expect it isn't).

pr-plan.md
```diff
@@ -0,0 +1,352 @@
...
+        ...
+```
+
+## `fme/core/step/output.py` (modified)
+
+```python
+@dataclasses.dataclass
+class StepOutput:
+    @classmethod  # NEW — encapsulate how per-step corrector deltas stack into a series
+    def stack_corrector_deltas(
```
mcgibbon: This breaks the contract of the corrector internals being hidden. Likely solution is to return a StackedStepOutput. Note this return value is what's going to have to be responsible for providing data to the data writer, so what we imagined as StepDiagnostics method(s) to give the data writer its information may actually be StackedStepOutput method(s).

Possibly replace each instance of StackedStepOutput with StackedStepDiagnostics in this comment, I'm not sure which is most appropriate.

pr-plan.md
```diff
@@ -0,0 +1,352 @@
+# Carry corrector deltas onto prediction data and add a step-diagnostics netCDF writer
```
mcgibbon: Not at this point in this plan, but if we're going to be using this "pr plan" cycle more regularly, it would help to include a relatively exhaustive list of user stories as part of the plan write-up, and have a validation write-up that verifies each of those user stories are met by the plan (just confirming each one, and writing any complications or tradeoffs down, it doesn't need a text entry for each one). Knowing what choices were made, e.g. what other seemingly-reasonable choices were considered and turned down and why, could also be helpful.

It's common in reviewing these plans to ask myself why a choice was made, this information would make that easier to process and potentially avoid some of the iteration cycles we've been going through if the agent asks itself the questions I've been asking.

Could also consider getting Claude to look at these pre-review reviews, and make its own pre-review steps for PR plans, or additions to your PR-plan creation skill. These are things that are very normal to have in design documents (e.g. https://rinaarts.com/how-to-write-an-effective-design-document/ talks about a "summary of requirements", "Decisions and tradeoffs", "non-goals", etc. and is just the first google result that came up for me), a "lit review" of design documents could help in shaping this skill or the self-review of its output.

mcgibbon (commented):
Significant comments, but I think you're at a stage where you could proceed with this PR. Meta comments are meant for the next PR plan. We could figure out the stacked corrector delta stuff here, but I think if you have Claude writing the PR itself, and it's able to reference these design review comments, it can figure out a good strategy better than it can in this pr-plan context.

pr-plan.md
jpdunc23: `pr-plan.md` is intended as a self-contained API-grounded skeleton doc for quick external reviewer consumption. It is produced only after two rounds of local planning prep and documentation, which includes "summary of requirements", "decisions and tradeoffs", "out-of-scope", an exhaustive list of user stories, etc. See [/to-prd ](https://github.com/mattpocock/skills/blob/main/skills/engineering/to-prd/SKILL.md) and [/to-issues](https://github.com/mattpocock/skills/blob/main/skills/engineering/to-issues/SKILL.md) in [mattpocock/skills](https://github.com/mattpocock/skills) if you're curious about my current local planning process.

Rather, than overwhelm reviewers by pushing those documents (which would also slow down my workflow), I prefer to answer your questions about the PR pre-planning process as they come up in your review. When it is unclear from `pr-plan.md` why a particular decision was made, this is usually helpful feedback to double check the reasoning. 

Personally, I feel the "pr plan" cycle is working to get the feedback I need in a farily efficient manner before spending many local cycles on implementation that will just be thrown out. I'll update my pr plan skill to include a bit more detail on the decision-making process so it is less opaque.

pr-plan.md
jpdunc23: Propagated throughout: new `SpatialMasking = StaticSpatialMasking | NullSpatialMasking` alias in `spatial_masking.py`; `build_output_spatial_masker` (ABC, implementations, and the `HasGetSpatialMask` protocol) now returns `SpatialMasking` instead of a bare `Callable`; the stepper's `output_process_func` param/attr become `output_masking`/`_output_masking` typed as `SpatialMasking`; and `NullPostProcessFn`/`PostProcessFnType` are deleted (the stepper was their only consumer; `NullSpatialMasking` already covers the no-op). A wiring like `output_masking = dataset_info.compute_derived_variables` is now a mypy error. No checkpoint impact — the masking is rebuilt from config on every load. Kept in this PR since `apply_output_masking`'s contract depends on it, but isolated as a standalone, behavior-preserving first commit — happy to split it into a preceding refactor PR if you'd prefer.

pr-plan.md
jpdunc23: Adopted: the stacking classmethod is now `StepOutput.stack_diagnostics(outputs) -> StepDiagnostics | None`, returning the stacked container itself (`None` when no step carried a delta), with `StepDiagnostics` moved to `fme/core/step/` so the return type respects the core→ace import rule. `Stepper.predict` just attaches the returned object, and the writer-facing export (`to_dataset(time)`) lives on it — the corrector internals stay hidden inside `StepOutput`/`StepDiagnostics`.

pr-plan.md

jpdunc23:
Pre-review agent: test-coverage note for the reviewer — there is no end-to-end test that runs the real evaluator/inference loop with `save_step_diagnostics: true` (in particular over multiple windows). The seam is covered pairwise instead: stepper-level tests prove `Stepper.predict` attaches a correct, masked delta series to `BatchData`, and `DataWriterConfig`-level tests prove a diagnostics-bearing `PairedData` round-trips through `autoregressive_step_diagnostics.nc` (including `save_names` filtering and time coarsening). This gap is deliberately deferred alongside the coupled-stepper follow-up, per the PR description. Flagging in case you'd rather see loop-level coverage land in this PR.

fme/core/step/step_diagnostics.py
```diff
@@ -0,0 +1,80 @@
+"""Per-sample step diagnostics carried on prediction data.
+
+``StepDiagnostics`` is an opaque container attached by ``Stepper.predict`` to
+the prediction ``BatchData``. Like ``StepperState``, the structure-preserving
+operations (device moves, ensemble broadcast, pin-memory) apply without
+inspecting its contents. Data consumers have two sanctioned read surfaces:
+``to_dataset`` for serialization (a self-describing, detached CPU export for
+the step-diagnostics writer), and the ``delta`` field directly for in-memory
+consumers that need the tensors on device (e.g. inference metrics).
```
jpdunc23, line 9: Note that the follow-on metrics PR will require the `delta` tensors.

pr-plan.md
mcgibbon: > It is produced only after two rounds of local planning prep and documentation, which includes "summary of requirements", "decisions and tradeoffs", "out-of-scope", an exhaustive list of user stories, etc.

That's great.

> Personally, I feel the "pr plan" cycle is working to get the feedback I need in a farily efficient manner before spending many local cycles on implementation that will just be thrown out.

Yes, I also feel this has been an improvement versus what we were doing before.

> Rather, than overwhelm reviewers by pushing those documents (which would also slow down my workflow), I prefer to answer your questions about the PR pre-planning process as they come up in your review.

I am finding my review of these documents is slowed down by not having more background.

I'm also sometimes curious about what line of reasoning led to Claude making certain decisions or not pre-emptively identifying certain issues in these documents - without any process output, I haven't been able to answer these questions at all. It feels like adding a "PR plan review" with explicit passes like I've been doing for pre-review would be really helpful, but I can't explore it if I don't understand this design process.

Straight up making the `/to-prd` output (after your own review and iteration on that output) visible could be a significant improvement, though it doesn't address "what other seemingly-reasonable choices were considered and turned down and why" and some of the sections will be more verbose than they need to be.

fme/core/step/step_diagnostics.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def sample_dim_size(self) -> int | None:
+        """Return the leading (sample) dim of the delta tensors, or None when
+        the delta is empty.
+        """
+        for tensor in self.delta.values():
+            return tensor.shape[0]
+        return None
+
+    def to_dataset(self, time: xr.DataArray) -> xr.Dataset:
```
mcgibbon, line 61: This PR is large and also quite pristene, so please handle this one issue follow on branch with a follow on PR, which prevents me having to re-review all of the existing changes.

The dataset this exports is a bit problematic in that it's a `step_diagnostics` netcdf whose variables e.g. `air_temperature_0` carry no indication that they're "the correction deltas" for each variable. The solution I proposed during the PR plan review was to support step diagnostics defining multiple named datasets. That means this method returning something that defines multiple named datasets, and each of those datasets getting appropriately handled in the dataset writer above it. The current implementation would define one such dataset - a corrector_deltas dataset.

Technically speaking we could return one such dataset to support our current needs, e.g. `to_named_dataset` that returns `tuple[str, xr.Dataset]`, but then the APIs kind of communicate the wrong intent about the code. It would imply that we want our step diagnostics to have only one dataset and to be able to name it, but the whole purpose behind the corrector supplying the name is to support the need to disambiguate between multiple datasets.

If you really want to hard-code it to one dataset because that's what we're using now I'm fine with that, but I think the APIs are clearer about their intent if they pre-emptively support multiple named datasets, and I expect we may want to take advantage of the ability to write multiple datasets fairly soon (e.g. we may want to write datasets indicating clipping masks, or to save separate sea ice clipping masks for 0 and 1-based values).

mcgibbon (approved):
LGTM, a follow-on requirement to fix an ambiguity this PR introduces but I think it's acceptable to merge it as-is, and would strongly prefer we handle the issue in a follow-on PR if at all (I'm also fine to handle it later, if you want to just know by convention step_diagnostics.nc carries deltas for now).

fme/core/step/step_diagnostics.py
jpdunc23: Thanks for the comment; agreed this is an oversight. Working on a follow-on PR.

PR #1340, opened 2026-07-04
Add GFDL OM4/CM4 ocean dataset pipeline on xarray_beam + Dataflow

jpdunc23:
Adds a self-contained `xarray_beam` pipeline on Dataflow to `scripts/gfdl_om4/` with entrypoint `make dataflow CONFIG=configs/<dataset>.yaml`. Reads native-grid source zarrs, interpolates and rotates velocities to tracer centers, regrids to a Gaussian grid with wetmask normalization, splits vertical levels into suffixed 2D variables, applies the derived-variable and masking conventions, and writes one sharded zarr v3 store per config.

Changes:
- `sea_ice_volume` units use plain m^3. The old store divided by 1000^3 but was mislabeled 1000 km^3.
- `areacello` fixed: exact quadrature areas at the 6371.0 km mean radius; the old areas used the polar radius (~0.5% low).
- Coastal velocity bias fixed: masked C-grid -> tracer interpolation plus face masks, replacing `fillna(0)`.
- `ocean_sea_ice_fraction` computed natively; `hfds_total_area` hoisted from `create_coupled_datasets.py`.
- `ocean_emulators` utilities ported in (no external dependency); regrid weights precomputed once as a versioned artifact.
- Fail-loud checks throughout: level counts, cross-stream time alignment, refusal to overwrite an existing store, smoke tests with output checks, cross-simulation wetmask identity.

- [ ] Tests added


pr-plan.md

scripts/gfdl_om4/pipeline/ocean_emulators_port.py
```diff
@@ -0,0 +1,145 @@
+"""Regridding utilities ported from the ai2cm fork of ocean_emulators
```
jpdunc23:
```suggestion
"""Regridding utilities ported from the ai2cm fork of m2lines/ocean_emulators
```

scripts/gfdl_om4/pipeline/ocean_emulators_port.py
```diff
@@ -0,0 +1,145 @@
...
+    ):
+        raise ValueError("`u` and `v` need to be on the same grid position as `angle`.")
+    theta = np.deg2rad(angle)
+    u_rotated = u * np.cos(theta) - v * np.sin(theta)
+    v_rotated = u * np.sin(theta) + v * np.cos(theta)
+    return u_rotated, v_rotated
+
+
+def _interpolate_right_to_center(
+    da: xr.DataArray, dim: str, center_dim: str, periodic: bool
```
spencerkclark: nit: maybe call the input `right` for maximum clarity
```suggestion
    right: xr.DataArray, dim: str, center_dim: str, periodic: bool
```

scripts/gfdl_om4/pipeline/ocean_emulators_port.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Center i sits between staggered points i-1 and i. Land points (NaN)
+    drop out of the average instead of being filled with zeros, so coastal
+    values are means of ocean neighbors only. Cells with no valid neighbor
+    come out NaN. Along a periodic dimension the first center wraps around;
+    otherwise the edge center uses its single in-domain neighbor.
+    """
+    if da.sizes[dim] != da[dim].size:
+        raise ValueError(f"Expected coordinate for staggered dim {dim}")
+    left = da.roll({dim: 1}, roll_coords=False) if periodic else da.shift({dim: 1})
```
spencerkclark: It could be worth clarifying in a comment or docstring that the non-periodic case simply uses a row of NaNs to form the first "left" boundary, and that this is natural in the case of MOM6, since the southern boundary is Antarctica, which is land and therefore all NaN.

jpdunc23:
## Notes for reviewers

- Deliberate, documented legacy-defect fixes (`areacello` from Gaussian quadrature at R=6371 km, ~0.5% above the old polar-radius areas; masked coastal velocity interpolation instead of fillna(0); explicit land-NaN exemption list; `sea_ice_volume` in `m^3` with divisor and units attr agreeing, where the old store divided to km³ but labeled `"1000 km^3"`).
- The sources' staggered velocities carry coastal faces that are *valid with value exactly 0.0* where MOM6's online z\*-remap unmasked land columns; averaged into the C-grid→center interpolation they would bias coastal speeds low. A one-time setup step (`pipeline/face_masks.py`) scans a source and publishes a per-simulation face-mask artifact (a face is flagged iff exactly 0.0 at every scanned timestep AND it has a dry tracer neighbor, fold-aware on the tripolar seam), verified against an independent census of the source's zero faces (16,960 u / 16,694 v surface faces, identical across piControl and 1pctCO2). Streams opt in via `face_mask_url`; flagged faces are treated as invalid before interpolation, and a per-chunk assertion fails the run if any face matching the fingerprint survives the masks (a stale artifact or changed source).
- Dropping invalid faces means some tracer-wet coastal cells have no valid face on an axis — a wall for that axis, where the model's resolved normal component is identically zero by no-normal-flow. That grid-relative component is set to 0.0 *before* rotation (rotation mixes the components, so filling after would also zero a valid along-wall component, e.g. the along-channel flow of a one-cell-wide channel). For `tauuo`/`tauvo`, whose staggered mask is stricter than `uo`/`vo`'s, this reproduces the legacy zero values at the 1373 affected surface cells (~0.14% of surface ocean); either way every rotated pair keeps the shared tracer-wetmask footprint exactly.
- Two legacy sea-ice conventions were deliberately changed: `UI`/`VI` follow the `HI` convention (zero where `sea_ice_fraction` is zero, NaN over land) instead of the old NaN-where-`EXT`==0 mask — `EXT` is a binary 15% concentration threshold, so that mask discarded velocities over thin ice and produced a time-varying NaN pattern; and fraction units use the GFDL-style `0-1` throughout.

scripts/gfdl_om4/pipeline/run.py
```diff
@@ -0,0 +1,973 @@
...
+  time-varying field equals the masks exactly at every timestep: training
+  assumes NaN exactly where ``mask_k == 0`` (a finite target at a masked
+  cell NaNs the loss; a NaN target at an unmasked cell NaNs metrics).
+  The source footprint drifts slightly from the reference-time wetmask
+  (z*-remapped bottom slivers dry/re-wet with sea level at a handful of
+  sub-surface cells), so each chunk is conformed to the wetmask before
+  regridding: values at cells wet at time t but outside the wetmask are
+  dropped, and cells inside the wetmask that are instantaneously dry are
+  filled from the level above (the water immediately overlying the
+  vacated sliver). See _conform_to_wetmask.
```
spencerkclark: Fine to keep `_conform_to_wetmask`—it's great that it facilitates this strict check even if we do not expect to use its primary funcationality—but reminder to update this section of the docstring.

scripts/gfdl_om4/pipeline/run.py
```diff
@@ -0,0 +1,973 @@
...
+        ds = ds.rename(stream.dim_renaming)
+        for staggered, tracer in STAGGERED_TO_TRACER_DIM.items():
+            if (
+                staggered in ds.dims
+                and tracer in ds.dims
+                and ds.sizes[staggered] == ds.sizes[tracer] + 1
+            ):
+                # Symmetric staggering carries both edges; drop the first
+                # point to match the right/north-edge convention of the
+                # shared interpolation and rotation machinery.
```
spencerkclark, line 144: Indeed awkward that the sea ice and ocean outputs follow different conventions here...

scripts/gfdl_om4/pipeline/run.py
```diff
@@ -0,0 +1,973 @@
...
+    wetmask.attrs = {}
+    return wetmask
+
+
+# Instantaneous ocean coverage may drift slightly from the reference-time
+# wetmask (cells with time-varying validity at sub-surface levels); observed
+# drift is ~0.01% of wet cells. Anything larger indicates an inconsistent
+# source (wrong grid, dropped mask) and fails the run rather than being
+# silently conformed.
+MAX_FOOTPRINT_DRIFT_FRACTION = 0.001
```
spencerkclark: Oh so this is still giving some lenience to a changing mask when `max_conformed_cells` is `None`?

scripts/gfdl_om4/pipeline/face_masks.py
```diff
@@ -0,0 +1,415 @@
...
+        zero_all = (
+            zero.all(axis=0) if zero_all is None else (zero_all & zero.all(axis=0))
+        )
+        logger.info(
+            "  %s: scanned steps %d..%d of %d",
+            da.name,
+            start,
+            min(start + TIME_BATCH, n_steps) - 1,
+            n_steps,
+        )
```
spencerkclark, line 167: nit: I think these operations should all be possible with xarray objects.  I would find it easier to follow with labeled dimensions.

scripts/gfdl_om4/pipeline/face_masks.py
spencerkclark, line 1: Oof a lot of work to address this issue.  We should find out why it's the case in MOM6 and whether they would be open to addressing it.

spencerkclark (commented):
A few minor comments as I gave this an initial rough read.  No red flags stood out so far.  It's great to have this workflow so well organized / easy to run now.

scripts/gfdl_om4/pipeline/run.py
jpdunc23: Updated in 4248ece4b — the masking-conventions section now describes `_conform_to_wetmask` as in practice a strict per-chunk footprint check (the production sources have shown no drift, and both the smoke tests and the production launch targets pin it to a no-op with `--max-conformed-cells 0`), with the fill-from-above repair documented as the fallback for a drifting source.

scripts/gfdl_om4/pipeline/run.py
jpdunc23: Agreed — the symmetric-vs-non-symmetric staggering difference between the ice and ocean stores is unfortunate. The dim renaming plus first-edge drop in `open_stream` confines it to that one spot, so everything downstream sees a single right/north-edge convention.

scripts/gfdl_om4/pipeline/run.py
jpdunc23: It was — with no `--max-conformed-cells`, drift up to `MAX_FOOTPRINT_DRIFT_FRACTION` (0.1% of wet cells per timestep) was conformed rather than fatal; that bound exists to catch a grossly inconsistent source while letting the sliver-scale drift seen in the trial sources through. Since the production sources have a static footprint, as of 4248ece4b the production `dataflow_*` Makefile targets now pass `--max-conformed-cells 0` just like the smoke tests, so any conforming is a hard failure and lenience is strictly opt-in (invoking `run-dataflow.sh` directly without the flag).

scripts/gfdl_om4/pipeline/face_masks.py
jpdunc23: Done in 4248ece4b — `_scan_structural` now accumulates xarray objects and reduces over the named time dimension, converting to numpy only at the return boundary. I left the neighbor/fold helpers in numpy since the tripolar seam reflection is most explicit as array indexing, but happy to convert those too if you'd prefer.

scripts/gfdl_om4/pipeline/ocean_emulators_port.py
jpdunc23: Agreed. This now reads:

```
    Cells with no valid neighbor
    come out NaN. Along a periodic dimension the first center wraps around.
    In the non-periodic case ``shift`` supplies a row of NaNs as the first
    center's missing "left" neighbor, so it uses its single in-domain
    neighbor — natural for MOM6, whose southern boundary is Antarctica:
    land, and therefore all-NaN there anyway.
```

scripts/gfdl_om4/pipeline/face_masks.py
jpdunc23: Definitely! I'm thinking you're probably best positioned to follow up on this, if you think it is worth doing so. Would you be willing to take that on?

scripts/gfdl_om4/configs/om4-1pctco2-1deg-5daily.yaml
jpdunc23, line 1: Add `-5daily` to the filenames for this an the other configs.

jpdunc23, line 1: Is naming these files / outputs stores with `om4` confusing given that the data comes from a coupled CM4 simulation?

Note that we've been training on output from this pipeline (`2026-07-15-om4-1pctco2-1deg-ocean-5daily.zarr`, `2026-07-15-om4-picontrol-1deg-ocean-5daily.zarr`, and the 4-degree variants) so renaming to, say, `cm4-picontrol-1deg-ocean-5daily` will also have a cost.

scripts/gfdl_om4/pipeline/run.py
jpdunc23: Perhaps we should attempt to simplify the pipeline by removing this `MAX_FOOTPRINT_DRIFT_FRACTION` and `max_conformed_cells` handling. We can extract it to a separate draft PR in case we have a future need.

scripts/gfdl_om4/configs/om4-1pctco2-1deg-5daily.yaml
jpdunc23: There are also inconsistencies in capitalization compared to existing stores: `1pctco2` here vs `1pctCO2` elsewhere, `picontrol` vs `piControl`, `cm4` vs `CM4`, etc.

scripts/gfdl_om4/pipeline/run.py
jpdunc23: Removed to https://github.com/ai2cm/ace/pull/1474

scripts/gfdl_om4/pipeline/run.py
jpdunc23: I decided to simplify `_conform_to_wetmask` to just do the strict check without the complicated NaN "sliver" fill-from-above handling. The more full-featured version is moved to https://github.com/ai2cm/ace/pull/1474 in case we ever need it.

scripts/gfdl_om4/configs/om4-1pctco2-1deg-5daily.yaml
jpdunc23: We're going to likely update this pipeline soon, so let's just keep as is for now and update the config / output store names for the next iteration.

scripts/gfdl_om4/pipeline/run.py
```diff
@@ -0,0 +1,871 @@
...
+    else:
+        return url
+
+
+def _assert_output_store_absent(path: str) -> None:
+    """Refuse to initialize into a pre-existing output store.
+
+    Output stores are written once and treated as immutable; initializing
+    the template into an existing store would corrupt or silently overwrite
+    it. Delete the store explicitly or pick a new output path.
```
jpdunc23:
```suggestion
    it.
```

scripts/gfdl_om4/pipeline/run.py
```diff
@@ -0,0 +1,871 @@
...
+
+    # Per-level ocean fraction and binary masks from the regridded wetmask.
+    level_count = wetmask.sizes[LEVEL_DIM]
+    surface_fraction = None
+    for k in range(level_count):
+        fraction = regridder(
+            wetmask.isel({LEVEL_DIM: k}, drop=True).astype("float64"), keep_attrs=False
+        ).fillna(0.0)
+        if k == 0:
+            surface_fraction = fraction
```
jpdunc23, line 512: Maybe we should also save `f"wet_fraction_{k}"` for each level in future iterations so that we can compute native-grid equivalent budgets.

scripts/gfdl_om4/configs/om4-1pctco2-1deg-5daily.yaml
spencerkclark: Indeed I would prefer `CM4` if we are to include a model name in the path, especially since we include `ocean` in the store name, which is enough to differentiate it from atmosphere and/or land datasets.  I do not have a strong opinion about capitalization.  

But I agree no need to change anything here, especially since there are training jobs already in flight.

scripts/gfdl_om4/pipeline/run.py
spencerkclark: Agreed that seems like a good idea.

scripts/gfdl_om4/pipeline/face_masks.py
spencerkclark: I'll reach out about this after I work through the review comments on the ACE2S-SHiELD+ paper.

spencerkclark (approved):
Thanks for factoring out the wetmask lenience and for the other updates following my initial comments.  This looks good to me now!

PR #1342, opened 2026-07-07
Add optional prediction_names to filter prediction files independently

elynnwu:
`DataWriterConfig.names` applies to both prediction and monthly netCDF files. Add an optional `prediction_names` field that, when set, overrides `names` for the prediction (autoregressive) files only, so a small subset can be saved at high frequency while the monthly aggregates retain all variables. Falls back to `names` when unset, so existing configs are unaffected.

Changes:
- `fme.ace.inference.data_writer.main.DataWriterConfig`: add optional `prediction_names` field; new private `_prediction_save_names` property returning `prediction_names` if set else `names`; both `build` and `build_paired` pass it as `save_names` to the raw/prediction writers while the monthly writers keep `names`. Adds a `__post_init__` warning when `prediction_names` is set but `save_prediction_files` is False.
- `test_data_writer.py`: add a warning test and an end-to-end test asserting the prediction file is subset to `prediction_names` while the monthly file still saves all variables.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/data_writer/main.py
+++ b/fme/ace/inference/data_writer/main.py
@@ -41,6 +41,12 @@ class DataWriterConfig:
             containing the monthly predictions and target values.
         names: Names of variables to save in the prediction and monthly
             netCDF files.
+        prediction_names: Names of variables to save in the prediction netCDF
+            files. When set, this overrides ``names`` for the prediction files
+            only (the monthly files continue to use ``names``); when None, the
+            prediction files fall back to ``names``. Useful for saving a small
+            subset in the high-frequency prediction files while retaining all
+            variables in the monthly aggregates.
         time_coarsen: Configuration for time coarsening of written outputs to the
             raw data writer.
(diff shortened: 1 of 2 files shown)


oliverwm1:
Can you use the more flexible `DataWriterConfig.files` API to achieve this? That was the intention of that feature. At one point we had plans to deprecate/remove the existing "raw" and "monthly" data writers, which I think we should still consider.

AnnaKwa:
> Can you use the more flexible `DataWriterConfig.files` API to achieve this? That was the intention of that feature. At one point we had plans to deprecate/remove the existing "raw" and "monthly" data writers, which I think we should still consider.

This was my comment too- seems like there is duplication between the older `save_prediction_files` and the newer file writer config options.

elynnwu:
Ahh, I totally missed that. Thanks both! Closing this PR in favor of just using `DataWriterConfig.files`. I'll add an example config showing how that's done instead.

PR #1344, opened 2026-07-07
Refactor OptimizationConfig to build the LR scheduler in its builder

mcgibbon:
Applies the Builder pattern's "Rule 2" (the impl should not build the sub-objects the builder is responsible for) to `fme/core/optimization.py`. Previously `OptimizationConfig.build` passed the raw scheduler *config* plus `parameters`/`optimizer_type`/`lr`/`max_epochs`/`kwargs` into `Optimization`, and `Optimization.__init__` built both the optimizer (the Adam/AdamW/FusedAdam if/elif) and the scheduler. This moves that construction into the builder.

Changes:
- `fme.core.optimization._build_optimizer`: new private module-level helper holding the Adam/AdamW/FusedAdam selection, returning a `torch.optim.Optimizer`.
- `fme.core.optimization.OptimizationConfig.build`: builds `parameters`, then the optimizer (via `_build_optimizer`), then the scheduler (`self.scheduler.build(optimizer, max_epochs)`), then passes the built `optimizer` and `scheduler` into `Optimization`. Public signature `(modules, max_epochs)` unchanged; the `resume_optimizer_ckpt_path` finetune-state load still runs after construction.
- `fme.core.optimization.Optimization.__init__`: no longer builds sub-objects. New signature takes `optimizer: torch.optim.Optimizer` and `scheduler: LRScheduler` (from `fme.core.scheduler`), plus the existing `enable_automatic_mixed_precision`, `use_gradient_accumulation`, `get_checkpoint`, `max_grad_norm`. gscaler construction, accumulated-loss init, and all other behavior unchanged.
- `fme.core.test_optimization` and `fme.core.generics.test_lr_tuning`: the direct-construction call sites (the only other places that built `Optimization` by hand) are routed through the real `OptimizationConfig(...).build(nn.ModuleList([...]), max_epochs)` path, following the corrector-template precedent (#1313), preserving each test's intent/params.

Behavior-preserving: no YAML config fields changed. `python -m pytest fme/core/test_optimization.py fme/core/generics/test_lr_tuning.py` passes (50 passed, 2 skipped CUDA-only); ruff + mypy clean on the changed files.

The ticket estimated two files; there are three (impl + the two test files that construct `Optimization` directly).

- [x] Tests added (existing tests re-routed through the build path; no new behavior to cover)

### DIFF
--- a/fme/core/generics/test_lr_tuning.py
+++ b/fme/core/generics/test_lr_tuning.py
@@ -1,7 +1,6 @@
 """Tests for run_lr_tuning_trial and LRTuningConfig."""
 
 import copy
-import itertools
 import unittest.mock
 from typing import Any
 
@@ -116,15 +115,11 @@ class _Stepper(TrainStepperABC["None", "_BatchData", "None", "None", "_TrainOutp
 
 
 def _build_optimization(modules: torch.nn.ModuleList) -> Optimization:
-    return Optimization(
-        parameters=itertools.chain(*[m.parameters() for m in modules]),
+    return OptimizationConfig(
         optimizer_type="Adam",
         lr=0.01,
-        max_epochs=10,
         scheduler=SchedulerConfig(),
-        enable_automatic_mixed_precision=False,
-        kwargs={},
-    )
+    ).build(modules, max_epochs=10)
 
 
 def _make_copy_ema(ema: EMATracker):
--- a/fme/core/optimization.py
+++ b/fme/core/optimization.py
@@ -11,7 +11,7 @@ from torch import nn
 
 from fme.core.device import get_device
 from fme.core.generics.optimization import OptimizationABC
-from fme.core.scheduler import SchedulerConfig, SequentialSchedulerConfig
+from fme.core.scheduler import LRScheduler, SchedulerConfig, SequentialSchedulerConfig
 from fme.core.typing_ import TensorDict, TensorMapping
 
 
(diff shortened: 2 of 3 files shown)


mcgibbon:
Claude: Independent pre-review (Opus, worktree-isolated, cold). Reviewed the diff against `origin/main` (3 files, +76/-115). This is a behavior-preserving Builder "Rule 2" refactor; verified by audit pass:

- **Rule 2 fix** ✅ `Optimization.__init__` no longer constructs the optimizer or scheduler — it takes a built `torch.optim.Optimizer` and `LRScheduler` directly. `OptimizationConfig.build` builds the optimizer (via the new module-level `_build_optimizer`), then the scheduler (`self.scheduler.build(optimizer, max_epochs)`), in the correct order (optimizer before scheduler, since the scheduler needs it), then passes both in.
- **Silent-failure pass** ✅ `_build_optimizer`'s `else` still raises `ValueError` loudly; no silent default introduced.
- **Behavior-change pass** ✅ Optimizer-type selection (`FusedAdam`→`AdamW(fused=True)` / `Adam` / `AdamW`) is byte-identical to the old if/elif. gscaler/AMP construction, accumulated-loss init, `use_gradient_accumulation`, `get_checkpoint`, `max_grad_norm` all unchanged in `__init__`. `resume_optimizer_ckpt_path` finetune-state load still runs after construction in `build`. `get_state`/`load_state`/scheduler stepping untouched. The only reordering (gscaler now built after the scheduler rather than before) is inconsequential — the two are independent, and `scheduler.build` needs only the optimizer.
- **Differential-test pass** ✅ The re-routed tests now genuinely exercise `OptimizationConfig(...).build(...)`. Each conversion preserves the original params; every previously-explicit `enable_automatic_mixed_precision=False`/`kwargs={}` matches the config defaults, so no test's intent or coverage changed. `nn.ModuleList([model])` references the same `nn.Parameter` objects, so the optimizer still operates on the real model weights (finetune/load-state tests remain valid).
- **AI-artifact / imports pass** ✅ `LRScheduler` import added and used; unused `itertools` removed from `test_lr_tuning.py`; `SchedulerConfig`/`SequentialSchedulerConfig` still used. No dead code, no Rule 1 leak (`_build_optimizer` is private module-level, callable by the builder).
- **PR description** ✅ Accurate against the diff (3 files, the file-count caveat, the deps checkbox correctly unchecked).

Verification: `FME_FORCE_CPU=1 python -m pytest fme/core/test_optimization.py fme/core/generics/test_lr_tuning.py` → 50 passed, 2 skipped (CUDA-only), matching the description. Pinned pre-commit hooks (ruff v0.8.1, ruff-format, mypy v1.15.0) all pass on the changed files. (Note: a newer local ruff 0.15.17 wanted to reformat an unrelated pre-existing `get_checkpoint` line and a stale local mypy flagged 602 pre-existing errors elsewhere — both are version artifacts, none in the changed files; the pinned CI versions are clean.)

No fixes were needed, so nothing was committed to the branch.

jpdunc23 (approved):
LGTM!

PR #1345, opened 2026-07-07
Write step diagnostics as named datasets, one file per name

jpdunc23:
Following review on #1335, step diagnostics should define multiple named datasets, each written to its own appropriately named file, so that variables like `air_temperature_0` in the output are identifiable as correction deltas rather than the diagnosed quantities themselves. This PR makes the export API plural (`to_datasets`) even though only the correction-delta dataset exists today, and moves the output from `autoregressive_step_diagnostics.nc` to `step_diagnostics/correction_deltas.nc`. Corrector-supplied dataset naming, per-dataset writer configuration, and additional dataset producers (e.g. clipping or sea-ice masks) are intentionally out of scope.

Changes:
- `fme.core.step.StepDiagnostics.to_dataset` becomes `to_datasets`, returning a mapping of dataset name to `xr.Dataset`; the delta export appears under the new constant `fme.core.step.step_diagnostics.CORRECTION_DELTAS` ("correction_deltas")
- `fme.ace.inference.data_writer.StepDiagnosticsWriter` now takes a per-name sub-writer factory and writes each named dataset to `step_diagnostics/{name}.nc`, creating the directory and one sub-writer per name lazily on first non-empty append; `save_names` filtering and time coarsening apply per file as before
- Behavior changes to note: with `save_step_diagnostics: true` but no corrector activity, no file is created at all (previously an empty skeleton `.nc` was written eagerly), and the netCDF `title` attribute now derives from the dataset name

- [x] Tests added


fme/ace/inference/data_writer/main.py
```diff
@@ -40,10 +40,12 @@ class DataWriterConfig:
...
-        save_step_diagnostics: Whether to enable writing of a netCDF file
-            containing per-step diagnostics of the prediction (currently the
-            corrector's per-step correction delta). No file content is
-            written when no corrector modified anything.
+        save_step_diagnostics: Whether to enable writing of netCDF files
+            containing per-step diagnostics of the prediction, one file per
+            named diagnostics dataset under a ``step_diagnostics/`` directory
+            (currently the corrector's per-step correction delta, written to
+            ``step_diagnostics/correction_deltas.nc``). Nothing is written
+            when no corrector modified anything.
```
jpdunc23, line 48: I can imagine wanted to write a subset of the step diagnostics, in which case we will eventually need a more granular configuration. For the time being with only one diagnostic, the boolean `save_step_diagnostics` will work fine.

fme/ace/inference/data_writer/step_diagnostics.py
```diff
@@ -6,47 +8,56 @@
...
-    Consumes datasets exported by the prediction's step diagnostics container;
-    it has no knowledge of the container's internals. Values are written as-is,
-    already in physical units.
+    Consumes named datasets exported by the prediction's step diagnostics
+    container; it has no knowledge of the container's internals. Values are
+    written as-is, already in physical units.
     """
 
-    def __init__(self, writer: RawDataWriter | TimeCoarsen):
+    def __init__(self, build_writer: Callable[[str], RawDataWriter | TimeCoarsen]):
```
jpdunc23: Change `build_writer` to `writer_factory`, here and elsewhere.

fme/ace/inference/data_writer/step_diagnostics.py
```diff
@@ -6,47 +8,56 @@
 
 VALID_TIME = "valid_time"
 
-STEP_DIAGNOSTICS_LABEL = "autoregressive_step_diagnostics"
+STEP_DIAGNOSTICS_DIR = "step_diagnostics"
```
jpdunc23, line 11: Dropped "autoregressive" since the diagnostics don't exactly feed one into the next. I think "step_diagnostics" already conveys the correspondence to `autoregressive_predictions.nc`.

fme/core/step/step_diagnostics.py
```diff
@@ -4,19 +4,23 @@
...
 import dataclasses
+from collections.abc import Mapping
 
 import xarray as xr
 
 from fme.core.device import get_device
 from fme.core.tensors import repeat_interleave_batch_dim
 from fme.core.typing_ import TensorMapping
 
+CORRECTION_DELTAS = "correction_deltas"
```
jpdunc23, line 24: This determines the output netCDF filename, i.e. `step_diagnostics/correction_deltas.nc`

fme/ace/inference/data_writer/main.py
jpdunc23: Agreed — leaving the boolean for now; a granular per-dataset configuration can come with the first PR that adds a second diagnostics dataset.

fme/ace/inference/data_writer/step_diagnostics.py
jpdunc23: Renamed `build_writer` to `writer_factory` in the writer, the config builder, and the test (7c7b1dc61).

fme/ace/inference/data_writer/step_diagnostics.py
jpdunc23: Sounds right — resolving.

fme/core/step/step_diagnostics.py
jpdunc23: Added a comment on the constant noting it determines the output filename `step_diagnostics/correction_deltas.nc` (7c7b1dc61).

fme/ace/inference/data_writer/step_diagnostics.py
```diff
@@ -6,47 +8,56 @@
...
-            str(name): torch.as_tensor(variable.values)
-            for name, variable in dataset.data_vars.items()
-        }
-        batch_time = xr.DataArray(dataset[VALID_TIME].values, dims=("sample", "time"))
-        self._writer.append_batch(data=data, batch_time=batch_time)
+        for name, dataset in datasets.items():
+            if len(dataset.data_vars) == 0:
+                continue
+            if name not in self._writers:
+                self._writers[name] = self._writer_factory(name)
```
mcgibbon, line 47: Agent: `Issue (follow-up OK):` The sub-writer is built lazily on a name's first non-empty append, but `RawDataWriter` places each batch at `current_lead_time_size` — the running size of its own `LEAD_TIME_DIM`, which starts at 0 for a freshly-created writer. So if a name's dataset is empty for the run's early batches and first becomes non-empty at batch *N > 0*, that name's writer starts writing at lead-time index 0 while carrying the lead-time coordinates from batch *N* — a silently truncated / misaligned lead-time axis relative to the prediction file, with no error raised.

Not a bug today: the sole producer (`correction_deltas`) is empty iff the corrector modifies zero variables, which is structural and constant across a run, so this writer is always created on batch 0 and lazy ≡ eager. The concern is that this plural/lazy machinery is built for the future producers the PR names (clipping / sea-ice masks), and those are exactly the diagnostics that can be empty early and appear later — silently tripping the invariant.

Fix now or open an issue linking this comment. Cheapest robust option: have the writer raise if a name first appears after any batch has been seen (the failure is otherwise silent); or at minimum document loudly here that a name must be non-empty from the run's first batch or its lead-time axis will misalign.

mcgibbon (approved):
LGTM. Claude caught an issue that would likely lead to a bug if we add diagnostics for clip-counts, it should be fixed but please either open an issue or work on it next as a follow-on PR. You could fix it in this PR, but if you do then I should re-review.

fme/ace/inference/data_writer/step_diagnostics.py
jpdunc23: Good catch — agreed this invariant is silent, and lazy creation only coincidentally works today because `correction_deltas` is structurally constant across a run. Opened #1355 to track it as a follow-up, with your suggested fix (raise when a name first appears after any batch has been seen) as the proposed approach.

PR #1346, opened 2026-07-07
Add per-entry aggregator config to coupled inline validation

mcgibbon:
Brings `fme.coupled` inline validation to parity with `fme.ace`: ace `InlineValidationConfig` has an `aggregator: OneStepAggregatorConfig` field plus a `build_aggregator_factory()` method (each validation entry configures its own `OneStepAggregator`), but coupled had none — `OneStepAggregatorConfig()` was hard-coded inside the coupled `OneStepAggregator.__init__`, so coupled validation entries had no per-entry aggregator configurability. One config applies to both the ocean and atmosphere sub-aggregators (separate per-component configs are a deferred, out-of-scope idea).

The default (`OneStepAggregatorConfig()`) reproduces today's behavior exactly, so this is behavior-preserving when `aggregator` is unspecified.

Changes:
- `fme.coupled.aggregator.OneStepAggregator.__init__`: new `config: OneStepAggregatorConfig | None = None` parameter (defaults to `OneStepAggregatorConfig()`), applied to both the ocean and atmosphere sub-aggregators
- `fme.coupled.train.train_config.InlineValidationConfig`: new `aggregator: OneStepAggregatorConfig` field and `build_aggregator_factory()` method mirroring ace
- `fme.coupled.train.train_config._get_validation_callback` / `_get_validate_stepper_callback`: build the aggregator from the per-entry config instead of a hard-coded default


mcgibbon:
Claude (pre-review agent): Ready for review.

Independent pre-review against the stacked base `refactor/train-entrypoint-builder-pattern` (three-dot diff only, #1333 refactor excluded). No fixes were needed; nothing was committed to the branch.

What I verified:
- **Config threading reaches both sub-aggregators.** `OneStepAggregator.__init__` uses the single `config` for both the `ocean` and `atmosphere` `config.build(...)` calls (aggregator.py:131,142). The test pair proves this genuinely: `test_one_step_aggregator_config_reaches_both_sub_aggregators` disables `snapshot` and asserts *no* `val/snapshot/` key survives — it would fail if the config reached only one component (the other's `atmos_var`/`ocean_var` snapshot key would remain) — and the default test confirms both keys are present otherwise.
- **Default path is behavior-preserving.** `config = config or OneStepAggregatorConfig()` reproduces the previous hard-coded default; the field default (`default_factory=lambda: OneStepAggregatorConfig()`) means the callbacks never pass None. `test_one_step_aggregator_default_config_emits_snapshot_for_both` covers this.
- **No dropped kwargs.** The new `build_aggregator_factory` passes exactly the old `make_factory` kwarg set (dataset_info, loss_scaling, save_diagnostics, output_dir) plus `config`; `channel_mean_names`/`variable_metadata` were not passed before and still aren't, so no regression.
- **LR-tuning path consistent with validation path.** `_get_validate_stepper_callback` constructs the aggregator inline with `config=entry_config.aggregator`, and its EMA logic (`run_validation_loop(..., ema=ema, validate_using_ema=validate_using_ema)`) is unchanged — this mirrors `fme.ace`'s `_get_validate_stepper_callback` exactly (ace also builds inline rather than via the factory).
- **Parity with `fme.ace` is faithful.** Field, `build_aggregator_factory` shape, docstrings, and callback routing match. The two deliberate divergences are correct: no `LegacyFlagOneStepAggregatorConfig` union (ace-only deprecated back-compat), and one config for both components (per-component configs deferred).
- **Checks pass:** `fme/coupled/test_aggregator.py` + `fme/coupled/train/test_train_config.py` = 50 passed; `pre-commit` (ruff, ruff-format, mypy) clean on all four changed files.

One minor, non-blocking observation for the reviewer: the LR-tuning path (`_get_validate_stepper_callback`) has no dedicated unit test that the per-entry `config` is threaded — it's covered indirectly (the aggregator-level test proves the threading mechanism, and this path passes the same one-line `config=entry_config.aggregator`). This matches the ace side's test coverage, so I left it as-is rather than add asymmetric coverage.

jpdunc23 (approved):
Aside from tests, the changes in this PR are pretty minor. Can you just squash-and-merge this into the https://github.com/ai2cm/ace/pull/1346 branch?

If you're set on getting #1346 merged first and having this as a separate commit on main, then I have a nit about the description: "Follow-on to https://github.com/ai2cm/ace/pull/1333 addressing the coupled validation-aggregator parity gap jpdunc23 raised in that review". "This stacks on https://github.com/ai2cm/ace/pull/1333 (base branch refactor/train-entrypoint-builder-pattern); rebase onto main after https://github.com/ai2cm/ace/pull/1333 merges" Could we make PR descriptions a little bit more self-contained? I think the agent PR process comments are distracting and should be isolated to the PR discussion. 

If you agree, I think it would be helpful to include this rule needs to be included in our repo agent guidance docs (as a separate PR).

mcgibbon:
> Could we make PR descriptions a little bit more self-contained? I think the agent PR process comments are distracting and should be isolated to the PR discussion.

Yes I've been independently noticing that agents tend to be hesitant to post comments on PRs, and instead want to put everything in the PR description. Maybe explicit instructions in our AGENTS.md, or maybe some text in the PR description template (which they're instructed to use), or both.

PR #1347, opened 2026-07-07
Add normalized correction metrics to the inference aggregators

jpdunc23:
Corrector-equipped runs currently write their per-step correction deltas to `autoregressive_step_diagnostics.nc` but log no metrics about them, so there is no at-a-glance answer to "how much does this model lean on its corrector?". This PR adds normalized-space correction metrics to both inference aggregators, computed from the `StepDiagnostics` attached to prediction data: an area-weighted global mean of the time-mean |normalized correction| per corrected variable (plus an opt-in signed time-mean map image) under the existing `time_mean_norm` group, and per-forecast-step area-weighted magnitude/std series under `mean_norm`. The normalized correction is `normalize(delta, apply_mean=False)` — i.e. `delta / std`, using the `apply_mean` option from #1353, since mean-centering a difference quantity would wrongly subtract the per-variable mean offset; all reductions go through the mask-aware `GriddedOperations`, so NaN off-mask cells drop out of the scalars. Metrics are uniform over all corrected variables and silent when no corrector ran (or in coupled inference, which does not attach step diagnostics). Inline training keeps the time-mean metrics but drops the per-step series, matching the existing time-series behavior.

The wiring mirrors the `StepDiagnostics` container rather than its current contents: each host aggregator carries a single optional `StepDiagnosticsAggregator` — a concrete wrapper owning a dict of per-concern sub-aggregators (one entry today — the correction-delta aggregator) and fanning record/log/flush out over it — so future `StepDiagnostics` fields get sibling entries inside the wrapper instead of more host surgery. Granular sub-aggregators for specific container fields (just `delta`, for now) live another level down and are equally invisible to the hosts.

Changes:
- `fme.ace.aggregator.inference.step_diagnostics` (new): the `StepDiagnosticsAggregator` wrapper, the `StepDiagnosticsSubAggregator` protocol, `StepDiagnosticsMetricConfig`, the `CorrectionDeltaAggregator`, and its granular `CorrectionDeltaTimeMeanAggregator` / `CorrectionDeltaMeanAggregator` sub-aggregators
- `InferenceEvaluatorAggregatorConfig.step_diagnostics` and `InferenceAggregatorConfig.step_diagnostics`: new `StepDiagnosticsMetricConfig` fields controlling metric granularity — `correction_scalars` (default on) and `correction_maps` (default off)
- Both inference aggregators gain `step_diagnostics_aggregator: StepDiagnosticsAggregator | None = None` and delegate record/log/flush to it
- `InferenceAggregatorConfig.build` gains an optional `normalize` argument (default `None`); the no-target inference entrypoint now passes `stepper.normalizer.normalize`. With no normalizer, a default `step_diagnostics` config silently skips the correction metrics (preserving current behavior for callers that omit `normalize`, e.g. the coupled builder), while an explicitly non-default config raises rather than silently dropping an opt-in
- `reduced.py`'s series-assembly dataclass is public (`SeriesData`), and `CorrectionDeltaTimeMeanAggregator` raises if a batch changes the sample count or variable set mid-accumulation instead of silently biasing the time-means

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+        ...
+
+    def _get_logs(self): ...                    # CHANGED — merge correction.summary_logs()
+    def get_summary(self): ...                  # CHANGED — same merge (loss key untouched)
+    def _get_inference_logs_slice(self, ...): ...  # CHANGED — merge correction.time_series_logs(step_slice)
+    def flush_diagnostics(self, subdir=None): ...  # CHANGED — merge correction.diagnostics()
+
+
+class InferenceAggregatorConfig:
+    log_correction_metrics: bool = True  # NEW
```
jpdunc23: The configuration should probably be more granular so that we can enable the new scalar metrics by default while disabling the new maps by default.

pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+    ): ...
+
+    def record_batch(self, correction_norm: TensorMapping) -> None: ...
+
+    def get_logs(self, label: str) -> dict[str, float | Image]:
+        # correction_magnitude/{name}: area-weighted global mean of the
+        #   time-mean |normalized correction| (mask-aware: reduction goes
+        #   through ops.area_weighted_mean(..., name=name), so NaN off-mask
+        #   cells drop out of the weights)
+        # correction_magnitude/channel_mean: arithmetic mean of those scalars
```
jpdunc23: I don't think we need the channel mean here. The interpretation will change a lot depending on what correctors are enable, which limits the usefulness of the channel mean metric. Please remove.

pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+
+```python
+def compute_correction_norm(
+    prediction: TensorMapping,
+    delta: TensorMapping,
+    normalize: Callable[[TensorMapping], TensorDict],
+) -> TensorDict:
+    """The normalized correction, over the delta's keys.
+
+    Computed as normalize(prediction) - normalize(prediction - delta) —
```
jpdunc23: Jeremy thought there may be a way to use `normalize` while disabling the centering step, e.g. something like `normalize(delta, center=False)`. I don't think that feature currently exists, but we should double check.

pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+    def record_batch(self, correction_norm: TensorMapping) -> None: ...
+
+    def get_logs(self, label: str) -> dict[str, float | Image]:
+        # correction_magnitude/{name}: area-weighted global mean of the
+        #   time-mean |normalized correction| (mask-aware: reduction goes
+        #   through ops.area_weighted_mean(..., name=name), so NaN off-mask
+        #   cells drop out of the weights)
+        # correction_magnitude/channel_mean: arithmetic mean of those scalars
+        #   over the corrected variables (mirrors time_mean_norm/rmse/channel_mean)
+        # correction_map/{name}: signed time-mean map image (diverging palette)
```
jpdunc23: Is the planned `correction_map` also normalized? I think denormalized maps are more useful and this would also match the data writer output which writes the deltas in denormalized space.

No need to update the plan one way or another for the time being, but please provide clarification.

pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+
+    def build(self, ...) -> "InferenceEvaluatorAggregator":  # CHANGED — forwards the flag
+        ...
+
+
+class InferenceEvaluatorAggregator:
+    def __init__(
+        self,
+        ...,
+        correction: CorrectionDeltaRecorder | None = None,  # NEW
```
jpdunc23: Rather than pass a `delta`-specific `correction: CorrectionDeltaRecorder`, let's try to establish a bit more generality for possible future additions to `StepDiagnostics` by passing `step_diagnostics_aggregators: dict[str, StepDiagnosticsAggregator] | None = None`.

pr-plan.md
```diff
@@ -0,0 +1,285 @@
...
+    log_correction_metrics: bool = True,  # NEW
+) -> "InferenceEvaluatorAggregator":
+    # NEW: when enabled, builds a CorrectionDeltaRecorder from
+    # dataset_info.gridded_operations / variable_metadata, n_timesteps,
+    # enable_time_series, and the existing normalize argument.
+    ...
+
+
+class InferenceEvaluatorAggregatorConfig:
+    log_correction_metrics: bool = True  # NEW — also on the Legacy flag config
```
jpdunc23: Configuration attr naming should also be more generic, e.g. `step_diagnostics: StepDiagnosticsMetricConfig`. `StepDiagnosticsMetricConfig` can control granularity of `StepDiagnostics`-related metrics (just on correction `deltas`, for now).

pr-plan.md
jpdunc23: Done — the flag is replaced by `StepDiagnosticsMetricConfig` with `correction_scalars: bool = True` and `correction_maps: bool = False`, so the new scalar metrics are on by default while map images are opt-in. See the `StepDiagnosticsMetricConfig` block in `pr-plan.md`.

pr-plan.md
jpdunc23: Removed — `correction_magnitude` is now per-variable only.

pr-plan.md
jpdunc23: Checked: `StandardNormalizer.normalize` has no centering toggle. It also turns out not to be needed — `normalize(pred) − normalize(pred − delta)` reduces algebraically to `delta / std` (the means cancel), so the planned formulation already is the uncentered normalize, with no normalizer change. Noted in the `compute_correction_norm` docstring in `pr-plan.md`.

pr-plan.md
jpdunc23: Clarification: the planned `correction_map` is normalized, consistent with its `time_mean_norm` group; denormalized per-cell deltas are already available from the step-diagnostics writer output. Maps are now off by default per the config-granularity thread, so the normalized map is opt-in.

pr-plan.md
jpdunc23: Done — both hosts now take `step_diagnostics_aggregators: dict[str, StepDiagnosticsAggregator] | None = None` (a new protocol), with the correction-delta aggregator as the sole entry for now; the granular delta sub-aggregators live behind it and are invisible to the hosts.

pr-plan.md
jpdunc23: Done — the config attribute is `step_diagnostics: StepDiagnosticsMetricConfig` on both aggregator configs, controlling granularity of `StepDiagnostics`-derived metrics (correction deltas only, for now).

pr-plan.md
mcgibbon: Would it be a better idea to add an `apply_mean: bool = True` option to normalize instead? Both because it's clearer to a reader (instead of having to explain that `normalize(pred) − normalize(pred − delta)` is algebraically identical), and because it avoids unneeded compute?

pr-plan.md
```diff
@@ -0,0 +1,343 @@
...
+    correction_maps: bool = False     # map images opt-in
+
+    def build(
+        self,
+        gridded_operations: GriddedOperations,
+        n_timesteps: int,
+        variable_metadata: Mapping[str, VariableMetadata] | None,
+        enable_time_series: bool,
+        normalize: Callable[[TensorMapping], TensorDict] | None,
+    ) -> dict[str, StepDiagnosticsAggregator]:
```
mcgibbon: We could isolate more of the concern of handling step diagnostics by having the new step-diagnostics-specific aggregator handle the dict[str, Aggregator] part of its existence and expose a single Aggregator. This could be a simple wrapper aggregator over such a dict. Otherwise, the main aggregator is doing low-level combine-the-aggregators logic which really belongs to the step diagnostics.

This becomes immediately relevant, since after these PRs are in I'm planning to associate the step class with a specific aggregator class, and the coupling would bite me there. It would be very helpful to have just one aggregator returned here.

pr-plan.md
jpdunc23: We could do this, though I'm not sure it really warrants blocking this PR on the refactor. 

Specific to this PR, I'm not too concerned about the extra `normalize` adding a lot of noticeable overhead relative to other things happening in the aggregators (e.g., spatial reductions). Is there a reason to suspect that an extra `normalize` call will significantly slow things down?

pr-plan.md
```diff
@@ -0,0 +1,343 @@
...
+        # {"correction": CorrectionDeltaAggregator(...)} — the key names the
+        # flush_diagnostics files. Empty dict when normalize is None (the
+        # correction metrics are normalized-space; callers that supply no
+        # normalizer keep current behavior) or when both flags are off.
+        ...
+
+
+def compute_correction_norm(
```
mcgibbon: nit: rename to `normalize_delta_without_mean`, as that's clearer about what this does.

pr-plan.md
mcgibbon: It's kind of confusing that `normalize(pred) − normalize(pred − delta) = delta / std`. Like, I still don't know it from having done the algebra, I'm kind of just accepting what it says as true. I'd prefer if each time someone reads this code they don't have to trust the algebra is correct or do a derivation.

One way you could accomplish this is by defining a `normalize_delta_without_mean(normalize, prediction, delta)` function which returns `normalize(pred) − normalize(pred − delta) = delta / std` and has a unit test with a few cases demonstrating that this is true. But I think it's just much simpler to add an optional `apply_mean` to `normalize`.

Actually, I realize now that that's exactly what you _do_ already have in the plan, it just wasn't named in a way that made that obvious. Posted a nit suggesting a rename.

I don't think this has a significant enough impact to block merge, but I don't see why we wouldn't do it. It turns an additional function in the plan with an indirect mechanism into an optional keyword argument on an existing method with clear action and less compute. I don't think this warrants its own separate PR (the same as how adding `normalize_delta_without_mean` doesn't require its own PR), you'd roll it into this one.

jpdunc23: The reason I think a separate PR is warranted is because the addition of `apply_mean: bool` to `normalize` does (I think) affect the required typing signature in a number of unrelated modules. I'm happy to add this as a separate PR in parallel, which can probably get reviewed quickly and merged first.

pr-plan.md
jpdunc23: Agreed this sounds nicer, let's go with it.

pr-plan.md
jpdunc23: Sounds good.

pr-plan.md
jpdunc23: Done — the plan now has the hosts hold a single optional `StepDiagnosticsAggregator`, a concrete wrapper owning the dict of per-concern sub-aggregators (which satisfy a `StepDiagnosticsSubAggregator` protocol) and the record/log/flush fan-out, so the combine logic lives with the step diagnostics rather than the hosts.

pr-plan.md
jpdunc23: Renamed to `normalize_delta_without_mean` in the plan (function and its tests).

pr-plan.md
jpdunc23: https://github.com/ai2cm/ace/pull/1353 merged, we should update the plan accordingly.

jpdunc23: Done — the plan now computes the normalized correction directly as `normalize(delta, apply_mean=False)` using the option #1353 added (the `normalize_delta_without_mean` helper is gone), and the normalize-passing annotations use the new `NormalizeFn` protocol. Latest `main` (including #1353) is merged into the branch.

jpdunc23:
Pre-review agent: I ran a pre-review audit of this PR ahead of human review, in a clean session separate from the one that built the change. All 19 new tests pass locally and ruff/mypy are clean. Three observations for the reviewer, none blocking:

1. **Silent skip when no normalizer is supplied.** `StepDiagnosticsMetricConfig.build` returns `None` whenever `normalize is None`, including when the user explicitly enabled `correction_scalars`/`correction_maps`. Both in-repo entrypoints now pass a normalizer, and the behavior is documented in the config docstring, but an explicit opt-in being silently dropped (rather than raising) is a deliberate trade-off the reviewer may want to sign off on.
2. **Unguarded accumulation invariants in `CorrectionDeltaTimeMeanAggregator`.** `record_batch` overwrites `self._n_samples` on every call and divides each variable's accumulated sum by the global timestep count, which assumes a constant sample count and a constant variable set across windows. The current inference loop guarantees both, but nothing asserts them — a future caller violating either would get silently biased time-means rather than an error.
3. **Multi-window accumulation is untested.** No test records two batches into the time-mean aggregator, so the accumulation arithmetic in (2) is exercised only for the single-window case. A two-call test with a hand-computed expected mean would lock it down.

Minor: `step_diagnostics.py` imports the private `_SeriesData` from `reduced.py` — the only cross-module use of that private name; consider making it public.

jpdunc23:
> 1. **Silent skip when no normalizer is supplied.** `StepDiagnosticsMetricConfig.build` returns `None` whenever `normalize is None`, including when the user explicitly enabled `correction_scalars`/`correction_maps`. Both in-repo entrypoints now pass a normalizer, and the behavior is documented in the config docstring, but an explicit opt-in being silently dropped (rather than raising) is a deliberate trade-off the reviewer may want to sign off on.

I think we already have precedents of raising errors in the aggregators when user config and build-time data availability are mismatched. Should we do the same here?

> 2. **Unguarded accumulation invariants in `CorrectionDeltaTimeMeanAggregator`.** `record_batch` overwrites `self._n_samples` on every call and divides each variable's accumulated sum by the global timestep count, which assumes a constant sample count and a constant variable set across windows. The current inference loop guarantees both, but nothing asserts them — a future caller violating either would get silently biased time-means rather than an error.

Is this a preexisting issue in other time mean aggregators? If not, we should follow their prior art. If so, do we just need to provide some better documentation of the `record_batch` assumptions?

> 3. **Multi-window accumulation is untested.** No test records two batches into the time-mean aggregator, so the accumulation arithmetic in (2) is exercised only for the single-window case. A two-call test with a hand-computed expected mean would lock it down.

Let's add this test.
 
> Minor: `step_diagnostics.py` imports the private `_SeriesData` from `reduced.py` — the only cross-module use of that private name; consider making it public.

Let's make it public.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,451 @@
...
+                if step_slice is not None:
+                    arr = arr[step_slice]
+                data.append(
+                    _SeriesData(
+                        metric_name=metric_name,
+                        var_name=var_name,
+                        data=self._dist.reduce_mean(arr).cpu().numpy(),
+                    )
+                )
+        return data
```
jpdunc23: This is a near copy of `MeanAggregator._get_series_data` with very minor changes: A maintainability headache. Similarly for `get_dataset()`. 

Can we factor out the existing `MeanAggregator._get_series_data` into a public helper method and reuse it here? Also for other repetitive implementations?

When factoring out existing code, avoid also changing the implementation, even though your implementation here is slightly more compact.

pr-plan.md

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: Factored the series assembly into public `get_series_data` and `series_data_to_dataset` helpers in `reduced.py`, lifted verbatim from `MeanAggregator`, and reused them in `MeanAggregator`, `SingleTargetMeanAggregator`, and `CorrectionDeltaMeanAggregator` — each keeps its own no-batches guard. Only intentional behavior change: `SingleTargetMeanAggregator.get_dataset` now returns an empty dataset instead of crashing when no metrics were recorded, inheriting `MeanAggregator`'s empty-safe branch.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,433 @@
...
+from fme.core.gridded_ops import GriddedOperations
+from fme.core.normalizer import NormalizeFn
+from fme.core.step.step_diagnostics import StepDiagnostics
+from fme.core.typing_ import TensorDict, TensorMapping
+from fme.core.wandb import Image
+
+from ..plotting import plot_paneled_data
+from .reduced import (
+    AreaWeightedSingleTargetReducedMetric,
+    _SeriesData,
```
jpdunc23: This should be made public, `SeriesData`.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,433 @@
...
+    @torch.no_grad()
+    def record_batch(
+        self,
+        prediction: TensorMapping,
+        step_diagnostics: StepDiagnostics | None,
+        i_time_start: int,
+    ) -> None:
+        if step_diagnostics is None:
+            # no corrector ran, or the prediction pipeline does not attach
+            # diagnostics (e.g. coupled inference)
```
jpdunc23: The `(e.g. coupled inference)` is at risk of being missed in a future PR when we port this to `fme.coupled`, so let's remove.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,433 @@
...
+        self._time_series = time_series
+
+    @property
+    def log_time_series(self) -> bool:
+        return self._time_series is not None
+
+    @torch.no_grad()
+    def record_batch(
+        self,
+        prediction: TensorMapping,
```
jpdunc23: Do we need `prediction`? I think the main reason this was included was because before https://github.com/ai2cm/ace/pull/1353 we needed the prediction for normalization, but that's no longer true. I guess it's nice to be able to check that the deltas have the same time axis shape as the predictions, and the predictions may be helpful for future `StepDiagnosticsSubAggregator` implementations, but they aren't strictly required in this PR.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,433 @@
...
+                "correction": CorrectionDeltaAggregator(
+                    normalize=normalize,
+                    time_mean=time_mean,
+                    time_series=time_series,
+                )
+            }
+        )
+
+
+class CorrectionDeltaAggregator:
```
jpdunc23, line 168: Is there a reason not to inherit `StepDiagnosticsSubAggregator`?

jpdunc23:
All four addressed in cd6dd05d2:

1. Followed the existing precedent, with one nuance: `StepDiagnosticsMetricConfig.build` now raises `ValueError` when the configuration was explicitly changed from the defaults but no normalizer is supplied. The default configuration (scalars on) keeps the silent skip — raising unconditionally would break the coupled inference aggregator builder, which constructs the per-component `InferenceAggregatorConfig` with defaults and no normalizer.
2. Partially preexisting: `TimeMeanAggregator` sets its sample count once rather than overwriting it, and a changed variable set there fails with a `KeyError` rather than biasing. I brought the correction time-mean aggregator up to (slightly beyond) that bar: `record_batch` documents its contiguous-windows assumptions and raises `ValueError` if a batch changes the sample count or the variable set mid-accumulation.
3. Added a two-window accumulation test with a hand-computed expected mean over unequal window lengths, plus tests for the two new invariant errors and the build-time error.
4. Renamed to public `SeriesData`.

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: Renamed to public `SeriesData` in cd6dd05d2, here and at its definition in `reduced.py`.

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: Removed in 4b6e61065.

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: Agreed — dropped `prediction` from the whole record path (protocol, wrapper, and `CorrectionDeltaAggregator`) in 4b6e61065. The record-time time-alignment check goes with it, since without the prediction there is nothing to compare against; the delta's alignment is guaranteed by `StepOutput.stack_diagnostics`'s construction. A future sub-aggregator that genuinely needs the prediction can widen the protocol when it arrives.

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: There is a concrete reason: explicitly inheriting from a `Protocol` class turns its `...` method stubs into inherited concrete methods (returning `None`), so a subclass that forgets to implement one would pass type checking and fail silently at runtime. With structural typing, conformance is checked where the aggregator dict is annotated as `dict[str, StepDiagnosticsSubAggregator]`, and a missing method is a type error there. Happy to add the explicit base if you prefer the discoverability, but the structural check is stricter.

mcgibbon: Maybe consider using ABCs if you want hard inheritance-type contracts. We only need to use Protocol when we need to support taking in an API the local context doesn't "control". If you want better type-checking of protocol implementation I sometimes put lines like:

```
_: type[MyProtocol] = MyNewClass
```

Below the class definition.

fme/ace/aggregator/inference/step_diagnostics.py
```diff
@@ -0,0 +1,450 @@
...
+        self,
+        step_diagnostics: StepDiagnostics | None,
+        i_time_start: int,
+    ) -> None: ...
+
+    def summary_logs(self) -> dict[str, Any]: ...
+
+    def time_series_logs(self, step_slice: slice) -> dict[str, Any]: ...
+
+    def diagnostics(self) -> dict[str, xr.Dataset]: ...
```
mcgibbon, line 48: Suggestion (optional): Use `StepDiagnosticsSubAggregator = CorrectionDeltaAggregator` (or whatever the correct form is for type variable assignment, and update this to equal the union of types when we have more than one such example. Or I don't mind you leaving it as-is, especially if this is replicating an existing pattern.

Optional: If you want a stronger guarantee that new implementations have the API, you could use an ABC instead of protocol.

mcgibbon (approved):
LGTM, only optional comments.

fme/ace/aggregator/inference/step_diagnostics.py
jpdunc23: Thanks for the suggestion, I added a type conformance check `_: type[StepDiagnosticsSubAggregator] = CorrectionDeltaAggregator`.

This is replicating the existing pattern for other aggregators (e.g., `SubAggregator(Protocol)` in fme/ace/aggregator/inference/data.py), so I think keeping the `Protocol` makes sense for now. That said, if you wanted to switch to an ABC for future `StepDiagnosticsSubAggregator` additions I wouldn't object.

PR #1348, opened 2026-07-08
Adding partial convolution option in Samudra

William-gregory:
This PR adds a partial convolution option to the Samudra model, motivated by the need to reduce grid-scale noise in Samudra rollouts (see comment below). The core change are to:

- `fme.ace.models.ocean.m2lines.samudra`
- `fme.ace.models.ocean.m2lines.layers`
- `fme.ace.registry.m2lines`

The partial convolution option has also been included in `fme.ace.models.ocean.m2lines.test_samudra`.

The PR relies on `in_names` in the `ModuleConfig` and `ModuleSelector` build methods. This was needed so that masks can be aligned with individual channels in the input tensor. Therefore, the PR also includes updates to:

- `fme.ace.models.registry`
- `fme.core.models.mlp`
- `fme.core.registry`
- `fme.core.step.single_module`


fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -3,11 +3,272 @@
...
+        super().__init__()
+        self.norm = nn.LayerNorm(channels)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = x.permute(0, 2, 3, 1)
+        x = self.norm(x)
+        return x.permute(0, 3, 1, 2).contiguous()
+
+
+class PartialConv2d(torch.nn.Module):
```
jpdunc23, line 23: What is the provenance of this class? The idea seems to come from https://github.com/NVIDIA/partialconv/blob/master/models/partialconv2d.py but the implementation looks pretty different. 

Assuming this is related to the NVIDIA project, please include a link to https://arxiv.org/abs/1804.07723 in the docstring.

Or if there is some other provenance, we give attribution somehow.

jpdunc23 (commented):
Thank you for adding this! These are very promising results, so we should definitely try to get this merged without too much delay.

For the PR description, please follow https://github.com/ai2cm/ace/blob/main/.github/pull_request_template.md. Plots are good for PR conversation, but not so good in the description which becomes the body of the commit message once the PR is squashed-and-merged.  Also, I recommend a more formal style for the description, e.g. avoiding first person ("Adds xyz feature to" vs "I'm adding this feature because"). The exact content of your current description would be good to copy paste into a comment in the PR discussion before making edits.

Asking @elynnwu if she can help review.

fme/ace/registry/m2lines.py
```diff
@@ -78,6 +83,7 @@ def build(
         n_in_channels: int,
         n_out_channels: int,
         dataset_info: DatasetInfo,
+        in_names: list[str] | None = None,
```
jpdunc23, line 86: Hmm, I think we can avoid this change. We have `dataset_info` here which has an attribute `spatial_mask_provider`, so ideally we should be able to construct the mask tensor with the right network channel order. The hard part is ensuring that the order is the same as the packer order.

@mcgibbon any thoughts?

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -3,11 +3,272 @@
...
 
+class ChannelLayerNorm(nn.Module):
+    def __init__(self, channels: int):
+        super().__init__()
+        self.norm = nn.LayerNorm(channels)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = x.permute(0, 2, 3, 1)
+        x = self.norm(x)
+        return x.permute(0, 3, 1, 2).contiguous()
```
jpdunc23, line 20: We should be able to avoid the permute here, which can be quite slow. See e.g. https://github.com/ai2cm/ace/blob/437397b8f8d16d9c5b525dd19a20ee6f01f53907/fme/core/models/conditional_sfno/layers.py#L95

jpdunc23: @mcgibbon how do you feel about models outside of `core/models/conditional_sfno` reusing some pieces? Or would it be better to copy-paste the implementation?

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -79,26 +92,87 @@ def __init__(
...
-                    a,
-                    b,
-                    dilation=self.dilation[i],
-                    n_layers=self.n_layers[i],
-                    pad=self.pad,
-                    norm=self.norm,
-                    norm_kwargs=self.norm_kwargs,
-                    upscale_factor=self.upscale_factor,
-                    checkpoint_strategy=self.checkpoint_strategy,
+            if i == 0:
```
jpdunc23, line 147: Why is it that only the first layer uses partial convolutions? Do the masks not flow through the layers?

jpdunc23: This seems somewhat incongruous with having `MaskedAvgPool2d` for every down layer, but maybe there's a good reason to do it this way.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -3,11 +3,272 @@
...
+
+        y = self.norm(features)
+        y = self.expand(y)
+        y = self.activation(y)
+        y = self.contract(y)
+
+        output = self.skip(residual_input) + y
+
+        # Latent features are only valid over actual ocean columns.
+        return output * ocean_column_mask
```
jpdunc23, line 236: If I understand correctly, `ocean_column_mask` is just the surface mask. The invalid area of the surface mask (i.e., land) is always a subset of the invalid area of deeper layers, so does `output * ocean_column_mask` actually change the output in any way?

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -6,8 +6,15 @@
...
+    BilinearUpsample,
+    ConvNeXtBlock,
+    MaskAwareInputBlock,
+    MaskedAvgPool2d,
+)
 from fme.ace.models.ocean.m2lines.utils import pairwise
+from fme.core.dataset_info import DatasetInfo
 
 
 class Samudra(torch.nn.Module):
```
jpdunc23, line 20: With this amount of changes to an existing model, I'm tempted to create a new class (e.g., `MaskAwareSamudra`) rather than risk introducing potentially subtle changes to existing checkpoints.

@elynnwu thoughts?

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -98,9 +359,7 @@ def __init__(
             )
         # Layer Norm
         elif norm == "layer":
-            convblock.append(
-                torch.nn.LayerNorm(in_channels * upscale_factor, **self.norm_kwargs)
-            )
+            convblock.append(ChannelLayerNorm(in_channels * upscale_factor))
```
elynnwu, line 362: This changes state-dict keys (`convblock.N.weight` → `convblock.N.norm.weight`), which breaks loading any existing `norm="layer"` Samudra checkpoint. This also drops `**self.norm_kwargs`, is this intended? Our baseline Samudra uses instance norm, so it may not be as critical.

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Yes, it seems like there're a lot of changes here that might be worth splitting it off. The one concern for doing that is we'll eventually diverge too much. 

I'm wondering if we should do a bit more ablations to nail down exactly what is causing this noisy behavior. I've previously also had this issue that eventually (mostly) went away due to the way masks are defined and training longer.

elynnwu: If possible, maybe we should try a these ablations:
1. Input masking only. Multiply the input by the per-channel wet mask (`x * channel_mask`) before the existing layer 0. No architecture change at all — layer 0 stays a plain` ConvNeXtBlock`, pooling stays plain `AvgPool`
2.  Input masking + mask-as-feature. Same as above, plus concatenate the per-channel mask as extra input channels (so layer 0's `in_channels` doubles, but it's still an ordinary convolution — no partial-conv renormalization, no masked pooling)

This will tell us if the effect is coming from masking the input, mask-conditioning, or the full partial-conv in this PR.

William-gregory:
Thanks @jpdunc23 and @elynnwu for looking through all of this! I'm pasting my original PR description with the figures here, following recommendation from @jpdunc23. I will look into your comments and ablation suggestions asap!

"In my training of uncoupled Samudra (using OM4 data), I was finding that my resultant bias maps were quite noisy. The noise was most prevalent near the surface, but could be seen at all depths. I was wondering if this noise was being introduced by convolutions and average pooling being computed across ocean and land grid cells at all depths, which then causes noise to mix across the channels. I therefore added a partial convolution operation to Samudra, which can be turned on in the config by `partial_convolutions: true`. I've found that this both reduces noise in the resultant bias maps and makes the model converge faster. The wandb plots below show simass rmse and sst bias during inline inference. The simass error without partial convolutions has a pretty consistent profile to what I've seen across my OM4 training and Ai2's CM4 training (flat for some number of epochs, and then begins to drop). With partial convolutions, simass error starts to drop from first epoch. The sst bias map then shows much less grid-scale noise.

To add the partial convolution functionality, I had to edit the `ModuleConfig` and `ModuleSelector` build methods (and therefore `core.step.single_module`) to pass `in_names` to the model. This is so that I can align each channel with its corresponding mask in `dataset_info`. Note that the partial convolution is just applied to the first layer of the network, where channels are defined on the physical domain."

simass_inline_inference
sst_bias_map

fme/ace/models/ocean/m2lines/samudra.py
William-gregory: Thanks @elynnwu. Just to clarify on 1. Is this different to what the original code is doing already? I was under the impression that inputs/outputs are already being masked before/after each forward pass?

fme/ace/models/ocean/m2lines/layers.py
William-gregory: Good point on the backwards compat. I think the code crashed when i tried layer norm with the initial `torch.nn.LayerNorm(in_channels * upscale_factor, **self.norm_kwargs)` strategy and it seemed I needed the permute step. Given we're just using instance norm anyway, maybe its best if I restore this to its original implementation?

William-gregory:
hey @jpdunc23 and @elynnwu, I did some additional testing based on @elynnwu's recommendations and found that the partial convolution isn't actually the reason behind these improvements! It is just coming from masking the input channels prior to the first `ConvNeXtBlock`. You can see this implementation on this new branch https://github.com/William-gregory/ace/commit/2ae6a6f0707dbae6d43f15286ab83f4e8903555d. I'm a little confused why this makes a difference, as I thought inputs were already being masked with `input_masking` in the stepper config. Would be good to hear your thoughts on this. If you think this alternative approach is worth pursuing then let me know whether its best to close this PR and open a new one.

elynnwu:
Thanks for the update @William-gregory . We just discussed this in the sync today, to clarify, you're applying the masking in the latent space right?

We're also wondering whether we should revisit different fill strategy in `input_masking`, specifically using `mean`:
```yaml
stepper:
  input_masking:
    mask_value: 0
    fill_value: mean
```

jpdunc23:
> Thanks for the update @William-gregory . We just discussed this in the sync today, to clarify, you're applying the masking in the latent space right?

Looked at this a bit closer after our sync. If I'm reading things right in https://github.com/William-gregory/ace/commit/2ae6a6f0707dbae6d43f15286ab83f4e8903555d, it does seem that Will's masking is in physical space, so I think what he's done should be equivalent to `stepper.input_masking.fill_value='mean'`. Still worth trying `stepper.input_masking.fill_value='mean'` with an experiment off of `main` to verify we see the same result.

William-gregory:
> Thanks for the update @William-gregory . We just discussed this in the sync today, to clarify, you're applying the masking in the latent space right?
> 
> We're also wondering whether we should revisit different fill strategy in `input_masking`, specifically using `mean`:
> 
> ```yaml
> stepper:
>   input_masking:
>     mask_value: 0
>     fill_value: mean
> ```

Thanks @elynnwu and @jpdunc23 for this suggestion. I can confirm that using `fill_value: mean` reproduces the result I had in https://github.com/William-gregory/ace/commit/2ae6a6f0707dbae6d43f15286ab83f4e8903555d, and gives (in my case) better training and less noisy rollouts than my default `fill_value: 0.0`. 

Sorry for making you go through all this effort reviewing the partial convolution code! I really appreciate all the input and glad we got to the bottom of it. If you're happy, I can close this PR.

jpdunc23:
> Thanks @elynnwu and @jpdunc23 for this suggestion. I can confirm that using `fill_value: mean` reproduces the result I had in [William-gregory@2ae6a6f](https://github.com/William-gregory/ace/commit/2ae6a6f0707dbae6d43f15286ab83f4e8903555d), and gives (in my case) better training and less noisy rollouts than my default `fill_value: 0.0`.
> 
> Sorry for making you go through all this effort reviewing the partial convolution code! I really appreciate all the input and glad we got to the bottom of it. If you're happy, I can close this PR.

@William-gregory No apologies necessary! In fact, this has been super helpful since we haven't revisited the decision to use `fill_value: 0.0` in a long while. In particular, we last looked at this before switching away from applying Gaussian smoothing to the 0.25 degree data prior to regridding, and I wonder if that choice explains why `fill_value: mean` is now working better.

I'll close this PR, but please feel free to re-open if you decide to experiment a bit more with partial convolutions.

PR #1350, opened 2026-07-08
Serialize ModuleConfig defaults in ModuleSelector.config

### LINKED ISSUE #596: Default `ModuleConfig` args not captured in wandb config parameters
For example, see [this run overview](https://wandb.ai/ai2cm/ace-samudra-cm4/runs/4nrk3kev/overview) and search for "builder" in the Config parameters section. `SamudraBuilder`'s default `pad` value is not captured.

This could cause issues if defaults ever change and make comparisons against default settings more difficult from the wandb UI.

DaPillah:
When a run is logged to Weights & Biases, only the config values explicitly set in the yaml file were captured. Default field values defined on the `ModuleConfig` dataclass were silently omitted, because `ModuleSelector` kept the raw yaml dict in its `config` field and never serialized the defaults that Python fills in when the `ModuleConfig` is instantiated. This PR normalizes `ModuleSelector.config` to the built instance's fully-defaulted values so defaults are captured wherever the config is serialized (wandb, checkpoints, logs).

Changes:
- `fme.core.registry.module.ModuleSelector.__post_init__`: after building the ModuleConfig instance, set `self.config = dataclasses.asdict(self._instance)` so default values are included in the serialized config.
- `fme.core.registry.test_module_registry`: add a mock builder with a default field and a test asserting the default appears in the serialized config (fails before the change, passes after).

Backwards compatibility: loading is unaffected. `ModuleConfig.from_state` uses `dacite(strict=True)` and every added key is a valid field of the target dataclass, so both older (partial) and newer (full) configs load correctly; the frozen/latest module backwards-compatibility tests still pass.

- [x] Tests added

Resolves #596

### DIFF
--- a/fme/core/registry/module.py
+++ b/fme/core/registry/module.py
@@ -156,6 +156,10 @@ class ModuleSelector:
                 f"got {self.type} (available: {CONDITIONAL_BUILDERS})"
             )
         self._instance = self.registry.get(self.type, self.config)
+        # Normalize config to include the built ModuleConfig's default values,
+        # so that defaults are captured when the config is serialized (e.g.
+        # logged to Weights & Biases). See issue #596.
+        self.config = dataclasses.asdict(self._instance)
 
     @property
     def module_config(self) -> ModuleConfig:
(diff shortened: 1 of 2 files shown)


mcgibbon (approved):
LGTM

PR #1353, opened 2026-07-09
Add `apply_mean` option to `StandardNormalizer.normalize` and a `NormalizeFn` protocol

jpdunc23:
Adds an explicit uncentered-normalize option to `StandardNormalizer`: `normalize(tensors, apply_mean=False)` divides by the standard deviation without subtracting the mean. This is helpful for normalizing increments, e.g. the corrector's `delta = corrected - network_output`, where subtracting the mean of the underlying field is not meaningful and the alternative is trusting the algebra of `normalize(pred) - normalize(pred - delta)`.

A named `NormalizeFn` protocol makes the option usable through a passed-in normalize function: a `Protocol` rather than a `Callable` alias so the kwarg keeps its name and default, leaving existing `normalize(x)` call sites valid. Its `tensors` parameter is positional-only so implementations may name their first parameter freely.

Changes:
- `fme.core.normalizer.StandardNormalizer.normalize` gains `apply_mean: bool = True`; when False it divides by the standard deviation without subtracting the mean (name filtering and NaN filling unchanged; `denormalize` and checkpoint serialization untouched)
- `fme.core.normalizer.NormalizeFn` protocol added; the normalize-callable annotations in `fme.ace.aggregator.inference.main` and `fme.coupled.aggregator` adopt it (no behavior change at call sites)

- [x] Tests added


mcgibbon (approved):
Wasn't expecting to need all these changes, but points out to me even more strongly that this is the right choice, since not all normalization strategies that exist are standard-scaling and alternative normalization schemes (if we ever use one) need to handle this flag differently.

PR #1354, opened 2026-07-09
Exempt `prescribed_prognostic_names` from post-corrector overlap guard

jpdunc23:
The post-corrector overlap guard in `step_with_adjustments` raised whenever any post-corrector adjustment wrote a variable the corrector already modified, treating `prescribed_prognostic_names` and the ocean surface temperature prescription identically. But prescribing a corrector-modified variable is an intentional "use the given value regardless of what the network or corrector produced" overwrite, not a misconfiguration, and blocking it prevented valid inference configurations from running. This narrows the guard so it only fires for the ocean prescription, and drops prescribed names from the reported corrector delta so the delta stays exact for every variable it still contains (the overwrite discards the corrector's value for a prescribed variable, so its delta no longer describes the returned output). The ocean overlap remains a hard error, and the guard runs against the full corrector delta before the filter, so an ocean surface temperature name that is also prescribed still raises.

Changes:
- `fme.core.step.single_module.step_with_adjustments`: overlap guard now checks
  only the ocean `surface_temperature_name`; `prescribed_prognostic_names` are
  exempt. Prescribed names are dropped from the returned
  `corrector_diagnostics.delta` (a no-op when the corrector never touched them).
- `fme/core/step/test_step.py`: retargeted the prescribed/corrected collision
  test to assert it now runs, yields the prescribed value, and omits that
  variable from the delta; added coverage that the ocean overlap still raises,
  that an ocean name which is also prescribed still raises, and that a
  multi-step rollout with a prescribed/corrected collision stacks
  `StepDiagnostics` without failing on inconsistent delta keys.

- [x] Tests added


Arcomano1234 (approved):
This looks good to me, thanks for getting the fix so quickly.

PR #1358, opened 2026-07-13
Allow filter_preserves_global_mean with spectral_ratio < 1 in the noise-conditioned SFNO

mcgibbon:
The noise-conditioned SFNO previously rejected `filter_preserves_global_mean` together with `spectral_ratio < 1`. This makes them compatible.

`filter_preserves_global_mean` means the spectral filter must not act on the l=0 (global-mean) mode: the per-mode weight at l=0 is bypassed so the filter cannot do global-level reasoning (its l=0 weight trains to zero gradient). The existing `spectral_ratio == 1` path already does this by passing the input's l=0 coefficient through unchanged in the spectral domain (`torch.cat([x[..., :1, :], xp[..., 1:, :]])`). That same passthrough works at `spectral_ratio < 1`: the l=0 that flows through is the reduced-channel bottleneck (`pre_proj`) representation, so it passes through the shared `pre_proj`/`post_proj` channel projections — a fixed mix, not per-mode reasoning — rather than the l=0-specific weight. The global mean is therefore largely, not exactly, preserved; this is the intended trade for keeping the filter free of l=0 reasoning without extra machinery.

Because the passthrough happens in the spectral domain, gated by `_has_global_mean` (the `l_start == 0` rank), it is distributed-correct exactly like the `spectral_ratio == 1` path — so the combination is now supported under spatial parallelism as well.

Changes:
- `fme.core.models.conditional_sfno.s2convolutions.validate_spectral_ratio`: no longer rejects `preserves_global_mean` with `spectral_ratio < 1`; docstring updated.
- `fme.core.models.conditional_sfno.s2convolutions.SpectralConvS2`: the l=0 spectral-domain passthrough now runs at any `spectral_ratio`; class docstring documents what `preserve_global_mean` does and the exact-vs-approximate distinction across ratios.
- Tests: replaced the config-rejection tests with a build test and behavioral tests asserting the l=0 per-mode weight is bypassed (no gradient, output invariant to it) at `spectral_ratio < 1`, at both `SpectralConvS2` and `SFNONet` level; added a combined end-to-end/gradient test; added a `@pytest.mark.parallel` test that the combination builds and forwards under spatial parallelism.

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/s2convolutions.py
+++ b/fme/core/models/conditional_sfno/s2convolutions.py
@@ -152,9 +152,9 @@ class SpectralConvS2(nn.Module):
     (global-mean) mode: the per-mode weight at l=0 is bypassed, so the filter
     cannot do global-level reasoning. Its l=0 weight then trains to zero
     gradient. At ``spectral_ratio == 1`` this holds the global mean exactly; at
-    ``spectral_ratio < 1`` the l=0 still passes through the shared channel
-    projections (pre_proj/post_proj), which is a fixed mix, not per-mode
-    reasoning, so the mean is largely but not exactly preserved.
+    ``spectral_ratio < 1`` the l=0 instead passes through the shared channel
+    projections (pre_proj/post_proj), so the mean is no longer held fixed --
+    but that is a fixed channel mix, not per-mode l=0 reasoning.
     """
 
     def __init__(
@@ -415,9 +415,12 @@ class SpectralConvS2(nn.Module):
             )
             xp = xp + self.lora_scaling * lora_update
             if self._preserve_global_mean and self._has_global_mean:
-                # Bypass the l=0 per-mode weight: pass the input's l=0 through
-                # unchanged so the filter does no global-mean reasoning. See the
-                # class docstring for the exact-vs-approximate behavior by ratio.
+                # Skip the per-mode weight at l=0 by passing the input's l=0
+                # through unchanged, so the filter cannot reason about or alter
+                # the global mean. With spectral_ratio < 1 this l=0 is the
+                # bottleneck (pre_proj) representation, so the mean still passes
+                # through the shared pre/post projections rather than being held
+                # exactly; that is intended (see the class-level note).
                 xp = torch.cat([x[..., :1, :], xp[..., 1:, :]], dim=-2)
             xp = xp.reshape(B, self.spectral_channels, H, W)
             x = xp.contiguous()
--- a/fme/core/models/conditional_sfno/s2convolutions.py
+++ b/fme/core/models/conditional_sfno/s2convolutions.py
@@ -240,6 +240,22 @@ class SpectralConvS2(nn.Module):
                 f"and out_channels={out_channels}."
             )
         if self._reduced_global_mean:
+            # The grid-space global-mean swap reads per-latitude quadrature
+            # weights from the transforms' ``weights`` buffer, which is only the
+            # full-grid, m=0 quadrature on a single spatial rank. Under spatial
+            # parallelism the buffer's latitude axis stays global (so it does
+            # not match the lat-sharded grid input) and its leading axis is the
+            # local-m partition (so ``[0, 0]`` is not the m=0 row off rank 0);
+            # the swap would then crash under H-parallelism and silently compute
+            # the wrong global mean under W-parallelism. The spectral_ratio == 1
+            # path preserves the global mean correctly under spatial parallelism
+            # via the l=0 coefficient on the l_start == 0 rank; matching that
+            # here is left as follow-up.
+            dist.require_no_spatial_parallelism(
+                "preserve_global_mean with spectral_ratio < 1 does not support "
+                "spatial parallelism; the grid-space global-mean swap needs the "
+                "single-rank, full-grid m=0 quadrature weights."
+            )
             # Latitude quadrature weights for the input (forward) grid and the
             # output (inverse) grid, uniform over longitude. inverse_transform
             # is an InverseRealSHT and carries no forward quadrature, so build a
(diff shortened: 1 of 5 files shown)


mcgibbon:
Claude (pre-review agent): Applied a fix to this PR — pushed as `9df238ace`.

**Finding (blocking): the grid-space global-mean swap is incorrect under spatial model parallelism.**

`_swap_global_mean` reads per-latitude quadrature weights from the SHT transform's `weights` buffer (`_latitude_quadrature_weights`, i.e. `weights[0, 0, :]`). That buffer is only the full-grid, m=0 quadrature on a **single spatial rank**. Under `torch_harmonics` distributed transforms:

- **H (polar) parallelism:** the buffer's latitude axis stays *global* (`weights` is split along m, not lat), while the block's grid input is lat-sharded. The `(1,1,nlat_global,1)` weight fails to broadcast against the local shard. Verified with a 2-rank probe (`FME_DISTRIBUTED_H=2`): `RuntimeError: The expanded size of the tensor (8) must match the existing size (16) at non-singleton dimension 2`.
- **W (azimuth) parallelism:** `weights` is split along its leading axis (m), so `weights[0, 0, :]` is the *local* first m — not the global m=0 row — off rank 0. The swap runs but silently returns the wrong global mean. Verified (`FME_DISTRIBUTED_W=2`): l=0 coefficients differ by up to 0.34 abs / 3.7 rel vs. the reference.

For contrast, the existing `spectral_ratio == 1` path (l=0 copy on the `l_start == 0` rank) is distributed-correct under both H=2 and W=2 — I verified this too. So this is a new gap specific to the reduced-channel path, not a pre-existing one.

**Fix applied:** reject `preserve_global_mean` + `spectral_ratio < 1` at construction under spatial parallelism via `Distributed.require_no_spatial_parallelism` (raises `SpatialParallelismNotImplemented`), plus a `@pytest.mark.parallel` regression test in `test_sfnonet.py` that asserts the guard fires under spatial parallelism while `spectral_ratio == 1` still builds. Single-rank behavior is unchanged; the feature works correctly on a single spatial rank.

**For the author / Jeremy to decide:** if the runs that need this feature will use spatial model parallelism, the guard makes the feature unusable and full distributed support is required (correct approach: local-lat-sliced quadrature weights matching the grid shard, with the true m=0 row — likely sourced from the model's area weights rather than the split SHT buffer). If those runs are single-spatial-rank, the guard is just a safety net and nothing more is needed.

mcgibbon:
Claude: Rewrote this PR (commit feb9e1a0d) to a much simpler form after clarifying intent with Jeremy. The earlier approach reconstructed the *exact* original C-channel global mean out of the reduced-channel bottleneck via a grid-space quadrature swap, which required `_swap_global_mean`, `_latitude_quadrature_weights`, an in==out restriction, a throwaway `out_transform`, quadrature-weight buffers, and a spatial-parallelism guard — hence the heavy commenting.

The real intent of `preserve_global_mean` is only that the l=0 per-mode filter weight is not applied (no global-level reasoning), not exact mean preservation. The existing `spectral_ratio == 1` spectral-domain l=0 passthrough already does that and works unchanged at `spectral_ratio < 1`, so all of the above was removed. Net source change vs main is now the rejection deletion plus a docstring paragraph and one short inline comment. As a bonus the spectral-domain passthrough is distributed-correct via `_has_global_mean`, so spatial parallelism is now supported instead of rejected.

Verified locally: the `s2convolutions`/`sfnonet` test modules pass on CPU, the new `@pytest.mark.parallel` test passes under both H=2 and W=2 via torchrun, and pre-commit (ruff/ruff-format/mypy) is clean on the changed files.

mcgibbon:
Pre-review agent (Claude): applied two documentation-only fixes to the PR branch (commit aedcde0), no runtime change.

1. `test_sfnonet_spectral_ratio_allows_preserve_global_mean` carried a stale comment describing the earlier, deleted approach ("the l=0 coefficient is restored in full-channel grid space, outside the reduced-channel bottleneck"). That is the opposite of what the merged code does — the l=0 per-mode weight is bypassed in the spectral domain, inside the reduced-channel representation. Replaced it with an accurate one-liner.
2. Trimmed the forward-path l=0 comment in `SpectralConvS2.forward` from six lines to three: it restated the class docstring's exact-vs-approximate explanation (and "cannot alter the global mean" read slightly at odds with the "not held exactly" caveat below it). The class docstring remains the single authoritative explanation.

Verified independently: main already ran the l=0 spectral-domain passthrough ungated by `spectral_ratio`; the only functional change vs main is dropping the `NotImplementedError` in `validate_spectral_ratio`, so the net diff is minimal and the intermediate grid-space machinery never reached main. Tests pass locally (CPU): `test_s2convolutions.py` + `test_sfnonet.py` 35 passed / 3 skipped; the `@pytest.mark.parallel` spatial-parallel test passes under torchrun --nproc-per-node 2.

fme/core/models/conditional_sfno/s2convolutions.py
elynnwu: `preserves_global_mean` `global_mean_arg_name` are not referenced in `validate_spectral_ratio` anymore, is there a reason to keep the signature here?

fme/core/models/conditional_sfno/s2convolutions.py
```diff
@@ -60,7 +60,9 @@ def validate_spectral_ratio(
         channels_name: name of ``channels`` to use in error messages.
         num_groups_name: name of ``num_groups`` to use in error messages.
         filter_type: if given, ``spectral_ratio < 1`` requires ``"linear"``.
-        preserves_global_mean: if True, ``spectral_ratio < 1`` is rejected.
+        preserves_global_mean: passed by callers but does not gate
+            ``spectral_ratio``; the two are compatible (see
+            ``SpectralConvS2.forward``).
         global_mean_arg_name: name of the global-mean flag for error messages.
```
elynnwu: doc should be updated here

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Claude: Good catch — no reason. They were dead once the function stopped gating on the global-mean flag. Removed `preserves_global_mean` and `global_mean_arg_name` from the signature (and the three call sites) in 3b6dcf791.

fme/core/models/conditional_sfno/s2convolutions.py
mcgibbon: Claude: Resolved by removing the two now-unused params — their docstring entries went with them. The compatibility note lives on `SpectralConvS2` where the passthrough actually happens (3b6dcf791).

elynnwu (approved):

PR #1359, opened 2026-07-13
Refactor HEALPix UNet configs/models to the Builder pattern (Rule 2)

mcgibbon:
Refactors the HEALPix UNet stack to the Builder pattern (Rule 2): each builder `Config` constructs its sub-objects from sub-config fields and passes the built objects to the implementation, so no implementation class receives a sub-config to build.

This is the pure Rule 2 follow-up to #1326 (HEALPix UNet code cleanup), which already introduced `HEALPixBuildContext`, made channel counts flow as explicit `.build()` args, and removed the in-place config mutation (`encoder.input_channels = …`, `.nside`) and the `dataclasses.replace(...).build()` pattern. Those violations are already gone on `main`; the residual addressed here is that implementation classes still called `.build()` on their sub-configs.

The three layers:

- **UNet.** `HEALPixUNet.__init__` no longer builds anything. `HEALPixUNetBuilder.build` (via a private `_build`) constructs the `HEALPixBuildContext`, validates the encoder/decoder level counts and the `nside` sequence, builds the encoder and decoder `nn.Module`s, and passes the built modules plus the runtime scalars the impl needs (`input_channels`, `output_channels`, and the resolved `nside` tuple used by `forward`'s face-size check) into `HEALPixUNet`. The impl holds no configs.
- **Encoder / decoder.** The per-level construction loop (channel threading and nside-ratio validation) moves from `UNetEncoder`/`UNetDecoder.__init__` into `UNetEncoderConfig._build` / `UNetDecoderConfig._build`. The config builds the ordered per-level modules (encoder: an `nn.Sequential(down?, conv)` per level; decoder: a per-level `DecoderLevel` module — a small `nn.Module` bundling the optional upsample + skip-concat + conv and running that branching in its own `forward` — plus the `output_layer`) and passes the built `nn.ModuleList`/modules to the impl, which just stores them. `UNetDecoder.forward` iterates the typed levels rather than reaching into string-keyed module dicts, so the impl is directly constructible from its signature and its `forward` needs no knowledge of how the levels were assembled.
- **Activations.** The ~7 conv/upsample block impls no longer take a `CappedGELUConfig` and call `activation.build()`. Each block config now passes an **activation factory** — the bound `activation.build` callable (or `None`) — as `activation_factory: Callable[[], nn.Module] | None`, and each impl calls `activation_factory()` exactly where it used to call `activation.build()`. This is deliberate: `CappedGELU` registers a `cap` buffer (so it appears in `state_dict`), and a block needs N *independent* activation instances (N depends on `n_layers`), so a single shared built instance would be wrong. The factory keeps the module instantiation count and order — and thus `state_dict` — identical. `SmoothedInterpolateConv`, which previously received a single built activation module, is folded into the same factory form for family consistency.

**Behavior-preserving and checkpoint-compatible.** The set and ordering of constructed submodules — and therefore `state_dict` keys — is unchanged, so a HEALPix checkpoint trained on `main` still loads. Verified by building representative models (BasicConv / ConvNeXt / MultiSymmetricConvNeXt convs; transposed-conv and smoothed-interpolate upsampling; karlbauer, isolatitude, and earth2grid padding; dealiased downsample) on both `main` and this branch under a fixed seed and confirming the full `state_dict` dumps (keys, order, shapes, dtypes) and parameter values are byte-identical across all 36 combinations.

Changes:
- `fme.ace.registry.hpx`: `HEALPixUNetBuilder.build` delegates to `_build`, which builds the context, validates, builds encoder/decoder, and assembles `HEALPixUNet`.
- `fme.ace.models.healpix.healpix_unet`: `HEALPixUNet.__init__` takes built encoder/decoder modules plus runtime scalars; no building, no configs.
- `fme.ace.models.healpix.healpix_encoder`, `.healpix_decoder`: construction loop hoisted into `UNetEncoderConfig._build` / `UNetDecoderConfig._build`; impls store built modules. The decoder's per-level unit is a typed `DecoderLevel` (submodules named `upsamp`/`conv`, so `state_dict` keys and order are unchanged from the previous `ModuleDict`); the encoder keeps its `nn.Sequential`-per-level structure.
- `fme.ace.models.healpix.healpix_blocks`: block impls take `activation_factory`; block configs pass `activation.build`; impls no longer reference `CappedGELUConfig`.
- `fme.ace.registry.test_hpx`: construct encoder/decoder/UNet through the config build path to match the new signatures (behavior identical).

- [x] Tests added (existing hpx tests updated to build via the config path; behavior-preserving, no new behavior to cover)

### DIFF
--- a/fme/ace/models/healpix/healpix_decoder.py
+++ b/fme/ace/models/healpix/healpix_decoder.py
@@ -91,7 +91,7 @@ class UNetDecoderConfig:
         n_channels = self.n_channels
         n_levels = len(n_channels)
 
-        decoder: List[nn.Module] = []
+        decoder: list[DecoderLevel] = []
         for n, curr_channel in enumerate(n_channels):
             up_sample_module = None
             level_nside = (
@@ -139,7 +139,7 @@ class UNetDecoderConfig:
         )
 
         return UNetDecoder(
-            decoder=nn.ModuleList(decoder),
+            decoder=decoder,
             output_layer=output_layer,
         )
 
@@ -185,17 +185,18 @@ class UNetDecoder(nn.Module):
 
     def __init__(
         self,
-        decoder: nn.ModuleList,
+        decoder: list[DecoderLevel],
         output_layer: nn.Module,
     ):
         """
         Args:
             decoder: Ordered per-level :class:`DecoderLevel` modules built by
-                :meth:`UNetDecoderConfig._build`.
+                :meth:`UNetDecoderConfig._build`; wrapped in an
+                ``nn.ModuleList`` here for submodule registration.
             output_layer: Final output-projection module.
         """
         super().__init__()
-        self.decoder = decoder
+        self.decoder = nn.ModuleList(decoder)
         self.output_layer = output_layer
 
     def forward(self, inputs: Sequence[torch.Tensor]) -> torch.Tensor:
--- a/fme/ace/models/healpix/healpix_decoder.py
+++ b/fme/ace/models/healpix/healpix_decoder.py
@@ -78,9 +78,9 @@ class UNetDecoderConfig:
     ) -> "UNetDecoder":
         """Construct the ordered per-level decoder modules plus output layer.
 
-        Builds a per-level ``{upsamp, conv}`` module dict (threading channel
-        counts and validating the nside upsample ratios) and the output layer,
-        then passes the built modules to :class:`UNetDecoder`.
+        Builds one :class:`DecoderLevel` per level (threading channel counts and
+        validating the nside upsample ratios) and the output layer, then passes
+        the built modules to :class:`UNetDecoder`.
         """
         dilations = self.dilations
         if dilations is None:
@@ -129,14 +129,7 @@ class UNetDecoderConfig:
                 ctx=ctx.layer(n_levels - 1 - n),
             )
 
-            decoder.append(
-                nn.ModuleDict(
-                    {
-                        "upsamp": up_sample_module,
-                        "conv": conv_module,
-                    }
-                )
-            )
+            decoder.append(DecoderLevel(upsamp=up_sample_module, conv=conv_module))
 
         output_layer = self.output_layer.build(
             in_channels=curr_channel,
@@ -151,8 +144,44 @@ class UNetDecoderConfig:
         )
 
 
+class DecoderLevel(nn.Module):
+    """One decoder level: optional upsample + skip-concat, then a conv block.
+
+    The deepest level (built first, ``upsamp is None``) receives the encoder
+    bottleneck and applies only its conv block. Each shallower level upsamples
+    the running activation, concatenates the matching encoder skip connection
+    along the channel dimension, then applies its conv block. Bundling the
+    upsample and conv into one module keeps this per-level branching structure
+    with the module that runs it, so :class:`UNetDecoder.forward` never reaches
+    into the level's internals.
+    """
+
+    def __init__(self, upsamp: Optional[nn.Module], conv: nn.Module):
+        super().__init__()
+        self.upsamp = upsamp
+        self.conv = conv
+        self.channel_dim = 1
+
+    def forward(self, x: torch.Tensor, skip: torch.Tensor) -> torch.Tensor:
+        """
+        Args:
+            x: Running activation from the deeper level.
+            skip: Matching encoder skip connection (unused when ``upsamp`` is
+                ``None``, i.e. the deepest level).
+        """
+        if self.upsamp is not None:
+            x = torch.cat([self.upsamp(x), skip], dim=self.channel_dim)
+        return self.conv(x)
+
+
 class UNetDecoder(nn.Module):
-    """Generic UNetDecoder that can be applied to arbitrary meshes."""
+    """Applies the decoder levels deepest-to-shallowest, then the output layer.
+
+    Receives the ordered per-level :class:`DecoderLevel` modules and the output
+    layer already built by :meth:`UNetDecoderConfig._build`; it constructs
+    nothing itself, so it is directly constructible from its signature and its
+    ``forward`` needs no knowledge of how the levels were assembled.
+    """
 
     def __init__(
         self,
@@ -160,15 +189,12 @@ class UNetDecoder(nn.Module):
         output_layer: nn.Module,
     ):
         """
-        Initialize the UNetDecoder.
-
         Args:
-            decoder: Ordered per-level ``{upsamp, conv}`` module dicts built by
+            decoder: Ordered per-level :class:`DecoderLevel` modules built by
                 :meth:`UNetDecoderConfig._build`.
             output_layer: Final output-projection module.
         """
         super().__init__()
-        self.channel_dim = 1
         self.decoder = decoder
         self.output_layer = output_layer
 
@@ -177,15 +203,13 @@ class UNetDecoder(nn.Module):
         Forward pass of the UNetDecoder.
 
         Args:
-            inputs: Sequence of tensors, one for each decoder level.
+            inputs: Sequence of tensors, one per decoder level (encoder outputs,
+                shallow-to-deep).
 
         Returns:
             The decoded values.
         """
         x = inputs[-1]
-        for n, layer in enumerate(self.decoder):
-            if layer["upsamp"] is not None:
-                up = layer["upsamp"](x)
-                x = torch.cat([up, inputs[-1 - n]], dim=self.channel_dim)
-            x = layer["conv"](x)
+        for n, level in enumerate(self.decoder):
+            x = level(x, inputs[-1 - n])
         return self.output_layer(x)
--- a/fme/ace/models/healpix/healpix_decoder.py
+++ b/fme/ace/models/healpix/healpix_decoder.py
@@ -62,65 +62,36 @@ class UNetDecoderConfig:
         Returns:
             UNet Decoder model.
         """
-        return UNetDecoder(
-            conv_block=self.conv_block,
-            up_sampling_block=self.up_sampling_block,
-            output_layer=self.output_layer,
-            n_channels=self.n_channels,
-            n_layers=self.n_layers,
-            output_channels=output_channels,
-            dilations=self.dilations,
-            ctx=ctx,
-        )
-
-
-class UNetDecoder(nn.Module):
-    """Generic UNetDecoder that can be applied to arbitrary meshes."""
-
-    def __init__(
-        self,
-        conv_block: ConvBlockConfig,
-        up_sampling_block: UpsamplingBlockConfig,
-        output_layer: ConvBlockConfig,
-        n_channels: Sequence = (64, 32, 16),
-        n_layers: Sequence = (1, 2, 2),
-        output_channels: int = 1,
-        dilations: Optional[list] = None,
-        ctx: HEALPixBuildContext | None = None,
-    ):
-        """
-        Initialize the UNetDecoder.
-
-        Args:
-            conv_block: Configuration for the convolutional block.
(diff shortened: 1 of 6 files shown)


mcgibbon:
Claude (pre-review): **Verdict — ready for review.** Independent pre-review, fresh session without the authoring context. No fixes applied; nothing warranted one.

**What I verified (not just trusted):**

- **Checkpoint/behavior preservation (the crux) — confirmed independently.** Built representative models on `origin/main` vs. this branch under a fixed seed and diffed full `state_dict` signatures (keys, order, shapes, dtypes, and per-tensor value hashes). Byte-identical across all four cases: BasicConv / ConvNeXt / MultiSymmetricConvNeXt convs × transposed-conv & smoothed-interpolate upsampling × karlbauer / isolatitude / earth2grid padding, plus dealiased downsample. Also confirmed the `forward` output is bit-identical for the isolatitude path (which exercises the `nside` face-size check). A `main`-trained HEALPix checkpoint will load unchanged.
- **Activation factory — correct.** The bound `activation.build` passed as `Callable[[], nn.Module] | None`, called once per activation site, produces the same number and order of independent instances as the old per-site `activation.build()`. `SmoothedInterpolateConv` (single site, previously handed a pre-built module) is folded in correctly — it now builds its one activation via `activation_factory()` at construction, identical in effect. Covered by the state_dict/forward diffs above.
- **Hoisted encoder/decoder loops — carried verbatim.** Channel threading and nside-ratio validation match `main`; module registration order (and thus `state_dict` order) is preserved. Validation is reached on every build path: `UNet*Config.build` does the nside-length check then delegates to `_build`, and `HEALPixUNetBuilder._build` re-checks levels/nside before building.
- **`DoubleConvNeXtBlock` — confirmed dead.** No `DoubleConvNeXtBlockConfig` exists and nothing instantiates it; it is unreachable through the registry. Converting it is harmless family-consistency scope creep.
- **Tests — not weakened.** They now build encoder/decoder/UNet through the real config path (arguably stronger). The public registry entry point `HEALPixUNetBuilder.build(dataset_info, …)` is exercised end-to-end by `test_hpx_init` and `test_HEALPixUNet_in_stepper`.

**Environment note:** unlike the authoring env, I installed `earth2grid` in the slot, so the earth2grid-gated paths actually ran — `test_hpx.py` = 59 passed, 7 skipped (all CUDA-only). `ruff` + `ruff-format` clean; `mypy` clean on the changed files (only the known rdkit-stubs env noise, absent in CI).

**For the reviewer's taste call (not blockers):**
- Whether to keep the dead `DoubleConvNeXtBlock` conversion in this PR or drop it to a separate cleanup.
- The `_build_hpx_unet` test helper calls the private `_build` directly; the public `build` path is covered separately by the stepper tests, so this is fine, but it's a stylistic choice worth a glance.

fme/ace/models/healpix/healpix_decoder.py
```diff
@@ -167,15 +138,40 @@ def __init__(
...
+            ctx=ctx.layer(0),
         )
 
+        return UNetDecoder(
+            decoder=nn.ModuleList(decoder),
+            output_layer=output_layer,
+        )
+
+
+class UNetDecoder(nn.Module):
```
mcgibbon, line 177: The decoder .forward is tightly coupled with the construction of the decoder nn.ModuleList. This presents a bit of a break of the initial guideline, we can't use a hard and fast rule here that "no sub-objects should be initialized inside the init". The solution is either reverting this init change, or something between that and this. In the end, we need the init to take in reasonable-to-construct objects or factories, in a way that the forward call will always work without having to do anything weird or unintuitive in the way it constructs the inputs. In addition to doing this refactor, comment here proposing how we might update the builder pattern pass in the pre-review to properly account for these concerns.

Generally speaking, I'm not too worried about neural network init methods initializing nn.Module's inside themselves, but it is potentially nicer still for it to dependency-inject these concerns. So weigh the options.

fme/ace/models/healpix/healpix_encoder.py
```diff
@@ -130,27 +104,40 @@ def __init__(
...
 
-            self.encoder.append(nn.Sequential(*modules))
+            encoder.append(nn.Sequential(*modules))
+
+        return UNetEncoder(encoder=nn.ModuleList(encoder))
+
 
-        self.encoder = nn.ModuleList(self.encoder)
+class UNetEncoder(nn.Module):
+    """Generic UNetEncoder that can be applied to arbitrary meshes."""
```
mcgibbon: Update the docstring to describe the actual responsibility of this class, since the encoder definition itself is basically separated.

fme/ace/models/healpix/healpix_decoder.py
mcgibbon: Claude: Went with the typed-level option you preferred, and kept the state_dict byte-identical.

**Change** (a612cf31a): introduced `DecoderLevel(nn.Module)` — it bundles the optional upsample + skip-concat + conv and owns that branching in its own `forward(x, skip)`. `UNetDecoder.forward` now just iterates typed levels (`x = level(x, inputs[-1-n])`); it no longer reaches into a string-keyed `ModuleDict` to reconstruct the structure, and `UNetDecoder` is directly constructible from its `(decoder, output_layer)` signature.

**State_dict unchanged (the critical constraint):** `DecoderLevel` names its submodules `upsamp`/`conv` (matching the old `ModuleDict` keys) and registers them in the same order, so keys, order, shapes, dtypes, and values are identical. Verified with a byte-level `state_dict` diff across 36 conv-block × upsample × downsample × padding combinations against the pre-change tip (itself already byte-identical to `main`) — no differences. `test_hpx.py`: 59 passed, 7 skipped (CUDA-only); ruff + mypy clean.

**Encoder:** left as `nn.Sequential`-per-level. A typed `EncoderLevel` with named attributes would rename `encoder.N.0/.1` → `encoder.N.down/.conv` and change the state_dict, and the encoder doesn't need it — its `forward` already applies each level generically. Only its docstring is updated (your other comment).

---

**Proposed refinement to the builder-pattern pass (for `nn.Module`s), for the pre-review checklist** — apply a concrete test rather than a mechanical rule:

> *Is the impl directly constructible from its `__init__` signature, and does its `forward` then always work — without the caller reconstructing a private internal structure?*

- **Composition seams** (a module wiring sub-networks — `HEALPixUNet` composing encoder+decoder, `UNetDecoder` holding per-level modules): inject already-built `nn.Module`s or factories. Impl runs; config wires.
- **Injected collaborators must be self-contained and well-typed:** a built `nn.Module` or a `Callable[[], nn.Module]` factory passes; a bare `nn.ModuleList` of string-keyed `ModuleDict`s whose keys / `None`-slots / ordering the `forward` depends on does **not** — the impl can't be built without replaying private logic. Prefer a small typed per-unit module (like `DecoderLevel`) that carries its branching in its own `forward`.
- **Leaf modules** whose `forward` is inseparable from a sequential, channel-threaded construction loop (no clean per-unit boundary) may keep building their own submodules in `__init__` from configs/factories — this is an accepted exception, not a violation. Forcing DI there buys nothing and yields opaque impls.
- **State_dict/checkpoint equivalence is a hard constraint** on any such refactor: preserve submodule attribute names and registration order, and *verify* it with a byte-level `state_dict` diff across the config matrix — don't assume it.

Happy to fold this into the builder-pattern-pass write-up as a follow-up if it looks right to you.

fme/ace/models/healpix/healpix_encoder.py
mcgibbon: Claude: Done (a612cf31a). Updated the `UNetEncoder` class docstring to describe its actual, reduced responsibility: it runs the per-level modules (each a downsample-then-conv `nn.Sequential`, or just a conv at the shallowest level) built by `UNetEncoderConfig._build`, feeding each level's output into the next and returning the per-level activations as the decoder's skip connections. It builds nothing itself and holds no config; the mesh, padding backend, and channel schedule live in the config.

mcgibbon:
Claude (pre-review): **Verdict — ready for review.** Independent pre-review of the follow-up commit `a612cf31a` (fresh session, no authoring context). No fixes applied; nothing warranted one.

Scope: verified the delta commit that addresses the two review comments, not the whole PR (the base was covered by the prior pre-review). Both comments are resolved: the decoder `forward` no longer reaches into a string-keyed `ModuleDict` (now iterates typed `DecoderLevel`s), and the `UNetEncoder` docstring is updated to its actual run-only responsibility.

**State_dict preservation (the hard constraint) — reproduced, not trusted.** Built HEALPixUNet models under a fixed seed across all **54** combinations (BasicConv / ConvNeXt / MultiSymmetricConvNeXt × transposed-conv & smoothed-interpolate upsampling × AvgPool / MaxPool / Dealiased downsample × karlbauer / isolatitude / earth2grid padding, with `earth2grid` installed so the gated paths actually run) and diffed full `state_dict` signatures — keys, registration order, shapes, dtypes, and a per-tensor SHA-256 of the values — between `a612cf31a` and its parent `734dfe412` (the reviewed builder tip). **Byte-identical in every case, zero differences.** Combined with the prior pre-review having established that parent ≡ `origin/main` byte-identical (comment 4963610115), a `main`-trained HEALPix checkpoint loads unchanged. (I compared against the parent rather than re-building on `main` because `main` predates the builder API; the state_dict equivalence chains transitively.) Confirmed the deepest level (built first, `upsamp is None`) emits only `conv.*` keys while shallower levels emit `upsamp.*` + `conv.*` — the `None` slot contributes nothing, exactly as the old `ModuleDict` did.

**Forward bit-identical.** A representative forward on the isolatitude path (which exercises the nside face-size check) produced an identical output hash and double-precision sum on `a612cf31a` vs parent.

**Design semantics — preserved.** `DecoderLevel.forward(x, skip)` reproduces the old inline loop exactly: when `upsamp is not None` it does `cat([upsamp(x), skip], dim=1)` then `conv`; when `None` (deepest level, fed the bottleneck) it skips the concat and applies `conv` only. `channel_dim=1` moved onto `DecoderLevel` intact. `UNetDecoder.forward`’s `level(x, inputs[-1-n])` passes the same skip tensor the old code indexed. No stale `ModuleDict` / `["upsamp"]` / `["conv"]` references remain in the models, registry, or tests.

**Test — meaningfully stronger.** The updated `test_HEALPixUNet_initialize` asserts each level is a `DecoderLevel`, its `conv` is a module, and `(level.upsamp is None) == (i == 0)` — this would fail if the deepest/shallower split were wrong, unlike a no-op.

**Checks:** `test_hpx.py` = 59 passed, 7 skipped (CUDA-only); `ruff check` + `ruff format --check` clean; `mypy` clean on the changed files (only the known missing-stub env noise, absent in CI).

mcgibbon:
Claude: Addressed both comments (the pair on `healpix_decoder.py` and `healpix_encoder.py` about `list` vs `ModuleList`) in **68679f253**.

`UNetDecoder.__init__` now takes `list[DecoderLevel]` and `UNetEncoder.__init__` takes `list[nn.Sequential]` — typed element info that a bare `nn.ModuleList` erases — and each wraps to `nn.ModuleList` internally for submodule registration. The configs' `_build` pass the plain lists.

state_dict is unchanged: `nn.ModuleList(list)` registers the same submodules in the same order as passing a pre-built `nn.ModuleList`. Re-verified byte-identical (keys, order, shapes, dtypes, values) across the 36-combo config matrix; `test_hpx.py` 59 passed / 7 skipped (CUDA-only); ruff + mypy clean.

(Replying as a PR comment rather than in-thread because the review is still pending/unsubmitted — the API rejects in-thread replies to pending-review comments. Happy to move these into the threads once the review is submitted.)

yikwill:
LGTM!

Arcomano1234 (approved):
Approved

PR #1360, opened 2026-07-15
fme/translate skeleton: component pool, freezing, checkpoint init (translate 1/6)

mcgibbon:
First PR of a six-PR series introducing `fme/translate`, a new package that provides one shared abstraction for two upcoming experiment programs: transfer learning (learned ERA5↔SHiELD transforms around a frozen SHiELD+ forward model) and multi-resolution latent stepping (per-resolution encoders/decoders around a shared coarse-latent backbone). Both reduce to *learned transforms wrapped around a backbone stepper*, differing in what is frozen and which loss terms are active. This PR lays the skeleton the rest of the series builds on; it adds no training, data-loading, objective, or inference logic.

The package sits at the coupled/downscaling tier (it may import `fme.core`, `fme.ace`, and `fme.downscaling`; nothing imports it) and follows the `fme.coupled.stepper` embed-a-stepper template.

The organizing concept is the **domain**: a named space where data or latent state lives — an ordered channel set on one grid. Components declare the domains they map between, and domains are the key that pairs components with data: `ComponentPoolConfig.build` takes one `DatasetInfo` per data-backed domain, keyed by domain name (the data-loading PR derives each entry from the data stream serving that domain). A 1°/2°/4° multi-resolution model is three domains, resolution-changing transforms between adjacent resolutions, and a backbone stepping in the 4° domain; `test_components.py` builds and runs exactly this chain at toy scale.

Changes:
- `fme.translate.DomainConfig` — a domain's channel declaration: an ordered union of plain variable names and `LatentChannels` blocks. A block names a group of free channels once and gives a count (`{name: z, channels: 512}` expands to `z_0`…`z_511`), so latent spaces are never enumerated by hand. A domain with no dataset behind it (a latent space) declares `grid_like: <domain>` and inherits that domain's grid at build.
- `fme.translate.TransformSelector` — a registry of transform module builders whose contract takes both the input and output domains' `DatasetInfo`s (the generic `ModuleConfig` takes one grid and cannot express a resolution-changing operator). Ships with `same_grid` (adapts any existing `ModuleSelector` entry; validates the grids match) and `interpolate` (1x1 conv + interpolation to the output grid — the trivial resize operator and test vehicle; SHT truncation and learned operators are later registry entries).
- `fme.translate.TransformConfig` — a trainable transform/encoder/decoder declaring `in_domain`/`out_domain`; channel counts come from the domains' expanded channel lists. Per-component freezing and name-matched partial checkpoint init are applied to the raw module before `wrap_module`.
- `fme.translate.BackboneConfig` — a backbone ace `Stepper` stepping within one declared `domain` (every stepper variable must be a channel of that domain, checked at build), its config either built fresh (`StepperConfig`) or sourced from a checkpoint (`CheckpointStepperConfig`). Freezing/init are threaded through `StepperConfig.get_stepper(..., parameter_initializer=...)` so they happen inside step construction, before DDP wrapping. When sourcing from a checkpoint, weights flow through the parameter-init path (mirroring `fme.coupled.stepper.CoupledParameterInitConfig`) so a frozen donor is never DDP-wrapped-then-frozen.
- `fme.translate.ComponentPoolConfig` — declares the named domains, transforms, and backbones; validates the wiring (domain references resolve, `grid_like` targets are data-backed, component names unique); dacite-round-trippable via `get_state`/`from_state`.
- `fme.translate.ComponentPool` — the built composite: holds the components and the resolved per-domain `DatasetInfo`s, exposes a flattened `.modules` `nn.ModuleList` (mirroring `CoupledStepper.modules`), fans out `set_train`/`set_eval`/`set_epoch`, and implements `get_state`/`from_state`/`load_state`. It has no forward/predict/objective logic and is not yet `Stepper`-compatible.
- `fme.translate.load_module_weights` — a `WeightsAndHistoryLoader` for the transform init path (loads a saved module `state_dict`).
- `fme/translate/examples/` — full example training configs for the two target programs, written against the intended config surface of the whole series (only `component_pool:` parses in this PR); committed for line-commenting, not runnable.

Freeze-before-wrap is the central correctness requirement: DDP registers gradient hooks at wrap time based on `requires_grad`, so freezing must precede wrapping (the established ordering in `fme/core/step/single_module.py`). A fully-frozen module wrapped fresh becomes a `DummyWrapper` with no live-gradient expectations. Tests cover this on tiny synthetic grids, including a test that a configured-frozen checkpoint-sourced backbone has all parameters frozen, is `DummyWrapper`-wrapped, and carries the donor's weights; plus the multi-resolution chain above, latent-block expansion, domain wiring validation, name-matched (full and partial/excluded) init, state round-trip, train/eval fan-out, and config validation.

Out of scope (later PRs): TranslateBatchData/data loaders binding streams to domains (2); objectives, weighted-sum train stepper, train entrypoint (3); Stepper-compatible export and evaluator (4); cut-point SFNO decomposition (5); latent-consistency objective and noise channels (6).

- [x] Tests added

### DIFF
--- a/fme/translate/components.py
+++ b/fme/translate/components.py
@@ -1,0 +1,392 @@
+"""Named pool of translate components and its built composite container.
+
+The translate abstraction is a *pool of named components* wrapped around a
+backbone stepper. Two kinds of component live in the pool:
+
+- **Transforms** (encoders / decoders / translation maps): trainable
+  ``nn.Module``s built via :class:`ModuleSelector`, each declaring its input
+  and output variable (or latent-channel) name sets. The number of input and
+  output channels is ``len(in_names)`` and ``len(out_names)``.
+- **Backbones**: full ace :class:`Stepper`s (normalization, ocean, corrector,
+  derived variables intact), either built fresh from a :class:`StepperConfig`
+  or sourced from a checkpoint via :class:`CheckpointStepperConfig`.
+
+The built composite (:class:`ComponentPool`) holds the components, exposes a
+flattened ``.modules`` (mirroring
+:class:`fme.coupled.stepper.CoupledStepper`), fans out train/eval/epoch
+control, and implements ``get_state``/``from_state``/``load_state``.
+
+Per-component freezing (:class:`FrozenParameterConfig`) and name-matched
+partial checkpoint initialization (:class:`ParameterInitializationConfig` ->
+``overwrite_weights``) are supported. Freezing *must* happen before DDP
+wrapping, because ``DistributedDataParallel`` registers gradient hooks at wrap
+time based on ``requires_grad``. This ordering is enforced by threading a
+:class:`ParameterInitializer` into ``StepperConfig.get_stepper`` for backbones
+(freeze happens inside step construction, pre-wrap) and by freezing the raw
+module before ``wrap_module`` for transforms.
+"""
+
+import dataclasses
+from typing import Any
+
+import dacite
+import torch
+from torch import nn
+
+from fme.ace.stepper.parameter_init import (
+    ParameterInitializationConfig,
+    StepperWeightsAndHistory,
+    WeightsAndHistoryLoader,
+    null_weights_and_history,
+)
+from fme.ace.stepper.single_module import (
+    CheckpointStepperConfig,
+    Stepper,
+    StepperConfig,
+    load_stepper,
+    load_weights_and_history,
+)
+from fme.core.dataset_info import DatasetInfo
+from fme.core.distributed import Distributed
+from fme.core.registry.module import Module, ModuleSelector
+from fme.core.training_history import TrainingHistory
+
+
+def load_module_weights(path: str | None) -> StepperWeightsAndHistory:
+    """Load a single module's ``state_dict`` for name-matched initialization.
+
+    The file at ``path`` must be a ``torch.save``-d module ``state_dict``
+    (a mapping of parameter name to tensor). Its keys must be a subset of the
+    destination module's ``state_dict`` keys; ``overwrite_weights`` matches by
+    name and, where a destination axis is larger, overwrites its initial slice.
+
+    Returns ``(None, TrainingHistory())`` when ``path`` is None so an unset
+    ``weights_path`` is a no-op.
+    """
+    if path is None:
+        return null_weights_and_history()
+    state_dict = torch.load(path, map_location="cpu", weights_only=False)
+    return [state_dict], TrainingHistory()
+
+
+def _stepper_weights_loader(stepper: Stepper) -> WeightsAndHistoryLoader:
+    """A weights loader that pulls each module's ``state_dict`` from a stepper.
+
+    Mirrors ``fme.coupled.stepper``'s checkpoint-sourced loader: it lets a
+    donor stepper's weights flow through the parameter-init path so that
+    freezing can be threaded through ``get_stepper`` (pre-wrap), rather than
+    freezing after ``load_stepper`` has already DDP-wrapped an unfrozen model.
+    """
+
+    def load(*_: Any) -> StepperWeightsAndHistory:
+        return [module.state_dict() for module in stepper.modules], (
+            stepper.training_history
+        )
+
+    return load
+
+
+@dataclasses.dataclass
+class TransformConfig:
+    """Configuration for a trainable transform / encoder / decoder component.
+
+    Parameters:
+        module: The module builder. Built with ``n_in_channels=len(in_names)``
+            and ``n_out_channels=len(out_names)``.
+        in_names: Input variable (or latent-channel) names.
+        out_names: Output variable (or latent-channel) names.
+        parameter_init: Freezing and name-matched weight initialization for this
+            transform. ``weights_path`` (if set) points to a module
+            ``state_dict`` file loaded by :func:`load_module_weights`; the
+            ``parameters[0].frozen`` config selects which parameters to freeze.
+    """
+
+    module: ModuleSelector
+    in_names: list[str]
+    out_names: list[str]
+    parameter_init: ParameterInitializationConfig = dataclasses.field(
+        default_factory=ParameterInitializationConfig
+    )
+
+    def _build_raw(self, dataset_info: DatasetInfo) -> Module:
+        return self.module.build(
+            n_in_channels=len(self.in_names),
+            n_out_channels=len(self.out_names),
+            dataset_info=dataset_info,
+        )
+
+    def build(self, dataset_info: DatasetInfo) -> Module:
+        """Build the transform, applying weight init and freezing before wrap.
+
+        Order: build the raw module, overwrite a subset of its weights from the
+        configured checkpoint (name-matched), freeze the configured parameters,
+        then DDP-wrap. Freezing precedes wrapping so a fully-frozen module is
+        wrapped as a ``DummyWrapper`` with no live-gradient expectations.
+        """
+        module = self._build_raw(dataset_info)
+        initializer = self.parameter_init.build(load_module_weights)
+        # Both operate on the raw (pre-wrap) module, so state-dict keys carry
+        # no "module." prefix and match an unwrapped donor's names directly.
+        initializer.apply_weights([module.torch_module])
+        initializer.freeze_weights([module.torch_module])
+        return module.wrap_module(Distributed.get_instance().wrap_module)
+
+    def build_for_load(self, dataset_info: DatasetInfo) -> Module:
+        """Build a wrapped transform without weight init or freezing.
+
+        Used when reconstructing from a saved state: the weights are restored by
+        a subsequent ``load_state``, so external initialization is skipped and,
+        as with ``Stepper.from_state``, the reloaded module is left unfrozen.
+        """
+        module = self._build_raw(dataset_info)
+        return module.wrap_module(Distributed.get_instance().wrap_module)
+
+
+@dataclasses.dataclass
+class BackboneConfig:
+    """Configuration for a backbone stepper component.
+
+    The stepper *configuration* is either built fresh (``stepper``) or read from
+    a checkpoint (``checkpoint``); exactly one must be provided. Freezing and
+    name-matched weight initialization are threaded through ``get_stepper`` via
+    ``parameter_init`` so that freezing happens before DDP wrapping.
+
+    When ``checkpoint`` is set and ``init_from_checkpoint`` is True (the default,
+    and the transfer-learning arm's central case of a frozen SHiELD+ donor), the
+    donor's weights are sourced through the parameter-init path: the config comes
+    from ``CheckpointStepperConfig.to_stepper_config()`` and the weights from the
+    loaded stepper's modules. This lets a frozen donor be built pre-wrap, so a
+    fully-frozen backbone is wrapped as a ``DummyWrapper`` rather than a
+    ``DistributedDataParallel`` expecting gradients.
+
+    Set ``init_from_checkpoint`` False to build fresh (random) weights over a
+    checkpoint-sourced config. For a non-frozen inference/fine-tune backbone,
+    ``load_stepper`` (used directly elsewhere) remains the convenience path.
+
+    Parameters:
+        stepper: A fresh stepper configuration. Mutually exclusive with
+            ``checkpoint``.
+        checkpoint: A checkpoint to source the stepper configuration from.
+            Mutually exclusive with ``stepper``.
+        parameter_init: Freezing and (for the fresh-config case) weight
+            initialization. When ``checkpoint`` is set and
+            ``init_from_checkpoint`` is True, ``weights_path`` must be None
+            because the weights come from the checkpoint.
+        init_from_checkpoint: Whether to source weights from ``checkpoint`` via
+            the parameter-init path. Only meaningful when ``checkpoint`` is set.
+    """
+
+    stepper: StepperConfig | None = None
+    checkpoint: CheckpointStepperConfig | None = None
+    parameter_init: ParameterInitializationConfig = dataclasses.field(
+        default_factory=ParameterInitializationConfig
+    )
+    init_from_checkpoint: bool = True
+
+    def __post_init__(self):
+        if (self.stepper is None) == (self.checkpoint is None):
+            raise ValueError(
+                "Exactly one of 'stepper' or 'checkpoint' must be provided "
+                "for a BackboneConfig."
+            )
+        if self.checkpoint is not None and self.init_from_checkpoint:
+            if self.parameter_init.weights_path is not None:
+                raise ValueError(
+                    "When sourcing backbone weights from 'checkpoint', "
+                    "parameter_init.weights_path must be None (the weights come "
+                    "from the checkpoint)."
+                )
+
+    def _config_and_loader(self) -> tuple[StepperConfig, WeightsAndHistoryLoader]:
+        if self.stepper is not None:
+            return self.stepper, load_weights_and_history
+        assert self.checkpoint is not None  # guaranteed by __post_init__
+        config = self.checkpoint.to_stepper_config()
+        if self.init_from_checkpoint:
+            donor = load_stepper(self.checkpoint.checkpoint_path)
+            return config, _stepper_weights_loader(donor)
+        return config, load_weights_and_history
+
+    def build(self, dataset_info: DatasetInfo) -> Stepper:
+        """Build the backbone stepper with freezing applied before DDP wrapping.
+
+        Freezing and weight initialization are carried by the
+        :class:`ParameterInitializer` threaded into ``get_stepper``; the stepper
+        applies the freeze inside step construction (pre-wrap).
+        """
+        config, loader = self._config_and_loader()
+        initializer = self.parameter_init.build(loader)
+        return config.get_stepper(
+            dataset_info=dataset_info,
+            parameter_initializer=initializer,
+        )
+
+
+@dataclasses.dataclass
+class ComponentPoolConfig:
+    """Configuration for a named pool of translate components.
+
+    Parameters:
+        transforms: Named trainable transform / encoder / decoder components.
+        backbones: Named backbone steppers.
+    """
+
+    transforms: dict[str, TransformConfig] = dataclasses.field(default_factory=dict)
+    backbones: dict[str, BackboneConfig] = dataclasses.field(default_factory=dict)
+
+    def __post_init__(self):
+        overlap = set(self.transforms).intersection(self.backbones)
+        if overlap:
+            raise ValueError(
+                f"Component names must be unique across transforms and "
+                f"backbones; found duplicates: {sorted(overlap)}."
+            )
+
+    def build(self, dataset_info: DatasetInfo) -> "ComponentPool":
+        """Build every component against ``dataset_info``.
+
+        Note: a single ``dataset_info`` (grid) is shared by all components in
+        this version; per-component grids for resolution-changing operators are
+        a later extension.
+        """
+        transforms = {
+            name: config.build(dataset_info) for name, config in self.transforms.items()
+        }
+        backbones = {
+            name: config.build(dataset_info) for name, config in self.backbones.items()
+        }
+        return ComponentPool(
+            config=self,
+            transforms=transforms,
+            backbones=backbones,
+            dataset_info=dataset_info,
+        )
+
+    def get_state(self) -> dict[str, Any]:
+        return dataclasses.asdict(self)
+
+    @classmethod
+    def from_state(cls, state: dict[str, Any]) -> "ComponentPoolConfig":
+        return dacite.from_dict(
+            data_class=cls,
+            data=state,
+            config=dacite.Config(
+                strict=True,
+                type_hooks={
+                    StepperConfig: StepperConfig.from_state,
+                },
+            ),
+        )
+
+
+class ComponentPool:
+    """The built composite of a translate component pool.
+
+    Holds the built transforms and backbones, exposes them for training via a
+    flattened ``.modules`` list, fans out train/eval/epoch control, and
+    round-trips its full state (config + weights) via
+    ``get_state``/``from_state``/``load_state``.
+
+    This container is deliberately minimal: it builds, holds, exposes modules,
+    and saves/loads state. It has no forward/predict/objective logic and is not
+    yet ``Stepper``-compatible.
+    """
+
+    def __init__(
+        self,
+        config: ComponentPoolConfig,
+        transforms: dict[str, Module],
+        backbones: dict[str, Stepper],
+        dataset_info: DatasetInfo,
+    ):
+        self._config = config
+        self._transforms = transforms
+        self._backbones = backbones
+        self._dataset_info = dataset_info
+
+    @property
+    def transforms(self) -> dict[str, Module]:
+        return self._transforms
+
+    @property
+    def backbones(self) -> dict[str, Stepper]:
+        return self._backbones
+
+    @property
+    def modules(self) -> nn.ModuleList:
+        """All trainable modules in the pool, as a flattened ``nn.ModuleList``.
+
+        Transforms contribute their wrapped torch module; backbones contribute
+        every module in their own ``.modules``. Ordering is transforms first
+        (in insertion order), then backbones.
+        """
+        result: list[nn.Module] = [
+            transform.torch_module for transform in self._transforms.values()
+        ]
+        for backbone in self._backbones.values():
+            result.extend(backbone.modules)
+        return nn.ModuleList(result)
+
+    def set_train(self) -> None:
+        for transform in self._transforms.values():
+            transform.torch_module.train()
+        for backbone in self._backbones.values():
+            backbone.set_train()
+
+    def set_eval(self) -> None:
+        for transform in self._transforms.values():
+            transform.torch_module.eval()
+        for backbone in self._backbones.values():
+            backbone.set_eval()
+
+    def set_epoch(self, epoch: int) -> None:
+        # Transforms have no epoch-dependent state; only backbones react.
+        for backbone in self._backbones.values():
+            backbone.set_epoch(epoch)
+
+    def get_state(self) -> dict[str, Any]:
+        return {
+            "config": self._config.get_state(),
+            "dataset_info": self._dataset_info.get_state(),
+            "transforms": {
+                name: transform.get_state()
+                for name, transform in self._transforms.items()
+            },
+            "backbones": {
+                name: backbone.get_state() for name, backbone in self._backbones.items()
+            },
+        }
+
+    def load_state(self, state: dict[str, Any]) -> None:
+        for name, transform in self._transforms.items():
+            transform.load_state(state["transforms"][name])
+        for name, backbone in self._backbones.items():
+            backbone.load_state(state["backbones"][name])
+
+    @classmethod
+    def from_state(cls, state: dict[str, Any]) -> "ComponentPool":
+        """Rebuild a pool from saved state.
+
+        Transforms are rebuilt from the pool config against the saved
+        ``dataset_info`` (without external init or freezing) and their weights
+        restored; backbones are rebuilt via the self-contained
+        ``Stepper.from_state`` (its checkpoint path need not still exist).
+        """
+        config = ComponentPoolConfig.from_state(state["config"])
+        dataset_info = DatasetInfo.from_state(state["dataset_info"])
+        transforms = {
+            name: transform_config.build_for_load(dataset_info)
+            for name, transform_config in config.transforms.items()
+        }
+        for name, transform in transforms.items():
+            transform.load_state(state["transforms"][name])
+        backbones = {
+            name: Stepper.from_state(backbone_state)
+            for name, backbone_state in state["backbones"].items()
+        }
+        return cls(
+            config=config,
+            transforms=transforms,
+            backbones=backbones,
+            dataset_info=dataset_info,
+        )
--- a/fme/translate/components.py
+++ b/fme/translate/components.py
@@ -1,15 +1,28 @@
 """Named pool of translate components and its built composite container.
 
-The translate abstraction is a *pool of named components* wrapped around a
-backbone stepper. Two kinds of component live in the pool:
+The translate abstraction is a *pool of named components* mapping between
+*named domains* (see :mod:`fme.translate.domains`), wrapped around a backbone
+stepper. Two kinds of component live in the pool:
 
 - **Transforms** (encoders / decoders / translation maps): trainable
-  ``nn.Module``s built via :class:`ModuleSelector`, each declaring its input
-  and output variable (or latent-channel) name sets. The number of input and
-  output channels is ``len(in_names)`` and ``len(out_names)``.
+  ``nn.Module``s built via the :class:`TransformSelector` registry, each
+  declaring the domain it maps from and the domain it maps to. Channel counts
+  come from the domains' expanded channel lists, and the builder receives both
+  domains' grids, so resolution-changing operators (a 1° -> 2° encoder) are
+  first-class.
 - **Backbones**: full ace :class:`Stepper`s (normalization, ocean, corrector,
   derived variables intact), either built fresh from a :class:`StepperConfig`
-  or sourced from a checkpoint via :class:`CheckpointStepperConfig`.
+  or sourced from a checkpoint via :class:`CheckpointStepperConfig`, stepping
+  within one declared domain.
+
+Domains are how components pair with data: :meth:`ComponentPoolConfig.build`
+takes a mapping of domain name to ``DatasetInfo``, one entry per data-backed
+domain (the data-loading layer provides these, pairing each data stream with
+the domain it serves); a latent domain declares ``grid_like`` and inherits the
+referenced domain's grid. A multi-resolution chain — e.g. 1°/2°/4° domains,
+1°->2° and 2°->4° transforms (and their inverses), and a backbone stepping in
+the 4° domain — is a pool with three domains, four transforms, and one
+backbone.
 
 The built composite (:class:`ComponentPool`) holds the components, exposes a
 flattened ``.modules`` (mirroring
@@ -27,6 +40,7 @@ module before ``wrap_module`` for transforms.
 """
 
 import dataclasses
+from collections.abc import Mapping
 from typing import Any
 
 import dacite
@@ -48,8 +62,11 @@ from fme.ace.stepper.single_module import (
 )
 from fme.core.dataset_info import DatasetInfo
 from fme.core.distributed import Distributed
-from fme.core.registry.module import Module, ModuleSelector
+from fme.core.registry.module import Module
 from fme.core.training_history import TrainingHistory
+
+from .domains import DomainConfig
+from .modules import TransformSelector
 
 
 def load_module_weights(path: str | None) -> StepperWeightsAndHistory:
@@ -91,31 +108,45 @@ class TransformConfig:
     """Configuration for a trainable transform / encoder / decoder component.
 
     Parameters:
-        module: The module builder. Built with ``n_in_channels=len(in_names)``
-            and ``n_out_channels=len(out_names)``.
-        in_names: Input variable (or latent-channel) names.
-        out_names: Output variable (or latent-channel) names.
+        module: The module builder, a :class:`TransformSelector` registry
+            entry. Built with the input and output domains' channel counts and
+            ``DatasetInfo``s (so resolution-changing builders see both grids).
+        in_domain: Name of the domain this transform maps from.
+        out_domain: Name of the domain this transform maps to.
         parameter_init: Freezing and name-matched weight initialization for this
             transform. ``weights_path`` (if set) points to a module
             ``state_dict`` file loaded by :func:`load_module_weights`; the
             ``parameters[0].frozen`` config selects which parameters to freeze.
     """
 
-    module: ModuleSelector
-    in_names: list[str]
-    out_names: list[str]
+    module: TransformSelector
+    in_domain: str
+    out_domain: str
     parameter_init: ParameterInitializationConfig = dataclasses.field(
         default_factory=ParameterInitializationConfig
     )
 
-    def _build_raw(self, dataset_info: DatasetInfo) -> Module:
+    def _build_raw(
+        self,
+        n_in_channels: int,
+        n_out_channels: int,
+        in_dataset_info: DatasetInfo,
+        out_dataset_info: DatasetInfo,
+    ) -> Module:
         return self.module.build(
-            n_in_channels=len(self.in_names),
-            n_out_channels=len(self.out_names),
-            dataset_info=dataset_info,
+            n_in_channels=n_in_channels,
+            n_out_channels=n_out_channels,
+            in_dataset_info=in_dataset_info,
+            out_dataset_info=out_dataset_info,
         )
 
-    def build(self, dataset_info: DatasetInfo) -> Module:
+    def build(
+        self,
+        n_in_channels: int,
+        n_out_channels: int,
+        in_dataset_info: DatasetInfo,
+        out_dataset_info: DatasetInfo,
+    ) -> Module:
         """Build the transform, applying weight init and freezing before wrap.
 
         Order: build the raw module, overwrite a subset of its weights from the
@@ -123,7 +154,9 @@ class TransformConfig:
         then DDP-wrap. Freezing precedes wrapping so a fully-frozen module is
         wrapped as a ``DummyWrapper`` with no live-gradient expectations.
         """
-        module = self._build_raw(dataset_info)
+        module = self._build_raw(
+            n_in_channels, n_out_channels, in_dataset_info, out_dataset_info
+        )
         initializer = self.parameter_init.build(load_module_weights)
         # Both operate on the raw (pre-wrap) module, so state-dict keys carry
         # no "module." prefix and match an unwrapped donor's names directly.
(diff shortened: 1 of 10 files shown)


mcgibbon:
Claude (pre-review): Applied one fix directly to the branch (commit 40b2c9043): corrected the `fme/translate/__init__.py` module docstring from "it imports `fme.core`, `fme.ace`, and `fme.downscaling`" to "it **may** import ...", since `components.py` currently imports only `fme.core` and `fme.ace` (the downscaling import lands with a later PR). Tests (15/15), ruff, ruff-format, and mypy all pass on the changed files after the fix.

fme/translate/components.py
mcgibbon, line 1: The design here looks like it's based from the one we have in ace, and it doesn't look to me like it's going to fit the bill. A few points:
- Configuration for latent channels is awkward, how do we name 512 latent channels? Ideally we'd name a block of latent channels and say how many channels there are, maybe we need a union over something representing variable names and something representing latent channels?
- How would we configure a model that works at 1-degree, 2-degree, and 4-degree, by having transforms from 1 to 2 degree and from 2 to 4-degree, and a 4-degree forward-stepping model? I don't see any representation of domains, and it's unclear how we could possibly pair these components with data.

Revisit the original requirements for this project, and revise the design.

mcgibbon: Claude: Revised the design to make **domains** first-class, addressing both points (pushed in cd116e09a; PR description updated to match).

- **Naming latent channels.** A domain's channels are now an ordered union of plain variable names and named latent-channel *blocks*: `DomainConfig(channels=["a", "b", LatentChannels(name="z", channels=512)])`. A block names the group once and gives a count; it expands to `z_0…z_511` so everything downstream (checkpoint name-matching, diagnostics) sees an ordinary ordered channel list. That's the union-over-{variables, latent-channels} you suggested.

- **Domains + the 1°/2°/4° model + data pairing.** Transforms declare `in_domain`/`out_domain`, backbones declare a `domain`, and `ComponentPoolConfig.build` takes `{domain_name: DatasetInfo}` — one entry per data-backed domain. That mapping *is* the component↔data pairing key (the data-loading PR will derive each entry from the stream serving that domain). Your example is exactly configurable and is now a test (`test_multi_resolution_chain_builds_and_runs`): three domains (`one_deg`/`two_deg`/`four_deg`), 1°→2° and 2°→4° transforms plus their inverses, and a backbone stepping in the `four_deg` domain. A latent domain (no dataset of its own) declares `grid_like: <domain>` and inherits that domain's grid.

- **Resolution-changing operators.** The generic `ModuleSelector.build` contract takes a single `DatasetInfo`, which can't express a resolution change, so transforms now build via a translate-local `TransformSelector` registry whose contract takes *both* the input and output domains' `DatasetInfo`s. `same_grid` adapts any existing `ModuleSelector` entry (SFNO, etc.); `interpolate` (1×1 conv + interpolation) is the trivial resize baseline. SHT truncation / transformer resizers are later registry entries.

Freeze-before-DDP-wrap, checkpoint sourcing, and state round-trip are unchanged in behavior; the frozen-donor transfer setup is still `BackboneConfig(checkpoint=…, init_from_checkpoint=True, parameter_init=<frozen>)`.

mcgibbon:
Claude: Per Jeremy's request, here are full example training configurations for the two target programs, written against the *intended* config surface through the whole PR sequence — only the `component_pool:` block is implemented in this PR; everything else is the shape the follow-on PRs must parse. The goal is to critique these now, before more code exists, and let the critique drive the design. A section at the end maps each block to its PR and lists the design issues that writing these surfaced.

The "Task representation" is the `objectives:` list: enumerated typed configs (translation / forward_prediction / latent_consistency), each naming the components it composes, the data streams it consumes, a loss, and a weight. Training computes all active objectives every optimizer step, backwards each sequentially, and steps once (weighted-sum, one optimizer — per the 2026-07-10 design decision). Validation and checkpoint selection use the same weighted sum.

## Example A — multi-resolution latent stepping (1°/2°/4°, 4° latent backbone)

Per-resolution encoders/decoders into a shared 4° latent, a forward stepper in that latent, forward prediction trained at all three resolutions against one shared backbone, and stochastic latent-consistency constraints between adjacent resolutions.

```yaml
experiment_dir: train_output
save_checkpoint: true
validate_using_ema: true
max_epochs: 40
logging: {log_to_screen: true, log_to_wandb: true, log_to_file: true, project: ace}

train_data:
  batch_size: 16
  num_data_workers: 4
  sampling: paired_by_time    # every stream in a batch is sampled at the same
                              # valid times (pre-coarsened datasets, shared calendar)
  streams:
    era5_1deg: {domain: atmos_1deg, dataset: {data_path: gs://.../era5-1deg-8layer/train}}
    era5_2deg: {domain: atmos_2deg, dataset: {data_path: gs://.../era5-2deg-8layer/train}}
    era5_4deg: {domain: atmos_4deg, dataset: {data_path: gs://.../era5-4deg-8layer/train}}

validation_data:
  batch_size: 16
  sampling: paired_by_time
  streams:
    era5_1deg: {domain: atmos_1deg, dataset: {data_path: gs://.../era5-1deg-8layer/valid, subset: {step: 5}}}
    era5_2deg: {domain: atmos_2deg, dataset: {data_path: gs://.../era5-2deg-8layer/valid, subset: {step: 5}}}
    era5_4deg: {domain: atmos_4deg, dataset: {data_path: gs://.../era5-4deg-8layer/valid, subset: {step: 5}}}

component_pool:
  domains:
    # identical channel sets at each resolution (only the grid differs, and the
    # grid comes from the dataset bound to the domain); YAML anchors avoid
    # writing the list three times. 8-level enumerations elided with "...".
    atmos_1deg: &atmos
      channels:
        - air_temperature_0        # ... _1 through _7
        - specific_total_water_0   # ...
        - eastward_wind_0          # ...
        - northward_wind_0         # ...
        - PRESsfc
        - surface_temperature
        - DSWRFtoa                 # forcings
        - ocean_fraction
        - land_fraction
        - sea_ice_fraction
        - global_mean_co2
    atmos_2deg: *atmos
    atmos_4deg: *atmos
    latent:
      grid_like: atmos_4deg        # latent state lives on the backbone's (4°) grid
      channels:
        - {name: z, channels: 512} # named block -> z_0 ... z_511
        # forcings are *channels of the latent domain*, overwritten from the 4°
        # stream each step (mirrors ace's existing prescribed/forcing overwrite
        # semantics) rather than produced by a separate forcing encoder:
        - DSWRFtoa
        - ocean_fraction
        - land_fraction
        - sea_ice_fraction
        - global_mean_co2

  transforms:
    # resolution-changing, noise-conditioned encoders into the shared latent.
    # ("noise_conditioned_sht_resize" / "sht_resize" are later registry
    # entries; only same_grid and interpolate exist in this PR.)
    encoder_1deg:
      in_domain: atmos_1deg
      out_domain: latent
      module:
        type: noise_conditioned_sht_resize
        config: {embed_dim: 256, num_layers: 4, n_noise_channels: 8}
    encoder_2deg:
      in_domain: atmos_2deg
      out_domain: latent
      module:
        type: noise_conditioned_sht_resize
        config: {embed_dim: 256, num_layers: 4, n_noise_channels: 8}
    encoder_4deg:
      in_domain: atmos_4deg
      out_domain: latent
      module:
        type: noise_conditioned_sht_resize
        config: {embed_dim: 256, num_layers: 4, n_noise_channels: 8}
    decoder_1deg:
      in_domain: latent
      out_domain: atmos_1deg
      module:
        type: sht_resize
        config: {embed_dim: 256, num_layers: 4}
    decoder_2deg:
      in_domain: latent
      out_domain: atmos_2deg
      module:
        type: sht_resize
        config: {embed_dim: 256, num_layers: 4}
    decoder_4deg:
      in_domain: latent
      out_domain: atmos_4deg
      module:
        type: sht_resize
        config: {embed_dim: 256, num_layers: 4}

  backbones:
    stepper_latent:
      domain: latent
      stepper:
        step:
          type: single_module
          config:
            builder:
              type: SphericalFourierNeuralOperatorNet
              config: {embed_dim: 384, num_layers: 8, operator_type: dhconv, scale_factor: 1}
            # "z" is a latent-block reference, expanded against the domain
            # declaration (z -> z_0 ... z_511) by BackboneConfig before the
            # StepperConfig is constructed — the stepper never enumerates
            # latent channels by hand.
            in_names: [z, DSWRFtoa, ocean_fraction, land_fraction, sea_ice_fraction, global_mean_co2]
            out_names: [z]
            normalization:
              network: {identity: true}  # NEW: latent channels have no dataset
                                         # stats; the encoder sets their scale
            ocean: null                  # no physical shell in the latent
            corrector: null

objectives:
  # ---- k=0 explicit reconstruction at each resolution ----
  - name: recon_1deg
    type: translation
    config:
      components: [encoder_1deg, decoder_1deg]
      input_stream: era5_1deg
      target_stream: era5_1deg
      n_ensemble: 2       # encoder is stochastic -> ensemble loss
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 1.0
  - name: recon_2deg
    type: translation
    config:
      components: [encoder_2deg, decoder_2deg]
      input_stream: era5_2deg
      target_stream: era5_2deg
      n_ensemble: 2
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 1.0
  - name: recon_4deg
    type: translation
    config:
      components: [encoder_4deg, decoder_4deg]
      input_stream: era5_4deg
      target_stream: era5_4deg
      n_ensemble: 2
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 1.0

  # ---- forward prediction at every resolution, one shared backbone ----
  - name: forward_4deg
    type: forward_prediction
    config:
      encoder: encoder_4deg
      backbone: stepper_latent
      decoder: decoder_4deg
      input_stream: era5_4deg
      target_stream: era5_4deg
      forcing_stream: era5_4deg   # forcings enter on the latent (4°) grid;
                                  # paired_by_time sampling time-aligns them
      n_forward_steps:
        start_value: 1
        milestones:
          - epoch: 8
            value: {outcomes: [{steps: 1, probability: 0.6}, {steps: 4, probability: 0.4}]}
      n_ensemble: 2
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 1.0
  - name: forward_2deg
    type: forward_prediction
    config:
      encoder: encoder_2deg
      backbone: stepper_latent
      decoder: decoder_2deg
      input_stream: era5_2deg
      target_stream: era5_2deg
      forcing_stream: era5_4deg   # note: cross-resolution — see issue 4 below
      n_forward_steps:
        start_value: 1
        milestones:
          - epoch: 8
            value: {outcomes: [{steps: 1, probability: 0.6}, {steps: 4, probability: 0.4}]}
      n_ensemble: 2
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 0.5
  - name: forward_1deg
    type: forward_prediction
    config:
      encoder: encoder_1deg
      backbone: stepper_latent
      decoder: decoder_1deg
      input_stream: era5_1deg
      target_stream: era5_1deg
      forcing_stream: era5_4deg
      n_forward_steps:
        start_value: 1
        milestones:
          - epoch: 8
            value: {outcomes: [{steps: 1, probability: 0.6}, {steps: 4, probability: 0.4}]}
      n_ensemble: 2
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 0.5

  # ---- stochastic latent consistency between adjacent resolutions:
  # noise-injected coarse encoding pulled toward the stop-gradient fine
  # encoding (stop-gradient is semantics of the objective type, not config) ----
  - name: consistency_1v2
    type: latent_consistency
    config:
      fine_encoder: encoder_1deg
      coarse_encoder: encoder_2deg
      fine_stream: era5_1deg
      coarse_stream: era5_2deg
      n_ensemble: 4
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 0.5
  - name: consistency_2v4
    type: latent_consistency
    config:
      fine_encoder: encoder_2deg
      coarse_encoder: encoder_4deg
      fine_stream: era5_2deg
      coarse_stream: era5_4deg
      n_ensemble: 4
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 0.5

optimization:
  optimizer_type: AdamW
  lr: 1.0e-4
  enable_automatic_mixed_precision: false

# in-training inference callbacks over exported composites: any encode/decode
# pairing around the shared backbone, evaluated with the standard aggregators
inference:
  - composite: {encoder: encoder_4deg, backbone: stepper_latent, decoder: decoder_4deg}
    n_forward_steps: 1460
    forward_steps_in_memory: 50
    loader: {dataset: {data_path: gs://.../era5-4deg-8layer/valid}, start_indices: {first: 0, n_initial_conditions: 4, interval: 365}}
  - composite: {encoder: encoder_1deg, backbone: stepper_latent, decoder: decoder_1deg}
    n_forward_steps: 1460
    forward_steps_in_memory: 50
    loader: {dataset: {data_path: gs://.../era5-1deg-8layer/valid}, start_indices: {first: 0, n_initial_conditions: 4, interval: 365}}
```

## Example B — transfer learning (frozen C96-SHiELD donor, ERA5↔SHiELD translators)

Translators operate directly on real fields (no latent domain), wrapped around a fully-frozen SHiELD+ donor stepper loaded from checkpoint. Cycle consistency in both directions. Both datasets at 4°, so the translators are same-grid (a resolution-changing registry entry drops in if the ERA5 side were 1°).

```yaml
experiment_dir: train_output
save_checkpoint: true
validate_using_ema: true
max_epochs: 20
logging: {log_to_screen: true, log_to_wandb: true, log_to_file: true, project: ace}

train_data:
  batch_size: 16
  num_data_workers: 4
  sampling: independent   # ERA5 and SHiELD are different realities; no
                          # objective consumes both streams in one term
  streams:
    era5:   {domain: era5,       dataset: {data_path: gs://.../era5-4deg-daily/train}}
    shield: {domain: shield_c96, dataset: {data_path: gs://.../shield-c96-4deg-daily/train}}

validation_data:
  batch_size: 16
  sampling: independent
  streams:
    era5:   {domain: era5,       dataset: {data_path: gs://.../era5-4deg-daily/valid}}
    shield: {domain: shield_c96, dataset: {data_path: gs://.../shield-c96-4deg-daily/valid}}

component_pool:
  domains:
    era5:
      channels:              # full list elided; the two variable sets differ
        - air_temperature_0  # ... _7
        - specific_total_water_0   # ...
        - eastward_wind_0    # ...
        - northward_wind_0   # ...
        - PRESsfc
        - surface_temperature
        - DSWRFtoa
        - ocean_fraction
        - land_fraction
        - sea_ice_fraction
        - global_mean_co2
    shield_c96:
      channels:              # the donor's full in/out union, incl. channels
        - air_temperature_0  # ERA5 lacks — the translators map between the
        - specific_total_water_0   # two sets; they need not match
        - eastward_wind_0    # ...
        - northward_wind_0   # ...
        - PRESsfc
        - surface_temperature
        - LHTFLsfc
        - SHTFLsfc
        - PRATEsfc
        - DSWRFtoa
        - ocean_fraction
        - land_fraction
        - sea_ice_fraction
        - global_mean_co2

  transforms:
    to_shield:
      in_domain: era5
      out_domain: shield_c96
      module:
        type: same_grid      # both domains on the same 4° grid
        config:
          module:
            type: SphericalFourierNeuralOperatorNet
            config: {embed_dim: 256, num_layers: 4, scale_factor: 1}
    to_era5:
      in_domain: shield_c96
      out_domain: era5
      module:
        type: same_grid
        config:
          module:
            type: SphericalFourierNeuralOperatorNet
            config: {embed_dim: 256, num_layers: 4, scale_factor: 1}

  backbones:
    donor:
      domain: shield_c96
      checkpoint:
        checkpoint_path: gs://.../t1-modern-pretrain/best_inference_ckpt.tar
      init_from_checkpoint: true   # weights sourced from the checkpoint...
      parameter_init:
        parameters:
          - frozen: {exclude: []}  # ...and every parameter frozen (empty
                                   # exclude list = freeze all), applied
                                   # before DDP wrapping

objectives:
  # end-to-end rollout scored on ERA5: translate in -> step frozen donor ->
  # translate out. Forcings (incl. SST for the donor's ocean config, CO2) are
  # translated per step through to_shield so they arrive in-donor-domain —
  # the same path the +4K / future-scenario perturbation harnesses exercise.
  - name: forward_era5
    type: forward_prediction
    config:
      encoder: to_shield
      backbone: donor
      decoder: to_era5
      input_stream: era5
      target_stream: era5
      forcing_stream: era5
      n_forward_steps:
        start_value: 1
        milestones:
          - epoch: 4
            value: {outcomes: [{steps: 1, probability: 0.5}, {steps: 4, probability: 0.5}]}
      n_ensemble: 2   # donor is the v2 noise-conditioned SFNO
      loss: {type: EnsembleLoss, kwargs: {crps_weight: 1.0, energy_score_weight: 1.0}}
      weight: 1.0

  # cycle consistency: translating to the other domain and back is identity
  - name: cycle_era5
    type: translation
    config:
      components: [to_shield, to_era5]
      input_stream: era5
      target_stream: era5
      loss: {type: MSE}
      weight: 0.2
  - name: cycle_shield
    type: translation
    config:
      components: [to_era5, to_shield]
      input_stream: shield
      target_stream: shield
      loss: {type: MSE}
      weight: 0.2

optimization:
  optimizer_type: AdamW
  lr: 1.0e-4
  enable_automatic_mixed_precision: false

inference:
  - composite: {encoder: to_shield, backbone: donor, decoder: to_era5}
    n_forward_steps: 1460
    forward_steps_in_memory: 50
    loader: {dataset: {data_path: gs://.../era5-4deg-daily/valid}, start_indices: {first: 0, n_initial_conditions: 4, interval: 365}}
```

## What lands where

| Config block | PR |
|---|---|
| `component_pool:` (domains / transforms / backbones, freeze, checkpoint init) | **1 (this PR)** — modulo the block-name expansion in backbone `in_names`/`out_names` and identity normalization, which are new (issues 1–2) |
| `train_data:` / `validation_data:` streams, `sampling:` | 2 |
| `objectives:` (`translation`, `forward_prediction`), `optimization:`, weighted-sum trainer, per-objective metric namespacing | 3 |
| `inference:` composites (Stepper-compatible export) | 4 |
| `latent_consistency` objective type, noise-conditioned encoder registry entries | 6 |

PR 5 (SFNO cut-point decomposition) doesn't appear in either config: it would show up as `parameter_init.weights_path` warm-starts on encoder/backbone/decoder, and as a bare-processor backbone in the latent-splice transfer variant.

## Issues writing these surfaced (starters for the critique)

1. **Backbone name lists vs latent blocks.** `StepperConfig.in_names`/`out_names` are flat lists, so a 512-channel latent reintroduces the enumeration problem *inside* the backbone config. Proposal shown: bare block names (`z`) in a backbone's name lists are expanded against its domain declaration by `BackboneConfig` before the `StepperConfig` is constructed — no `fme.ace` changes. Needs a collision rule (block name shadowing a variable name).
2. **Latent normalization.** `NormalizationConfig` requires stats files; latent channels have none (the encoder sets their scale, per the design decision). Needs an identity-normalization option (`network: {identity: true}` above is invented).
3. **How forcings enter a latent backbone.** Shown: forcing variables are *channels of the latent domain*, overwritten from the forcing stream each step (mirrors ace's prescribed/forcing overwrite semantics); encoders technically output those channels and the overwrite discards them. The alternative — a forcing encoder producing a dedicated latent block — requires transforms to target a channel *subset* of a domain, which the current domain algebra can't express. This wants an explicit decision.
4. **Cross-resolution `forcing_stream`.** `forward_1deg` pulls forcings from the 4° stream (the latent grid), which silently requires paired-by-time sampling across streams. The coupling between an objective's stream references and the loader's `sampling:` mode is currently implicit; misconfiguration (independent sampling + cross-stream forcing) would be garbage-in with no error.
5. **`grid_like` requires a data-backed domain.** A latent grid with no dataset at that resolution can't be declared (fine here — 4° data exists; a gap if we ever want a latent-only resolution). Explicit grid spec on a domain is the alternative.
6. **Chained vs direct multi-hop transforms.** `translation` takes a component *list*, but `forward_prediction` takes a single encoder/decoder — a chained 1°→2°→4° encode isn't expressible there. Direct per-resolution encoders are used here because `latent_consistency`'s fine/coarse signature wants single-hop encoders into the same latent; if hop-chaining (parameter sharing across resolutions) is wanted, `forward_prediction`'s encoder/decoder should become lists too.
7. **Domains carry no grid identity.** A domain is just a channel list; its grid comes from whichever dataset the loader binds to it. Binding the 2° dataset to `atmos_1deg` is silent — nothing in the config detects it. A nominal grid declaration (e.g. expected `img_shape`) per data-backed domain would make stream↔domain misbinding fail loudly.
8. **Loss-scale coupling in the weighted sum.** Latent-consistency CRPS is in raw latent units while physical-objective CRPS is in normalized units, so objective `weight:`s double as scale normalizers and checkpoint selection inherits the sensitivity. Per-objective loss normalization (or documented weight-tuning guidance) may be needed.

mcgibbon:
Does your multi-resolution configuration example take the 1-degree latent space and upscale it to 2-degree (with let's say module A) and then take the 2-degree latent and upscale it to 4-degree with module B, where if we were to encode the 2-degree data its latent would also use module B to go to 4-degree?

mcgibbon:
Let's put these example configurations into the repo for easier line-commenting.

mcgibbon:
Claude: No — as written it doesn't chain. Each resolution has its own direct encoder into the shared 4° latent (`encoder_1deg: atmos_1deg → latent`, `encoder_2deg: atmos_2deg → latent`, `encoder_4deg: atmos_4deg → latent`), so no module is shared between the 1° and 2° encode paths: 2° data encodes through `encoder_2deg` alone, and nothing in the 1° path reuses it. The latent-consistency objectives pull the two encodings toward each other statistically, but share no parameters. This is issue 6 in the design-issues list — I used direct encoders because `forward_prediction` and `latent_consistency` take a *single* encoder, so a chained encode isn't expressible in the objectives as drafted.

The scheme you describe — shared upscalers, where 2° data's latent rides the same module B as the 1° path — is already expressible at the component-pool level (per-resolution latent domains, transforms between them):

```yaml
domains:
  latent_1deg: {grid_like: atmos_1deg, channels: [{name: z, channels: 512}, <forcings>]}
  latent_2deg: {grid_like: atmos_2deg, channels: ...}
  latent_4deg: {grid_like: atmos_4deg, channels: ...}   # backbone domain
transforms:
  encoder_1deg: {in_domain: atmos_1deg, out_domain: latent_1deg, ...}
  encoder_2deg: {in_domain: atmos_2deg, out_domain: latent_2deg, ...}
  up_1to2:     {in_domain: latent_1deg, out_domain: latent_2deg, ...}   # module A
  up_2to4:     {in_domain: latent_2deg, out_domain: latent_4deg, ...}   # module B
```

What's missing is the objective side: `encoder:`/`decoder:` need to become component *chains* (as `translation.components` already is):

```yaml
- name: forward_1deg
  type: forward_prediction
  config:
    encoder: [encoder_1deg, up_1to2, up_2to4]
    backbone: stepper_latent
    decoder: [down_4to2, down_2to1, decoder_1deg]
    ...
- name: forward_2deg
  config:
    encoder: [encoder_2deg, up_2to4]        # shares module B with the 1° path
    ...
```

`latent_consistency` also gets cleaner under chains: 1°-vs-2° consistency compares `[encoder_1deg, up_1to2]` against `[encoder_2deg]` *in `latent_2deg`* — per-hop, in the intermediate latent — instead of both paths only meeting in the 4° latent.

I'd make chain-valued `encoder:`/`decoder:` the config contract regardless: a direct encoder is a length-1 chain, so it subsumes both schemes and settles issue 6. Whether the example should *train* the chained/shared-B scheme or the current independent-encoders scheme is the real design decision. My read: chained is the better default — module B gets gradient from every finer-resolution path and cross-resolution consistency becomes structural rather than purely statistical — at the cost of a deeper 1° encode and per-resolution intermediate latents to define. If you confirm the chained scheme, I'll revise the in-repo example (`fme/translate/examples/multi-resolution-latent.yaml`) to it.

mcgibbon:
Claude: Done in 161d5e9b9 — [`fme/translate/examples/`](https://github.com/ai2cm/ace/blob/161d5e9b9/fme/translate/examples/README.md) now holds both configs (`multi-resolution-latent.yaml`, `transfer-learning.yaml`), transcribed verbatim from the comment above, plus a README carrying the status caveat (intended surface, only `component_pool:` parses in this PR, not runnable) and the what-lands-where table.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+  streams:
+    era5_1deg: {domain: atmos_1deg, dataset: {data_path: gs://.../era5-1deg-8layer/train}}
+    era5_2deg: {domain: atmos_2deg, dataset: {data_path: gs://.../era5-2deg-8layer/train}}
+    era5_4deg: {domain: atmos_4deg, dataset: {data_path: gs://.../era5-4deg-8layer/train}}
+
+validation_data:
+  batch_size: 16
+  sampling: paired_by_time
+  streams:
+    era5_1deg: {domain: atmos_1deg, dataset: {data_path: gs://.../era5-1deg-8layer/valid, subset: {step: 5}}}
```
mcgibbon: It's not obvious what the meaning of the key is in this mapping (whether it's referenceble and if so by what name), can we make this a list of mapping where era5_1deg (if needed at all) is the value of a key (name? what's the difference between name and domain?).

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+experiment_dir: train_output
+save_checkpoint: true
+validate_using_ema: true
+max_epochs: 40
+logging: {log_to_screen: true, log_to_wandb: true, log_to_file: true, project: ace}
+
+train_data:
+  batch_size: 16
+  num_data_workers: 4
+  sampling: paired_by_time    # every stream in a batch is sampled at the same
```
mcgibbon: Are there other options? Should we allow other options at all?

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+        - PRESsfc
+        - surface_temperature
+        - DSWRFtoa                 # forcings
+        - ocean_fraction
+        - land_fraction
+        - sea_ice_fraction
+        - global_mean_co2
+    atmos_2deg: *atmos
+    atmos_4deg: *atmos
+    latent:
```
mcgibbon: We need latent_1deg, latent_2deg, and latent_4deg.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+      grid_like: atmos_4deg        # latent state lives on the backbone's (4°) grid
+      channels:
```
mcgibbon: Maybe we can allow channels to be [int, list[str]]? To avoid the awkward name: z, channels: 512.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+        - {name: z, channels: 512} # named block -> z_0 ... z_511
+        # forcings are *channels of the latent domain*, overwritten from the 4°
+        # stream each step (mirrors ace's existing prescribed/forcing overwrite
+        # semantics) rather than produced by a separate forcing encoder:
+        - DSWRFtoa
+        - ocean_fraction
+        - land_fraction
+        - sea_ice_fraction
+        - global_mean_co2
```
mcgibbon: No, these should be encoded by the 1-degree encoder into latent_1deg and then translated to latent_2deg and latent_4deg.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+
+  transforms:
+    # resolution-changing, noise-conditioned encoders into the shared latent.
+    # ("noise_conditioned_sht_resize" / "sht_resize" are later registry
+    # entries; only same_grid and interpolate exist in PR 1.)
+    encoder_1deg:
+      in_domain: atmos_1deg
+      out_domain: latent
```
mcgibbon: These should each go to latent_ndeg, and then we want upsampler and downsampler modules that go between resolutions.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+        config: {embed_dim: 256, num_layers: 4, n_noise_channels: 8}
+    encoder_4deg:
+      in_domain: atmos_4deg
+      out_domain: latent
+      module:
+        type: noise_conditioned_sht_resize
+        config: {embed_dim: 256, num_layers: 4, n_noise_channels: 8}
+    decoder_1deg:
+      in_domain: latent
+      out_domain: atmos_1deg
```
mcgibbon: Issue: while the encoder_1deg and decoder_1deg are both in the atmos_1deg _domain_, we'd like the input fields for the encoder to be able to differ from the output fields of the decoder. That means the channel lists belong to these transforms, not to the domains. Domains are a data loading concept primarily. Perhaps the channel list of the domains could be automatically derived from the channels referenced by the transforms.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+                                         # stats; the encoder sets their scale
+            ocean: null                  # no physical shell in the latent
+            corrector: null
+
+objectives:
+  # ---- k=0 explicit reconstruction at each resolution ----
+  - name: recon_1deg
+    type: translation
+    config:
+      components: [encoder_1deg, decoder_1deg]
```
mcgibbon, line 192: We should be able to keep this objective even if the inputs and outputs are different channels, FYI.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,249 @@
...
+
+optimization:
+  optimizer_type: AdamW
+  lr: 1.0e-4
+  enable_automatic_mixed_precision: false
+
+# in-training inference callbacks over exported composites: any encode/decode
+# pairing around the shared backbone, evaluated with the standard aggregators
+inference:
+  - composite: {encoder: encoder_4deg, backbone: stepper_latent, decoder: decoder_4deg}
```
mcgibbon: I think we'll need a list of components, not a mapping of names, especially when you update to separate the latent spaces.

fme/translate/examples/transfer-learning.yaml
```diff
@@ -0,0 +1,146 @@
...
+  # end-to-end rollout scored on ERA5: translate in -> step frozen donor ->
+  # translate out. Forcings (incl. SST for the donor's ocean config, CO2) are
+  # translated per step through to_shield so they arrive in-donor-domain —
+  # the same path the +4K / future-scenario perturbation harnesses exercise.
+  - name: forward_era5
+    type: forward_prediction
+    config:
+      encoder: to_shield
+      backbone: donor
+      decoder: to_era5
```
mcgibbon: Again we probably want a list of components, not an explicit encoder/backbone/decoder.

mcgibbon:
Claude: Revised both example configs per your review in 207bacb75. This comment originally carried the point-by-point responses because the review was still pending (in-thread replies were blocked); now that it's submitted, each response lives in its thread — including two design consequences flagged for your eye, in the latent-resampler thread (where the resamplers get gradient) and the channels-on-transforms thread (latent domains still self-declare their width).

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — streams are now a [list of `{name?, domain, dataset}`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L15-L28). `name` is the handle objectives reference and defaults to `domain`; it exists because a domain is a channel space several streams could share (e.g. ERA5 and IFS both feeding `atmos_1deg`), while `era5_1deg` names the source. This example keeps explicit names; the transfer-learning example omits them to show the default.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Removed the knob entirely in 207bacb75: streams that co-occur in some objective are sampled at the same valid times, streams that never co-occur are sampled independently. That derives `paired_by_time` for this example (the consistency and forcing_stream terms pair everything transitively) and `independent` for the transfer example, with nothing to configure or get wrong.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — [`latent_1deg`/`latent_2deg`/`latent_4deg`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L45-L69), each `grid_like` its resolution's atmos domain, with learned resamplers between them (see the encoder/decoder threads below).

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Adopted in 207bacb75 — `channels` is now `int | list` (and a list may mix an int entry with names): [`channels: 512`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L50-L52) for the pure latents. An int declares an anonymous block named after its domain, so the backbone writes `in_names: [latent_4deg, DSWRFtoa, ...]`. `latent_4deg` uses the [mixed-list form](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L56-L69) because it also carries the forcing channels the overwrite fills.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — each encoder is now same-grid into its own resolution's latent ([encoders](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L71-L107)), and the resolution change happens in shared latent-to-latent resamplers, so the 1° and 2° encode paths into the 4° latent share `latent_2to4` — the module sharing your earlier question described.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — decoders come out of their own resolution's latent, and [four learned `sht_resize` resamplers](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L140-L172) (`latent_1to2`, `latent_2to4`, `latent_4to2`, `latent_2to1`) move between latent resolutions; the decode paths share `latent_4to2`. One consequence worth your eye: with the fine side of `latent_consistency` behind the stop-gradient, the downsamplers train only through the forward objectives' encode paths and the upsamplers only through their decode paths. If that's too weak, a latent cycle objective (`translation` with `components: [latent_1to2, latent_2to1]` scored in latent space) is expressible in this surface — but it needs a latent-domain "target stream", which streams-of-datasets can't provide. Flagging rather than solving.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — encoders declare [`in_channels`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L78-L90) (prognostics + forcings), decoders declare [`out_channels`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L115-L121) (prognostics only, deliberately different), and the data domains are [empty declarations](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L37-L44) whose load lists are derived from the transforms referencing them. One wrinkle: the latent domains still declare their own channel width (nothing loads them, and every transform touching a latent must agree on its width anyway). The alternative — encoder declares `out_channels: 512` — puts the same number on many transforms with a consistency check. I preferred the domain declaration; easy to flip.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Kept in 207bacb75, with the semantics [stated on the objective](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L192-L195): the loss compares the chain's output channels (the decoder's `out_channels`) against the target stream, so encoder-only inputs (forcings) simply aren't reconstructed and differing in/out channel sets are fine.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Done in 207bacb75 — `forward_prediction` and `inference` composites take a [flat chain with exactly one backbone](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L224-L264): components before it encode once at the initial time, the backbone rolls out in its own domain, components after it decode every output step. `latent_consistency` takes [`fine_components`/`coarse_components` chains](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/multi-resolution-latent.yaml#L286-L299) that must land in the same latent domain (e.g. fine `[encoder_1deg, latent_1to2]` vs coarse `[encoder_2deg]` in `latent_2deg`).

fme/translate/examples/transfer-learning.yaml
mcgibbon: Claude: Done in 207bacb75 — [`components: [to_shield, donor, to_era5]`](https://github.com/ai2cm/ace/blob/207bacb75063602539d6ff4c248cfb96ee2d5cc7/fme/translate/examples/transfer-learning.yaml#L111) for the forward objective and the same chain for the inference composite, with the one-backbone split semantics as in the multi-resolution example.

fme/translate/examples/multi-resolution-latent.yaml
```diff
@@ -0,0 +1,320 @@
...
+      channels:
+        - 512
+        # forcing channels of the backbone's domain, overwritten from the 4°
+        # stream each step (mirrors ace's existing prescribed/forcing
+        # overwrite semantics) rather than produced by any transform:
+        - DSWRFtoa
+        - ocean_fraction
+        - land_fraction
+        - sea_ice_fraction
+        - global_mean_co2
```
mcgibbon: Why do we need this? We should be encoding and translating the n-degree version.
```suggestion
```

mcgibbon: Claude: Done in 5c5cd06b4 — and you're right that your earlier comment ("these should be encoded by the 1-degree encoder into latent_1deg and then translated...") was anchored on these forcing lines; I mis-read it as being about the encoder blocks, so it went unapplied last round. Now: [all three latent domains are pure `channels: 512`](https://github.com/ai2cm/ace/blob/5c5cd06b444ed5916a5636b5c5d09c7edb5b4a02/fme/translate/examples/multi-resolution-latent.yaml#L45-L60), the backbone's [`in_names` is just `[latent_4deg]`](https://github.com/ai2cm/ace/blob/5c5cd06b444ed5916a5636b5c5d09c7edb5b4a02/fme/translate/examples/multi-resolution-latent.yaml#L176-L177), and the forward objectives' forcing streams are [same-resolution](https://github.com/ai2cm/ace/blob/5c5cd06b444ed5916a5636b5c5d09c7edb5b4a02/fme/translate/examples/multi-resolution-latent.yaml#L261) (the n-degree version, carried into the latent through the objective's encode chain — which also removes the old cross-resolution forcing_stream wrinkle).

One mechanism question this raises: at rollout steps beyond the initial time, the encode chain must inject forcings *without* data prognostics (feeding target-time prognostics into the encoder would leak the answer into the rollout). So the encoder needs a forcing-only application — masked/absent prognostic inputs on the same module, or a small dedicated forcing-encoder transform per resolution sharing the latent resamplers. I noted the constraint [on the objectives](https://github.com/ai2cm/ace/blob/5c5cd06b444ed5916a5636b5c5d09c7edb5b4a02/fme/translate/examples/multi-resolution-latent.yaml#L217-L224) without picking a mechanism.

fme/translate/examples/multi-resolution-latent.yaml
mcgibbon: Claude: Correction — this comment was anchored on the latent forcing channels, not the encoder blocks, and I mis-scoped my reply above to the encoders. The forcing channels are now removed from the latent domain in 5c5cd06b4; see [the new thread](https://github.com/ai2cm/ace/pull/1360#discussion_r3669813763) for the applied change and one mechanism question it raises.

PR #1362, opened 2026-07-15
extend per channel loss to validation

Arcomano1234:
Validation currently only reports scalar loss and per-variable RMSE, so there's no direct per-variable counterpart to training's `train/mean/loss/<var>` for spotting which channels drive the loss. This seems to have been an oversight in the previous per-channel PR as the stepper already populates `TrainOutput.per_channel_losses `during validation (loss-schedule steps run with optimize=True regardless of the null optimizer), this PR simply has the validation aggregator consume that field, reusing the same `PerChannelLossAggregator` as training so validation emits directly comparable `val/mean/loss/<var>` metrics.

Changes:
- `fme.ace.aggregator.one_step.OneStepAggregator` now records batch.per_channel_losses and logs `<label>/mean/loss/<var>` in get_summary, mirroring TrainAggregator.
- `fme.ace.aggregator.one_step.OneStepAggregatorConfig` and` build_one_step_aggregator `gain a `per_channel_loss: bool = True` flag (matching `TrainAggregatorConfig.per_channel_loss`) to toggle the behavior.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/one_step/main.py
+++ b/fme/ace/aggregator/one_step/main.py
@@ -5,7 +5,10 @@ from collections.abc import Sequence
 
 import torch
 
-from fme.ace.aggregator.loss_metrics import PerStepLossAggregator
+from fme.ace.aggregator.loss_metrics import (
+    PerChannelLossAggregator,
+    PerStepLossAggregator,
+)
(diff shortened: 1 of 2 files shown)


brianhenn (approved):
LGTM thanks

PR #1363, opened 2026-07-16
Make coupled-dataset path handling work for cloud and local stores

jpdunc23:
The coupled-dataset creation script assumed local filesystem paths: existence checks on output zarrs and stats directories, and the output writer's overwrite guard, used `os.path`, which silently misbehaves on `gs://` URLs. The writer's guard was therefore inert on cloud paths, overwriting an existing store rather than raising. This meant multi-terabyte production stores had to be mirrored to a local volume before coupled processing could run. Path-existence decisions now resolve the filesystem from the URL scheme via a shared fsspec helper, so cloud and local stores share one code path; the staged compute-or-read-existing behavior, debug mode, and subsample mode are unchanged on local paths.

Changes:
- `scripts/data_process/fs_utils.py` (new): `path_exists`, `is_dir`, `makedirs`, `is_local` helpers resolving the filesystem from the URL scheme via `fsspec.core.url_to_fs`, dropping the path's cached listing so a stale miss cannot defeat an overwrite guard
- `scripts/data_process/create_coupled_datasets.py`: all path-existence checks (output stores, stats dirs, merged-stats dirs, ensemble category dirs) go through `path_exists`
- `scripts/data_process/writer_utils.py`: `OutputWriterConfig.write`'s overwrite/initialize guard uses the scheme-aware `is_dir`, accepting `gs://` output stores
- `scripts/data_process/get_stats.py`, `scripts/data_process/merge_stats.py`: hardcoded `gs:` prefix checks and local-only `os.makedirs` replaced with the shared helpers
- New end-to-end test running `CoupledDatasetsConfig.write_datasets_and_stats` on tiny synthetic zarrs: staged outputs and merged stats produced, re-runs resume on existing outputs without recomputing, debug mode writes nothing

- [x] Tests added


scripts/data_process/fs_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def is_dir(path: str) -> bool:
+    """Return whether the path is a directory (or object-store prefix)."""
+    fs, fs_path = _fs_and_path(path)
+    return fs.isdir(fs_path)
+
+
+def makedirs(path: str) -> None:
+    """Create a directory if it doesn't exist; a no-op on object stores."""
```
elynnwu: [nit] technically gcsfs attempts bucket creation if it doesn't exist which can raise a permissions error rather than no-op

scripts/data_process/fs_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def _fs_and_path(path: str) -> tuple[fsspec.AbstractFileSystem, str]:
+    return fsspec.core.url_to_fs(path)
+
+
+def path_exists(path: str) -> bool:
+    """Return whether a file, directory, or object-store prefix exists."""
+    fs, fs_path = _fs_and_path(path)
+    return fs.exists(fs_path)
```
elynnwu, line 28: Do you think we will want to make this utils more accessible to other use? If so, consider adding `fs.invalidate_cache()` after `_fs_and_path`, same for `is_dir`. 

`url_to_fs` returns a cached, long-lived instance whose directory listing cache is not invalidated by writes made through other instances (e.g. zarr v3 writes via a separate asynchronous instance). This does not affect the current workflow since every check happens either before the corresponding write or after writes made through the same sync instance.

elynnwu (approved):

scripts/data_process/fs_utils.py
jpdunc23: Good catch — docstring corrected in a4b56addd.

Left the behavior as a plain delegation rather than short-circuiting on non-local paths: a `gs://` output directory whose bucket doesn't exist is a config error, and surfacing the permissions error here beats a silent no-op followed by an opaque failure later inside the zarr write. `makedirs(..., exist_ok=True)` never removes anything, so there's no clobber risk either way.

— 🤖 Drafted by Claude (Opus 5) via Claude Code, posted on @jpdunc23's behalf.

scripts/data_process/fs_utils.py
jpdunc23: Added it in a2d661b22, in `_fs_and_path` rather than in the two helpers individually, so all of them get it and a future helper can't silently skip it.

You're right that it doesn't affect the current workflow, but I think it earns its place beyond hygiene: `OutputWriterConfig.write` uses `is_dir` to decide whether to raise before `initialize_store`, so a stale negative there wouldn't surface as a confusing error, it would overwrite an existing store. The call sites are all coarse control-flow checks rather than anything in a loop, so the extra listing costs nothing measurable.

Two limits worth naming: this only fixes staleness in the calling process's own cache — it does nothing for two concurrent runs, where both see "not there" and both initialize. Nothing here guards that today and I'd rather not widen this PR to take it on.

On the broader question — keeping the module script-local for now, since it's imported as a `scripts/data_process/` sibling and nothing outside that directory can reach it. Happy to promote it into `fme` in its own PR when something needs it.

— 🤖 Drafted by Claude (Opus 5) via Claude Code, posted on @jpdunc23's behalf.

PR #1364, opened 2026-07-16
Ocean-sourced sea ice for coupled dataset creation

jpdunc23:
Some production ocean stores have no separate 6-hourly sea ice dataset: sea ice fields live in the ocean store at 5-daily snapshot instants, including a natively-computed ocean-relative `ocean_sea_ice_fraction`. This PR lets `compute_coupled_sea_ice` take sea surface fraction and sea ice concentration from the ocean store, selecting the source by priority: a separate sea ice dataset when configured; else the ocean store's sea ice fields when no `window_avg` is configured (the gate keeps existing no-sea-ice-dataset windowed configs, like the E3SM coupled configs, on the legacy atmosphere-sourced path unchanged); else the atmosphere's own sea ice fraction. A new `CoupledSeaIceConfig.use_atmosphere_sea_ice_fraction_fallback` flag (default False) makes reaching the atmosphere source a loud error by default, so a misconfigured production run — a missing ocean sea ice variable or an accidentally-configured window average — can never silently define the coupled surface from the atmosphere field. On the ocean-sourced path the existing coherence logic (modified sea surface fraction where the ocean is wet, concentration clipped to [0, 1], sea ice and ocean fractions recomposed) is applied at 5-daily resolution and the fractions are forward-filled onto the atmosphere's 6-hourly index — no window average, which is unnecessary at the ocean cadence. Within the ocean source, the native `ocean_sea_ice_fraction` is preferred over re-deriving the clipped ratio. `hfds_total_area` is passed through from the input ocean store instead of being re-derived when it is already present.

Changes:
- `coupled_dataset_utils.compute_coupled_sea_ice` selects the sea-ice source by priority (sea ice dataset > ocean store when un-windowed > atmosphere) and gains the ocean-sourced 5-daily path
- `coupled_dataset_utils.CoupledSeaIceConfig` gains `use_atmosphere_sea_ice_fraction_fallback` (default False); the E3SM coupled config sets it to true explicitly to stay on the legacy atmosphere-sourced path
- `coupled_dataset_utils.OceanInputFieldsConfig` gains `sea_ice_fraction_name` and `ocean_sea_ice_fraction_name`
- `coupled_dataset_utils.compute_coupled_ocean` passes `hfds_total_area` through when the input ocean store already provides it
- `coupled_dataset_utils.compute_coupled_ocean` no longer includes the atmosphere surface temperature in the coupled ocean dataset; it was previously retained by mistake when `include_ts` was configured

- [x] Tests added


scripts/data_process/coupled_dataset_utils.py
```diff
@@ -406,12 +415,21 @@ class CoupledSeaIceConfig:
             to the output dataset. If window_avg is also configured, then the window-
             average is applied to ocean grid cells only based on the forumula
             (1 - ofrac_window_avg) * ts + ofrac_window_avg * ts_window_avg.
+        use_atmosphere_sea_ice_fraction_fallback: If true (default), the
```
elynnwu: We talked about just removing this option. On second thought, I think we keep it for now to do a release so the paper can cite and have data processing reproducible. We can open another PR removing it later. Maybe it's better to default to False and set it to True in E3SM manually?

elynnwu (approved):

PR #1365, opened 2026-07-16
Point CM4 coupled configs at the OM4 GCS production stores

jpdunc23:
Updates `CM4-piControl-200yr-coupled.yaml` and `CM4-1pctCO2-140yr-coupled.yaml` in place to read the `2026-07-15` 1-degree production stores on GCS and write coupled outputs and stats to `gs://vcm-ml-intermediate`, with sea ice sourced from the ocean store's native 5-daily fields: no separate sea ice input, no sea-ice window average, and `use_atmosphere_sea_ice_fraction_fallback: false`.

Time chunk sizes align with the stores' shards (atmosphere 360, ocean 365). The surface-flux window timestamps are right-labeled at the ocean stores' interval-end instants (the previous stores used midpoint-style labels), and the 1pctCO2 timestamps follow that store's model years (0001–0141). The 1pctCO2 config's precomputed sea ice mask points at the piControl coupled ocean output, which must exist before the 1pctCO2 run.

Changes:
- `configs/CM4-piControl-200yr-coupled.yaml`, `configs/CM4-1pctCO2-140yr-coupled.yaml`: repointed at the GCS production stores with ocean-sourced sea ice; no compatibility with their previous `/climate-default` contents

- [ ] Tests added


jpdunc23:
Stacked on #1364 (ocean-sourced sea ice), which is itself stacked on #1363 (cloud path handling); the base is set to `scripts/coupled-ocean-sourced-sea-ice` so the diff shows only the config changes. Will retarget to `main` as those land. Debug-mode validation against the GCS stores passed; the subsample smoke run is still pending.

scripts/data_process/configs/CM4-1pctCO2-140yr-coupled.yaml
```diff
@@ -28,29 +31,26 @@ coupled_datasets:
...
   coupled_sea_ice:
-    window_avg:
-      window_timedelta: 120h
-      first_timestamp: "0211-01-06T00:00:00"
-      shift_timestamps_to_avg_interval_midpoint: true
+    # No window_avg and no sea_ice input dataset: the sea ice fraction is
+    # sourced from the ocean store's native 5-daily fields and forward-filled
+    # onto the atmosphere's 6-hourly time index.
     include_ts: true
+    use_atmosphere_sea_ice_fraction_fallback: false
```
elynnwu, line 38: [nit] the default is False already, can skip this line (plus we might remove this config option later)

scripts/data_process/configs/CM4-piControl-200yr-coupled.yaml
```diff
@@ -27,29 +30,26 @@ coupled_datasets:
...
   coupled_sea_ice:
-    window_avg:
-      window_timedelta: 120h
-      first_timestamp: "0151-01-06T00:00:00"
-      shift_timestamps_to_avg_interval_midpoint: true
+    # No window_avg and no sea_ice input dataset: the sea ice fraction is
+    # sourced from the ocean store's native 5-daily fields and forward-filled
+    # onto the atmosphere's 6-hourly time index.
     include_ts: true
+    use_atmosphere_sea_ice_fraction_fallback: false
```
elynnwu, line 37: same as above

elynnwu (approved):

PR #1366, opened 2026-07-16
Add two-track (global/local latent) SFNO step

mcgibbon:
Adds a two-track (global / local latent) SFNO as a new step type. The
spherical harmonic (spectral) operator acts on only a *global* subset of the
latent channels; the remaining *local* channels see only pointwise operations,
so their horizontal gradients are never distorted by the network's spatial
mixing. This is intended as a more physical prior for out-of-sample climates
and lets column-local variable-transformation bases (e.g. RH-like or
convective-basis inputs) participate without the SHT distorting between-column
differences.

The global track reuses the single-track conditional SFNO's encoder, decoder,
and per-block submodule names and widths, so with zero local channels and all
options off the module tree and parameter names are byte-for-byte identical to
`SphericalFourierNeuralOperatorNet` — an existing single-track checkpoint loads
into the two-track network with no remap.

Changes:
- `fme.core.models.conditional_sfno.latents.Latents`: wraps a concatenated
  `[global | local]` latent, exposing `.global_channels` / `.local_channels` /
  `.all` tensors and per-track `__add__`, hiding the split index from callers.
- `fme.core.models.conditional_sfno.two_track_sfnonet`: `TwoTrackSFNONetConfig`,
  `TwoTrackFourierNeuralOperatorBlock`, and
  `TwoTrackSphericalFourierNeuralOperatorNet` (+ `get_lat_lon_two_track_sfnonet`).
  Per block, the pointwise parts (conditional norm, activation, MLP,
  inner/outer skips) act on the full concatenation; the spectral filter reads
  and writes only the global slice, with the local filter-stage output zero and
  carried by the inner-skip residual over the full width. Three default-off
  options: `feed_global_to_local` (option 1), `parallel_conv1x1` (option 2),
  `per_track_layer_norm` (option 3).
- `fme.ace.registry.two_track_sfno.TwoTrackSFNOBuilder` (registered
  `TwoTrackSFNO`): builds the noise-conditioned two-track network via
  `build_two_track`, which takes the four per-track channel counts; the
  single-tensor `build` raises since the split is not known there. Defaults:
  `embed_dim=256`, `local_embed_dim=256`, `noise_embed_dim=32`. `spectral_ratio`
  (default 1.0) is sized against the *global* width (the filter's width), so it
  can be scaled up to keep the actual filter (channels-per-group x groups) equal
  to a single-track baseline's when the global track is narrowed.
- `fme.ace.step.two_track.TwoTrackStep` / `TwoTrackStepConfig` (registered
  `two_track`): explicit `global_in_names` / `local_in_names` /
  `global_out_names` / `local_out_names`; post-init raises if a variable
  appears on both tracks on the same side. Inputs are packed
  `global + local` into one tensor (the network splits at
  `len(global_in_names)`); outputs unpack `global + local`. Orchestration
  (normalization / corrector / ocean / residual prediction) reuses
  `step_with_adjustments`, mirroring `SingleModuleStep`.
- `fme/ace/step/example_configs/two_track_baseline.yaml`: documented baseline
  step config (options off, some variables on the local track), exercised by a
  CPU smoke train-step test.

Out of scope for this first pass (not added): global-mean removal,
`include_channel_mask_inputs`, secondary decoder, input dropout, LoRA,
local (DISCO) blocks, and non-linear filters. Only `filter_type='linear'`
is supported.

L=0 *output* equivalence (the two-track forward reproducing the single-track
forward byte-for-byte at zero local width) holds only when the spectral filter
returns no round-trip residual, because the two-track block takes its
inner/outer skip from the full normed input rather than the filter's returned
residual. Two things force a round trip and so break output equivalence:
`filter_residual=True`, and `data_grid='equiangular'` (which gives the first
and last blocks a forward/inverse transform on different grids). The baseline
uses the default `data_grid='legendre-gauss'` (all transforms share a grid),
where equivalence holds and is covered by the regression test. Because the
module tree and parameter names are identical regardless of grid, an
`equiangular` single-track checkpoint would load without reproducing its
output — so `TwoTrackSFNOBuilder.__post_init__` rejects `local_embed_dim=0`
with `data_grid='equiangular'`, failing loudly on the backwards-incompatibility
rather than loading a wrong-output checkpoint. `local_embed_dim` must also be
nonzero if and only if the local track carries input or output channels
(validated in `build_two_track`): the practical config always has a nonzero
local track, and `local_embed_dim=0` is reserved for warm-starting from a
single-track checkpoint on `legendre-gauss`.

- [x] Tests added

### DIFF
--- a/fme/ace/registry/two_track_sfno.py
+++ b/fme/ace/registry/two_track_sfno.py
@@ -1,0 +1,167 @@
+import dataclasses
+from typing import Literal
+
+import torch
+
+from fme.ace.registry.registry import ModuleConfig, ModuleSelector
+from fme.ace.registry.stochastic_sfno import NoiseConditionedModel
+from fme.core.dataset_info import DatasetInfo
+from fme.core.models.conditional_sfno.layers import ContextConfig
+from fme.core.models.conditional_sfno.two_track_sfnonet import (
+    TwoTrackSFNONetConfig,
+    get_lat_lon_two_track_sfnonet,
+)
+
+
+@ModuleSelector.register("TwoTrackSFNO")
+@dataclasses.dataclass
+class TwoTrackSFNOBuilder(ModuleConfig):
+    """Builder for a two-track (global / local latent) noise-conditioned SFNO.
+
+    The network splits its latent into a global track (which passes through the
+    spherical harmonic transform) and a local, pointwise-only track. Unlike the
+    single-track builders, the per-track *channel* counts are not known until
+    the owning two-track step assigns variables to tracks, so this builder is
+    driven by ``build_two_track`` (called directly by the step) rather than the
+    standard ``build``. It is still registered in the module registry so the
+    config loads from YAML like any other builder.
+
+    Attributes:
+        embed_dim: Total latent width (global + local).
+        local_embed_dim: Width of the local latent track. 0 recovers the
+            single-track network exactly.
+        noise_embed_dim: Dimension of the noise conditioning channels. 0
+            disables noise conditioning.
+        noise_type: "gaussian" or "isotropic" noise.
+        context_pos_embed_dim: Dimension of the context positional embedding.
+        label_embed_dim: Dimension of a learned label embedding (0 = one-hot).
+        data_grid: Grid type for the spherical harmonic transforms.
+        feed_global_to_local: OPTION 1 (default off).
+        parallel_conv1x1: OPTION 2 (default off).
+        per_track_layer_norm: OPTION 3 (default off).
+
+    The remaining attributes mirror the single-track SFNO config.
+    """
+
+    embed_dim: int = 256
+    local_embed_dim: int = 0
+    filter_type: Literal["linear"] = "linear"
+    noise_embed_dim: int = 256
+    noise_type: Literal["isotropic", "gaussian"] = "gaussian"
+    context_pos_embed_dim: int = 0
+    label_embed_dim: int = 0
+    global_layer_norm: bool = False
+    num_layers: int = 12
+    use_mlp: bool = True
+    mlp_ratio: float = 2.0
+    activation_function: str = "gelu"
+    encoder_layers: int = 1
+    pos_embed: bool = True
+    big_skip: bool = True
+    checkpointing: int = 0
+    data_grid: Literal["legendre-gauss", "equiangular"] = "legendre-gauss"
+    hard_thresholding_fraction: float = 1.0
+    filter_residual: bool = False
+    filter_output: bool = False
+    normalize_big_skip: bool = False
+    affine_norms: bool = False
+    filter_num_groups: int = 1
+    drop_rate: float = 0.0
+    drop_path_rate: float = 0.0
+    feed_global_to_local: bool = False
+    parallel_conv1x1: bool = False
+    per_track_layer_norm: bool = False
+
+    def __post_init__(self):
+        if self.context_pos_embed_dim > 0 and self.pos_embed:
+            raise ValueError(
+                "context_pos_embed_dim and pos_embed should not both be set"
+            )
+        # Validate the network config (embed_dim / local_embed_dim / groups).
+        self._net_config()
+
+    def _net_config(self) -> TwoTrackSFNONetConfig:
+        return TwoTrackSFNONetConfig(
+            embed_dim=self.embed_dim,
+            local_embed_dim=self.local_embed_dim,
+            filter_type=self.filter_type,
+            global_layer_norm=self.global_layer_norm,
+            num_layers=self.num_layers,
+            use_mlp=self.use_mlp,
+            mlp_ratio=self.mlp_ratio,
+            activation_function=self.activation_function,
+            encoder_layers=self.encoder_layers,
+            pos_embed=self.pos_embed,
+            drop_rate=self.drop_rate,
+            drop_path_rate=self.drop_path_rate,
+            hard_thresholding_fraction=self.hard_thresholding_fraction,
+            big_skip=self.big_skip,
+            checkpointing=self.checkpointing,
+            filter_num_groups=self.filter_num_groups,
+            filter_residual=self.filter_residual,
+            filter_output=self.filter_output,
+            normalize_big_skip=self.normalize_big_skip,
+            affine_norms=self.affine_norms,
+            feed_global_to_local=self.feed_global_to_local,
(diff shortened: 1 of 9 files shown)


mcgibbon:
Pre-review agent: independent pre-review (no authoring context). Verdict below; one item worth a decision before merge, plus a doc fix I applied.

## Verdict: ready for review, with one design question flagged

The change is well-structured and does what the description says. The spectral filter is sized to and reads/writes only the global slice; local channels never enter the SHT (filter is constructed with `global_channels`, local filter-stage output is a zero pad carried by the inner skip). The both-sides validation, the global-then-local pack/unpack order, and the three default-off options are all wired and tested as described. Tests are meaningful (option 1 uses a forward hook to probe exactly the edge it adds; the L=0 regression loads `state_dict(strict=True)` and compares outputs at `atol=0`). `import fme`, the two touched test files (27 tests), ruff, ruff-format, and mypy are all green on the branch.

## Flagged for your decision: L=0 output equivalence is grid-conditional

The headline claim — a single-track checkpoint loads into the two-track net and reproduces its output byte-for-byte at zero local width — holds for the default `data_grid='legendre-gauss'` but **silently breaks for `data_grid='equiangular'`**. Verified empirically (embed_dim=16, 2 layers, local=0):

```
grid=legendre-gauss  max|diff|=0.000e+00   (equivalent)
grid=equiangular     max|diff|=4.089e-01   (NOT equivalent)
```

Mechanism: the two-track block always takes its inner/outer skip from the full normed input `x_norm`. The single-track block takes the skip from the spectral filter's *returned residual*, which equals `x_norm` only when `SpectralConvS2._round_trip_residual` is false. That flag is true not only under `filter_residual=True` but also whenever a block's forward/inverse transforms are on different grids — which is exactly the first and last blocks when `data_grid='equiangular'` (`trans_down`/`itrans_up` are equiangular, `trans`/`itrans` are legendre-gauss). Middle blocks are unaffected.

So the regression test (which uses `legendre-gauss`) is a genuine equivalence, not tautological — but it does not cover the equiangular regime, where an existing checkpoint loads without error (module tree + names are identical) yet does not reproduce its output. The single-track noise-conditioned SFNO you'd warm-start from defaults to `legendre-gauss`, so the intended path is fine; the risk is a silent wrong-output trap for an equiangular warm-start.

I did **not** change the numerical behavior — the block's use of `x_norm` for the skip is a deliberate, documented design choice (and switching the global skip to the round-tripped residual would change equiangular global-track behavior at local>0, which is your call, not a mechanical fix). Decision for you: (a) accept + keep the limitation documented, (b) make the global skip grid-independent so the invariant holds for all grids, or (c) forbid equiangular for the warm-start path.

## Applied fix (documentation only, pushed to the branch)

Commit c57d267bf: corrected the module and block docstrings and the PR description, which previously attributed non-equivalence solely to `filter_residual=True`, to also name the `equiangular` round-trip; and added a comment to the regression test pinning its coverage boundary (why `legendre-gauss`, and that `equiangular` diverges). No behavior change.

fme/ace/registry/two_track_sfno.py
```diff
@@ -0,0 +1,167 @@
...
+        parallel_conv1x1: OPTION 2 (default off).
+        per_track_layer_norm: OPTION 3 (default off).
+
+    The remaining attributes mirror the single-track SFNO config.
+    """
+
+    embed_dim: int = 256
+    local_embed_dim: int = 0
+    filter_type: Literal["linear"] = "linear"
+    noise_embed_dim: int = 256
```
mcgibbon: Let's set default 32.

fme/ace/registry/two_track_sfno.py
```diff
@@ -0,0 +1,167 @@
...
+        data_grid: Grid type for the spherical harmonic transforms.
+        feed_global_to_local: OPTION 1 (default off).
+        parallel_conv1x1: OPTION 2 (default off).
+        per_track_layer_norm: OPTION 3 (default off).
+
+    The remaining attributes mirror the single-track SFNO config.
+    """
+
+    embed_dim: int = 256
+    local_embed_dim: int = 0
```
mcgibbon: Let's set default 256 for local as well, while we want to technically support zero (and no local inputs) in practice this will always be nonzero (otherwise we'd just use the existing step). Post-init validation should ensure local embed dim is zero if and only if the local input and output channels are both zero (if either are non-zero it must be nonzero).

fme/ace/registry/two_track_sfno.py
mcgibbon: Claude: Done in 2e9cd5b75 — noise_embed_dim default is now 32.

fme/ace/registry/two_track_sfno.py
mcgibbon: Claude: Done in 2e9cd5b75 — local_embed_dim default is now 256. build_two_track now enforces the iff: local_embed_dim must be > 0 when the local track has any input or output channels, and must be 0 when it has none (with the error pointing at the single-track step for the no-local case).

mcgibbon:
Claude: Implemented your equiangular decision (option c — fail loudly) in 2e9cd5b75, plus the two inline requests.

- `TwoTrackSFNOBuilder.__post_init__` now raises when `local_embed_dim=0` and `data_grid='equiangular'`, with an error naming the backwards-incompatibility (single-track checkpoint loads but doesn't reproduce its output on that grid). legendre-gauss (the warm-start path) is unaffected.
- `local_embed_dim` default → 256, `noise_embed_dim` default → 32.
- `build_two_track` enforces local_embed_dim > 0 iff the local track carries channels (points at the single-track step when there's no local track).

Added 5 tests (equiangular guard raises; legendre-gauss L=0 allowed; both iff directions). Full two-track suite (31 tests across the two files) plus ruff, ruff-format, and mypy are green on the branch. PR description updated to describe the guard as a hard error.

mcgibbon:
Claude: Added a `spectral_ratio` passthrough in 04905bbf6 (was in the first-pass out-of-scope list). The two-track spectral filter is sized to the *global* width, so a narrowed global latent would otherwise shrink the actual filter. `spectral_ratio` (default 1.0, preserving the L=0 equivalence) is now on the config/builder and validated against the global width — so per the plan, when the global track is halved (e.g. 512→256) you double spectral_ratio to keep the filter (channels-per-group × groups) equal to the single-track baseline's. +3 tests (builds/runs at ratio<1; validated against global width; builder threads it); full two-track suite now 34, ruff/format/mypy green.

PR #1367, opened 2026-07-16
Thread StepperState through CoupledStepper

elynnwu:
`CoupledStepper` drops each component's `StepperState` (noise-generator random state and corrector state) at every coupled-step boundary: per-step initial conditions are rebuilt without it, the SST prescription discards it from the atmosphere state, and the terminal `StepOutput`s are built with `stepper_state=None`. As a result, stochastic modules reseed at each coupled step, correctors re-capture their reference values, and coupled restart files carry no embedded state, so a run resumed from a restart is not bitwise-identical to an uninterrupted run. This PR threads the per-component stepper states through the coupled rollout and into the restart serialization added in #1341.

Coupled runs do not currently carry a seeded random state, so existing stochastic runs are unaffected. But a random state attached to the initial condition (e.g. restored from a restart) now persists through the rollout instead of being silently dropped. 

One behavioral change for existing runs is for checkpoints with dry-air conservation enabled: the corrector's reference dry-air mass was previously re-captured at every coupled-step boundary. Since each step's output is corrected to exactly match the reference, this caused no physical drift, but it made coupled runs inconsistent with uncoupled ACE and would have prevented segmented runs from being bitwise-identical to unsegmented ones. The reference is now captured once from the run's initial condition and held, matching the uncoupled behavior.

Changes:
- `fme.coupled.stepper.ComponentStepPrediction` carries a `stepper_state`
- `fme.coupled.stepper.CoupledStepper.get_prediction_generator` threads each component's terminal stepper state into the next coupled step's initial conditions; `_prescribe_ic_sst` preserves the incoming atmosphere stepper state; `_process_prediction_generator_list` attaches the terminal states to the component `StepOutput`'s, so `BatchData` serialization embeds them in `ocean/restart.nc` and `atmosphere/restart.nc`

- [x] Tests added

### DIFF
--- a/fme/coupled/stepper.py
+++ b/fme/coupled/stepper.py
@@ -46,6 +46,7 @@ from fme.core.ocean import OceanConfig
 from fme.core.ocean_data import OCEAN_FIELD_NAME_PREFIXES, OceanData
 from fme.core.optimization import NullOptimization
 from fme.core.step.output import StepOutput
+from fme.core.stepper_state import StepperState
 from fme.core.tensors import add_ensemble_dim, unfold_ensemble_dim
 from fme.core.timing import GlobalTimer
 from fme.core.training_history import TrainingHistory, TrainingJob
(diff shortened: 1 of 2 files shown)


fme/coupled/stepper.py
```diff
@@ -772,10 +773,12 @@ def __init__(
         realm: Literal["ocean", "atmosphere"],
         data: TensorDict,
         step: int,
+        stepper_state: StepperState | None = None,
```
AnnaKwa: Are there any remaining usages of this class not in this diff where the stepper state is left as None? If it's always going to be threaded through can this just be a required arg?

elynnwu: Updated to not default to None, but it can actually be none (e.g., random seed not set)

AnnaKwa (approved):

PR #1369, opened 2026-07-16
Add whole-run zarr outputs across inference segments

elynnwu:
Segmented inference writes each segment's outputs to its own directory, which fragments long-run outputs (e.g. monthly means) into per-segment files. This PR makes zarr outputs configured through `DataWriterConfig.files` accumulate into a single whole-run store shared by all segments, written to any fsspec-compatible location (including directly to a remote bucket), so a preempted-and-resumed run produces the same single stores as an uninterrupted one. netCDF outputs are unaffected and `MonthlyDataWriter` / `save_monthly_files` are untouched.

Block-mean `time_coarsen` needs no special handling across segments: it is stateless per batch and its validation requires the per-segment step count to be divisible by the coarsen factor, so no coarsening block can straddle a segment boundary; whole-run store offsets are simply expressed in coarsened units.

Changes:
- `fme.ace.inference.data_writer.zarr.ZarrWriterAdapter` and `SeparateICZarrWriterAdapter` take a `start_timestep`: the store is sized for the whole run, each segment region-writes its slice of the time axis, and resumed segments open the existing store in append mode; only the run's first segment may create the store, since its initial condition times determine the store's time coordinates
- `fme.ace.inference.data_writer.monthly_zarr.MonthlyZarrWriter`, built from a `files` entry with zarr format and `monthly_mean` time coarsening (previously rejected), with the month axis pre-allocated from the whole run's length; because monthly accumulation is not idempotent, `finalize` writes a netCDF snapshot to the segment directory (before the restart files, so restarts remain the completion marker) and the next segment's writer restores from it, making segment re-runs exact
- `fme.ace.inference.data_writer.zarr.ZarrWriterConfig` gains an optional `path` for placing stores outside the experiment directory
- `fme.ace.inference.data_writer.segment.SegmentContext` threaded at runtime from the coupled segmented driver through `DataWriterConfig` / `FileWriterConfig` build paths; no user-facing config changes and non-segmented runs are unchanged

- [x] Tests added

### DIFF
--- a/fme/ace/inference/data_writer/file_writer.py
+++ b/fme/ace/inference/data_writer/file_writer.py
@@ -17,7 +17,9 @@ from fme.core.typing_ import Slice
 
 from .dataset_metadata import DatasetMetadata
 from .monthly import MonthlyDataWriter
+from .monthly_zarr import MONTHLY_SNAPSHOT_DIR, MonthlyZarrWriter
 from .raw import NetCDFWriterConfig, RawDataWriter
+from .segment import SegmentContext
 from .time_coarsen import (
     MonthlyCoarsenConfig,
     PairedTimeCoarsen,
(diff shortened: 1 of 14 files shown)


elynnwu:
The idea for this is that we can write the monthly zarr to gcs like this:
```yaml
data_writer:
  ocean:
    save_prediction_files: false
    save_monthly_files: false
    files:
      - label: output               # -> gs://bucket/output.zarr
        save_reference: false       # predictions only; no "_predictions" suffix
        time_coarsen:
          method: monthly_mean
        format:
          name: zarr
          path: gs://bucket         # store dir; {label}.zarr is created inside
        # names: [sst, o_prog]      # optional: subset of variables to save
  atmosphere:
    save_prediction_files: false
    save_monthly_files: false
    files:
      - label: output               # -> gs://bucket/atmosphere/output.zarr
        save_reference: false
        time_coarsen:
          method: monthly_mean
        format:
          name: zarr
          path: gs://bucket/atmosphere
```

PR #1372, opened 2026-07-17
  Fix wandb run naming in segmented inference

### LINKED ISSUE #471: Segmented runs do not have desired wandb names for each segment

brianhenn:
Segmented inference (`python -m fme.ace.inference <cfg> --segments N`) runs N segments sequentially in one process. It was meant to give each segment its own wandb run named `<base>-segment_NNNN`, but instead every segment logged to a single shared run: later segments restarted their step counter at 0 and clobbered earlier segments' metrics, leaving only the last segment's values. Two wandb behaviors combined to cause this: repeated `wandb.init()` in a process returns the already-active run by default (`reinit="return_previous"` in scripts), and `WANDB_NAME` is snapshotted only on the first init, so mutating it per segment was ignored. This fixes both so each segment gets its own correctly-named run.

Changes:
- `fme.core.wandb.WandB.finish`: ends the active run so the next `init` starts a new run rather than reusing it.
- `fme.core.logging_utils.LoggingConfig._configure_wandb`: passes `name=os.environ.get("WANDB_NAME")` to `wandb.init` explicitly. This is the shared init used by all entrypoints, but it is a no-op for every single-`init`-per-process path (training, evaluation, single inference): the explicit value is exactly what wandb already reads from `WANDB_NAME` on that first init, so run names are unchanged. It only has an effect when `WANDB_NAME` changes mid-process after the first init — i.e. the segmented case fixed here.
- `fme.ace.inference.run_segmented_inference`: finishes each segment's run, and configures top-level logging with wandb disabled so no run lingers before the first segment; the pre-loop `configure_logging` in `main` is removed.
- `fme.core.testing.MockWandB`: adds `finish()`, records started runs, and models wandb reading `WANDB_NAME` only once, so tests exercise the real per-init naming behavior.

Resumption is unaffected. Inference does not use wandb resumption (`resumable=False`), so a segment re-run after a preemption simply starts a fresh run with the same name — it never resumes an existing run. Training's resume path does a single init where the explicit `name` equals the value wandb would have read from the environment anyway, so a resumed run keeps its existing name exactly as before.

- [x] Tests added

Resolves #471

### DIFF
--- a/fme/ace/inference/inference.py
+++ b/fme/ace/inference/inference.py
@@ -47,6 +47,7 @@ from fme.core.generics.inference import get_record_to_wandb, run_inference
 from fme.core.labels import BatchLabels
 from fme.core.logging_utils import LoggingConfig
 from fme.core.timing import GlobalTimer
+from fme.core.wandb import WandB
 
 from .evaluator import resolve_variable_metadata
 
@@ -348,7 +349,6 @@ def main(
             with GlobalTimer():
                 return run_inference_from_config(config)
         else:
-            config.configure_logging(log_filename="inference_out.log")
             run_segmented_inference(config, segments)
 
 
@@ -486,6 +486,14 @@ def run_segmented_inference(config: InferenceConfig, segments: int):
             "cannot re-broadcast it consistently. Run with n_ensemble_per_ic=1, "
             "or run a single non-segmented inference for ensemble runs."
         )
+    # Configure top-level logging without a wandb run; each segment owns its run.
+    top_level_logging = dataclasses.replace(config.logging, log_to_wandb=False)
+    top_level_logging.configure_logging(
+        config.experiment_dir,
+        "inference_out.log",
+        config=dataclasses.asdict(config),
+        resumable=False,
+    )
     logging.info(
         f"Starting segmented inference with {segments} segments. "
         f"Saving to {config.experiment_dir}."
@@ -505,6 +513,8 @@ def run_segmented_inference(config: InferenceConfig, segments: int):
                 os.environ["WANDB_NAME"] = f"{original_wandb_name}-{segment_label}"
             with GlobalTimer():
                 run_inference_from_config(config_copy)
+            # Finish this segment's run so the next segment starts a fresh one.
+            WandB.get_instance().finish()
         config_copy.initial_condition = InitialConditionConfig(
             path=restart_path, engine="netcdf4"
         )
--- a/fme/core/logging_utils.py
+++ b/fme/core/logging_utils.py
@@ -153,6 +153,9 @@ class LoggingConfig:
         notes = _get_wandb_notes(_get_beaker_id())
         wandb.init(
             config=config_copy,
+            # wandb reads WANDB_NAME only on the first init in a process, so pass
+            # it explicitly for later runs (e.g. segments) to be named correctly.
+            name=os.environ.get("WANDB_NAME"),
             project=self.project,
             entity=self.entity,
             experiment_dir=experiment_dir,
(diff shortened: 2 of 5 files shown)


fme/ace/inference/inference.py
```diff
@@ -486,6 +486,14 @@ def run_segmented_inference(config: InferenceConfig, segments: int):
             "cannot re-broadcast it consistently. Run with n_ensemble_per_ic=1, "
             "or run a single non-segmented inference for ensemble runs."
         )
+    # Configure top-level logging without a wandb run; each segment owns its run.
+    top_level_logging = dataclasses.replace(config.logging, log_to_wandb=False)
```
brianhenn, line 490: Pre-review agent: this block inlines the body of `InferenceConfig.configure_logging` plus a manual `dataclasses.asdict(config)`. It could reuse the existing method instead, which removes the duplication:

```python
dataclasses.replace(
    config, logging=dataclasses.replace(config.logging, log_to_wandb=False)
).configure_logging("inference_out.log")
```

Non-blocking readability suggestion.

elynnwu (commented):
I was curious about the preempt behavior here, and I think it's doing the right thing for cancelled mid-segment and cancelled between segments. If you also think it's appropriate for this PR, can you add a test for resuming a partial segment job? Since the mock now models both `finish()` and resumption, add a small test where `segment_NNNN/wandb_run_id` already exists and the segment resumes the same id, not create a new run

fme/core/logging_utils.py
```diff
@@ -153,6 +153,9 @@ def _configure_wandb(
         notes = _get_wandb_notes(_get_beaker_id())
         wandb.init(
             config=config_copy,
+            # wandb reads WANDB_NAME only on the first init in a process, so pass
+            # it explicitly for later runs (e.g. segments) to be named correctly.
+            name=os.environ.get("WANDB_NAME"),
```
elynnwu, line 158: If a future resume path (or a manual restart) ever launches with a different `WANDB_NAME`, the run silently renames with no warning. I think it's fine but do we want to assert name-consistency on resume or have a warning?

brianhenn: This PR doesn't change renaming behavior either way -- it only changes when the `WANDB_NAME` is read from the env (before only once at process start by the wandb backend, but now by us whenever `wandb.init` is called), thus allowing new wandb names for segments.

A resuming job (like training with `ResumeResultsConfig` set currently) *can* rename the job if a different `WANDB_NAME` from the prior wandb run name is supplied. In inference, `resumable=False` so a new wandb run is always created and no renaming can happen (again this PR doesn't change anything). If in the future we changed inference to be resumable, then the rename possibility would be there, but that's a separate issue from this change I think.

brianhenn:
> If you also think it's appropriate for this PR, can you add a test for resuming a partial segment job? Since the mock now models both finish() and resumption, add a small test where segment_NNNN/wandb_run_id already exists and the segment resumes the same id, not create a new run

The short answer is that there's no resumption (ACE or wandb) *within* segments/inference. For wandb, what happens is that the preempted run is left incomplete, and a new run (with the same name) is created that proceeds from the start. That behavior is unchanged with this PR, and a test would only be able to confirm that there's no resumption, so I am skipping it for now. If in the future we go to resumable inference we'll need to tackle what to do with the preempted wandb run, but I don't think anything here makes that task harder.

I changed the PR description to try to make it clearer that most wandb behavior around inference is not changed by this PR.

elynnwu (approved):

PR #1375, opened 2026-07-17
Enforce downscaling prediction always uses same patch size as trained on

AnnaKwa:
It is easy for a user to configure the input regions in `downscaling.predict` or `downscaling.inference` to be smaller or larger than the patch size used to train the model. In these cases (if patching is not configured for larger regions) the model will happily run on the different input shape but produce very biased outputs. This PR add a helper function to check the coarse data shape against the model's coarse input shape in those two entrypoints (inference entrypoint already checked for this) and raise an error in those cases. Inference is updated to use this same function for its check.

- Enforced that `PatchPredictorConfig. divide_generation, composite_prediction` must be either both True or both False. In practice, we don't mix True/False in any of the inference/predict/evaluator entrypoints.
- Added helper function `check_input_shape_supported` which is used across inference/predict/evaluator entrypoints to enforce that either i) prediction region is same size as the model training patch size, or ii) regions larger than the input size are divided into the model training patch size and later composited together. 

- [x] Tests added

### DIFF
--- a/fme/downscaling/evaluator.py
+++ b/fme/downscaling/evaluator.py
@@ -38,31 +38,22 @@ class Evaluator:
         model: DiffusionModel | DenoisingMoEPredictor | PatchPredictor,
         experiment_dir: str,
         n_samples: int,
-        patch_data: bool = False,
     ) -> None:
         self.data = data
         self.model = model
         self.experiment_dir = experiment_dir
         self.n_samples = n_samples
         self.dist = Distributed.get_instance()
-        self.patch_data = patch_data
 
     def run(self):
         aggregator = GenerationAggregator(
             self.data.dims,
             self.model.downscale_factor,
-            include_positional_comparisons=False if self.patch_data else True,
+            include_positional_comparisons=True,
             percentiles=[99.99, 99.9999],
         )
 
-        if self.patch_data:
-            batch_generator = self.data.get_patched_generator(
-                coarse_yx_patch_extent=self.model.coarse_shape,
-            )
-        else:
-            batch_generator = self.data.get_generator()
-
-        for i, batch in enumerate(batch_generator):
+        for i, batch in enumerate(self.data.get_generator()):
             with torch.no_grad():
                 logging.info(f"Generating predictions on batch {i + 1}")
                 outputs = self.model.generate_on_batch(batch, n_samples=self.n_samples)
@@ -211,7 +202,7 @@ class EvaluatorConfig:
             name="evaluator",
         )
         evaluator_model: DiffusionModel | DenoisingMoEPredictor | PatchPredictor
-        if self.patch.divide_generation and self.patch.composite_prediction:
+        if self.patch.needs_patch_predictor:
             evaluator_model = PatchPredictor(
                 model,
                 coarse_yx_patch_extent=model.coarse_shape,
@@ -220,19 +211,11 @@ class EvaluatorConfig:
         else:
             evaluator_model = model
 
-        if self.patch.divide_generation and not self.patch.composite_prediction:
-            # Subdivide evaluation into patches, do not composite them together
-            # No maps will be saved for this configuration.
-            patch_data = True
-        else:
-            patch_data = False
-
         return Evaluator(
             data=dataset,
             model=evaluator_model,
             experiment_dir=self.experiment_dir,
             n_samples=self.n_samples,
-            patch_data=patch_data,
         )
 
     def _build_event_evaluator(
--- a/fme/downscaling/evaluator.py
+++ b/fme/downscaling/evaluator.py
@@ -205,7 +205,6 @@ class EvaluatorConfig:
         if self.patch.needs_patch_predictor:
             evaluator_model = PatchPredictor(
                 model,
-                coarse_yx_patch_extent=model.coarse_shape,
                 coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
             )
         else:
@@ -243,7 +242,6 @@ class EvaluatorConfig:
         ):
             evaluator_model = PatchPredictor(
                 model=model,
-                coarse_yx_patch_extent=model.coarse_shape,
                 coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
             )
         else:
--- a/fme/downscaling/evaluator.py
+++ b/fme/downscaling/evaluator.py
@@ -25,6 +25,7 @@ from fme.downscaling.predictors import (
     DenoisingMoEPredictor,
     PatchPredictionConfig,
     PatchPredictor,
+    check_input_shape_supported,
 )
 from fme.downscaling.requirements import DataRequirements
 from fme.downscaling.typing_ import FineResCoarseResPair
@@ -203,6 +204,12 @@ class EvaluatorConfig:
         coarse_lon = dataset.coarse_extent_latlon_coords.lon
         # No-op when coarse_lon does not cross the prime meridian.
         model = model.with_rolled_lon(coarse_lon)
+        check_input_shape_supported(
+            model.coarse_shape,
+            dataset.coarse_shape,
+            self.patch,
+            name="evaluator",
+        )
         evaluator_model: DiffusionModel | DenoisingMoEPredictor | PatchPredictor
         if self.patch.divide_generation and self.patch.composite_prediction:
             evaluator_model = PatchPredictor(
@@ -242,6 +249,12 @@ class EvaluatorConfig:
         coarse_lon = dataset.coarse_extent_latlon_coords.lon
         # No-op when coarse_lon does not cross the prime meridian.
         model = model.with_rolled_lon(coarse_lon)
+        check_input_shape_supported(
+            model.coarse_shape,
+            dataset.coarse_shape,
+            self.patch,
+            name=f"event {event_config.name}",
+        )
         if (dataset.coarse_shape[0] > model.coarse_shape[0]) or (
             dataset.coarse_shape[1] > model.coarse_shape[1]
         ):
--- a/fme/downscaling/evaluator.py
+++ b/fme/downscaling/evaluator.py
@@ -207,7 +207,6 @@ class EvaluatorConfig:
         if self.patch.divide_generation and self.patch.composite_prediction:
             evaluator_model = PatchPredictor(
                 model,
-                coarse_yx_patch_extent=model.coarse_shape,
                 coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
             )
         else:
@@ -247,7 +246,6 @@ class EvaluatorConfig:
         ):
             evaluator_model = PatchPredictor(
                 model=model,
-                coarse_yx_patch_extent=model.coarse_shape,
                 coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
             )
         else:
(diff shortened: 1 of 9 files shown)


fme/downscaling/inference/inference.py
```diff
@@ -76,38 +77,21 @@ def _get_generation_model(
         input_shape = (len(coarse_coords.lat), len(coarse_coords.lon))
         # No-op when coarse_lon does not cross the prime meridian.
         base_model = self.model.with_rolled_lon(coarse_coords.lon)
-        model_patch_shape = base_model.coarse_shape
-
-        if model_patch_shape == input_shape:
-            # short circuit, no patching necessary
+        check_input_shape_supported(
```
AnnaKwa, line 80: This has the same behavior as the previous checks in inference.py, I just updated so that all entrypoints use the same code for the shape check.

fme/downscaling/predictors/composite.py
```diff
@@ -28,15 +28,15 @@ class PatchPredictionConfig:
     composite_prediction: bool = False
     coarse_horizontal_overlap: int = 1
 
-    @property
-    def needs_patch_data_generator(self):
```
AnnaKwa, line 32: In practice we never ended up using this mode in the inference entrypoints (break up into patches and report metrics on the non-composited patches) so I removed it to be consistent with the added requirement in the post init that `divide_generation=True requires composite_prediction=True.`

fme/downscaling/predictors/composite.py
```diff
@@ -28,15 +28,15 @@ class PatchPredictionConfig:
...
-        # If final predictions are not composited together, the BatchData is divided
-        # into patches before being passed to the top level no-target generation call.
-        # If final predictions are composited together, the BatchData is divided
-        # into patches within the PatchPredictor's generation method.
+    def __post_init__(self):
         if self.divide_generation and not self.composite_prediction:
-            return True
-        return False
+            raise ValueError(
+                "divide_generation=True requires composite_prediction=True."
```
AnnaKwa, line 38: This requirement is not strictly related to the other changes in the PR related to enforcing the model predicts on the same patch size it was trained on. I thought it simplified configuration though since we never use this mode (see comment above).

fme/downscaling/predictors/composite.py
```diff
@@ -49,30 +49,25 @@ class PatchPredictor:
     """
     Model prediction wrapper for generating a full-extent prediction
     by dividing the input into a grid of patches.
+
+    Patch size is inferred from the model's coarse_shape used in training.
```
frodre, line 57: I think this needs a rebase and force push? I thought the patch extents were dealt with in the previous PR

fme/downscaling/predictors/composite.py
frodre: Yeah, probably worth a deprecation for this option and eventual removal.

fme/downscaling/predictors/composite.py
```diff
@@ -170,6 +165,39 @@ def generate_on_batch_no_target(
         return prediction
 
 
+def check_input_shape_supported(
+    model_coarse_shape: tuple[int, int],
+    input_shape: tuple[int, int],
+    patch: PatchPredictionConfig,
+    name: str = "",
+) -> None:
+    model_shape = tuple(model_coarse_shape)
```
frodre: Why do we need the conversion when type is specified?

fme/downscaling/inference/inference.py
```diff
@@ -76,38 +77,21 @@ def _get_generation_model(
...
+            base_model.coarse_shape,
+            input_shape,
+            output.patch,
+            name=f"output {output.name}",
+        )
+        if tuple(base_model.coarse_shape) == tuple(input_shape):
```
frodre: This comparison is checked in the validation function.  Could switch to the patch configs boolean for whether to return the base model or a patch predictor.

fme/downscaling/test_evaluator.py
```diff
@@ -216,6 +216,52 @@ def test_evaluator_rolls_for_seam_crossing(tmp_path):
     assert (experiment_dir / "evaluator_maps_and_metrics.nc").exists()
 
 
+def test_evaluator_raises_when_larger_without_patching(tmp_path):
```
frodre, line 219: This is a really heavy setup  (saving model, saving data, reloading) for checking the retrieval of the generation model passes shapes.  The previous evaluator test was an end-to-end integration because there are a few different pathways that needed to be verified for seam rolling against the model/real data.

Could we mock the data/model instead of going through the construction and saving?  Similar to what happens in the inference tests, but maybe there are too many connected pieces to mock

fme/downscaling/test_evaluator.py
```diff
@@ -216,6 +216,52 @@ def test_evaluator_rolls_for_seam_crossing(tmp_path):
...
+    """The default evaluator refuses a region larger than the model patch when
+    patch prediction is not configured."""
+    coarse_shape = (4, 4)
+    downscale_factor = 2
+    fine_shape = (
+        coarse_shape[0] * downscale_factor,
+        coarse_shape[1] * downscale_factor,
+    )
+    paths = global_data_paths_helper(tmp_path)
+    full_fine_coords = LatLonCoordinates(
```
frodre: I think these should be the full data shape, not the model patch shape.

fme/downscaling/test_predict.py
```diff
@@ -225,6 +225,82 @@ def test_predictor_runs_seam_crossing(tmp_path, cls):
         assert value.shape[-2:] == fine_shape
 
 
+def _build_model_and_data_at_extent(tmp_path, lat_extent, lon_extent):
+    """A (4, 4)-patch model plus GriddedData subset to the given extent."""
+    coarse_shape = (4, 4)
```
frodre: Similar to evaluator. Pretty heavy setup for checking the model again, but maybe mocks are too complicated as well.

frodre (changes requested):
Thanks @AnnaKwa,  I think this branch needs a rebase and force push since it has some of the commits from #1376.  If we can simplify the entrypoint tests, I think that would be a win.  Otherwise minor comments and looks good.

fme/downscaling/predictors/composite.py
AnnaKwa: Not needed- I think Claude auto inserted that because `PatchPredictor.coarse_shape` lacked a type hint. I added the type hint there instead.

fme/downscaling/inference/inference.py
```diff
@@ -76,38 +77,20 @@ def _get_generation_model(
...
-
-        if model_patch_shape == input_shape:
-            # short circuit, no patching necessary
+        check_input_shape_supported(
+            base_model.coarse_shape,
+            input_shape,
+            output.patch,
+            name=f"output {output.name}",
+        )
+        if output.patch.needs_patch_predictor is False:
```
frodre:
```suggestion
        if not output.patch.needs_patch_predictor:
```

fme/downscaling/predictors/composite.py
frodre, line 21: Add some information to the docstring about the deprecation and expected settings.

fme/downscaling/predictors/composite.py
```diff
@@ -165,6 +165,39 @@ def generate_on_batch_no_target(
         return prediction
 
 
+def check_input_shape_supported(
+    model_coarse_shape: tuple[int, int],
+    input_shape: tuple[int, int],
+    patch: PatchPredictionConfig,
+    name: str = "",
+) -> None:
+    in_shape = tuple(input_shape)
```
frodre: Leftover tuple conversion

fme/downscaling/inference/inference.py
frodre: This was also my bad.  Claude flagged that this changes the behavior slightly because now if the shape matches and we have patching configured, we'd still get a `PatchPredictor` despite not needing it.  Unfortunately means we still need the duplicated model/data shape check for these conditional branches.

fme/downscaling/predict.py
```diff
@@ -114,6 +115,12 @@ def _get_generation_model(self):
         coarse_lon = self.data.coarse_extent_latlon_coords.lon
         # No-op when coarse_lon does not cross the prime meridian.
         base_model = self.model.with_rolled_lon(coarse_lon)
+        check_input_shape_supported(
+            base_model.coarse_shape,
+            self.data.shape,
+            self.patch,
+            name=f"event {self.event_name}",
+        )
         if self.patch.needs_patch_predictor:
```
frodre: Same behavior where we return a PatchPredictor even when we don't need one (model size and data size match).

fme/downscaling/predictors/composite.py
```diff
@@ -28,15 +28,15 @@ class PatchPredictionConfig:
...
-    def needs_patch_data_generator(self):
-        # If final predictions are not composited together, the BatchData is divided
-        # into patches before being passed to the top level no-target generation call.
-        # If final predictions are composited together, the BatchData is divided
-        # into patches within the PatchPredictor's generation method.
+    def __post_init__(self):
         if self.divide_generation and not self.composite_prediction:
-            return True
-        return False
+            raise ValueError(
```
frodre, line 37: The error conditions should probably have a unit test.

fme/downscaling/evaluator.py
```diff
@@ -203,28 +195,26 @@ def _build_default_evaluator(self) -> Evaluator:
...
         model = model.with_rolled_lon(coarse_lon)
+        check_input_shape_supported(
+            model.coarse_shape,
+            dataset.coarse_shape,
+            self.patch,
+            name="evaluator",
+        )
         evaluator_model: DiffusionModel | DenoisingMoEPredictor | PatchPredictor
-        if self.patch.divide_generation and self.patch.composite_prediction:
+        if self.patch.needs_patch_predictor:
```
frodre: For the events, we already use the coarse dataset shape check to see if we need the `PatchPredictor` wrap. We should add that consistency with events (and `inference` and `predict`) here.

frodre (commented):
Claude flagged a behavior change in inference that I think is worth addressing here. (Even though this was slightly inconsistent in other locations before the PR)  We only want a `PatchPredictor` wrap to occur when we have a domain size that's larger than the model patch size.  Right now, if the domain shape is equivalent to the patch size, we would still get a patched model in some locations.  Only other minor thing to update is the patch configuration docstring with the new constraints and a test for those errors.

fme/downscaling/predict.py
```diff
@@ -121,12 +121,12 @@ def _get_generation_model(self):
...
         )
-        if self.patch.needs_patch_predictor:
-            return PatchPredictor(
-                base_model,
-                coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
-            )
-        return base_model
+        if base_model.coarse_shape == self.data.shape:
+            return base_model
+        return PatchPredictor(
```
frodre, line 126: Need the conditional check `needs_patch_predictor` to decide whether or not we wrap

frodre (approved):
One small update, but LGTM!

fme/downscaling/predict.py
AnnaKwa: This is checked within `check_input_shape_supported` if the data shape is larger than the model coarse shape, so if that check passes in addition to the post init checks on `PatchPredictionConfig` and model shape!= data shape, then `needs_patch_predictor` is implicitly True.

PR #1376, opened 2026-07-17
Remove coarse_yx_patch_extent arg from `PatchPredictor` init 

AnnaKwa:
This is a small refactor that enforces that the patch size used in the `PatchPredictor` is always the same as the model coarse shape. This PR precedes https://github.com/ai2cm/ace/pull/1375 which adds a hard requirement to downscaling inference to disallow model predictions on ranges or patches that are not equal to the trained input data size (leads to large errors).

### DIFF
--- a/fme/downscaling/predict.py
+++ b/fme/downscaling/predict.py
@@ -117,7 +117,6 @@ class EventDownscaler:
         if self.patch.needs_patch_predictor:
             return PatchPredictor(
                 base_model,
-                self.data.shape,
                 coarse_horizontal_overlap=self.patch.coarse_horizontal_overlap,
             )
         return base_model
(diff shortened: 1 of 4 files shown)


fme/downscaling/predict.py
```diff
@@ -117,7 +117,6 @@ def _get_generation_model(self):
         if self.patch.needs_patch_predictor:
             return PatchPredictor(
                 base_model,
-                self.data.shape,
```
AnnaKwa, line 120: this line was actually a bug- would have used a patch size equal to the full input data size. We never ran into this because we always generated events with 16x16 deg range.

frodre (approved):
LGTM, do we need to note in the docstrings anywhere that we're now just using the default patch size of the model all the time?  Not that it would be outside of expectations, but probably good to make sure there's an explicit nod to it.

PR #1377, opened 2026-07-20
aimip_forcing: add daily variant for the daily 1° model

mcgibbon:
The AIMIP forcing and initial-condition scripts under `scripts/aimip_forcing/` were built for our submitted six-hourly model (ACE2.1-ERA5). Running the daily 1° model (e.g. `nobgd4ek`) through the AIMIP evaluation needs inputs that match its daily time step and its near-surface prognostic state (`TMP2m`, `Q2m`, `UGRD10m`, `VGRD10m`), which the published inputs lack. This adds a backward-compatible daily path; the six-hourly defaults and outputs are unchanged.

Changes:
- `scripts/aimip_forcing/interpolate_aimip_forcing.py`: `--freq` (default `6h`) threads the output time step through the time-coordinate and repeated-insolation builders. `--extension-start ""` disables the synthetic time extension and insolation repeat used to cover 2023–2024, for the case where the forcing source already spans the full window — the daily zarr runs through 2025, so the daily forcing uses that source's own time coordinate and `DSWRFtoa` directly. The extension and insolation-repeat windows are now CLI options rather than hardcoded literals.
- `scripts/aimip_forcing/create_aimip_ic_datasets.py`: `--include-near-surface` appends the daily model's near-surface prognostics to the IC variable set (default off).
- `scripts/aimip_forcing/Makefile`: daily targets (`create_daily_aimip_evaluation_datasets` and its components) that reuse the shared spatial regrid (the 1° grid is identical) and set the daily parameters. The forcing and IC/repeat-first steps are decoupled as in the six-hourly workflow, so the build is two steps: `make process_daily_forcing` then `make create_daily_aimip_evaluation_datasets`.
- `scripts/aimip_forcing/README.md`: documents the daily variant and the build-time parameters that must be confirmed against the daily zarr before running.

Build-time parameters left as overridable Makefile variables and flagged in the README (this PR is the script change only; the dataset build is not run here):
- `DAILY_ERA5_GCS_DATA` — the GCS path of the daily zarr the model trained on (mounted in training as `/climate-default/2026-03-19-era5-1deg-8layer-daily-1940-2025.zarr`).
- The daily calendar hour (assumed 06Z) and the IC source timestamps `DAILY_IC_TIMESTAMPS` (which `create_ic` reads by exact match) and prepend timestamp must be timestamps that exist in that zarr.

- [x] Tests added — no automated tests exist for these pipeline scripts; validated by `pre-commit` (ruff, ruff-format, mypy all pass), a synthetic smoke test of the daily code paths (`--freq 1D` yields 24 h steps; monthly→daily interpolation is NaN-free; the near-surface IC set is disjoint from the base prognostics), and `make -n` of the daily targets.
- [ ] If dependencies changed, "deps only" image rebuilt — n/a, no dependency change.

### DIFF
--- a/scripts/aimip_forcing/Makefile
+++ b/scripts/aimip_forcing/Makefile
@@ -78,10 +78,86 @@ create_modified_forcing: upload_modified_forcing
 create_aimip_evaluation_datasets: create_aimip_ics create_modified_forcing
 	@echo "All AIMIP evaluation datasets created"
 
+# ---------------------------------------------------------------------------
+# Daily variant (for the daily 1-degree model, e.g. nobgd4ek)
+#
+# Reuses the shared spatial regrid ($(LOCAL_REGRIDDED_FILE)); the 1-degree
+# Gaussian grid is identical for the six-hourly and daily models. Only the time
+# step and the initial-condition variable set differ, so this variant just
+# reinterpolates the regridded monthly forcing to a daily step and builds ICs
+# that include the daily model's near-surface prognostics.
+#
+# CONFIRM BEFORE RUNNING:
+#   - DAILY_ERA5_GCS_DATA: the GCS path of the daily zarr (the training config
+#     mounts it as /climate-default/2026-03-19-era5-1deg-8layer-daily-1940-2025.zarr).
+#   - The daily calendar hour (assumed 06Z below) and the IC / prepend timestamps
+#     must match timestamps that actually exist in that zarr.
+# ---------------------------------------------------------------------------
+DAILY_ERA5_GCS_DATA ?= gs://vcm-ml-intermediate/2026-03-19-era5-1deg-8layer-daily-1940-2025.zarr
+DAILY_START_TIME ?= 1978-10-01T00:00:00
+DAILY_END_TIME ?= 2024-12-31T23:59:59
+DAILY_OUTPUT_ZARR_NAME = 2026-07-20-aimip-era5-1deg-daily-forcing-1978-2024.zarr
+LOCAL_DAILY_OUTPUT_ZARR := $(LOCAL_DATA_DIR)/$(DAILY_OUTPUT_ZARR_NAME)
+GCS_PATH_DAILY_FORCING ?= gs://vcm-ml-intermediate/$(DAILY_OUTPUT_ZARR_NAME)
+DAILY_IC_OUTPUT_DIR := $(LOCAL_DATA_DIR)/aimip_ics_daily
+GCS_DAILY_IC_PATH ?= gs://vcm-ml-intermediate/2026-07-20-AIMIP-evaluation-ICs-daily/
+DAILY_IC_TARGET ?= 1978-09-30T06:00:00
+DAILY_FORCING_FIRST_STEP ?= 1978-10-01T06:00:00
+DAILY_MODIFIED_FORCING_NAME = 2026-07-20-aimip-era5-1deg-daily-forcing-1978-2024-repeat-first.zarr
+LOCAL_DAILY_MODIFIED_FORCING := $(LOCAL_DATA_DIR)/$(DAILY_MODIFIED_FORCING_NAME)
+GCS_PATH_DAILY_MODIFIED_FORCING ?= gs://vcm-ml-intermediate/$(DAILY_MODIFIED_FORCING_NAME)
+
+$(LOCAL_DAILY_OUTPUT_ZARR): $(LOCAL_REGRIDDED_FILE)
+	python interpolate_aimip_forcing.py $(LOCAL_REGRIDDED_FILE) $(LOCAL_DAILY_OUTPUT_ZARR) \
+	  --freq 1D \
+	  --ace2-era5-gcs-data $(DAILY_ERA5_GCS_DATA) \
+	  --start-time $(DAILY_START_TIME) \
+	  --end-time $(DAILY_END_TIME) \
+	  --extension-start ""
+
+upload_processed_daily_forcing: $(LOCAL_DAILY_OUTPUT_ZARR)
+	gsutil -m cp -r $(LOCAL_DAILY_OUTPUT_ZARR) $(GCS_PATH_DAILY_FORCING)
+
+process_daily_forcing: upload_processed_daily_forcing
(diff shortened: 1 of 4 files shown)


mcgibbon:
Pre-review agent (Claude): independent pre-review, no prior context on this change. Verdict below; one correctness fix applied and pushed to the branch (`1089efda7`).

## Fix applied (pushed to the branch)
**Daily IC source timestamps would have failed against the 06Z daily zarr.** The daily IC target (`$(DAILY_IC_OUTPUT_DIR)/.done`) passed `--target-timestamp` but no `--ic-timestamp`, so it fell back to `create_aimip_ic_datasets.py`'s six-hourly defaults (`IC_TIMESTAMPS`, all at `00:00:00`). `create_ic` does `era5.sel(time=ic_timestamp)` (exact label match), and the daily zarr carries one step per day at the assumed 06Z calendar hour — so every IC member would raise `KeyError`. The README's "confirm before running" list didn't cover this: it named `DAILY_IC_TARGET`, which is only *assigned* (`assign_coords(time=...)`), never selected, so it need not exist in the zarr — while the timestamps that *are* selected (the IC members) weren't exposed at all.

Fix: added an overridable `DAILY_IC_TIMESTAMPS` (the same five days at 06Z) and threaded it via `--ic-timestamp`; corrected the README "confirm" list to name the timestamps actually read from the zarr (`DAILY_IC_TIMESTAMPS`, `DAILY_FORCING_FIRST_STEP`) and to note that `DAILY_IC_TARGET` is a relabel only.

Also in the same commit — **documented the forcing-build ordering.** `create_daily_aimip_evaluation_datasets` = `create_daily_aimip_ics create_daily_modified_forcing`; the modified-forcing step reads `GCS_PATH_DAILY_FORCING` back from GCS, which only `process_daily_forcing` populates. The README's single documented command therefore would have failed at the prepend step. Reworded to the two-step flow (`make process_daily_forcing` then `make create_daily_aimip_evaluation_datasets`). This matches the six-hourly workflow's existing decoupling (its eval target likewise omits `process_aimip_forcing`), so I left the make graph as-is rather than adding the dependency. **Optional for the author:** if you'd prefer the single command to be self-contained, add `process_daily_forcing` as the first prerequisite of `create_daily_aimip_evaluation_datasets` (serial make updates prereqs left-to-right) and revert the README to one command.

## Verified, no change needed
- **Backward compatibility:** with defaults, `interpolate_aimip_forcing.py` is behavior-identical to main. `--freq` defaults to `6h`; `DEFAULT_EXTENSION_START`/`DEFAULT_REPEAT_SOURCE_START`/`_END` equal the old hardcoded literals; both `if extension_start:` branches reproduce the old `get_time_coordinate` + `get_repeated_insolation` + `concat` path exactly.
- **No-extension branch** (`--extension-start ""`): `time_coord` is the source zarr's own (decoded `use_cftime=False`) time coordinate, `DSWRFtoa` is used directly, and the three merge inputs (interpolated AIMIP forcing, `HGTsfc`, `DSWRFtoa`) all share that identical time coordinate — merge aligns cleanly. `--freq` threads correctly into both builders in the extension path.
- **IC near-surface flag:** `--include-near-surface` appends exactly `TMP2m/Q2m/UGRD10m/VGRD10m`, disjoint from `PROGNOSTIC_VARIABLES`; default off preserves the old set.
- **Makefile:** daily zarr / IC / modified-forcing outputs use distinct names, so no clobbering of six-hourly outputs; reuses `$(LOCAL_REGRIDDED_FILE)`; `.PHONY` updated.

## One observation (not blocking, no change made)
Under `--extension-start ""`, `--freq` is never read (neither `get_time_coordinate` nor `get_repeated_insolation` is called), so `--freq 1D` in the daily forcing target has no effect — the daily step comes entirely from the daily zarr's time coordinate. Harmless and arguably self-documenting; flagging so it isn't mistaken for what sets the output step. (I reworded the README so it no longer attributes the daily step to `--freq 1D`.)

## Verdict: ready for review (after the pushed fix)
Scope is a clean, backward-compatible daily path; six-hourly defaults and outputs are unchanged. The remaining deferred build-time parameters (`DAILY_ERA5_GCS_DATA`, the 06Z calendar-hour assumption, IC/prepend timestamps) are correctly flagged in the README as requiring confirmation against the daily zarr before the build is run.

scripts/aimip_forcing/Makefile
```diff
@@ -78,10 +78,91 @@ create_modified_forcing: upload_modified_forcing
...
+	  --include-near-surface
+	touch $(DAILY_IC_OUTPUT_DIR)/.done
+
+upload_daily_aimip_ics: $(DAILY_IC_OUTPUT_DIR)/.done
+	gsutil -m cp $(DAILY_IC_OUTPUT_DIR)/*.nc $(GCS_DAILY_IC_PATH)
+
+create_daily_aimip_ics: upload_daily_aimip_ics
+	@echo "Daily AIMIP IC datasets uploaded to $(GCS_DAILY_IC_PATH)"
+
+$(LOCAL_DAILY_MODIFIED_FORCING): data_dir
```
brianhenn, line 143: If I am following correctly this step shouldn't occur in the daily workflow. It was a hack in the 6-hourly workflow to get the first predicted timestep to be at 0Z even though the forcing started at 0Z (so we needed to fill in forcing at previous 18Z) so that we could then coarsen to daily to meet the AIMIP spec for daily data (averages over [0, 6, 12, 18Z]). Since the daily data is already coarsened (and FWIW timestamped at 6Z -- average of [6, 12, 18, 0Z]? -- which isn't quite compatible with the AIMIP spec) we don't need to do this.

scripts/aimip_forcing/create_aimip_ic_datasets.py
```diff
@@ -26,15 +26,21 @@
...
 
 
 def create_ic(
     era5: xr.Dataset,
     ic_timestamp: str,
     target_timestamp: np.datetime64,
 ) -> xr.Dataset:
-    ic = era5.sel(time=ic_timestamp)
-    return ic.assign_coords(time=target_timestamp)
+    # Select with a single-element list so `time` stays a length-1 dimension
```
brianhenn, line 39: FWIW I found that the inference loaded worked fine with a scalar IC/sample coordinate, but maybe that's out of date now.

PR #1378, opened 2026-07-20
Two-track SFNO: RH-input variant (saturation-normalized humidity on the local track)

mcgibbon:
Adds an RH-input variant of the two-track (global/local latent) SFNO. The base of this PR is the non-RH two-track experiment branch (`experiment/2026-07-17-local-global-sfno-oos`), so the diff isolates exactly what the RH representation adds on top of the two-track baseline.

The research question: expressing humidity as `q/q_sat` (a relative-humidity-like, O(1), climate-invariant channel) keeps humidity in-sample as the climate warms. The two-track point is that this pointwise RH representation must ride the local (never-spectral) track, off the spherical-harmonic path — so every derived `relative_humidity_<level>` channel is routed there.

Changes:
- `fme.core.humidity`, `fme.core.match_names`, `fme.core.step.saturation_normalization` — self-contained modules ported verbatim (`saturation_specific_humidity`, fnmatch name resolution, and the paired forward/inverse `q <-> q/q_sat` transform carrying humidity under internal `relative_humidity_<level>` channels with identity normalizer stats).
- `fme.core.step.single_module.step_with_adjustments` — gains an optional `saturation_normalization` parameter. Default `None` leaves behavior byte-identical to before (the L=0 single-track parity of the two-track PR is untouched). When set, `forward_transform` runs on the network input before normalization and `inverse_transform` maps the denormalized output back to `q` before the GMR inverse / corrector / ocean, so those adjustments operate in physical `q` space.
- `fme.ace.step.two_track.TwoTrackStepConfig` / `TwoTrackStep` — new `saturation_normalization` field, validated in `__post_init__` (via `resolve_all_fields`, mirroring the residual-prediction coherence check). Derived RH names get identity stats in the network and loss normalizers. The step packs derived input channels **after** the local segment (`in_packer = global_in + local_in + derived_input`), so they land in the local track that the network splits off at `len(global_in_names)`, and grows `local_in_channels` accordingly. Prognostic names are remapped through `residual_names`.
- Two new aimip-like training arms (`...-twotrack-rh`, `...-twotrack-rh-noco2`) with `saturation_normalization: [{names: [specific_total_water_*], input: append, prediction: false}]` — `q` stays on the global track and is still predicted in physical `q`; RH rides as an additional local input. Both pass `fme.ace.validate_config` and register a `run_training` call.

Scope / deviations:
- Only `input: append` with `prediction: false` is fully supported by the two-track step — the case this experiment needs. `input: replace` and `prediction: true` would move the source `q` across tracks (an ambiguous cross-track routing), so both are rejected with a clear error rather than guessed at; use the single-module step for those.
- The ported `test_saturation_normalization.py` step-level tests were rewritten to exercise `resolve_all_fields` directly and the two-track step, since the rh-branch's `single_module` config wiring (and its `SharedGlobalMeanRemovalConfig.qsat_scaled_names`) was deliberately **not** ported (it is a stale revert on an old main). The transform-behavior and build tests are kept verbatim; two-track routing, end-to-end, and validation tests live in `test_two_track.py`.

- [x] Tests added

### DIFF
--- a/configs/baselines/aimip-like/run-train.sh
+++ b/configs/baselines/aimip-like/run-train.sh
@@ -148,3 +148,5 @@ run_training "train-4deg-daily-v2-era5-only-noresid-noclip-noco2.yaml" "train-4d
 run_training "train-4deg-daily-v2-era5-only-noresid-noclip-notnorm.yaml" "train-4deg-daily-v2-era5-only-noresid-noclip-notnorm-rs0" 1
 run_training "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack.yaml" "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rs0" 1
 run_training "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-noco2.yaml" "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-noco2-rs0" 1
+run_training "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh.yaml" "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh-rs0" 1
+run_training "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh-noco2.yaml" "train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh-noco2-rs0" 1
--- a/configs/baselines/aimip-like/train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh-noco2.yaml
+++ b/configs/baselines/aimip-like/train-4deg-daily-v2-era5-only-noresid-noclip-twotrack-rh-noco2.yaml
@@ -1,0 +1,554 @@
+seed: 0
+experiment_dir: /results
+save_checkpoint: true
+checkpoint_every_n_batches: 5000
+validate_using_ema: true
+ema_checkpoint_save_epochs:
+  start: 5
+  step: 5
+max_epochs: 120
+ema:
+  decay: 0.999
+train_aggregator:
+  ensemble_metrics: true
+inference:
+  - name: 10year
+    weight: 0.0
+    epochs:
+      start: 1
+      step: 2
+    n_forward_steps: 3652
+    forward_steps_in_memory: 73
+    n_ensemble_per_ic: 1
+    loader:
+      start_indices:
+        times:
+          - '2015-01-01T00:00:00'
+          - '2015-01-02T00:00:00'
+          - '2015-01-03T00:00:00'
+          - '2015-01-04T00:00:00'
+          - '2015-01-05T00:00:00'
+          - '2015-01-06T00:00:00'
+          - '2015-01-07T00:00:00'
+          - '2015-01-08T00:00:00'
+      dataset:
+        data_path: /climate-default
+        file_pattern: 2026-04-17-era5-4deg-8layer-daily-1940-2025.zarr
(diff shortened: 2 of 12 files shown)


mcgibbon:
Claude: Independent pre-review (Opus, fresh context, worktree-isolated). Correctness-focused review gating the 2 RH training arms. Verdict: **ready to launch.** No correctness issues found; one minor (non-blocking) messaging nit below.

**Verified against the stated correctness priorities:**

1. **Local-track routing is correct.** `in_packer = Packer(global_in_names + local_in_names + derived_input_names)` places the derived `relative_humidity_<level>` channels strictly after the global segment. The network splits input at `global_in_channels` (`two_track_sfnonet.py`: `global_in = x[..., :global_in_channels]`, `local_in = x[..., global_in_channels:]`), and `build_two_track` is sized with `global_in_channels = len(global_in_names)` (unchanged) and `local_in_channels = len(local_in_names) + len(derived_input_names)`. So the 8 RH channels ride the local (never-spectral) track; the source q stays on the global track. Packer order and network split agree. In the -rh config: 34 global inputs, 10 local + 8 RH = 18 local. Confirmed by `test_saturation_append_routes_derived_rh_to_local_track` and end-to-end run.

2. **q_sat anchoring is correct.** `forward_transform` computes q_sat from the raw pre-normalization input (`saturation_normalization.forward_transform(input, network_input)` in `step_with_adjustments`, using `input`'s temperature + surface pressure), inserted before `normalizer.normalize`. `inverse_transform` runs after `denormalize` and before the GMR-inverse / corrector / ocean / prescribed steps, so those operate in physical space. inverse uses the cached input-step q_sat.

3. **`saturation_normalization=None` parity is byte-identical.** Read the restructured pre-normalize block: with sat=None, GMR-present path (`normalize(gmr_output)` then merge extras) and no-GMR path (`network_input=input`, `normalize(input)`) both reproduce the original exactly. Only a brand-new code path activates when sat is set; no existing caller's behavior changes. #1366's L=0 single-track equivalence is preserved.

4. **Config correctness confirmed.** `specific_total_water_0..7` on the global track (in + out); `air_temperature_0..7` and `PRESsfc` present as global inputs so q_sat resolves at every matched level; 8 derived RH = 8 matched stw levels. `-rh` differs from `-twotrack` by ONLY the `saturation_normalization` block; `-rh-noco2` differs from `-rh` by exactly the established noco2 diff. Both pass `validate_config --config_type train`.

5. **Guards are real.** `input='replace'` and `prediction=True` raise loudly in `TwoTrackStep.__init__` before any misrouting; residual-prediction/RH coherence is validated in `__post_init__`. All tests confirm the raises.

6. **No signature-insertion breakage:** the new `saturation_normalization` param sits mid-signature, but all 5 callers of `step_with_adjustments` pass every argument by keyword, so nothing shifts.

**Tests / checks run** (slot venv, CPU): `test_two_track.py` + `test_saturation_normalization.py` + `test_humidity.py` + `test_match_names.py` = 50 passed; full `fme/core/step/` = 159 passed, 1 skipped; ruff/ruff-format/mypy clean on all changed files.

**Minor (non-blocking):** the two guard messages in `two_track.py` direct users to "the single-module step" for `replace`/`prediction`, but `SingleModuleStepConfig` does not wire `saturation_normalization` (only `step_with_adjustments` accepts the param; no config field). A user following that advice would hit a dacite strict-mode rejection, not a working fallback. Not a silent failure and not launch-gating — worth softening the message or wiring the single-module config in a follow-up.

PR #1379, opened 2026-07-20
Add cBottle-style label dropout for combined-source training

mcgibbon:
Combined-source (multi-dataset) training conditions the model on a per-source label via the `LabelEncoding` → `ConditionalLayerNorm` path. To let such a model learn a genuine *unconditional* mode alongside its per-source conditional ones, this adds cBottle-style label dropout ([arXiv:2505.06474](https://arxiv.org/pdf/2505.06474)): on a configured fraction of training samples, the per-source label is withheld and the model sees the empty/all-zeros label instead. This is a data-loading feature only — no model-architecture change.

Changes:
- `fme.ace.data_loading.config.DataLoaderConfig`: add `label_dropout_rate` (default `0.0`), validated to `[0, 1]` in `__post_init__`.
- `fme.ace.data_loading.batch_data.BatchData.from_sample_tuples`: before encoding, draw a per-sample `Bernoulli(label_dropout_rate)` mask and replace each dropped sample's label set with the empty set. The empty set encodes to an all-zeros one-hot row (via the existing `LabelEncoding.encode` path) — a per-sample route to the unconditional state, rather than zeroing the whole batch.
- `fme.ace.data_loading.getters`: thread `label_dropout_rate` through `CollateFn`; in `get_gridded_data`, apply it only when `train=True` (forced to `0.0` on the validation path, so validation metrics keep true per-source conditioning).

Zero-init / identity finding: the label-conditioning Linear layers are zero-initialized, so an all-zeros one-hot yields identity (additive-zero) conditioning at init — no change was needed. In `ConditionalLayerNorm.forward` the one-hot enters additively as `scale += W_scale_labels(labels)` and `bias += W_bias_labels(labels)`; `reset_parameters` sets both `W_scale_labels` and `W_bias_labels` weight *and* bias to `0.0` (`fme/core/models/conditional_sfno/layers.py`), and the swin conditioning path's `adaln_labels` is likewise zeros-initialized (`fme/core/models/swin_transformer/swin_layers.py`). An all-zeros one-hot therefore contributes only the (zero-at-init) label-bias term, which is exactly identity conditioning. After training, the withheld state maps to the learned label-bias term, shared across all unconditional samples and distinct from any active one-hot combination — that is the unconditional mode.

No-op guard: when the dataset has no labels (`available_labels is None` ⇒ `label_encoding is None`, e.g. `builder.conditional: false`), the dropout branch is never reached, so `from_sample_tuples` returns `labels=None`; a `label_dropout_rate > 0` then has no effect and does not crash.

Tests:
- `test_batch_data.py`: `label_dropout_rate=1.0` zeros all labels; `0.0` leaves them unchanged; a seeded run is per-sample and deterministic; `rate>0` with no `label_encoding` is a no-op.
- `test_data_loader.py`: through `get_gridded_data`, `rate=1.0` on the train path emits only zeroed labels; the validation path preserves labels despite `rate=1.0`; an out-of-range rate is rejected.
- `test_train.py::test_train_with_label_dropout`: a `NoiseConditionedSFNO` (conditional) model trains end-to-end with `label_dropout_rate=1.0`, confirming the `ConditionalLayerNorm` forward handles the withheld-label state.

- [x] Tests added

### DIFF
--- a/fme/ace/data_loading/batch_data.py
+++ b/fme/ace/data_loading/batch_data.py
@@ -741,7 +741,15 @@ class BatchData:
         horizontal_dims: list[str] | None = None,
         label_encoding: LabelEncoding | None = None,
         allow_missing_variables: bool = False,
+        label_dropout_rate: float = 0.0,
     ) -> "BatchData":
+        """Collate sampled dataset items into a ``BatchData``.
+
+        ``label_dropout_rate`` is the probability that each sample has its source
+        label withheld (replaced by the empty/all-zeros label), so the model learns
+        an unconditional mode. It has no effect when ``label_encoding is None`` and
+        should be left at 0 on the validation/inference paths.
+        """
         (
             sample_data,
             sample_times,
@@ -764,7 +772,17 @@ class BatchData:
                 raise ValueError("label_encoding must be provided if labels are used.")
             labels = None
         else:
-            labels = label_encoding.encode(list(sample_labels), device="cpu")
+            encoded_labels = list(sample_labels)
+            if label_dropout_rate > 0.0:
+                # cBottle-style label dropout: withhold each dropped sample's label
+                # by replacing it with the empty set, which encodes to an all-zeros
+                # one-hot row (the unconditional state).
+                dropped = torch.rand(len(encoded_labels)) < label_dropout_rate
+                encoded_labels = [
+                    set() if drop else label_set
+                    for drop, label_set in zip(dropped.tolist(), encoded_labels)
+                ]
+            labels = label_encoding.encode(encoded_labels, device="cpu")
         return BatchData.new_on_cpu(
             data=batch_data,
             time=batch_time,
--- a/fme/ace/data_loading/config.py
+++ b/fme/ace/data_loading/config.py
@@ -40,6 +40,12 @@ class DataLoaderConfig:
             Each pool slot holds one window of
             ``batch_size * (time_buffer + n_timesteps)`` tensors in memory.
             Requires ``time_buffer > 0``.
+        label_dropout_rate: Probability, in [0, 1], that each training sample has
+            its source label withheld (replaced by the empty/all-zeros label). This
+            is cBottle-style label dropout: it forces the model to learn a genuine
+            unconditional mode alongside the per-source conditional ones. Applied
+            per-sample on the training path only (never validation or inference),
+            and a no-op when the dataset has no labels (``available_labels is None``).
 
     Note:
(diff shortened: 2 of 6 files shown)


mcgibbon:
Claude: Independent pre-review complete — **ready for review**, no fixes applied (nothing rose above optional-comment level). Verified in an isolated worktree: train-path-only gate (validation/inference/coupled callers all take the `0.0` default), no-op guard when labels are unused, the identity-at-init claim (both `W_scale_labels`/`W_bias_labels` and the swin `adaln_labels` Linear are zero-init → all-zeros one-hot is additive-zero conditioning), genuinely per-sample zeroing via `LabelEncoding.encode`, sound per-worker RNG, and that the 7 unit tests + medium-duration smoke-train exercise the claims.

Two optional, non-blocking observations left for your call:

1. **Silent no-op footgun.** Setting `label_dropout_rate > 0` on a label-less dataset (`conditional: false`) silently does nothing (documented + tested, so defensible). A one-line `logging.warning` in `get_gridded_data` when `label_dropout_rate > 0` and `label_encoding is None` would catch a misconfiguration.
2. **Coupled training does not thread dropout.** `coupled/train/train_config` uses `get_gridded_train_data`, which does not pass the kwarg — out of scope here, since combined-source training runs through the ace gridded `ConcatDataset` path this PR wires.

mcgibbon:
Claude: Per Jeremy, this feature merges only once the label-dropout experiments show improvement on the response axes — so the runs gate the merge (not the reverse), and the PR stays open until they are evaluated.

Launched 2026-07-20 on the `experiment/2026-07-16-combined-shield-era5` branch (this PR's commit cherry-picked on, `label_embed_dim: 2` added per Jeremy), 4 arms at 1 GPU, ai2/ace high priority:

| arm | label_dropout_rate | wandb |
|---|---|---|
| B' (no-dropout anchor) | 0.0 | [j2ox5jvv](https://wandb.ai/ai2cm/ace/runs/j2ox5jvv) |
| C | 0.10 | [bntku16k](https://wandb.ai/ai2cm/ace/runs/bntku16k) |
| D | 0.50 | [tfort5ji](https://wandb.ai/ai2cm/ace/runs/tfort5ji) |
| E | 0.90 | [v9irj2he](https://wandb.ai/ai2cm/ace/runs/v9irj2he) |

Evaluated against the running `label_embed_dim=0` anchor (B, tklu8mif) and the label-off control (A, d3oopmrt) on the +4K and future-scenario response harnesses. Tracked in research investigation 2026-07-16-combined-shield-era5-training.

PR #1380, opened 2026-07-20
Remove mse_fractional_components validation metric

Arcomano1234:
Remove approximate `mse_fractional_components` validation metrics now that true per-channel loss is logged, and drop the `loss_scaling`/`effective_loss_scaling` plumbing that existed only to compute them.

Changes:
- `OneStepDeterministicAggregator`: remove `_get_loss_scaled_mse_components` and stop emitting `*/mean/mse_fractional_components/*`
- `build_one_step_aggregator`, `OneStepAggregatorConfig.build`, `LegacyFlagOneStepAggregatorConfig.build`, coupled `OneStepAggregator`, ace/coupled train configs, and inference evaluator: stop accepting/passing `loss_scaling` into validation aggregators
- `fme.core.loss`, ace and coupled steppers: remove now-unused `effective_loss_scaling` properties and the `CoupledTensorMapping` type
- Delete `test_deterministic.py` covering the removed fractional-component helper; update related train/aggregator/loss tests

- [x] Tests added

Should finally resolve everything in #485


mcgibbon (changes requested):
Can/should we remove effective_loss_scaling and associated plumbing?

> After the PR, effective_loss_scaling survives only as a definition-and-delegation chain with no production reader left:
> - fme/core/loss.py:307 (base computed property: loss_normalizer_std / weight)
> - fme/core/loss.py:897–898 (delegation)
> - fme/ace/stepper/single_module.py:1811–1817 (delegation)
> - fme/coupled/stepper.py:1455–1458 and 1924–1925 (delegation)
> - plus fme/coupled/test_loss.py:37 (mock.effective_loss_scaling = {} — a Mock(spec=StepLoss) setup line that would also need removing if the property goes)
> 
> It's a pure reporting property — not used in actual loss computation and not required by any ABC/abstractmethod. So by the PR's own rationale ("drop the loss_scaling plumbing that existed only to compute [the metric]"), this is the same plumbing one layer deeper, and it's now dead.

Arcomano1234:
> Can/should we remove effective_loss_scaling and associated plumbing?

Yeah looking at the code I think it makes sense to remove these are they are completely unused

mcgibbon (approved):
LGTM

PR #1381, opened 2026-07-21
Small fix to `StaticSpatialMaskingConfig` docs

jpdunc23:
Fixes incorrect wording on the `fill_value` description in the `StaticSpatialMaskingConfig`. The fill value is used _inside_ of masked regions, not 'outside' of them (i.e., valid regions).

Changes:
- No API changes, docs only.

- [ ] Tests added


elynnwu (approved):
lgtm

PR #1382, opened 2026-07-21
Switch to datetime-based directory labels for segmented inference

spencerkclark:
This PR switches from labeling subdirectories in segmented runs with the segment number to labeling them with the start date of the first (or only) ensemble member of the segment.

Changes:
- Segment directories by default are now labeled with the start date of the first (or only) ensemble member of the segment, using the `strftime` format of `"%Y%m%dT%H"`.
- Adds a command line option, `--segment-label-format`, to the inference entrypoint to modify this `strftime` format if desired, e.g. if using a sub-hourly timestep for instance.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/inference.py
+++ b/fme/ace/inference/inference.py
@@ -495,8 +495,24 @@ def _get_segment_label(
     n_forward_steps: int,
     segment_label_format: str,
 ) -> str:
-    segment_start_time = initialization_time + timestep * segment * n_forward_steps
-    return segment_start_time.strftime(segment_label_format)
+    segment_length = n_forward_steps * timestep
+    current_start_time = initialization_time + segment * segment_length
+    current_label = current_start_time.strftime(segment_label_format)
+
+    if segment > 0:
+        previous_start_time = initialization_time + (segment - 1) * segment_length
+        previous_label = previous_start_time.strftime(segment_label_format)
+        if previous_label == current_label:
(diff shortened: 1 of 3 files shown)


fme/ace/inference/inference.py
```diff
@@ -463,20 +469,60 @@ def run_inference_from_config(config: InferenceConfig):
...
         config: inference configuration to be used for each individual segment. The
             provided initial condition configuration will only be used for the first
             segment.
         segments: total number of segments desired. Only missing segments will be run.
+        segment_label_format: strftime format used to render each segment's start
+            time—specifically, the start time of the first (or only) ensemble
+            member—into its directory/wandb-run label. Defaults to hour precision,
+            which is truncated (not rounded), so distinct segment start times that
+            share an hour would collide; pass a more precise format
+            (e.g. ``"%Y%m%dT%H%M%S"``) if that's a concern for your timestep or
```
elynnwu: [nit] consider raising an error if consecutive computed labels are equal instead of silently erroring

elynnwu (approved):
I'm good with `segment_%Y%m%dT%H` as you have implemented

fme/ace/inference/inference.py
spencerkclark: Indeed that's a good idea.  I added a check in 46dd1bb.

PR #1384, opened 2026-07-23
Make evaluating all forward steps configurable per inline-validation entry

jpdunc23:
Adds `evaluate_all_steps: bool = True` to each inline-validation entry, in both the ACE and coupled trainers. The default keeps today's behavior: every validation batch rolls out the full data window and logs a loss at every step. Setting it to `false` evaluates only the steps the train stepper would evaluate for that batch, so validation cost stays commensurate with training cost.

That matters under a long-tailed stochastic step schedule. With steps sampled from {1, 2, 4, 12, 20} the validation window is sized to the schedule max, so every validation batch rolls out 20 steps while training averages around 3. What opting out costs is metric density: `loss_step_N` then averages only the batches that sampled more than N steps, so it is noisier at longer leads.

Sparse per-step losses let ranks hold different `loss_step_N` key sets — unequal batch counts, or a rank with no batches at all — which the previous per-key `reduce_mean` in `PerStepLossAggregator` would turn into a hang. The aggregator now agrees on the global key set first, via `gather_object`/`scatter_object`, then reduces sums and counts over that set. Means are count-weighted unconditionally, replacing an unweighted mean of per-rank means. Reported per-step losses therefore shift slightly wherever ranks have unequal batch counts, including in existing full-window runs; the shift is small next to batch noise, and the weighted validation loss used for checkpoint selection is unaffected.

That reduction surfaced a latent bug in `ModelTorchDistributed.scatter_object`: it scattered from `_data_rank == 0` rather than global rank 0, which deadlocks under spatial parallelism, where every spatial rank has `_data_rank == 0`. It now scatters from global rank 0, matching `gather_object`.

The flag governs inline validation only. The post-epoch evaluation of training data in `train_one_epoch` keeps evaluating all steps, so its per-step losses stay dense and comparable across epochs.

`LossSchedule.init_for_epoch` now raises `EpochNotProvidedError` for any non-constant schedule when a batch carries no epoch. Previously only milestone schedules raised, so a milestone-free stochastic schedule silently evaluated every data step instead. Constant fixed-step schedules still accept an epoch-less batch, where the two evaluation modes coincide anyway.

Changes:
- `fme.ace.train.train_config.InlineValidationConfig` and `fme.coupled.train.train_config.InlineValidationConfig`: new `evaluate_all_steps: bool = True` field; configs omitting it parse and behave as before
- `fme.core.generics.validation.run_validation_loop` / `run_validation` and `fme.core.generics.trainer.ValidationTask`: new `evaluate_all_steps` pass-through
- `fme.ace.aggregator.loss_metrics.PerStepLossAggregator`: count-weighted reduction over an agreed key set, deadlock-free with mismatched or empty per-rank keys
- `fme.core.distributed.model_torch_distributed.ModelTorchDistributed.scatter_object`: scatter from global rank 0
- `fme.ace.stepper.loss_schedule.LossSchedule.init_for_epoch`: raises for any non-constant schedule without an epoch, using the new `TimeLengthSchedule.is_constant`

- [x] Tests added


pr-plan.md
```diff
@@ -0,0 +1,164 @@
...
+```python
+class PerStepLossAggregator:
+    def record(self, metrics: dict[str, torch.Tensor]) -> None:  # CHANGED — sums/counts kept as tensors for collective reduction
+        ...
+
+    def get_logs(self, label: str) -> dict[str, float]:  # CHANGED — distributed reduction reworked, see below
+        ...
+```
+
+### Critical detail — deadlock-free, count-weighted reduction
```
jpdunc23: NB: With new count-weighted reduction and `evaluate_all_steps=False`, the uncertainty in each validation `loss_step_N` metric is inversely (sqrt) proportional to 1 minus the trainer's cumulative step sampling probabilities, i.e. if `p = P(steps <= N)` and `B` is the total number of validation batches then `loss_step_N` is computed from `B * (1 - p)` batches on average.

Also note that with our existing `seed=0` for validation, the number of steps for each validation batch are consistent from epoch to epoch.

Together this implies that larger validation batch sizes increases the probability that long leads won't be logged at all, i.e. it is best to set the batch size small enough so that `B * (1 - p) >> 1`.

pr-plan.md
```diff
@@ -0,0 +1,198 @@
...
+  deterministic runs; negligible vs. batch noise. Chosen over preserving the old definition to
+  avoid a validation-only fork: the same aggregator serves training-side `loss_step_N` logs, and
+  `get_logs` calls are already rank-symmetric there.
+
+## `fme/core/generics/validation.py` (modified)
+
+```python
+def run_validation_loop(
+    ...,
+    evaluate_all_steps: bool = True,  # NEW — passed to stepper.train_on_batch; all callers pass explicitly, default is dense-by-default for the generic layer
```
jpdunc23: The default here is conservative: Don't silently drop metrics on the generic. Callers make a conscious choice to override.

pr-plan.md
jpdunc23: Some explanation along these lines of the `evaluate_all_steps=False` case should make it into the inline docs / docstrings.

pr-plan.md
```diff
@@ -0,0 +1,198 @@
...
+The post-epoch train-data evaluation loop (`train_one_epoch`'s
+`train_on_batch(..., evaluate_all_steps=True)`) is intentionally unchanged — this PR scopes the
+flag to inline validation.
+
+## `fme/ace/train/train_config.py` (modified)
+
+```python
+@dataclasses.dataclass
+class InlineValidationConfig:
+    evaluate_all_steps: bool = False  # NEW — False evaluates only the steps the train stepper samples for the batch (the stochastically-sampled count under a schedule, the fixed count otherwise); True rolls out the full data window with dense per-step metrics
```
jpdunc23: Default `False` here seems desirable to me to avoid the issue I encountered: Unexpectedly slow validation due to every batch rolling out the full.

On the other hand, default `True` is more coherent with the case of fixed integer `n_forward_steps` (note that `False` behaves identically when `n_forward_steps` is an integer) and avoids any confusion about the uncertainty in the `loss_step_N` metrics.

I think default `True` is also @mcgibbon preference, so we should go with that.

pr-plan.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+- Default `False` restores the pre-existing sampled-rollout behavior: with a long-tailed step
+  schedule (e.g. sampled from {1, 2, 4, 12, 20}) the validation window is sized to the schedule
+  max, so evaluating every step makes validation ~×13–16 slower than training-commensurate cost.
+  Users who monitor long-lead metrics opt in per entry. The default flip and the count-weighting
+  change are called out in the PR description/changelog, not in docstrings or runtime logging.
+- Under the default, `loss_step_N` at different leads aggregates over different numbers of
+  batches (unbiased but noisier at long leads); interpreting them in light of the step-sampling
+  probabilities is documented as the user's responsibility.
```
jpdunc23: See comment above. We should use default `True`.

pr-plan.md
```diff
@@ -0,0 +1,198 @@
...
+) -> AggregatorSummary:
+    ...
+```
+
+## `fme/core/generics/trainer.py` (modified)
+
+```python
+@dataclasses.dataclass
+class ValidationTask(Generic[BD, TO]):
+    evaluate_all_steps: bool = True  # NEW — both trainers pass explicitly; dense-by-default at the generic layer
```
jpdunc23: Another possibility that came up during planning is `evaluate_n_steps: "sampled" | "all" | int`. I personally don't like this choice because the `int` handling is tricky for little added value, and `"sampled"` doesn't really make sense for fixed integer `n_forward_steps`.

jpdunc23: Note the interpretation of `evaluate_all_steps: bool`:

> the flag doesn't encode an outcome, it encodes a policy — `False` = "evaluate what training would evaluate for this batch", `True` = "evaluate the full window regardless"

pr-plan.md
```diff
@@ -0,0 +1,198 @@
...
+- No `__post_init__` validation added (nothing invalid to reject). A fixed-integer step-count
+  mode was considered and dropped as speculative; a later widening to `bool | int` stays
+  config-compatible.
+
+## `fme/coupled/train/train_config.py` (modified)
+
+```python
+@dataclasses.dataclass
+class InlineValidationConfig:
+    evaluate_all_steps: bool = False  # NEW — mirrors the ACE field; False evaluates only the steps the coupled loss samples for the batch, True rolls out the full window with dense per-realm per-step metrics
```
jpdunc23: Same here: default `True`.

pr-plan.md
jpdunc23: Agreed — the plan now commits the field docstring to this guidance: `loss_step_N` averages ~`B·(1−p)` batches, per-batch step draws are epoch-fixed under the validation seed, and leads with `B·(1−p)` ≲ 1 may never be logged (so prefer smaller validation batch sizes). See the `fme/ace/train/train_config.py` section of `pr-plan.md`.

pr-plan.md
jpdunc23: Rationale now carried inline next to the parameter in `pr-plan.md`: the generic layer never silently drops metrics; callers consciously override.

pr-plan.md
jpdunc23: Done — the plan now uses `evaluate_all_steps: bool = True` in both configs, with the decision rationale (and the rejected `False` alternative) inline; the sampled speedup is opt-in via `false`, and the PR description no longer frames a default flip.

pr-plan.md
jpdunc23: Resolved by the default change to `True` — see the config section of `pr-plan.md`.

pr-plan.md
jpdunc23: Recorded inline in the config section of `pr-plan.md`: the bool encodes a policy (`False` = what training would evaluate for the batch), and the `evaluate_n_steps: "sampled" | "all" | int` spelling is noted as rejected. The docstring will state the policy reading.

pr-plan.md
jpdunc23: Done — the coupled default is `True`, mirroring the ACE field.

pr-plan.md
```diff
@@ -0,0 +1,212 @@
...
+- Per family: build dense sum and count vectors of length `global_max + 1` (zeros where the rank
+  never saw the step), one `reduce_sum` each; mean = sum/count; steps with global count 0 are
+  skipped. Non-step keys get the same sum/count treatment keyed by name.
+- Family names and non-step key names are assumed rank-symmetric (as today — every rank that
+  records batches produces the same families); the delta this PR must handle is per-step key sets
+  differing across ranks within a family. This robustness is live for both trainers, since
+  coupled sampled evaluation also produces sparse per-step key sets.
+- Count-weighting is unconditional (no legacy path). Reported values shift slightly vs. today's
+  unweighted mean-of-rank-means whenever ranks have unequal batch counts — including existing
+  deterministic runs; negligible vs. batch noise. Chosen over preserving the old definition to
```
jpdunc23: "negligible vs. batch noise": What is the worst case here?

jpdunc23:
Pre-review agent: results of a self pre-review audit of this PR, for reviewer orientation.

1. ✅ **Silent-failure pass** — An empty aggregator returns `{}` (tested); keys with a global count of 0 are skipped explicitly; the new `evaluate_all_steps` field parses under strict dacite with omitted/`true`/`false` all tested in both trainers. The default `True` means every previously-valid YAML behaves identically.
2. ✅ **Behavior-change pass** — Two intended changes to existing paths, both stated in the PR description: `PerStepLossAggregator` means become count-weighted (values shift slightly when rank batch counts are unequal), and `ModelTorchDistributed.scatter_object` now scatters from global rank 0 instead of `_data_rank == 0` (latent deadlock fix under spatial parallelism; the two coincide in pure data-parallel, so no behavior change there). The only other `run_validation` caller (`fme/ace/inference/evaluator.py`) inherits the default and is unaffected.
3. ✅ **Differential-test pass** — The parallel aggregator test uses rank-dependent values/counts, so count-weighted and mean-of-rank-means results are numerically different (e.g. 0.667 vs 0.5 for step 0 with 2 ranks), and the pre-PR code would hang outright on the mismatched key sets. The passthrough tests assert the actual flag values reach `train_on_batch`.
4. ✅ **Sibling/mirror pass** — ACE and coupled both get the field, wiring, docstrings, and tests; the LR-tuning `validate_stepper` path is wired in both. The post-epoch train-data evaluation loop is intentionally out of scope (noted in `pr-plan.md`).
5. ✅ **Config back-compat pass** — Purely additive field with a behavior-preserving default; no checkpoint-persisted state touched.
6. ✅ **Split heuristic** — 10 files, one feature plus the small distributed fix it required; no split warranted.
7. ✅ **Abstract-or-final / inheritance-depth pass** — Only flat test helper classes added.
8. ✅ **AI-agent artifact pass** — The in-function `time_length_probabilities` import in `fme/coupled/test_stepper.py` matches three pre-existing instances in that file, so it was left as the file's idiom.
9. ✅ **Builder pattern pass** — `_get_validation_callback` reading `entry_config.evaluate_all_steps` follows the existing pattern for `weight`/`name` in the same functions; no new violation class.

Overall: CI green, tests differential, coverage mirrored across ACE/coupled, and the description already flags both behavior changes. The one open item is the unresolved `pr-plan.md` thread on the worst-case size of the count-weighting shift.

pr-plan.md
jpdunc23: Worst case: the count-weighted mean differs from the old mean-of-rank-means by at most the between-rank spread of per-key means times how far the count weights deviate from uniform. For existing `evaluate_all_steps=True` runs per-rank batch counts are equal or off by one, so that deviation is ~R/(2B). Concretely, with R=8 ranks and B=100 batches (per-rank counts 12 or 13) the weight deviation is 0.02; if per-batch loss noise is 10% of the loss value, rank means spread about ±3%, so the shift is ≤ ~0.2% of the loss value — about 5× below the metric's own batch-sampling noise of 10%/√100 = 1%, and that's an upper bound. Even the degenerate extreme (2 ranks, 3 batches: weights 1/3, 2/3 vs 1/2, 1/2, i.e. at most 1/6 of the between-rank difference) gives ≈ 2.3% against a 5.8% noise floor. (Under `false`, per-key counts are arbitrarily uneven, but there's no old value to compare — the previous per-key `reduce_mean` would hang.) The worked bound is now in the count-weighting section of `pr-plan.md`.

pr-plan.md
```diff
@@ -0,0 +1,221 @@
+# Make evaluating all forward steps configurable per inline-validation entry
```
mcgibbon: Agent: Chore: flagging this plan file for removal before merge in case it's not already on your radar — none of the `pr-plan.md` files on other feature branches have landed on main, so it looks like a branch-side artifact.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -302,7 +302,7 @@ def gather_object(self, obj: T) -> list[T] | None:
 
     def scatter_object(self, obj: T | None) -> T:
         """Scatter a picklable object from the root process to all processes."""
-        if self._data_rank == 0:
+        if self._rank == 0:
```
mcgibbon, line 316: Agent: Suggestion (optional): this root-rank fix is only covered indirectly, through the new aggregator parallel test. A direct `test_scatter_object` mirroring `fme/core/distributed/parallel_tests/test_gather_object.py` (~15 lines) would pin the global-root contract that `PerStepLossAggregator.get_logs` now depends on, cheaply.

fme/ace/aggregator/loss_metrics.py
```diff
@@ -23,12 +29,33 @@ def record(self, metrics: dict[str, torch.Tensor]) -> None:
...
+        if len(keys) == 0:
+            return {}
+        device = get_device()
+        sums = torch.zeros(len(keys), device=device)
+        counts = torch.zeros(len(keys), device=device)
+        for i, key in enumerate(keys):
+            if key in self._sums:
+                sums[i] = self._sums[key].to(device)
+                counts[i] = self._counts[key]
+        sums = dist.reduce_sum(sums)
```
mcgibbon, line 42: Agent: Suggestion (optional, follow-up): `PerChannelLossAggregator` below keeps the old per-locally-observed-key `reduce_mean` — rank-mean-of-means weighting under unequal batch counts, and a hang if a rank records nothing. It's lower stakes (training-side only, where key sets are dense and counts near-equal), but this file now carries two different reduction definitions side by side; the same count-weighted rework would unify them in a follow-up.

fme/ace/train/train_config.py
```diff
@@ -73,6 +73,19 @@ class InlineValidationConfig:
             "val_0", changing its wandb keys and output directory.
         weight: weight for this validation's loss in the combined checkpoint
             selection metric. Must be non-negative.
+        evaluate_all_steps: if True (default), evaluate every forward step in
```
mcgibbon: Agent: Note (no action needed): with a plain-probabilities schedule, `LossSchedule.init_for_epoch(None)` early-returns without building samplers, so a validation batch carrying `epoch=None` would make `evaluate_all_steps: false` silently behave as dense evaluation — no cost saving, and nothing raises. Real loaders stamp the epoch, so this is theoretical; just noting that the new flag makes that pre-existing path user-adjacent for the first time.

mcgibbon (commented):
Agent: **This is an AI review**, posted for the author's reference and benefit. All comments are optional — no verdict is expressed or implied.

General observations from the review:

- The gather/scatter key-universe agreement in `PerStepLossAggregator` is a clean design — no contiguity assumptions, and it handles `loss_step_N` and coupled `loss/{realm}_step_{k}` uniformly. The parallel test is well-differentiated: its expected values (e.g. step-0 mean 2/3) differ from what the old mean-of-rank-means would produce (0.5), so it catches weighting regressions and not just hangs, and the `empty_rank` case pins the deadlock fix. The deliberate choice to accept a small, well-bounded metric shift rather than fork a legacy reduction path reads as the right call, and the error-bound analysis in the description is exactly what's needed to sign off on it.
- Note: the post-epoch train-data evaluation (`trainer.py`, `train_on_batch(..., evaluate_all_steps=True)` in `train_one_epoch`) intentionally keeps dense evaluation — the plan file states this scoping, but that file is presumably being deleted before merge. A sentence in the PR description would let the scope decision survive.

pr-plan.md
jpdunc23: Agreed, plan is to delete before merge but leaving in for the time being for the benefit of reviewers (if they choose to look at it).

fme/core/distributed/model_torch_distributed.py
jpdunc23: Let's add this test.

fme/ace/train/train_config.py
jpdunc23: Would a defensive raise be cheap to add?

fme/ace/aggregator/loss_metrics.py
jpdunc23: Sounds like a good follow-up. Would it be cheap to do it now or add a lot to the diff?

fme/core/distributed/model_torch_distributed.py
jpdunc23: Added in 73e696886: `fme/core/distributed/parallel_tests/test_scatter_object.py` mirrors `test_gather_object` — the root scatters a list and every rank asserts it receives it. Passes under both the `torch` and spatial-parallel `model` (2×1) backends via torchrun.

fme/ace/aggregator/loss_metrics.py
jpdunc23: Mechanically cheap — `PerChannelLossAggregator.record` already keeps weighted sums and counts, so the same universe-agreement + dense `reduce_sum` rework is ~20 lines plus a small parallel test, and the two aggregators could share a helper. Doing it as a follow-up PR instead: it extends the count-weighted metric-definition change to the training-side per-channel losses, widening what this PR's description has to explain, while the hang risk it guards against (a rank recording nothing) doesn't arise on that path today.

fme/ace/train/train_config.py
jpdunc23: Was cheap — added in 73e696886. `init_for_epoch(None)` now raises `EpochNotProvidedError` for any non-constant schedule, not just milestone schedules, so the silent dense-evaluation path is a hard failure. Constant schedules (fixed-int `n_forward_steps`, or a single-outcome probabilities entry, no milestones) still accept `epoch=None` via a new `TimeLengthSchedule.is_constant` property — there the epoch-less fallback (evaluate the full window) is identical to sampled behavior anyway. Tests cover both sides.

fme/ace/aggregator/loss_metrics.py
```diff
@@ -6,7 +6,13 @@
 
 
 class PerStepLossAggregator:
-    """Accumulates per-step loss metrics across batches and produces means."""
+    """Accumulates per-step loss metrics across batches and produces means.
+
+    The distributed reduction must not depend on locally-observed keys, since
```
mcgibbon: I can't tell what this sentence means, please clean it up or delete it.

fme/ace/aggregator/loss_metrics.py
```diff
@@ -23,12 +29,32 @@ def record(self, metrics: dict[str, torch.Tensor]) -> None:
 
     def get_logs(self, label: str) -> dict[str, float]:
         dist = Distributed.get_instance()
+        # Every rank must issue an identical sequence of collective calls, so
```
mcgibbon: Non-blocking: AI text-density needs to be revised or deleted here.

fme/ace/train/train_config.py
```diff
@@ -73,6 +73,18 @@ class InlineValidationConfig:
...
```
mcgibbon: Chore: revise AI text density

fme/coupled/train/train_config.py
```diff
@@ -109,6 +109,19 @@ class InlineValidationConfig:
             "val_0", changing its wandb keys and output directory.
         weight: weight for this validation's loss in the combined checkpoint
             selection metric. Must be non-negative.
+        evaluate_all_steps: if True (default), evaluate every forward step in
```
mcgibbon: Chore: revise AI text density, perhaps base on the existing human-written docstring?

fme/core/generics/validation.py
```diff
@@ -28,6 +28,7 @@ def run_validation_loop(
     validate_using_ema: bool = False,
     compute_derived_variables: bool = True,
     log_progress: bool = False,
+    evaluate_all_steps: bool = True,
```
mcgibbon, line 31: For another time/PR, but I notice many of these variables are defined and documented in two functions in this file, perhaps we should compartmentalize them into a dataclass/simplenamespace later.

mcgibbon (approved):
LGTM

PR #1386, opened 2026-07-23
Fix annual aggregator year-completeness threshold

### LINKED ISSUE #1385: Annual aggregator year-completeness threshold is silently zero for timesteps >= 1 day
`_get_min_samples` in `fme/ace/aggregator/inference/annual.py` computes

```python
steps_per_day = datetime.timedelta(days=1) // timestep
return 362 * steps_per_day
```

For timesteps of one day or longer (e.g. the 5-daily ocean emulator), the floor division yields 0, so the filter `counts > min_samples` keeps every year containing even a single sample. Near-empty partial years then appear in annual series and wandb plots — e.g. run `ai2cm/ace-samudra-cm4/7il0bgfc` retains trailing years holding only the rollout's final instant, which land at ~−3.5 W/m² in the implied-OHC-tendency annual series and contaminate raw means. The 6-hourly atmosphere series from the same experiment correctly drops those years, so paired atmosphere/ocean analyses disagree on year counts.

Fix: compute the threshold with true division, e.g. `362 * (datetime.timedelta(days=1) / timestep)` (the strict `>` comparison works fine with a float threshold).

Relatedly, the 362-day cutoff also drops near-complete first years when the initial condition is nudged a few days into a year (~360 days of samples); lowering the cutoff to 350 days alongside the division fix would address both.

jpdunc23:
The annual global-mean inference aggregator keeps a year in the annual mean series only if it has more than a threshold number of samples, but the threshold was computed with floor division of `timedelta(days=1)` by the timestep — silently zero for timesteps of one day or longer. Ocean (5-daily) annual series therefore retained near-empty partial years, including trailing years holding only the rollout's final end-of-interval instant, which land at wildly unphysical values and disagree with the 6-hourly atmosphere series from the same experiment about which years exist. Separately, the 362-day cutoff dropped near-complete first years from rollouts whose initial condition is offset a few days into a year. This PR computes the threshold with true division so it applies at any timestep, and lowers the cutoff to 350 days so offset first years are kept.

Behavior changes: annual series at timesteps of one day or longer now drop partial years (previously all years were kept); annual series now keep years with more than 350 days of samples that the 362-day cutoff previously dropped. Interior full years are unaffected.

Changes:
- `fme.ace.aggregator.inference.annual._get_min_samples`: threshold computed with true (not floor) division of timedeltas; cutoff extracted to a named constant `MIN_COMPLETE_YEAR_DAYS = 350` with the offset-start rationale documented at the definition.

- [x] Tests added

Resolves #1385

### DIFF
--- a/fme/ace/aggregator/inference/annual.py
+++ b/fme/ace/aggregator/inference/annual.py
@@ -408,9 +408,16 @@ def to_dataset(data: TensorMapping, time: xr.DataArray) -> xr.Dataset:
     return xr.Dataset(data_vars, coords={"valid_time": time})
 
 
-def _get_min_samples(timestep: datetime.timedelta) -> int:
-    steps_per_day = datetime.timedelta(days=1) // timestep
-    return 362 * steps_per_day
+# A year is kept in the annual mean series only if it holds more samples than
+# this many days' worth. 350 leaves ~15 days of slack below a full year so that
+# a rollout whose initial condition is nudged a few days into a year keeps its
+# first year, while genuinely partial (e.g. mid-year-start) years are dropped.
+MIN_COMPLETE_YEAR_DAYS = 350
+
+
+def _get_min_samples(timestep: datetime.timedelta) -> float:
+    steps_per_day = datetime.timedelta(days=1) / timestep
+    return MIN_COMPLETE_YEAR_DAYS * steps_per_day
 
 
 @dataclasses.dataclass
--- a/fme/ace/aggregator/inference/annual.py
+++ b/fme/ace/aggregator/inference/annual.py
@@ -410,7 +410,7 @@ def to_dataset(data: TensorMapping, time: xr.DataArray) -> xr.Dataset:
 
 # A year is kept in the annual mean series only if it holds more samples than
 # this many days' worth. 350 leaves ~15 days of slack below a full year so that
-# a rollout whose initial condition is nudged a few days into a year keeps its
+# a rollout whose initial condition is offset a few days into a year keeps its
 # first year, while genuinely partial (e.g. mid-year-start) years are dropped.
 MIN_COMPLETE_YEAR_DAYS = 350
 
--- a/fme/ace/aggregator/inference/test_annual.py
+++ b/fme/ace/aggregator/inference/test_annual.py
@@ -393,6 +393,108 @@ def test_annual_aggregator():
     assert isinstance(logs["test/a"], plt.Figure)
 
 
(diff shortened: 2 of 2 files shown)


fme/ace/aggregator/inference/test_annual.py
```diff
@@ -393,6 +393,108 @@ def test_annual_aggregator():
...
+    # year, which must be dropped.
+    years = _get_kept_years(
+        cftime.DatetimeNoLeap(2000, 7, 1), FIVE_DAILY_TIMESTEP, n_time=37 + 73
+    )
+    assert years == [2001]
+
+
+def test_annual_aggregator_keeps_nudged_start_year_six_hourly():
+    # a rollout starting a few days into the year (e.g. a nudged initial
+    # condition) covers 360 of 365 days of its first year, which is kept.
```
jpdunc23: What do you mean by "nudged" here? Nudging has a specific technical meaning (e.g., in reanalysis) so this is somewhat confusing. Maybe "offset" would be better?

jpdunc23: Agreed — renamed throughout to "offset" in 992f8fe: the test is now `test_annual_aggregator_keeps_offset_start_year_six_hourly`, its comment says "an initial condition offset to skip leading invalid data", and the rationale comment on `MIN_COMPLETE_YEAR_DAYS` in `annual.py` now reads "offset a few days into a year".

fme/ace/aggregator/inference/annual.py
```diff
@@ -408,9 +408,16 @@ def to_dataset(data: TensorMapping, time: xr.DataArray) -> xr.Dataset:
...
 
 
-def _get_min_samples(timestep: datetime.timedelta) -> int:
-    steps_per_day = datetime.timedelta(days=1) // timestep
-    return 362 * steps_per_day
+# A year is kept in the annual mean series only if it holds more samples than
+# this many days' worth. 350 leaves ~15 days of slack below a full year so that
+# a rollout whose initial condition is offset a few days into a year keeps its
+# first year, while genuinely partial (e.g. mid-year-start) years are dropped.
+MIN_COMPLETE_YEAR_DAYS = 350
```
jpdunc23, line 415: I lowered this from `362` to `350` to avoid throwing away the first year of ACE outputs in coupled jobs where the ocean's first timestamp is Jan 6, not Jan 1. As a result, the overall initial condition is Jan 6 and we throw away atmosphere obs for Jan 1 - Jan 5.

spencerkclark (approved):
Good catch @jpdunc23, thanks.  This looks good to me.

PR #1387, opened 2026-07-24
Remove just added `--segment-label-format` option

spencerkclark:
After further discussion following #1382 we decided it would be best to keep the directory structure of segmented runs as predictable as possible.  We therefore remove the option to change the format. For now we lock in the hourly precision timestamp for better human readability, but we could consider changing that in the future if a need ever comes up for a model with a sub-hourly timestep or if it ever makes sense to start from initial conditions misaligned from the ACE timestep.

Changes:
- Removes the just added `--segment-label-format` option from the `inference` entrypoint.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/__main__.py
+++ b/fme/ace/inference/__main__.py
@@ -1,7 +1,7 @@
 from fme.core.cli import get_parser
 from fme.core.distributed.distributed import Distributed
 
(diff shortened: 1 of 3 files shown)


oliverwm1 (approved):
LGTM

spencerkclark:
Thanks @oliverwm1

Issue #579, opened 2025-03-07
Unordered Lists cast to/from Set in stepper code


mcgibbon:
In our stepper code we currently have many casts of list to set and back in order to satisfy `List[str]` type constraints, even though these lists are unordered and could be typed as Set. If we change these to Set return types, we can avoid this casting boilerplate and make it clearer that the values are not ordered.

Note this only applies to actually-unordered lists, our base config values and values derived from them by concatenation should still be list (e.g. to preserve ML channel orders).

FaresIbrahim32:
hello My name is Fares Ibrahim . I am a student from codepath and I wanted to work on this

mcgibbon:
To fully close the issue we will need to fix these occurrences in the StepABC and its implementations, which I was considering as "stepper" code when writing this issue.

PR #1391, opened 2026-07-26
Type unordered stepper names as set[str]

FaresIbrahim32:
## Summary

This PR updates unordered stepper name properties to use `set[str]` instead of `list[str]` and removes the redundant casts that were only there to satisfy the old typing.

## Why

The affected stepper properties represent unordered collections of names. Typing them as lists was misleading and forced unnecessary `list`/`set` conversions at several call sites.

## Changes

- Re-typed the unordered stepper name properties as `set[str]`.
- Removed redundant casts around those values.
- Kept order-sensitive list values unchanged.
- Added a regression test for the new typing behavior.

## Related issue

Related to #579 (partial - `StepABC` and its implementations still have unordered `list[str]` values that should also become `set[str]` to fully close out the issue, per @mcgibbon's comment below)


mcgibbon:
PR looks good and can merge. It doesn't close the issue because I didn't make something clear in writing that issue - the StepABC and its implementations also use many list[str] that should be set[str] since they're unordered. Those will also need to be translated over to fully close it out. Nonetheless, this PR does what it says and it's an improvement. Please remove the "Resolves" line from the PR description.

FaresIbrahim32:
Thanks for the review and for clarifying the scope, @mcgibbon! I've updated the PR description to say "Related to #579" instead of "Resolves #579" so it won't auto-close the issue on merge.

Happy to take a follow-up pass on `StepABC` and its implementations to fully close out #579 if that's useful - let me know if you'd like that as a separate PR or added to this one.

mcgibbon (approved):
I’ve deleted the dev notes. LGTM! Thank you.

PR #1394, opened 2026-07-29
Add translate named data streams and paired-by-time multi-resolution loading

mcgibbon:
PR 2 of the `fme/translate` series, targeting the long-lived base branch
`feature/translate-skeleton` (PR #1360) rather than `main`.

Translate training reads several named data streams at once — the same state at
1°/2°/4°, or ERA5 alongside SHiELD. Some of those streams must be sampled at the
same valid times (a translation objective comparing 1° and 2° fields of one state
is meaningless if the two are drawn from different times) and some must not be
(the transfer-learning program's two streams are unpaired distributions). This
adds `fme/translate/data/`: named streams, that pairing, and the multi-stream
`GriddedDataABC` implementation training will consume.

Whether two streams are sampled together is **derived, never configured**. Each
objective declares what it needs from each stream it consumes; every objective
ties its streams together, and the connected components of the resulting graph
are the *pairing groups*. Streams in a group share a time index; groups are
sampled independently. Grouping is transitive, so objectives over (a, b) and
(b, c) put all three on one index. A `sampling:` knob in the config could
contradict the objectives; a derivation cannot.

The requirements type is the seam PR 3 (objectives) will produce: it constructs
one `ObjectiveDataRequirements` per objective and calls
`TranslateDataRequirements.from_objectives`, which merges per-stream variable
names as an ordered union and per-stream window length as the *pointwise maximum*
of the objectives' `IntSchedule`s (ace loads a window of the scheduled length and
each objective samples a prefix of it, so loading the longest window any objective
asks for serves them all). Nothing in the seam depends on the component pool: the
loader takes variable names from the requirements, not from the pool's domain
channel lists.

Alignment within a group is validated when the group is built, since pairing by
index is otherwise silently wrong: the streams must share a timestep, and their
sample start times must match index by index, with the first offending index and
its two times named in the error. A group's length is the minimum over its
streams, because objectives asking for different window lengths leave different
numbers of valid start times.

`n_samples`/`n_batches`/seeding semantics under unequal groups: `n_batches` is the
minimum over the groups (iteration stops with the shortest, so no group is
re-drawn mid-epoch), and `n_samples` is `n_batches * batch_size` — samples drawn
*per group*, not summed across groups, so it stays the count a per-batch metric
divides by. Each group's sampler is built with its own `seed_offset`; without
that, two equal-length groups receive the same permutation and are index-paired
in fact while claiming to be independent (a test fails if the offset is removed).

`TranslateGriddedData.dataset_info` is keyed by **domain**, not stream: that is
the pairing key `ComponentPoolConfig.build` consumes, and streams sharing a
domain are checked to describe compatible datasets via
`DatasetInfo.assert_compatible_with`, with their variable metadata and labels
unioned.

Changes:
- `fme.translate.data.requirements`: `StreamRequirements`,
  `ObjectiveDataRequirements`, and `TranslateDataRequirements.from_objectives` —
  the objective-to-data seam and its merge (name union, pointwise-max schedule,
  pairing groups).
- `fme.translate.data.config`: `StreamConfig` (`domain`, `dataset`, optional
  `name` defaulting to the domain via `stream_name`) and
  `TranslateDataLoaderConfig`, mirroring `fme.ace.data_loading.config.DataLoaderConfig`'s
  field names and validating in `__post_init__`.
- `fme.translate.data.batch_data`: `TranslateBatchData`, a
  `Mapping[str, BatchData]` generalizing `fme.coupled`'s two-component batch
  type, plus `TranslateCollateFn`.
- `fme.translate.data.dataset`: `PairedStreamDataset`, an N-stream
  `GenericDataset` that pairs by index and validates timestep and start-time
  alignment at build time.
- `fme.translate.data.dataloader`, `.gridded_data`, `.getters`:
  `TranslateDataLoader` (one per pairing group, all behavior from
  `GenericDataLoader`), `TranslateGriddedData` (zips the groups, publishes
  domain-keyed `dataset_info`), and `get_gridded_data(config, requirements, train)`.
- `fme.core.distributed.Distributed.get_sampler`: new `seed_offset` argument
  (default 0, existing callers unaffected) so several samplers over equal-length
  datasets can be shuffled independently rather than in lockstep. Added here
  rather than mutating the returned sampler so the spatial-parallel `num_replicas`
  and `rank` logic stays in one place.
- `fme/translate/examples/`: the README table and the two configs' headers now
  say which blocks parse — `train_data:`/`validation_data:` with this PR, and
  `component_pool:` only once PR 1's remaining rework (channels-on-transforms
  with derived domain load lists, int-declared latent blocks, identity
  normalization) lands, which is tracked separately and is why the loader takes
  variable names from the requirements seam rather than the pool config.

Design points worth a reviewer's attention:
- `PairedStreamDataset` is translate-local rather than a reuse of
  `fme.downscaling`'s `FineCoarsePairedDataset`, which is hard-wired to a
  fine/coarse pair and to single-timestep items (its `BatchItemDatasetAdapter`
  squeezes the time dimension away) and so cannot carry the time windows a
  forward-prediction objective needs. `fme.coupled.CoupledDataset` is the closer
  analogue but is likewise fixed at two named components. What is reused is the
  machinery under both: `GenericDataset`/`GenericDataLoader` supply subsetting,
  epoch propagation, shuffle control, `batch_size` and `n_samples`.
- Unequal but integer-ratio timesteps within a group (`fme.coupled`'s
  `n_steps_fast` index scaling) is a deliberate non-feature; the docstring names
  it as the path if an objective ever needs it.
- Batches are moved to the device but not spatially scattered, following
  `fme.coupled`: spatial model parallelism would need each stream scattered
  against its own grid, which no translate module supports yet. Data-parallel
  sharding is handled. `get_gridded_data` calls
  `Distributed.require_no_spatial_parallelism`, so a spatially-parallel run
  fails rather than handing every co-rank the whole global field.
- `DataRequirements.allow_missing_variables` is refused rather than ignored:
  `StreamConfig.get_dataset` does not plumb it to `DatasetABC.build`, so a
  requirements object carrying it would load every variable and report success.
- ace's `time_buffer` window-reuse options are not offered: they draw several
  output batches from one preloaded window, which would break the index pairing.
- A configured stream that no objective consumes is an error, not a silent
  unused load.

- [x] Tests added

### DIFF
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -205,13 +205,27 @@ class Distributed:
         dataset: torch.utils.data.Dataset,
         shuffle: bool,
         drop_last: bool = False,
+        seed_offset: int = 0,
     ) -> torch.utils.data.Sampler:
+        """
+        Get a sampler over ``dataset`` for this rank.
+
+        Args:
+            dataset: The dataset to sample from.
+            shuffle: Whether to shuffle the samples.
+            drop_last: Whether to drop the incomplete tail of the dataset.
+            seed_offset: Added to the distributed seed. Two samplers built with
+                different offsets produce different permutations of
+                equal-length datasets in the same epoch; leave it at 0 unless
+                sampling several datasets that must be shuffled independently
+                of one another rather than in lockstep.
+        """
         return torch.utils.data.DistributedSampler(
             dataset,
             shuffle=shuffle,
             num_replicas=self._distributed.total_data_parallel_ranks,
             rank=self._distributed.data_parallel_rank,
-            seed=self._seed,
+            seed=self._seed + seed_offset,
             drop_last=drop_last,
         )
 
--- a/fme/translate/data/batch_data.py
+++ b/fme/translate/data/batch_data.py
@@ -1,0 +1,146 @@
+"""A batch of named data streams, and its collate function.
+
+:class:`TranslateBatchData` generalizes
+:class:`fme.coupled.data_loading.batch_data.CoupledBatchData` from two fixed
+components to an arbitrary set of named streams: where the coupled type has
+``ocean_data`` and ``atmosphere_data`` fields and fans every method out over
+both, this one holds a ``dict[str, BatchData]`` and fans out over its keys.
+
+The surface is deliberately narrower than the coupled type's — per-stream
+access, device movement, and the ``epoch`` a trainer needs. Time-window
+manipulation (``prepend``, ``get_start``, ``remove_initial_condition``) is done
+per stream by the objectives, which know which of their streams is the input,
+the target, and the forcing; hoisting those onto a whole-batch fan-out would
+apply them to streams that should not receive them.
+"""
+
+import dataclasses
+from collections.abc import Iterator, Mapping, Sequence
+
+from fme.ace.data_loading.batch_data import BatchData
+from fme.core.dataset.dataset import DatasetItem
+from fme.core.labels import LabelEncoding
+
+__all__ = ["TranslateBatchData", "TranslateCollateFn"]
+
+
+@dataclasses.dataclass
+class TranslateBatchData:
+    """A batch holding one :class:`BatchData` per named data stream.
+
+    Parameters:
+        streams: The batch's per-stream data, keyed by stream name.
+    """
+
+    streams: dict[str, BatchData]
+
+    def __post_init__(self):
+        if not self.streams:
+            raise ValueError("A TranslateBatchData must hold at least one stream.")
+        epochs = {name: batch.epoch for name, batch in self.streams.items()}
+        if len(set(epochs.values())) > 1:
+            raise ValueError(
+                "All streams in a batch must carry the same epoch (they are "
+                f"drawn in step by the trainer), got {epochs}."
+            )
+
+    @property
+    def epoch(self) -> int | None:
+        """The epoch every stream in this batch was drawn in.
+
+        Consumed by ace's ``LossSchedule.init_for_epoch``, which needs the epoch
+        of the data rather than of the trainer loop.
+        """
+        return next(iter(self.streams.values())).epoch
+
+    def __getitem__(self, name: str) -> BatchData:
+        return self.streams[name]
+
+    def __contains__(self, name: str) -> bool:
+        return name in self.streams
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self.streams)
+
+    def __len__(self) -> int:
+        return len(self.streams)
+
+    def to_device(self) -> "TranslateBatchData":
+        return TranslateBatchData(
+            streams={name: batch.to_device() for name, batch in self.streams.items()}
+        )
+
+    def to_cpu(self) -> "TranslateBatchData":
+        return TranslateBatchData(
+            streams={name: batch.to_cpu() for name, batch in self.streams.items()}
+        )
+
+    def pin_memory(self) -> "TranslateBatchData":
+        """Page-lock every stream's tensors; called by torch's DataLoader."""
+        self.streams = {
+            name: batch.pin_memory() for name, batch in self.streams.items()
+        }
+        return self
+
+    @classmethod
+    def merge(cls, batches: Sequence["TranslateBatchData"]) -> "TranslateBatchData":
+        """Combine batches over disjoint stream sets into one.
+
+        Used to assemble the independently-sampled pairing groups' batches into
+        the single batch a training step sees.
+        """
+        streams: dict[str, BatchData] = {}
+        for batch in batches:
+            overlap = sorted(set(batch.streams) & set(streams))
+            if overlap:
+                raise ValueError(
+                    f"Cannot merge batches sharing the streams {overlap}; each "
+                    "stream belongs to exactly one pairing group."
+                )
+            streams.update(batch.streams)
+        return cls(streams=streams)
+
+
+class TranslateCollateFn:
+    """Collates per-stream samples into a :class:`TranslateBatchData`.
+
+    One instance serves one pairing group: its keys are that group's streams,
+    and it is called with the group's paired samples (see
+    :class:`fme.translate.data.dataset.PairedStreamDataset`). Defined at module
+    level so it can be pickled to data-loader worker processes.
+    """
+
+    def __init__(
+        self,
+        horizontal_dims: Mapping[str, list[str]],
+        label_encodings: Mapping[str, LabelEncoding | None],
+    ):
+        """
+        Args:
+            horizontal_dims: Each stream's horizontal dimension names, used
+                when writing batches to netCDF.
+            label_encodings: Each stream's label encoding, or None for a stream
+                whose dataset provides no labels.
+        """
+        if set(horizontal_dims) != set(label_encodings):
+            raise ValueError(
+                "horizontal_dims and label_encodings must cover the same "
+                f"streams, got {sorted(horizontal_dims)} and "
+                f"{sorted(label_encodings)}."
+            )
+        self.horizontal_dims = dict(horizontal_dims)
+        self.label_encodings = dict(label_encodings)
+
+    def __call__(
+        self, samples: Sequence[Mapping[str, DatasetItem]]
+    ) -> TranslateBatchData:
+        return TranslateBatchData(
+            streams={
+                name: BatchData.from_sample_tuples(
+                    [sample[name] for sample in samples],
+                    horizontal_dims=dims,
+                    label_encoding=self.label_encodings[name],
+                )
+                for name, dims in self.horizontal_dims.items()
+            }
+        )
--- a/fme/translate/data/batch_data.py
+++ b/fme/translate/data/batch_data.py
@@ -25,8 +25,14 @@ __all__ = ["TranslateBatchData", "TranslateCollateFn"]
 
 
 @dataclasses.dataclass
-class TranslateBatchData:
+class TranslateBatchData(Mapping[str, BatchData]):
     """A batch holding one :class:`BatchData` per named data stream.
+
+    A read-only ``Mapping`` from stream name to that stream's data, since that
+    is how objectives will reach their streams: ``batch["era5_1deg"]``,
+    ``batch.items()``, ``**batch``. Inheriting ``Mapping`` rather than
+    hand-rolling the surface keeps ``keys``/``values``/``items``/``get`` in
+    agreement with ``__getitem__``.
 
     Parameters:
         streams: The batch's per-stream data, keyed by stream name.
@@ -56,9 +62,6 @@ class TranslateBatchData:
     def __getitem__(self, name: str) -> BatchData:
         return self.streams[name]
 
-    def __contains__(self, name: str) -> bool:
-        return name in self.streams
-
     def __iter__(self) -> Iterator[str]:
         return iter(self.streams)
 
@@ -76,7 +79,14 @@ class TranslateBatchData:
         )
 
     def pin_memory(self) -> "TranslateBatchData":
-        """Page-lock every stream's tensors; called by torch's DataLoader."""
+        """Page-lock every stream's tensors *in place*, and return self.
+
+        Deliberately unlike ``to_device``/``to_cpu``, which return copies:
+        ``BatchData.pin_memory`` and ``fme.coupled``'s both mutate and return
+        self, because torch's DataLoader calls this on a batch it has just
+        collated and is about to hand over, so no other holder of the un-pinned
+        tensors can be surprised. Do not call it on a batch someone else keeps.
+        """
         self.streams = {
             name: batch.pin_memory() for name, batch in self.streams.items()
         }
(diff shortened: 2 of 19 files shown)


mcgibbon (commented):
Claude: pre-review pass (independent agent, no authoring context). Verdict: **ready for review after the four fixes I pushed** — 58c0a7a. I found no correctness bug in the pairing, alignment, merge or sizing logic; the things I changed are a silently-wrong path under spatial parallelism, two requirements that were silently ignored, one unreachable error branch, and a test locking in an invariant the design depends on. Details in the thread comments; two design questions I deliberately left for Jeremy rather than settling here.

What I verified rather than took on trust:

- **`seed_offset` placement and back-compat.** Every existing caller (`fme/ace/data_loading/getters.py:71`, `fme/downscaling/data/config.py`, the two test modules) passes no offset and gets `seed=self._seed` exactly as before. The stated reason for not offsetting the epoch instead holds, and is in fact stronger than the description says: `GenericDataLoader.set_epoch` calls `self._dataset.set_epoch(epoch)`, which reaches `XarrayDataset._global_epoch`; that feeds both `DatasetItem`'s epoch (→ `BatchData.epoch` → `LossSchedule.init_for_epoch` at `fme/ace/stepper/single_module.py:1632`) **and** `_ensure_epoch_synchronized`'s `_n_timesteps_schedule.get_value(epoch)`. So an epoch offset would corrupt the loss schedule *and* shift the window-length schedule per group. Offsetting the seed is right.
- **Independent-group sampling.** The independence test is load-bearing: patching `seed_offset=group_index` to `seed_offset=0` in `getters.py` makes `test_independently_sampled_streams_are_not_time_locked` fail (`assert not True`), everything else still passing. On the RandomSampler question — there is no such path here. `Distributed.get_sampler` always returns a `DistributedSampler`, and translate always goes through it, so the mechanism is uniform serially and distributed (ace only reaches `RandomSampler` via `sample_with_replacement`, which this config deliberately does not expose). That also means `set_epoch`/`alternate_shuffle`'s `isinstance(..., DistributedSampler)` branches are always taken.
- **Pointwise-max schedule.** Correct at every epoch, including 0 and epochs between milestones: each input schedule is piecewise-constant with breakpoints only at its own milestones, so the max is piecewise-constant with breakpoints only in the union — evaluating at the union epochs and holding between them is exact. `ValidatedMilestones` rejects a milestone at epoch ≤ 0, so `start_value = max(get_value(0))` is not shadowed by an epoch-0 milestone.
- **Group length is not off by one.** `XarrayDataset._n_initial_conditions = total_timesteps - n_timesteps.max_value + 1`, so `min(len(dataset))` over a group is exactly the shared valid-start count (8 times / window 4 → 5, matching the test), and the alignment prefix compare is over equal-length slices. Alignment and timestep are validated in `PairedStreamDataset.__init__`, which runs before the loader exists, so no sample can be drawn misaligned.
- **`GriddedDataABC` conformance.** All six abstract members implemented; `n_samples`/`n_batches`/`batch_size` match `fme.ace`'s `GriddedData` semantics (local batch size, `n_batches * batch_size`). `TranslateDataLoader` adds no `__init__` args, so `GenericDataLoader.subset`'s `self.__class__(dataset=, sampler=, collate_fn=, **kwargs)` re-instantiation is safe — and `subset` freezes each group's sampler order independently, which keeps within-group pairing while leaving groups independent.
- Tests pass (79 in `fme/translate`, 479 in `fme/core/distributed` + `fme/core/test_rand.py` + `fme/ace/data_loading`), `pre-commit run --files` clean on every changed file.

Fixes pushed (each its own commit):

1. `9ff7dbb` — the duplicate-pairing-group check in `TranslateDataRequirements.__post_init__` was unreachable (a stream in two groups also fails the partition check, which ran first and misreported it), and `DataRequirements.allow_missing_variables` was silently dropped by the loader. Reordered, and refuse the flag.
2. `84ddd70` — `require_no_spatial_parallelism` guard in `get_gridded_data`.
3. `4a0c199` — test that a scheduled window keeps the group's start-time count.
4. `58c0a7a` — `num_data_workers` is per pairing group; drop a docstring guarantee that isn't checked.

Revert any of these freely if you disagree — 2 in particular is a behavior change I judged worth making, not an obvious one.

Two things I think are Jeremy's call, not mine, both flagged in threads below: whether `n_batches = min` over groups is the right definition of an epoch when groups are very unequal, and whether an unconsumed configured stream should error.

fme/core/distributed/distributed.py
```diff
@@ -205,13 +205,27 @@ def get_sampler(
...
+                sampling several datasets that must be shuffled independently
+                of one another rather than in lockstep.
+        """
         return torch.utils.data.DistributedSampler(
             dataset,
             shuffle=shuffle,
             num_replicas=self._distributed.total_data_parallel_ranks,
             rank=self._distributed.data_parallel_rank,
-            seed=self._seed,
+            seed=self._seed + seed_offset,
```
mcgibbon, line 228: Claude: this is the right place — the alternative (mutating the returned sampler's `seed`) would either duplicate the `num_replicas`/`rank` logic or reach into `DistributedSampler` internals, and adding a full `seed=` override would break `set_seed`'s single point of control. Back-compatible for all four existing call sites.

One property worth a line of comment, because it is easy to over-claim from the docstring: `DistributedSampler.__iter__` seeds with `self.seed + self.epoch`, so offset *k* at epoch *e* yields the same permutation as offset 0 at epoch *e+k*. Within any single epoch the offsets differ and the groups are genuinely independent, which is the thing that matters here (and `alternate_shuffle`'s `alternate_seed(epoch)` keeps them distinct too). But the offsets are a shift along one seed axis, not independent streams, so nothing later should rely on cross-epoch independence between two groups. If you want that property outright, `self._seed + seed_offset * <large stride>` gets it for free.

fme/translate/data/gridded_data.py
```diff
@@ -0,0 +1,239 @@
...
+    the groups: a group with fewer batches would otherwise have to be re-drawn
+    mid-epoch, correlating its samples within the epoch.
+    """
+
+    def __init__(self, loaders: Sequence[DataLoader[TranslateBatchData]]):
+        if not loaders:
+            raise ValueError("At least one pairing group's loader is required.")
+        self._loaders = list(loaders)
+
+    def __len__(self) -> int:
```
mcgibbon, line 38: Claude: **for Jeremy, not settled in pre-review.** `min` over groups makes the epoch as long as the *shortest* pairing group. The reasoning given (a shorter group re-drawn mid-epoch correlates its samples within the epoch) is right, but the cost lands squarely on the transfer-learning program, where the two streams are independent groups with no reason to be the same length: if ERA5 has 10x the samples of SHiELD, an "epoch" touches a tenth of ERA5, and everything keyed to epochs (LR schedule, `n_forward_steps` milestones, `max_epochs`, validation cadence) is now denominated in the short stream. Coverage over many epochs is fine, since each epoch reshuffles — this is about what "epoch" means, not about wasted data.

The alternatives are all worse in their own way (cycling the short group correlates it; padding it duplicates samples; per-group epoch counters make the schedule ambiguous), so I don't think `min` is wrong. But it deserves to be a decision you made rather than a consequence you find out about when the first transfer-learning run's LR schedule finishes 10x early. Worth at least a line in the config docstring telling the user to size the streams comparably.

fme/translate/data/getters.py
```diff
@@ -0,0 +1,171 @@
...
+    unknown = sorted(required - configured)
+    if unknown:
+        raise ValueError(
+            f"The objectives reference data streams that are not configured: "
+            f"{unknown}. Configured streams are {sorted(configured)}."
+        )
+    unused = sorted(configured - required)
+    if unused:
+        raise ValueError(
+            f"These data streams are configured but no objective consumes them: "
```
mcgibbon, line 45: Claude: on the debatable one — I think the error is right, and for a stronger reason than "don't silently load unused data". Since pairing is *derived* from the objectives, a stream no objective names has no group, so the failure mode of tolerating it is not a wasted read, it's a stream that is configured, never sampled, and never mentioned again. The overwhelmingly likely cause is a typo in a stream name or an objective edited without its stream, and both are exactly what you want to fail on. It is also the easy direction to relax later: turning an error into a warning breaks nobody.

The one case to keep an eye on is PR 3 introducing a train-only objective, since `train_data:` and `validation_data:` are validated against the same requirements — a stream only a train-side objective consumes would still have to appear in `validation_data:`. That's a fine constraint today (both example configs mirror their stream lists) but it is worth remembering when the objective list grows a train/validation distinction.

fme/translate/data/gridded_data.py
```diff
@@ -0,0 +1,239 @@
...
+        horizontal_coordinates=properties.horizontal_coordinates,
+        vertical_coordinate=properties.vertical_coordinate,
+        spatial_mask_provider=properties.spatial_mask_provider,
+        timestep=properties.timestep,
+        variable_metadata=properties.variable_metadata,
+        all_labels=properties.all_labels,
+    )
+
+
+def _domain_dataset_info(
```
mcgibbon, line 57: Claude: two things about comparing every stream against `entries[0]` rather than pairwise or cumulatively.

`assert_compatible_with` skips a field entirely when *either* side is `None` (`fme/core/dataset_info.py:100-135`). So if the first stream has `infer_timestep: false` (`timestep=None`) or a `NullVerticalCoordinate`, that field goes unchecked against every other stream serving the domain, and streams 2 and 3 are never compared to each other — three streams can pass while two of them disagree. Then the returned `DatasetInfo` takes `horizontal_coordinates`, `vertical_coordinate`, `spatial_mask_provider` and `timestep` from the reference alone, so the domain can be published with `timestep=None` even though the streams that were compatible-by-omission all carry one.

Not reachable from either example config (both name real datasets with inferred timesteps), so I left it. If you want it airtight, compare each new stream against the accumulated `info` and rebuild `info` as you go, rather than holding the reference fixed.

fme/translate/data/requirements.py
```diff
@@ -0,0 +1,220 @@
...
+
+    names: list[str]
+    n_timesteps: int | IntSchedule = 1
+
+    def __post_init__(self):
+        if not self.names:
+            raise ValueError("StreamRequirements requires at least one variable name.")
+
+    @property
+    def n_timesteps_schedule(self) -> IntSchedule:
```
mcgibbon, line 56: Claude: nit — this constructs a whole `DataRequirements` (and copies `names`) to read a two-branch property. `IntSchedule.from_constant` is right there, and the union is normalized in one place either way:

```python
if isinstance(self.n_timesteps, IntSchedule):
    return self.n_timesteps
return IntSchedule.from_constant(self.n_timesteps)
```

The delegation reads as if it were load-bearing (the comment says it is) when what it buys is avoiding a duplicated three-line branch, at the cost of a reader chasing into `fme.ace.requirements` to find out that's all it does. Your call; I didn't want to churn the seam.

fme/translate/data/dataset.py
```diff
@@ -0,0 +1,139 @@
...
+                )
+            _assert_start_times_match(
+                reference_name,
+                reference_times,
+                name,
+                dataset.sample_start_times[: self._length],
+            )
+
+    @property
+    def stream_names(self) -> list[str]:
```
mcgibbon: Claude: `stream_names` has no caller and no test. Its natural use would have been letting `TranslateGriddedData` check that its group loaders' streams partition `stream_properties` — the guarantee its docstring claimed — but `GenericDataLoader` doesn't expose its dataset, so the check can't be made from there. I softened that docstring in 58c0a7a instead. Either drop this property or wait for PR 3 to actually need it; as it stands it reads like an API someone forgot to wire up.

Same observation, lower stakes, for `TranslateGriddedData.variable_metadata` and `TranslateBatchData.to_cpu` — no consumer yet. Those two have direct counterparts in `fme.ace`/`fme.coupled` that PRs 3-4 will plausibly want, so forward-declaring them is more defensible.

fme/translate/data/batch_data.py
```diff
@@ -0,0 +1,146 @@
...
+        return TranslateBatchData(
+            streams={name: batch.to_device() for name, batch in self.streams.items()}
+        )
+
+    def to_cpu(self) -> "TranslateBatchData":
+        return TranslateBatchData(
+            streams={name: batch.to_cpu() for name, batch in self.streams.items()}
+        )
+
+    def pin_memory(self) -> "TranslateBatchData":
```
mcgibbon, line 81: Claude: `pin_memory` mutates `self.streams` and returns `self`, while `to_device`/`to_cpu` build a new `TranslateBatchData`. `BatchData`'s own methods are all copy-returning, so the odd one out here is a reader trap even though torch's pin-memory hook doesn't care which you do. Making it `return TranslateBatchData(streams={...})` costs nothing and makes the three read the same.

Separately, on the surface as a whole: this defines `__getitem__`/`__contains__`/`__iter__`/`__len__` with `Mapping` semantics (`__iter__` yields keys) without being a `Mapping`, so `batch.keys()`, `batch.items()` and `**batch` all fail while `sorted(batch)` and `name in batch` work. Either inherit `collections.abc.Mapping[str, BatchData]` and get the rest for free, or drop the dunders and let callers use `.streams` — the half-Mapping is the version a reader has to test to understand. Not blocking; the tests only use the dunders that exist.

PR #1395, opened 2026-07-29
Avoid CUDA context creation in DataLoader workers

yyexela:
Inference dataloaders run with a forkserver multiprocessing context and persistent workers, and their `worker_init_fn` enters a `Distributed` context so each worker can decide which samples belong to its rank. Constructing `TorchDistributed` called `torch.cuda.set_device`, which creates a CUDA context (~520 MiB) in every worker process. A training config with six inline inference loaders at `num_data_workers: 4` therefore leaves 24 idle CUDA contexts holding 12.2 GiB on each GPU, which is enough to turn an otherwise-fitting training step into `torch.OutOfMemoryError` during backward.

DataLoader workers only read `rank` and `world_size`, which `fme.ace.data_loading.inference.InferenceDataset` uses for sharding; they never run collectives and never place tensors on the GPU. Workers now take that metadata from the launcher's environment variables and skip both process group and CUDA device initialization. Objects pickled into the workers hold only CPU tensors, since `DatasetProperties.to_device` returns a new instance and is applied in `GriddedData`/`InferenceGriddedData` rather than on the dataset itself, so no CUDA context is re-created by unpickling.

This covers both `fme.ace.data_loading.getters._forkserver_worker_init_fn` and `fme.coupled.data_loading.getters._coupled_forkserver_worker_init_fn`, since the fix is in the backend they both build.

Changes:
- `fme.core.distributed.torch_distributed._in_dataloader_worker` and `_rank_metadata_from_env`: new helpers that detect a DataLoader worker and read rank, world size and local rank from the torchrun or srun environment variables
- `fme.core.distributed.torch_distributed.TorchDistributed.__init__`: in a DataLoader worker, populate rank metadata from the environment and return without calling `torch.distributed.init_process_group` or `torch.cuda.set_device`

- [x] Tests added

### DIFF
--- a/fme/core/distributed/test_distributed.py
+++ b/fme/core/distributed/test_distributed.py
@@ -3,9 +3,10 @@ import os
 import pytest
 import torch
 import torch.multiprocessing as mp
+import torch.utils.data
 
 from fme import get_device
-from fme.core.distributed import Distributed
+from fme.core.distributed import Distributed, torch_distributed
 from fme.core.distributed.torch_distributed import (
     _gather_irregular,
     _pad_tensor_at_end,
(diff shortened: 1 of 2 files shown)


yyexela:
Superseded by #1396 — same commit, branch renamed `oom_fix` -> `fix/oom_fix` to match the repo naming convention. Closing this one.

PR #1396, opened 2026-07-29
Avoid CUDA context creation in DataLoader workers

yyexela:
Inference dataloaders run with a forkserver context and persistent workers, and their `worker_init_fn` enters a `Distributed` context so each worker knows which samples belong to its rank. Constructing the backend called `torch.cuda.set_device`, creating a ~520 MiB CUDA context in every worker. A training config with six inline inference loaders at `num_data_workers: 4` therefore left 24 idle contexts holding 12.2 GiB on each GPU, enough to turn an otherwise-fitting training step into `torch.OutOfMemoryError` during backward.

Workers only read sharding metadata — `data_parallel_rank` and `total_data_parallel_ranks` in `fme.ace.data_loading.inference.InferenceDataset`, `rank` and `world_size` in the coupled equivalent. They never run collectives and never place tensors on the GPU. They now take that metadata from the launcher's environment variables and skip both process group and CUDA device initialization. Objects pickled into workers hold only CPU tensors, since `DatasetProperties.to_device` returns a new instance and is applied in `GriddedData`/`InferenceGriddedData` rather than on the dataset, so unpickling re-creates no context.

Workers were also joining the process group. That was harmless — torchrun's agent hosts the TCPStore so every rank connects as a client, `_store_based_barrier` is already satisfied when a worker increments the counter, and NCCL communicator creation is lazy without a `device_id`. The CUDA context was the whole cost.

Both `TorchDistributed` and `ModelTorchDistributed` are reachable from a worker, so both are fixed, which covers `fme.ace.data_loading.getters._forkserver_worker_init_fn` and `fme.coupled.data_loading.getters._coupled_forkserver_worker_init_fn`. Under spatial parallelism a worker cannot build the DeviceMesh that normally supplies data-parallel rank and size, so it derives them from the row-major `(data, h, w)` mesh layout; a parallel test checks that arithmetic against a real mesh. Attributes a worker cannot populate — the device id, the spatial process groups, the mesh — are left unset, so using one raises rather than reporting a value that disagrees with the process's actual state.

Changes:
- `fme.core.distributed.torch_distributed._in_dataloader_worker` and `_rank_metadata_from_env`: new helpers that detect a DataLoader worker and read rank and world size from the torchrun or srun environment variables, raising the same "initialized without torchrun or srun" `ValueError` as the non-worker path when neither launcher is detected
- `fme.core.distributed.torch_distributed.TorchDistributed.__init__`: in a worker, populate rank metadata from the environment and return without calling `torch.distributed.init_process_group` or `torch.cuda.set_device`
- `fme.core.distributed.model_torch_distributed.ModelTorchDistributed.__init__`: same guard, additionally skipping `DistributedManager.initialize` and the DeviceMesh, and deriving data-parallel rank and size arithmetically
- `fme.core.distributed.model_torch_distributed._check_world_size_divisible`: new helper, shared by the worker and mesh paths

- [x] Tests added

### DIFF
--- a/fme/core/distributed/model_torch_distributed.py
+++ b/fme/core/distributed/model_torch_distributed.py
@@ -106,7 +106,6 @@ class ModelTorchDistributed(DistributedBackend):
     ):
         spatial_size = h_size * w_size
         if _in_dataloader_worker():
-            # see _in_dataloader_worker; the (data, h, w) mesh is row-major
             self._rank, self._world_size = _rank_metadata_from_env()
             _check_world_size_divisible(self._world_size, spatial_size)
             self._data_size = self._world_size // spatial_size
--- a/fme/core/distributed/model_torch_distributed.py
+++ b/fme/core/distributed/model_torch_distributed.py
@@ -35,12 +35,24 @@ from ._gloo_patch import patch_gloo_alltoall
 from .base import DistributedBackend
 from .external.pnd_manager import DistributedManager
 from .non_distributed import DummyWrapper
-from .torch_distributed import _gather_irregular
+from .torch_distributed import (
+    _gather_irregular,
+    _in_dataloader_worker,
+    _rank_metadata_from_env,
+)
 
 logger = logging.getLogger(__name__)
 
 
 T = TypeVar("T")
+
+
+def _check_world_size_divisible(world_size: int, spatial_size: int) -> None:
+    if world_size % spatial_size != 0:
+        raise ValueError(
+            f"world_size must be divisible by h_size * w_size, "
+            f"got world_size={world_size} and h*w={spatial_size}"
+        )
 
 
 class _AutogradAllReduce(torch.autograd.Function):
@@ -92,18 +104,27 @@ class ModelTorchDistributed(DistributedBackend):
         w_size: int = 1,
         verbose: bool = False,
     ):
+        spatial_size = h_size * w_size
+        if _in_dataloader_worker():
+            # see _in_dataloader_worker for why workers skip process group and
+            # CUDA device initialization, which here also means skipping the
+            # DeviceMesh. The (data, h, w) mesh is row-major over global ranks,
+            # so the data-parallel index and size follow from arithmetic; the
+            # spatial groups, mesh and device id are left unset, so that using
+            # them raises rather than silently misbehaving without a mesh.
+            self._rank, self._world_size = _rank_metadata_from_env()
+            _check_world_size_divisible(self._world_size, spatial_size)
+            self._data_size = self._world_size // spatial_size
+            self._data_rank = self._rank // spatial_size
+            return
+
         # Initialize PhysicsNeMo DistributedManager.
         DistributedManager.initialize()
         self._dm = DistributedManager()
 
         # Build a 3-D (data, h, w) DeviceMesh; data=-1 auto-sizes the
         # data-parallel dimension to absorb all remaining ranks.
-        spatial_size = h_size * w_size
-        if self._dm.world_size % spatial_size != 0:
-            raise ValueError(
-                f"world_size must be divisible by h_size * w_size, "
-                f"got world_size={self._dm.world_size} and h*w={spatial_size}"
-            )
+        _check_world_size_divisible(self._dm.world_size, spatial_size)
         mesh = self._dm.initialize_mesh(
             mesh_shape=(-1, h_size, w_size),
             mesh_dim_names=("data", "h", "w"),
--- a/fme/core/distributed/model_torch_distributed.py
+++ b/fme/core/distributed/model_torch_distributed.py
@@ -106,12 +106,7 @@ class ModelTorchDistributed(DistributedBackend):
     ):
         spatial_size = h_size * w_size
         if _in_dataloader_worker():
-            # see _in_dataloader_worker for why workers skip process group and
-            # CUDA device initialization, which here also means skipping the
-            # DeviceMesh. The (data, h, w) mesh is row-major over global ranks,
-            # so the data-parallel index and size follow from arithmetic; the
-            # spatial groups, mesh and device id are left unset, so that using
-            # them raises rather than silently misbehaving without a mesh.
+            # see _in_dataloader_worker; the (data, h, w) mesh is row-major
             self._rank, self._world_size = _rank_metadata_from_env()
             _check_world_size_divisible(self._world_size, spatial_size)
             self._data_size = self._world_size // spatial_size
--- a/fme/core/distributed/parallel_tests/test_dataloader_worker.py
+++ b/fme/core/distributed/parallel_tests/test_dataloader_worker.py
@@ -1,0 +1,49 @@
+"""
+Tests that the DataLoader worker code path reports the same sharding metadata
+as a fully initialized backend.
+
+A worker skips the process group, the CUDA device and (under spatial
+parallelism) the DeviceMesh, so it derives rank and data-parallel metadata from
+the launcher's environment variables and mesh arithmetic instead. Inference
+dataloaders use that metadata to decide which samples belong to their rank, so
+a mismatch with the real backend would silently drop or duplicate samples.
+
+These tests need a launcher to be meaningful and skip without one.
+"""
+
+import os
+
+import pytest
+import torch.utils.data
+
+from fme.core.device import using_srun
+from fme.core.distributed import Distributed
+from fme.core.distributed.model_torch_distributed import ModelTorchDistributed
+from fme.core.distributed.torch_distributed import TorchDistributed
+
+
+def _build_worker_backend(monkeypatch):
+    """Build the configured backend as a DataLoader worker would see it."""
+    monkeypatch.setattr(torch.utils.data, "get_worker_info", lambda: object())
+    if os.environ.get("FME_DISTRIBUTED_BACKEND") == "model":
+        return ModelTorchDistributed(
+            h_size=int(os.environ["FME_DISTRIBUTED_H"]),
+            w_size=int(os.environ["FME_DISTRIBUTED_W"]),
+        )
+    return TorchDistributed()
+
+
+@pytest.mark.parallel
+def test_worker_sharding_metadata_matches_backend(monkeypatch):
+    if "RANK" not in os.environ and not using_srun():
+        pytest.skip("requires torchrun or srun")
+    if os.environ.get("FME_DISTRIBUTED_BACKEND") == "none":
+        pytest.skip("requires a distributed backend")
+    dist = Distributed.get_instance()
+
+    worker = _build_worker_backend(monkeypatch)
+
+    assert worker.rank == dist.rank
+    assert worker.total_ranks == dist.world_size
+    assert worker.data_parallel_rank == dist.data_parallel_rank
+    assert worker.total_data_parallel_ranks == dist.total_data_parallel_ranks
--- a/fme/core/distributed/test_distributed.py
+++ b/fme/core/distributed/test_distributed.py
@@ -177,6 +177,7 @@ def _forbid_cuda_and_process_group_init(monkeypatch):
 
     monkeypatch.setattr(torch.distributed, "init_process_group", fail)
     monkeypatch.setattr(torch.cuda, "set_device", fail)
+    # get_device() calls current_device(), which lazily initializes CUDA
     monkeypatch.setattr(torch.cuda, "current_device", fail)
 
 
@@ -187,6 +188,11 @@ def _pretend_dataloader_worker(monkeypatch):
 LAUNCHER_ENVS = [
     pytest.param(_set_torchrun_env, id="torchrun"),
     pytest.param(_set_srun_env, id="srun"),
+]
+
+LAUNCHER_ENVS_AND_DEVICE_IDS = [
+    pytest.param(_set_torchrun_env, 1, id="torchrun"),
+    pytest.param(_set_srun_env, 0, id="srun"),
 ]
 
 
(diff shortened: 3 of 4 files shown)


jpdunc23:
Pre-review agent: I ran a pre-review pass over this PR to make the reviewer's job
easier. Summary up front: **structurally this is in good shape** — 68 lines, one
commit, one concern, description states the mechanism and the memory arithmetic
checks out (24 × ~520 MiB ≈ 12.2 GiB). No splitting or restructuring needed. Two
items below need an author decision; the rest are nits.

## Verified

- **The new test is genuinely differential.** Removing the early-return guard from
  `TorchDistributed.__init__` makes
  `test_dataloader_worker_skips_cuda_and_process_group_init` fail on the
  `init_process_group` stub. It is not a test that would pass against the old code.
- **The safety argument holds.** Every worker-side consumer of `Distributed` reads
  only sharding metadata: `fme/ace/data_loading/inference.py` (lines 292, 339) uses
  `total_data_parallel_ranks`/`data_parallel_rank`, and
  `fme/coupled/data_loading/inference.py` (lines 168, 190) uses `world_size`/`rank`.
  No collectives, no GPU tensor placement, in either the ace or coupled path.
- **`torch.cuda.set_device` is the right culprit.** It lands in `cudaSetDevice`,
  and cu126 is past the CUDA 12.0 change that made that eagerly create the device's
  primary context.

## 1. Consider dropping `_device_id` from the worker branch

```python
if using_gpu():
    self._device_id = local_rank
```

This is the only CUDA-adjacent call left in the branch whose whole purpose is to
avoid touching CUDA: `using_gpu()` → `get_device()` → `torch.device("cuda",
torch.cuda.current_device())`, and `torch.cuda.current_device()` calls
`torch.cuda._lazy_init()`. Reading the `_lazy_init` → `_cuda_init` →
`CUDAHooks::initCUDA()` chain, that path sets up the caching allocator and the p2p
cache but never calls `cudaSetDevice`, so I don't believe it creates a primary
context today. I couldn't measure it (no GPU on hand), so that's inference from
source, not observation.

The reason to drop it anyway is that the value is unusable and misleading:

- `_device_id` is read only by `_device_ids` (`torch_distributed.py:203`), which
  feeds `wrap_module` and `barrier` — both of which need the process group the
  worker deliberately didn't join, so they'd raise regardless.
- It records `local_rank` while the process's actual CUDA current device is still
  0, because `set_device` was skipped. Anything that did trust it would disagree
  with `get_device()`.

Leaving it unset makes the branch CUDA-free by construction and turns accidental
use into a loud `AttributeError` instead of a plausible-looking wrong device id.
The corresponding `assert dist._device_id == 3` in the test would come out too
(which also removes a private-attribute assertion).

If it stays, a measured before/after GPU memory figure in the description would be
worth adding — it both validates the diagnosis and settles the `using_gpu()`
question empirically.

## 2. `ModelTorchDistributed` has the same bug and is unfixed

`Distributed.get_instance()` builds `ModelTorchDistributed` under
`FME_DISTRIBUTED_BACKEND=model` (`distributed.py:110-113`), and `worker_init_fn`
reaches it via `Distributed.context()` → `get_instance()`, so it *is* reachable in
a DataLoader worker. Its `__init__` calls `DistributedManager.initialize()`, which
does both `init_process_group` and `torch.cuda.set_device`
(`fme/core/distributed/external/pnd_manager.py:473-481`), plus `initialize_mesh`.
So under spatial parallelism a worker still gets the full CUDA-context cost.

Worth an explicit decision either way — fix here, or state in the description that
spatial parallelism plus forkserver inference loaders is out of scope for this PR
so the reviewer doesn't have to re-derive the question.

## 3. The srun branch of `_rank_metadata_from_env` is untested

```python
if using_srun():
    return int(os.environ["SLURM_PROCID"]), int(os.environ["SLURM_NTASKS"]), 0
```

This is the one place a typo would be invisible in CI — note that
`pnd_manager.py:226` reads `SLURM_NPROCS` for the same quantity where this reads
`SLURM_NTASKS`. Parametrizing the existing test over the torchrun and srun
environments looks cheap and is the best coverage-per-line available here.

Also uncovered: that a *non*-worker still calls `set_device`. A guard that
accidentally became always-true wouldn't be caught by the current tests.

## 4. Nits

- `monkeypatch.setattr(torch.utils.data, "get_worker_info", ..., raising=False)` —
  `raising=False` is unnecessary (the attribute exists) and would mask a rename.
- `test_not_in_dataloader_worker_in_main_process` asserts torch's own behavior
  rather than anything this PR changes. Harmless, but it isn't buying coverage.
- In `if _in_dataloader_worker() and self.is_available():`, the second condition
  can't be false in practice — `Distributed.__init__` only constructs
  `TorchDistributed` when `is_available()` is already true. Worth either dropping
  it or noting why it's there, since a reviewer will pause on it. (Relatedly, the
  new branch makes `is_available()` load-bearing, and its
  `a and b or c` precedence at `torch_distributed.py:108` is pre-existing but
  arguably not what was intended.)

## One suggestion for the description

A reviewer will ask "were workers joining the process group this whole time?" They
were, and it's worth pre-empting with one sentence. I checked under
`torchrun --nproc-per-node=2` with a forkserver loader: `is_initialized()` is
`False` on worker entry, so the worker really does call `init_process_group`, and
it succeeds. It's harmless because torchrun's agent hosts the TCPStore
(`TORCHELASTIC_USE_AGENT_STORE=True`, so every rank including rank 0 connects as a
client), `_store_based_barrier` is already satisfied when the worker increments the
counter so it returns immediately rather than hanging, and NCCL communicator
creation is lazy without a `device_id` argument. The parent's own `barrier()` still
passes afterwards. Net cost was an extra store client per worker plus the CUDA
context — i.e. the description's framing is correct, and saying so explicitly saves
the reviewer the detour.

Tests pass locally: `11 passed, 1 skipped` in
`fme/core/distributed/test_distributed.py`, with `origin/main` merged in.

jpdunc23 (approved):
Looks good, please have a look at https://github.com/ai2cm/ace/pull/1396#issuecomment-5134891121 before merging. Up to you whether / how to address 1. Seems like the srun testing would be easy to add, but optional as well.

yyexela:
Thanks for the pre-review — all items addressed in `b2daccd4..4619d958`.

**1. `_device_id` in the worker branch** — dropped, as suggested. `_rank_metadata_from_env` now returns only `(rank, world_size)`, so `LOCAL_RANK` isn't read and `using_gpu()` isn't called on the worker path. The branch is CUDA-free by construction, and the `_device_id` assertion is out of the test.

**2. `ModelTorchDistributed`** — brought into scope rather than deferred. Same guard, additionally skipping `DistributedManager.initialize` and the DeviceMesh. Since the mesh normally supplies data-parallel rank and size, the worker derives them from the row-major `(data, h, w)` layout: `data_rank = rank // (h*w)`, `data_size = world_size // (h*w)`. That layout was already relied on by `parallel_tests/test_backend.py::test_reduce_mean_data_parallel_group_only`, and the new parallel test checks the arithmetic against a real mesh. The divisibility check is now `_check_world_size_divisible`, shared by both paths.

**3. srun coverage** — worker tests are parametrized over the torchrun and srun environments. Added `test_dataloader_worker_guard_does_not_affect_main_process`, which asserts the main process still calls `set_device`; forcing the guard always-true fails it. `SLURM_NTASKS` is left as-is: it matches this file's own srun branch (`torch_distributed.py:86`), whereas `SLURM_NPROCS` is vendored `pnd_manager` code.

**4. Nits** — `raising=False` removed; `and self.is_available()` removed; `test_not_in_dataloader_worker_in_main_process` deleted. The `is_available()` precedence oddity is untouched as pre-existing.

**5. Description** — added the paragraph on workers having joined the process group and why that was harmless, plus the two-backend scope and the derivation rationale.

Verification: forcing the guard false fails all four worker tests, forcing it true fails both main-process tests, and breaking `rank // spatial_size` fails the parallel mesh test. `make cpu_test_all_parallel` passes all 8 configs; `fme/core/distributed/` is 68 passed, 3 skipped; pre-commit clean.

No GPU available here, so the 12.2 GiB figure remains arithmetic (24 × ~520 MiB), not a measurement — worth confirming on a GPU run before merge.

yyexela:
Pre-review agent: pre-review pass over the branch. One code change, plus two test cleanups:

- `_rank_metadata_from_env` raised `KeyError: 'RANK'` when a worker was constructed with no launcher environment (reachable via a forced `FME_DISTRIBUTED_BACKEND=torch`), because the worker guard sits above the existing `else: raise ValueError(...)`. It now raises that same `ValueError`. New unit test `test_dataloader_worker_without_launcher_env_raises`.
- `test_dataloader_worker_guard_does_not_affect_main_process` no longer reads `dist._device_id`; the expected device id is parametrized per launcher (torchrun -> `LOCAL_RANK`, srun -> 0).
- Restored a dropped comment explaining why `torch.cuda.current_device` is stubbed in `_forbid_cuda_and_process_group_init` (production code reaches it via `get_device()`, not directly).

Worth knowing while reviewing:

- The unit tests in `test_distributed.py` are what pin the guard: they stub `init_process_group` / `DistributedManager.initialize` to raise, so they fail if the guard is removed. The parallel test `test_dataloader_worker.py` instead validates the `(data, h, w)` mesh arithmetic against a real mesh, and would still pass without the guard. It is only non-trivial for `h * w > 1`, which five of the eight `make cpu_test_all_parallel` configs cover.
- `NonDistributed` needs no equivalent guard, since it never calls `torch.cuda.set_device`, so single-process runs were never affected.
- Verified: `_worker_loop` sets `_worker_info` before calling `init_fn` (torch 2.7.1), so `get_worker_info()` is non-`None` inside `worker_init_fn`.
- Verified the worker really needs this metadata rather than the `worker_init_fn` being vestigial: `InferenceDataset._get_batch_data` and `__getitem__` both call `Distributed.get_instance()`.
- Verified nothing worker-reachable touches the attributes the worker leaves unset. `_device_id` is only read by `_device_ids` (used by `wrap_module` and `barrier`), and `_h_size` / `_w_size` only via `get_local_slices`, whose only data-path caller is `localize()` in `gridded_data.py` — all main process.

Ran locally: `test_distributed.py` 16 passed / 1 skipped; the new parallel test across all eight `cpu_test_all_parallel` configs; pre-commit clean.

jpdunc23:
Re-review of the current head (`94f73a31`), which now covers both backends. **No blockers — this is ready to merge.** My approval stands.

All four items from the earlier pass are addressed as described, and I re-verified the parts that were previously argued rather than tested.

## Verified on the new code

- **The guard is load-bearing in both directions.** Forcing `_in_dataloader_worker()` to return `False` fails all four worker tests (both launchers × both backends); forcing it to return `True` fails both `test_dataloader_worker_guard_does_not_affect_main_process` cases. Neither direction can rot silently.
- **The mesh arithmetic is right, and the parallel test actually pins it.** For a row-major `(data, h, w)` mesh, rank *r*'s data-parallel group is the ranks sharing its `(h, w)` coordinates, and *r*'s index within that sorted group is `r // (h*w)` — which is what the worker computes. Mutating `self._rank // spatial_size` to `% spatial_size` fails `test_worker_sharding_metadata_matches_backend` on 6 of 8 ranks under `NPROC=8 H=2 W=2`. Worth noting the `NPROC=4 H=2 W=2` config alone wouldn't have caught it (`data_size == 1`, so every rank's `data_parallel_rank` is 0) — the 8-rank config is what gives this test teeth, and CI runs it.
- **Attribute names line up with the properties they feed.** `TorchDistributed` sets `self.world_size` (plain attribute, matching the rest of that class) while `ModelTorchDistributed` sets `self._world_size` (matching its `total_ranks` property) — the asymmetry is each class's own convention, not a slip.
- **Everything worker-reachable is populated.** `fme/ace/data_loading/inference.py` reads only `total_data_parallel_ranks` and `data_parallel_rank` (lines 297, 349); the coupled dataset reads only `rank` and `world_size`. All four are set on both worker branches. The attributes left unset — `_device_id`, `_local_rank`, the spatial groups and sizes, `_mesh`, `_dm` — are reachable only via `get_local_slices`/`wrap_module`/`barrier`, none of which a worker touches.
- **Spatial parallelism config can't diverge between parent and worker.** `spatial_parallelism` is only ever populated from `FME_DISTRIBUTED_BACKEND`/`FME_DISTRIBUTED_H`/`FME_DISTRIBUTED_W` in `Distributed.get_instance()` (`distributed.py:110-113`) — never from a YAML config — so a worker rebuilding the singleton from the environment always selects the same backend with the same mesh shape as its parent. That's what makes the arithmetic safe to trust.
- **No other worker path is affected.** Only the two inference getters install a `worker_init_fn`; `fme/downscaling/` sets a multiprocessing context but no init fn. And for a fork-context loader the parent's singleton is inherited, so `__init__` never re-runs.

`fme/core/distributed/` is 69 passed / 3 skipped; the new parallel test passes under `NPROC=4 H=2 W=2` and `NPROC=8 H=2 W=2`; pre-commit clean; all CI checks green including the GPU job.

## Two non-blocking notes

1. **The 12.2 GiB is still arithmetic (24 × ~520 MiB), not a measurement.** The mechanism isn't in doubt — `cudaSetDevice` eagerly creates the primary context — and the GPU CI job passing says nothing regressed. But since the motivation is a specific training config that OOM'd, it'd be worth confirming the recovered headroom on that run once it's back up, purely to close the loop on the diagnosis. Nothing to change in the code.
2. **For the record: `SLURM_NTASKS` vs `SLURM_NPROCS`.** Under srun *and* the model backend, the worker reads `SLURM_NTASKS` while the parent's `DistributedManager.initialize_slurm` reads `SLURM_NPROCS` (`pnd_manager.py:226`) for the same quantity. Slurm sets both to the same value (`SLURM_NPROCS` is a deprecated alias), so there's no live divergence, and keeping `SLURM_NTASKS` is the right call — it matches the srun branch in the same file. Just flagging it so the next person who reads the two side by side doesn't have to re-derive that they agree.

PR #1398, opened 2026-07-29
Attempt an orderly distributed teardown on termination in every entrypoint

mcgibbon:
**Handle termination signals in one place for every entrypoint**, and **tear the process group down before the restart checkpoint.** Preempted ACE jobs cordon nodes: a rank SIGKILLed with its NCCL communicators still open drops its NVLink peers, the survivors fault with `CUDA error: Invalid access of peer GPU memory over nvlink or a hardware error`, and the GPUs need rebooting. 18 of 18 recent ≥4-GPU preemptions in `ai2/ace` escalated at least one rank to SIGKILL. Two things got us there. `Trainer.__init__` installed the only SIGTERM handler in the codebase, so inference, evaluation and downscaling jobs had none and neither did training before the Trainer was constructed; the handler now lives in `Distributed.context()`, which all 11 entrypoints already wrap their work in. And that handler wrote a multi-GB restart checkpoint *before* `dist.shutdown()`, which is collective, so the other ranks blocked inside it for the 7–15 s of rank 0's write and were SIGKILLed there. That order is now reversed, which matters because torch elastic's agent gives every rank on a node a shared 30 s rather than 30 s each (`PContext.close` defaults to `timeout=30` and `LocalElasticAgent._shutdown` calls it without an argument), so one slow rank spends its peers' allowance.

This is best-effort rather than a guarantee. The handler runs only when the main thread returns to the interpreter, so a rank blocked in a C-level call — a collective's stream sync above all — never starts the teardown, and a rank that does start one cannot finish it while such a peer still holds the communicator. A [deliberate 4-rank preemption](https://beaker.org/orgs/ai2/workspaces/ace/work/01KYWME86AF7C3XRG6M3TBJ4WH) showed both halves: SIGTERM did reach every rank, two ran their handler and hung in the teardown until the 20 s watchdog exited them, two never ran theirs and were SIGKILLed, and no restart checkpoint was written because the watchdog's `os._exit` skips the post-shutdown callbacks. That node stayed clean of SXid errors, but the graceful path never completed, so it is not evidence that the graceful path protects one. Making mid-step preemption safe is still work to be done in a follow-up.

Changes:
- `fme.core.distributed.shutdown`: new module. `handle_termination_signals` tears the backend down on SIGTERM/SIGINT, runs the callbacks registered with `add_post_shutdown_callback`, then exits `128 + signum`. A watchdog thread bounds the collective teardown at 20 s (`DEFAULT_TEARDOWN_TIMEOUT`, 10 s under elastic's 30 s) and reports through a raw write to stderr, since the main thread it races may be holding a logging lock it will never release. Nothing is installed off the main thread or in a DataLoader worker, and the handler is inert in any process other than the one that installed it — forked workers inherit it and would otherwise tear down a fork-inherited process group and race the parent's checkpoint write.
- `fme.core.distributed.Distributed.context`: handles termination signals for the lifetime of the context, unless called with `handle_signals=False` by a caller that owns the process's own response to them.
- `fme.core.generics.Trainer`: registers its interrupt-time restart checkpoint as a post-shutdown callback instead of installing its own handler. Removes `chain_signal_handler`, now unused.
- `TorchDistributed.shutdown` and `ModelTorchDistributed.shutdown`: skip the collective when the process group is already gone, so a signal arriving during the normal end-of-run teardown is harmless. Both log at INFO — at DEBUG these lines were invisible in every preemption log.
- `fme.core.device.in_dataloader_worker`: the "is this process a DataLoader worker" check, in one place rather than three.
- `conftest.py`: enters the session-scoped `Distributed.context()` with `handle_signals=False` so Ctrl-C keeps interrupting pytest rather than becoming a caught teardown, and resets the callback registry between tests.
- Two effects outside the preemption path, across all 16 call sites: SIGINT raises `SystemExit` rather than `KeyboardInterrupt`, so an interrupted run exits without a traceback, and a termination signal exits `128 + signum` — 143 for SIGTERM, 130 for SIGINT — rather than 1.

- [x] Tests added


mcgibbon:
Claude: relationship to #1396 ("Avoid CUDA context creation in DataLoader workers"), which was filed in response to the same message from the Beaker team.

The two are independent and both worth landing. #1396 adds no signal handling, so it does not give a preempted job a graceful exit — but it does stop inference DataLoader workers from creating CUDA contexts and joining the process group, which removes a set of GPU-holding processes that were also being killed abruptly. It helps at the margin on the same symptom.

One overlap to note for whichever lands second: #1396 adds `_in_dataloader_worker()` in `torch_distributed.py` and this PR does the same `torch.utils.data.get_worker_info() is not None` check inline in `shutdown.py`. They should be de-duplicated onto the shared helper at that point.

mcgibbon:
Claude: **the independent pre-review did not run** — the agent and its two audit subagents all terminated with `You've hit your org's monthly spend limit`. So this PR has not passed the pre-review gate and is deliberately still a draft.

What has been verified:
- CI green on d31adbf49: build, cpu, gpu, cpu-very-fast-no-healpix, pre-commit, GitGuardian.
- Locally: 1424 passed in `fme/core`; 832 passed across `fme/ace/data_loading`, `fme/ace/test_train.py`, `fme/ace/inference`, `fme/coupled`.
- Each new test was confirmed to fail without its fix: reverting the `Distributed.context()` wiring, the `deadline.cancel()`, and the re-entrancy guard each makes the corresponding test fail.

Worth a reviewer's attention specifically: whether tearing down the process group *before* the restart-checkpoint write is definitely safe. I traced `save_checkpoint` → `stepper.get_state()` / `optimization.get_state()` / `EMATracker.get_state()` and found only local state-dict reads with no collectives, but a mistake there would silently lose restart checkpoints on every preemption, so it deserves a second pair of eyes.

jpdunc23:
Pre-review agent: I ran a pre-review of this branch with `main` merged in, from zero prior context. `pytest fme/core/distributed/test_shutdown.py fme/core/distributed/test_distributed.py fme/core/generics/test_trainer.py` passes (85 passed, 1 skipped) and `ruff check` is clean.

The diagnosis in the description is unusually well-evidenced, and the core insight — teardown before the checkpoint write, handler at `Distributed.context()` rather than on the `Trainer` — looks right. The findings below are all in the new module's edges, and two of them undercut the PR's own stated goal.

## Audit pass summary

### 1. Silent-failure pass — ✅

~**B1. Forked DataLoader workers inherit the handler *and* the callback registry; the guard meant to prevent this is in the wrong place.**~

**Correction (added after posting): B1 as written below is wrong — it does not fire on SIGTERM, so it is not a real failure mode for preemption, and this pass should be read as ✅ rather than ❌.**

The claim below is that a forked DataLoader worker runs the inherited handler's teardown on SIGTERM. It does not. torch's `_worker_loop` calls `signal_handling._set_worker_signal_handlers()` (`torch/utils/data/_utils/signal_handling.py:23-27`), which installs a **C-level** SIGTERM handler in the worker, overwriting CPython's — so the inherited Python `handle` is dead code for SIGTERM. My original evidence checked the handler's *disposition* (`signal.getsignal(SIGTERM).__qualname__`), which is indeed inherited, but not whether it actually fires. It doesn't. Verified on this branch: worker `SigCgt: 00000001000044c2` vs driver `0000000100014002`, and `killpg(pgid, SIGTERM)` kills the workers with torch's own `RuntimeError: DataLoader worker (pid ...) is killed by signal: Terminated` while producing **no** worker teardown markers.

What survives is a narrower case: torch installs nothing for **SIGINT**, so under `killpg(pgid, SIGINT)` forked workers do run the full teardown (verified — both workers wrote shutdown and callback markers and exited 130). That is an interactive-Ctrl-C scenario, not a preemption one, and these nodes do not typically run interactive training or debugging jobs — so it is a minor edge case, not a blocker.

Two further corrections to the framing below: the fork-inheritance hazard is **not introduced by this PR** — `Trainer.__init__` on `main` already installs `chain_signal_handler(SIGTERM/SIGINT, on_terminate)` (`fme/core/generics/trainer.py:354-355`) ending in `dist.shutdown()`, so it pre-exists; this PR widens it to every job kind and makes it deterministic. And "the failure mode this PR exists to remove" overstates it accordingly. The suggested pid guard remains cheap and correct as belt-and-braces for the SIGINT case, but it is **not** a merge blocker, and item 1 of the recommendations at the end of this comment should be read in that light.

<details>

`fme/core/distributed/shutdown.py:129-134` skips installing a handler when `get_worker_info()` is not `None`, and the comment says "DataLoader workers enter this context only to learn their rank." That is only true of the **forkserver** path (`fme/ace/data_loading/getters.py:158-163`, `fme/coupled/data_loading/getters.py:50`) — and those workers are `spawn`-descended, so they never had the parent's handler to begin with. The path that *is* dangerous is the default one: `fme/ace/data_loading/getters.py:116-117` sets `mp_context = None` and `worker_init_fn = None`, i.e. **fork**, and forked children inherit Python-level signal handlers. torch does not reset SIGTERM/SIGINT in `_worker_loop` (it only installs C-side SIGBUS/SIGSEGV handlers, `torch/utils/data/_utils/worker.py:253`).

I confirmed this empirically: inside `handle_termination_signals`, a `num_workers=2` DataLoader worker reports `our_handler_installed=1, n_callbacks=1, is_worker=1` with `start_method=fork`. And torchrun delivers the signal to them — `SubprocessHandler` launches each rank with `start_new_session=True` and closes with `os.killpg(self.proc.pid, death_sig)` (`torch/distributed/elastic/multiprocessing/subprocess_handler/subprocess_handler.py:56,74`), so every DataLoader child of every rank gets SIGTERM directly. Consequences, in order of severity:

- Each worker calls `Distributed.get_instance().shutdown()` → `destroy_process_group()` on a CUDA/NCCL context inherited across fork. That is exactly the undefined-behaviour case; at best it hangs until its own 20 s watchdog `os._exit(143)`s it, at worst it aborts or itself faults the fabric — the failure mode this PR exists to remove.
- The parent's unwinding then blocks: `multiprocessing.util._exit_function` joins daemonic children after terminating them, so the parent waits on N workers each burning up to `DEFAULT_TEARDOWN_TIMEOUT`. The watchdog has already been cancelled at that point (see B2), so nothing bounds it.
- The `Trainer` callback is inherited too. It is usually inert in the child because `self._current_epoch_num_batches_seen` is frozen at fork time (0 at iterator creation, so the `> 0` check fails) — but on a **resumed mid-epoch run** it is non-zero at fork time (`fme/core/generics/trainer.py:551`), so forked workers each `torch.save` a stale multi-GB restart checkpoint and race the parent's `os.replace` onto `latest_checkpoint_path`. Last writer wins: a silent rollback of training progress.

Suggested change — in `handle_termination_signals`, capture the installing pid and make the handler refuse to act in any other process:

```python
    installed_pid = os.getpid()

    def handle(signum, frame):
        if os.getpid() != installed_pid:
            # inherited across fork (DataLoader workers with the default
            # start method); the process that installed this owns the teardown
            signal.signal(signum, signal.SIG_DFL)
            signal.raise_signal(signum)
            return
        ...
```

The `get_worker_info()` comment at `:129-134` should then be corrected or dropped, since the pid guard subsumes it.

Other silent-failure items are clean: `_tear_down` (`:91-107`) swallows per-step failures deliberately and logs via `logger.exception`; catching `BaseException` around callbacks so a `sys.exit()` in one cannot hijack the exit code is correct and tested; the empty-registry case is a no-op by construction. No configs or dacite unions are touched.
</details>

### 2. Behavior-change pass on existing code paths — ⚠️

**B2. The watchdog bounds the checkpoint write, contradicting the PR's own rationale.**

<details>

`shutdown.py:166` arms the deadline, `_tear_down` runs `shutdown()` **and then all callbacks** under it (`:98-107`), and `:173` cancels only afterwards. The description says the checkpoint "now runs on borrowed time where overrunning the grace period is harmless" — but the code hard-kills it at 20 s, *stricter* than torchrun's 30 s. With the description's own numbers (7–15 s for the write) plus a multi-second NCCL `destroy_process_group`, a real preemption can `os._exit` mid-`torch.save` and lose the restart checkpoint the reordering was designed to preserve. The watchdog's stated purpose is to stop a wedged rank holding its *peers*; once `shutdown()` returns the peers are already safe, and the remaining deadline protects nobody. Suggested change — cancel the deadline as soon as `shutdown()` returns, before the callbacks:

```python
def _tear_down(shutdown, deadline: threading.Timer) -> None:
    try:
        shutdown()
    except BaseException:
        logger.exception("Failed to shut down the distributed backend.")
    finally:
        deadline.cancel()  # peers are safe; callbacks run on the scheduler's clock
    for callback in _post_shutdown_callbacks:
        ...
```

</details>

**G1. `ModelTorchDistributed.shutdown`'s new guard is both redundant and lossy.**

<details>

`fme/core/distributed/model_torch_distributed.py:456-460` returns before `DistributedManager.cleanup()`. But `cleanup()` is already internally guarded on its own `_shared_state["_is_initialized"]` (`fme/core/distributed/external/pnd_manager.py:499-511`) *and* unconditionally does `DistributedManager._shared_state = {}` (`:512`). So the new early return changes behaviour: the manager's shared state is no longer reset. `torch.distributed.is_initialized()` is not the right predicate for this backend. Suggested change — guard only the log line and let `cleanup()` decide:

```python
    def shutdown(self):
        if torch.distributed.is_initialized():
            logger.info("Shutting down rank %d", self._rank)
        DistributedManager.cleanup()
```

The `TorchDistributed` guard (`torch_distributed.py:235-240`) *is* needed — bare `destroy_process_group()` raises when uninitialised — so that one should stay.

</details>

**G3. SIGINT semantics changed from `KeyboardInterrupt` to `SystemExit`, unstated.**

<details>

The old `chain_signal_handler` called `prev_handler`, which for SIGINT is `signal.default_int_handler` and therefore *raised `KeyboardInterrupt`*; `sys.exit(1)` was unreachable. Now every entrypoint exits 130 via `SystemExit` with no traceback. Defensible, but it is a user-visible change across 16 call sites (including `docs/conf.py:267` and everything under `scripts/`), and it belongs in the description. Also unstated: exit code `1` → `128 + signum` — worth a sentence on whether Beaker's or torchrun's retry accounting distinguishes 143 from 1.

Correctly stated and correct: DEBUG→INFO on both `shutdown` log lines; the teardown-before-checkpoint reordering; the `chain_signal_handler` removal (grep confirms no remaining references). `save_checkpoint` is indeed root-only and collective-free (`trainer.py:668-697`), so the "runs after the PG is gone" claim holds.

</details>

### 3. Differential-test pass — ⚠️

The suite is genuinely good — I checked each new test against a no-op of the code it covers and all of them fail correctly. `test_a_signal_after_a_swallowed_exit_kills_the_process` is a sharp test in particular (without the `teardown_complete` branch the second signal falls into the `tearing_down` no-op and `exited == []`). `test_resume_after_preemption` in `test_trainer.py` remains differential: `n_checkpointed_batches` is 25 for the preempt path vs 20 for the exception path.

**G2. Neither new `is_initialized()` early return has a test.** Two new branches, zero coverage, and one of them is wrong (G1). Suggested addition to `test_distributed.py`:

<details>

```python
def test_shutdown_is_a_noop_when_the_process_group_is_already_gone(monkeypatch):
    destroyed = []
    monkeypatch.setattr(torch.distributed, "is_initialized", lambda: False)
    monkeypatch.setattr(
        torch.distributed, "destroy_process_group", lambda *a: destroyed.append(a)
    )
    TorchDistributed.shutdown(cast(TorchDistributed, SimpleNamespace(rank=0)))
    assert destroyed == []
```

Also: `test_deadline_does_not_outlive_a_successful_teardown` (`test_shutdown.py:150-163`) asserts a guarantee the code cannot give — `Timer.cancel()` is a no-op once `give_up` has begun, so a clean teardown can still exit 143 if the timer fires in the microseconds before `:173`. The test only covers the common case; a comment saying so would stop a future reader trusting it too far.

And there is **no test for B1** — a test that a forked child does not run the teardown is the one that would have caught it.

</details>

### 4. Sibling / mirror coverage pass — ✅

All three `DistributedBackend.shutdown` implementations are accounted for (`non_distributed.py:139` is already `return`). Both worker-init mirrors (`fme/ace/data_loading/getters.py`, `fme/coupled/data_loading/getters.py`) are covered by the shared module. `fme/core/generics/trainer.py` is the shared trainer, so ace/coupled/downscaling all get the change together — no mirror PR needed.

### 5. Config back-compat pass — ✅ (N/A)

No loaded configs, no checkpoint-persisted state, no YAML schema touched. Checkpoint *contents* unchanged (`trainer.py:680-692`).

### 6. Split heuristic — ✅

9 files, one concern, no new config, no shim. Should not be split. One structural nit: commits 2 and 3 are successive patches to the module introduced in commit 1, so a commit-by-commit reviewer reads three revisions of `shutdown.py`. Since this squash-merges the cost is low, but squashing them would make a commit-walking reviewer's job strictly easier.

### 7. Abstract-or-final / inheritance-depth pass — ✅

No new classes, no inheritance changes.

### 8. AI-agent artifact pass — ⚠️

- `test_shutdown.py:195` `import threading`, `:213` `import torch.utils.data`, `:215` `from fme.core.distributed import shutdown as shutdown_module` — all inside test bodies; these should be lifted to the module top.
- `:215` and `:222` reach `handle_termination_signals` through the module alias, but it is already imported at `:11`. Redundant indirection — call the imported name.
- `:218` passes `raising=False` to `monkeypatch.setattr(torch.utils.data, "get_worker_info", ...)`, but the attribute exists, so this only disables typo protection. Drop it.
- The `conftest.py:192` in-fixture import matches the neighbouring `reset_global_timer` (`:180`), so that one is consistent with local style — leave it.
- No `Any` returns, no `isinstance`/`type: ignore` pairs, no abbreviations, no stray comments. The docstrings and comments in the new module are unusually good; the "instants before the handler is installed remain unprotected" paragraph (`shutdown.py:17-25`) is exactly the honesty a reviewer wants.

### 9. Builder pattern pass — ⚠️ (design comment, not a rule violation)

No configs involved. But the PR introduces process-global mutable state (`_post_shutdown_callbacks`, `:47`) whose correctness across the test suite depends on a root-`conftest.py` autouse fixture, and `handle_termination_signals` reaches out to clear a registry it does not own (`:184`) — asymmetrically, since the two early-return paths (`:124`, `:129`) do not clear. A reviewer will likely ask why the context does not simply yield a registrar:

```python
with handle_termination_signals(instance.shutdown) as on_shutdown:
    yield on_shutdown
```

which makes the lifetime explicit, deletes `clear_post_shutdown_callbacks`, and removes the conftest fixture entirely. Not blocking — but worth deciding before merge, because the global-registry version is what future callers will copy.

## Overall state and suggested next steps

**Needs revision before review** — not because the design is wrong, but because two defects would keep the reported symptom alive in production, and a third makes the test suite worse to use. In order:

1. **B1** (pid guard in `handle`) — the one I would not merge without. Forked DataLoader workers running `destroy_process_group()` on a fork-inherited CUDA context, with the `Trainer` checkpoint callback attached, on every rank of every preempted training job, is the same class of fault the PR is trying to eliminate — plus a real (if narrow) path to a silently rolled-back restart checkpoint. Worth a regression test alongside.
2. **B2** (cancel the deadline after `shutdown()`, not after the callbacks) — currently the watchdog is stricter than the scheduler and can truncate the very checkpoint write the reordering was designed to protect.
3. **G1** (guard the log line, not `DistributedManager.cleanup()`), plus the G2 tests for both early returns.
4. **Description additions**: the SIGINT `KeyboardInterrupt`→`SystemExit` change; the exit-code change and whether Beaker/torchrun care; and Beaker's actual grace period — the 20 s default is justified against torchrun's 30 s only, and if Beaker's is shorter the watchdog never fires in time.
5. **A decision to make before the reviewer hits it**: because `conftest.py:145-148` enters `Distributed.context()` at session scope, the handler is now installed for the entire test suite. Ctrl-C during `make test` no longer stops it — the first SIGINT tears the backend down and becomes a *test failure*, the session continues with the process group destroyed, and a second Ctrl-C `os._exit(130)`s with no summary, no `--lf` cache, no tmpdir cleanup. The `teardown_complete` hard-exit in commit 3 reads as a workaround for this rather than a fix. Cleanest resolution I can see: add `handle_signals: bool = True` to `Distributed.context()` and pass `False` from the conftest fixture, so tests that want the handler install it themselves (they all already do — `test_shutdown.py` and `preempt_after_calls_patch` each open their own context).
6. **The one gap the description leaves open**: it presents excellent *before* evidence (18 of 18 recent ≥4-GPU preemptions escalating to SIGKILL) and no *after* evidence. Has a real ≥4-GPU Beaker job been preempted on this branch and confirmed to exit cleanly without SXid/NVLink faults? That single data point is what turns this from a plausible fix into a verified one, and it is likely the first thing the Beaker admins who asked for this will want.

Once 1–3 land, this is a tight, well-documented, well-tested change that should review in about an hour. Nothing here warrants restructuring or splitting.

jpdunc23:
Re-review agent: follow-up to the [pre-review report](https://github.com/ai2cm/ace/pull/1398#issuecomment-5133185030), covering the five fix commits pushed since it (`87480a768..b86b160b7`), with `main` (`cc931ee94`) merged in locally.

`pytest fme/core/distributed/test_shutdown.py fme/core/distributed/test_shutdown_dataloader.py fme/core/distributed/test_distributed.py fme/core/generics/test_trainer.py` → **93 passed, 1 skipped** (was 85 + 1). `pre-commit` on all nine touched files — ruff, ruff-format, mypy — passes.

## Status of the pre-review items

| # | Item | Status |
|---|---|---|
| 1 | **B1** — pid guard in `handle` | ✅ `shutdown.py:199-209`, plus **two** tests: an in-process `os.fork()` case and a real-DataLoader/real-process-group subprocess test |
| 2 | **B2** — cancel the deadline after `shutdown()`, not after the callbacks | ✅ `_shut_down_backend` cancels in `finally`; tested both ways (collective succeeded / collective raised) |
| 3 | **G1** — guard the log line, not `DistributedManager.cleanup()` | ✅ exactly as suggested |
| 3 | **G2** — tests for both `is_initialized()` early returns | ✅ both added |
| 5 | `handle_signals=False` for the session-scoped conftest context | ✅ exactly the suggested resolution, with a test |
| 8 | AI-agent artifacts (in-body imports, module alias, `raising=False`) | ✅ all four |
| 9 | Global callback registry vs. a yielded registrar | ✅ decided and recorded in the description |
| 4 | Description: SIGINT and exit-code changes | ✅ stated, as is Beaker's 5 min. ⚠️ still silent on whether Beaker's or torchrun's retry accounting distinguishes 143 from 1 |
| 6 | After-evidence from a real preemption | ❌ still open |

## No new defects in the delta

I looked specifically for regressions the fixes could introduce, and found none:

- The pid guard runs before every phase check, so a forked child can never reach the teardown.
- `signal.signal(signum, SIG_DFL)` followed by `raise_signal` is safe in the child — Python signal handlers always run on the main thread, and the child touches no CUDA before dying.
- Dropping the early return in `ModelTorchDistributed.shutdown` cannot `KeyError`: `cleanup()` tests `"_is_initialized" in DistributedManager._shared_state` rather than doing a bare lookup (`pnd_manager.py:499`), so it tolerates a manager that was never initialized.
- The fork test's pipe protocol cannot pass vacuously — the parent closes `write_fd` after forking, so EOF is reachable only via the child's death, and the `WTERMSIG == SIGTERM` assertion catches a `raise_signal` that quietly killed nothing.

Two things worth calling out as better than the report asked for. The `_Phase.CALLBACKS` escalation branch is a consequence of B2 that the report did not name: once the callbacks are unbounded, a rank stalled in a checkpoint write would be deaf to every signal until SIGKILL, and that branch is the escape hatch. And `test_shutdown_dataloader.py` goes well past the suggested regression test — it drives a real forked-worker DataLoader inside a real `Distributed.context()`, signals the process group the way torchrun's `SubprocessHandler` does, and asserts the preconditions (two workers, `start_method=fork`, handler actually inherited) so the negative assertion cannot pass vacuously. Its docstring correctly explains why it tests SIGINT rather than SIGTERM.

## On the concern raised in the PR discussion

@mcgibbon [asked](https://github.com/ai2cm/ace/pull/1398#issuecomment-5123461255) for a second pair of eyes on whether tearing the process group down *before* the restart-checkpoint write is definitely safe. I traced it independently and it holds. `_should_save_checkpoints` → `is_root()` reads a cached rank (`distributed.py:410`), and `save_checkpoint`'s payload is `stepper.get_state()` / `_ema.get_state()` / `optimization.get_state()`, the last being `optimizer.state_dict()` on a plain torch optimizer. The decisive check is that there is **no `ZeroRedundancyOptimizer` and no FSDP anywhere in `fme/`** — that is what would have made `state_dict()` collective. Nothing on that path touches the process group.

## Outstanding before merge

1. **The after-evidence.** Unchanged in character from the pre-review: the description has excellent *before* data (18 of 18 recent ≥4-GPU preemptions escalating to SIGKILL) and no *after* data point. One real ≥4-GPU Beaker preemption on this branch, confirmed to exit cleanly with no SXid/NVLink fault, is what turns this from plausible into verified, and it is likely the first thing the Beaker admins who asked for this will want. It is entangled with a question this PR does not answer: jobs launch as `gantry run ... -- torchrun ...`, so the elastic agent is a grandchild of the container entrypoint, and none of this fires unless the scheduler's SIGTERM actually reaches it. A single preemption on this branch would settle both at once.
2. **One line in the description** on whether the exit-code change (1 → `128 + signum`) matters to Beaker's or torchrun's retry accounting.

## Recommendation

**Ready for external review.** Every blocking item from the pre-review is resolved, and resolved the way the report suggested rather than worked around. Item 1 above is a merge gate in my view, but it is evidence to gather rather than a code change; item 2 is a sentence.

fme/core/distributed/model_torch_distributed.py
```diff
@@ -454,5 +454,11 @@ def get_disco_conv_s2(self, *args, **kwargs) -> nn.Module:
         return thd.DistributedDiscreteContinuousConvS2(*args, **kwargs)
 
     def shutdown(self):
-        logger.debug("Shutting down rank %d", self._rank)
+        # a termination signal can arrive during the normal end-of-run teardown,
```
Arcomano1234: Nit optional: comment seems quite verbose for a minor point

fme/core/distributed/shutdown.py
```diff
@@ -0,0 +1,266 @@
...
+def _hard_exit_after(timeout: float, exit_code: int) -> threading.Timer:
+    """Exit the process if the graceful teardown has not finished in time.
+
+    A blocked collective does not return to the interpreter, so a Python signal
+    handler or ``signal.alarm`` would never run; only a separate thread can
+    enforce the deadline.
+    """
+
+    def give_up() -> None:
+        logger.error(
```
Arcomano1234: Suggestion: Fable flagged this but it seems logging handler locks are RLocks: re-entrant for the main thread (so the handler's own logger.info calls are safe), but the watchdog runs on its own thread. If the signal lands while the main thread is mid-log-emit (holding the lock) — plausible, training logs constantly — and the teardown then wedges in NCCL, the watchdog thread blocks acquiring the lock and never reaches os._exit. The rank rides to torchrun's SIGKILL and the SXid fault returns. One-line fix: replace logger.error in give_up with a raw os.write(2, b"..."), which needs no locks.

fme/core/distributed/shutdown.py
```diff
@@ -0,0 +1,266 @@
...
+    killing outright.
+    """
+
+    RUNNING = enum.auto()
+    COLLECTIVE = enum.auto()
+    CALLBACKS = enum.auto()
+    COMPLETE = enum.auto()
+
+
+def add_post_shutdown_callback(callback: Callable[[], None]) -> None:
```
Arcomano1234, line 72: Claude : Non-blocking: nothing here prevents duplicate registration, and `Trainer.__init__` registers its restart-checkpoint callback as a per-construction side effect. One-Trainer-per-process holds everywhere today, so this is fine as-is — but a second construction in the same process (future lr-tuning/sweep code, say) would silently write the restart checkpoint twice on termination.

Since the registry is the API future callers will copy, worth a sentence in this docstring so the constraint is discoverable, e.g.:

Registering the same callback twice runs it twice; callers that may be constructed more than once per process should guard their own registration.

fme/core/distributed/shutdown.py
```diff
@@ -0,0 +1,266 @@
+"""Tear the distributed backend down cleanly when a job is preempted.
```
Arcomano1234, line 1: Question: this file has a lot of inline comments that I assume are Claude generated. Do we want to have this style. IMO for this rather hard to follow / not obvious because there are a lot of edge cases code its fine

Arcomano1234 (approved):
Left a few non-blocking comments / style questions, the only thing that should be considered is the edge case of using a logger during `_hard_exit_after `

fme/core/distributed/shutdown.py
jpdunc23: Confirmed and fixed — reproduced by holding the handler lock across the signal: the control exits 143, but with the lock held the process never exits at all (mine sat for 5.5 hours before I killed it). `give_up` now writes to fd 2 instead, with a regression test that deadlocks on the old code.

fme/core/distributed/shutdown.py
jpdunc23: Added, close to your wording. Left the behaviour alone since one Trainer per process holds everywhere today, and a registry that silently de-duplicated would be the more surprising API for the next caller.

fme/core/distributed/model_torch_distributed.py
jpdunc23: Fair — trimmed to just the part that stops someone reinstating the early return.

fme/core/distributed/shutdown.py
jpdunc23: Keeping the style, but you prompted a redundancy pass and there was some: the point that the watchdog bounds the collective and nothing after it was being made in five separate places, and is now made once in the module docstring. ~36 lines lighter with nothing dropped.

yyexela:
`test_shutdown_dataloader.py::test_dataloader_workers_do_not_tear_down_when_the_group_is_signalled` fails on macOS. The driver is passed as `python -c`, and macOS defaults to `spawn`, which re-imports `__main__` in each worker:

```
AttributeError: Can't get attribute '_Dataset' on <module '__main__' (built-in)>
RuntimeError: DataLoader worker (pid(s) 5232) exited unexpectedly
```

The behavior under test only exists under `fork`, where a worker inherits an already-installed handler. A spawned worker starts from a fresh interpreter with no handler, so there is nothing to decline. Forcing `multiprocessing_context="fork"` would test a loader production never builds, so I skipped on the precondition instead:

```python
_DEFAULT_START_METHOD = multiprocessing.get_start_method()


@pytest.mark.medium_duration
@pytest.mark.skipif(
    _DEFAULT_START_METHOD != "fork",
    reason=(
        f"DataLoader workers default to {_DEFAULT_START_METHOD} here, so no "
        "worker inherits the handler this test checks is declined"
    ),
)
def test_dataloader_workers_do_not_tear_down_when_the_group_is_signalled(tmp_path):
```

Reading the start method rather than checking `sys.platform` also covers Python 3.14, where Linux defaults to `forkserver`. Runs unchanged on CI.

yyexela:
(got a failure using this branch locally but you can ignore if the issue doesn't actually matter)

jpdunc23:
> (got a failure using this branch locally but you can ignore if the issue doesn't actually matter)

Thanks @yyexela ! Fixed. We might want to target macOS for (some subset of) our CI tests.

PR #1402, opened 2026-08-03
Bump to python=3.12 and torch=2.10

oliverwm1:
Bump to python=3.12 and torch=2.10 in our shared `fme` conda environment, docker image and CI testing environments. This PR does not change the dependencies or minimum required version of `fme` package. Bumping to python=3.12 will let us use the latest version of zarr-python.

Changes:
- update `Makefile`, CI workflows, constraints.txt and Dockerfile and related scripts to get python=3.12

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,10 @@
-FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime AS base
+# FLAVOR is `runtime` for the production image, and `devel` for the deps-only
+# image, which needs the nvcc shipped in the devel variant to build the CUDA
+# extensions. This tag provides Python 3.12 on Ubuntu 24.04; keep the torch
+# version in constraints.txt in sync with it.
+ARG PYTORCH_TAG=2.10.0-cuda12.8-cudnn9
+ARG FLAVOR=runtime
+FROM pytorch/pytorch:${PYTORCH_TAG}-${FLAVOR} AS base
 
 ENV FME_DIR=/full-model
 ENV DGLBACKEND=pytorch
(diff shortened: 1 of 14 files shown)


docker/install_cuda_extensions.sh
```diff
@@ -1,19 +1,55 @@
...
+CUDA_APT_SUFFIX=12-8
+CUDA_INSTALL_DIR=/usr/local/cuda-12.8
+
 echo "=== Installing CUDA extensions ==="
 
-# Step 1: Install CUDA toolkit via conda
+# Step 1: Install CUDA toolkit via apt
 echo ""
-echo "[1/3] Installing CUDA toolkit (nvcc) via conda..."
-conda install -y -c nvidia cuda-nvcc=12.8 cuda-toolkit=12.8
```
oliverwm1, line 16: conda is no longer available in the base image, hence why this script had to change.

docker/Dockerfile
```diff
@@ -14,10 +14,14 @@ RUN apt-get update -y && apt-get install -y \
 # Install gcloud- used for monthly netcdf data processing script
 # https://cloud.google.com/sdk/docs/install#deb
 RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | \
-  tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
-  apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - && apt-get update -y && apt-get install google-cloud-cli -y
```
oliverwm1, line 18: according to claude: "apt-key is gone in Ubuntu 24.04, so dearmor the key into the keyring directly."

AnnaKwa (approved):
The apex build worked ok with these changes. Claude pointed out that in the updated deps-only image tag c9fb617a4 the nvcc source was from the base vs in the current head it's installed via apt cuda-toolkit-12-8, then purged. Not sure if that is actually an issue but to be careful you could rebuild the image and update that file.

oliverwm1:
> The apex build worked ok with these changes. Claude pointed out that in the updated deps-only image tag [c9fb617](https://github.com/ai2cm/ace/commit/c9fb617a42d0c8200851f9477ef074ee0dc21d5d) the nvcc source was from the base vs in the current head it's installed via apt cuda-toolkit-12-8, then purged. Not sure if that is actually an issue but to be careful you could rebuild the image and update that file.

Thanks—I think I just did the image build before actually committing the dockerfile changes, hence Claude's understandable confusion. But to be safe I rebuilt and updated the image tag with the latest commit on this branch.

PR #1404, opened 2026-08-03
Carry corrector step diagnostics through coupled inference

jpdunc23:
Coupled inference now surfaces the same corrector diagnostics that single-component inference does, per realm. Each component stepper's per-step correction delta is carried through `CoupledStepper` onto that realm's prediction data, so a coupled run whose ocean and/or atmosphere stepper has a corrector writes `<realm>/step_diagnostics/correction_deltas.nc` when that realm's writer config enables it, and logs the correction metrics normalized with that realm's stepper normalizer. A realm whose stepper has no corrector stays silent; a corrector that declares variables but leaves them unchanged reports a zeros series, matching the single-component behavior. The existing `fme.ace` step-diagnostics writer and correction-metrics aggregator are reused once per realm rather than reimplemented, so there is no coupled-specific diagnostics container.

Defaults preserve current coupled behavior: no new files, unchanged rollout values, and no writer config surface added. Correction metric scalars are on by default, so a corrector-equipped coupled run gains those logged scalars — in coupled inference, the coupled evaluator, and coupled inline validation, which all build through the same two configs — and nothing else changes.

Changes:
- `fme.coupled.stepper.ComponentStepPrediction` carries the component step's corrector diagnostics, threaded from both the inner atmosphere and the outer ocean yield sites.
- `fme.coupled.stepper.CoupledStepper._predict` returns the prognostic state alongside the prediction and attaches each realm's stacked series after the derived-variable windowing, which is the only point where the series and the data are time-aligned. The `get_end` block that `predict` and `predict_paired` each duplicated moves into it.
- `fme.coupled.data_loading.batch_data.CoupledBatchData.with_step_diagnostics` attaches a per-realm series, re-validating each series against its batch.
- `fme.coupled.data_loading.batch_data.CoupledPairedData.from_coupled_batch_data` delegates each realm to `PairedData.from_batch_data` instead of hand-copying fields, which forwards `step_diagnostics` and fixes a pre-existing silent drop of `n_ensemble` to its default of 1.
- `fme.coupled.aggregator.InferenceEvaluatorAggregatorConfig` and `fme.coupled.aggregator.InferenceAggregatorConfig` gain a `step_diagnostics` field applied to both realms; the latter's `build` gains optional `ocean_normalize` / `atmosphere_normalize` arguments, mirroring the `fme.ace` `InferenceAggregatorConfig.build` signature it wraps once per realm. Coupled inference supplies both from each component stepper's normalizer.
- `fme.coupled.aggregator.InferenceEvaluatorAggregatorConfig.build` gains an `enable_time_series` argument, and coupled inline validation passes `False` — mirroring the `fme.ace` training path — so the correction per-step series (whose contiguous-window time bookkeeping inline validation violates) is never built there, while the time-mean correction scalars remain.

`pr-plan.md` at the repo root is the API roadmap for the change, including the constraint that an atmosphere corrector cannot touch the `OceanConfig`-prescribed surface temperature variable, and the test-setup threading the end-to-end tests need.

- [x] Tests added

### DIFF
--- a/pr-plan.md
+++ b/pr-plan.md
@@ -1,0 +1,321 @@
+# Carry corrector step diagnostics through coupled inference
+
+Threads each component stepper's per-step correction `delta` through `CoupledStepper` onto
+that realm's prediction data, and surfaces the existing step-diagnostics writer and
+correction-metrics aggregator per realm in the coupled inference configs. No new
+coupled-specific diagnostics container: the `fme.ace` components are reused once per realm.
+
+---
+
+## `fme/coupled/stepper.py` (modified)
+
+```python
+class ComponentStepPrediction:
+    def __init__(
+        self,
+        realm: Literal["ocean", "atmosphere"],
+        data: TensorDict,
+        step: int,
+        stepper_state: StepperState | None,
+        corrector_diagnostics: CorrectorDiagnostics,  # NEW
+    ):
+        ...
+
+    # Required, not defaulted: this class is the seam where the component
+    # StepOutput's diagnostics are currently dropped, so a new yield site must be
+    # forced to say what it carries rather than silently defaulting to empty.
+
+    @property
+    def corrector_diagnostics(self) -> CorrectorDiagnostics:  # NEW
+        ...
+```
+
+```python
+class CoupledStepper:
+    def get_prediction_generator(
+        self,
+        initial_condition: CoupledPrognosticState,
+        forcing_data: CoupledBatchData,
+        optimizer: OptimizationABC,
+    ) -> Generator[ComponentStepPrediction, None, None]:
+        # CHANGED — both yield sites (the inner atmosphere loop and the outer
+        # ocean step) pass the component StepOutput's corrector_diagnostics
+        ...
+
+    def _process_prediction_generator_list(
+        self,
+        output_list: list[ComponentStepPrediction],
+        forcing_data: CoupledBatchData,
+    ) -> CoupledBatchData:
+        # CHANGED — the per-realm StepOutput reconstruction carries
+        # corrector_diagnostics instead of defaulting it to empty. Attaches
+        # nothing: the returned batch still passes through time-changing ops.
+        ...
+
+    def _predict(
+        self,
+        initial_condition: CoupledPrognosticState,
+        forcing: CoupledBatchData,
+        compute_derived_variables: bool = False,
+    ) -> tuple[CoupledBatchData, CoupledPrognosticState]:
+        # CHANGED — also returns the end state (was: prediction only), so the
+        # stacked per-realm diagnostics can be attached after get_end.
+        # Rejected: attaching in predict() and predict_paired() separately, which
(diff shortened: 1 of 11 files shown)


pr-plan.md
```diff
@@ -0,0 +1,321 @@
...
+        # nothing: the returned batch still passes through time-changing ops.
+        ...
+
+    def _predict(
+        self,
+        initial_condition: CoupledPrognosticState,
+        forcing: CoupledBatchData,
+        compute_derived_variables: bool = False,
+    ) -> tuple[CoupledBatchData, CoupledPrognosticState]:
+        # CHANGED — also returns the end state (was: prediction only), so the
```
jpdunc23: More clear: `also returns the prognostic state`

pr-plan.md
```diff
@@ -0,0 +1,321 @@
...
+    ...
+```
+
+---
+
+## Open Questions
+
+- The correction-metrics granularity is one `step_diagnostics` field applied to both realms.
+  Worth per-realm fields instead if enabling `correction_maps` for only the ocean is a real
+  use case?
```
jpdunc23: We can live without granular per-realm configurability for now.

pr-plan.md
```diff
@@ -0,0 +1,321 @@
...
+- `InferenceAggregatorConfig.build` takes the two normalizers as optional keyword arguments
+  to match the single-component signature and keep existing call sites working. Making them
+  required would force every caller to opt in explicitly — preferable?
```
jpdunc23: What did the `fme.ace` version do? Mirror.

pr-plan.md
jpdunc23: Done — the `_predict` comment now reads "also returns the prognostic state (was: prediction only)", and the two other places that said "end state" now say prognostic state too.

pr-plan.md
jpdunc23: Agreed — dropped the open question and recorded the single shared field as a decision under Design choices. Worth noting it isn't only deferral: `correction_maps` is the only setting that would want per-realm control, and `CorrectionDeltaTimeMeanAggregator.get_dataset` hard-codes `dims = ("lat", "lon")`, so per-realm maps would break on a realm whose horizontal dims differ. Splitting the field later is a config-breaking change, now stated as accepted.

pr-plan.md
jpdunc23: `fme/ace/aggregator/inference/main.py`'s `InferenceAggregatorConfig.build` takes `normalize: NormalizeFn | None = None` — optional. Mirrored: one optional normalizer per realm, now a decision under Design choices rather than an open question.

Also corrected the stated reason. It said "keep existing call sites working", which was wrong — coupled inference is the only production call site of that build, and this PR changes it. The real reason is signature parity with the `fme.ace` config it wraps twice, and the availability rule that makes optional safe lives in `StepDiagnosticsMetricConfig.build` rather than in the caller.

Issue #1273, opened 2026-06-12
Corrector training features: pre-corrector loss optimization and corrector regularization


jpdunc23:
## What to build

Two corrector training features, unified behind a single `StepOutputLoss` object (per Jeremy's suggestion): a Loss-like object that takes a `StepOutput` and folds in both the pre-corrector loss target and the correction-magnitude penalty, jointly configurable. Landed as two commits in one PR. This is PR 3 of a 3-PR split of #1218 and #1222, depending only on #1271.

Both features operate in the stepper's loss-normalized space (`get_loss_normalizer`), consistent with how `StepLoss` already normalizes, so there is no normalizer mismatch between the main loss and the penalty.

### `StepOutputLoss` (in `fme/core/loss.py`, alongside `StepLoss`)

Wraps the existing `StepLoss` (the main loss, unchanged) plus an optional regularization `WeightedMappingLoss`, a reg weight, and an optional pre-corrector exclusion matcher.

- Call signature: `__call__(step_output, target_step, step, data_mask) -> StepOutputLossResult`, unfolding the ensemble dim internally (n_ensemble fixed at build).
- `StepOutputLossResult` exposes `.main` (a `LossOutput`), `.regularization` (a `Tensor | None`), and `.total() == main.total() + regularization`.
- Exposes `needs_uncorrected_grad: bool` (true iff either feature is configured), used by the stepper to decide detaching.

### Commit 1 — pre-corrector optimization (from #1222)

- Realized inside `StepOutputLoss.__call__`: the main-loss target overrides corrected values with `step_output.uncorrected` for every corrector-modified variable not matched by the exclusion matcher, then defers to the wrapped `StepLoss`.
- `exclude_names_and_prefixes` uses the shared name-and-prefix matcher consolidated with the spatial-masking convention (bare name matches 2D + all 3D levels, trailing-underscore prefix matches all levels, explicit `name_<level>` matches exactly).
- The exclusion matcher and its warn-once validation of unmatched entries live on the loss object, since it is what sees `step_output.uncorrected.keys()` each step.
- Rollout state and returned predictions always use fully corrected outputs; only the loss target changes.
- Adds a new `corrector_enabled: bool` abstract property on `StepABC`, implemented by every step type. The `TrainStepper` raises at build time if `corrector_loss` is configured but `corrector_enabled` is `False` (a config mistake that would otherwise train standard with no signal). A first-step warn-once covers the case where a corrector runs but `uncorrected` comes back empty.

[...]

PR #1405, opened 2026-08-03
Carry per-variable correction categories on CorrectorDiagnostics

jpdunc23:
A corrector's `delta` says how much each variable was adjusted but not what kind of adjustment it was, and the corrector-optimization work (ai2cm/ace#1273) needs that distinction to pick sensible defaults for which corrected variables to train against. Each correction now labels every field it writes with the facts that hold for that write, and `CorrectorDiagnostics` carries those labels alongside `delta`, keyed identically.

The facts are defined by what the delta *is*, since that is what a consumer reads. Exactly one of three holds per write: `budget` means the delta is the network's own violation of a constraint over its own fields, `overwrite` means it is disagreement with a target built from other fields in which the network's value for that variable plays no part, and `residual` means it is a term the network does not control and so does not depend on that value at all. The discriminator is two questions — does the delta depend on the network's value for this variable, and if so, what makes it vanish — which settles the cases a looser wording leaves ambiguous: an exogenous constant inside a budget target (`unaccounted_heating`, `ConserveDryAir`'s IC-seeded reference mass) is still `budget`, because the delta there still moves with the network's value. A fourth fact, `straight_through`, records that the write went through the straight-through estimator and combines with one of the other three. A label describes the operation rather than the grid cells, and is not conditional on the delta being nonzero: `ForcePositive` and `SurfaceEnergyFluxCorrection` both write unconditionally.

The labels are per field because three corrections write fields in different categories in a single call. `IceBudgetCorrection` rebalances the network's budget terms and then reconstructs each prognostic from them, discarding the network's own prediction. `MoistureBudgetCorrection` scales the precipitation or evaporation rate by a global-mean ratio, leaving the network's field in place, but recomputes the advection tendency as `twp_total_tendency - (evaporation - precipitation)`, which discards the network's own advection tendency entirely — a reconstruction exactly parallel to the ice prognostics, not a budget adjustment — and separately clips the frozen precipitation rate to the total. `SeaIceFractionCorrection` clamps the fraction while separately zeroing ice-dependent fields where the corrected fraction is zero, so even a correction uniform in the exclusive fact is not uniform in the set. They are sets because `delta` is already the cumulative effect when several corrections touch one variable, so the labels compose the same way the quantity they describe does; `CorrectionSequence` unions them and has no error path.

`SurfaceEnergyFluxCorrection` is labeled from its method. Under `residual_prediction` the corrected value is `net_flux * ocean_fraction + gen_hfds`, so `delta = net_flux * ocean_fraction` contains no `hfds` at any ocean fraction and nothing the network could produce would change it. Under `prescribed` the corrected value is `net_flux * ocean_fraction + gen_hfds * (1 - ocean_fraction)`, so the delta vanishes exactly when the network agrees with `net_flux`, with the ocean fraction only scaling how much of that disagreement is charged.

This is plumbing with no consumer: no config surface, no change to any corrector's numerical behavior, and existing configs and checkpoints are unaffected because the labels are derived from code rather than from config or checkpoint state. Correction metrics stay label-blind, and the labels stop at `StepOutput.corrector_diagnostics` — `StepDiagnostics`, the inference and serialization surface, still carries `delta` only.

Changes:
- `fme.core.corrector.registry.Correction` returns a new `fme.core.corrector.output.CorrectionResult` in place of a `(modified, corrector_state)` tuple, carrying a set of categories per written field and raising `ValueError` for a write that does not declare exactly one of `budget`, `overwrite`, or `residual`
- `fme.core.corrector.output.CorrectorDiagnostics` gains a `categories` mapping and validates in `__post_init__` that its keys equal `delta`'s; the union across corrections is deliberately not held to the per-write rule, since `{budget, overwrite}` on one name is legitimate there
- `fme.core.corrector.output.build_corrector_diagnostics` takes a name-to-categories mapping in place of its `touched_names` iterable

- [ ] Tests added

### DIFF
--- a/pr-plan.md
+++ b/pr-plan.md
@@ -1,7 +1,7 @@
 # Carry per-variable correction categories on `CorrectorDiagnostics`
 
 Each `Correction` labels every variable it writes with the facts that hold for that write — what
-happened to the network's value, and whether the straight-through estimator was involved — and
+the delta means, and whether the straight-through estimator was involved — and
 `CorrectorDiagnostics` carries those labels per name alongside `delta`. Pure plumbing: no config
 surface, no consumer, no change to any corrector's numerical behavior.
 
@@ -16,17 +16,20 @@ CorrectionCategory = Literal["budget", "overwrite", "residual", "straight_throug
 These are independent facts, not mutually exclusive kinds: a variable carries the
 set of them that hold, and each has exactly one consequence for a consumer.
 
-Exactly one of these three answers "what happened to the network's value":
+The three below are defined by **what the delta is**, because that is what a
+consumer reads. Exactly one holds per write:
 
-- ``budget``: it was adjusted to satisfy a conservation or budget constraint. The
-  corrected value still carries the network's value, and the delta is the
-  network's violation of the constraint.
-- ``overwrite``: it was replaced by a value built from other fields -- a derived
-  reference, a reconstruction from tendencies, a mask, or a bound. The delta
-  measures disagreement with that replacement, which is a proxy for the truth
-  rather than the truth itself.
-- ``residual``: a term the network does not control was added to it. The network's
-  value is a residual against that term, so the delta does not depend on it.
+- ``budget``: the delta is the network's own violation of a constraint computed
+  over the network's own fields. It vanishes when the network already satisfies
+  the constraint, and the corrected value is the network's value adjusted (offset
+  or scaled) to satisfy it.
+- ``overwrite``: the delta is the network's disagreement with a target computed
+  from other fields, with the network's value for this variable playing no part
+  in that target. It vanishes when the network already agrees. A derived
+  reference, a reconstruction from other tendencies, a mask, or a bound.
+- ``residual``: the delta is a term the network does not control. It does not
+  depend on the network's value for this variable and does not vanish for any
+  value the network could produce for it.
 
 ``straight_through`` is an orthogonal fact about the gradient path, and combines
 with one of the above:
@@ -39,11 +42,23 @@ with one of the above:
   optimization override would collide with a gradient path already there -- does
   not apply.
 
-A ``Literal`` alias rather than an ``Enum``: the repo's idiom for a closed set of
-string choices is ``Literal`` (e.g. ``SurfaceEnergyFluxCorrectionConfig.method``),
-and correctors are constructed through ``dacite.Config(strict=True)`` with no
-``cast`` list, so an ``Enum`` would need extra dacite configuration the moment a
-category becomes config-facing.
+``residual`` is the third distinct sense of that word within this call stack's
+reach, and does not mean either of the other two:
+
+- ``step_with_adjustments(residual_prediction=...)``: the *network* predicts a
+  tendency against its own previous state.
+- ``SurfaceEnergyFluxCorrectionConfig.method="residual_prediction"``: the
+  network's ``hfds`` is a residual on top of an atmosphere-derived flux.
+- ``CorrectionCategory`` ``residual``: this *delta* is an exogenous term, so it
+  says nothing about the network's value.
+
+The second implies the third and is currently its only instance; the first is
+unrelated. Renaming the category (``exogenous``, ``imposed``) was rejected:
+``residual`` is the accurate word for the relationship, and ``exogenous`` is
+already used for forcing variables in ``fme/coupled``.
+
+A ``Literal`` alias rather than an ``Enum``, because the repo's idiom for a closed
+set of string choices is ``Literal`` (e.g. ``SurfaceEnergyFluxCorrectionConfig.method``).
 """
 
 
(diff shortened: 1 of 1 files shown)


pr-plan.md
```diff
@@ -0,0 +1,455 @@
...
+``straight_through`` is an orthogonal fact about the gradient path, and combines
+with one of the above:
+
+- ``straight_through``: this write went through ``replace_value_keep_gradient``,
+  so the forward value is the corrected one while the gradient flows as if the
+  correction had not happened. It names the gradient path rather than the
+  operation: a bound enforced without the estimator is an ``overwrite`` and
+  nothing more, because the reason to distinguish it -- that an explicit
+  optimization override would collide with a gradient path already there -- does
+  not apply.
```
jpdunc23, line 43: A bit verbose and obtuse. Please revise and speak plainly.

pr-plan.md
```diff
@@ -0,0 +1,455 @@
...
+---
+
+## `fme/core/corrector/output.py` (modified)
+
+```python
+CorrectionCategory = Literal["budget", "overwrite", "residual", "straight_through"]  # NEW
+"""A fact about what a correction did to one variable it wrote.
+
+These are independent facts, not mutually exclusive kinds: a variable carries the
+set of them that hold, and each has exactly one consequence for a consumer.
```
jpdunc23, line 17: This could be replaced with "Multiple categories may apply simultaneously."

pr-plan.md
```diff
@@ -0,0 +1,455 @@
...
+  operation: a bound enforced without the estimator is an ``overwrite`` and
+  nothing more, because the reason to distinguish it -- that an explicit
+  optimization override would collide with a gradient path already there -- does
+  not apply.
+
+A ``Literal`` alias rather than an ``Enum``: the repo's idiom for a closed set of
+string choices is ``Literal`` (e.g. ``SurfaceEnergyFluxCorrectionConfig.method``),
+and correctors are constructed through ``dacite.Config(strict=True)`` with no
+``cast`` list, so an ``Enum`` would need extra dacite configuration the moment a
+category becomes config-facing.
```
jpdunc23: Fine as commentary on the plan, but not necessary in the final docstring.

pr-plan.md
```diff
@@ -0,0 +1,455 @@
...
+"""
+
+
+@dataclasses.dataclass
+class CorrectionResult:  # NEW
+    """What one ``Correction`` did to ``gen_data`` in a single call.
+
+    Replaces the ``(modified, corrector_state)`` tuple so a correction can label
```
jpdunc23, line 69: `Replaces the ``(modified, corrector_state)`` tuple`  is a PR dev process comment and doesn't belong in a docstring.

pr-plan.md
```diff
@@ -0,0 +1,455 @@
...
+    def __post_init__(self) -> None:
+        """Reject keys that differ from ``modified``'s, and any write not carrying
+        exactly one of ``budget``/``overwrite``/``residual``.
+
+        A single write cannot both preserve and replace the network's value, nor
+        both depend and not depend on it, so a set holding two of the three is an
+        authoring error. Requiring one rather than allowing none also means a
+        write can never be labeled ``{"straight_through"}`` alone, which would say
+        nothing
+        about how a consumer should treat it.
```
jpdunc23: Too verbose for a docstring. Write `pr-plan.md` docstrings as if they were part of the final PR merge candidate. Commentary on the plan goes outside of code blocks.

pr-plan.md
mcgibbon: FWIW I've been trying GPT 5.6 in some of my personal projects and it's much less prone to these... very hard to parse sentences.

pr-plan.md
```diff
@@ -0,0 +1,671 @@
...
+
+Not part of this PR, but the reason there are four. Each fact carries one consequence, and the
+intended consumer (corrector-aware training) composes them by membership:
+
+| fact | consequence for the consumer |
+|---|---|
+| `budget` | its presence is the reason to train against this variable |
+| `overwrite` | not by itself a reason to train against it — and **not** a veto over `budget` |
+| `residual` | a veto: never train against it by default, and an explicit override is incoherent, not merely off — the delta does not depend on the network's value, and the value is a residual against a derived term rather than an estimate of the reference |
+| `straight_through` | an explicit override collides with the gradient path already there — guard it |
```
mcgibbon, line 263: ST feels like a flag, not a category. Does this mean a change could have multiple of these facts?

It's not clear to me whether clipping frozen precipitation to not exceed total precipitation would have budget, overwrite, or residual attached to it - arguments could be made for or against for all three of them. I'm not sure that this is a useful set of flags, both because it's not clear when one is applied and it's not clear what it being applied would mean (there's no clear effect of these flags in the plan, just bookkeeping).

We had previously discussed your completing the work for pre-corrector optimization and correction regularization, after which I could make PRs to introduce a new optimized-derived-output concept. I would still like to complete that work. I have a strong idea of what that design should look like - I haven't been implementing it because I'm avoiding introducing merge conflicts to your ongoing corrector work.

My general feeling is that all of the actions of the corrector can be cleanly split into "this is a field derived from network output fields, which you never would define as a delta to be regularized" and "this is a correction, which is a candidate to regularize". All existing examples we have discussed so far fit into this split and make a lot more sense code-wise once we do, and we haven't thought of a counter-example which does not.

Past that, some of the ideas in this plan could potentially be applied in how we configure which deltas to include or not include in a regularization, but I think we should defer that discussion.

jpdunc23:
This plan is superseded by the https://github.com/ai2cm/ace/pull/1426 opt-in configuration approach.

PR #1407, opened 2026-08-04
Agree a batch boundary across ranks so a preempted job leaves the training loop together

jpdunc23:
## READ FIRST — decisions that are the maintainer's, not this branch's

Three questions this change cannot settle from inside itself. Each changes what ships, so each wants an answer before merge rather than after.

1. **Does the per-batch agreement need a config switch to disable it?** Today there is none: every training job, preempted or not, pays one blocking host-side gloo all-reduce per iteration on the critical path. A switch buys an escape hatch if the cost bites at production world size; it costs a supported configuration in which the fabric fault returns. The only measurement is 630 µs at 8 gloo ranks on an idle workstation over loopback, with no multi-node figure anywhere in the branch.
2. **Should the teardown/checkpoint split inside torchrun's 30 s window be re-cut?** `DEFAULT_TEARDOWN_TIMEOUT` claims 20 s of the 30 s, which leaves the restart checkpoint nothing once a rank reaches the boundary 7 s or more after its signal — the expected case for a preemption landing inside a multi-GB periodic checkpoint write. Cutting the ceiling to about 5 s and giving the checkpoint 20 s would fix that, and `_ABORT_BACKSTOP` already bounds a hung teardown at the ceiling plus 5 s; against it, a teardown that genuinely needs more than 5 s would then be aborted rather than allowed to finish. The arithmetic on both sides is derived rather than measured — no test waits out a real budget or a real teardown.
3. **Is the watchdog's new `ProcessGroup.abort()` wanted before any GPU evidence?** It is added to the one path whose purpose is avoiding a fabric fault, and whether an aborted communicator leaves the fabric better off than one dropped by `os._exit` is unverified; it may fault the fabric sooner instead. `os._exit` remains as a named backstop, so removing the abort is a one-line reversal either way.

---

A preempted whole-node training job faults its own GPU fabric on the way out, and the node is then cordoned. Internal analysis of preemptions on our H100 clusters over a two-week window found that 109 of the 145 preemptions which interrupted the training loop of an 8-GPU job faulted the fabric — 75%, scoped to 8-GPU jobs specifically, and counting priority preemptions and drains together. The mechanism is that one rank exits while a peer still has a collective in flight; the peer's next NVLink access is refused, the driver raises an unrecoverable NVSwitch error, and Beaker cordons the node, which kills co-tenants as well as the preempted job.

ai2cm/ace#1398 made the teardown orderly once a rank runs its handler, and said what it could not fix: a rank whose main thread is inside a C-level collective never services the signal, so it never starts the teardown, and a rank that did start one cannot finish it while such a peer still holds the communicator. In a deliberate 4-rank preemption during that change, two ranks hung in the teardown until the 20 s watchdog exited them and two never ran their handler at all.

This change makes the ranks agree on a batch boundary and leave the training loop together, so no rank leaves while a peer still has a collective in flight. A termination signal arriving inside the loop is recorded rather than acted on; every rank then finishes its current batch, exchanges a stop flag and its batch index at the boundary, and tears the backend down having agreed. Because no rank leaves early, a rank that was blocked in a collective when the signal arrived returns to the interpreter, runs its own pending handler, and reaches the same boundary. Outside the loop the immediate teardown is unchanged.

The mechanism is a component any entrypoint can adopt, not a `Trainer` detail: a caller with its own batch loop, holding no `Trainer` and importing nothing from `fme.ace`, can adopt it using only public names from `fme.core.distributed`, and a test constructs exactly that caller. It does not, however, address the case for putting this in the launcher or the framework so that jobs from other codebases benefit — that cannot be answered from inside this repository.

Three things are deliberately true and worth stating rather than discovering later.

The agreement runs on a dedicated world-wide gloo group rather than on the NCCL group. An unbounded all-reduce on the NCCL group would hang the healthy ranks whenever one rank never enters it, which is worse than today's behaviour; and `reduce_max` on the spatial-parallel backend operates on the data-parallel subgroup, so a stop originating on a spatial co-rank would never reach its data-parallel peers. Each exchange carries its own deadline, because a gloo group can neither be created with a short timeout nor have its timeout narrowed afterwards — `_set_pg_timeout` accepts a gloo group and has no effect on it, since the timeout is read once at context construction.

The agreement group is never destroyed, and a rank that abandoned an exchange exits hard. `ProcessGroupGloo`'s destructor waits without bound for its work queue to drain, and it overrides neither `shutdown()` nor `abort()`, so no call reclaims a gloo group that has an abandoned exchange outstanding. The group is therefore held for the life of the process, and holding it is what makes the post-shutdown callbacks reachable when an exchange is abandoned: had the teardown dropped the group, the handler would never leave its collective phase and the restart checkpoint would be lost by construction. Anyone editing this code must not drop that reference — a superseded group is moved aside, never released.

Interpreter finalization is not a way out of it either. That destructor also runs during `Py_Finalize`, where it joins the worker thread still holding the abandoned operation, so `sys.exit` waits for the wedged peer's socket to close: measured on torch 2.7.1, 119.3 s against a peer parked for 120 s and 24.3 s against one parked for 25 s, versus 0.05 s with `os._exit`. Whether it joins or leaks depends on what else still references the group, so the wait is not merely long but unbounded and non-deterministic — a rank left to finalize is SIGKILLed by torchrun and the launcher reads a signal death rather than the `128 + signum` the design promises. So once an exchange has been abandoned the teardown ends in `os._exit`, taken after the backend is released and after the callbacks have run: unlike the watchdog's hard exit it drops no communicator and skips no work, and it is the whole of the design's bound on that path.

The interrupt-time restart checkpoint is load-bearing here. The agreed boundary is chosen so that the checkpoint written by the post-shutdown callback is consistent, and the cooperative stop's value depends on that write happening. Removing the callback registry would remove it.

What this does not fix. A rank wedged for a reason independent of the signal still times out the agreement, still forces the healthy ranks out, and can still fault; that residual rate is unknown and is not zero. A rank that dies without running Python at all — a segfault, the OOM killer, SIGKILL — is not covered, and neither is it by the exception path. The mechanism is scoped to the training loop, so a preemption landing in validation, inference or startup keeps today's behaviour. Interrupting an epoch-boundary checkpoint write is a known regression: a non-root rank inside the loop defers toward a boundary that root, being outside the loop, will not reach.

What stays argued rather than measured. **No preemption trial has been run against this branch on real hardware.** The mechanism's exit paths pass under NCCL in CI's GPU job, at two ranks, including the torchrun cases that observe a rank's process death; production runs eight, where signal skew is widest. But the fault this change exists to prevent has never been provoked against the fix: the intended 4-GPU trial did not run because both H100 clusters were at zero free GPUs, and the 8-GPU whole-node trial is deliberately gated on a maintainer's approval. So the claim that the fabric fault stops happening is a design argument plus CPU and two-rank NCCL evidence, not a measurement. Relatedly, whether aborting a communicator releases a wedged peer, as opposed to only the aborting rank's own main thread, is untested — the one test that reaches the abort stubs it — and it is the difference between the fault disappearing and merely preserving the checkpoint. The per-batch cost is argued structurally and from CPU timing rather than benchmarked; the exchange is one small CPU all-reduce per optimizer step, on a step that already ends in a blocking gradient all-reduce, and there is no switch to disable it — a supported configuration in which the fault returns would be worse than the cost, but whether that judgement holds is decision 1 above rather than something this branch settled.

Changes:
- `fme.core.distributed.cooperative_stop`: new module. `cooperative_stop` wraps a batch loop, and `CooperativeStop.agreed` exchanges a stop reason and batch index at each boundary, returning whether to stop. Carries `PeerStopRequested` and `UnequalLoopLength`.
- `fme.core.distributed.stop_agreement`: new module. `StopAgreement` with gloo and single-rank implementations, and the cached and uncached group constructors — the cache keyed on the identity of the default group the agreement belongs to, so a world that has been destroyed and re-initialised gets a new agreement group and the stale one is moved aside. `timeout_contract_verified` reports whether the installed torch is one of the releases the operation-timeout behaviour the short deadline rests on was read against. It never fails a job: group construction warns once, and a rank on an unverified release passes the default group's own timeout where it would have passed a short one, which is `main`'s behaviour rather than a new failure mode. `default_group_timeout` reads that figure off the group instead of restating it, because `DistributedManager` passes `init_process_group` no timeout and so takes NCCL's default 10 minutes on GPU.
- `fme.core.distributed.shutdown`: a termination signal inside a registered scope records intent instead of tearing down, via `defer_termination` and `PendingStop`; `terminate` performs the teardown and callbacks for the deferred and exception paths, ending in `os._exit` where the caller asked for it with `PendingStop.require_hard_exit` — on the exception path too, which prints the propagating traceback itself before exiting, since that traceback was the only reason that path previously declined to exit at all; the phase machine gains a stop-requested state in which a repeat signal is ignored, because the scheduler and torchrun each signal every rank; the teardown watchdog gains an abort ahead of its hard exit, and a cancelled flag so a released main thread is not exited mid-write; `write_marker` emits one line per event to stderr.
- `fme.core.distributed.Distributed.stop_agreement` and `Distributed.abort`, with implementations on each backend, and the agreement group created alongside the process group.
- `fme.core.generics.Trainer.train_one_epoch`: the batch loop runs inside `cooperative_stop` and agrees at each boundary. The periodic checkpoint write is skipped once a stop is pending, because it would delay the agreed stop by a whole multi-GB `torch.save` to record the index the terminate-time restart checkpoint is about to record after the teardown; and `save_restart_checkpoints_on_terminate` skips a write the periodic checkpoint has just made, which is the case where the signal landed inside that write. An epoch left at an agreed stop is not counted as complete: the epoch counter and the epoch's batch counter are together the resume point, so recording a truncated epoch as complete would silently skip the rest of its data on resume, and the loop does not rely on the exit having happened to avoid that. The truncated epoch's train-evaluation diagnostics are written under that epoch's own number rather than into the last completed epoch's subdirectory, which is where an un-incremented counter would have put them.
- Evidence in the container log: every rank emits its agreed batch index, which bound was in force on it, and its teardown returning, by a direct write to stderr, because non-root ranks are pinned to `logging.ERROR` and `logger.info` from them never reaches the log. The whole `fme-stop:` event vocabulary — every event name, which module writes it, what it means, and what a reader does with it — is listed in one place, in `fme.core.distributed.shutdown`'s module docstring.

- [x] Tests added

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -190,15 +190,22 @@ def reset_global_timer():
 
 @pytest.fixture(autouse=True)
 def reset_post_shutdown_callbacks():
-    # the registry is process-global and nothing clears it between tests -- the
-    # session-scoped context above installs no handler, so its usual
+    # both registries are process-global and nothing clears them between tests --
+    # the session-scoped context above installs no handler, so its usual
     # clear-on-exit never runs either. Without this, every Trainer built by an
-    # earlier test would still be called on a signal raised by a later one
-    from fme.core.distributed.shutdown import clear_post_shutdown_callbacks
+    # earlier test would still be called on a signal raised by a later one, and a
+    # deferral left behind by a failed test would make a later test's signal
+    # deferred to a scope nobody polls
+    from fme.core.distributed.shutdown import (
+        clear_pending_stop,
+        clear_post_shutdown_callbacks,
+    )
 
     clear_post_shutdown_callbacks()
+    clear_pending_stop()
     yield
     clear_post_shutdown_callbacks()
+    clear_pending_stop()
 
 
 @pytest.fixture(autouse=True)
--- a/fme/core/distributed/__init__.py
+++ b/fme/core/distributed/__init__.py
@@ -1,3 +1,10 @@
+from .cooperative_stop import cooperative_stop
 from .distributed import Distributed
+from .shutdown import add_post_shutdown_callback
 
-__all__ = ["Distributed"]
+# `add_post_shutdown_callback` is exported alongside the other two because an
+# adopter that wants work to run after the teardown -- writing a restart
+# checkpoint, say -- should not have to reach into
+# `fme.core.distributed.shutdown` for it. Those three names are the whole
+# adopter-facing surface of the cooperative stop.
+__all__ = ["Distributed", "add_post_shutdown_callback", "cooperative_stop"]
--- a/fme/core/distributed/base.py
+++ b/fme/core/distributed/base.py
@@ -6,6 +6,8 @@ from typing import TypeVar
 import torch
 import torch.nn as nn
 
+from .stop_agreement import StopAgreement
+
 T = TypeVar("T")
 
 
@@ -57,6 +59,33 @@ class DistributedBackend(ABC):
 
     @abstractmethod
     def reduce_max(self, tensor: torch.Tensor) -> torch.Tensor | None: ...
+
+    @abstractmethod
+    def stop_agreement(self) -> StopAgreement:
+        """The group ranks use to agree on leaving a loop together.
+
+        It must be **world-wide**, which is why it is not `reduce_max`: that is
+        data-parallel-only under a spatially-parallel backend, while the teardown
+        is world-wide, so reusing it would let a stop originating on a spatial
+        co-rank never reach its data-parallel peers.
+
+        Created with the backend and **not** destroyed with it. ``shutdown()`` is
+        a no-op on a gloo group and its destructor is unbounded, so the group is
+        held for the life of the process and disposed of by the process's exit.
+        """
+        ...
+
+    @abstractmethod
+    def abort(self) -> None:
+        """Abort this rank's communicators, for the teardown watchdog.
+
+        Called only when the collective teardown has overrun its deadline, in
+        place of dropping the communicators by a hard exit. It targets the
+        **default** group and never the agreement group: gloo's ``abort()`` is an
+        empty default, so this is NCCL-only in effect, and aborting the agreement
+        group would achieve nothing while the design needs it kept anyway.
+        """
+        ...
 
     @abstractmethod
     def gather(
--- a/fme/core/distributed/cooperative_stop.py
+++ b/fme/core/distributed/cooperative_stop.py
@@ -22,6 +22,25 @@ The whole adopter-facing surface is one context manager::
 Leaving the context on a stop performs the same teardown the signal handler would
 have performed -- the backend release, the post-shutdown callbacks, the exit code
 -- entered from the loop's boundary rather than from wherever the signal landed.
+**Every** path that leaves the loop ends that way: a stop read out of an exchange,
+a stop given up on, and an exception. A rank that broke out of the loop and carried
+on into another collective, whose peers had already left, is the fault this module
+exists to prevent.
+
+**What is covered is the one loop that adopts this, and nothing else.** A signal
+arriving during the validation pass, during inline inference, or inside
+``run_lr_tuning_trial`` is acted on where it lands, exactly as before this module
+existed -- and those paths issue aggregator reductions of their own, so a rank
+signalled there still tears the backend down from wherever it was. Widening the
+cover means wrapping those loops too; nothing here does it for them.
+
+**It costs a blocking host-side collective on every iteration** of the loop it
+wraps, over a world-wide gloo group, and there is no switch to turn it off. One
+exchange measured 630 us at 8 gloo ranks -- on an idle workstation, over loopback,
+with no figure above 8 ranks and nothing asserting it, so treat that as a lower
+bound of unknown looseness rather than a cost. It also removes the run-ahead
+today's async NCCL gradient all-reduce allows, capping a rank's host-side lead at
+one iteration.
 
 **Every deadline the design chooses lives here**, with the one module that
 consumes them, rather than split across the two below it in the layering.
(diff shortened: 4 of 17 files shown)


mcgibbon (commented):
My comment from standup can kind of be distilled into: Can we implement this safety with only changes in Distributed, or as much as possible with changes there? For example by having checks for collective shutdown whenever a worker attempts to enter a distributed operation? If we need this concern to exist in other parts of the code, why do we need it, and what's the minimal version of that?

jpdunc23:
> My comment from standup can kind of be distilled into: Can we implement this safety with only changes in Distributed, or as much as possible with changes there? For example by having checks for collective shutdown whenever a worker attempts to enter a distributed operation? If we need this concern to exist in other parts of the code, why do we need it, and what's the minimal version of that?

Will look into this, but my sense is no: collective ops happen outside of the `distributed` API, e.g. the most relevant is `loss.backward()` which reduces gradients. That's why this PR requires changes to `Trainer.train_one_epoch()`, to ensure the ranks are coordinated at the boundary between each batch.

mcgibbon:
> > My comment from standup can kind of be distilled into: Can we implement this safety with only changes in Distributed, or as much as possible with changes there? For example by having checks for collective shutdown whenever a worker attempts to enter a distributed operation? If we need this concern to exist in other parts of the code, why do we need it, and what's the minimal version of that?
> 
> Will look into this, but my sense is no: collective ops happen outside of the `distributed` API, e.g. the most relevant is `loss.backward()` which reduces gradients. That's why this PR requires changes to `Trainer.train_one_epoch()`, to ensure the ranks are coordinated at the boundary between each batch.

It's wild to me because I thought this is what `destroy_process_group` is supposed to do, i.e. prevent workers from continuing to attempt distributed operations and get all other workers to shut down. Seeing this PR it's still not clear to me what the issue is, how the changes here (attempting to ensure all workers complete the same batch count) solve it, or what (possibly simpler) alternatives exist. I think it's important we establish this before we merge such far-reaching code. I do see some parts of the code (or motivational comments) I don't think are required (e.g. worrying about bookkeeping for extra reads since the last distributed checkpoint).

loss.backward calls distributed operations because of DistributedDataParallel, does that mean the bug is in its implementation?

jpdunc23:
> It's wild to me because I thought this is what `destroy_process_group` is supposed to do, i.e. prevent workers from continuing to attempt distributed operations and get all other workers to shut down. Seeing this PR it's still not clear to me what the issue is, how the changes here (attempting to ensure all workers complete the same batch count) solve it, or what (possibly simpler) alternatives exist. I think it's important we establish this before we merge such far-reaching code. I do see some parts of the code (or motivational comments) I don't think are required (e.g. worrying about bookkeeping for extra reads since the last distributed checkpoint).
> 
> loss.backward calls distributed operations because of DistributedDataParallel, does that mean the bug is in its implementation?

Some key points extracted from Fable 5's overly verbose reply to your comment:

> `destroy_process_group` doesn't do either thing you describe. It sends no message to other ranks and doesn't stop anyone from attempting collectives — it's a local resource release that the docs require every rank to call after all in-flight collectives complete. [...] Calling it (or just exiting) while a peer is still inside an all-reduce is the fault

> The issue in one sequence: `SIGTERM` lands on all ranks at slightly different times, and a rank only sees it when its main thread returns to the interpreter. A rank blocked in a collective returns only when its peers feed that collective.

> Why agreeing on a batch count solves it: NCCL already requires every rank to issue the identical collective sequence in identical order [...] If every rank finishes batch `k` and none starts `k+1`, every in-flight collective has all its participants and completes, and teardown starts with an empty pipeline. Stop-at-step-boundary is the standard pattern for this (torchft, NeMo's preemption callback)

> Is the bug in DDP? No. `backward()` is just where our densest collectives live [...] PyTorch's own answers to uncoordinated exit are restart-after-death (torchelastic) and timeout-then-abort (the NCCL watchdog) — and the abort path is what we watch fault the fabric today.

jpdunc23:
8-GPU production-like beaker test run: https://beaker.org/orgs/ai2/workspaces/climate-titan/work/01KZ7Z6DTSH0NBYWAHDM5RFDCG?taskId=01KZ7Z6DV2874D82R7B1DBJSS2&jobId=01KZ7Z6DYFWZGSZMGX9TQH0YYH

mcgibbon:
I had a chat with Fable and it sounds like most of the complexity of the solution in this PR is needed only so that we can ensure consistency of the model across ranks when writing the checkpoint that gets written on-abort. Otherwise we could use NCCL abort instead. I'd propose we consider going that route and remove the write-checkpoint-on-interrupt machinery altogether (it hasn't been working well anyways, usually falling back on the write-every-n-batches).

I have Fable drafting a PR so we can see how much more minimal this might be. Also it's not 100% guaranteed to work and would need to be tested:
>  The open question a simpler design must answer empirically is whether abort (on each rank, potentially invoked from a non-main thread while the main thread is wedged in the collective) reliably quiesces the fabric without itself faulting it.

PR #1409, opened 2026-08-04
Resolve inline inference epochs against the training epochs only

### LINKED ISSUE #1408: `evaluate_before_training: true` silently skips inline inference when `max_epochs` is odd
`TrainConfig.get_inference_epoch_sets` puts epoch 0 into the candidate list *before* applying each entry's `epochs` slice:

```python
start_epoch = 0 if self.evaluate_before_training else 1
all_epochs = list(range(start_epoch, self.max_epochs + 1))
return [set(all_epochs[entry.epochs.slice]) for entry in inference]
```

The method is identical in `fme/ace/train/train_config.py` and `fme/coupled/train/train_config.py`. Two defects follow from that one line.

### 1. Pre-training inference depends on `max_epochs` parity

With `epochs: {start: -1, step: -2}` — the last epoch and every other one working back — the candidate list has `max_epochs + 1` entries when the flag is set, so epoch 0 is selected only when `max_epochs` is even.

| `max_epochs` | resolved epochs | runs at epoch 0 |
| --- | --- | --- |
| 15 | 1, 3, …, 15 | no |
| 20 | 0, 2, …, 20 | yes |

The skip is silent. `Trainer.train` logs `Starting inline inference before training` before consulting the epoch set, so a skipped run is indistinguishable from a real one in the logs except by elapsed time — 4 ms against roughly an hour in two runs that differed only in `max_epochs`.

### 2. The slice's meaning depends on the flag

`epochs: {step: 2}` at `max_epochs: 6` selects training epochs 2, 4, 6 under the coupled trainer (flag defaults true) and 1, 3, 5 under ACE (flag defaults false). A flag that should only add a pre-training evaluation re-phases the whole training schedule. Both behaviors are currently pinned by tests: `test_get_inference_epoch_sets_per_config` in `fme/coupled/train/test_train_config.py`, and `test_get_inference_epoch_sets_per_config_zero_weight` in `fme/ace/train/test_train_config.py`.

jpdunc23:
`evaluate_before_training: true` only produced a pre-training inference pass when `max_epochs` was even: `get_inference_epoch_sets` prepended epoch 0 *before* applying each entry's positional `epochs` slice, so an end-anchored slice such as `epochs: {start: -1, step: -2}` reached epoch 0 only at even parity — silently, since the trainer logs `Starting inline inference before training` before consulting the epoch set. The same prepend made the slice's meaning depend on the flag: at `max_epochs: 6` with `epochs: {step: 2}`, the coupled package (flag true by default) selected training epochs 2, 4, 6 while ACE (flag false) selected 1, 3, 5. Each entry's slice now resolves against the training epochs `1..max_epochs` regardless of the flag, and epoch 0 is unioned into every entry's set — no per-entry opt-out, including `weight: 0` entries and entries whose slice resolves empty — when the flag is true. That is how ACE already behaved, so the coupled package now agrees with ACE rather than adopting a new convention, but coupled configs with a start-anchored slice do change which epochs they run inference on. One config class newly fails at startup: an ACE config whose `weight > 0` entries shared a schedule only because epoch 0 shifted their phase now trips `_validate_weighted_inference_epochs` instead of selecting checkpoints on an incomparable metric.

Changes:
- `fme.ace.train.train_config.TrainConfig.get_inference_epoch_sets` and `fme.coupled.train.train_config.TrainConfig.get_inference_epoch_sets`: resolve each entry's `epochs` slice against the training epochs `1..max_epochs`, then union epoch 0 into every returned set when `evaluate_before_training` is true. The signature and return type are unchanged, and `get_inference_epochs` still derives the union.
- `fme.ace.train.train_config.InlineInferenceConfig.epochs` and `fme.coupled.train.train_config.InlineInferenceConfig.epochs` docstrings: state that the value is a positional slice over the training epochs and that epoch 0 is additionally included when `evaluate_before_training` is true.
- Tests in both packages covering `max_epochs` parity under a true flag, the slice's independence from the flag, an entry whose slice selects no training epoch, and a guard that epoch 0 stays absent when the flag is false. Both config-building test helpers now accept `evaluate_before_training`, defaulting to each class's own default.

- [x] Tests added

Resolves #1408

### DIFF
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -120,7 +120,10 @@ class InlineInferenceConfig:
         forward_steps_in_memory: number of forward steps to take before
             re-reading data from disk
         n_ensemble_per_ic: number of initial condition based ensembles
-        epochs: epochs on which to run inference. By default runs inference every epoch.
+        epochs: positional slice over the training epochs 1 to max_epochs, so
+            index 0 selects epoch 1. By default runs inference every training
+            epoch. Epoch 0 is additionally included when
+            evaluate_before_training is true.
         aggregator: configuration of inline inference aggregator.
         name: name used as wandb log prefix and output subdirectory. If None,
             defaults to "inference" when there is a single inference config
@@ -557,9 +560,12 @@ class TrainConfig:
         inference = self.inference_list
         if not inference:
             return []
-        start_epoch = 0 if self.evaluate_before_training else 1
-        all_epochs = list(range(start_epoch, self.max_epochs + 1))
-        return [set(all_epochs[entry.epochs.slice]) for entry in inference]
+        training_epochs = list(range(1, self.max_epochs + 1))
+        pre_training_epochs = {0} if self.evaluate_before_training else set()
+        return [
+            set(training_epochs[entry.epochs.slice]) | pre_training_epochs
+            for entry in inference
+        ]
 
     def get_inference_epochs(self) -> list[int]:
         epoch_sets = self.get_inference_epoch_sets()
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -383,8 +383,8 @@ class TrainConfig:
             must be run in segments, e.g. due to wall clock limit.
         save_per_epoch_diagnostics: Whether to save per-epoch diagnostics from
             training, validation and inline inference aggregators.
-        evaluate_before_training: Whether to run validation and inline inference before
-            any training is done.
+        evaluate_before_training: Whether to run all validation and inline
+            inference before any training is done.
         save_best_inference_epoch_checkpoints: Whether to save a separate checkpoint
             for each epoch where best_inference_error achieves a new minimum.
             Checkpoints are saved as best_inference_ckpt_XXXX.tar.
--- a/fme/ace/train/train_config.py
+++ b/fme/ace/train/train_config.py
@@ -122,8 +122,7 @@ class InlineInferenceConfig:
         n_ensemble_per_ic: number of initial condition based ensembles
         epochs: positional slice over the training epochs 1 to max_epochs, so
             index 0 selects epoch 1. By default runs inference every training
-            epoch. Epoch 0 is additionally included when
-            evaluate_before_training is true.
+            epoch.
         aggregator: configuration of inline inference aggregator.
         name: name used as wandb log prefix and output subdirectory. If None,
             defaults to "inference" when there is a single inference config
--- a/fme/coupled/train/train_config.py
+++ b/fme/coupled/train/train_config.py
@@ -150,7 +150,10 @@ class InlineInferenceConfig:
         n_coupled_steps: number of coupled forward steps to take
         coupled_steps_in_memory: number of coupled forward steps to take before
             re-reading data from disk
-        epochs: epochs on which to run inference. By default runs inference every epoch.
+        epochs: positional slice over the training epochs 1 to max_epochs, so
+            index 0 selects epoch 1. By default runs inference every training
+            epoch. Epoch 0 is additionally included when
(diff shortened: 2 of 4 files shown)


fme/ace/train/train_config.py
```diff
@@ -120,7 +120,10 @@ class InlineInferenceConfig:
         forward_steps_in_memory: number of forward steps to take before
             re-reading data from disk
         n_ensemble_per_ic: number of initial condition based ensembles
-        epochs: epochs on which to run inference. By default runs inference every epoch.
+        epochs: positional slice over the training epochs 1 to max_epochs, so
+            index 0 selects epoch 1. By default runs inference every training
+            epoch. Epoch 0 is additionally included when
+            evaluate_before_training is true.
```
jpdunc23:
```suggestion
        epochs: positional slice over the training epochs 1 to max_epochs, so
            index 0 selects epoch 1. By default runs inference every training
            epoch.
```

fme/coupled/train/train_config.py
```diff
@@ -150,7 +150,10 @@ class InlineInferenceConfig:
         n_coupled_steps: number of coupled forward steps to take
         coupled_steps_in_memory: number of coupled forward steps to take before
             re-reading data from disk
-        epochs: epochs on which to run inference. By default runs inference every epoch.
+        epochs: positional slice over the training epochs 1 to max_epochs, so
+            index 0 selects epoch 1. By default runs inference every training
+            epoch. Epoch 0 is additionally included when
+            evaluate_before_training is true.
```
jpdunc23:
```suggestion
        epochs: positional slice over the training epochs 1 to max_epochs, so
            index 0 selects epoch 1. By default runs inference every training
            epoch.
```

fme/coupled/train/train_config.py
```diff
@@ -520,9 +522,12 @@ def get_inference_epoch_sets(self) -> list[set[int]]:
...
             return []
-        start_epoch = 0 if self.evaluate_before_training else 1
-        all_epochs = list(range(start_epoch, self.max_epochs + 1))
-        return [set(all_epochs[entry.epochs.slice]) for entry in inference]
+        training_epochs = list(range(1, self.max_epochs + 1))
+        pre_training_epochs = {0} if self.evaluate_before_training else set()
+        return [
+            set(training_epochs[entry.epochs.slice]) | pre_training_epochs
+            for entry in inference
+        ]
```
jpdunc23, line 530: We can probably de-duplicate this (and the tests) in `fme.ace` and `fme.coupled`, but for now I'd ask the reviewer to let this PR keep the status quo of duplication so we can get this bugfix merged.

In a follow-up PR or sequence we should see if we can de-duplicate a lot of the config boilerplate.

fme/ace/train/train_config.py
```diff
@@ -127,7 +127,9 @@ class InlineInferenceConfig:
         forward_steps_in_memory: number of forward steps to take before
             re-reading data from disk
         n_ensemble_per_ic: number of initial condition based ensembles
-        epochs: epochs on which to run inference. By default runs inference every epoch.
+        epochs: positional slice over the training epochs 1 to max_epochs, so
+            index 0 selects epoch 1. By default runs inference every training
+            epoch.
```
elynnwu, line 132: Consider restoring "Epoch 0 is additionally included when evaluate_before_training is true." that was deleted in an earlier commit.

elynnwu (approved):

fme/ace/train/train_config.py
jpdunc23: I removed it because I was afraid it might mislead readers into thinking that epoch 0 is included in the slicing, which is the bug this PR fixes. 

I updated the `evaluate_before_training` docstring to hopefully be a bit more clear that it runs for all inline validation and inference configs.

PR #1410, opened 2026-08-05
Upload a Beaker dataset for time-coarsened stats

brianhenn:
`upload_stats.py` uploaded only the native-resolution stats to Beaker, so the time-coarsened (e.g. daily) stats that `get_stats.py` and `combine_stats.py` already compute could not be published. It now uploads one Beaker dataset per configured stats directory, driven by a new `time_coarsen.beaker_dataset`.

`upload_stats.py` imports `StatsConfig` and `TimeCoarsenConfig` from `get_stats.py` rather than declaring its own, which makes `stats.beaker_dataset` optional: omit it to upload only the coarsened stats. `Config.__post_init__` rejects a config naming no Beaker dataset at all, so that cannot silently become a no-op. Every checked-in config sets it, so none change behavior.

Changes:
- `scripts/data_process/upload_stats.py`: builds and uploads one `UploadSpec` per Beaker dataset, replacing the single hard-coded upload.
- `scripts/data_process/get_stats.py`: `TimeCoarsenConfig` gains `beaker_dataset`, and `factor` for the dataset description.
- `scripts/data_process/compute_dataset_argo_workflow.yaml`: provisions `get_stats.py` into the `upload-beaker-stats` step, as other steps already do.
- `scripts/data_process/test_config.py`: unit tests over the native/coarsened combinations, including the rejection case.

- [x] Tests added

### DIFF
--- a/scripts/data_process/get_stats.py
+++ b/scripts/data_process/get_stats.py
@@ -89,10 +89,15 @@ class TimeCoarsenConfig:
     Attributes:
         data_output_directory: Directory to save the coarsened datasets as zarr stores.
         stats_output_directory: Directory to save the stats of the coarsened datasets.
+        factor: Factor by which the time dimension is coarsened.
+        beaker_dataset: Name of the Beaker dataset to create from the coarsened stats.
+            If None, the coarsened stats are not uploaded to Beaker.
     """
 
     data_output_directory: str
     stats_output_directory: str
+    factor: int
+    beaker_dataset: str | None = None
 
 
 @dataclasses.dataclass
--- a/scripts/data_process/compute_dataset_argo_workflow.yaml
+++ b/scripts/data_process/compute_dataset_argo_workflow.yaml
@@ -100,6 +100,8 @@ spec:
             parameters:
             - name: python_script
               value: "{{workflow.parameters.upload_stats_script}}"
+            - name: stats_script
+              value: "{{workflow.parameters.get_stats_script}}"
             - name: config
               value: "{{workflow.parameters.config}}"
     - name: compute-fme-dataset-individual
@@ -317,6 +319,7 @@ spec:
       inputs:
         parameters:
           - name: python_script
+          - name: stats_script
           - name: config
       container:
         image: us-central1-docker.pkg.dev/vcm-ml/full-model/atmosphere-processing:v2025.12.0
@@ -332,6 +335,10 @@ spec:
           - |
             cat << EOF > script.py
             {{inputs.parameters.python_script}}
+            EOF
+
+            cat << EOF > get_stats.py
+            {{inputs.parameters.stats_script}}
             EOF
 
             cat << EOF > config.yaml
--- a/scripts/data_process/test_config.py
+++ b/scripts/data_process/test_config.py
@@ -90,17 +90,20 @@ def test_valid_create_coupled_ic_config(filename):
     )
 
 
+NATIVE_DATA_DIRECTORY = "gs://bucket/native-6hourly"
 NATIVE_STATS_DIRECTORY = "gs://bucket/native-stats"
 COARSENED_STATS_DIRECTORY = "gs://bucket/native-daily-stats"
 COARSENED_DATA_DIRECTORY = "gs://bucket/native-daily"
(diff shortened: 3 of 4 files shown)


brianhenn:
Pre-review agent: three judgment calls in this PR that I left as-authored, flagged so the reviewer can go straight to them rather than re-deriving the context.

**1. `factor` is added to `get_stats.TimeCoarsenConfig`, but `get_stats.py` never reads it.** It exists only so `upload_stats.py` can put the coarsening factor in the Beaker dataset description. The obvious alternative is for `upload_stats.py` to import `TimeCoarsenConfig` from `time_coarsen.py`, which already declares `factor` and is the fullest copy of that schema. The cost of that alternative is in the Argo step: `time_coarsen.py` itself does `from get_stats import StatsConfig`, so `upload-beaker-stats` would have to be provisioned with two scripts instead of one, and would pull `xarray` in at import. The chosen split looks like the cheaper option, but it does leave an unused field in a config class.

**2. Three partial copies of the `time_coarsen:` schema now exist** — `get_stats.py`, `time_coarsen.py`, `combine_stats.py` — with `factor` in two of them. All three parse the same YAML block via non-strict dacite, so they can't disagree on a value, but they can drift on which fields are required. Consolidating them is out of scope here; worth deciding whether it's worth a follow-up.

**3. `_upload_specs` / `_describe` are free functions that read `Config` and its children's fields** (`config.stats.exclude_runs`, `config.time_coarsen.factor`), which is the config-as-parameter-bag shape AGENTS.md steers away from. The PR does move in the right direction — `UploadSpec` is a leaf dataclass, so `_upload` no longer touches the config at all. The tidier version (an `upload_spec()` method on `StatsConfig` / `TimeCoarsenConfig`) is blocked by module layout: those classes live in `get_stats.py`, which would then need to import `UploadSpec` from `upload_stats.py`, and `upload_stats.py` already imports from `get_stats.py`. So this reads as a local optimum rather than an oversight.

Checks that came back clean: all 15 configs with a `time_coarsen:` block already set `factor`, so making it required breaks none of them (and existing `test_get_stats_valid` parametrizes `UploadStatsConfig` over every config, so it's guarded). `combine_stats.py` already writes the coarsened `combined/` directory the new upload reads. `get_stats_script` was already a declared workflow parameter used by two other templates. The refactored description string is byte-identical to the previous one. Moving the already-exists check into per-spec `_upload` correctly continues to the next dataset instead of exiting.

scripts/data_process/get_stats.py
```diff
@@ -89,10 +89,15 @@ class TimeCoarsenConfig:
...
         stats_output_directory: Directory to save the stats of the coarsened datasets.
+        factor: Factor by which the time dimension is coarsened.
+        beaker_dataset: Name of the Beaker dataset to create from the coarsened stats.
+            If None, the coarsened stats are not uploaded to Beaker.
     """
 
     data_output_directory: str
     stats_output_directory: str
+    factor: int
+    beaker_dataset: str | None = None
```
Arcomano1234, line 100: Question: Does this now lead to a silently not upload to beaker now and exits successfully. Claude thinks this was not the previous behavior where it would fail loudly and exit the argo workflow.

Arcomano1234 (approved):
Approved although I think the behavior changed for the beaker_dataset which can lead to unexpected silent "successful" argo job with out uploading the stats. Lets double check this and then it should be good to go. I think the things missing from this PR will be in the follow on PR I'm working on

scripts/data_process/get_stats.py
brianhenn: (claude): You're right, and thanks for catching it. Measured on a config with no `stats.beaker_dataset`: `origin/main` exits 1 with `MissingValueError`, this branch exited 0 — so the argo step went green without uploading. The warning I'd added doesn't change the exit status.

The field had to become optional to express a coarsen-only run (upload just the daily stats when the native ones are already in Beaker), so rather than revert it I've made `Config.__post_init__` reject a config that names *no* Beaker dataset at all. Omitting one of the two is now deliberate; omitting both — or mistyping the key, which non-strict dacite would otherwise ignore — fails loudly again.

PR #1411, opened 2026-08-05
Temporal Downscaling Stage 1

jhhuangchloe:
Adds an endpoint-conditioned video diffusion model ("video PMD") for temporal
interpolation: given two observed daily snapshots (e.g. 0h/24h), it diffuses
only the residual over a linear-interpolation baseline to infill 3-hourly
interior frames, with endpoints pinned exactly. Includes per-channel temporal
noise kernels (Brownian-bridge/OU/RBF), subset-frame training so the model can
answer variable query sets, and an optional marginal-consistency loss (L_marg)
that penalizes disagreement between full-grid and subset predictions sharing
the same noise. Two interchangeable backbones: a compact periodic VideoUNet,
and a SongUNetv2 wrapper with temporal attention injected via forward hooks.

Changes:
- `fme.downscaling.noise`: Brownian-bridge, OU, and RBF temporal noise-mixing
  kernels, optionally one per channel.
- `fme.downscaling.modules.video_modules.VideoUNet`/`VideoEDMPrecond`: compact
  5-D (B, C, T, H, W) backbone with calendar embedding and temporal attention.
- `fme.downscaling.modules.video_song_unet.VideoSongUNet`: SongUNetv2 wrapped
  for video via forward-hook temporal attention; `physicsnemo_unets_v2.Conv2d`
  gains a `periodic` flag for longitude-circular padding on global fields (and
  a guard against the untested up/down resample-filter case, currently inert).
- `fme.downscaling.video_models.VideoDiffusionModel`/`VideoDiffusionModelConfig`:
  the model itself, including subset-frame generation and the L_marg loss.
- `fme.downscaling.data.config.PairedDataLoaderConfig.build_video`: paired
  fine/coarse video-clip data loading.
- `fme.downscaling.aggregators.main.LossVsNoiseAggregator`: accepts a
  per-channel sigma; the shared-sigma "all_channels" total is now a per-sample
  sum across channels rather than a mean (see class docstring) -- historical
  wandb `loss_vs_noise/all_channels` curves logged before this change are on a
  different scale.
- `fme.downscaling.video_train`: training entry point (checkpointing, EMA,
  wandb visualization panels, optional held-out test split).
- `fme.downscaling.video_inference`: test-set ensemble inference from a
  trained checkpoint, writing a zarr store matching the input format.
- `fme.core.logging_utils.LoggingConfig`: optional `name` field for the wandb
  run display name.
- `configs/experiments/2026-*-video-pmd-*`: dated run configs from the sweep
  that developed this model.
- `configs/baselines/downscaling-temporal/train.yaml`: canonical reference
  config (5-channel, per-channel OU kernel, subset-frame training).

- [x] Tests added

### DIFF
--- a/configs/baselines/downscaling-temporal/train.yaml
+++ b/configs/baselines/downscaling-temporal/train.yaml
@@ -1,0 +1,192 @@
+# Canonical training config for the endpoint-conditioned video-PMD temporal
+# interpolation model (global 1deg, 24h -> 3h). Per-channel Ornstein-Uhlenbeck
+# noise kernels (temporal_noise_correlation: per_channel), plus subset-frame
+# training so the model can answer variable query sets. See
+# fme/downscaling/video_train.py and configs/experiments/2026-*-video-pmd-*
+# for the sweep this was selected from.
+experiment_dir: /results
+max_epochs: 200
+validate_using_ema: true
+validate_interval: 10
+test_interval: 20
+max_val_batches: 20
+max_test_batches: 50
+generate_n_samples: 1
+log_loss_vs_noise: true
+logging:
+  log_to_screen: true
+  log_to_wandb: true
+  log_to_file: true
+  project: ace-temporal-interp
+  name: global-1degree-24to3-5ch-per-channel-kernel-subset-cons0-v1
+  entity: ai2cm
+optimization:
+  lr: 0.0001
+ema:
+  decay: 0.9999
+train_data:
+  n_timesteps: 9
+  fine:
+    - data_path: /climate-default/2026-06-25-temporal-diffusion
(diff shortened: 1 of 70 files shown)


AnnaKwa (changes requested):
Since the temporal downscaling is a very large feature, could you split this into multiple smaller PRs that each implement a component of the workflow, starting from the bottom and building up the dependency chain? e.g. the new video data components and related tests, new modules, training, inference, and experiment configs would all be separate PRs.

jhhuangchloe:
Splitting into a stacked series: #1412 (data) → #1413 (modules) → #1415 (training) → #1416 (inference) → #1417 (configs). Closing this in favor of that stack.

PR #1412, opened 2026-08-06
Add video/paired-clip data loading

jhhuangchloe:
PairedDataLoaderConfig.build_video builds fine/coarse loaders of video clips
(n_timesteps consecutive frames with an explicit leading time axis), reusing
the existing paired-dataset machinery, as the data-layer foundation for an
upcoming endpoint-conditioned video diffusion model for temporal
interpolation (video PMD). This PR is data-loading only; no model code.

Changes:
- `fme.downscaling.data.config.PairedDataLoaderConfig.build_video`: builds a
  paired fine/coarse DataLoader of video clips of `n_timesteps` consecutive
  frames, spaced `clip_start_stride` apart.
- `fme.downscaling.data.datasets`: new `VideoBatchItem`, `VideoBatchData`,
  `PairedVideoBatchData`, `PairedVideoGriddedData`,
  `VideoBatchItemDatasetAdapter`, `VideoFineCoarsePairedDataset` types.
- `fme.downscaling.data.time_encoding.compute_calendar_features`: per-frame
  UTC (day_of_year, second_of_day) for a clip, consumed by the video model in
  a later PR.

- [x] Tests added


fme/downscaling/data/datasets.py
```diff
@@ -608,6 +609,282 @@ def generate_from_patches(
             yield PairedBatchData(fine=fine_batch, coarse=coarse_batch)
 
 
+# Video (temporal) data containers: clips of ``T`` frames with an explicit
```
AnnaKwa: Since this module is already quite long, and the video dataset objects will probably be called in different instances from the existing data objects, could you split the new code here into a `video_datasets.py` module? `compute_calendar_features` could also go in there.

fme/downscaling/data/config.py
```diff
@@ -448,6 +452,8 @@ class PairedDataLoaderConfig:
     topography: str | None = None
     sample_with_replacement: int | None = None
     drop_last: bool = False
+    n_timesteps: int = 1
```
AnnaKwa: Can you add docstrings for these new args?

fme/downscaling/data/config.py
```diff
@@ -604,6 +623,120 @@ def build(
             coarse_extent_latlon_coords=example.coarse.latlon_coordinates,
         )
 
+    def build_video(
```
AnnaKwa, line 685: Instead of adding to `PairedDataLoaderConfig`, can you instead make a `PairedVideoLoaderConfig` class that inherits from this but has different build method, the  additional args, and the clip_start_stride property? Then when later training config classes type hint the data loader arg, having the `PairedVideoLoaderConfig` in the type hint will make it more specific about what the data config should be.

AnnaKwa: Note this rule we have for inheritance in the repo:

> Methods defined on a class with at least one in-repo subclass must be `@abstractmethod` or `@final`. Exception: `__init__` on `nn.Module` subclasses.

i.e. build method on subclass would get a `@final` decorator

AnnaKwa (changes requested):
Generally looks good! Just some requests for reorganization to separate out the classes responsible for the video work, since they will be used separately from the existing spatial downscaling code.

jhhuangchloe:
Thanks for the review, all items addressed in 191be9aa..3018bebb.

1. **Split video code into `video_datasets.py`** — `VideoBatchItem`, `VideoBatchItemDatasetAdapter`, `VideoBatchData`, `VideoFineCoarsePairedDataset`, `PairedVideoBatchItem`, `PairedVideoBatchData`, and `PairedVideoGriddedData` moved out of `datasets.py` into the new `fme/downscaling/data/video_datasets.py`. `compute_calendar_features` moved there too (from `time_encoding.py`, which is now deleted since it was a single-function module). Renamed `test_video.py` → `test_video_datasets.py` to match, consistent with this directory's existing 1:1 test/module naming.

2. **Docstrings for `n_timesteps`/`time_stride`** — added on the new `PairedVideoLoaderConfig` (see #3), covering both the fixed args and the derived `clip_start_stride` property.

3. **`PairedVideoLoaderConfig` subclass** — `n_timesteps`, `time_stride`, `clip_start_stride`, and `build_video` all moved off `PairedDataLoaderConfig` into a new `PairedVideoLoaderConfig(PairedDataLoaderConfig)`. Confirmed before moving that none of this was load-bearing on the base: `.build()` never reads `self.n_timesteps` (it takes timestep count from `requirements` instead), and no non-video caller or config ever set `n_timesteps`/`time_stride` on a plain `PairedDataLoaderConfig`.

4. **`@final`/`@abstractmethod`** — since `PairedDataLoaderConfig` now has an in-repo subclass, marked its remaining methods (`build`, `_get_sampler`, `_mp_context`, `_repeat_if_requested`, `_first_data_config`, `coarse_full_config`) `@final`; `build_video` is `@final` on the subclass. Left `__post_init__` unmarked — it's the dataclass-native override-and-chain hook (`PairedVideoLoaderConfig.__post_init__` calls `super().__post_init__()`), matching the one existing precedent for this pattern in the repo (`inference/work_items.py`'s `SliceWorkItem`/`LoadedSliceWorkItem`).

Verification: `fme/downscaling/` — pytest 453 passed; pre-commit (ruff, ruff-format, mypy) clean.

Downstream branches in the stack (`video-pmd-modules`, `-training`, `-inference`, `-configs`) imported these symbols and type-hinted against `PairedDataLoaderConfig`, so I've rebased all of them on top of this and fixed the handful of spots that needed `PairedVideoLoaderConfig`/the new import paths (#1413, #1415, #1416, #1417 are all updated and still mergeable).

fme/downscaling/data/config.py
```diff
@@ -485,11 +495,13 @@ def _mp_context(self):
         return mp_context
 
     @property
+    @final
     def coarse_full_config(
         self,
     ) -> Sequence[XarrayDataConfig | MergeNoConcatDatasetConfig]:
         return _full_configs(self.coarse)
 
+    @final
```
AnnaKwa, line 504: Oops, I misspoke about putting the decorator on the subclass, these edits to put `@final` on the parent are correct.

AnnaKwa (approved):
Thanks for the updates, LGTM!

PR #1413, opened 2026-08-06
Add video diffusion model backbones and VideoDiffusionModel

jhhuangchloe:
Adds the model layer for video PMD: an endpoint-conditioned diffusion model
that, given two observed daily snapshots (e.g. 0h/24h), diffuses only the
residual over a linear-interpolation baseline to infill 3-hourly interior
frames, with endpoints pinned exactly. Builds on the data loading from #1412.

Changes:
- `fme.downscaling.noise`: Brownian-bridge, OU, and RBF temporal noise-mixing
  kernels that keep noise pinned at clip endpoints, optionally one per
  channel (`per_channel_noise_kernel`/`per_channel_kernel_length_scale`).
- `fme.downscaling.modules.video_modules.VideoUNet`/`VideoEDMPrecond`: a
  compact 5-D (B, C, T, H, W) backbone, periodic in longitude, with
  cBottle-style calendar embedding and temporal attention, wrapped in EDM
  preconditioning (Karras 2022).
- `fme.downscaling.modules.video_song_unet.VideoSongUNet`: wraps the existing
  SongUNetv2 image backbone for video via forward-hook temporal attention on
  its attention-resolution blocks.
- `fme.downscaling.modules.physicsnemo_unets_v2.layers.Conv2d`: new
  `periodic` flag for longitude-circular padding on global lat/lon fields
  (main conv branch only; a guard prevents silently combining it with a
  resample filter wider than the current default, which the up/down branch
  doesn't yet support circularly).
- `fme.downscaling.video_models.VideoDiffusionModel`/`VideoDiffusionModelConfig`:
  the model itself, including subset-frame generation (train on a random
  subset of interior frames so the model can answer variable query sets) and
  an optional marginal-consistency loss (L_marg) tying full-grid and
  subset-native predictions that share the same noise.

- [x] Tests added

### DIFF
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -47,6 +47,8 @@ class ModelOutputs:
     channel_losses: TensorDict = dataclasses.field(default_factory=dict)
     sigma: torch.Tensor | None = None
     per_sample_channel_loss: TensorDict = dataclasses.field(default_factory=dict)
+    # Optional marginal-consistency term (video PMD), for logging only.
+    marginal_consistency_loss: torch.Tensor | None = None
 
 
 def _rename_normalizer(
--- a/fme/downscaling/models.py
+++ b/fme/downscaling/models.py
@@ -47,8 +47,6 @@ class ModelOutputs:
     channel_losses: TensorDict = dataclasses.field(default_factory=dict)
     sigma: torch.Tensor | None = None
     per_sample_channel_loss: TensorDict = dataclasses.field(default_factory=dict)
-    # Optional marginal-consistency term (video PMD), for logging only.
-    marginal_consistency_loss: torch.Tensor | None = None
 
 
 def _rename_normalizer(
--- a/fme/downscaling/modules/video_modules.py
+++ b/fme/downscaling/modules/video_modules.py
@@ -5,6 +5,7 @@ and temporal attention, wrapped in EDM preconditioning (Karras 2022).
 """
 
 import math
+from typing import Literal
 
 import torch
 import torch.nn as nn
@@ -21,7 +22,9 @@ class NoiseEmbedding(nn.Module):
     mapped through an MLP -- matches HiRO/SongUNet noise conditioning.
     """
 
-    def __init__(self, dim: int, embedding_type: str = "positional"):
+    def __init__(
+        self, dim: int, embedding_type: Literal["positional", "fourier"] = "positional"
+    ):
         super().__init__()
         if dim % 2 != 0:
             raise ValueError("NoiseEmbedding dim must be even.")
@@ -315,7 +318,7 @@ class VideoUNet(nn.Module):
         attention_levels: tuple[int, ...] = (1, 2),
         temporal_attention_levels: tuple[int, ...] | None = None,
         num_freqs: int = 4,
-        noise_embedding_type: str = "positional",
+        noise_embedding_type: Literal["positional", "fourier"] = "positional",
     ):
         super().__init__()
         self.levels = len(channel_mult)
--- a/fme/downscaling/modules/video_modules.py
+++ b/fme/downscaling/modules/video_modules.py
@@ -1,0 +1,477 @@
+"""Network building blocks for video (temporal) diffusion downscaling.
+
+Compact 5-D ``(B, C, T, H, W)`` backbone with cBottle-style calendar embedding
+and temporal attention, wrapped in EDM preconditioning (Karras 2022).
+"""
+
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from fme.downscaling.modules.physicsnemo_unets_v2.layers import (
+    FourierEmbedding,
+    PositionalEmbedding,
+)
+
+
+class NoiseEmbedding(nn.Module):
+    """Positional (DDPM++) or Fourier (NCSN++) embedding of the log-noise level,
+    mapped through an MLP -- matches HiRO/SongUNet noise conditioning.
+    """
+
+    def __init__(self, dim: int, embedding_type: str = "positional"):
+        super().__init__()
+        if dim % 2 != 0:
+            raise ValueError("NoiseEmbedding dim must be even.")
+        if embedding_type == "fourier":
+            self.embed: nn.Module = FourierEmbedding(dim)
+        elif embedding_type == "positional":
+            self.embed = PositionalEmbedding(dim)
+        else:
+            raise ValueError(f"unknown noise embedding_type {embedding_type!r}")
+        self.mlp = nn.Sequential(nn.Linear(dim, dim), nn.SiLU(), nn.Linear(dim, dim))
+
+    def forward(self, c_noise: torch.Tensor) -> torch.Tensor:
+        return self.mlp(self.embed(c_noise.float()))
+
+
+class FrequencyEmbedding(nn.Module):
+    """Periodic sinusoidal encoding on the circle for x in [0, 1)."""
+
+    def __init__(self, num_freqs: int):
+        super().__init__()
+        self.num_freqs = num_freqs
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        k = torch.arange(1, self.num_freqs + 1, device=x.device, dtype=x.dtype)
+        ang = 2 * math.pi * x.unsqueeze(-1) * k
+        return torch.cat([ang.cos(), ang.sin()], dim=-1)
+
+
+class CalendarEmbedding(nn.Module):
+    """cBottle-style absolute-time embedding (diurnal + seasonal).
+
+    ``second_of_day`` is shifted by longitude into local solar time so the
+    diurnal phase is correct per pixel.
+    """
+
+    def __init__(self, num_freqs: int = 4):
+        super().__init__()
+        self.embed = FrequencyEmbedding(num_freqs)
+        self.out_channels = 4 * num_freqs
+
+    def forward(
+        self,
+        day_of_year: torch.Tensor,
+        second_of_day: torch.Tensor,
+        lon: torch.Tensor,
+        height: int,
+    ) -> torch.Tensor:
+        lon = lon.to(second_of_day.device, second_of_day.dtype)
+        local = second_of_day[:, :, None] + lon[None, None, :] * 86400.0 / 360.0
+        local = local % 86400.0
+        diurnal = self.embed(local / 86400.0)
+        diurnal = diurnal.permute(0, 3, 1, 2)
+        diurnal = diurnal.unsqueeze(3).expand(-1, -1, -1, height, -1)
+
+        seasonal = self.embed((day_of_year / 365.25) % 1.0)
+        seasonal = seasonal.permute(0, 2, 1)[:, :, :, None, None]
+        seasonal = seasonal.expand(-1, -1, -1, height, diurnal.shape[-1])
+        return torch.cat([diurnal, seasonal], dim=1)
+
+
+def _groups(channels: int) -> int:
+    g = min(32, channels)
+    while channels % g != 0:
+        g -= 1
+    return g
+
+
+class PeriodicConv3d(nn.Module):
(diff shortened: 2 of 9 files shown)


fme/downscaling/video_models.py
```diff
@@ -0,0 +1,972 @@
...
+    x0 = field[..., 0:1, :, :]
+    xT = field[..., n_times - 1 : n_times, :, :]
+    return (1 - w) * x0 + w * xT
+
+
+@dataclasses.dataclass
+class VideoDiffusionModelConfig:
+    """Configuration for the temporal-interpolation video diffusion model."""
+
+    out_names: list[str]
```
AnnaKwa, line 126: Can you add docstrings for these classes? That helps clarify args like sigma_min (is for generation, but not that clear from the name). ex. see `DiffusionModelConfig`

fme/downscaling/models.py
```diff
@@ -47,6 +47,8 @@ class ModelOutputs:
     channel_losses: TensorDict = dataclasses.field(default_factory=dict)
     sigma: torch.Tensor | None = None
     per_sample_channel_loss: TensorDict = dataclasses.field(default_factory=dict)
+    # Optional marginal-consistency term (video PMD), for logging only.
```
AnnaKwa: To keep all the video-specific code together I suggest reverting this original class and making a `VideoModelOutputs` in the video_models module, with the video-specific loss field.

fme/downscaling/noise.py
```diff
@@ -56,6 +57,142 @@ def sample(self, batch_size: int, device: torch.device) -> torch.Tensor:
...
+
+    Endpoints land exactly on 0 and 1; interior frames on ``k / (T - 1)``.
+    """
+    return torch.linspace(0.0, 1.0, n_timesteps, dtype=torch.float64)
+
+
+def brownian_bridge_mixing_matrix(tau: torch.Tensor) -> torch.Tensor:
+    """Temporal mixing matrix for endpoint-pinned, time-correlated noise.
+
+    Given the normalized frame times ``tau`` (1-D, length ``T``, with
```
AnnaKwa: Is there a reference that can be linked via a url instead of putting the full description here?

fme/downscaling/noise.py
```diff
@@ -56,6 +57,142 @@ def sample(self, batch_size: int, device: torch.device) -> torch.Tensor:
...
+    )
+    mixing[1 : 1 + n_interior, 1 : 1 + n_interior] = chol
+    return mixing.to(torch.float32)
+
+
+def ou_mixing_matrix(tau: torch.Tensor, length_scale: float) -> torch.Tensor:
+    """Temporal mixing matrix for endpoint-pinned, Ornstein-Uhlenbeck-
+    correlated noise: ``corr(s, t) = exp(-|s - t| / length_scale)``.
+
+    Same endpoint-pinned-interior-Cholesky structure as
```
AnnaKwa: In general for these function definitions, just describe them briefly and remove the longer details. e.g. for this one I'd just leave it at the first two lines.

fme/downscaling/noise.py
```diff
@@ -56,6 +57,142 @@ def sample(self, batch_size: int, device: torch.device) -> torch.Tensor:
...
+            "Brownian-bridge noise needs at least 3 frames (2 endpoints + 1 "
+            f"interior); got tau with shape {tuple(tau.shape)}."
+        )
+    n_timesteps = tau.shape[0]
+    n_interior = n_timesteps - 2
+    tau_i = tau[1:-1].to(torch.float64)
+    s = tau_i.reshape(-1, 1)
+    t = tau_i.reshape(1, -1)
+    cov = torch.minimum(s, t) - s * t  # interior Brownian-bridge covariance
+    std = cov.diagonal().sqrt()
```
AnnaKwa: Can you just call `_interior_cholesky_mixing_matrix` in here with the appropriate corr_fn?

AnnaKwa (changes requested):
Claude likes to write very long inline comments describing the code in detail. In general we want to avoid leaving them in the PRs since they make the code very long, and future-Claude could to read those and can assume they are the truth about how the code operates when maybe it is not correct, or the code itself has changed in the meantime. For this PR and others, could you remove those and just leave the inline comments you feel are necessary?

jhhuangchloe:
Thanks for the review, all items addressed in abae57dc.

1. **Verbose inline comments** — trimmed throughout. Converted the `VideoDiffusionModelConfig` field comments into the docstring below (one line per field, matching `DiffusionModelConfig`'s style — no separate inline comments duplicating it). Also found and removed comments that were flat-out duplicated elsewhere: the `subset_min_interior + 1` constraint was explained in both `__post_init__` and `train_on_batch`; `frame_index`'s meaning was explained in both `TemporalAttention.forward` and again at its `train_on_batch` call site. Cut to one place each. Kept the handful that explain a real non-obvious "why" (e.g. the DDP buffer-versioning workaround in `FIRBlur.forward`).

2. **Docstrings for `sigma_min` etc.** — `VideoDiffusionModelConfig` now has a full `Parameters:` docstring (`sigma_min`/`sigma_max` explicitly called out as "for generation"), and `VideoDiffusionModel` has a short class docstring.

3. **`VideoModelOutputs`** — `ModelOutputs` is back to its pre-video shape; `marginal_consistency_loss` now lives on a new `VideoModelOutputs(ModelOutputs)` in `video_models.py`, and `train_on_batch` returns that instead.

4. **`noise.py` URL reference** — `brownian_bridge_mixing_matrix`'s inline covariance derivation is replaced with a link to the standard definition (Wikipedia's Brownian bridge page); kept the subset-marginal property paragraph since it's referenced from `video_models.py`.

5. **`noise.py` trim `ou_mixing_matrix`/`rbf_mixing_matrix`** — both cut to description + `Args:`, dropping the speculative empirical claims that referenced a report file not present in the repo (kept the RBF near-singularity caution, since that's a real numerical gotcha, not narrative).

6. **`brownian_bridge_mixing_matrix` calls `_interior_cholesky_mixing_matrix`** — refactored to build the Brownian-bridge correlation analytically (`(min(s,t) - s*t) / sqrt(var(s)*var(t))`) and delegate to the shared scaffold, matching `ou_mixing_matrix`/`rbf_mixing_matrix`. Verified numerically identical via the existing tests (`test_noise.py`'s own reference implementation is literally the old inline code, so this is a direct regression check).

Verification: `fme/downscaling/` — pytest 488 passed; pre-commit (ruff, ruff-format, mypy) clean.

fme/downscaling/noise.py
AnnaKwa: Full description can be removed now that the reference link is provided.

fme/downscaling/video_models.py
```diff
@@ -0,0 +1,982 @@
...
+    n_timesteps: int
+    normalization: NormalizationConfig | None = None
+    sigma_min: float = 0.002
+    sigma_max: float = 80.0
+    churn: float = 0.0
+    num_diffusion_generation_steps: int = 18
+    model_channels: int = 64
+    n_heads: int = 4
+    num_freqs: int = 4
+    noise_embedding_type: str = "positional"
```
AnnaKwa:
```suggestion
    noise_embedding_type: Literal["positional", "fourier"]
```
etc. for other args that must be in some subset

fme/downscaling/modules/video_modules.py
```diff
@@ -0,0 +1,477 @@
...
+    PositionalEmbedding,
+)
+
+
+class NoiseEmbedding(nn.Module):
+    """Positional (DDPM++) or Fourier (NCSN++) embedding of the log-noise level,
+    mapped through an MLP -- matches HiRO/SongUNet noise conditioning.
+    """
+
+    def __init__(self, dim: int, embedding_type: str = "positional"):
```
AnnaKwa:
```suggestion
    def __init__(self, dim: int, embedding_type: Literal["positional", "fourier"]):
```
etc. for other similar args in this module

fme/downscaling/modules/video_song_unet.py
```diff
@@ -0,0 +1,87 @@
...
+        self,
+        in_channels: int,
+        out_channels: int,
+        img_resolution: list[int],
+        seq_length: int,
+        model_channels: int = 128,
+        channel_mult: tuple[int, ...] = (1, 2, 2, 2),
+        num_blocks: int = 2,
+        n_heads: int = 8,
+        attn_resolutions: tuple[int, ...] = (22,),
```
AnnaKwa, line 29: Do you want the video UNet to allow no attention resolutions?

fme/downscaling/modules/video_song_unet.py
```diff
@@ -0,0 +1,87 @@
...
+        num_freqs: int = 4,
+        periodic: bool = True,
+    ):
+        super().__init__()
+        self.seq_length = seq_length
+        self.calendar = CalendarEmbedding(num_freqs)
+        self.unet = SongUNetv2(
```
AnnaKwa, line 49: There are additional init args for `SongUNetv2` that aren't passed here and thus aren't available to be configured in the video model, could you make those configurable in the `VideoDiffusionModelConfig` and pass them through to build? 

in particular the `use_apex_gn` can speed up training on the B200s but it defaults to False and can't be set here.

AnnaKwa (changes requested):
Thanks for the updates, mostly LGTM but I noticed another important thing in the module building that needs to be updated too (should be able to configure all possible SongUNetv2 args)

jhhuangchloe:
Thanks for the follow-up, all items addressed in 5bb785e6.

1. **Brownian-bridge full description** — removed now that the reference link is there; left only the subset-marginal-property paragraph (still referenced from `video_models.py`).

2. **`Literal[...]` types** — `noise_embedding_type`, `backbone`, `temporal_noise_correlation` (`video_models.py`), `embedding_type` (`video_modules.py`'s `NoiseEmbedding`/`VideoUNet`), and `per_channel_noise_kernel`'s dict values are all `Literal[...]` now instead of `str`. Kept the existing runtime checks in `__post_init__` alongside the type hints — those still matter since dacite/yaml-loaded configs never go through mypy. This did catch one thing: `test_invalid_temporal_noise_correlation_rejected` intentionally passes an invalid value to exercise that runtime guard, so it needed a `# type: ignore[arg-type]` (justified inline).

3. **"Do you want the video UNet to allow no attention resolutions?"** — Yes: the "simple" backbone already lets us do a no-spatial-attention ablation via `attention_levels=[]`, so the songunet backbone should support the same via `attn_resolutions=[]` for consistency between the two. That exposed a bug: `attn_resolutions=tuple(self.attn_resolutions or [default_attn])` treated an explicit empty list the same as unset, silently falling back to the default resolution instead of actually disabling attention. Fixed to check `is None` instead. Added `test_attn_resolutions_empty_list_disables_non_bottleneck_attention`, which also surfaces that `SongUNetv2`'s bottleneck block always gets attention regardless of `attn_resolutions` (`bottleneck_attention` defaults to `True` and isn't exposed here) — noted in the test docstring.

4. **`use_apex_gn`** — added to `VideoDiffusionModelConfig` (songunet-only) and threaded through `VideoSongUNet` to `SongUNetv2`, replacing the hardcoded `False`. Since it requires a GPU (`SongUNetv2`'s own docstring: "Need to set this as False on cpu"), added a `__post_init__` check. Scoped to just this arg per our earlier discussion rather than exposing all of `SongUNetv2`'s remaining init args.

Verification: `fme/downscaling/` — 491 pytest passed; pre-commit (ruff, ruff-format, mypy) clean.

AnnaKwa (approved):
Sorry what I meant in the comment about passing the init args was that I was requesting that all possible init args for `SongUNetv2` be configurable (can leave as their same defaults) and passed through, not just the apex_gn one. That was just an example of one that definitely needed to be included, but I think all should be configurable. 

I'm approving since I trust you can make that correction without it needing to be reviewed, but please update before merging.

jhhuangchloe:
Thanks for the follow-up! All items addressed in 5bb785e6..d01a6a3e, including the correction on `SongUNetv2` args.

PR #1415, opened 2026-08-06
Add video PMD trainer

jhhuangchloe:
Adds the training entry point for VideoDiffusionModel from #1413.

Changes:
- `fme.downscaling.video_train`: `VideoTrainerConfig`/`VideoTrainer` --
  checkpointing (EMA-aware save/restore), wandb visualization panels
  (GT/linear-baseline/pred/diff frame grids), and an optional held-out test
  split evaluated every test_interval epochs.
- `fme.downscaling.aggregators.main.LossVsNoiseAggregator`: accepts a
  per-channel sigma (in addition to a shared per-sample sigma), binning each
  (sample, channel) pair at its own noise level. With a shared sigma, the
  "all_channels" total is now a per-sample sum across channels rather than a
  mean (see updated class docstring and test) -- pre-existing wandb
  loss_vs_noise/all_channels curves logged before this change are on a
  different scale.
- `fme.core.logging_utils.LoggingConfig`: optional `name` field for the
  wandb run display name.
- `fme/downscaling/configs/video_train_xshield.yaml`,
  `video_train_xshield_smoke.yaml`: example training configs.

- [x] Tests added


fme/downscaling/configs/video_train_xshield.yaml
```diff
@@ -0,0 +1,111 @@
...
+  name: regional-30-60-1degree-24to3
+  entity: ai2cm
+optimization:
+  lr: 0.0001
+ema:
+  decay: 0.9999
+train_data:
+  n_timesteps: 9
+  fine:
+    - data_path: gs://vcm-ml-scratch/andrep
```
AnnaKwa: Don't want to have tests rely on reading from GCS data- if an integration test is necessary, use a helper function to write test data to a temporary dir. ex. see `fme.downscaling.test_utils.create_test_data_on_disk`

fme/downscaling/configs/video_train_xshield_smoke.yaml
```diff
@@ -0,0 +1,114 @@
+# CPU smoke config (tiny + capped) for video_train_xshield.yaml.
+# Run: FME_FORCE_CPU=1 python -m fme.downscaling.video_train \
+#   fme/downscaling/configs/video_train_xshield_smoke.yaml
+experiment_dir: /tmp/video_smoke
```
AnnaKwa: Run integration tests with a temporary dir context, and overwrite `experiment_dir` with that path, see `test_evaluator_config.yaml` and usages for example

fme/downscaling/aggregators/main.py
```diff
@@ -104,35 +113,71 @@ def record_batch(self, outputs: ModelOutputs) -> None:
...
+            in_range = (log_sigma >= self._log10_sigma_min) & (
+                log_sigma <= self._log10_sigma_max
+            )
+            if torch.any(in_range):
+                # Indices in [0, n_bins-1] for values within the sigma range.
+                bin_indices = torch.bucketize(log_sigma[in_range], self._inner_edges)
+                self._accumulate(
+                    values=channel_loss[in_range], bin_indices=bin_indices, name=name
+                )
+                if channelwise:
```
AnnaKwa: The previous behavior should be the same as if channelwise is True but all channels have the same sigma. Can these aggregator changes be streamlined to use the same path for all cases?

jhhuangchloe:
Thanks for the review, all items are addressed in 15568be9..b3732dd2.

1. **Streamline the shared-sigma/channelwise aggregator paths** — good catch: shared sigma is just channelwise sigma broadcast to every channel, so it doesn't need its own special case. `LossVsNoiseAggregator.record_batch` now goes through a single per-(sample, channel)-point path for both, which also removes the separate second sigma-binning pass that only ran for shared sigma. One semantic consequence worth flagging: `all_channels` is now a mean across `(sample, channel)` pairs in both cases, rather than a sum for shared sigma and independent points for channelwise — updated the class docstring and the two tests that had locked in the old two-path behavior (`test_loss_vs_noise_aggregator_get_wandb`'s total-count, and renamed/rewrote `test_loss_vs_noise_aggregator_all_channels_totals_sum_across_channels` to `test_loss_vs_noise_aggregator_shared_sigma_matches_channelwise_path`).

2. **`video_train_xshield.yaml` / `video_train_xshield_smoke.yaml` reading from GCS** — neither file was actually wired to a pytest test despite living in the `fme/downscaling/configs/` fixture directory. `video_train_xshield.yaml` (the full, 20-epoch, personal-scratch-bucket config) was an early prototype superseded by the canonical config and the sweep configs already under `configs/`, so I deleted it. `video_train_xshield_smoke.yaml` is now a genuine fixture: `data_path`/`experiment_dir` are `/will/be/overwritten` placeholders swapped for local synthetic data at test time, following `test_evaluator_config.yaml`'s pattern. Dropped the 4th output channel (`PRATEsfc`, with its own log-uniform noise distribution + log-transform) to match the 3 synthetic channels `data_paths_helper` provides — that combination is already covered directly at the model level by `test_video_models.py`'s per-channel-noise tests. Added `test_video_train_xshield_smoke_config_runs`, which loads the fixture through the same dacite path `video_train.py`'s CLI uses and runs one training step end-to-end against the synthetic data.

Verification: `fme/downscaling/` — pytest 500 passed; pre-commit (ruff, ruff-format, mypy) clean.

fme/downscaling/aggregators/main.py
```diff
@@ -104,35 +109,52 @@ def record_batch(self, outputs: ModelOutputs) -> None:
...
-        bin_indices = torch.bucketize(log_sigma[in_range], self._inner_edges)
+        if sigma.dim() not in (1, 2):
+            raise ValueError(
+                "Expected sigma to have shape (batch,) or (batch, channel)"
+            )
 
-        per_channel: TensorDict = {}
-        for name, loss in outputs.per_sample_channel_loss.items():
+        for i, (name, loss) in enumerate(outputs.per_sample_channel_loss.items()):
+            # Register every channel up front (regardless of whether any sample
```
AnnaKwa: Remove verbose comments unless you think it is necessary for understanding the code

fme/downscaling/video_train.py
```diff
@@ -0,0 +1,524 @@
...
+                if (
+                    self.config.max_test_batches is not None
+                    and b >= self.config.max_test_batches
+                ):
+                    break
+                generated = self.model.generate(
+                    batch, n_samples=self.config.generate_n_samples
+                )
+                for c, name in enumerate(names):
+                    gt = batch.fine.data[name].float()
```
AnnaKwa: Nit: variable name `gt` is unclear what it refers to, can you change this to something that is more obvious to someone reading the code?

fme/downscaling/video_train.py
```diff
@@ -0,0 +1,524 @@
...
+        self.model.module.train()
+        wandb = WandB.get_instance()
+        loss_vs_noise = (
+            LossVsNoiseAggregator() if self.config.log_loss_vs_noise else None
+        )
+        batch: PairedVideoBatchData
+        epoch_loss = 0.0
+        n_batches = 0
+        for i, batch in enumerate(self.train_data.loader):
+            if (
```
AnnaKwa, line 243: Missing `park_if_terminating` call (added recently) that other trainers use- see the non-video trainer for example usage.

fme/core/logging_utils.py
```diff
@@ -155,7 +158,7 @@ def _configure_wandb(
             config=config_copy,
             # wandb reads WANDB_NAME only on the first init in a process, so pass
             # it explicitly for later runs (e.g. segments) to be named correctly.
-            name=os.environ.get("WANDB_NAME"),
+            name=self.name or os.environ.get("WANDB_NAME"),
```
AnnaKwa, line 161: This is bugged for the segmented inference case where if `self.name` is set, the per-segment naming `{name}-segment_N` is not correctly set.

AnnaKwa (changes requested):
A few more comments, most notably the current version of the logging utils wandb name change will break segmented inference naming.

jhhuangchloe:
Thanks for the review, all addressed in 15568be9..57de147a.

1. **Aggregator streamlining** — unified shared/channelwise sigma into one per-(sample, channel) path in `record_batch`. `all_channels` is now a mean in both cases (was sum for shared sigma). Updated docstring + tests.
2. **GCS-dependent configs** — deleted `video_train_xshield.yaml` (superseded prototype). `video_train_xshield_smoke.yaml` is now a real fixture (local synthetic data, `/will/be/overwritten` placeholders) with `test_video_train_xshield_smoke_config_runs` exercising it end-to-end.
3. **Still-verbose comments** — trimmed further; removed one that duplicated the class docstring.
4. **Missing `park_if_terminating`** — added to all 3 per-batch loops in `video_train.py`, matching the non-video trainer.
5. **`gt` naming nit** — renamed to `truth` (3 spots), matching existing usage in `_interior_generation_mae`.
6. **Real bug**: segmented-inference per-segment wandb naming only worked via the `WANDB_NAME` env var — setting `logging.name` in config instead silently reused one name across all segments. Fixed at the source (`config_copy.logging.name` set per segment) + added a test.

AnnaKwa (commented):
Is there a reason why the logging feature to add an optional wandb name via config is needed, as opposed to the default way of setting `WANDB_NAME` via the beaker job submission command? If not, I think it'd be best to just remove the logging and test_segmented changes from this PR so there isn't any added complexity and potential for bugs down the line.

jhhuangchloe:
> Is there a reason why the logging feature to add an optional wandb name via config is needed, as opposed to the default way of setting `WANDB_NAME` via the beaker job submission command? If not, I think it'd be best to just remove the logging and test_segmented changes from this PR so there isn't any added complexity and potential for bugs down the line.

Thanks for the suggestion! Reverted the `logging.name`-based fix and its test entirely, back to the original env-var-only mechanism.

AnnaKwa (approved):

PR #1416, opened 2026-08-06
Add video PMD test-set ensemble inference

jhhuangchloe:
Adds test-set ensemble inference for a trained video PMD checkpoint (from #1415), writing a zarr store that covers the full test period with no gaps.

Changes:
- `fme.downscaling.video_inference`: `VideoInferenceConfig` / `VideoInferenceRunner` / `main` entrypoint -- given a trained checkpoint, runs ensemble inference over a test split and writes a single zarr store with dims (time, ensemble, latitude, longitude). Endpoint frames are the observed ground truth broadcast across the ensemble dimension; interior frames are the generated ensemble. A `frame_source` time-coordinate flags observed (0) vs. generated (1) frames. Config validates up front (`__post_init__`) that data/model `n_timesteps` match, clip stride is tumbling, no repeats or replacement sampling, and `ensemble_chunk_size > 0`.
- `fme.downscaling.video_inference._bare_module`: unwraps `DummyWrapper`/`DDP` with an informative `RuntimeError` when assumptions aren't met.
- `fme.downscaling.video_inference._splice_observed_endpoints`, `_clip_write_slice`, `_validate_contiguous_clips`, `_warn_if_idle_ranks`: extracted helpers for endpoint broadcasting, per-clip write slicing, contiguous-tiling validation, and idle-rank warnings, each with dedicated tests.
- `fme.downscaling.data.config.build_video`: accepts a `drop_last` override so test-set inference can pass `False` to guarantee full coverage; also fixes the inner `DataLoader` to respect the override (previously hardcoded `drop_last=True` independent of the sampler-level flag). Existing training/validation callers are unaffected (default `None` preserves prior behavior).
- `fme.downscaling.data.config._expand_clip_starts_to_frame_times`: expands per-clip start times to the full sorted, deduplicated frame-time axis and computes each clip's index into it -- expands per clip rather than from a global anchor, so it only extrapolates within a clip's own span.
- `fme.downscaling.data.video_datasets.PairedVideoGriddedData`: extended with `clip_start_times` (renamed from `all_times`), `timestep`, `frame_times`, `clip_start_indices`, and `fine_extent_latlon_coords` so the inference layer sources all time/coordinate information from the data layer instead of re-deriving it.
- Zarr output uses sharding (`shard = (n_timesteps - 1, n_ensemble, lat, lon)`) to keep object count proportional to clips rather than individual frame/member pairs.
- Each ensemble chunk is moved to CPU immediately after `generate()` rather than accumulating on GPU, bounding GPU memory to `ensemble_chunk_size`.

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated -- no dependency changes


fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    EMA shadow, exactly mirroring ``VideoTrainer._ema_context()``.
+
+    ``model.module`` is always wrapped (``DummyWrapper`` outside torchrun,
+    ``DistributedDataParallel`` under it), both under the attribute name
+    ``module``. DDP's own ``state_dict()``/``load_state_dict()`` transparently
+    forward to that inner module (no prefix), which is what training's saved
+    checkpoint contains -- but ``DummyWrapper`` does *not* override those
+    methods, so loading directly into ``model.module`` only works under DDP.
+    Loading into ``getattr(model.module, "module", model.module)`` instead
+    reaches the raw net directly in both cases.
```
frodre: This all seems like experiment/setup information that should not be logged as a docstring because it could become stale/incorrect quite quickly.  I would just make sure that the dummy-wrapper vs. ddp handling gets handled gracefully (informative error) so it fails in an understandable way if our assumptions aren't met.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    attrs = {name: dict(ds[name].attrs) for name in out_names if name in ds}
+    return time, attrs
+
+
+@dataclass
+class VideoInferenceConfig:
+    """Config for running test-set inference with a trained video PMD model.
+
+    ``model`` and ``data`` are pasted verbatim from the training config's
+    ``model:`` and ``test_data:`` blocks -- no new schema for those.
```
frodre: Use our docstring format to describe the configuration components.  Remove the long comments below.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    # Cap the number of batches processed per rank; for smoke tests.
+    max_batches: int | None = None
+    # If True, overwrite an existing store at output_path (mode="w") instead
+    # of the safe default (mode="w-", fail if it already exists). Use this
+    # while iterating on a run that keeps failing/retrying; leave False once
+    # a run is expected to succeed, so a completed store can't be clobbered
+    # by accident.
+    overwrite: bool = False
+
+    def configure_logging(self, log_filename: str) -> None:
```
frodre, line 190: This doesn't use the config building format we've started to use more consistently throughout the project.  If you have the `pre-review` skill from this project available it will tell you some adjustments to make for the building from a config.  

In a basic sense, while they are tightly coupled by virtue of proximity, we don't want the building using handoffs to separate functions which have to have explicit knowledge of deeper layers of the config component APIs in order to work.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    griddata = config.data.build_video(
+        train=False, requirements=config.model.data_requirements
+    )
+
+    time, var_attrs = _reference_time_and_attrs(config.data, model.out_names)
+    n_time = len(time)
+    n_timesteps = config.model.n_timesteps
+    clip_stride = n_timesteps - 1  # tumbling clips share only their boundary frame
+
+    # Observed at every clip boundary (0, clip_stride, 2*clip_stride, ...),
```
frodre: Long comments are discouraged because they can mislead the agents if things change since there's no dependency graph.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    # generated everywhere else. If the dataloader's drop_last truncates the
+    # final partial batch, the tail of this array is simply never written and
+    # will show up as a gap (fill value) in the output.
+    frame_source = np.ones(n_time, dtype=np.int8)
+    frame_source[0::clip_stride] = 0
+
+    # griddata.fine_coords reflects the *pre*-lat_extent/lon_extent-crop
```
frodre: Another long comment.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+    for i, batch in enumerate(itertools.chain([first_batch], batch_iterator)):
+        if config.max_batches is not None and i >= config.max_batches:
+            break
+
+        remaining = config.n_ensemble
+        ensemble_chunks: dict[str, list[torch.Tensor]] = {
+            name: [] for name in model.out_names
+        }
+        while remaining > 0:
+            n = min(config.ensemble_chunk_size, remaining)
```
frodre: How big are the chunks usually for 25km data?  If we're generating too many objects, could also consider sharding where the shard is the unit of work, with a reasonable inner chunk size (e.g., time=1)

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+        train=False, requirements=config.model.data_requirements
+    )
+
+    time, var_attrs = _reference_time_and_attrs(config.data, model.out_names)
+    n_time = len(time)
+    n_timesteps = config.model.n_timesteps
+    clip_stride = n_timesteps - 1  # tumbling clips share only their boundary frame
+
+    # Observed at every clip boundary (0, clip_stride, 2*clip_stride, ...),
+    # generated everywhere else. If the dataloader's drop_last truncates the
```
frodre: Shouldn't we always pin drop last to a value that would not remove outputs in the test set? Usually we want full coverage.  Also making sure the fill value is not 0, so it's obvious what's gone should be important if full coverage is not necessary.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,318 @@
...
+        for b in range(batch_size):
+            start_idx = int(np.searchsorted(time, clip_times[b, 0]))
+            if time[start_idx] != clip_times[b, 0]:
+                raise ValueError(
+                    f"Clip start time {clip_times[b, 0]} not found in the "
+                    "reference test time axis; data/config mismatch."
+                )
+            is_last_clip = start_idx + (n_timesteps - 1) == n_time - 1
+            n_frames_to_write = n_timesteps if is_last_clip else n_timesteps - 1
+            time_slice = slice(start_idx, start_idx + n_frames_to_write)
```
frodre: Should convert to to a helper function and write tests so it's clear what cases are covered and what's expected.

frodre (changes requested):
Thanks @jhhuangchloe!  I am requesting a few changes to bring this closer in line with some things in the rest of our codebase.

* use our /pre-review skill to address the configuration and build expecations
* remove some of the long experiment-related details and longer comments, those can go stale quickly and mislead agents.  Always prefer self-documenting code through naming, functions, etc.
* Elaborate on expectations for parallel generation with `drop_last=True`.  I wouldn't expect this to be something we should allow for an inference where the output should cover the full set, but also not sure of exactly the use case in mind here.

jhhuangchloe:
Thanks for the review, all addressed in e3187a09e..726cb4465.

1. **Config/build pattern** — ran this through the `pre-review` skill's builder-pattern rules. `run_inference` reached into `config.data.build_video()` and `config.model.data_requirements` directly; `VideoInferenceConfig.build()` now assembles the model + data loader (mirroring `VideoTrainerConfig.build() -> VideoTrainer`) and returns a `VideoInferenceRunner`. First pass still had the runner holding the whole config and reading `config.data`/`config.model.n_timesteps`/etc. from inside `run()` — same violation, one layer down. Fixed: `build()` now does all config-reading and hands the runner only built objects (model, griddata, time, var_attrs) and plain values; the runner has no knowledge of `VideoInferenceConfig`'s structure.
2. **DummyWrapper/DDP handling** — replaced the docstring explanation + silent fallback with `_bare_module()`, which raises an informative `RuntimeError` if the wrapping assumption isn't met.
3. **Long comments** — trimmed the config docstring, `frame_source` block, and `fine_coords` block; the last is now a one-line `_post_crop_latlon` helper.
4. **Endpoint-splicing/write helper** — extracted `_splice_observed_endpoints` and `_clip_write_slice`, each with dedicated tests covering middle-clip, last-clip, and mismatch cases. Writing these caught a real bug: a clip start past the end of the time axis raised `IndexError` instead of the intended `ValueError` (fixed).
5. **drop_last** — forced `False` always at the inference call site. Also found and fixed a second gap: `build_video`'s underlying `DataLoader` hardcoded `drop_last=True` independent of the sampler-level flag. `build_video` now takes a `drop_last` override controlling both; training/validation callers unaffected.
6. **Fill value** — `ZarrWriter`/`_initialize_zarr` now take an optional `fill_value` (existing callers unaffected); the video-inference writer passes `NaN`, so any cell that somehow goes unwritten reads as obviously missing instead of a silent 0.
7. **Chunking** — `determine_zarr_chunks` never actually subdivides lat/lon for realistic 25km-ish grids (full frame is ~4-4.5MB, under the 10MB target), so `chunks = {time:1, ensemble:1, lat:n_lat, lon:n_lon}` always. Object count = n_time × n_ensemble × n_variables regardless of `target_mb` — e.g. a full-year 3-hourly test set with n_ensemble=32 and 4 vars is ~374k objects. Confirms the concern; sharding (shard = one GPU-work unit, inner chunk time=1) would collapse that a lot.

fme/core/writer.py
```diff
@@ -96,6 +96,7 @@ def _initialize_zarr(
     array_attributes: dict[str, dict[str, str]] | None = None,
     group_attributes: dict[str, str] | None = None,
     mode: str = "w-",
+    fill_value: float | None = None,
```
frodre: The fill-value adjustment is a change that would need to be separated into it's own PR.  However, let's make sure that it's actually necessary before that happens. As I understand, the core assumption is that you get our test-set inference should be full coverage over the requested inference period, right?  If that's the case, why do we need a fill value?

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+class VideoInferenceConfig:
+    """Config for running test-set inference with a trained video PMD model.
+
+    ``model`` and ``data`` are pasted verbatim from the training config's
+    ``model:`` and ``test_data:`` blocks -- no new schema for those.
+
+    Args:
+        checkpoint_path: Path to the trained checkpoint to load.
+        model: Model config, pasted verbatim from training's ``model:`` block.
+        data: Data config, pasted verbatim from training's ``test_data:`` block.
```
frodre: remove the "pasted" text

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+                "n_ensemble": str(self.n_ensemble),
+                "use_ema": str(self.use_ema),
+            },
+            nondim_coords={
+                "frame_source": xr.DataArray(frame_source, dims=[TIME_NAME]),
+            },
+            mode="w" if self.overwrite else "w-",
+            time_calendar="julian",
+            # Unwritten cells (e.g. a killed/retried run) stay NaN rather than
+            # a silently-plausible-looking 0.
```
frodre: I could see fill value being useful for allowing runs to be pre-empted / restarted.  Still should be a separate PR.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+    subset = src.subset
+    if not isinstance(subset, TimeSlice):
+        raise ValueError(
+            "Expected a TimeSlice subset on data.fine[0] for video inference, "
+            f"got {type(subset)}."
+        )
+    ds = ds.sel(time=slice(subset.start_time, subset.stop_time))
+    time = ds["time"].values
+    attrs = {name: dict(ds[name].attrs) for name in out_names if name in ds}
+    return time, attrs
```
frodre: I think the built `PairedVideoGriddedData` object has a representation of the times used that we could source from?

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+                ),
+                "checkpoint_path": self.checkpoint_path,
+                "n_ensemble": str(self.n_ensemble),
+                "use_ema": str(self.use_ema),
+            },
+            nondim_coords={
+                "frame_source": xr.DataArray(frame_source, dims=[TIME_NAME]),
+            },
+            mode="w" if self.overwrite else "w-",
+            time_calendar="julian",
```
frodre: Could we also grab the calendar from the gridded data to be consistent?

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+        return model
+
+    def build(self) -> "VideoInferenceRunner":
+        """Build the model, load its checkpoint, and set up the test-set loader."""
+        model = self.build_model(get_device())
+        griddata = self.data.build_video(
+            train=False, requirements=self.model.data_requirements, drop_last=False
+        )
+        time, var_attrs = _reference_time_and_attrs(self.data, model.out_names)
```
frodre: If grabbing from the gridded data, then this portion can be removed.  Attributes should be accessible in a similar from our loaders.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+
+            remaining = self.n_ensemble
+            ensemble_chunks: dict[str, list[torch.Tensor]] = {
+                name: [] for name in model.out_names
+            }
+            while remaining > 0:
+                n = min(self.ensemble_chunk_size, remaining)
+                generated = model.generate(batch, n_samples=n)
+                for name in model.out_names:
+                    ensemble_chunks[name].append(generated[name])
```
frodre: Should probably offload these gathering operations to CPU rather than keeping each chunk on GPU given the chunk is chosen in order to fit within memory.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+            nondim_coords={
+                "frame_source": xr.DataArray(frame_source, dims=[TIME_NAME]),
+            },
+            mode="w" if self.overwrite else "w-",
+            time_calendar="julian",
+            # Unwritten cells (e.g. a killed/retried run) stay NaN rather than
+            # a silently-plausible-looking 0.
+            fill_value=np.nan,
+        )
+        writer.initialize_store(data_dtype=np.float32)
```
frodre, line 343: This involves a blocking operation that will deadlock if there are not any batches for other ranks. Not really something that would come up with expected usage, but we should a check that fails fast on all ranks if that's the case. It probably needs to be at the top of this run method, something that checks to total times from the loader vs. the world size.  Might have to add some validation that checks that the paired video dataset is not configured with any repeats, or take that into consideration.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+
+        # Observed at every clip boundary (0, clip_stride, 2*clip_stride, ...),
+        # generated everywhere else.
+        frame_source = np.ones(n_time, dtype=np.int8)
+        frame_source[0::clip_stride] = 0
+
+        batch_iterator = griddata.get_generator()
+        first_batch = next(batch_iterator)
+        lat, lon = _post_crop_latlon(first_batch)
+        n_lat, n_lon = len(lat), len(lon)
```
frodre: This is the type of information that should be stored on the `GriddedData` for video datasets.  We do something similar for the paired loaders with the coarse data for patching and lon 0 crosses, but you'd just want to store the fine instead.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+        griddata = self.data.build_video(
+            train=False, requirements=self.model.data_requirements, drop_last=False
+        )
+        time, var_attrs = _reference_time_and_attrs(self.data, model.out_names)
+        return VideoInferenceRunner(
+            model=model,
+            griddata=griddata,
+            time=time,
+            var_attrs=var_attrs,
+            n_timesteps=self.model.n_timesteps,
```
frodre: We are handing the model, so why not just use the model value within the runner?

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+    data: PairedVideoLoaderConfig
+    output_path: str
+    experiment_dir: str
+    logging: LoggingConfig
+    n_ensemble: int = 32
+    ensemble_chunk_size: int = 8
+    use_ema: bool = True
+    max_batches: int | None = None
+    overwrite: bool = False
+
```
frodre, line 160: Post init validation that checks the data n_timesteps matches model n_timesteps, and that the clip stride on the data config is consistent with the timesteps, since we don't want any stride mismatch to mess up the inference coverage.

frodre: In general, it would be good to not rederive this sort of information and have a single source of truth if possible.

fme/downscaling/test_video_inference.py
```diff
@@ -0,0 +1,123 @@
...
+    assert result == slice(8, 17)
+
+
+def test_clip_write_slice_raises_on_start_time_mismatch():
+    time = np.array(
+        [cftime.DatetimeProlepticGregorian(2020, 1, d) for d in range(1, 5)]
+    )
+    bad_start = cftime.DatetimeProlepticGregorian(2020, 2, 1)
+    with pytest.raises(ValueError, match="not found in the reference test time axis"):
+        _clip_write_slice(bad_start, time, n_timesteps=2)
```
frodre: Could use a quick end-to-end test of the inference, since it is an entrypoint.

fme/downscaling/video_inference.py
frodre: It would be similar to `all_times`, but that becomes an overloaded term in that this really refers to the clip or "frame" start times, right?  It should use different naming than `all_times` which gives full coverage in our other data configurations.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+    checkpoint_path: str
+    model: VideoDiffusionModelConfig
+    data: PairedVideoLoaderConfig
+    output_path: str
+    experiment_dir: str
+    logging: LoggingConfig
+    n_ensemble: int = 32
+    ensemble_chunk_size: int = 8
+    use_ema: bool = True
+    max_batches: int | None = None
```
frodre: In considering the consistency checks for runtime, `max_batches` has the issue for testing that it would need a conditional to ignore in this case.  It might be simpler codewise to remove this option.

Typically for testing, we've just adjusted the times on the `XarrayDataConfig` to reduce the compute.  This would of course have to be done so it gives complete clip, but it's not overly complicated.

frodre (changes requested):
Thanks @jhhuangchloe for the updates.  The build process is looking much cleaner.  I have some more updates requested. Here are the big ones, the rest are mentioned in individual comments.

1. The fill-value would need to be a separate PR, but I don't think it's necessary to have this change.  Remove its usage and we can introduce it when needed.  My thinking is that partially written outputs are a result of a crashed or preempted beaker run, and we don't support resumption with this inference yet.  We would typically not use crashed run output without caution, so the zeros are not an issue.
2. There are spots where you are re-deriving coordinates relating to time, lat, lon, and tiling.  Most of these are already implemented in certain locations, or if not, should be added to objects (like this variant of `GriddedData`) that traditionally have been source of truth and consistency checks rather than yet another derivation.  I highlighted what should be used and validated in the comments.
3. It would be nice to fail fast in the cases of tiling coverage mismatches, or clips / world size mismatche for the sampler.
4. The build and run need some test coverage.  Make sure it's still fast to run using patterns we do for other APIs.  No need to have a multi-gpu integration test, but the fast failures should be checked as well (e.g., world size and num available batches mismatch)

fme/downscaling/video_inference.py
frodre: Also check that the `ensemble_chunk_size` > 0

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,406 @@
...
+    last clip also writes its own trailing endpoint.
+    """
+    n_time = len(time)
+    start_idx = int(np.searchsorted(time, clip_start_time))
+    if start_idx >= n_time or time[start_idx] != clip_start_time:
+        raise ValueError(
+            f"Clip start time {clip_start_time} not found in the reference "
+            "test time axis; data/config mismatch."
+        )
+    is_last_clip = start_idx + (n_timesteps - 1) == n_time - 1
```
frodre: For this to be triggered the total size of the dataset needs to exactly match.  worth a check that there's remainder in the times of the dataset for the given clip size.

`(n_time - 1) % (n_timesteps - 1) == 0`

jhhuangchloe:
Thanks for the second round, addressed in b579c5f95..c1a567697.

1. **Fill value** — reverted entirely (both the `ZarrWriter` change and its usage here); happy to revisit as its own PR if/when pre-emption support needs it.
2. **Re-derived coordinates/time/attrs** — added `fine_extent_latlon_coords` and `timestep` to `PairedVideoGriddedData` (mirrors `PairedGriddedData.coarse_extent_latlon_coords`; both were already computed in `build_video`, just never exposed). `video_inference.py` now sources lat/lon and the time calendar from `griddata`, and reconstructs the full per-frame time axis from `griddata.all_times` (clip starts) + the new `timestep` field — no more re-opening the input zarr directly, so the `XarrayDataConfig`/`TimeSlice` isinstance requirements on `data.fine[0]` are gone too. Variable attrs come from `griddata.variable_metadata[name].as_attrs()`. Also dropped the redundant `n_timesteps` constructor arg on the runner — it already has `model`, which carries `n_timesteps` itself.
3. **Fail-fast checks** — added `_validate_world_size` (a rank with zero clips would `StopIteration` before reaching `writer.initialize_store()`'s barrier, hanging every other rank — this is a real deadlock, now checked at the top of `run()` before anything barrier-requiring) and `__post_init__` validation: `data.n_timesteps == model.n_timesteps`, `clip_start_stride == n_timesteps - 1` (tumbling only), `data.repeat == 1`, `ensemble_chunk_size > 0`. (The separate full-coverage check I'd added is gone too — once the time axis is reconstructed from the clips themselves, it tiles evenly by construction, so that check could never fail.)
4. **GPU memory** — each ensemble chunk now moves to CPU right after `generate()` instead of accumulating on GPU across the whole while-loop before `torch.cat` — `ensemble_chunk_size` was chosen to bound `generate()`'s memory, which the accumulation was undoing.
5. **max_batches** — removed. It conflicted with the new fail-fast checks (would've needed a bypass just for testing); tests now trim the input time range instead, matching the pattern elsewhere.
6. **Test coverage** — added an end-to-end test of `build().run()` against a small synthetic zarr store, plus tests for every new validation/helper (including `_reference_time_axis`).

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+            remaining = self.n_ensemble
+            ensemble_chunks: dict[str, list[torch.Tensor]] = {
+                name: [] for name in model.out_names
+            }
+            while remaining > 0:
+                n = min(self.ensemble_chunk_size, remaining)
+                generated = model.generate(batch, n_samples=n)
+                # ensemble_chunk_size bounds generate()'s GPU memory use;
+                # move each chunk off-GPU immediately so accumulating chunks
+                # here doesn't undo that bound.
```
frodre: Remove this comment

fme/downscaling/data/video_datasets.py
```diff
@@ -310,8 +311,10 @@ class PairedVideoGriddedData:
     n_timesteps: int
     dims: list[str]
     variable_metadata: Mapping[str, VariableMetadata]
-    all_times: xr.CFTimeIndex
-    fine_coords: LatLonCoordinates
+    all_times: xr.CFTimeIndex  # clip start times
```
frodre: Suggest renaming to `clip_start_times`

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+    all_times: xr.CFTimeIndex, n_timesteps: int, timestep: datetime.timedelta
+) -> np.ndarray:
+    """Full per-frame time axis spanning every tumbling clip's coverage.
+
+    ``all_times`` holds each clip's start time. Tumbling clips share their
+    boundary frame, so consecutive clips together cover
+    ``(len(all_times) - 1) * (n_timesteps - 1) + n_timesteps`` distinct
+    frames at ``timestep`` spacing, starting from the first clip's start time.
+    """
+    n_time = (len(all_times) - 1) * (n_timesteps - 1) + n_timesteps
```
frodre: Why can't this also be moved into the data layer?  From the underlying dataset we have the start time, the timestep, total number of clips.  Seems like we shoudl be able to construct all the information we need without having to reconstruct it all here.  Essentially just use the clip time starts as indexing into frame times, slices, etc.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+def _clip_write_slice(clip_start_time, time: np.ndarray, n_timesteps: int) -> slice:
+    """Time-axis slice a clip should write to, given its start time.
+
+    Tumbling clips share their boundary frame with the next clip, so every
+    clip but the last writes only its first ``n_timesteps - 1`` frames; the
+    last clip also writes its own trailing endpoint.
+    """
+    n_time = len(time)
```
frodre: Similarly, I would imagine being able to get this information from the data rather than reconstructing here.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+            )
+        if self.data.clip_start_stride != self.model.n_timesteps - 1:
+            raise ValueError(
+                "Test-set inference requires tumbling clips (clip_start_stride "
+                f"== n_timesteps - 1); got clip_start_stride="
+                f"{self.data.clip_start_stride} for n_timesteps="
+                f"{self.model.n_timesteps}. A different stride would leave "
+                "gaps or overlaps in the output's write coverage."
+            )
+        if self.data.repeat != 1:
```
frodre, line 175: Should also make sure that we cannot use `sample_with_replacement` for the underlying data

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+            data_shape=(n_time, self.n_ensemble, len(lat), len(lon)),
+            bytes_per_element=4,
+        )
+
+        writer = ZarrWriter(
+            path=self.output_path,
+            dims=DIMS,
+            coords=coords,
+            data_vars=model.out_names,
+            chunks=chunks,
```
frodre, line 325: Chunks always default to (1, 1) for time ensemble, which would create a large number of files. We should use the sharding to fix that.  For our other downscaling inference it's usually set to the size of the output on GPU, but here I think we could just the clip length (`ntimesteps-1`) and the ensemble size.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+            f"Only {n_clips} clip(s) available but {world_size} rank(s) "
+            "requested; every rank needs at least one clip."
+        )
+
+
+@dataclass
+class VideoInferenceConfig:
+    """Config for running test-set inference with a trained video PMD model.
+
+    ``model`` and ``data`` are pasted verbatim from the training config's
```
frodre: Still has remaining pasted section in the description

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
...
+def _validate_world_size(n_clips: int, world_size: int) -> None:
+    """A rank with zero clips would raise ``StopIteration`` before ever
+    reaching the barrier in ``writer.initialize_store()``, hanging every
+    other rank. This check is deterministic and identical on every rank, so
+    it can safely fail fast without any collective communication.
+    """
+    if n_clips < world_size:
+        raise RuntimeError(
+            f"Only {n_clips} clip(s) available but {world_size} rank(s) "
+            "requested; every rank needs at least one clip."
```
frodre: You updated the pre-inference loop so a stop iteration would not fire before reaching the barrier making the check unnecessary and the text incorrect.  This is no longer strictly necessary, but you could just switch to issuing a warning that there will be an idle GPU.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,422 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, Allen Institute for AI
+# SPDX-License-Identifier: Apache-2.0
+"""Test-set inference for the endpoint-conditioned video diffusion model.
+
+Given a trained checkpoint, runs an ensemble of samples over a test split:
+observed daily endpoint snapshots (0h, 24h) in, infilled 3-hourly interior
+frames (3h..21h) out. Writes a single zarr store matching the input dataset's
```
frodre: There are some specifics here to the experiment which should be removed.

fme/downscaling/video_inference.py
frodre: Here's a summary from claude of something I think would be more refined for handling times.

## Clip/frame consolidation

### Problem

`run()` reconstructs the output frame axis and the clip-boundary layout from a single anchor plus stride arithmetic, when the data layer already knows both.

`_reference_time_axis` reads only `all_times[0]` — starts `1..n-1` contribute nothing but `len()`. The count and the values come from different places, so they can't disagree visibly. It also extrapolates across the entire run, including across any `XarrayConcat` boundary, which nothing validates.

### Data layer — `PairedVideoGriddedData` / `build_video`

| Field | Meaning |
| --- | --- |
| `clip_start_times` | rename of `all_times`; one entry per clip, semantics unchanged |
| `frame_times` | sorted union of `{t_k + j * timestep : j in 0..n_timesteps-1}` over all clip starts; dedupes shared boundaries, length `n_clips * (n_timesteps - 1) + 1`. This is the store's `time` coordinate. |
| `clip_start_indices` | indices into `frame_times` where each clip begins, so `frame_times[clip_start_indices] == clip_start_times` |

Expand **per clip start**, not from one global anchor. `_get_timestep` (`fme/core/dataset/xarray.py:1011`) raises if a dataset's times aren't uniformly spaced, so a non-`None` `timestep` is already proof of within-dataset uniformity — and a clip never spans a dataset boundary, so per-clip expansion relies only on the guarantee that actually holds.

The total time length ends at the clip boundary, i.e., times past the last clip's end that don't have a corresponding endpoint can't be generated, so excluding them is correct.

Also move the `timestep is None` check out of `VideoInferenceConfig.build()` and into `build_video` — it's the data layer's own precondition, and that's where the uniformity guarantee is established.

No changes needed in `fme/core`.

### Inference layer — `video_inference.py`

1. Delete `_reference_time_axis`. `run()` uses `griddata.frame_times`.
2. Drop the `time` parameter from `VideoInferenceRunner.__init__` and the `time=` kwarg in `build()` — it's passed straight through and stored for nothing else.
3. `frame_source`: replace `frame_source[0::clip_stride] = 0` with

   ```python
   frame_source[griddata.clip_start_indices] = 0
   frame_source[-1] = 0  # final frame is an observed endpoint, not a clip start
   ```

   The current stride expression covers the final frame only incidentally, because the spacing happens to be uniform.
4. `_clip_write_slice`: `is_last_clip` becomes `start_idx == clip_start_indices[-1]` (equivalently, no other clip starts at `start_idx + n_timesteps - 1`) instead of `start_idx + (n_timesteps - 1) == n_time - 1`.
5. Delete `_validate_world_size`, or downgrade it to a warning. `initialize_store()` barriers unconditionally before the loop and the loop contains no collectives, so an empty rank skips nothing — the deadlock it describes can't occur. It also rejects configurations that work: with `ContiguousDistributedSampler` at `drop_last=False`, 3 clips on 4 ranks gives ranks 0–2 an empty index range and rank 3 all three, which is correct, just serialized.

### Stays in inference

- The `searchsorted` lookup — a batch carries its clip's *times* (`batch.fine.time`), not its index.
- The `n_timesteps` vs `n_timesteps - 1` rule — that's output-splicing policy (don't double-write a shared boundary), not a dataset property. A metrics consumer over the same clips would want all frames every time.

### Net effect

`build()` shrinks, the runner loses a field, three stride/position assumptions disappear, and the invariants become testable in `test_video_datasets.py` against real dataset fixtures instead of hand-built `CFTimeIndex`es in the inference tests.

frodre (commented):
Okay, this is quite close.  I have a few cleanups and one more substantial change that will hopefully consolidate most of the clip/frame handling.

fme/downscaling/video_inference.py
jhhuangchloe: Removed.

fme/downscaling/data/video_datasets.py
jhhuangchloe: `PairedVideoGriddedData.all_times` is now `clip_start_times`.

fme/downscaling/video_inference.py
jhhuangchloe: Implemented as specified:
- `clip_start_times` (renamed), `frame_times`, `clip_start_indices` added to `PairedVideoGriddedData`; `timestep is None` now raises in `build_video`.
- `_reference_time_axis` deleted; `run()` uses `griddata.frame_times`.
- `time` param dropped from `VideoInferenceRunner.__init__`/`build()`.
- `frame_source` now set via `frame_source[griddata.clip_start_indices] = 0` + `frame_source[-1] = 0`.
- `_clip_write_slice`'s `is_last_clip` now compares against the last clip's index rather than arithmetic on `n_time`.
- `_validate_world_size` deleted, replaced by a warning.
- `searchsorted` lookup and the `n_timesteps`/`n_timesteps - 1` splicing rule stayed in inference.

fme/downscaling/video_inference.py
jhhuangchloe: Updated. It now takes `frame_times` and the last clip's index straight from `griddata`.

fme/downscaling/video_inference.py
jhhuangchloe: Added to `__post_init__`: raises if `data.sample_with_replacement is not None`.

fme/downscaling/video_inference.py
jhhuangchloe: Added `shards={time: n_timesteps - 1, ensemble: n_ensemble}` to the `ZarrWriter` call, matching the other downscaling inference's convention of sharding to one unit of GPU work. Verified with a test asserting the actual shard shape on the written store, not just that the argument was passed.

fme/downscaling/video_inference.py
jhhuangchloe: Fixed

fme/downscaling/video_inference.py
jhhuangchloe: `_validate_world_size` (raised) is now `_warn_if_idle_ranks` (logs a warning, doesn't block a valid, just-serialized run).

fme/downscaling/video_inference.py
jhhuangchloe: Removed the "(0h, 24h)"/"(3h..21h)" cadence specifics from the module docstring; kept the general description.

jhhuangchloe:
Thanks for the reply! Replied to individual threads (see above); addressed in 6d63ab178..00f28c91b.

fme/downscaling/data/config.py
```diff
@@ -740,7 +764,12 @@ def build_video(
                 f"Expected clips of {self.n_timesteps} timesteps, got "
                 f"{dataset_fine.sample_n_times}."
             )
-        all_times = dataset_fine.sample_start_times
+        clip_start_times = dataset_fine.sample_start_times
```
frodre: nit: this is not yet the `clip_start_times` since it isn't strided.  Would be more readable as something like `fine_start_times` which gets translated into `clip_start_times` down below.

fme/downscaling/data/video_datasets.py
```diff
@@ -310,8 +311,14 @@ class PairedVideoGriddedData:
...
-    all_times: xr.CFTimeIndex
-    fine_coords: LatLonCoordinates
+    clip_start_times: xr.CFTimeIndex
+    timestep: datetime.timedelta  # spacing between consecutive frames
+    frame_times: xr.CFTimeIndex  # every distinct frame time covered by a clip
+    clip_start_indices: (
+        np.ndarray
+    )  # frame_times[clip_start_indices] == clip_start_times
+    fine_coords: LatLonCoordinates  # full-domain fine coordinates
+    fine_extent_latlon_coords: LatLonCoordinates  # post-crop fine coordinates
```
frodre: nit: Rather than using inline comments, expand a docstring with attribute information for the data class above

fme/downscaling/test_video_inference.py
```diff
@@ -0,0 +1,321 @@
...
+        experiment_dir=str(tmp_path / "train_exp"),
+        logging=LoggingConfig(
+            log_to_screen=False, log_to_wandb=False, log_to_file=False
+        ),
+        ema=EMAConfig(decay=0.99),
+    )
+    trainer = trainer_config.build()
+    trainer.train()
+    ckpt_path = str(tmp_path / "e2e.ckpt")
+    _save_checkpoint(trainer, ckpt_path)
```
frodre, line 293: nit: a bit wierd to require training a model in an inference test, could we just build a small model and save.  Although I don't love all the layers of fixtures in `inference/test_inference.py`, I think we could have some in-file helpers that generate a small model, build it, and save a checkpoint, rather than having it train and save a model in this file.  Up to you though because this works as and it doesn't take a significant amount of time as far as I can tell.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,402 @@
...
+        model, griddata = self.model, self.griddata
+        frame_times = griddata.frame_times
+        last_clip_start_idx = griddata.clip_start_indices[-1]
+
+        _warn_if_idle_ranks(len(griddata.clip_start_times), dist.world_size)
+
+        logger.info(f"Number of parameters: {count_parameters(model.modules)}")
+
+        n_time = len(frame_times)
+        n_timesteps = model.n_timesteps
```
frodre, line 296: Claude is worried about a couple more pathways to non-contiguous inputs for data that would trip up the contiguous clip assertion.  I don't think they are configured by default, but multi-dataset `XarrayConcatDataset` could break across the dataset time concatenation, or using `Subset` or `RepeatedInterval`.  Regardless, a quick check to make sure our assumptions hold can be a last line of defense.  Something along these lines

```
expected_starts = np.arange(0, len(frame_times) - n_timesteps + 1, n_timesteps - 1)
if not np.array_equal(griddata.clip_start_indices, expected_starts):
    raise ValueError(
        f"Clip starts {griddata.clip_start_indices.tolist()} do not tile the "
        f"{len(frame_times)}-frame time axis in steps of {n_timesteps - 1}; "
        "some frames would be left unwritten. Check for multiple entries in "
        "data.fine/data.coarse or a strided data subset."
    )
```

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,402 @@
...
+        """Build the model from config and load trained weights.
+
+        The training config used ``validate_using_ema: true``, so the checkpoint
+        that was selected as "best" was evaluated with EMA-swapped weights. To
+        reproduce that quality at inference time, load the raw state dict first
+        (establishes buffers) and then overwrite the trainable params with the
+        EMA shadow, exactly mirroring ``VideoTrainer._ema_context()``.
+        """
+        model = self.model.build()
+        ckpt = torch.load(self.checkpoint_path, map_location=device, weights_only=False)
```
frodre:
```suggestion
        ckpt = torch.load(self.checkpoint_path, map_location="cpu", weights_only=False)
```

frodre: We load the optimizer state to cpu rather than gpu, because we don't need it in these inference settings.

fme/downscaling/video_inference.py
```diff
@@ -0,0 +1,402 @@
...
+    is_last_clip = start_idx == last_clip_start_idx
+    n_frames = n_timesteps if is_last_clip else n_timesteps - 1
+    return slice(start_idx, start_idx + n_frames)
+
+
+def _warn_if_idle_ranks(n_clips: int, world_size: int) -> None:
+    if n_clips < world_size:
+        logger.warning(
+            f"Only {n_clips} clip(s) available but {world_size} rank(s) "
+            "requested; some ranks will sit idle."
```
frodre:
```suggestion
            "requested; all but one rank will sit idle."
```

contiguous hands off `items // world_size` chunks to each  rank and the remainder goes to a single rank

frodre (approved):
Thanks @jhhuangchloe for all your work getting this one updated.  A few minor points to consider, but overall I think we can merge this once those are addressed.

jhhuangchloe:
PR title and description updated by the `write-pr-prose` skill, running on **claude-opus-4-6** (Opus 4.6, 1M context). No warnings from the context-gathering script; all sections (existing prose, issue comments, reviews, inline review comments, and full diff) were retrieved successfully.

PR #1419, opened 2026-08-06
Add a Beaker urgent-priority balancer

mcgibbon:
Beaker does not enforce our urgent-priority allocation (72 GPU slots on `ai2/jupiter`, 32 on `ai2/titan`), so queueing urgent jobs can silently put the team over it. That effectively bans queueing urgent work. This adds a script that keeps us inside the allocation automatically, so urgent jobs can be queued freely.

Users can set an env var on their jobs `CM_PRIORITY` to `low`, `normal`, `high` or `urgent`. The balancer only modifies such jobs, other urgent jobs still count against the allocation.

When we have more than the allocated number of urgent slots (running or in-queue), the balancer downgrades the lowest-priority CM_PRIORITY jobs (longest-running breaks ties). When we have fewer, it upgrades the highest-priority job (most time in queue breaks ties). It also will downgrade running beaker-urgent-priority jobs with lower CM_PRIORITY to make space for higher CM_PRIORITY jobs that are stuck in the queue (but not for jobs already running, as we would need to stop them to upgrade the priority).

Decisions are made per **replica group**, not per job. A multi-node job is all-or-nothing.

When downgrading a running job, priority is changed in-place without pre-emption. Upgrading priority of a running job would require stopping it, so rather than do that we never upgrade the priority of jobs already running.

Two kinds of job count against the allocation but are never changed in either direction: **interactive sessions**, which have a person attached to them, and jobs at **`immediate` priority**, which outranks urgent and which Beaker only grants with a human-supplied reason.

The allocation is the team's, not one workspace's. The balancer manages `ai2/ace` and also **counts the urgent slots held in `ai2/climate-titan`** (`--observe`, repeatable; `--no-observe` to disable), without ever modifying anything there. Without this it would grant urgent against slots that are already occupied, handing the same allocation out twice — not hypothetical, since climate-titan currently holds 33 of the 32 titan slots. `CM_PRIORITY` is not read in an observed workspace: opting in is what makes a job the balancer's to move, and nothing there is. Beaker returns the whole org's unfinished jobs in one call, so observing another workspace costs a name lookup rather than a second pass.

A job is balanced against the clusters its slots are charged to. A **running** job occupies one cluster and is balanced against that one alone, whatever it was submitted eligible for — its slots can be seen and reclaimed from that cluster's budget. A **queued** job targeting multiple clusters is managed **pessimistically**: granting it urgent requires room on every cluster it could land on and charges all of them, since nothing can preempt an urgent job and granting it commits the allocation on whichever cluster it picks. Demoting a queued multi-cluster job frees all of its clusters. This means a multi-cluster grant costs more headroom than a single-cluster one when allocations are tight, which is the right tradeoff — the alternative is to grant urgent against a budget the job might consume elsewhere.

Changes:
- `scripts/beaker_balancer/balance.py` — the balancer. Pure `decide()` computes the priority changes; `fetch_jobs`/`set_priority` wrap Beaker. Supports `--dry-run`, `--workspace`, `--limit CLUSTER=SLOTS`, `--observe`/`--no-observe` and `--interval`, and is stateless and convergent so it is safe to run from cron. Fails soft per job, so a teammate's job it lacks permission on is logged and skipped rather than failing the pass, and orders demotions before promotions so any prefix of a partially applied pass stays within allocation.
- `scripts/beaker_balancer/test_balance.py` — decision-logic tests, including randomised populations of replica groups, sessions and `immediate` jobs asserting the invariants that matter: a pass never pushes usage above the allocation, any prefix of a pass stays within it, a second pass is always a no-op, a group is never left holding urgent on only some ranks, sessions and `immediate` jobs are never touched, and a placed job is never raised. Usage is computed with an oracle written independently of the production accounting, and a further test asserts the random populations actually reach the states that matter.
- `scripts/beaker_balancer/test_beaker_io.py` — the Beaker-facing layer driven against a fake client built from real protobuf messages, covering env-var and placement-constraint parsing, replica-group and session detection, slot accounting, dry-run, and fail-soft behaviour.
- `scripts/beaker_balancer/README.md` — the allocation model, opt-in instructions, and the rationale behind each rule.

Cluster and workspace names given on the command line are resolved at startup, since a typo would silently manage or count nothing and look exactly like a quiet cluster or an idle workspace.

Verified against the live `ai2/ace` workspace: slot accounting reproduces the true per-cluster urgent usage (independently cross-checked), and feeding the real job population through `decide()` converges in one pass without exceeding either allocation.

- [x] Tests added

### DIFF
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -171,12 +171,19 @@ committed future occupant, not a maybe.
 
 Being charged to a cluster and being balanced against it are the same question,
 so they are answered the same way. A placed job is managed against the single
-cluster it landed on. A queued job targeting multiple clusters is managed
-**pessimistically** against all of them: granting it urgent requires room on
-every cluster, and the grant charges all of them. Demoting it frees all of
-them. This means a multi-cluster grant costs more than a single-cluster one
-when headroom is tight, which is the right tradeoff — the alternative is to
-grant urgent against a budget the job might consume elsewhere.
+cluster it landed on. A queued job targeting multiple budgeted clusters is
+managed **pessimistically** against all of them: granting it urgent requires
+room on every cluster, and the grant charges all of them. Demoting it frees
+all of them. This means a multi-cluster grant costs more than a single-cluster
+one when headroom is tight, which is the right tradeoff — the alternative is
+to grant urgent against a budget the job might consume elsewhere.
+
+A job targeting a mix of budgeted and non-budgeted clusters — e.g. both
+`ai2/jupiter` (budgeted) and `ai2/ceres` (no allocation) — is managed against
+the budgeted ones only. While queued, only the budgeted clusters are checked
+and charged. If it lands on a non-budgeted cluster, its slots stop counting
+against the allocation entirely, since it is not consuming a budgeted cluster's
+slots.
 
 A placed multi-cluster job is different: it is on one cluster and only that
 cluster's budget is affected, whether the job is granted or demoted. It could
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -32,13 +32,18 @@ count per pass:
   to one cluster if you want it balanced. Most `ai2/ace` jobs today target
   `titan+jupiter`, so expect this to be the common case during adoption.
 - **It targets a cluster with no allocation** (`ai2/saturn`, `ai2/ceres`).
-- **It is part of a replica group.** The allocation walk would grant urgent to
-  some ranks and not others, which for a synchronised-start group wastes the
-  slots it did grant. It is also unconfirmed whether `UpdateJobSourcePriority`
-  acts per job or per source; if per source, one call would move every replica.
+- **It is an interactive session.** There is a person attached to it, and
+  taking their priority away mid-session is not the script's call.
+- **It is at `immediate` priority.** Beaker requires a human-supplied reason for
+  `immediate`, so it represents a deliberate decision the balancer should not
+  quietly undo. `immediate` is not an accepted `CM_PRIORITY` value either: it is
+  not something the script may hand out.
 - **It has landed on a node the script cannot resolve to a cluster.** Its slots
   are then charged to every budget as a precaution, which is inconsistent with
   managing it against one.
+
+Sessions and `immediate` jobs still count against the allocation — see
+[Accounting](#accounting).
 
 ### What may surprise you
 
@@ -55,33 +60,61 @@ count per pass:
   a `CM_PRIORITY` level, the job that has held its GPUs longest is the first to
   lose urgent, on the grounds that it has already had its turn.
 
+## Replica groups
+
+Beaker runs a multi-node job as one job per rank. The balancer decides for the
+whole **replica group** at once and grants urgent all-or-nothing. A lone job is
+simply a group of one.
+
+A half-granted group is the worst available outcome: it cannot start until every
+rank is placed, so it waits at the priority of its lowest rank while the granted
+ranks hold slots the allocation has already spent. A four-rank group needing 80
+slots against 72 free therefore gets nothing, rather than stranding 64 slots on
+a job that still cannot run.
+
+A group is managed only if *every* rank is: one unlabelled rank, one session, or
+one at `immediate` makes the whole group untouchable. The same applies if the
+ranks disagree about their `CM_PRIORITY`, or if the group comes back short —
+only unfinished jobs in one workspace are read, so granting the ranks that
+happen to be visible would half-grant the real group. Skipped groups are counted
+in the log.
+
+Splits *below* urgent are left alone deliberately. They cost the allocation
+nothing, and raising a queued rank whose siblings are stuck at a lower priority
+is what gets the group placed.
+
+The grouping is keyed on the replica group rather than the workload, since one
+workload can hold independent tasks that have no reason to move together.
+
 ## What it does
 
 Each pass recomputes the whole allocation from scratch and applies the
 difference:
 
-1. Read every unfinished job in the workspace.
-2. Subtract the slots held at urgent by jobs it cannot modify. Those are a
-   fixed charge; only what is left is available.
-3. Walk the managed jobs in `CM_PRIORITY` order, granting urgent while the
-   job's cluster has room.
-4. Set every managed job to urgent if granted, or to its resting priority if
+1. Read every unfinished job in the workspace and group it with its ranks.
+2. Subtract the slots held at urgent or `immediate` by jobs it cannot modify.
+   Those are a fixed charge; only what is left is available.
+3. Walk the managed groups in `CM_PRIORITY` order, granting urgent while the
+   group's cluster has room for all of it.
+4. Set every managed group to urgent if granted, or to its resting priority if
    not.
 
 Recomputing rather than adjusting incrementally is what makes step 3 work: a
-`high` job is always considered before any `normal` job, so it never has to
-wait for slots a lesser job already took. If `normal` jobs are holding urgent
+`high` group is always considered before any `normal` one, so it never has to
+wait for slots a lesser one already took. If `normal` jobs are holding urgent
 slots that a `high` job needs, they are dropped to make room.
 
-Within one `CM_PRIORITY` level, urgent is granted to running jobs that already
-hold it first — so a queued job always gives up its slot before a running one —
-ordered so the job that has held a GPU longest is the first to lose it. Queued
-jobs follow, longest-waiting first. The two orderings use different clocks:
-time on GPU for running jobs, and time since entering the queue for queued
-ones. The queue clock resets when a job is preempted and requeued, so a job
-that has been bounced does not keep seniority it no longer has.
+Within one `CM_PRIORITY` level, urgent is granted to running groups that already
+hold it first — so a queued group always gives up its slot before a running one
+— ordered so the group that has held its GPUs longest is the first to lose it.
+Queued groups follow, longest-waiting first. The two orderings use different
+clocks: time on GPU for running groups, and time since entering the queue for
+queued ones. A group's queue clock is that of its *latest* rank, because it
+cannot start until the last one is placed; since the clock resets when a rank is
+preempted and requeued, a group that has been bounced does not keep seniority it
+no longer has.
 
-A job too large for the remaining slots is skipped and smaller ones behind it
+A group too large for the remaining slots is skipped and smaller ones behind it
 still get a chance. Nothing is reserved for it, because a level that could not
 fit has already taken everything it was entitled to.
 
@@ -100,13 +133,17 @@ BeakerPermissionsError: cannot increase priority for running job "01K..."
 ```
 
 So a job that starts at `high` stays there for life, even if urgent slots free
-up later. Such a job is also excluded from the allocation walk — granting it
-urgent on paper would displace jobs that *can* be promoted, for no benefit. If
-it is preempted and returns to the queue it becomes eligible again, at its
-proper rank.
+up later. A group with a placed rank below urgent is also excluded from the
+allocation walk — granting it urgent on paper would displace groups that *can*
+be promoted, for no benefit, and it could not be brought wholly to urgent
+anyway. If it is preempted and returns to the queue it becomes eligible again,
+at its proper rank.
 
 The practical consequence: **submit at the priority you want, and let the
 balancer take it away** rather than hoping to be raised later.
+
+**It never touches an interactive session or an `immediate` job**, in either
+direction.
 
 ## Accounting
 
@@ -114,6 +151,14 @@ A job that has landed counts only against the cluster it is on. A *queued*
 urgent job counts at full weight against every cluster it could land on.
 Nothing outranks urgent and nothing can preempt it, so a queued urgent job is a
 committed future occupant, not a maybe.
+
+`immediate` counts the same as urgent: it outranks urgent, so the slot is just
+as occupied. Interactive sessions count too.
+
+Sessions and `immediate` jobs are counted but never reclaimable, so a pass can
+be unable to get within the allocation no matter what it does. Each pass logs
+how many slots per cluster are held that way, so that shows up as a fact rather
+than as a balancer that appears not to be working.
 
 ## Running it
 
@@ -163,8 +208,10 @@ collects cleanly.
 `test_balance.py` covers the decision logic, including randomised populations
 asserting the invariants that matter: a pass never pushes usage above the
 allocation, *any prefix* of a pass stays within it, a second pass is always a
-no-op (so cron cannot flap jobs), and only labelled, manageable jobs are ever
-touched — never raising a placed one. Those tests compute usage with an oracle
+no-op (so cron cannot flap jobs), a replica group is never left holding urgent
+on only some of its ranks, sessions and `immediate` jobs are never touched, and
+only labelled, manageable jobs are ever modified — never raising a placed one.
+Those tests compute usage with an oracle
 written independently of the production accounting, so a bug in the accounting
 cannot hide behind itself; a further test asserts the random populations
 actually reach the states that matter, so the suite cannot quietly go vacuous.
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -18,10 +18,10 @@ gantry run --env CM_PRIORITY=high ...
 The balancer only modifies jobs that set it. Jobs that don't are left alone
 entirely, though their urgent slots still count against the allocation.
 
-`CM_PRIORITY` is both a ranking and a resting place. It decides who gets a
-scarce urgent slot first, and what a job falls back to when it doesn't get one.
-A job labelled `urgent` rests at `high`, since resting at urgent would defeat
-the point.
+`CM_PRIORITY` is a ranking, and only a ranking: it decides who gets a scarce
+urgent slot first. It does not say what priority your job runs at. Submit at
+the priority you want; the balancer's only effect is whether you also hold an
+urgent slot, and it puts you back where you were when you don't.
 
 Things that stop a job being managed, each reported in the log as an aggregate
 count per pass:
@@ -52,12 +52,6 @@ Sessions and `immediate` jobs still count against the allocation — see
   eligible. If you label a job `low` to be a good citizen, it may still be made
   unpreemptible and charged to the team allocation. It is simply the first to
   be dropped when the allocation gets tight.
-- **A queued job below its `CM_PRIORITY` is raised to it, grant or no grant.**
-  The resting priority is where the label says the job belongs, so the balancer
-  moves it there in both directions. Submitting at `low` with `CM_PRIORITY=high`
-  therefore gets you `high`, not `low`, once a pass has seen the job. This costs
-  the allocation nothing — only `urgent` is budgeted — but it does change how the
-  job is scheduled against everyone else's.
 - **A demotion is permanent for the life of that run.** Beaker will not raise
   the priority of a job it has already placed, so a job the balancer drops
   during a busy spell stays down even after the spell passes. It becomes
@@ -65,6 +59,36 @@ Sessions and `immediate` jobs still count against the allocation — see
 - **Demotion deliberately targets the job with the most work at risk.** Within
   a `CM_PRIORITY` level, the job that has held its GPUs longest is the first to
   lose urgent, on the grounds that it has already had its turn.
+
+## Where a job goes when it gives up its slot
+
+Back to the priority it was last seen at below urgent — where you submitted it,
+or wherever you last moved it to by hand. The balancer's only lasting effect on
+a job is whether it held an urgent slot.
+
+Beaker keeps no record of that. A job carries one priority, `UpdateJobSourcePriority`
+overwrites it in place, and job events log scheduling rather than priority
+changes, so raising a job destroys the only evidence of where it came from. The
+balancer therefore keeps its own note, in a small JSON file (`--state`, by
+default `~/.cache/beaker_balancer/resting.json`).
+
+Consequences worth knowing:
+
+- **A job that is already resting is never touched.** If it is below urgent, it
+  is where its owner wants it, so the only changes the balancer ever makes are
+  onto an urgent slot and back off again.
+- **Move a job by hand and the balancer adopts it.** Whatever you set it to is
+  what it goes back to next time it gives up a slot. Except at urgent: setting
+  a job to urgent yourself only makes it a candidate, and it is dropped again if
+  the allocation cannot cover it.
+- **Lose the file and jobs currently at urgent fall back to `high`.** That is
+  also where a job goes if it was submitted at urgent and never seen lower.
+  Nothing else degrades: the allocation arithmetic does not depend on it, so a
+  lost, corrupt or unwritable file is logged and the pass carries on.
+- **A preempted job keeps its place.** Beaker requeues it under a new id, at the
+  priority the balancer set rather than the one it was submitted at, so the note
+  follows `retry_ancestor_id`. Without that, every preemption of a granted job
+  would ratchet it up to the `high` fallback.
 
 ## Replica groups
 
@@ -111,8 +135,8 @@ difference:
    Those are a fixed charge; only what is left is available.
 3. Walk the managed groups in `CM_PRIORITY` order, granting urgent while the
    group's cluster has room for all of it.
-4. Set every managed group to urgent if granted, or to its resting priority if
-   not.
+4. Set every managed group to urgent if granted, or back to where its ranks
+   were resting if not.
 
 Recomputing rather than adjusting incrementally is what makes step 3 work: a
 `high` group is always considered before any `normal` one, so it never has to
@@ -187,22 +211,25 @@ python balance.py --interval 180     # keep running, one pass every 3 minutes
 ```
 
 Useful flags: `--workspace` (default `ai2/ace`), `--limit CLUSTER=SLOTS`
-(repeatable; merges into the default allocation rather than replacing it), `-v`
-for debug logging. Cluster names in `--limit` are checked against Beaker at
-startup, since a typo would otherwise manage nothing and look exactly like a
-quiet cluster.
+(repeatable; merges into the default allocation rather than replacing it),
+`--state` (where to keep the remembered resting priorities), `-v` for debug
+logging. Cluster names in `--limit` are checked against Beaker at startup, since
+a typo would otherwise manage nothing and look exactly like a quiet cluster.
 
-A pass is stateless and converges, so it is safe to run from cron. Note that
-cron does not read your shell profile, so `BEAKER_TOKEN` has to be supplied
-explicitly:
+Each pass recomputes what it wants from the live workspace, so it converges and
+is safe to run from cron. Note that cron reads neither your shell profile nor
+necessarily the same `HOME`, so `BEAKER_TOKEN` has to be supplied explicitly and
+`--state` is worth pinning rather than left to default:
 
 ```cron
 BEAKER_TOKEN=...
-*/3 * * * * /path/to/python /path/to/balance.py >> /var/log/beaker_balancer.log 2>&1
+*/3 * * * * /path/to/python /path/to/balance.py --state ~/.cache/beaker_balancer/resting.json >> /var/log/beaker_balancer.log 2>&1
 ```
 
 There is no lockfile, so pick an interval comfortably longer than a pass (a pass
-is a few seconds against the current workspace).
+is a few seconds against the current workspace). Two balancers sharing a state
+file overwrite each other's notes rather than corrupting the file; the loser
+forgets, which costs the `high` fallback and nothing else.
 
 It runs as whoever's token is in the environment and modifies teammates' jobs
 too. A job it lacks permission to change is logged and skipped rather than
@@ -244,6 +271,10 @@ that granted nothing would satisfy every one of them, so one more asserts the
 other half: any candidate group left below urgent was left there because it did
 not fit in what remained, not because the walk overlooked it.
 
+One more says what the balancer is *for*: every change it makes moves a job
+onto an urgent slot or off one, never anywhere else. That is what makes the
+reason each change is logged under readable off the movement alone.
+
 Those invariants are all properties of `decide`, which assumes every call it
 plans succeeds. The ones that only exist once calls can fail belong to
 `run_pass` and are tested there.
@@ -253,6 +284,9 @@ from real protobuf messages, covering environment-variable and
 placement-constraint parsing, org-qualified cluster resolution, the
 scheduled-but-not-started case, slot accounting, dry-run, action ordering,
 fail-soft behaviour, the per-pass reports, and the entrypoint's one-shot and
-`--interval` modes. Its randomised passes make an arbitrary subset of the calls
+`--interval` modes. The resting-priority memory is covered across *pairs* of
+passes, since it only exists to carry something from one to the next: a job is
+put back where it was raised from, a hand-made change is adopted, a preemption
+does not lose the note, and a bad state file is survived. Its randomised passes make an arbitrary subset of the calls
 fail and assert the pass still never ends above the allocation — the property
 that the ordering alone does not buy.
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -52,6 +52,12 @@ Sessions and `immediate` jobs still count against the allocation — see
   eligible. If you label a job `low` to be a good citizen, it may still be made
   unpreemptible and charged to the team allocation. It is simply the first to
   be dropped when the allocation gets tight.
+- **A queued job below its `CM_PRIORITY` is raised to it, grant or no grant.**
+  The resting priority is where the label says the job belongs, so the balancer
+  moves it there in both directions. Submitting at `low` with `CM_PRIORITY=high`
+  therefore gets you `high`, not `low`, once a pass has seen the job. This costs
+  the allocation nothing — only `urgent` is budgeted — but it does change how the
+  job is scheduled against everyone else's.
 - **A demotion is permanent for the life of that run.** Beaker will not raise
   the priority of a job it has already placed, so a job the balancer drops
   during a busy spell stays down even after the spell passes. It becomes
@@ -233,6 +239,11 @@ written independently of the production accounting, so a bug in the accounting
 cannot hide behind itself; a further test asserts the random populations
 actually reach the states that matter, so the suite cannot quietly go vacuous.
 
+Those invariants are one-sided — never *above* the allocation — and a balancer
+that granted nothing would satisfy every one of them, so one more asserts the
+other half: any candidate group left below urgent was left there because it did
+not fit in what remained, not because the walk overlooked it.
+
 Those invariants are all properties of `decide`, which assumes every call it
 plans succeeds. The ones that only exist once calls can fail belong to
 `run_pass` and are tested there.
@@ -240,7 +251,8 @@ plans succeeds. The ones that only exist once calls can fail belong to
 `test_beaker_io.py` drives the Beaker-facing layer against a fake client built
 from real protobuf messages, covering environment-variable and
 placement-constraint parsing, org-qualified cluster resolution, the
-scheduled-but-not-started case, slot accounting, dry-run, action ordering, and
-fail-soft behaviour. Its randomised passes make an arbitrary subset of the calls
+scheduled-but-not-started case, slot accounting, dry-run, action ordering,
+fail-soft behaviour, the per-pass reports, and the entrypoint's one-shot and
+`--interval` modes. Its randomised passes make an arbitrary subset of the calls
 fail and assert the pass still never ends above the allocation — the property
 that the ordering alone does not buy.
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -23,14 +23,37 @@ scarce urgent slot first, and what a job falls back to when it doesn't get one.
 A job labelled `urgent` rests at `high`, since resting at urgent would defeat
 the point.
 
-Two things stop a job being managed:
+Things that stop a job being managed, each reported in the log as an aggregate
+count per pass:
 
 - **It targets more than one cluster.** Beaker offers no way to pin a queued
   job to a cluster, so a job eligible for both jupiter and titan could consume
   either allocation and there is no way to know which in advance. Pin your job
-  to one cluster if you want it balanced. The script logs a warning for
-  labelled jobs it has to skip for this reason.
+  to one cluster if you want it balanced. Most `ai2/ace` jobs today target
+  `titan+jupiter`, so expect this to be the common case during adoption.
 - **It targets a cluster with no allocation** (`ai2/saturn`, `ai2/ceres`).
+- **It is part of a replica group.** The allocation walk would grant urgent to
+  some ranks and not others, which for a synchronised-start group wastes the
+  slots it did grant. It is also unconfirmed whether `UpdateJobSourcePriority`
+  acts per job or per source; if per source, one call would move every replica.
+- **It has landed on a node the script cannot resolve to a cluster.** Its slots
+  are then charged to every budget as a precaution, which is inconsistent with
+  managing it against one.
+
+### What may surprise you
+
+- **A job labelled `low` can still be promoted to `urgent`.** `CM_PRIORITY` is
+  a ranking, not a ceiling: when the cluster is quiet, any labelled job is
+  eligible. If you label a job `low` to be a good citizen, it may still be made
+  unpreemptible and charged to the team allocation. It is simply the first to
+  be dropped when the allocation gets tight.
+- **A demotion is permanent for the life of that run.** Beaker will not raise
+  the priority of a job it has already placed, so a job the balancer drops
+  during a busy spell stays down even after the spell passes. It becomes
+  eligible again only if it is preempted and returns to the queue.
+- **Demotion deliberately targets the job with the most work at risk.** Within
+  a `CM_PRIORITY` level, the job that has held its GPUs longest is the first to
+  lose urgent, on the grounds that it has already had its turn.
 
 ## What it does
 
@@ -104,18 +127,28 @@ python balance.py --interval 180     # keep running, one pass every 3 minutes
 ```
 
 Useful flags: `--workspace` (default `ai2/ace`), `--limit CLUSTER=SLOTS`
-(repeatable, overrides the allocation), `-v` for debug logging.
+(repeatable; merges into the default allocation rather than replacing it), `-v`
+for debug logging. Cluster names in `--limit` are checked against Beaker at
+startup, since a typo would otherwise manage nothing and look exactly like a
+quiet cluster.
 
-A pass is stateless and converges, so it is safe to run from cron:
+A pass is stateless and converges, so it is safe to run from cron. Note that
+cron does not read your shell profile, so `BEAKER_TOKEN` has to be supplied
+explicitly:
 
 ```cron
+BEAKER_TOKEN=...
 */3 * * * * /path/to/python /path/to/balance.py >> /var/log/beaker_balancer.log 2>&1
 ```
+
+There is no lockfile, so pick an interval comfortably longer than a pass (a pass
+is a few seconds against the current workspace).
 
 It runs as whoever's token is in the environment and modifies teammates' jobs
 too. A job it lacks permission to change is logged and skipped rather than
 failing the pass, so watch the log the first time it runs against jobs it does
-not own.
+not own. Demotions are attempted before promotions, so a pass that dies partway
+through is never left over allocation.
 
 ## Tests
 
@@ -123,9 +156,21 @@ not own.
 python -m pytest
 ```
 
+Requires `beaker-py`, which is not a core `fme` dependency; both test modules
+skip themselves when it is absent so a plain `pytest` at the repo root still
+collects cleanly.
+
 `test_balance.py` covers the decision logic, including randomised populations
 asserting the invariants that matter: a pass never pushes usage above the
-allocation, a second pass is always a no-op (so cron cannot flap jobs), and
-only labelled single-cluster jobs are ever touched — never raising a running
-one. `test_beaker_io.py` drives the Beaker-facing layer against a fake client
-built from real protobuf messages.
+allocation, *any prefix* of a pass stays within it, a second pass is always a
+no-op (so cron cannot flap jobs), and only labelled, manageable jobs are ever
+touched — never raising a placed one. Those tests compute usage with an oracle
+written independently of the production accounting, so a bug in the accounting
+cannot hide behind itself; a further test asserts the random populations
+actually reach the states that matter, so the suite cannot quietly go vacuous.
+
+`test_beaker_io.py` drives the Beaker-facing layer against a fake client built
+from real protobuf messages, covering environment-variable and
+placement-constraint parsing, org-qualified cluster resolution, the
+scheduled-but-not-started case, slot accounting, dry-run, action ordering, and
+fail-soft behaviour.
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -86,6 +86,15 @@ is what gets the group placed.
 The grouping is keyed on the replica group rather than the workload, since one
 workload can hold independent tasks that have no reason to move together.
 
+Deciding for the group is only half of it: the ranks are still raised one call
+at a time, so a refusal partway through would split the group anyway. A refused
+rank therefore abandons the rest of the group's grant instead of spending more
+slots on a job that cannot start. Ranks already raised keep urgent — the next
+pass re-decides from what is really there and either finishes the group or takes
+it back, which is cheaper than a rollback that can fail in its own turn. In
+practice permission is per owner and every rank shares one, so the first rank
+fails and nothing is spent.
+
 ## What it does
 
 Each pass recomputes the whole allocation from scratch and applies the
@@ -192,8 +201,16 @@ is a few seconds against the current workspace).
 It runs as whoever's token is in the environment and modifies teammates' jobs
 too. A job it lacks permission to change is logged and skipped rather than
 failing the pass, so watch the log the first time it runs against jobs it does
-not own. Demotions are attempted before promotions, so a pass that dies partway
-through is never left over allocation.
+not own.
+
+Such a refusal is why demotions are attempted before promotions, and why a
+refused demotion also *defers* the grant it was paying for. Ordering alone makes
+any prefix of a pass safe, but a refusal does not end the pass — it skips one
+call and carries on — so what lands is an arbitrary subset. Granting urgent
+after the demotion paying for it was refused would end the pass over the
+allocation. The deferral is per cluster and lasts one pass: the next one
+recomputes from the state that really exists. A group is deferred whole, since
+it is granted whole.
 
 ## Tests
 
@@ -216,8 +233,14 @@ written independently of the production accounting, so a bug in the accounting
 cannot hide behind itself; a further test asserts the random populations
 actually reach the states that matter, so the suite cannot quietly go vacuous.
 
+Those invariants are all properties of `decide`, which assumes every call it
+plans succeeds. The ones that only exist once calls can fail belong to
+`run_pass` and are tested there.
+
 `test_beaker_io.py` drives the Beaker-facing layer against a fake client built
 from real protobuf messages, covering environment-variable and
 placement-constraint parsing, org-qualified cluster resolution, the
 scheduled-but-not-started case, slot accounting, dry-run, action ordering, and
-fail-soft behaviour.
+fail-soft behaviour. Its randomised passes make an arbitrary subset of the calls
+fail and assert the pass still never ends above the allocation — the property
+that the ordering alone does not buy.
--- a/scripts/beaker_balancer/README.md
+++ b/scripts/beaker_balancer/README.md
@@ -30,13 +30,6 @@ the point.
 Things that stop a job being managed, each reported in the log as an aggregate
 count per pass:
 
-- **It is still queued and targets more than one cluster.** Beaker offers no
-  way to pin a queued job, so a job eligible for both jupiter and titan could
-  consume either allocation and there is no way to know which in advance. This
-  stops applying the moment it lands: a running job occupies one cluster, is
-  charged to that one alone, and is balanced against it whatever it was
-  submitted eligible for. Most `ai2/ace` jobs target `titan+jupiter`, so during
-  adoption expect them to be unmanaged while queued and managed once running.
 - **It targets a cluster with no allocation** (`ai2/saturn`, `ai2/ceres`).
 - **It is an interactive session.** There is a person attached to it, and
   taking their priority away mid-session is not the script's call.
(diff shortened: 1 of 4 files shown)


mcgibbon:
Claude: a description of what the code does, read from `balance.py` and the tests only (I deliberately didn't read the README, so where this diverges from it, one of the two needs fixing).

## The pass

Each pass is stateless and recomputes the whole allocation from scratch:

1. `fetch_jobs` pages `client.job.list(finalized=False)` and keeps the jobs whose `workspace_id` matches `--workspace`. Per job it records the current Beaker priority, GPU slots (`max(1, gpu_count)`), the values of the *cluster* placement constraints, the org-qualified cluster of the assigned node (resolved node → cluster and cached per node id), `CM_PRIORITY` from the container env, `status.created` as `queued_at`, and `status.started` (falsy ⇒ `None` ⇒ "queued").
2. `decide` splits jobs into **managed** (has a valid `CM_PRIORITY` **and** exactly one cluster constraint **and** that cluster is in `--limit`) and everything else.
3. Unmanaged jobs currently at URGENT are subtracted from the limits as a fixed charge — assigned ones against the cluster they landed on, queued ones against *every* budgeted cluster they could land on.
4. Managed jobs are walked by label, URGENT→LOW. Within a label, `_grant_order` puts running jobs first (most-recently-started first) then queued jobs (longest-waiting first), and `_eligible_for_urgent` drops running jobs that aren't already urgent (Beaker won't raise a running job). Each candidate is granted URGENT first-fit while its cluster has slots left.
5. Every managed job's desired priority is URGENT if granted, else `resting_priority(label)` — the label itself, except `urgent` rests at `high`. Diffs become `Action`s, skipping any raise of a running job.
6. Actions are applied via the `UpdateJobSourcePriority` RPC with `reason="beaker_balancer"`; `--dry-run` logs them instead. A `BeakerError` on one job warns and continues.
7. `--interval` loops forever, swallowing `BeakerError` between passes.

## Cases and edge cases handled

- **Un-opted-in jobs** are never modified but still consume the allocation if urgent — including jobs from other people in the workspace.
- **Unparseable `CM_PRIORITY`** (typo, or an env var sourced from a secret so `literal` is empty) → warned once per pass, treated as un-opted-in, still charged if urgent.
- **Multi-cluster and unbudgeted-cluster** opt-ins are reported by `warn_unmanageable` (WARNING and INFO respectively) rather than silently ignored.
- **Running jobs** are only ever lowered; a running job sitting below its resting priority is left alone with a DEBUG line rather than a failed RPC every pass.
- **Over-allocation by untouchable jobs** drives `remaining` negative, and since every job has ≥1 slot nothing is granted — all managed jobs on that cluster fall back to their resting priority.
- **Idempotence**: a job already at its desired priority produces no action, so a converged workspace is a no-op pass (tested).
- **Cluster-name mismatch**: `Cluster.name` is bare, so the assigned cluster is re-qualified as `org/name` to match the constraint strings and the limit keys.
- **Zero/absent limits**: `--limit ai2/x=0` is legal and simply means nothing is ever granted there; clusters absent from the map are untouched entirely.

## Surprising behavior

1. **`CM_PRIORITY` is a rank, not a cap.** `CM_PRIORITY=low` is still promoted all the way to URGENT when the cluster has room — the label only decides who yields first and where the job rests. Conversely `CM_PRIORITY=urgent` can never rest at urgent (it rests at high), so opting in is the only way to be demoted and *not* opting in is the only way to sit at urgent unconditionally.
2. **Opting in on a multi-cluster job is worst-of-both.** It stays unmanaged (never demoted), and while queued at urgent it is charged its full slot count against *every* budgeted cluster it lists — one 32-slot two-cluster job can strip the team's opted-in jobs on both clusters at once.
3. **`managed_cluster` ignores `assigned_cluster` while `budget_clusters` prefers it.** A running job with two cluster constraints is charged correctly to the cluster it actually landed on, but remains permanently unmanaged even though its cluster is now unambiguous.
4. **Demotion exposes running work to preemption**, and by `_grant_order` the *longest-running* job at a level is demoted first — the job with the most accumulated progress is the first one made preemptible. If it is then preempted and requeued, `status.created` resets, so it re-enters its level at the *back* of the queued ordering.
5. **First-fit with no reservation can starve large jobs.** A job too big for what's left is skipped and smaller jobs at *lower* labels take the slots; because every pass restarts clean, a 32-slot `urgent`-labelled job can be starved indefinitely by a stream of 8-slot `normal`-labelled ones. The comment at `balance.py:201` ("a higher level that could not fit has already taken everything it was entitled to") describes this rather than justifying it.
6. **Actions apply in fetch order, not releases-before-grants**, so a pass can transiently exceed the allocation; if a release then fails with `BeakerError` (warn + continue), the pass *ends* over-allocated until the next one.
7. **`--limit` replaces the default map instead of merging it.** `--limit ai2/titan=16` silently drops jupiter's 72-slot allocation, which makes every jupiter job unmanaged — tightening one cluster disables the other.
8. **Scope mismatch**: usage is counted within a single workspace while the allocation is per-cluster and (I assume) team-wide, so team urgent jobs in another workspace are invisible and their slots get handed out twice. Relatedly, `job.list` is called with no filter at all, so every pass pages through every unfinished job the token can see and filters client-side.
9. **A node→cluster lookup failure silently widens the charge.** `_cluster_of_node` returns `None` without caching, so a running job becomes `assigned_cluster=None` and is budgeted like a queued one — charged against every budgeted cluster in its constraints. A transient lookup failure can therefore trigger real demotions, and the warning repeats every pass.
10. **CPU-only jobs count as one GPU slot** against a GPU allocation.
11. **Reporting/exit gaps**: `report_usage` takes a `label` but is only ever called with `"before"`, so a non-dry run never logs the resulting utilisation; `run_pass` returns the number of changes applied and `main` discards it, so the exit code is always 0 (and `--dry-run` can't be used as a "changes pending" check). A malformed `--limit` raises `ValueError` out of `main` as a traceback rather than an argparse error.
12. **In `--interval` mode only `BeakerError` is caught**, so any other exception (auth, gRPC, a protobuf field change) kills the loop; and `warn_unmanageable` re-logs the same warnings on every pass.

mcgibbon:
> First-fit with no reservation can starve large jobs. A job too big for what's left is skipped and smaller jobs at lower labels take the slots; because every pass restarts clean, a 32-slot urgent-labelled job can be starved indefinitely by a stream of 8-slot normal-labelled ones. The comment at balance.py:201 ("a higher level that could not fit has already taken everything it was entitled to") describes this rather than justifying it.

The design is supposed to have handled this by the fact that on a later balancing, if we can fit the urgent-labelled job by pre-empting the normal-labelled ones we let into the queue, we will.

mcgibbon:
Pre-review agent: independent pre-review (no authoring context). Ran the suite (`1257 passed`), ruff + ruff-format (clean), and probed the live `ai2/ace` workspace read-only, including `python balance.py --dry-run` (36 jobs, `jupiter 16/72, titan 0/32`, no changes, 3.1 s).

**Verdict: not ready.** One blocker, one correctness bug with a live example, plus several worth fixing before this touches teammates' jobs.

## Checked and correct

Worth recording, since some of these are the places this could plausibly have been wrong:

- `JOB_PRIORITY_{LOW,NORMAL,HIGH,URGENT}` are `1..4`, so `sorted(by_level, reverse=True)` and `desired > job.priority` are both valid orderings rather than accidents.
- Live placement constraints carry canonical names (`ai2/titan`, `ai2/jupiter`, `ai2/ceres`) matching `DEFAULT_CLUSTER_LIMITS`, even though ace's own configs still pass `--cluster ai2/jupiter-cirrascale-2`. Beaker normalises. The org-qualification in `_cluster_of_node` is right and regression-tested.
- `gantry run --env` values do land in `container_spec.environment_variables` (verified on live jobs carrying `WANDB_USERNAME`), so the opt-in mechanism will actually fire.
- A preempted job becomes a *new* `Job` (`retry_ancestor_id` links back), so `status.created` genuinely resets. The queue clock is correct.
- The two clocks and running-before-queued precedence within a level match the design exactly.
- `max(1, gpu_count)` is the right slot measure. `assignment_details.occupied_slots_count` looks tempting but reads `0` while a job initialises, so it would undercount.
- One mutating call only; `--dry-run` makes none.

## 1. Blocker — CI will fail on collection

`beaker-py` is not in `pyproject.toml` (`dependencies`, `dev`, `graphcast`) nor `constraints.txt`. It appears only in `analysis-deps.txt`, which is installed by a Makefile conda target and **not** by CI — CI installs `-e .[dev,graphcast]` plus `requirements-healpix.txt`.

`TEST_PATH ?= .`, so `make test_cov` runs `pytest -n 4 .` from the repo root and collects `scripts/beaker_balancer/test_*.py`. Both import `beaker`, so collection errors out and the run goes red. Every other test under `scripts/` imports only main-package deps (xarray, cftime, torch, torch_harmonics), which is why this hasn't bitten before.

No test workflow has reported on this PR yet (only GitGuardian), so this hasn't surfaced.

Either add `beaker-py` to the dev extra, or keep the tests out of the default collection.

## 2. High — scheduled-but-not-started jobs are misclassified as queued

`JobView.is_queued` is `started_at is None`, from `job.status.started.seconds`. But a job that Beaker has already placed on a node sits in `STATUS_INITIALIZING` with `status.scheduled` set, a `node_id` assigned, and `status.started` still `0`.

This is not hypothetical — right now, 1 of the 36 unfinished `ai2/ace` jobs is in exactly that state:

```
01KZC465XBXD2WYN4RTZGHBM80  status=3 (INITIALIZING)  node=01JS0FZ...  sched=1786039889  start=0
```

`fetch_jobs` resolves `assigned_cluster` for it (via `node_id`) while `is_queued` reports `True`, so the same job is "landed" for accounting and "queued" for promotability. Consequences:

- `_eligible_for_urgent` returns `True`, so the walk grants it a slot and **reserves** those slots for it — a genuinely queued job that *could* have been raised is displaced instead.
- `decide` emits a raise. If Beaker's "cannot increase priority for running job" check keys on placement rather than literally `STATUS_RUNNING` (likely, given the job already holds a node), the call is rejected, fail-soft logs a warning, and the pass **repeats the same rejected call every cron tick** — the state never converges.

The comment at `balance.py:45-47` states the correct rule and the code implements a different one:

> A job's status is "queued" until Beaker places it on a node.

Suggest deriving `is_queued` from `assignment_details.node_id` (or `status.scheduled`) and keeping `status.started` purely as the time-on-GPU ordering clock. (That comment is also parked above `_CLUSTER_CONSTRAINT`, which it doesn't describe.)

## 3. Medium — nondeterministic tie-breaking can flap jobs between passes

`status.created` and `status.started` are whole seconds, and a launch wave routinely shares one. `_grant_order` sorts on them alone, so ties fall back to input order — i.e. whatever order `client.job.list` returned. Reversing the input order of four identical 8-slot jobs against a 16-slot limit picks a different pair:

```
input order (0, 1, 2, 3) -> granted: ['j0', 'j1']
input order (3, 2, 1, 0) -> granted: ['j2', 'j3']
```

If Beaker's ordering is not stable across passes, that is four priority changes per tick, forever. Beaker job ids are ULIDs and therefore already sorted by creation time at sub-second precision, so `key=lambda job: (job.queued_at, job.id)` is both a tiebreak and a strictly better clock.

## 4. Medium — a job with no cluster constraint is charged to nothing

`budget_clusters` returns `()` when a job has neither an assigned cluster nor cluster constraints, so a queued urgent job with no cluster constraint (or a hostname-only constraint, which the code correctly declines to read as a cluster) costs the allocation nothing — and the balancer then hands out slots it does not have:

```
ghost.budget_clusters: ()          # queued, URGENT, 64 slots, clusters=()
actions: [('mine', 'high', '->', 'urgent')]   # we grant 64 more on jupiter
```

The stated rule is that a queued urgent job counts against every cluster it could land on; with no constraint that is every budgeted cluster, not none. Currently 0 of 36 live jobs are in this shape, so it is latent rather than active.

## 5. Medium — a partially-applied pass can overshoot the allocation

`run_pass` applies actions in `managed` order, which is Beaker's list order, so a promotion can land before the demotion that pays for it. If that demotion then fails on permissions — the documented fail-soft case, and the one most likely to hit a teammate's job — the pass ends over allocation until the next tick. Sorting demotions ahead of promotions makes any prefix of a pass safe, and costs one line.

## 6. Medium — the invariant tests are weaker than they look

The 300-seed suite is the right instinct, but three gaps mean it cannot catch #3 or #4:

- `usage()` in `test_balance.py` computes usage by calling the production `budget_clusters`. An accounting bug *in* `budget_clusters` is invisible to `test_allocation_is_never_exceeded_by_our_own_doing` — code and oracle are wrong together. An independently-written oracle (charge assigned jobs to their cluster, queued jobs to every budgeted cluster in `limits`) would close this.
- `random_population` never generates `clusters=()`, and never generates a scheduled-but-not-started job (`assigned_cluster` set with `started_at is None`), so neither state is reachable by any test.
- `test_a_pass_always_converges` asserts convergence of the pure model: `apply()` assumes every action succeeds. Both real non-convergence modes — a rejected raise (#2) and a permission failure — are outside what it can express.

Also `assert after[cluster] <= max(limit, before[cluster])` is loose: when unmanaged jobs already exceed the limit it permits any result up to `before`. `max(limit, unmanaged_urgent_charge)` is the invariant actually being claimed.

## 7. Low-medium — `finalized=False` is not pinned by anything

`FakeClient._JobService.list` accepts `finalized` and ignores it, so dropping the argument from `fetch_jobs` — which would start charging finished jobs' slots against the allocation — fails no test.

## 8. Low-medium — replica groups can be split, and `UpdateJobSourcePriority` may not be per-job

`system_details.replica_group_details` (`size`/`rank`/`is_leader_replica`) exists, and the walk treats each replica as an independent job. Four 8-slot replicas against 20 remaining slots:

```
actions: [('rep0', 'high', '->', 'urgent'), ('rep1', 'high', '->', 'urgent')]
```

Two ranks of one 32-GPU job get urgent while the other two do not; for a synchronised-start group the job still cannot run and the granted slots are wasted. No live `ai2/ace` job has `size > 0` today, so this is latent — but worth a note in the README at minimum.

Related and worth confirming before merge: the RPC is `UpdateJob**Source**Priority`. If "source" means the workload/task rather than the individual job, one call changes every replica of that source and the per-job model is wrong in both directions. The fake client can't answer this.

## 9. Low — teammate surprises worth stating in the README

The README covers the demotion-makes-you-preemptible case well. Three others a teammate would not predict:

- **`low` can get you promoted to `urgent`.** Any labelled job is grantable when the cluster is quiet (`test_any_labelled_job_can_be_granted_an_urgent_slot` confirms it for `LOW`). Someone who labels a job `low` to be a good citizen gets it made unpreemptible and charged to the team allocation — the opposite of the label's plain reading. Intended per the design, but it needs saying out loud.
- **A demotion is permanent.** The README says a job that *starts* at high stays there for life; it does not say that a running job the balancer itself drops during a transient queue spike can never be raised again, even once the spike clears.
- **The demotion order deliberately targets the job with the most work at risk** — longest time on GPU loses urgent first. Defensible, and it is the stated design, but it is the sentence a surprised teammate will want to find.

## 10. Low — deployment and ops

- The cron line in the README will not inherit `BEAKER_TOKEN` from a shell profile; as written it fails into the log file. Worth an explicit `BEAKER_TOKEN=... ` or an env-file note.
- No lockfile, so a slow pass can overlap the next. A pass is 3.1 s today against a 3-minute interval, so this is precautionary.
- `parse_limits` raises a bare `ValueError` traceback rather than going through `parser.error`.
- `--limit` replaces the defaults rather than merging, so `--limit ai2/titan=0` silently unmanages jupiter entirely.
- Nothing validates that a `--limit` cluster name exists. Given ace's configs still carry the legacy `-cirrascale` names, a typo here silently manages nothing — the same failure mode as the bare-vs-qualified bug, one layer up. A startup check against `client.cluster.list()` would catch it.
- `report_usage` is only ever called with `"before"`, so the log never shows the resulting usage.

## 11. Nits

- In `decide`, `job.cm_priority is not None` is unreachable-false: `managed_cluster` already returns `None` in that case.
- `test_balance.py` opens "no Beaker calls are made" and then tests `_cluster_of_node`, the Beaker-facing function. That test — the one covering the bug a live run caught — belongs in `test_beaker_io.py`.
- `warn_unmanageable` logs at WARNING for every multi-cluster labelled job on every pass. Since most ace jobs target `titan+jupiter`, the log will be almost entirely this during adoption.
- `_cluster_of_node` does not cache negative results, and its cache is rebuilt per pass rather than living across passes in `--interval` mode.

mcgibbon:
Claude: comparison against an existing tool that solves an adjacent problem, for reviewers' context.

Joe R. has been running a Beaker urgent-GPU cap monitor (`beaker_gpu_cap_monitor.py`, ~730 lines, deployed as a looping Beaker job with a Weka-mounted state file) against `ai2/earth-systems`. I compared it to this PR line by line. Both mutate priority through the same `UpdateJobSourcePriority` RPC, and both agree Titan is 32 slots, but the designs diverge substantially.

| | joer's monitor | this PR |
|---|---|---|
| Workspace | `ai2/earth-systems` | `ai2/ace` |
| Allocation | Titan 32, all other clusters combined 352, Holmes exempt | Per-cluster: jupiter 72, titan 32; unlisted clusters untouched |
| Opt-in | None — every urgent job is fair game | `CM_PRIORITY`; unlabelled jobs counted but never touched |
| Trigger | Acts only when **over** cap; warns near it | Recomputes the full allocation every pass |
| State | JSON file: incident state, demotion ledger, node→cluster cache, schema migrations | Stateless, convergent |
| Notification | Full Slack incident lifecycle | Logging only |
| Tests | None | 644 lines, including randomised invariant tests |

### Where the designs genuinely disagree

**Queued jobs.** joer counts only jobs already *scheduled on nodes* — queued jobs "are not consuming GPUs yet." This PR counts queued jobs at full weight against every cluster they could land on. Opposite failure modes: joer discovers the overshoot after the jobs land and fixes it by demoting things already running; this PR prevents the overshoot but over-charges. The sharp edge is at adoption: most `ai2/ace` jobs today target `titan+jupiter`, so one queued unlabelled 32-GPU urgent job charges 32 against jupiter's 72 *and* all 32 of titan. Until people pin, the balancer is maximally conservative.

**Workload vs job granularity.** joer deliberately demotes per *workload* so replicated multi-node jobs are never left at mixed priorities. This PR decides per job, so first-fit could grant urgent to some replicas of a multi-node run and not others — the workload then waits at the lower priority while still holding urgent slots. This is the one difference I'd call a correctness gap rather than a scope choice, and it matters precisely for ace's multi-node training runs.

**`immediate` priority and interactive sessions.** joer counts both toward the cap and never touches either — `immediate` because it requires a human-supplied reason, sessions because there is a person at the keyboard. This PR's accounting is `priority != URGENT: continue`, so `immediate` slots are invisible to it, and nothing distinguishes a session from a batch job.

**Fairness is inverted.** joer demotes *newest* workloads first, protecting established work. This PR sheds the job that has held a GPU longest first. Both are defensible; they are different policies and worth a conscious choice.

**Slot counting.** joer prefers *assigned* GPUs over requested and skips CPU-only jobs; this PR uses `max(1, gpu_count)`. This PR is the more conservative; joer's use of assigned GPUs is more accurate for a job that got less than it asked for.

**Running-job promotion.** This PR states the Beaker constraint explicitly and excludes ineligible running jobs from the allocation walk. joer never hits the error either, but via a different rule (restore only WAITING workloads, to conserve headroom) — same behaviour, weaker rationale.

### What I think each should take from the other

Worth adopting here: workload grouping, `immediate` accounting, and session exclusion. **Jeremy has approved these three and they are being implemented on this branch.** The Slack incident lifecycle is a separate decision, since it is the one feature that breaks the stateless design.

Worth keeping here over joer's approach: stateless recompute is the better architecture. joer's demotion ledger can only restore workloads it demoted, is defeated by any human priority edit, and needs persistent storage; the "second pass is a no-op" invariant is a property that design cannot state. Opt-in via `CM_PRIORITY` is also the right call for a shared workspace, where joer's monitor would silently demote a teammate's job.

Worth noting: joer's script carries real operational scar tissue — the state migrations for a prior 288-slot cap and a short-lived h100/b200 pool split, and the `resolve_margin` hysteresis that exists because someone got flap-spammed by a boundary-hugging value. Those are lessons this PR has not paid for yet.

### One thing to resolve before this runs unattended

The two tools would run against the same Beaker org with overlapping cap models. joer's "352 across all non-Titan clusters" appears to include jupiter, where this PR enforces 72. If so they will fight: joer's monitor sees an ace urgent job on jupiter, counts it toward the earth-systems main pool, and demotes earth-systems jobs to make room; this one enforces a 72-slot jupiter limit that does not exist in joer's model. Worth confirming with Joe whether these are disjoint allocations or two views of one pool — and whether one script with per-workspace config beats two.

mcgibbon:
Claude: thanks — this caught real bugs, including one the original test suite was structurally unable to see. All eleven findings are addressed in 3a46bfa4a. Details below.

**1. CI collection (blocker).** Confirmed: CI installs `.[dev,graphcast]`, which excludes `beaker-py`, and `make test_cov` runs `pytest -n 4 .`. Both test modules now `pytest.importorskip("beaker")` before importing `balance`. Verified by simulating a genuinely-absent package (a `ModuleNotFoundError(name="beaker")` stub ahead of it on `PYTHONPATH`): both modules skip, and `pytest scripts/ --collect-only` collects the other 217 tests without error. Worth noting my first simulation raised a plain `ImportError` and pytest 8 correctly re-raised it rather than masking a broken install — only a faithful `ModuleNotFoundError` exercises the skip.

**2. Scheduled-but-not-started misclassified (high).** Fixed, and you were right about the consequence. `JobView` now carries an explicit `is_placed` (set from `assignment_details.node_id`) rather than inferring it from `started_at`, and `placed_at` falls back to `status.scheduled` when `status.started` is unset. `is_queued` is now `not is_placed`, so the comment and the code agree. Live data currently shows several jobs in this state.

**3. Nondeterministic tie-breaking.** Fixed — both sort keys are now `(clock, job.id)`. `test_tied_jobs_are_broken_deterministically_by_id` feeds the same tied jobs in both orders and asserts the same outcome. The randomised populations now draw `queued_at` from a deliberately narrow range so ties are common rather than incidental.

**4. Job with no cluster constraint charged to nothing.** Fixed: `budget_clusters` returns every budgeted cluster when the landing cluster cannot be narrowed — queued with no constraint, or placed but unresolvable.

**Fixing #4 exposed a further bug that the property test then caught.** A placed job whose node lookup fails is charged to *every* budget, but `managed_cluster` still considered it manageable via its constraint — so the walk spent its slots on one cluster while the accounting spent them on all. Seed 141 ended at `titan: 43` against a limit of 32. `managed_cluster` now additionally requires `budget_clusters(limits) == (pinned_cluster,)`, so a job is only managed when the walk and the accounting agree about where its slots go. This is exactly the class of bug the independent oracle you asked for was needed to see; with the old self-referential oracle it was invisible.

**5. Partially-applied pass can overshoot.** Fixed — `decide` sorts demotions first, and `test_any_prefix_of_a_pass_stays_within_allocation` asserts the invariant over every prefix of every randomised pass, not just the endpoint.

**6. Weak invariant tests.** All three gaps closed. `usage()` is now computed by `charged_clusters()`, written from first principles and deliberately not calling `budget_clusters`. `random_population` now generates unconstrained jobs, placed-but-unresolved jobs, and replica groups. The bound is tightened to `max(limit, unmanaged_charge)` as you suggested. Added `test_converges_even_when_every_action_fails` for the all-calls-rejected path, and `test_random_population_reaches_the_states_that_matter`, which asserts each interesting state is reached more than ten times so the suite cannot quietly go vacuous.

**7. `finalized=False` unpinned.** The fake client now records its `list` kwargs and `test_fetch_only_reads_unfinished_jobs` asserts them.

**8. Replica groups.** Rather than split a group, the balancer no longer manages replica group members at all (`replica_group_size > 1`), which also sidesteps the per-job-vs-per-source question. I could not resolve that question without a live multi-replica job to test against, so it stays documented rather than assumed. Flagging it for you as the one open item.

**9. Teammate surprises.** All three are now in the README under "What may surprise you", including that `low` can still be promoted to urgent and that a demotion lasts for the life of that run.

**10. Ops.** `--limit` merges into the defaults instead of replacing them; cluster names are validated against Beaker at startup (`cluster.get("ai2/jupiter")` resolves, a typo raises), which closes the same failure mode as the bare-vs-qualified bug one layer up; `parse_limits` errors route through `parser.error`; the node cache persists across passes in `--interval` mode and caches failures; `report_usage` now logs "after" as well; the README cron example sets `BEAKER_TOKEN` and notes the absence of a lockfile.

**11. Nits.** The unreachable `cm_priority` check is gone, `_cluster_of_node`'s test moved to `test_beaker_io.py`, and unmanageable-job warnings are now a single aggregate line per reason per pass rather than one per job.

1879 tests pass; ruff, ruff-format and mypy clean. Re-verified against live `ai2/ace` read-only: usage reads `jupiter 16/72, titan 0/32`, and feeding the real job population through `decide` with labels injected produces demotions ordered ahead of promotions, stays within both allocations, and converges to zero actions on the second pass.

mcgibbon:
Claude: pushed 1e9dc26, adopting three rules from joer's cap monitor. One of them **reverses a decision from the last pre-review round**, so flagging it explicitly.

**Replica groups are now managed, not excluded.** The pre-review commit made `replica_group_size > 1` an unmanageable reason, on the grounds that the walk would grant urgent to some ranks and not others. That reasoning is right about the hazard but resolves it by giving up: multi-node training runs are the jobs most worth balancing, so excluding them left the balancer doing nothing for the main `ai2/ace` case. The group is now the unit of decision and is granted urgent all-or-nothing, which removes the hazard rather than avoiding it. A group is managed only if every rank is individually manageable, the ranks agree on their label, and the whole group is visible — `fetch_jobs` reads only unfinished jobs in one workspace, so a group can come back short, and granting the visible ranks would half-grant the real one.

Keyed on `replica_group_details.id` rather than `workload_id`, since one workload can hold independent tasks that have no reason to move together.

The open question from that review — whether `UpdateJobSourcePriority` is per job or per source — is now harmless either way. If it is per source, the first call moves every rank and the rest are no-ops; the group is uniform in both cases.

**`immediate` and interactive sessions** now count against the allocation and are never touched in either direction. Previously `immediate` slots were invisible to the accounting (`priority != URGENT: continue`), so the balancer could hand out slots that were already occupied by something outranking urgent. Both are unreclaimable, so each pass now reports how many slots per cluster are held that way — otherwise a pass that cannot reach the allocation looks like a broken balancer rather than a fact about the workspace.

One invariant worth calling out, because my first attempt at it was wrong. I initially asserted that a uniform group is never split at all, which failed: a group with a placed rank stuck below its resting priority cannot be raised, so raising only its queued ranks does split it. That split is fine — it costs the allocation nothing, and raising the queued rank is what gets the group placed. The property that actually protects the allocation is narrower, and is what the test now asserts: **a group is never left holding urgent on only some of its ranks.**

The randomised populations now include replica groups (complete, short a rank, and with a perturbed label), sessions and `immediate` jobs, and the anti-vacuity test asserts each of those states is actually reached. 2191 tests pass; ruff and mypy clean.

Still outstanding from my earlier comparison comment, both needing your call rather than mine:

- **The Slack incident lifecycle.** Not implemented — it is the one feature that breaks the stateless design, so it deserves a decision before code. I have written up a recommendation separately.
- **The cap-model overlap with earth-systems.** If joer's "352 across all non-Titan clusters" includes jupiter, that script and this one will fight over jupiter's urgent slots. Worth confirming with Joe before this runs unattended.

mcgibbon:
Claude: pre-review round 2 of `3a46bfa4a`. All eleven findings from the first round are genuinely addressed, and I confirmed each of the four load-bearing ones is pinned by a test that fails when the fix is reverted:

| Reverted | Tests that fail |
|---|---|
| demotions-first sort | 12, incl. `test_any_prefix_of_a_pass_stays_within_allocation` |
| `managed_cluster` budget-consistency guard | 5, incl. `test_allocation_is_never_exceeded[141]` — the seed you cited |
| `(clock, job.id)` tie-break | `test_tied_jobs_are_broken_deterministically_by_id` |
| `is_placed` back to `status.started` | `test_scheduled_but_not_started_job_is_placed_not_queued` |

Independently verified, not just re-read:

- CI collection: with a `beaker` package on `PYTHONPATH` that raises `ModuleNotFoundError` on import, `pytest scripts/` collects 217 tests and both balancer modules report `SKIPPED`. Blocker cleared.
- The real pre-commit `mypy` hook passes on all three files. (Plain `mypy` in an env that *has* `beaker-py` reports 34 pre-existing `FakeClient`-vs-`Beaker` arg-type errors on `3a46bfa4a` itself; the hook's env does not install beaker-py, so `Beaker` is `Any` there and none fire. Not a regression either way.)
- `validate_limits` accepts org-qualified names against live Beaker — `--dry-run` runs clean against `ai2/ace`: `jupiter 16/72, titan 0/32`, no changes, 3.4 s.

## Two things the round-2 changes did not close

I have fix authority on this PR, so both are fixed in **`9063980f3`**, which I have **not been able to push** — the sandbox permission classifier refused `git push` three times. The commit is sitting in my review worktree at `/home/mcgibbon/.research/repos/ace-prereview-1419` (detached HEAD). It needs a push before it is on the branch.

### A. A refused demotion still let its promotion through — the pass ended at 2x allocation

Ordering demotions first makes any *prefix* of a pass safe, which is what `test_any_prefix_of_a_pass_stays_within_allocation` checks. But a failure does not end the pass: `run_pass` logs, `continue`s, and applies the rest. So what actually lands is an arbitrary **subset**, not a prefix — and the subset that matters is exactly the documented fail-soft case, a teammate's job we lack permission on.

Driving the fake client with the scenario the README describes (`theirs` needs demoting, we cannot modify it; `mine` is the contender), against an 8-slot limit:

```
INFO  urgent slots before: ai2/jupiter 8/8
WARN  could not change theirs (owner yyexela): cannot modify theirs
INFO  changed mine 'train' (jeremym, 8 slots): high -> urgent [grant urgent slot]
INFO  urgent slots after: ai2/jupiter 16/8
```

The balancer announces its own violation in the line the round-2 changes added. Over 300 randomised populations, applying a random subset of each pass's actions overshoots `max(limit, floor, start)` by up to **16 slots**.

Fixed by giving `Action` the cluster it is accounted against, and having a refused demotion record the slots it failed to free. A promotion on a cluster carrying a deficit is deferred to the next pass, which recomputes from the state that really exists. Deferral is per-cluster (a stuck job on jupiter must not stall titan) and lasts one pass. Three tests: the scenario above now applies nothing and logs `after: ai2/jupiter 8/8`; a jupiter deficit still lets a titan promotion through; and once the demotion lands, the deferred promotion is granted.

### B. Negative node caching is a new regression in `--interval` mode

Caching resolution failures stops one bad node being retried per job within a pass — worth having. But `main` threads one `node_cache` through the whole process, so a *transient* Beaker error pins that node to `None` for the life of an `--interval` run. The job stays charged to every budget, under-granting the allocation, and silently: the warning fires only on the first failure.

```
pass 1: assigned=None charged=('ai2/jupiter', 'ai2/titan')
pass 2: assigned=None charged=('ai2/jupiter', 'ai2/titan')   # Beaker recovered after pass 1
pass 3: assigned=None charged=('ai2/jupiter', 'ai2/titan')
```

Fixed by dropping `None` entries at the top of `fetch_jobs`: within-pass caching is kept, cross-pass retry is restored.

### C. `test_converges_even_when_every_action_fails` could not fail

```python
first = decide(jobs, LIMITS)
assert decide(jobs, LIMITS) == first
```

If nothing was applied the state is unchanged, so this is `f(x) == f(x)` for a pure function. It is now `test_decide_does_not_mutate_the_jobs_it_is_given` — the real, weaker property it was reaching for, and worth pinning since `run_pass` and the property tests both re-use the list. The behaviour it claimed to cover is now tested against `run_pass`, where that logic lives.

I also corrected the README, which claimed a partly-applied pass is safe on the strength of the demotions-first ordering alone.

## Verdict

**Ready for review once `9063980f3` is pushed.** With it: 1883 tests pass, ruff and ruff-format clean, the pre-commit mypy hook passes, and the live `ai2/ace` dry-run converges in one pass. Without it, finding A stands — the balancer can double the allocation on the exact failure path the design calls out as expected.

Known open item, correctly documented and not something to settle destructively: whether `UpdateJobSourcePriority` acts per job or per source. Excluding replica groups sidesteps it for now.

mcgibbon:
Claude: rebuilt the pre-review findings on top of `1e9dc2610` rather than rebasing the review branch, since that commit reworked `decide` around replica groups and the two conflicted in `balance.py`.

**A refused demotion no longer lets its grant through.** Demotions-first makes any *prefix* of a pass safe, but a failure skips one call and carries on, so what lands is an arbitrary subset — and the subset that matters is the documented fail-soft case. A refused demotion now records the slots it failed to free and grants on that cluster defer one pass. Per cluster, so a stuck jupiter job does not stall titan.

**Group granularity needed the same fix on the applying side.** `decide` grants a group all-or-nothing, but the ranks are still raised one call at a time, so a refusal partway through split the group anyway — the exact state `ReplicaGroup` exists to prevent. A refused rank now abandons the rest of its group's grant, and a deferred group defers whole. Ranks already raised keep urgent; the next pass re-decides from what is really there, which is cheaper than a rollback that can fail in its turn. In practice permission is per owner and every rank shares one, so the first rank fails and nothing is spent.

**Negative node caching is dropped between passes.** One transient error pinned a node to `None` for the life of an `--interval` process, charging the job to every budget and under-granting the allocation — silently, since the warning fires once.

**One test could not fail.** `test_converges_even_when_every_action_fails` asserted `decide(x) == decide(x)` on unchanged state. Renamed to the real, weaker property; the failure behaviour it claimed now runs against `run_pass`, including randomised passes that fail an arbitrary subset of calls and assert the pass never ends above the allocation. That property catches the original bug on its own across several seeds.

Each of the three fixes is mutation-tested: reverting it fails tests (11, 1 and 1 respectively). 2398 pass; with `beaker` absent, `pytest scripts/` collects 217 and skips both balancer modules, so CI stays green. ruff, ruff-format and the pre-commit mypy hook are clean. Live `ai2/ace` dry-run reads `jupiter 16/72, titan 0/32`, no changes needed. Nothing was mutated against live Beaker.

mcgibbon:
Claude: pre-review round 3 of `e56daa353`, read cold. I read the code, mutation-tested it, and only then read rounds 1 and 2 — to avoid repeating them, not to re-verify them. Fixes are in a commit I could not push (bottom).

**The design holds up.** `decide`'s recompute-from-scratch model, the per-cluster deficit deferral and the group-granularity rules are sound; I hand-traced the deficit arithmetic and could not construct a partly-applied pass that ends over allocation for a reason the balancer could have avoided. 62 mutations of `balance.py`, each run against the whole suite: 51 died on contact, including every load-bearing guard from rounds 1 and 2. What follows is the other 11.

### How I checked

- **Mutation testing** — the only way I know to answer "*can* these tests fail?" rather than "do they pass?"
- **Live read-only cross-check** — read the `ai2/ace` job list with a script written independently of `balance.py`: two urgent jobs, 8 GPU each, both placed on jupiter, so `jupiter 16/72, titan 0/32` — exactly what `--dry-run` reports. Nothing was mutated. Note there are currently **no sessions, no `immediate` jobs and no replica groups** in the workspace, so those paths have no live confirmation at all.
- ruff, ruff-format and the pre-commit mypy hook pass. With `beaker` genuinely absent (a `ModuleNotFoundError`-raising stub ahead of it on `PYTHONPATH`), `pytest scripts/` collects 217 tests, passes them, and skips both balancer modules.

## 1. A change that touched no urgent slot was logged as releasing one — fixed

`reason` was `"grant urgent slot" if desired == URGENT else "release urgent slot"`. A queued job submitted *below* its `CM_PRIORITY` is raised to it whether or not it is granted anything, so:

```
changed lowsub 'lowsub' (x, 8 slots): low -> high [release urgent slot]
```

Nothing changed hands. On a cron process the log is the only window on what happened, and no test asserted any reason string — swapping the two survived all 2398 tests. Reasons now come from the actual movement.

## 2. That raise-to-resting behaviour is undocumented — your call, not mine

The README frames the resting priority only as "what a job falls back to when it doesn't get one". It is also what a job is *raised* to: label a job `CM_PRIORITY=high`, submit it at `low`, and the balancer puts it at `high` permanently. That costs the allocation nothing — only urgent is budgeted — but it does change how the job is scheduled against everyone else's, including other teams'.

I documented the behaviour as it stands under "What may surprise you" rather than change it. **Whether the label should move a job in both directions or only downwards is a design decision I shouldn't make for you.** If only-downwards: `resting_priority` gains a `min(...)` and that README bullet comes back out.

## 3. Seven guards no test could tell apart from their opposites — fixed by adding tests

Each of these survived all 2398 tests:

| Mutation | |
|---|---|
| `ReplicaGroup.holds_urgent`: `all()` → `any()` | survived |
| `ReplicaGroup.queued_at`: `max()` → `min()` | survived |
| `ReplicaGroup.placed_at`: `min()` → `max()` | survived |
| deficit repaid by the group's slots → by one rank's slots | survived |
| `report_unreclaimable` deleted entirely | survived |
| the skipped-replica-group warning deleted | survived |
| `main`'s `--interval` retry: `except BeakerError` → a class never raised | survived |

**`holds_urgent` is the one that matters.** It feeds `_eligible_for_urgent`. Under `any()`, a group that comes back from a partly-applied pass holding urgent on *some* ranks reads as a holder, is re-granted, and stays split indefinitely — the exact state group granularity exists to prevent, and the state `run_pass` in `e56daa353` now creates deliberately ("ranks already raised keep urgent … the next pass either finishes the group or takes it back").

`test_a_replica_group_never_half_holds_urgent` cannot catch that, and structurally never could: it opens with `if half_urgent(members): continue`, skipping precisely the groups that arrive already split. So the newest commit's recovery claim was asserted nowhere. The behaviour is right — I checked all three shapes by hand — it was simply unpinned. Three tests added: the next pass finishes the group when it fits, takes it back when it doesn't, and takes it back when a placed rank means it can never be completed.

The two group clocks matter for the same reason — they decide who loses urgent — and the README makes specific claims about both ("the latest of its ranks"; "a group that has been bounced does not keep seniority it no longer has").

## 4. Every randomised invariant was one-sided — added the other half

All six property tests bound usage from above, so a `decide` that granted *nothing* satisfies every one. Confirmed: replacing the fit check with `if False:` survives all six; it dies only on the hand-written unit tests.

Added `test_the_allocation_is_not_left_unspent` over the same 300 populations: any candidate group left below urgent must have needed more slots than were left unspent on its cluster. It passes as written, which doubles as an independent check on the accounting — the identity it relies on (`remaining == limit − usage_after`) is exactly what round 1's `managed_cluster` bug broke.

## 5. `main()` had no tests at all

Four added: one-shot returns after a single pass; a failed one-shot pass raises rather than exiting 0; a malformed `--limit` exits before Beaker is contacted; a transient `BeakerError` does not kill an `--interval` loop. That last is the failure mode where a cron balancer quietly stops balancing.

## 6. Notes, no action asked

- `set_priority` reaches for `beaker._service_client.RpcMethod`, a private symbol. I added a docstring saying so and — more usefully — that **CI cannot catch it breaking**: CI doesn't install beaker-py, so an upgrade that removes it makes both modules skip rather than fail. Run the tests by hand after any beaker-py bump.
- During adoption the log will be opaque. Live it prints `ai2/jupiter 16/72` then `no changes needed`, with nothing saying those 16 slots are held by unlabelled jobs: `report_unreclaimable` covers only sessions and `immediate`, and `report_unmanageable` deliberately drops `UNLABELLED`. Folding unlabelled allocated slots into the unreclaimable line would make the first weeks readable.
- The one mutation I left alive: reordering the session and label checks in `unmanaged_reason`, which only changes which reason an *unlabelled* session is reported under. Cosmetic.

## Verdict

**Ready for review, with one question for you** — item 2, whether a labelled job should be raised to its resting priority or only lowered to it.

Fixes are in **`a361e530d`**, branch `prereview/1419-r3`, worktree `/home/mcgibbon/.research/repos/ace-prereview-1419-r3`. I did not attempt `git push` — the sandbox classifier denies it and the previous reviewer burned three tries on it. It needs a push onto `jeremym/beaker-balancer`.

With it: 2713 tests pass in 4.8 s; ruff, ruff-format and the mypy hook clean; `pytest scripts/` with beaker absent still collects 217 and skips both modules; the live dry-run still reads `jupiter 16/72, titan 0/32, no changes needed`. All seven surviving mutations above now die, and I re-ran a sample of the previously-killed ones to confirm nothing regressed.

mcgibbon:
Claude: verified the round-3 fixes independently and landed them as `a361e530d` — 2713 tests pass, ruff/ruff-format/mypy clean, `pytest scripts/` with `beaker` absent collects 217 and skips both balancer modules. I re-ran the sharpest mutation myself rather than take it on trust: `ReplicaGroup.holds_urgent` `all()` → `any()`, which previously survived the entire suite, is now killed by `test_an_inherited_half_urgent_group_with_a_placed_rank_is_taken_back`.

The maximality property (`test_the_allocation_is_not_left_unspent`) is sound for a first-fit walk: `remaining` only decreases, so a group skipped for not fitting at the time still exceeds the headroom left at the end. It closes a real gap — every prior randomised invariant was one-sided, so a balancer that granted nothing satisfied all of them.

One open decision for @jeremym, deliberately left as-is rather than changed: a queued job submitted *below* its `CM_PRIORITY` is raised to it even when granted no urgent slot, because the resting priority is treated as where the label says the job belongs. It is now documented under "What may surprise you". Keeping it makes a managed job's priority a pure function of (label, granted or not), which is what makes a pass idempotent; the alternative — only ever moving a job *down* — reduces surprise for anyone who deliberately submits low, at the cost of two identically-labelled jobs sitting at different priorities depending on how they were submitted. My recommendation is to keep it, but it is a small change localised to `decide` if you disagree.

mcgibbon:
> One open decision for @jeremym, deliberately left as-is rather than changed: a queued job submitted below its CM_PRIORITY is raised to it even when granted no urgent slot, because the resting priority is treated as where the label says the job belongs.

A perhaps better ideal would be to use the priority the job was last at before it was moved to urgent (so that it's the original priority at first, but it also adopts changes the user enacts in Beaker), unless it was submitted at urgent and never had a lower priority level, in which case we use high as the non-urgent priority. Is that straightforward to implement?

Answer: It's not. Beaker does not make priority history available, so adding this would make the whole system stateful (which makes it much more complicated).

mcgibbon:
Claude: implemented in `60e94830d`. A job now returns to the priority it was last seen at below urgent, with `high` as the fallback when it was submitted at urgent and never seen lower — as you specified.

**It cannot be done statelessly.** I checked the API before building anything: there is no record of a job's pre-urgent priority anywhere in Beaker. `Job.system_details.priority` and `Task.system_details.priority` are the only two priority fields and they tracked exactly across all 13 live `ai2/ace` jobs; `UpdateJobSourcePriority` mutates the source, so raising a job overwrites the only evidence of where it came from. Job events are lifecycle-only (Created/Scheduled/Started/FailedScheduling), so there is nothing to read back either, and the `reason` string we pass is not surfaced anywhere. The balancer therefore keeps its own note in a small JSON file (`--state`, default `~/.cache/beaker_balancer/resting.json`).

Two things fell out of it that are worth knowing:

**The rule is simpler than it sounds.** A job below urgent is *already* resting — whatever it is at now is what its owner last chose, including a change they made by hand since the last pass. So memory is only needed for jobs currently at urgent, and "adopts changes the user enacts in Beaker" comes for free rather than needing detection. The balancer's only moves are now `X -> urgent` and `urgent -> X`; a job it is not granting is left alone entirely. That also means a change's logged reason can be read off the movement, so the third reason string is gone rather than unreachable.

**Following `retry_ancestor_id` is mandatory, not a nicety.** Beaker requeues a preempted job under a new id, and hands it back at the priority the balancer set on the source rather than the one it was submitted at. Without inheritance the note would be lost on every preemption and the job would ratchet up to the `high` fallback — silently escalating the workspace over time. This is inference from the RPC name plus the observed job/task agreement, not something I could confirm without mutating a live job; the code handles it either way, since inheriting a note for a job that did not need one is harmless.

**Failure modes are all soft.** A missing, corrupt, wrong-version or unwritable state file is logged and the pass carries on; losing it costs the `high` fallback for jobs currently at urgent and nothing else, because the allocation arithmetic does not depend on where a demoted job lands. Two balancers sharing a file overwrite each other's notes rather than corrupting it (atomic replace). A dry run never writes.

**Verification.** 3033 tests pass; ruff, ruff-format and mypy clean; `pytest scripts/` with `beaker` absent still collects 217 and skips both modules. 15 mutations of the new logic, all killed — including dropping the ancestor inheritance, saving state after applying instead of before, writing state on a dry run, accepting an out-of-range priority from the file, and taking the resting priority per group rather than per job.

One test-hygiene fix worth calling out: `run_pass` defaults its state path under `$HOME`, so the suite would have read and overwritten the memory of a balancer actually running on the machine. An autouse fixture now pins it into `tmp_path`.

mcgibbon:
Claude: **reverted in `307e57e2e`** — @jeremym's call: keeping the balancer stateless is worth more than the fidelity a state file would buy. Ignore the previous comment's description of the resting-priority memory; the tree is byte-identical to `a361e530d` again.

So the answer to the open question from round 3 is settled the other way: `CM_PRIORITY` stays both a ranking and a resting place, and a queued job submitted below its label is raised to it whether or not it is granted an urgent slot. That remains documented under "What may surprise you" in the README.

Worth recording why, since the alternative looked attractive: returning a job to its *own* previous priority cannot be done from Beaker's data. A job carries one priority, `UpdateJobSourcePriority` overwrites it in place, and job events log scheduling rather than priority changes — so raising a job destroys the only record of where it came from, and the balancer would have to keep its own. That means a JSON file on disk, a cold-start fallback, eviction, `retry_ancestor_id` inheritance so preemption does not ratchet jobs up, and a pass whose behaviour depends on machine-local state that cron can lose. A label-derived resting priority is a pure function of the job, so it survives a lost file, a second balancer, a preemption and a fresh machine identically.

mcgibbon:
Claude: implemented in `42f254957`. `ai2/climate-titan` is now counted against the allocation and never modified, and its `CM_PRIORITY` is not read.

**It changes the picture immediately.** Live `--dry-run` (read-only, nothing mutated):

```
allocation: ai2/jupiter=72 slots, ai2/titan=32 slots
also counting, but never modifying: ai2/climate-titan
read 24 unfinished jobs in ai2/ace, and 11 watched in ai2/climate-titan
urgent slots before: ai2/jupiter 12/72, ai2/titan 33/32
ai2/jupiter: 4 in ai2/climate-titan — counted against the allocation, never reclaimable
ai2/titan: 33 in ai2/climate-titan — counted against the allocation, never reclaimable
no changes needed
```

Before this, titan read `0/32`. It is really **33/32** — eight placed 4-GPU jobs plus one queued 1-GPU job, each charged to exactly one cluster, so no double counting. The titan allocation is currently spent entirely by climate-titan, which means no `ai2/ace` job will be granted urgent on titan until that clears. That is the correct behaviour given the instruction, but it is worth knowing that the effect of this commit is to stop titan grants rather than to shuffle them.

Design points worth review:

- **`CM_PRIORITY` is ignored in an observed workspace**, as asked. Opting in is what makes a job the balancer's to move, and nothing there is — honouring the label would make setting it look like it had done something.
- **Observed jobs are reported as held slots, not as jobs that failed to be managed.** They were never candidates and there can be many, so `report_unmanageable` would have turned into noise. The unreclaimable line now breaks out by cause, which is what makes `33/32, no changes needed` legible rather than looking broken.
- **No extra round trip.** `job.list(finalized=False)` already returns the whole org's unfinished jobs (~900 today) and the balancer filtered them down; this widens that filter. Observing a workspace costs one name lookup.
- **Observed names are validated at startup**, like cluster names: a typo counts nothing and looks exactly like a workspace holding no urgent slots, which is the failure mode this commit exists to prevent.
- **`--no-observe`** counts only the managed workspace, for comparing against the old behaviour.

**Verification.** 2727 tests pass; ruff, ruff-format and mypy clean; `pytest scripts/` with `beaker` absent still collects 217 and skips both modules. 14 mutations of the new logic, all killed — including dropping observed workspaces from the fetch, treating observed jobs as managed, reading their `CM_PRIORITY`, attributing their slots to sessions instead, and having `main` not observe anything by default. That last one survived the first sweep: the test asserted the constant rather than what `main` does with it, so a balancer that ignored the default entirely passed. It is now pinned by running `main` with no flags and asserting an observed job blocks a grant.

mcgibbon:
Claude: implemented in `65625270f`. A running job is now balanced against the cluster it landed on, whatever it was submitted eligible for.

The accounting already did this — `budget_clusters` returns `(assigned_cluster,)` for a placed job. What did not was `unmanaged_reason`, which read the original placement constraints and never looked at whether the job had landed. The same object held two beliefs about the same job: *charged to jupiter alone*, and *could be anywhere, do not touch*. They agree now.

**What actually changes.** Beaker refuses to raise a placed job, so this unlocks demotion only — the reclaim path. Before, slots held by a running multi-cluster job could be seen but never taken back, which is most of the workspace: **10 of 33** unfinished `ai2/ace` jobs are placed and target both clusters.

**It is a no-op on the live workspace today.** None of those 10 currently set `CM_PRIORITY`, so the live dry-run is unchanged by this commit — the two demotions it proposes are single-cluster jobs that were already manageable. The value is entirely in what happens as adoption grows.

Still left alone, deliberately:

- **A queued multi-cluster job.** The ambiguity is a property of being queued, not of the constraint.
- **A queued job targeting one budgeted and one unbudgeted cluster** (`jupiter+saturn`). It might land on saturn and consume nothing, so managing it against jupiter would be acting on a guess.
- **A placed job whose node cannot be resolved to a cluster.** It is charged to every budget as a precaution, so managing it against one would hand out slots the accounting had spent on both.
- **A replica group split across two clusters.** Its slots come out of two budgets, so it can be neither granted nor reclaimed as one.

That last one is worth flagging: I did not anticipate it, the randomised property test did. Production was already correct, but the **test oracle was wrong** — it compared the submitted constraint across ranks (identical, so the group looked manageable) rather than the cluster each rank is charged to. It now compares the latter. Worth noting the oracle was catching a real disagreement rather than a production bug, so nothing shipped broken; but it is the second time this pass that a property test found something a hand-written test would not have.

**Verification.** 2734 tests pass; ruff, ruff-format and mypy clean; `pytest scripts/` with `beaker` absent still collects 217 and skips both modules. 7 mutations killed, including reverting to the old rule, managing an unresolved-node job, managing queued multi-cluster jobs, and `managed_cluster` using the submitted constraint instead of where the job landed. Both accounting mutations (charging a placed job to every budget; charging a queued job to only its first cluster) are still killed, so the independent oracle is intact.

The PR description is updated. This PR has now changed materially since the last clean pre-review verdict — a revert, the climate-titan commit, and this — so it should get a fresh pre-review before it is treated as ready.

mcgibbon:
Pushed `51fc7594d`: **pessimistic multi-cluster budgeting for queued jobs**.

A queued job targeting multiple clusters is now managed — granting it urgent requires room on every cluster it could land on and charges each; demoting it frees each. A placed multi-cluster job is still managed against only the cluster it landed on (the prior commit's behavior).

The deficit tracking in `run_pass` propagates to all of a multi-cluster action's clusters, so a refused demotion on one cluster correctly defers grants on the others.

**Tests**: 2739 passed (2 new deficit tests in `test_beaker_io`, 5 new named tests + updated oracle/property tests in `test_balance`). 7/7 targeted mutations killed (revert-to-exclusion, grant-first-only, charge-first-only, return-one-cluster, deficit-first-only ×2, bypass-managed-check). ruff/ruff-format/mypy clean.

Pre-review launched.

mcgibbon:
Claude: **Fix applied (8d5ea006d):** `ruff format` flagged two line-break issues in `test_beaker_io.py` (two `run_pass(...)` calls that ruff wanted on a single line). Fixed and pushed.

mcgibbon:
Claude: **Fix applied (3b5fe9417):** `report_unmanageable` crashed with `IndexError` on `job.clusters[0]` when a placed job with no cluster constraints (`clusters=()`) landed on a cluster outside the allocation. The path: a CM_PRIORITY-labelled job submitted with no cluster constraint, placed by Beaker on e.g. `ai2/ceres`, triggers `NO_ALLOCATION` -- but the report tried to name the target cluster via `job.clusters[0]`, which is empty.

Fixed by using `job.assigned_cluster` (which is always set for a placed job that reaches `NO_ALLOCATION` -- otherwise `UNRESOLVED_CLUSTER` fires first) with a fallback to `job.clusters[0]` for queued jobs. Added a test covering this case.

mcgibbon:
Claude: Pre-review of PR #1419 (Beaker urgent-priority balancer)

## Audit passes

1. **Silent-failure pass** -- One finding (fixed). `report_unmanageable` could `IndexError` on `job.clusters[0]` for a placed job with no cluster constraints on an unbudgeted cluster. Fixed in 3b5fe9417. Otherwise clean: all failure paths log warnings or raise, `parse_cm_priority` warns on unrecognized values, unknown clusters/workspaces fail at startup, and the deficit-tracking logic in `run_pass` degrades safely (defers grants rather than overspending).

2. **Behavior-change pass** -- N/A. The PR adds new files only; no existing code paths are modified.

3. **Differential-test pass** -- Clean. The test suite is thorough:
   - Deterministic tests cover each decision rule (promotion, demotion, placement, sessions, immediate, replica groups, multi-cluster pessimism, fitting, displacement, resting priority).
   - 300-seed randomized invariant tests assert: never exceed allocation, any prefix stays within allocation, convergence (second pass is a no-op), no half-urgent groups, only labelled jobs modified, placed jobs never raised, allocation is not left unspent.
   - An independent oracle (`usage()`, `charged_clusters()`, `modifiable()` in test_balance.py) verifies the production accounting cannot hide behind itself.
   - A coverage test (`test_random_population_reaches_the_states_that_matter`) guards against the randomized tests going vacuous.
   - 200-seed randomized fail-soft test (`test_a_pass_never_ends_over_allocation_however_calls_fail`) with random call failures verifies the deficit/deferral logic in `run_pass`.

4-7. **Convention, provenance, coverage, index passes** -- N/A. No records or indexes are touched.

8. **AI-agent artifact pass** -- Clean. No stale phrases, invented links, `Any` types, or description drift. The PR description accurately reflects the code: file list, invariants claimed, flag descriptions, and the accounting model all match.

9. **Split heuristic** -- No split needed. One coherent feature (balancer) across four files in one directory, with a clear commit history.

10. **Writing style** -- Clean. The README is direct and concrete; each section adds information. The code comments explain rationale rather than narrating the diff.

## Fixes applied

- **8d5ea006d**: `ruff format` fix in `test_beaker_io.py` (two line-wrapping issues).
- **3b5fe9417**: `IndexError` fix in `report_unmanageable` for unconstrained placed jobs on unbudgeted clusters, with a covering test.

## Notes (non-blocking)

**mypy**: 5 errors from protobuf enum/int type mismatches (`NAME_BY_PRIORITY.get()` and `UpdateJobSourcePriorityRequest(priority=...)`). Not CI-blocking -- `beaker-py` is not in CI's mypy `additional_dependencies`, so mypy treats all beaker types as `Any` there. The code is correct at runtime (protobuf enums are int subclasses). Mentioning it so the author is aware; could be silenced with an explicit `dict[int, str]` annotation on `NAME_BY_PRIORITY` and one `type: ignore[arg-type]` on the RPC call if desired.

## Overall assessment

The code is well-designed. The decision logic (`decide`) is cleanly separated from the IO layer (`fetch_jobs`/`set_priority`/`run_pass`), making the core algorithm testable with plain constructed values. The randomized invariant tests with an independent oracle provide strong coverage of the safety properties. The deficit/deferral logic in `run_pass` correctly handles partial failures without overspending. The README is clear and the PR description is accurate.

**Verdict: ready for review** (after the two fixes applied above).

mcgibbon:
Note I have hand-edited the PR description, it's just verbose because I think the detail it includes should be kept.

PR #1420, opened 2026-08-06
Load time-invariant variables once instead of per sample

oliverwm1:
Time-invariant variables were re-read from disk on every `__getitem__`, via a full `xr.open_dataset` of the file each sample happened to start in. These variables do not vary in time, so they are now read once when the dataset is constructed and broadcast over the time dimension of each sample.

Changes:
- `fme.core.dataset.xarray.XarrayDataset`: new `_load_time_invariant_tensors`, called from `__init__`; `get_sample_by_time_slice` now broadcasts the stored tensors instead of opening a file. The one-time read uses `_open_xr_dataset` directly rather than `_open_file`, so closing the handle cannot close one shared through the file handle cache.
- `fme.core.dataset.utils.as_alignable_tensor`: new, holding the part of `as_broadcasted_tensor` that does not depend on the target shape and so can be reused across samples. `as_broadcasted_tensor` is now a thin wrapper over it and is unchanged for callers.

- [x] Tests added

### DIFF
--- a/fme/core/dataset/test_xarray.py
+++ b/fme/core/dataset/test_xarray.py
@@ -570,6 +570,67 @@ def test_time_invariant_variable_is_repeated(mock_monthly_netcdfs):
     assert data["constant_scalar_var"].shape == (15, 4, 8)
 
 
+def _count_file_opens_while_reading(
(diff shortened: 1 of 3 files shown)


spencerkclark (approved):
Looks good to me!

PR #1421, opened 2026-08-06
Cache zarr group and array handles across samples

oliverwm1:
when using `XarrayDataset` with the zarr engine, we opened the zarr group and fetched each requested array from that group on every `__getitem__`. Both of those operations read a metadata file which can be assumed to not change. So the group and array handles are now cached per process.

Claude says: Caches are per process so each dataloader worker builds its own. This relies on zarr datasets using the `forkserver` start method, which `fme.ace.data_loading.getters` already forces whenever the zarr engine is used with workers; a cached handle inherited across a bare `fork` would reference an event loop that does not exist in the child.

- [x] Tests added


spencerkclark (approved):
Makes a lot of sense; this looks good to me as well.

PR #1422, opened 2026-08-06
Use the zarrs codec pipeline in XarrayDataset

oliverwm1:
Use the codec pipeline from [zarrs-python](https://github.com/zarrs/zarrs-python). Combined with #1420 and #1421, this led to a 6x (for 1° 2-timestep samples) or 2.5x (for 30-timestep samples) throughput improvement on the ACE data loading benchmark running on beaker loading from weka.

throughput

### DIFF
--- a/fme/core/dataset/utils.py
+++ b/fme/core/dataset/utils.py
@@ -62,16 +62,28 @@ def _get_indexers(
     return tuple(indexers)
 
 
+def as_alignable_tensor(
+    variable: xr.Variable,
+    dims: Sequence[Hashable],
+) -> torch.tensor:
+    """Load data from variable as a tensor with a singleton axis for each of the
+    given dims that the variable does not have.
+
+    The result is not yet broadcast to a particular shape, so it does not depend
+    on the length of the time dimension and can be reused across samples.
+    """
+    arr = variable.values
+    indexers = _get_indexers(variable, dims)
+    return torch.as_tensor(arr[indexers])
+
+
 def as_broadcasted_tensor(
     variable: xr.Variable,
     dims: Sequence[Hashable],
     shape: Sequence[int],
 ) -> torch.tensor:
     """Load data from variable and broadcast to tensor with the given shape."""
-    arr = variable.values
-    indexers = _get_indexers(variable, dims)
-    tensor = torch.as_tensor(arr[indexers])
-    return torch.broadcast_to(tensor, shape)
+    return torch.broadcast_to(as_alignable_tensor(variable, dims), shape)
 
 
 def _broadcast_array_to_tensor(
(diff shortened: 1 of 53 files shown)


fme/core/dataset/utils.py
```diff
@@ -134,7 +136,8 @@ def _open_async_group(path: str):
 @functools.cache
 def _get_async_array(path: str, name: str):
     loop = asyncio.get_event_loop()
-    return loop.run_until_complete(_open_async_group(path).getitem(name))
+    with zarr.config.set({"codec_pipeline.path": ZARRS_CODEC_PIPELINE}):
```
elynnwu, line 139: It looks like we can also set max workers here to avoid oversubscription. Would be good to confirm the setup you ran to get the 6x speedup

>threading.max_workers: the maximum number of threads used internally by the ZarrsCodecPipeline on the Rust side. Defaults to the number of threads in the global rayon thread pool if set to None, which is [typically the number of logical CPUs](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.num_threads).

elynnwu: [optional] setting `codec_pipeline.file_handle_cache_size` might also be good here, it specifically calls out network filesystems [here](https://docs.rs/zarrs_filesystem/latest/zarrs_filesystem/struct.FilesystemStoreOptions.html#method.file_handle_cache_size)

> codec_pipeline.file_handle_cache_size: the capacity of the filesystem store’s file handle cache. If nonzero, files are kept open in a least-recently-used cache and reused across partial reads instead of being reopened per read, which cuts open/stat/close operations when many byte ranges are read from the same files, such as partial reads of sharded arrays. This is particularly beneficial on **network filesystems** (e.g. Lustre, NFS), where each metadata operation is a server round trip.

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "wandb[media]>=0.19.0",
     "xarray",
     "zarr>=3",
+    "zarrs>=0.2.3",
```
elynnwu, line 46: `zarrs` pin `zarr>=3.1`, so we should probably do the same.

fme/core/dataset/utils.py
oliverwm1: > It looks like we can also set max workers here to avoid oversubscription. Would be good to confirm the setup you ran to get the 6x speedup

I ran with the default options for zarrs.

> codec_pipeline.file_handle_cache_size

I suspect this wouldn't help that much with training where we are shuffling data and so file access is irregular and we wouldn't get many cache hits. Maybe it would help more for loading data for inference where we do sometimes access chunks from the same shard in subsequent calls. Still, will leave this fine tuning for later.

pyproject.toml
oliverwm1: updated

fme/core/dataset/utils.py
elynnwu: Make sense, okay let's not worry about this additional settings.

elynnwu (approved):

PR #1425, opened 2026-08-10
Abort NCCL communicators on termination instead of collective teardown

mcgibbon:
Preempted multi-GPU training jobs fault the GPU fabric on the way out, and Beaker cordons the node: a rank that dies while a peer's NCCL kernels still hold NVLink accesses into its memory raises SXid errors. #1398 addressed this with a collective teardown on SIGTERM, but a Python signal handler only runs when the main thread returns to the interpreter — a rank wedged inside a collective never starts the teardown, so its peers cannot finish theirs.

This PR removes all cross-rank coordination from the exit path. On SIGTERM/SIGINT, each rank aborts its own NCCL communicators (stopping its kernels and releasing wedged threads), waits a short grace period so its peers' aborts finish too, then exits with `128+signum`. The signal is observed on a dedicated listener thread via `signal.set_wakeup_fd`, which works even while the main thread is blocked in a collective. Nothing outside `fme.core.distributed` is aware of the mechanism except the Trainer's optional post-abort checkpoint, keeping these concerns separated.

Writing a restart checkpoint on preemption is kept where it is safe: without model/domain parallelism, rank 0 holds the full model state, so after the abort it saves a restart checkpoint locally — no cross-rank cooperation needed. The save waits until the main thread has blocked in the shutdown context's exit (a main thread running compute between collectives keeps mutating model state until its next collective raises), and is skipped rather than risk a torn snapshot if that never happens. Under model/domain parallelism the state is sharded and cannot be assembled without the aborted communicators, so those jobs resume from the every-n-batches checkpoint instead.

Changes:
- `fme.core.distributed.shutdown`: rewritten around a wakeup-fd listener thread — abort, post-abort callbacks, grace period, exit. Repeated signals are ignored once the exit is underway, so a second Ctrl-C cannot cut the grace period short; an `os.register_at_fork` hook disarms fork-started DataLoader workers, so a signal sent to a worker cannot masquerade as the parent's preemption.
- `DistributedBackend.abort` (new): locally release the backend's communicators. Non-collective and callable from a non-main thread, unlike `shutdown`; covers every process group, mesh subgroups included.
- `Trainer`: registers its restart-checkpoint write as a post-abort callback, skipped under model/domain parallelism.

- [x] Tests added

Resolves the node-cordoning issue addressed by #1407.


fme/core/generics/trainer.py
```diff
@@ -314,36 +313,10 @@ def __init__(
...
-                        "restart checkpoints as it is unsafe to do so"
-                    )
-                elif not self._started_training:
-                    logging.info(
-                        "Not saving restart checkpoints as training has not started"
-                    )
-                else:
-                    self._save_restart_checkpoints()
-
-        add_post_shutdown_callback(save_restart_checkpoints_on_terminate)
```
mcgibbon, line 345: Removing this as it wasn't really working anyways, and would make the solution required to fix the error much more complicated.

mcgibbon:
Claude: Validation and process notes, kept out of the description:

**Hardware validation.** Before implementing, the abort-then-exit mechanism was validated by a standalone 8-GPU DDP probe (the `nccl_abort_probe.py` in this PR): torchrun SIGTERM'd mid-`backward()` with one rank deliberately absent from the gradient all-reduces, so the surviving ranks were wedged in live NCCL collectives when the signal arrived. Runs on both B200 ([01KZPNQ7CCVMJ0CB06NRYAQHC9](https://beaker.org/ex/01KZPNQ7CCVMJ0CB06NRYAQHC9), titan) and H100 ([01KZPP6TG742HM4YBHPWMW2TXK](https://beaker.org/ex/01KZPP6TG742HM4YBHPWMW2TXK), jupiter) passed: NCCL was on genuine NVLink peer transport (P2P/CUMEM + NVLS), every abort returned in under a second, all wedged ranks' main threads were released, no rank exited before every abort had returned, a fresh NCCL job on the same GPUs succeeded afterwards, and neither node was cordoned.

**Motivation for replacing #1398's approach outright**: over a recent two-week window, ~75% of 8-GPU training-loop preemptions faulted the fabric despite the collective teardown.

**Fork hazard found during implementation.** The wakeup fd is inherited by fork-started DataLoader workers and feeds the parent's pipe, so a SIGTERM delivered to a worker (e.g. `DataLoader._shutdown_workers`' terminate path) would read as the parent's own preemption and kill the job. Demonstrated empirically, closed with the `os.register_at_fork` disarm, and pinned by `test_shutdown_dataloader.py`.

**Why a listener thread + wakeup fd rather than `pthread_sigmask`+`sigwait`**: threads that exist before the mask is set (torch spawns many at import) keep SIGTERM unblocked, and a process-directed signal delivered to any of them takes the default action — verified locally. The wakeup fd hears the signal regardless of which thread receives it.

fme/core/distributed/shutdown.py
mcgibbon, line 1: Claude, the changes to this file are quite extensive. While the new state looks generally nicer and more concise, can you please clearly describe the changes made here?

fme/core/distributed/nccl_abort_probe.py
mcgibbon, line 1: Please describe for reviewers the motivation for committing this probe, or why they might want to ask for it to be removed before merge.

fme/core/distributed/shutdown.py
mcgibbon: Claude: A map of the rewrite, old → new:

**Unchanged**: the interface (`handle_termination_signals` context manager entered by `Distributed.context()`); the guards (installed only on the main thread, never in DataLoader workers); the `128+signum` exit convention; dispositions restored on context exit.

**How the signal is observed** — the core change. The old design did its work inside a Python signal handler, which runs only when the main thread returns to the interpreter; a rank wedged in a collective's stream sync never got there, so the teardown never started on exactly the ranks that mattered. Now the handler is a no-op that exists only so CPython delivers the signal, and delivery is observed on a dedicated listener thread through `signal.set_wakeup_fd` — the C-level handler writes the signal number to a pipe at delivery time, whatever the main thread is doing.

**What is done about the signal**: previously a collective `destroy_process_group` bounded by a 20s watchdog timer, then best-effort post-shutdown callbacks (the restart-checkpoint write), then `sys.exit`. Now: `abort()` (local `ncclCommAbort` of every process group — releases this rank's kernels and wedged threads without needing any peer), a grace-period sleep so peers' aborts finish while our memory is still mapped, then `os._exit(128+signum)`.

**Deleted with nothing replacing them**, because the multi-step teardown they sequenced no longer exists: the `_Phase` machine (repeated-signal handling), the `_hard_exit_after` watchdog thread (abort is local and prompt, so there is nothing left to bound), and the `add_post_shutdown_callback`/`clear_post_shutdown_callbacks` registry (its only client was the Trainer's checkpoint-on-terminate, removed in this PR).

**Fork protection changed shape**: the old handler checked `os.getpid()` at signal time and re-raised with `SIG_DFL` in fork children. The wakeup fd needs stronger handling — unlike a handler, it acts at C level before any Python check can run, and a fork-started DataLoader worker's inherited fd feeds the *parent's* pipe, so a signal sent to a worker would read as the parent's own preemption. An `os.register_at_fork` hook now restores default dispositions and clears the wakeup fd in every forked child.

**Listener lifecycle**: the pipe and thread are created at context entry; on normal exit the previous dispositions and wakeup fd are restored and a sentinel byte stops the thread, so repeated context entries (the test suite) don't leak threads. The listener ignores bytes from other handled signals (torch's DataLoader installs a SIGCHLD handler, so routine worker exits write to the same pipe).

fme/core/distributed/nccl_abort_probe.py
mcgibbon: Claude: **Motivation**: the probe is the harness that validated the abort mechanism on real hardware before this PR was written — an 8-GPU DDP job, SIGTERM'd mid-`backward()` with one rank deliberately absent from the gradient all-reduces so the others wedge in live NCCL collectives; it asserts every abort returns, no rank exits before the last abort returned, wedged main threads are released, and a fresh NCCL job on the same GPUs then succeeds. Committed, its 2-GPU pytest wrappers give GPU CI the only coverage of a *real* NCCL wedge-and-abort (the unit tests in `test_shutdown.py` stub the abort), and the probe itself remains the ready-made tool for re-validating on a new NCCL/driver/hardware generation, as was done on B200 and H100 for this PR.

**Why a reviewer might ask for its removal**: it is ~400 lines that deliberately re-implement the listener mechanism inline rather than importing it from `fme` (so it runs standalone on a deps-only image, no `fme` install), which means it does not regression-test the production code path itself and its copy of the mechanism can drift from `shutdown.py`'s; and the wrappers add a couple of minutes of GPU CI. If that trade reads poorly, the fallback is to drop the probe + wrappers from the repo and keep them on a branch for future hardware validation.

mcgibbon:
Claude: An independent pre-review agent audited the PR; `8d1d65f97` addresses its findings. The two that mattered:

- **The grace period was enforced only by an accidental coupling.** Aborting the communicators *releases* a main thread blocked in a collective, which then raises out of the training loop and unwinds; only the context exit's `listener.join(timeout=10.0)` stopped it from exiting ahead of the grace period with the exception's exit code (reproduced with `grace_period=12` → exit 1, grace not honored). Now, once a signal has arrived the listener explicitly owns the process exit: leaving the context blocks unboundedly on it (scheduler SIGKILL as backstop if the abort hangs), documented in the docstring, with a regression test.
- **Single-process runs hard-exited on Ctrl-C** — no traceback, no `finally`/`atexit`, buffered output lost, with no fabric to protect. The listener is now only installed when `is_distributed()`; single-process jobs keep default signal behavior.

Also from the review: the final `destroy_process_group` (itself a collective) now runs while the listener still protects the process; a termination signal racing the context's exit wins over the stop sentinel instead of being dropped; the at-fork disarm closes its inherited pipe fds; the probe's healthcheck job kills its torchrun tree on any interrupt so a test timeout can't orphan two GPUs; and comments now cover the `TORCH_NCCL_ASYNC_ERROR_HANDLING` question and what the probe does and doesn't regression-test.

fme/core/generics/trainer.py
jpdunc23: This seems to be working more reliably in https://github.com/ai2cm/ace/pull/1407. Rather than remove this feature in this PR, could you put it behind a module-level hard-coded flag, e.g. `_SAVE_PREEMPT_CHECKPOINTS=False`?

fme/core/distributed/model_torch_distributed.py
```diff
@@ -471,3 +471,11 @@ def shutdown(self):
...
             logger.info("Shutting down rank %d", self._rank)
         DistributedManager.cleanup()
+
+    def abort(self):
+        if not torch.distributed.is_initialized():
+            return
+        # covers the manager's mesh subgroups too: passing no group aborts
+        # every registered process group. The manager's own state is left
+        # behind, but the process is about to exit.
+        torch.distributed.distributed_c10d._abort_process_group()
```
jpdunc23, line 481: Do you have any reservations about using a private `distributed_c10d` method? AFAIK `_abort_process_group` is an experimental feature, and part of the reason I decided not to explore this PR's approach was because of the private namespace. 

If other popular repos (e.g., decent number of stars) are using `_abort_process_group` in production then that would alleviate some of that concern.

jpdunc23 (commented):
Left a couple of comments. My main concern right now is about the private external namespace / reliance on a potentially experimental feature in `torch.distributed.distributed_c10d._abort_process_group()`. This doesn't necessarily need to block merge, especially if we have good evidence that this PR prevents the faults.

fme/core/generics/trainer.py
mcgibbon: Let's re-introduce this feature but disable it if we're using model/domain parallelism, something we are able to check from Distributed.

mcgibbon: Claude: Done in ebaa13f04 — re-introduced live rather than behind a flag, per this follow-up. The Trainer registers `save_restart_checkpoints_on_terminate` as a post-abort callback, which the listener thread runs between the abort and the grace-period exit; registration is skipped when `world_size != total_data_parallel_ranks` (the same check `require_no_spatial_parallelism` uses), since a domain-parallel stepper's state is sharded and assembling it would need the aborted communicators. The save path is post-abort-safe: `is_root()` reads a rank cached at init, and `save_checkpoint` is local reads plus a tmp-write-and-rename.

Coverage restored: the "preempt" arm of `test_resume_after_interrupted_training_during_epoch` (the callback preserves mid-epoch progress and training resumes from it), plus a listener-level test that callbacks run after the abort and a raising callback cannot block the exit.

fme/core/distributed/model_torch_distributed.py
mcgibbon: Claude: I looked into this. `_abort_process_group` is a thin orchestrator over the *bound* API `ProcessGroup.abort()`: it loops every group's `.abort()` inside `_group_start()`/`_group_end()` (avoiding a known multi-communicator abort deadlock, [pytorch#119797](https://github.com/pytorch/pytorch/issues/119797)) and then clears torch's Python-side `_world` bookkeeping. Hand-rolling the same loop ourselves would touch more private surface (`_world.pg_names`, `_group_start`), not less — enumerating the mesh subgroups has no public path.

Production precedent for the mechanism: [torchft](https://github.com/meta-pytorch/torchft) (meta-pytorch's fault-tolerance library, behind the [fault-tolerant Llama run](https://pytorch.org/blog/fault-tolerant-llama-training-with-2000-synthetic-failures-every-15-seconds-and-no-checkpoints-on-crusoe-l40s/) with thousands of injected failures) is built on `pg.abort()`/`backend.abort()` against live in-flight collectives — and *continues training* afterwards, a strictly harder use than ours; its monarch example calls `_abort_process_group()` itself. The two open pytorch issues about the function ([#151159](https://github.com/pytorch/pytorch/issues/151159), [#149418](https://github.com/pytorch/pytorch/issues/149418)) are both abort-then-*reinit* patterns; we abort-then-exit within seconds, and the 8-GPU probes exercised exactly our pattern — abort from a second thread while peers were wedged mid-`backward()` — on H100 and B200 without incident.

If the symbol changes at a torch upgrade, the failure is loud and contained: the unit tests `mock.patch` it (patching a missing attribute errors in CPU CI), the 2-GPU GPU-CI probe exercises the real abort, and at runtime an abort that raises is caught and the rank still exits after the grace period — degrading to the status quo ante (possible cordon), not a new failure mode. So no strong reservations; if you'd rather minimize the private surface at the cost of the deadlock guard and bookkeeping, the torchft-style `hasattr(pg, 'abort')` loop is a small mechanical swap.

mcgibbon: (531 stars on torchft)

fme/core/distributed/nccl_abort_probe.py
```diff
@@ -0,0 +1,421 @@
+"""Empirical probe: on SIGTERM, does aborting the NCCL communicators from a
```
Arcomano1234: Nit: I do find this a little strange to have in `fme/core/distributed` is there a better place to put this?

fme/core/distributed/shutdown.py
```diff
@@ -1,255 +1,224 @@
-"""Tear the distributed backend down cleanly when a job is preempted.
+"""Quiesce the GPU fabric before a preempted job exits.
```
Arcomano1234: Most Claude speak line of all time

fme/core/distributed/shutdown.py
```diff
@@ -1,255 +1,224 @@
...
+    previous_handlers = {sig: signal.getsignal(sig) for sig in TERMINATION_SIGNALS}
+    previous_fd = signal.set_wakeup_fd(write_fd, warn_on_full_buffer=False)
     for sig in TERMINATION_SIGNALS:
-        signal.signal(sig, handle)
+        signal.signal(sig, _ignore_signal)
+    _pipe_fds = (read_fd, write_fd)
+    _armed = True
     try:
         yield
     finally:
```
Arcomano1234, line 256: I was iterating on this with claude:
### The context's `finally` re-arms the default signal disposition before the grace period finishes

`shutdown.py`'s `finally` (lines 207-224) restores the previous SIGTERM/SIGINT handlers (210-211) and the previous wakeup fd (212) **before** blocking in `listener.join()` (218). In a torchrun worker the previous SIGTERM disposition is `SIG_DFL`, so for the whole duration of that join the process is back to being killed instantly by a second signal.

That window is not small, and it opens at the worst moment. `abort()` is what *releases* the wedged main thread: `ncclCommAbort` stops the local kernels, the collective raises out of the training loop, and the main thread unwinds into the `finally` within milliseconds of the abort — i.e. at the *start* of the listener's work, not the end. It then sits at `SIG_DFL` while the listener still has to run the post-abort callbacks (the restart checkpoint, deliberately unbounded — 7-15 s per #1398's own measurements) and then `time.sleep(grace_period)`.

A signal arriving in that window kills the rank before the grace period completes, while a peer's aborts may not have finished — the fabric fault this PR exists to prevent. It also contradicts the module's own stated invariant at lines 162-163: *"nothing on the way out may `os._exit` or `SIG_DFL` its way around it."* The `finally` does the `SIG_DFL` itself.

#### Reproduction

No GPUs needed — this is pure signal handling, so `abort()` can be stubbed with something that just releases the main thread. Parent sends a second signal 1 s after the first, with `grace_period=5.0`:

| Case | Exit code | Died after | Grace period honored |
|---|---|---|---|
| SIGTERM, one signal | 143 | 5.13 s | yes |
| SIGTERM, two signals | −15 | 1.11 s | **no** |
| SIGINT, two signals | −2 | ~5.5 s | incidentally |

The first row is the design working as intended. The second is the defect: deterministic across three runs, the rank dies by signal 4 s into a 5 s grace period, and the listener's `Exiting with code 143.` line never prints — it was killed mid-window. The exit code is −15 (killed by signal) rather than a clean 143, so torchrun records a signal-killed worker rather than a graceful one.

The SIGINT row differs because the restored disposition there is `default_int_handler`, not `SIG_DFL`. The second SIGINT raises `KeyboardInterrupt` out of `listener.join()`, the main thread escapes the context with a traceback, and the process dies with −2. The grace period survived in all three runs, but only because the daemon listener won a race against interpreter finalization — nothing guarantees that, and the exit code is wrong either way.

<details>
<summary>Repro scripts</summary>

`repro_child.py`:

```python
import sys
import threading
import time

from fme.core.distributed.shutdown import handle_termination_signals

GRACE = 5.0
released = threading.Event()


def abort():
    # stands in for ncclCommAbort: stops local kernels and releases the main
    # thread that was blocked in the collective's stream sync
    sys.stderr.write("ABORT called\n")
    sys.stderr.flush()
    released.set()


start = time.monotonic()
try:
    with handle_termination_signals(abort, grace_period=GRACE):
        sys.stderr.write("READY\n")
        sys.stderr.flush()
        released.wait()  # main thread "wedged" in the collective
        # the abort released it; the collective raises out of the training loop
        raise RuntimeError("collective aborted")
except RuntimeError:
    pass

# only reachable if the listener did NOT own the exit
sys.stderr.write(f"MAIN THREAD ESCAPED after {time.monotonic() - start:.2f}s\n")
```

`repro_parent.py`:

```python
import signal
import subprocess
import sys
import time

GRACE = 5.0
sig = getattr(signal, sys.argv[1])
send_second = sys.argv[2] == "double"

proc = subprocess.Popen(
    [sys.executable, "repro_child.py"], stderr=subprocess.PIPE, text=True
)
while "READY" not in (proc.stderr.readline() or ""):
    pass

start = time.monotonic()
proc.send_signal(sig)
if send_second:
    time.sleep(1.0)
    proc.send_signal(sig)  # arrives during abort/callbacks/grace period

code = proc.wait(timeout=60)
elapsed = time.monotonic() - start
print(f"exit code: {code}   died after: {elapsed:.2f}s   (grace = {GRACE}s)")
print("child stderr:", " | ".join(l for l in proc.stderr.read().split("\n") if l))
print(f"grace period honored: {elapsed >= GRACE}")
```

Run with `python repro_parent.py SIGTERM double`.

</details>

#### Where the second signal comes from

Worth being precise, because it sets the severity. Torch elastic's `_close` (`torch/distributed/elastic/multiprocessing/api.py`) sends the death signal **once**, waits 30 s, then escalates straight to SIGKILL — it never re-sends SIGTERM. So under Kubernetes-style preemption (SIGTERM to PID 1 only) each worker most likely receives exactly one SIGTERM, from the agent. I did not verify whether Beaker signals the whole cgroup instead.

The reliable double-signal path is Ctrl-C, and it takes only a **single keystroke**: the tty delivers SIGINT to the entire foreground process group, hitting every rank directly, and the agent then catches its own copy and forwards the same signal via `_pcontext.close(death_sig)` (`local_elastic_agent.py:360-368` forwards `e.sigval`, not a hardcoded SIGTERM). A second Ctrl-C is near-certain on top of that, since the job now deliberately takes 10-20 s to die.

So this is most reachable interactively rather than on the preemption path the PR targets — but an interactive Ctrl-C on a multi-GPU node faults the fabric the same way.

#### Suggested fix

Move the restoration after the join, so it only runs on the non-terminating path. In the terminating case `join()` never returns, since the listener `os._exit`s:

```python
finally:
    _armed = False
    _pipe_fds = None
    os.write(write_fd, bytes([_STOP_LISTENING]))
    if terminating.is_set():
        listener.join()  # never returns; the listener owns the exit
    else:
        listener.join(timeout=grace_period + 10.0)
        for sig, handler in previous_handlers.items():
            signal.signal(sig, handler)
        signal.set_wakeup_fd(previous_fd)
        os.close(write_fd)
        os.close(read_fd)
```

That closes the window, but it deepens the loss of escalation relative to #1398: with `_ignore_signal` installed for the whole protected period, a hung `abort()` leaves the rank deaf to everything but SIGKILL — which torchrun supplies at 30 s and an interactive launch does not.

Both are fixable together by making the listener's exit condition time-based rather than sequence-based: stamp the time when `abort()` returns, let the listener keep reading the pipe so a repeated signal abandons the remaining callbacks, then always sleep out whatever remains of `grace_period` measured from the abort before `os._exit`. Escalation then costs the operator the checkpoint — which is what they are asking for — but never the fabric-protecting wait, which is the one thing that must not be skippable.

Either way this is worth a test: there is currently no case where a second signal arrives during the protected window, which is why it got through.

Arcomano1234 (commented):
Left a few comments. I am curious your thoughts about the finally: branch in `shutdown.py`I think its a real bug.

fme/core/distributed/nccl_abort_probe.py
mcgibbon: Let's just remove this file, if we really need to we can recover it from this PR but I don't expect to need it again.

fme/core/distributed/shutdown.py
mcgibbon: Yes, let's reduce changes here, maintaining "when a job is preempted" at the end and using plainer terms than "Quiesce" and "fabric" (which are both jargon).

fme/core/distributed/shutdown.py
mcgibbon: Claude please write a test and consider whether this fix is appropriate.

fme/core/distributed/nccl_abort_probe.py
mcgibbon: Claude: Removed in 42add9856, along with its 2-GPU test wrappers and its entry in `test_distributed_context.py`'s exclusion list. Recoverable from this PR's history if a new hardware generation ever needs re-validation.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Reworded in 42add9856 — the docstring now opens with main's original "Tear the distributed backend down safely when a job is preempted." and reuses its error-message description in place of the fabric/quiesce phrasing; the only remaining changes are the ones describing the new mechanism (self-abort + grace period, and the wakeup-fd listener replacing the handler).

fme/core/distributed/shutdown.py
mcgibbon: Claude: Confirmed — the bug is real, and the suggested fix is the right one. Done in 42add9856.

**Test**: `test_a_repeated_signal_cannot_cut_the_grace_period_short`, parametrized over SIGTERM and SIGINT. It stages exactly the reported scenario — the abort releases a wedged main thread, which unwinds into the context's exit; a second signal arrives 0.5 s later, mid-grace-period — and asserts the process still exits with `128+signum` after the full grace period, without the main thread escaping the context. Against the old code it reproduces both rows of the table: SIGTERM dies at −15 mid-grace, and SIGINT raises `KeyboardInterrupt` out of `listener.join()` and escapes the context.

**Fix**: the suggested ordering, with one addition for the race where a signal was delivered but not yet read at context exit — the timed join runs first, and `terminating` is re-checked after it, so a signal that wins that race also holds the main thread with the ignore-handlers still installed rather than timing out into the restore.

**On the escalation trade-off**: I considered the time-based variant (a repeated signal abandons the remaining callbacks but always waits out the grace period from the abort) and deliberately didn't take it. The deafness window the simple fix accepts is bounded in the cases that matter — checkpoint + grace is ~10–20 s, inside torchrun's 30 s SIGTERM-to-SIGKILL escalation, and a scheduler preemption SIGKILLs at its own deadline regardless. The genuinely bad case is a *hung* checkpoint write on an interactive run, where only `kill -9` gets you out. The escalation variant covers that, but needs the callbacks on their own thread plus de-duplication of the first keystroke's near-simultaneous signal pair (the tty delivers SIGINT to every rank directly *and* the agent forwards it ms later) — and a mis-dedupe would silently skip the checkpoint on every interactive Ctrl-C. That felt like more mechanism than the residual risk warrants, but it's a contained follow-up if the deafness worries you.

mcgibbon:
Claude: An independent red-team review (adversarial, correctness-focused) was run against head 42add985; 269ddd9b7 addresses what it found. Three real defects, each now fixed with a regression test that fails against the previous commit:

1. **An unguarded stderr write could silently forfeit the whole protection** (confirmed by repro). The listener's first message was written before `abort()` and outside any try/except; with stderr's reader gone (EPIPE — plausible mid-preemption if the log collector dies first), the write raised and killed the listener thread with `terminating` already set. The abort never ran (the exact fault this PR prevents), and the main thread joined a dead thread and walked on, deaf to everything but SIGKILL. `write_stderr` (now public) never raises.

2. **The post-abort checkpoint could race a still-running main thread** (new to this design — the old handler ran the save *on* the main thread). The abort only releases a rank blocked in a collective; one running compute between collectives keeps mutating model/optimizer/EMA state until its next collective raises, up to ~a batch overlapping the save — a torn `latest_checkpoint_path` that resume would silently trust. Callbacks now run only once the main thread has blocked in the shutdown context's exit (an event set in the `finally`, after which that thread only joins the listener), and are skipped after `park_timeout` (5 s) rather than snapshot alongside a live main thread. This also makes the EMA-context guard sound: a parked main thread has unwound every inner context manager.

3. **The at-fork disarm left module state set in the child** (confirmed by repro), so a second-generation fork (dataset code using multiprocessing inside a fork-started worker) re-closed fd numbers the worker had reused. The disarm now clears `_armed`/`_pipe_fds`.

Also from the review: the Trainer callback no longer touches the logging module (a handler lock held by a thread blocked on a stalled stderr would deadlock the exit path — the same reason `shutdown.py` always used raw writes); callbacks are cleared on normal context exit so a later context can't fire them against torn-down state; and the benign sub-millisecond window where a signal racing normal context exit is dropped is now documented in place.

The review could not break the main-line signal flow (entry/exit orderings, byte protocol, fork disarm, the repeated-signal fix) and confirmed the tests assert what their names claim.

Arcomano1234 (approved):
Approved, I am not seeing any more obvious issues although this code is honestly pretty complicated and I am not sure I have a perfect grasp of everything its doing and why. I guess my question is do we just merge it as long as it fixes the problems and doesnt lead to any other problems we live with the fact we might need to rely on Claude to maintain this?

PR #1432, opened 2026-08-11
Post-corrector residual tracking, presentation, and optimization for prescribed prognostics

jpdunc23:
Variables in `prescribed_prognostic_names` are overwritten with the target value after the corrector runs, so the model's own output for those fields is discarded: their loss contribution is zero and downstream metrics compare the target with itself. This PR records the post-corrector residual `r' = y − x'` at that overwrite, presents the model's corrected output `x' = y − r'` to aggregators and writers while the rollout continues to consume exactly `y`, and adds an opt-in training-loss optimization of `x'` against `y` for selected prescribed names — so the model trains through the corrector on the fields it is prevented from drifting on. The new config is grouped with `corrector_loss` under one `loss_features` surface.

Changes:
- `fme.core.step.output.StepOutput` — new `prescribed_residual` mapping carried per step, `presented_output` recovering `x'`, and `stack_diagnostics` stacking the residual series
- `fme.core.step.step_diagnostics.StepDiagnostics` — residual series carried through the structural ops and exported as a `prescribed_residuals` dataset by the step-diagnostics writer
- `fme.core.step.single_module.step_with_adjustments` — computes `r'` at the prescribed-prognostic overwrite; `detach_prescribed_residuals` (default detached) threaded from `StepABC.set_detach_prescribed_residuals` through all step classes, forwarded by `MultiCallStep`
- `fme.core.loss.StepLoss` — optional `residuals` argument and `StepLossPrescribedArgs`: the loss prediction for selected names becomes `predict − r' = x'`, targets untouched; unselected prescribed names keep their zero contribution
- `fme.core.loss_features` — new `LossFeaturesConfig` grouping `corrector` (the former `TrainStepperConfig.corrector_loss`) with `PrescribedPrognosticOptimizationConfig`; opt-in `names_and_prefixes` selection validated at build against the prescribed names, ocean-written names rejected
- `fme.ace.stepper.single_module` — `Stepper.build_loss(..., loss_features=...)` builds both features and flips the two detach flags independently; residual masking in `Stepper.step`; presentation applied in `Stepper.predict` and `TrainStepper._accumulate_loss` (training `gen_data`, validation metrics, inline-inference aggregators, and inference writers all see `x'`; the next-step input and restart state stay exactly `y`)

- [ ] Tests added

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -104,6 +104,11 @@ from fme.ace.stepper.time_length_probabilities import (
 from fme.core.cli import ResumeResultsConfig
 from fme.core.corrector.atmosphere import AtmosphereCorrectorConfig
 from fme.core.corrector.ice import IceCorrectorConfig
+from fme.core.corrector.loss import (
+    CorrectorLossConfig,
+    CorrectorRegularizationConfig,
+    PreCorrectorOptimizationConfig,
+)
 from fme.core.corrector.ocean import OceanCorrectorConfig
 from fme.core.dataset.concat import ConcatDatasetConfig
 from fme.core.dataset.merged import MergeDatasetConfig, MergeNoConcatDatasetConfig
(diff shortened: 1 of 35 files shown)


jpdunc23:
Plan-review stage: the diff against `feature/corrector-loss-config` is `pr-plan.md` only (it replaces the base branch's retained plan file in the branch view; the diff against the base shows just this PR's plan). Implementation follows on this PR after plan review.

mcgibbon:
If I understand correctly, the motivation for this is the idea that optimizing the network to produce a target it doesn't actually use (discards with overwrite) will improve some scientific behavior - can we avoid merging a feature like this into main until we have a specific behavior we expect to improve with this pathway, and show experimentally that it does?

PR #1434, opened 2026-08-13
Inference-time corrector-disable override, coupled-inference seed, and the CM4 stratospheric-humidity ablation configs

elynnwu:
The CM4 1pctCO2 coupled emulator responds to stratospheric humidity
(`specific_total_water_0`), apparently including in the deep ocean — where a
stratospheric perturbation has no fast physical path. Separating a genuine
learned sensitivity from an artifact of the energy correctors requires
evaluation-time ablations of a *fixed* checkpoint. Prescribing a prognostic from
forcing was already supported; disabling a corrector was not, because corrector
options are serialized into the checkpoint. Coupled inference also had no seed,
so two runs differing only in a stepper override also differed by an
uncontrolled noise draw from the stochastic atmosphere.

This adds the two missing evaluation-time controls and the ablation configs that
use them. Branch targets `exper/2026-07-16-jamesd`.

Changes:
- `fme.ace.stepper.StepperOverrideConfig.disable_corrections` — a list of
  corrector option names to switch off when loading a checkpoint for inference.
  A correction is disabled by resetting its option to that field's declared
  default, which is the "not applied" value for every option that is a
  correction, so the rule is uniform across corrector types. Options that are
  not corrections are listed in `NON_CORRECTION_OPTIONS` and rejected:
  `corrector_disabled_epochs` schedules the corrector during training, and
  `keep_gradient_through_clamps` only selects whether the clamps use a
  straight-through estimator (a no-op under `no_grad`), so resetting either
  would ablate nothing. An unknown, non-correction, or already-off name raises:
  each would make an ablation silently reproduce the un-ablated run.
- Plumbed along the existing `prescribed_prognostic_names` path —
  `fme.core.corrector.registry.CorrectorConfigABC`,
  `fme.core.registry.corrector.CorrectorSelector` (which also refreshes its
  serialized `config` mapping, the path the ocean corrector takes),
  `fme.core.step.step.StepConfigABC`/`StepSelector`, each step config holding a
  corrector, `MultiCallStepConfig`, `StepperConfig` and `Stepper` (rebuilding
  the step from the mutated config and reloading its trained state, as
  `replace_ocean` does). It is abstract on `StepConfigABC`, as its
  `replace_ocean`/`replace_prescribed_prognostic_names` neighbours are, so a step
  type with no corrector has to raise deliberately.
- `fme.coupled.inference.evaluator` admits the new override: unlike
  `ocean`/`multi_call`/`derived_forcings` it changes no variable names, so it
  cannot leave `CoupledStepperConfig`'s cached cross-component forcing-name sets
  stale.
- `OceanCorrectorConfig.remove_deprecated_keys` no longer crashes on an explicit
  `ocean_heat_content_correction: null` — pre-existing, but newly reachable now
  that an ablation writes that value.
- `seed` for coupled inference — `seed: int | None` on the coupled
  `InferenceEvaluatorConfig` and `InferenceConfig`, plus `apply_config_seed` on
  `CoupledPrognosticState` and the coupled `InferenceGriddedData`, mirroring
  `fme.ace`. The coupled stepper already carries `stepper_state` across outer
  coupled steps, so a seeded IC propagates for the whole rollout, independent of
  `coupled_steps_in_memory`.
- `configs/experiments/cm4_1pctCO2-train_0046to0125/{coupled_baseline,atmos_baseline}/q0_ablations/`
  — 12 arms (8 coupled, 4 uncoupled) as one evaluator config per realm plus
  `experiments.txt` rows that differ only in the `override` dotlist. All arms
  share one IC (`0046-01-06T00:00:00`), 80 noleap years with the realms matched
  step for step (5840 coupled = 116800 6-hourly), and `seed: 0`. A smoke config
  at ~1.1 yr is included for a pre-campaign check of one arm of each kind. The
  uncoupled arms log the aggregator's per-variable correction magnitude, which
  measures how much work the energy correction is doing; the coupled arms cannot,
  because `fme.coupled` drops the corrector's per-step deltas before they reach a
  writer or aggregator (noted in the configs, tracked separately).

- [x] Tests added


elynnwu:
Claude: **Blocking — `save_step_diagnostics: true` is inert in the coupled evaluator, so none of the 8 coupled arms will produce the correction deltas the campaign is built around.**

The PR description says "the coupled arms write 50-day-mean correction deltas (the coupled evaluator has no step-diagnostics aggregator)". The aggregator is indeed missing, but so is the *writer* path: the coupled stepper never plumbs step diagnostics through at all.

Static proof:

- `grep -rn step_diagnostics fme/coupled/` returns **zero** hits (excluding tests).
- `fme.ace`'s `Stepper.predict` attaches them explicitly (`fme/ace/stepper/single_module.py:1326`, `step_diagnostics=StepOutput.stack_diagnostics(output_list)`). The coupled `CoupledStepper._predict` / `_process_prediction_generator_list` has no equivalent — `ComponentStepPrediction` doesn't even carry the field, and the `StepOutput`s it builds pass only `output` and `stepper_state`.
- `CoupledPairedData.from_coupled_batch_data` (`fme/coupled/data_loading/batch_data.py:200`) constructs each realm's `PairedData` without `step_diagnostics`, so it defaults to `None`.
- `PairedDataWriter.append_batch` writes only `if ... batch.step_diagnostics is not None`.

So it is structurally impossible for the coupled path to write a delta file, regardless of whether a corrector modified anything.

Verified empirically: I ran the coupled evaluator on the mock coupled fixture with an atmosphere corrector that definitely fires (`force_positive_names=["a_prog"]`) and `save_step_diagnostics: true` on both realms:

```
PROBE ocean step_diagnostics dir exists: False
PROBE atmosphere step_diagnostics dir exists: False
```

Knock-on effect: `time_coarsen` in both coupled configs is inert too. Its only two consumers are the raw prediction writer (disabled here via `save_prediction_files: false`) and the step-diagnostics writer (never fed). Monthly files ignore it. So the `coarsen_factor: 10` / `coarsen_factor: 200` arithmetic and the "~30 GB per variable" reasoning in the comments describe a file that is never written.

Two ways out, and the choice is yours rather than something I should pick:

1. Plumb step diagnostics through the coupled stepper — carry the per-step diagnostics on `ComponentStepPrediction`, stack them per realm, attach to each realm's `BatchData` at the same final-reconstruction point `fme.ace` uses (after the derived-variables prepend/remove, so they never pass through a time-changing `BatchData` method), and thread them into `CoupledPairedData.from_coupled_batch_data`. That is a reviewable feature in its own right and should probably be its own commit with its own test.
2. Drop `save_step_diagnostics` and `time_coarsen` from both coupled configs, and reword the design so the coupled arms don't claim a direct corrector-work measurement (the uncoupled arms keep theirs via `aggregator.step_diagnostics`, which does work).

I did not apply either.

elynnwu:
Claude: **Applied fix (81af853) — the "reset to the field's default" rule is not correct for every option: `keep_gradient_through_clamps`.**

The docstring and the PR description both claim the field default is "the 'not applied' value for every option on every corrector config". I checked all four corrector configs by introspection. It holds for every field except one:

| config | field | default | disabling it removes a correction? |
|---|---|---|---|
| Atmosphere | `conserve_dry_air` | `False` | yes |
| Atmosphere | `zero_global_mean_moisture_advection` | `False` | yes |
| Atmosphere | `moisture_budget_correction` | `None` | yes |
| Atmosphere | `force_positive_names` | `[]` | yes |
| Atmosphere | `total_energy_budget_correction` | `None` | yes |
| Atmosphere | `clip_frozen_precipitation` | `False` | yes |
| Atmosphere / Ocean | `keep_gradient_through_clamps` | `False` | **no** |
| Ocean | `force_positive_names`, `sea_ice_fraction_correction`, `surface_energy_flux_correction`, `ocean_heat_content_correction` | `[]` / `None` | yes |
| Ice | `budget_correction` | `None` | yes |

`keep_gradient_through_clamps` only selects whether the `force_positive_names` clamp and the sea-ice-fraction bound/rebalance use a straight-through estimator. The forward value is clamped either way, and under `torch.no_grad()` at inference the flag is a total no-op. So `disable_corrections(["keep_gradient_through_clamps"])` was accepted and ablated nothing — verified before the fix:

```
keep_gradient_through_clamps accepted; force_positive still on: ['so_0'] False
```

That is precisely the silent-no-op the unknown-name and already-off errors exist to prevent, so it belongs in the same exclusion as `corrector_disabled_epochs`.

Fix: replaced the hardcoded `{"corrector_disabled_epochs"}` literal with a `NON_CORRECTION_OPTIONS: ClassVar[frozenset[str]]` on `CorrectorConfigABC` that subclasses extend, and added `keep_gradient_through_clamps` to it on `AtmosphereCorrectorConfig` and `OceanCorrectorConfig`. The error message's "whose corrections are [...]" list now excludes it too. Tests: extended the existing rejection parametrization with a `gradient_shaping` case, plus one asserting the clamp is left untouched, and one covering rejection through the `CorrectorSelector`.

Note `clip_frozen_precipitation` stays disable-able — it does modify the forward data, so it is a genuine correction.

elynnwu:
Claude: **Applied fix (6b691f1) — the two `noOHCC` arms produced a corrector config that could not be reloaded.**

`OceanCorrectorConfig.remove_deprecated_keys` tested membership on the value without first checking it is a dict:

```python
elif (
    "ocean_heat_content_correction" in state_copy
    and "method" in state_copy["ocean_heat_content_correction"]
    ...
```

An explicit `ocean_heat_content_correction: null` therefore raised:

```
TypeError: argument of type 'NoneType' is not iterable
  fme/core/corrector/ocean.py:292, in remove_deprecated_keys
```

This is pre-existing on `exper/2026-07-16-jamesd` (identical code there), but previously unreachable in practice because a null was only ever *absent* from a serialized config. This PR makes writing an explicit null routine: `CorrectorSelector.disable_corrections` does `self.config = dataclasses.asdict(...)`, so `ocean_stepper_override.disable_corrections=[ocean_heat_content_correction]` — rows `noOHCC` and `q0-noOHCC` — puts `ocean_heat_content_correction: null` into the serialized config.

The arms would not have crashed at launch: `StepSelector._step_config_instance` and `CorrectorSelector._corrector_config_instance` are both cached in `__post_init__`, and `StepperConfig.get_stepper` builds from the cached mutated instance rather than re-reading `config`, so nothing reconstructs the selector during the run. But any later reload of the ablated config crashes — which includes the smoke-test check the config itself calls for ("the dumped config.yaml shows the corrector option as null"), a rerun of the arm from its dumped config, and a `CoupledStepper.get_state()`/`from_state()` round-trip.

Fix: guard the branch with `isinstance(state_copy.get(...), dict)`, matching the style the neighbouring `sea_ice_fraction_correction` branch already uses. Left the `# FIXME: don't merge!` `constant_temperature` -> `scaled_temperature` coercion alone — that is this experiment branch's own marker.

Test added: `test_reload_of_a_disabled_ocean_heat_content_correction`, which fails before the fix.

I also extended `test_disable_corrections_through_a_corrector_selector` to round-trip through `CorrectorSelector.from_state(dataclasses.asdict(selector))`. Refreshing the serialized `config` mapping is only worth doing if the result loads, and the original test only inspected the dict. That round-trip is otherwise faithful — I confirmed nested sub-configs, `force_positive_names`, and a non-zero `corrector_disabled_epochs` on the wrapped config all survive `asdict` -> `from_state` unchanged.

elynnwu:
Claude: **Conditional on how the step-diagnostics finding is resolved — the coupled atmosphere `names` list would filter out the TEBC delta even if the writer worked.**

`DataWriterConfig._build_step_diagnostics_writer` passes `save_names=self.names` to the step-diagnostics `RawDataWriter` (`fme/ace/inference/data_writer/main.py:183`), so `names` filters the delta file, not just the prediction and monthly files. The existing test on the base branch locks this in: `test_step_diagnostics_writer_writes_delta_series` asserts `"b" not in ds.data_vars` when `save_names=["a"]`. The `names:` docstring ("Names of variables to save in the prediction and monthly netCDF files") is stale on that point.

The delta keys are plain field names (`StepDiagnostics.to_datasets` uses `self.delta` keys verbatim), so this maps directly onto what each ablated correction modifies:

- **Ocean — correct as written.** `_correct_hfds` picks `hfds` if present in `gen_data`, else `hfds_total_area`; the training config's ocean `out_names` has `hfds_total_area`, which is in the config's `names`. So the SEFC delta lands. `_force_conserve_ocean_heat_content` modifies `thetao_*` and `sst`, and `sst`/`thetao_9`/`thetao_18` are in `names`. Both ocean arms are covered. (`ocean_heat_content` in `names` is a derived variable that never appears in a delta — harmless, just inert for this file.)
- **Atmosphere — not covered.** `_force_conserve_total_energy` returns only `air_temperature_*` (`fme/core/corrector/atmosphere.py:662-664`), and the config's atmosphere `names` is `surface_temperature, specific_total_water_0, LHTFLsfc, SHTFLsfc, PRATEsfc`. No `air_temperature_*`, so the TEBC delta — the entire measurement for `noTEBC` / `q0-noTEBC` — is filtered out. What would land instead is the `force_positive_names` delta on `specific_total_water_0` and `PRATEsfc`, plus the `advection_and_precipitation` moisture correction's `PRATEsfc` term; `conserve_dry_air`'s `PRESsfc` delta is also filtered out.

If you take the plumb-it-through route, one level is sufficient rather than all eight: `temperature_correction` is `energy_correction / area_weighted_mean(...)`, i.e. a per-sample scalar added identically to every level and every grid cell, so `air_temperature_0` captures the correction in full.

I have not applied this, since it is moot under the drop-the-config resolution.

elynnwu:
Claude: Non-blocking notes on the parts I checked and either found sound or would change only on your say-so.

**1. The `CoupledStepperConfig` stale-cache claim is correct.** This was flagged as the most serious defect if wrong, so I chased it: `__post_init__` really does cache `_ocean_forcing_exogenous_names`, `_atmosphere_forcing_exogenous_names`, `_shared_forcing_exogenous_names`, `_atmosphere_to_ocean_forcing_names`, `_ocean_to_atmosphere_forcing_names` (`fme/coupled/stepper.py:279-341`) — so the caches are real, and the docstring's "computed on demand" phrasing in `apply_coupled_stepper_config_inference_overrides` is looser than the code. But every one derives from `stepper.input_only_names` / `output_names` / `all_names`, and none of those consults the corrector: `SingleModuleStepConfig.input_names` is `in_names` (union `ocean.forcing_names`), `output_names` is `out_names` plus secondary-decoder names. I grepped all four corrector-holding step configs (`single_module`, `fcn3`, `radiation`, `secondary_module`) — `self.corrector` appears only in the docstring, the field declaration, and `get_step`/`_build`, never in a name-producing property. So `disable_corrections` cannot invalidate them. Claim holds.

**2. The `Stepper.disable_corrections` rebuild-and-reload is sound, including for the stateful correction.** `conserve_dry_air`'s reference dry-air mass lives in `CorrectorState.global_dry_air_mass`, which is threaded through `StepperState` on the prognostic state, not in the step's checkpoint state — so at load time, before any rollout, there is nothing stateful to lose, whichever correction is disabled. The `Step.get_state()` -> `load_state()` corrector round-trip carries only `EpochScheduledCorrector`'s `corrector_disabled` flag, and `load_state` raises when `_disabled_epochs > 0` and that key is absent — but `_disabled_epochs` comes from `corrector_disabled_epochs`, which `disable_corrections` refuses to touch, so old and new correctors always agree and the key is always present or always absent together. `test_load_stepper_with_disable_corrections_override` covers the trained weights surviving (-3 -> -2 confirms the +1 module is intact).

Worth noting for the reader: the coupled arms don't actually take this path. `load_stepper` for a single coupled checkpoint calls `apply_stepper_override` on the built component `Stepper`s, and `load_stepper_config` separately takes the config path via `apply_stepper_override_to_stepper_config`; the config path needs no rebuild at all.

**3. Coupled `apply_config_seed`: both components share one seed value.** Today this is harmless — the ocean is deterministic. But `RandomState.from_seed` is a plain `manual_seed`, and a single generator drives the whole batch, so if the ocean component ever becomes stochastic the two components would draw *identical* noise fields on comparable shapes, imposing a spurious ocean-atmosphere noise correlation. The docstring's "they are separate models, so their noise sequences do not interact" answers the wrong worry — the risk is correlation, not interaction. Cheap insurance: derive distinct per-component seeds (`seed` / `seed + 1`, or hash the realm name in) and say so in the docstring. Your call whether it's worth the churn now.

The rest of the seed change checks out. `data.apply_config_seed(config.seed)` is placed after ensemble broadcasting in both runners (matching `fme.ace`), and the seed does reach the whole rollout across memory windows: `Looper` feeds each window's returned prognostic state into the next `predict`, `predict_paired` builds it from `gen_data.<realm>_data.get_end(...)`, and both `subset_names` and `select_time_slice` propagate `stepper_state` (`fme/ace/data_loading/batch_data.py:844,886`), with `_prescribe_ic_sst` explicitly preserving it too (`fme/coupled/stepper.py:1016`).

**4. `zero_global_mean_moisture_advection` is a latent silent-wrong-thing for a *future* arm.** `AtmosphereCorrectorConfig`'s docstring states the invariant "`zero_global_mean_moisture_advection` must be True if using a `moisture_budget_correction` option other than `None`", but nothing enforces it — no `__post_init__` on any corrector config, and `disable_corrections` uses `setattr`, which would bypass one anyway. The CM4 atmosphere sets `moisture_budget_correction: advection_and_precipitation`, so `disable_corrections=[zero_global_mean_moisture_advection]` — a plausible next arm in this same campaign — would silently produce a documented-invalid corrector. Not this PR's bug (unenforced on main too), but this PR is what makes it reachable from a config line. A validation in `disable_corrections`, or an `__post_init__` on `AtmosphereCorrectorConfig`, would close it.

**5. `StepConfigABC.disable_corrections`: I'd make it `@abc.abstractmethod`.** Its two nearest siblings, `replace_ocean` and `replace_prescribed_prognostic_names`, are both abstract, and `replace_ocean` is abstract even though not every step type has an ocean — so the repo's convention here is "abstract, subclass decides". All six in-tree subclasses already override `disable_corrections`, which makes the concrete raise dead code today; the only cost of flipping it is a one-line addition to `MockStepConfig` in `test_step_registry.py`. The upside is that a new step type is forced to make the decision rather than inheriting a raise it may not have considered. Judgment call and you reasoned about it explicitly, so I left it alone.

**6. Things I verified and found correct, for the record.**

- All 24 combinations (12 arms x both configs) tokenize through `evaluate.sh`'s `eval "OVERRIDE_ARGV=($OVERRIDE_ARGS)"` exactly as intended and pass the real `fme.{coupled,ace}.validate_config --config_type evaluator`. I reproduced the `cut -f8` extraction and the `eval` re-tokenization verbatim. The `base` arms' empty override field yields a zero-element array, and `--override` is `nargs="*"`, so `--override` with no value is accepted rather than an argparse error.
- Numerics: 5840 x 20 = 116800 matches the uncoupled `n_forward_steps`; 5840 x 5 days = 29200 = 80 noleap years exactly, so 0046-01-06 -> 0126-01-06 is right; 5839 is indeed prime. `coupled_steps_in_memory: 10` divides 5840 and 80. `forward_steps_in_memory: 20` divides 116800 and 1600.
- The `time_coarsen` factors are internally consistent with how `get_data_writer` validates them per realm — ocean against `(coupled_steps_in_memory, n_coupled_steps)` and atmosphere against those times `n_inner_steps` (`fme/coupled/inference/evaluator.py:378-397`) — so 10 and 200 both divide their realm's window and total. (Inert, per the blocking comment, but not *wrong*.)
- `experiments.txt` rows carry 13 (coupled) / 11 (atmos) fields against 14/11-field headers; the omitted trailing `shared_mem` reads as empty through `cut -f14` and falls back to 20GiB. Consistent with the sibling files on the base branch, which stop at 7 fields.
- Lint and mypy pass on every changed file; the affected suites (`fme/core/corrector`, `fme/core/registry`, `fme/core/step`, `fme/coupled`, `fme/ace/stepper`) are 704 passed / 2 skipped including my additions.

elynnwu:
Claude: **Verdict: not ready for review.** One blocking item.

Pre-review of `0d02b8f5` (base `8f67d45d`), by an agent without the authoring session's context.

**Blocks merge**

1. `save_step_diagnostics: true` is inert in the coupled evaluator — the coupled stepper never plumbs step diagnostics through, so none of the 8 coupled arms will write a correction-delta file, and `time_coarsen` in those configs has no consumer either. Verified statically and by running the coupled evaluator. Half the campaign's stated measurement doesn't happen. Needs either plumbing step diagnostics through the coupled path (its own commit + test) or removing the config and the claim. Details and the two options in the comment above; I did not choose for you.

Contingent on how (1) is resolved: if you plumb it through, the coupled atmosphere `names` list also needs an `air_temperature_*` entry, or the TEBC delta is filtered out of the file even once it is written. `air_temperature_0` alone suffices — the correction is uniform in space and height.

**Fixed and pushed** (both have failing-first tests; lint, mypy, and the affected suites are green)

- `81af853` — `keep_gradient_through_clamps` was accepted by `disable_corrections` but disables no correction, so naming it silently reproduced the un-ablated run. The "reset to the field's declared default" rule is correct for every other option on all four corrector configs; this one is a gradient-shaping flag, not a correction. Replaced the hardcoded exclusion with an extensible `NON_CORRECTION_OPTIONS` class variable.
- `6b691f1` — the two `noOHCC` arms wrote `ocean_heat_content_correction: null` into the serialized corrector config, which `OceanCorrectorConfig.remove_deprecated_keys` then crashed on (`TypeError: argument of type 'NoneType' is not iterable`). Pre-existing on the base branch but newly reachable. The arms would have launched fine; any later reload of the ablated config — including the smoke test's own "the dumped config.yaml shows the corrector option as null" check — would not.

**Checked and sound**

The claim flagged as most serious if wrong — that `disable_corrections` cannot leave `CoupledStepperConfig`'s cached cross-component forcing-name sets stale — holds. The caches are real, but no name-producing property on any of the four corrector-holding step configs consults the corrector. The `Stepper` rebuild-and-reload path preserves the trained state and handles corrector state correctly, including when the disabled correction is the stateful one (`conserve_dry_air`'s reference lives in the rollout's `CorrectorState`, not the checkpoint, and `disable_corrections` runs before any rollout). The `CorrectorSelector` `asdict` refresh round-trips faithfully through serialization. The seed reaches the whole rollout across memory windows. All 24 arm/config combinations tokenize as intended through `evaluate.sh` and pass the real validators, and the step arithmetic (5840 x 20 = 116800, 80 noleap years, the `time_coarsen` divisibility per realm) checks out.

Four non-blocking suggestions are in the comment above — the shared per-component seed, the unenforced `zero_global_mean_moisture_advection` invariant, and making `StepConfigABC.disable_corrections` abstract. None gates the merge.

Re-request pre-review once (1) is resolved.

elynnwu:
Claude: Delta re-review, `6b691f128..60c595492` (plus my `fe65a3e4b`). Taking the four checks in order.

**1. Coupled configs internally consistent — one orphaned claim, fixed in `fe65a3e4b`.**

The `save_step_diagnostics` and `time_coarsen` blocks are gone from both realms in both configs, and nothing orphaned survives in the 80yr config. But the smoke config's header checklist still read:

```
# to confirm the override reached the checkpoint (the dumped config.yaml shows
# the corrector option as null), that every requested monthly name is actually
# present in the prediction data (an absent name only warns), and that the
# step-diagnostics files appear.
```

That last clause sat eight lines above the new comment explaining that the coupled path writes no such files — two contradictory instructions in one file, in the one config whose entire job is to be the pre-campaign checklist. And because the uncoupled smoke config says "See that config for what to check", the stale instruction reached both realms. Dropped the clause; the remaining two checks are both still valid and both still achievable.

I swept the rest: the only other `step_diagnostics` hits under `q0_ablations/` are the uncoupled configs' `aggregator.step_diagnostics` blocks, which are the mechanism that does work. No `time_coarsen` and no delta/coarsening arithmetic left anywhere in the coupled configs, so nothing describes a file that isn't written.

All 24 arm/config combinations still tokenize through `evaluate.sh`'s `eval "OVERRIDE_ARGV=($OVERRIDE_ARGS)"` and pass the real validators — 24 ok, 0 fail, re-run against the updated configs.

**2. Nothing leaked from the reverted plumbing attempt.** Confirmed two ways: `git diff --stat 6b691f128..HEAD -- fme/coupled/` is empty, and grepping the additions in `git diff 8f67d45dd..HEAD -- fme/coupled/` for `step_diagnostics|corrector_diagnostics|stack_diagnostics` returns nothing. The coupled diff against base is byte-for-byte the same shape I reviewed in round one (`batch_data.py` +13, `gridded_data.py` +7, `test_batch_data.py` +62, `inference/evaluator.py` +22/-4, `inference/inference.py` +7, `inference/test_evaluator.py` +48, `test_stepper.py` +5) — the seed feature and the `disable_corrections` acceptance test, nothing else. The revert is clean.

**3. The abstract-method change is complete.** All seven `StepConfigABC` subclasses in the repo implement `disable_corrections`: `SingleModuleStepConfig`, `FCN3StepConfig`, `SeparateRadiationStepConfig`, `SecondaryModuleStepConfig` (delegate to their corrector), `MultiCallStepConfig` (delegates to the wrapped step), `StepSelector` (delegates and refreshes `config`), and `MockStepConfig` (raises). I grepped for the subclass pattern across all of `fme/`, not just `fme/core/step`, so nothing in `ace`/`diffusion`/`downscaling` is missed. Verified the abstractness actually binds: `'disable_corrections' in StepConfigABC.__abstractmethods__` is `True`, so a future step type cannot skip it. `pytest fme/core/step fme/ace/stepper fme/coupled fme/core/corrector` is **696 passed, 2 skipped**; ruff, ruff-format and mypy pass on every changed file.

**4. Dropping the coupled correction-delta measurement — not a blocker.** Reasoning in the next comment.

**One thing I fixed beyond the diff: the PR description.** Since it becomes the squash-merge message, its first bullet still claimed the field default is "the 'not applied' value for every option on every corrector config" — which `NON_CORRECTION_OPTIONS` now exists specifically to contradict. Rewrote that bullet to state the rule plus both exclusions and why each is excluded, and noted that `disable_corrections` is abstract on `StepConfigABC`. (`gh pr edit` fails on this repo with a Projects-classic GraphQL deprecation error; the REST `PATCH /repos/ai2cm/ace/pulls/1434` works.)

I also read `tasks/2026-08-13-coupled-step-diagnostics-plumbing.md` on the research repo's `main` — it's landed, and its account of the `i_time_start = n_timesteps_seen` indexing failure matches both your description and the code, so the configs' "plumbing the coupled side is its own task" is a live reference rather than a dangling one.

elynnwu:
Claude: **On whether dropping the coupled correction deltas leaves the campaign able to answer its question — yes, it does. Not a blocker.** But there is one cheap addition I'd make before burning 80 years x 8 arms.

The deltas were *mechanism* evidence, not the identification strategy. The question is whether the emulator's deep-ocean response to `specific_total_water_0` is learned physics or a corrector artifact, and the thing that answers it is the 2x2 itself:

- (baseline, corrector on) vs (q0, corrector on) — the effect under investigation
- (baseline, corrector off) vs (q0, corrector off) — does the effect survive without the corrector?

If the q0-driven deep-ocean response collapses in the corrector-off pair, the corrector is implicated; if it survives, it isn't. That contrast is measured entirely from the retained monthly fields (`thetao_9`, `thetao_18`, `ocean_heat_content`, `sst`), all of which are still written. Nothing about the causal test depended on the per-step delta file.

The loss is also smaller than it looks for the two ocean arms, because the corrector-modified fields happen to be in the retained `names`:

- `noSEFC`: `_correct_hfds` writes to `hfds_total_area` for this model, and `hfds_total_area` is in the ocean monthly `names`. The arm-vs-arm monthly difference is an integrated measure of what the correction was doing — coarser than a 50-day per-step delta, and confounded with the model's own response, but present.
- `noOHCC`: the correction modifies `thetao_*` and `sst`, and `sst`/`thetao_9`/`thetao_18` plus derived `ocean_heat_content` are all retained.

The real gap is `noTEBC` on the **coupled** side: `_force_conserve_total_energy` acts only on `air_temperature_*`, and the coupled atmosphere `names` has no air-temperature level. So the coupled arms have no handle at all on the field the prime-suspect corrector touches. The uncoupled arms do measure TEBC's magnitude directly, on the same corrector and the same checkpoint lineage — so "how much work is TEBC doing" is answered there, and the coupled arms answer "does it reach the ocean". That division works.

**The suggestion:** add `air_temperature_0` to the coupled configs' atmosphere `names` (both the 80yr and the smoke config).

```diff
   atmosphere:
     save_prediction_files: false
     save_monthly_files: true
     names:
       - surface_temperature
+      - air_temperature_0
       - specific_total_water_0
```

Why it's worth a line:

- It is the field TEBC acts on, and level 0 is the same top layer the perturbed humidity (`specific_total_water_0`) lives on — the layer where a corrector-mediated artifact and the perturbation coincide.
- With TEBC off the atmosphere's total energy budget is open, so temperature drift *is* the signature of the correction's absence. Without any air-temperature field recorded, that drift is invisible in the coupled arms and you would be inferring it from `surface_temperature` alone.
- I confirmed `air_temperature_0` is in the checkpoint's atmosphere `out_names`, so it writes real data rather than tripping the absent-name warning.
- The asymmetry matters: one extra monthly variable costs almost nothing, whereas discovering during analysis that you needed it means re-running 80 model years x 8 arms.

I have **not** applied this — it changes what the campaign records, which is your call rather than a defect I should patch. If you want it, it's the two-line diff above and the arms revalidate unchanged.

One smaller note in the same vein: `ocean_heat_content` and `hfds_total_area` were partly justified by the step-diagnostics file (derived variables never appear in a corrector delta). They're still worth keeping for the monthly files — `ocean_heat_content` is arguably the single most direct readout of the deep-ocean response the investigation is about — so no change needed, just flagging that their rationale is now purely the monthly path.

elynnwu:
Claude: **Verdict: ready for review.**

Delta re-review `6b691f128..60c595492`, plus my `fe65a3e4b`. The round-one blocker is resolved.

**The blocker is properly closed.** The scope decision is the right call — the plumbing genuinely is blocked on two-rate time alignment in the per-realm `InferenceEvaluatorAggregator`, not on the writer side, and that is its own piece of work with its own test burden. What matters for a merge is that the configs no longer request something the code cannot deliver, and they no longer do. The revert is clean: `git diff 6b691f128..HEAD -- fme/coupled/` is empty and no `step_diagnostics`/`corrector_diagnostics`/`stack_diagnostics` addition appears anywhere in the coupled diff against base, so `fme/coupled/` differs from `exper/2026-07-16-jamesd` only by the seed feature and the `disable_corrections` acceptance test. The gap is recorded in `tasks/2026-08-13-coupled-step-diagnostics-plumbing.md` with the actual failure mode, so the config comments point somewhere real.

Dropping the coupled correction-delta measurement does **not** block the campaign. The 2x2 design is the identification strategy and it runs entirely off retained monthly fields; the deltas were mechanism color. Detail in the comment above, along with one non-blocking suggestion I'd take before launch (add `air_temperature_0` to the coupled atmosphere `names` — it is the field TEBC acts on, and the cost asymmetry against a re-run of 80 years x 8 arms is stark). Your call; not a gate.

**Fixed in this pass**

- `fe65a3e4b` — the coupled smoke config's checklist still told the operator to confirm "that the step-diagnostics files appear", eight lines above the new comment explaining that no such files are written. The uncoupled smoke config defers to that checklist, so the stale instruction covered both realms. This was the only orphaned claim; I swept the rest of `q0_ablations/` and no delta, coarsening, or step-diagnostics reference survives on the coupled side.
- PR description — its first bullet still asserted the field default is "the 'not applied' value for every option on every corrector config", which `NON_CORRECTION_OPTIONS` exists to contradict. Rewritten to state the rule plus both exclusions and why each is excluded, and to note `disable_corrections` is now abstract. Worth knowing for next time: `gh pr edit` fails on this repo with a Projects-classic GraphQL deprecation error; `gh api --method PATCH repos/ai2cm/ace/pulls/1434 --input -` works.

**Verified**

- Abstract-method change complete: all seven `StepConfigABC` subclasses repo-wide implement `disable_corrections`, and `'disable_corrections' in StepConfigABC.__abstractmethods__` is `True`, so it binds.
- `pytest fme/core/step fme/ace/stepper fme/coupled fme/core/corrector` — 696 passed, 2 skipped. ruff, ruff-format, mypy pass on every changed file.
- All 24 arm/config combinations tokenize through `evaluate.sh` and pass the real `fme.{coupled,ace}.validate_config` against the updated configs — 24 ok, 0 fail.
- Everything cleared in round one still holds: the cached forcing-name sets cannot go stale, the rebuild-and-reload preserves trained and corrector state, the `CorrectorSelector` `asdict` refresh round-trips, and the seed reaches the whole rollout across memory windows.

The three round-one non-blocking items I did not act on remain open and remain non-blocking: the shared per-component seed (latent until the ocean becomes stochastic), and the unenforced `zero_global_mean_moisture_advection` invariant (a trap for a future arm, not this one). The third — making `StepConfigABC.disable_corrections` abstract — is done.

Good to go to elynn.

elynnwu:
Claude: Taken — applied in `a43dac033`. You're right that the coupled `noTEBC` arms had no handle on the field the correction touches, and the cost asymmetry against re-running 80 years x 8 arms decides it. Added `air_temperature_0` to both coupled configs' atmosphere monthly names, with a comment giving the reason (it is the layer the prescribed humidity lives on, and an open energy budget shows up as temperature drift). All 16 coupled arm/config combinations revalidate.

Agreed on the rest: the 2x2 is the identification strategy and it runs off the retained monthly fields, so the coupled correction-delta gap is a loss of mechanism color rather than of the causal test. `ocean_heat_content` and `hfds_total_area` stay for the monthly path, as you say.

Thanks for catching the stale smoke checklist and the PR-description claim — both were exactly the kind of thing that survives an author's own reread.

PR #1435, opened 2026-08-13
Restructure the shutdown path so the policy layer reads without systems expertise

mcgibbon:
The shutdown path from #1425 interleaves three concerns in one file, and only one of them needs low-level expertise. This PR separates them so a reader fluent in signals and distributed training — but not threads, pipes, or file descriptors — can review the termination policy and later modify it (step order, timeouts, callback conditions) without touching the delivery mechanism. Structure and naming carry the load; no behavior change is intended. The only added prose is three one-time definitions the naming cannot carry alone: what a "listener" is (`shutdown.py`), what an "fd" is, and what `Thread.join` does (`_signal_listener.py`).

- The signal-delivery plumbing (pipe, wakeup fd, no-op routing handler, at-fork disarm) moves to `fme.core.distributed._signal_listener.SignalListener`, whose contract is one sentence: run a callback on a dedicated thread when one of the given signals arrives, whatever the main thread is doing. The module stays explicit for a reader who does know those concepts.
- `fme.core.distributed.shutdown` keeps only policy: `_abort_and_exit` is a linear sequence of named steps (`_abort_local_communicators`, `_run_post_abort_callbacks`, `_wait_for_peer_aborts`, then `os._exit`), each carrying its invariant in its name and docstring.
- The two cross-thread flags become an `_ExitCoordination` object whose method names state the facts they record — `mark_listener_owns_exit()`, `wait_until_training_state_frozen(timeout)` — so no policy code outside that wrapper touches `threading.Event`.

Changes:
- `fme.core.distributed._signal_listener` (new): `SignalListener`, plus the at-fork disarm and the routing handler (renamed from `_ignore_signal` to `_route_to_listener`; it routes the signal to the listener rather than ignoring it).
- `fme.core.distributed.shutdown`: policy only; `handle_termination_signals` renamed to `abort_and_exit_on_termination` (the name now states the outcome), `park_timeout`/`DEFAULT_PARK_TIMEOUT` to `state_freeze_timeout`/`DEFAULT_STATE_FREEZE_TIMEOUT`. The API other modules import (`add_post_abort_callback`, `write_stderr`, `TERMINATION_SIGNALS`) is unchanged.
- `Distributed.has_spatial_parallelism` (new property): replaces world-size arithmetic in `require_no_spatial_parallelism`, in the Trainer's restart-checkpoint guard (which now reads as the checkpoint-safety condition it is), and in three test sites that spelled the same predicate by hand.
- Tests: renames and the three predicate conversions above; the suite's assertions are unchanged.

- [x] Tests added (no new functionality; the existing suite pins behavior across the restructure)

### DIFF
--- a/fme/core/distributed/_signal_listener.py
+++ b/fme/core/distributed/_signal_listener.py
@@ -1,0 +1,178 @@
+"""Deliver signals to a dedicated thread, whatever the main thread is doing.
+
+A Python signal handler runs only when the main thread returns to the
+interpreter, and a preempted rank's main thread is typically blocked inside a
+collective's stream sync, so a handler could never run in time. Instead,
+CPython's C-level handler writes each delivered signal number to a wakeup fd
+(``signal.set_wakeup_fd``; an fd, or file descriptor, is the integer handle
+the OS gives an open file or pipe). `SignalListener` points that fd at a pipe
+and reads it from its own thread, so the signal is observed regardless of
+what the main thread is doing.
+
+This module is pure delivery mechanism: what to *do* about a termination
+signal is `fme.core.distributed.shutdown`'s concern.
+"""
+
+import os
+import signal
+import threading
+import types
+from collections.abc import Callable
+from typing import Any
+
+# what signal.getsignal returns and signal.signal accepts
+_SignalHandler = (
+    Callable[[int, types.FrameType | None], Any] | int | signal.Handlers | None
+)
+
+# tells the sentinel apart from signal numbers, which are all positive
+_STOP_LISTENING = 0
+
+# The wakeup fd and signal dispositions are process-global, so at most one
+# listener is active, and the at-fork hook must be able to find it.
+_active_listener: "SignalListener | None" = None
+_at_fork_registered = False
+
+
+def _route_to_listener(signum: int, frame: types.FrameType | None) -> None:
+    # exists so CPython delivers the signal (writing it to the wakeup fd)
+    # instead of taking the default action; the listener thread does the work
+    pass
+
+
+def _disarm_in_child() -> None:
+    """Undo the parent's signal setup in a forked child.
+
+    A fork-started DataLoader worker inherits the routing handler and the
+    wakeup fd, which still feeds the parent's pipe: a signal delivered to the
+    worker would masquerade as the parent's own. Restore the defaults so the
+    child dies as it would have without the inheritance. This runs in the
+    child's main thread -- fork's surviving thread is re-designated as such
+    before the at-fork hooks run -- so the signal module cooperates.
+    """
+    global _active_listener
+    listener = _active_listener
+    if listener is None:
+        return
+    for sig in listener._signals:
+        signal.signal(sig, signal.SIG_DFL)
+    signal.set_wakeup_fd(-1)
+    os.close(listener._read_fd)
+    os.close(listener._write_fd)
+    # clear the state, or a second-generation fork (dataset code forking a
+    # worker's own subprocess) re-runs this on fd numbers the worker has
+    # since reused
+    _active_listener = None
+
+
+class SignalListener:
+    """Run a callback on a dedicated thread when one of ``signals`` arrives.
+
+    ``on_signal`` runs on the listener's own thread, so it must not assume the
+    main thread's cooperation -- and if it never returns (it may end the
+    process), the thread never finishes. While the listener is armed the
+    process's own disposition for ``signals`` is a no-op handler: nothing
+    happens on delivery except the callback.
+
+    The lifecycle is ``start()``, then exactly one of two teardowns, the
+    caller choosing by whether a signal arrived:
+
+    - no signal: ``request_stop()``, ``wait_until_finished()``, ``dismantle()``
+      to restore the previous signal setup;
+    - signal arrived: the listener stays armed -- so further signals stay
+      no-ops -- and ``block_until_process_exit()`` parks the calling thread
+      until ``on_signal`` ends the process.
+    """
+
+    def __init__(
+        self,
+        signals: tuple[signal.Signals, ...],
+        on_signal: Callable[[signal.Signals], None],
+    ) -> None:
+        self._signals = signals
+        self._on_signal = on_signal
+        self._read_fd = -1
+        self._write_fd = -1
+        self._previous_handlers: dict[signal.Signals, _SignalHandler] = {}
+        self._previous_wakeup_fd = -1
+        self._thread = threading.Thread(
+            target=self._read_pipe_until_signal_or_stop,
+            name="termination-listener",
+            daemon=True,
+        )
+
+    def start(self) -> None:
+        """Arm the listener: point the wakeup fd at a fresh pipe, route
+        ``signals`` to it, and start the thread that reads it. Main thread
+        only (only it may install signal handlers).
+        """
+        global _active_listener, _at_fork_registered
+        if not _at_fork_registered:
+            os.register_at_fork(after_in_child=_disarm_in_child)
+            _at_fork_registered = True
+        self._read_fd, self._write_fd = os.pipe()
+        os.set_blocking(self._write_fd, False)
+        self._thread.start()
+        self._previous_handlers = {sig: signal.getsignal(sig) for sig in self._signals}
+        self._previous_wakeup_fd = signal.set_wakeup_fd(
+            self._write_fd, warn_on_full_buffer=False
+        )
+        for sig in self._signals:
+            signal.signal(sig, _route_to_listener)
+        _active_listener = self
+
+    def _read_pipe_until_signal_or_stop(self) -> None:
+        while True:
+            data = os.read(self._read_fd, 64)
+            # the wakeup fd sees every signal CPython delivers (torch's
+            # DataLoader installs a SIGCHLD handler, for one); act only on
+            # ours. Checked before the sentinel: a signal racing the stop
+            # request must win, or the process walks on with its protection
+            # already removed.
+            for sig in self._signals:
+                if sig in data:
+                    self._on_signal(sig)
+                    return
+            if not data or _STOP_LISTENING in data:
+                return
+
+    def request_stop(self) -> None:
+        """Ask the listener thread to finish, assuming no signal arrived.
+
+        A signal already delivered but not yet read still wins over the stop
+        request. One delivered between the thread consuming the sentinel and
+        ``dismantle()`` restoring the dispositions is dropped; the window is
+        microseconds wide and the process is exiting anyway.
+        """
+        global _active_listener
+        _active_listener = None
+        os.write(self._write_fd, bytes([_STOP_LISTENING]))
+
+    def wait_until_finished(self, timeout: float) -> None:
+        """Block up to ``timeout`` seconds for the listener thread to finish
+        (``Thread.join``: wait for a thread to end).
+        """
+        self._thread.join(timeout=timeout)
+
+    def block_until_process_exit(self) -> None:
+        """Park the calling thread until ``on_signal`` ends the process.
+
+        Never returns: the listener thread is running an ``on_signal`` that
+        exits, and joining a thread that never finishes blocks forever. The
+        process's external kill (the scheduler's SIGKILL) is the backstop if
+        ``on_signal`` hangs.
+        """
+        self._thread.join()
+
+    def dismantle(self) -> None:
+        """Restore the pre-``start()`` signal setup and close the pipe.
+
+        Only after ``wait_until_finished`` confirms no signal arrived: this
+        removes the process's protection, and closes fds the listener thread
+        would otherwise still be reading.
+        """
+        for sig, handler in self._previous_handlers.items():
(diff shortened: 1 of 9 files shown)


mcgibbon:
Claude (pre-review agent): pre-review of this PR, run cold in an isolated worktree by an agent that did not author it. Comment-only — nothing was pushed to the branch.

## Audit passes

- **Behavior preservation vs `feature/nccl-abort-probe`** ✅ — traced the whole path statement by statement; no semantic drift found. Specifically verified: arm order (`register_at_fork` → `os.pipe` → `set_blocking` → thread start → `getsignal` → `set_wakeup_fd` → `signal.signal` → mark active) is unchanged; the listener body's order (`mark_listener_owns_exit` before the stderr message, abort, callbacks, grace sleep, `os._exit(128+signum)`) is unchanged, including the `if not _post_abort_callbacks: return` guard and the identical skip message; the signal-before-sentinel priority in the read loop is preserved with the same first-match-in-`signals`-order semantics; the two-teardown asymmetry in `finally` is preserved (sentinel + de-register unconditionally, then bounded join only when the listener does not own the exit, then re-check and either park forever or restore-and-close); fd lifecycle is unchanged (normal exit closes both in `dismantle()`, terminating path closes nothing, fork child closes both). Base's two guards `_armed` / `_pipe_fds is not None` collapse to the single `_active_listener is None` check — equivalent, since both were always set together. Base cleared `_armed` a couple of statements earlier in `finally` than `request_stop()` now clears `_active_listener`; the only consumer is the at-fork hook, so no observable difference.
- **Readability goal in the policy file** ✅ — `shutdown.py` contains no `join`, no fd, no pipe, no `signal.signal`, no `register_at_fork`. The only `threading.Event` is inside `_ExitCoordination` (see the open question below). "fd" is defined once, at first use, in `_signal_listener.py`; "listener" is defined once, in `shutdown.py`.
- **Moved comments/docstrings correct in their new location** ⚠️ — three mismatches, F1/F3/F4 below.
- **Differential tests** ✅ — the test diff is renames only, as claimed; the base suite already pins each invariant the restructure could break (repeated-signal, released-main-thread-cannot-exit-first, disposition restore, second-generation fork fd recycling).
- **Coverage / stragglers** ✅ for the renames (no `handle_termination_signals`, `park_timeout`, `_ignore_signal`, `_pipe_fds`, `_armed` left anywhere in the repo); ⚠️ for the new property, F7 below.
- **Lint / types / tests** ✅ — `ruff check` and `ruff format --check` clean; `mypy` reports nothing in the three changed modules; `FME_FORCE_CPU=1 pytest fme/core/distributed/test_shutdown.py test_shutdown_dataloader.py` → 19 passed.
- **PR description accuracy** ⚠️ — F8 below.
- **Structure** ✅ — one commit, one reason, 7 files. No split warranted.

## Findings

**F1 — `_signal_listener.py:77-84`, medium.** The class docstring says the caller performs "exactly one of two teardowns", listing `request_stop()` only on the no-signal path. The caller (`shutdown.py:242`) calls `request_stop()` unconditionally, before it knows which path it is on. This is load-bearing rather than cosmetic: `request_stop()` is where `_active_listener = None` happens, and that is what preserves the base branch's `_armed = False` at the top of `finally` — the at-fork hook stops finding the listener on *both* paths. A reviewer checking the two-teardown asymmetry (the thing this PR most needs checked) hits a docstring that contradicts the code. Suggest: state that `request_stop()` always runs, note that on the signal path its sentinel goes into a pipe nobody reads again and its real effect is the de-registration, then branch on `wait_until_finished()` + `dismantle()` vs `block_until_process_exit()`.

**F2 — `_signal_listener.py:139-149`, medium.** `request_stop()`'s docstring documents the sentinel race but never mentions that it clears `_active_listener`, i.e. that it is the at-fork disarm's off-switch. Consequence worth naming: between `request_stop()` and `dismantle()`, a fork inherits the routing handler and a wakeup fd still pointing at the parent's open pipe, and `_disarm_in_child` no-ops. This hole exists identically on the base branch, so it is not a regression — but the restructure gave the step a name, so the name should carry the consequence.

**F3 — `shutdown.py:51-52`, low.** The `DEFAULT_STATE_FREEZE_TIMEOUT` comment says "see _ExitCoordination.training_state_frozen". No such name exists — the class has `_training_state_frozen` (private), `mark_training_state_frozen()`, and `wait_until_training_state_frozen()`. Point it at `wait_until_training_state_frozen`.

**F4 — `_signal_listener.py:73-75`, low.** "nothing happens on delivery except the callback" reads as though every delivery runs the callback. `_read_pipe_until_signal_or_stop` returns after the first match, so the callback runs once and every later delivery is a true no-op — which is exactly the "a repeated signal cannot cut the grace period short" invariant (`shutdown.py:202-204`, `test_a_repeated_signal_cannot_cut_the_grace_period_short`). Say "once, on the first matching signal".

**F5 — `_signal_listener.py:43-65`, low.** `_disarm_in_child` reaches into `listener._signals`, `listener._read_fd`, `listener._write_fd` from module scope. Nothing breaks (same module), but the class docstring presents `dismantle()` as the thing that closes the pipe, so the third fd-closing path is invisible from the class. Either add a line to the class docstring or make it a method the hook calls, so "who closes what" is answerable from one place.

**F6 — `distributed.py:215-220`, low.** The new property's docstring says "world_size > total_data_parallel_ranks" while the body is `!=`. Equivalent in practice, but this is now a stated contract rather than an inline comment.

**F7 — coverage, low.** Three sites still spell the same predicate as arithmetic: `fme/ace/data_loading/test_data_loader.py:1360` and `:1397` (`world_size != total_data_parallel_ranks`), and `fme/core/models/conditional_sfno/test_sfnonet.py:497` (`world_size == total_data_parallel_ranks`, a skip guard). All tests. Either convert them or note in the description that the property was scoped to production call sites.

**F8 — PR description, low.**
- "no explanatory comments were added" is not accurate: the PR deliberately adds the definition of "listener" (`shutdown.py:16-18`), the definition of "fd" (`_signal_listener.py:7-8`), and the `Thread.join` gloss (`_signal_listener.py:153`). Those additions are the point of the PR, but the sentence invites a reviewer to distrust the diff on meeting them.
- "policy code never touches `threading.Event` directly" — `shutdown.py:102-103` constructs two of them, inside `_ExitCoordination`, in the policy file. True as intended; worth qualifying to "outside `_ExitCoordination`".
- `DEFAULT_PARK_TIMEOUT` → `DEFAULT_STATE_FREEZE_TIMEOUT` is a module-level rename not mentioned (only the parameter rename is). Nothing in the repo imports it, so it is harmless, but "Public API unchanged" reads as covering constants.

## Open question

`_ExitCoordination` lives in the policy file and is the only `threading.Event` there. Given the stated audience, is the class-as-shield enough, or should it move to the plumbing module? My read: it should stay — the facts it records (who owns the exit; is training state frozen) are policy, and moving them would split the policy story across two files. Flagging it because it is the one judgment call in the restructure.

## Assessment

The PR does what it says. The policy file now reads as a linear sequence of named steps with no threading, pipe, or fd vocabulary, and the plumbing module is a self-contained delivery mechanism whose contract fits in a sentence. I found no behavior change. Review effort is low: one commit, a mechanical rename set, and one genuinely new abstraction (`SignalListener`) whose surface is five methods.

**Verdict: ready for review once F1 and F3 are applied** — F1 because it contradicts the code on the invariant a reviewer is most likely to check, F3 because it is a dangling reference. F2 and F4 are cheap and improve the same reading. F5–F8 are optional.

mcgibbon:
Claude: Applied the pre-review findings in 9e613d4:

- **F1/F4/F5**: the `SignalListener` lifecycle docstring now matches the code — `request_stop()` is documented as issued unconditionally (before the caller can know whether a signal arrived), only the first matching signal runs the callback, and the forked child's automatic teardown is named as the third path.
- **F2**: `request_stop()`'s docstring states that it stands down the at-fork disarm and why.
- **F3**: the `DEFAULT_STATE_FREEZE_TIMEOUT` comment now points at `mark_training_state_frozen`.
- **F6**: `has_spatial_parallelism`'s docstring operator matches the body (`!=`).
- **F7**: the three test sites (`test_data_loader.py:1360,1397`, `test_sfnonet.py:497`) now use the property.
- **F8**: PR description amended — the added-prose claim now names the three one-time definitions, the `threading.Event` claim is qualified to "outside `_ExitCoordination`", and the `DEFAULT_PARK_TIMEOUT` constant rename is listed.

On the open question: `_ExitCoordination` stays in the policy file, per the argument in the review — the facts it records are policy.

Checks after the fixes: pre-commit (ruff, ruff-format, mypy) clean on the touched files; 2-rank spatial-parallel run of the F7-touched test files passes.

mcgibbon:
Claude (pre-review agent): delta re-review of 9e613d478 against aeec16a07.

## Findings re-check

All eight are addressed, and I found nothing new introduced.

- **F1** — `request_stop()` is now hoisted out of the branch and onto the common path, which matches `shutdown.py`. The "signal already delivered still wins" clause is the right thing to say there.
- **F2** — `request_stop()` now leads with "and stand down the at-fork disarm" and names `_disarm_in_child` in the body.
- **F3** — resolves now. You pointed at `mark_training_state_frozen` where I had suggested `wait_until_training_state_frozen`; `mark_` is the better choice, since the phrase a reader is chasing is "stop touching training state", which is what that method's docstring defines.
- **F4** — "Only the first matching signal runs it: the thread reads nothing further" is exactly the invariant.
- **F5** — the third-teardown paragraph puts the fork path in the class contract, and "the child's copy of the pipe" is precise about what gets closed.
- **F6** — operator matches the body.
- **F7** — all three sites read correctly, including the negation at `test_sfnonet.py:497` (`==` becomes `not ...`). The two `test_data_loader.py` guards are straight substitutions of an identical predicate.
- **F8** — the added-prose sentence now names all three definitions, the Event claim is scoped to "outside that wrapper", `DEFAULT_PARK_TIMEOUT` is listed, and "Public API" became "The API other modules import", which is the accurate claim.

`ruff check` and `ruff format --check` clean on all changed files. The delta touches only docstrings, one comment, and three test predicates; no executable logic outside those predicates changed, and each is logically identical to what it replaced.

## Blocker: the PR needs a rebase before it can be reviewed

Not your doing — #1425 merged while this was open. It was **squash**-merged, so `d0428b074` is not an ancestor of `main`. GitHub retargeted this PR's base from `feature/nccl-abort-probe` to `main`, and the result is that it now reports `CONFLICTING` and its Files-changed view lists **15 files instead of 7** — it has absorbed all of #1425 (`base.py`, `non_distributed.py`, `torch_distributed.py`, `model_torch_distributed.py`, `test_distributed.py`, `test_trainer.py`). Reviewing the PR as GitHub currently presents it means re-reviewing #1425 on top of a conflicted diff.

The fix is mechanical and I verified it is content-preserving: `d0428b074` and `origin/main` (6f44c926b) have **byte-identical trees**, so replaying just the two restructure commits is clean.

```
git rebase --onto origin/main d0428b074 refactor/shutdown-readability
git push --force-with-lease
```

I ran that on a throwaway branch: it rebased both commits with no conflicts and produced a tree identical to 9e613d478's. Nothing else needs to change.

## Two optional nits

- `_signal_listener.py:150-154` — "stops `_disarm_in_child` from acting in children forked after the teardown has begun, when the fds it would close may already be gone" is true once `dismantle()` has run. In the window between `request_stop()` and `dismantle()` the fds are still open, and the effect there is the opposite one: a child forked in that window keeps the routing handler and a wakeup fd feeding the parent's live pipe, and is not disarmed. Identical to the base branch's behaviour, so nothing to fix — but if you want the sentence to cover the whole window, that is the half it currently leaves out.
- "Tests: renames only; the suite's assertions are unchanged" now sits slightly behind the three `has_spatial_parallelism` conversions. The bullet above it already names them, so this is only about the word "only".

## Assessment

The restructure is done and the docstrings now say what the code does, which was the one thing standing between a reviewer and the invariants this PR asks them to check. Review effort is low once the diff isolates the change again: two commits, a mechanical rename set, and one new five-method abstraction.

**Verdict: ready for review once rebased onto `origin/main`.** The code is finished; the rebase is the only thing in the way, and it is a one-liner with a verified-clean result.

fme/core/distributed/_signal_listener.py
```diff
@@ -0,0 +1,187 @@
+"""Deliver signals to a dedicated thread, whatever the main thread is doing.
```
Arcomano1234, line 1: Question: Do we want to have this module be private? I can't remember but I feel like we rarely have private modules in ACE. I don't really have a strong opinion and defer to you but just wanted to raise this in case it was not intentional.

Arcomano1234 (approved):
Approved with just a single comment / question. The tests are failing but don't seem to related to changes in this PR seems we are still having flaky tests or something strange with pip on the CI.

fme/core/distributed/_signal_listener.py
mcgibbon: It's marked private mainly to signal "hey you probably don't need to read this file". It intentionally hides code that would be confusing for a scientist.

I'll leave it since you don't have a strong opinion though I also don't have a strong opinion.

PR #1436, opened 2026-08-13
Add 6h-mean surface channels to ERA5 ingest pipeline

mcgibbon:
Add 8 daily-mean surface channels to the ERA5 ingest pipeline. We need
daily means of surface temperature, pressure, 2m temperature/dewpoint/humidity,
and 10m winds as prediction targets; ERA5 archives no time-integrated means
for instantaneous fields, so a 24-hourly-snapshot mean built by our own
pipeline is the best obtainable daily mean.

Changes:
- `scripts/era5/pipeline/xr-beam-pipeline.py`: add Stream 5 — reads
  `skin_temperature`, `2m_temperature`, `2m_dewpoint_temperature`,
  `surface_pressure`, and `10m_u/v_component_of_wind` hourly from
  ARCO-ERA5 over `flux_time_slice`, regrids each hourly stamp to the
  target Gaussian grid, derives `Q2m` (from regridded dewpoint + pressure)
  and wind speed (√(u²+v²)) per stamp, then averages the 6-stamp
  [T−5h, T] window via `_average_hourly_to_6hourly`. Outputs 8 channels:
  `surface_temperature_mean`, `PRESsfc_mean`, `TMP2m_mean`, `DPT2m_mean`,
  `Q2m_mean`, `UGRD10m_mean`, `VGRD10m_mean`, `WIND10m_mean`.
  `DESIRED_ATTRS` extended with long_name and units for each new channel;
  `_make_template` processes one 6h chunk of the new stream.
- `scripts/era5/Makefile`: bump all three `OUTPUT_PATH_*` variables to
  `2026-08-13` dates to match the config `runs:` keys.
- `scripts/data_process/configs/era5-{1,2,4}deg-8layer-1940-2025.yaml`:
  add the 8 new `*_mean` channels to `time_coarsen.window_names` (so the
  factor-4 daily coarsening averages them). All dated output paths bumped
  to `2026-08-13` (runs, data_output_directory, stats_output_directory,
  stats.output_directory, stats.beaker_dataset) because each stage's
  skip-if-exists guard would silently no-op on reused names.

- [ ] Tests added

### DIFF
--- a/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
@@ -1,5 +1,5 @@
 runs:
-  2026-08-13-era5-4deg-8layer-1940-2025: ""
+  2026-08-13-era5-4deg-8layer-1940-2025: ""  # no real data source, config only for computing stats
 data_output_directory: gs://vcm-ml-intermediate
 stats:
   output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-stats-1990-2019
--- a/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
@@ -88,6 +88,7 @@ time_coarsen:
     - DSWRFtoa
     - LHTFLsfc
     - PRATEsfc
+    - PRESsfc_mean
     - Q2m_mean
     - SHTFLsfc
     - TMP2m_mean
--- a/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
@@ -1,16 +1,16 @@
 runs:
-  2026-03-19-era5-4deg-8layer-1940-2025: ""  # no real data source, config only for computing stats
+  2026-08-13-era5-4deg-8layer-1940-2025: ""
 data_output_directory: gs://vcm-ml-intermediate
 stats:
-  output_directory: gs://vcm-ml-intermediate/2026-03-19-era5-4deg-8layer-stats-1990-2019
-  beaker_dataset: 2026-03-19-era5-4deg-8layer-stats-1990-2019
+  output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-stats-1990-2019
+  beaker_dataset: 2026-08-13-era5-4deg-8layer-stats-1990-2019
   start_date: "1990-01-01"
   end_date: "2019-12-31"
   data_type: ERA5
 time_coarsen:
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-1940-2025
-  stats_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-stats-1990-2019
+  data_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-1940-2025
+  stats_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-stats-1990-2019
   snapshot_names:
     - DPT2m
     - PRESsfc
@@ -83,16 +83,23 @@ time_coarsen:
     - surface_temperature
   window_names:
     - DLWRFsfc
+    - DPT2m_mean
     - DSWRFsfc
     - DSWRFtoa
     - LHTFLsfc
     - PRATEsfc
+    - Q2m_mean
     - SHTFLsfc
+    - TMP2m_mean
+    - UGRD10m_mean
     - ULWRFsfc
     - ULWRFtoa
     - USWRFsfc
     - USWRFtoa
+    - VGRD10m_mean
+    - WIND10m_mean
     - global_mean_co2
+    - surface_temperature_mean
     - tendency_of_total_water_path_due_to_advection
     - total_frozen_precipitation_rate
   constant_prefixes:
--- a/scripts/era5/pipeline/xr-beam-pipeline.py
+++ b/scripts/era5/pipeline/xr-beam-pipeline.py
@@ -228,39 +228,32 @@ DESIRED_ATTRS = {
         "units": "kg/m**2",
     },
     "surface_temperature_mean": {
-        "long_name": "6-hourly mean skin temperature",
+        "long_name": "Mean skin temperature",
         "units": "K",
-        "averaging_window": "[T-5h, T]",
     },
     "TMP2m_mean": {
-        "long_name": "6-hourly mean 2m air temperature",
+        "long_name": "Mean 2m air temperature",
         "units": "K",
-        "averaging_window": "[T-5h, T]",
     },
     "DPT2m_mean": {
-        "long_name": "6-hourly mean 2m dewpoint temperature",
+        "long_name": "Mean 2m dewpoint temperature",
         "units": "K",
-        "averaging_window": "[T-5h, T]",
     },
     "Q2m_mean": {
-        "long_name": "6-hourly mean 2m specific humidity",
+        "long_name": "Mean 2m specific humidity",
         "units": "kg/kg",
-        "averaging_window": "[T-5h, T]",
     },
     "UGRD10m_mean": {
-        "long_name": "6-hourly mean 10m U component of wind",
+        "long_name": "Mean 10m U component of wind",
         "units": "m/s",
-        "averaging_window": "[T-5h, T]",
     },
     "VGRD10m_mean": {
-        "long_name": "6-hourly mean 10m V component of wind",
+        "long_name": "Mean 10m V component of wind",
         "units": "m/s",
-        "averaging_window": "[T-5h, T]",
     },
     "WIND10m_mean": {
-        "long_name": "6-hourly mean 10m wind speed",
+        "long_name": "Mean 10m wind speed",
         "units": "m/s",
-        "averaging_window": "[T-5h, T]",
     },
     **{
         f"{soil_type}_soil_type_fraction": {
@@ -684,12 +677,7 @@ def process_mean_flux(
 
 
 def _process_surface_mean(ds: xr.Dataset, output_grid: str) -> xr.Dataset:
-    """Regrid surface fields hourly, derive Q2m and wind speed, then average.
-
-    Processing order: regrid each hourly stamp → derive nonlinear channels
-    (Q2m from dewpoint+pressure, wind speed from u+v) at the target grid →
-    average the 6 stamps → drop intermediates → rename to *_mean output names.
-    """
+    """Regrid surface fields hourly, derive Q2m and wind speed, then average."""
     xr.set_options(keep_attrs=True)
     regridded = _regrid(ds, output_grid)
     regridded = regridded.drop_vars(["latitude", "longitude"])
@@ -1360,8 +1348,7 @@ def main():
                 output_grid=args.output_grid,
                 check_data_validity=args.check_data_validity,
             )
-            | "surface_mean_ConsolidateChunks"
-            >> xbeam.ConsolidateChunks(output_shards)
+            | "surface_mean_ConsolidateChunks" >> xbeam.ConsolidateChunks(output_shards)
             | "surface_mean_to_zarr"
             >> xbeam.ChunksToZarr(
                 output_store,
--- a/scripts/era5/pipeline/xr-beam-pipeline.py
+++ b/scripts/era5/pipeline/xr-beam-pipeline.py
@@ -239,6 +239,10 @@ DESIRED_ATTRS = {
         "long_name": "Mean 2m dewpoint temperature",
         "units": "K",
     },
+    "PRESsfc_mean": {
+        "long_name": "Mean surface pressure",
+        "units": "Pa",
+    },
     "Q2m_mean": {
         "long_name": "Mean 2m specific humidity",
         "units": "kg/kg",
@@ -693,6 +697,7 @@ def _process_surface_mean(ds: xr.Dataset, output_grid: str) -> xr.Dataset:
 
     output = xr.Dataset()
     output["surface_temperature_mean"] = regridded["skin_temperature"]
+    output["PRESsfc_mean"] = regridded["surface_pressure"]
     output["TMP2m_mean"] = regridded["2m_temperature"]
     output["DPT2m_mean"] = regridded["2m_dewpoint_temperature"]
     output["Q2m_mean"] = q2m
--- a/scripts/era5/pipeline/xr-beam-pipeline.py
+++ b/scripts/era5/pipeline/xr-beam-pipeline.py
@@ -119,6 +119,15 @@ FULL_37_MODEL_LEVEL_SURFACE_VARS = [
     "10m_v_component_of_wind",
 ]
 
+FULL_37_SURFACE_MEAN_VARS = [
+    "skin_temperature",
+    "2m_temperature",
+    "2m_dewpoint_temperature",
+    "surface_pressure",
+    "10m_u_component_of_wind",
+    "10m_v_component_of_wind",
+]
+
 # Soil type definitions from the ECMWF documentation: https://codes.ecmwf.int/grib/param-db/43
 # undefined is not part of the defintions, but it appears to be the fill value.
 # Some cells with land_fraction > 0 have this value, so it still seems relevant
@@ -217,6 +226,41 @@ DESIRED_ATTRS = {
     "surface_snow_amount": {
         "long_name": "Surface snow amount",
         "units": "kg/m**2",
+    },
+    "surface_temperature_mean": {
+        "long_name": "6-hourly mean skin temperature",
+        "units": "K",
+        "averaging_window": "[T-5h, T]",
+    },
+    "TMP2m_mean": {
+        "long_name": "6-hourly mean 2m air temperature",
+        "units": "K",
+        "averaging_window": "[T-5h, T]",
+    },
+    "DPT2m_mean": {
+        "long_name": "6-hourly mean 2m dewpoint temperature",
+        "units": "K",
+        "averaging_window": "[T-5h, T]",
+    },
+    "Q2m_mean": {
+        "long_name": "6-hourly mean 2m specific humidity",
+        "units": "kg/kg",
+        "averaging_window": "[T-5h, T]",
+    },
+    "UGRD10m_mean": {
+        "long_name": "6-hourly mean 10m U component of wind",
+        "units": "m/s",
+        "averaging_window": "[T-5h, T]",
+    },
+    "VGRD10m_mean": {
+        "long_name": "6-hourly mean 10m V component of wind",
+        "units": "m/s",
+        "averaging_window": "[T-5h, T]",
+    },
+    "WIND10m_mean": {
+        "long_name": "6-hourly mean 10m wind speed",
+        "units": "m/s",
+        "averaging_window": "[T-5h, T]",
     },
     **{
         f"{soil_type}_soil_type_fraction": {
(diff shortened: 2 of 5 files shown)


mcgibbon:
Claude: Pre-review of PR #1436

## Audit passes

1. **Silent-failure** -- clean. The pipeline additions introduce no silent-failure paths; variable lists are fixed, output datasets are built explicitly, and DESIRED_ATTRS matching is defensive.

2. **Behavior-change on existing code paths** -- findings in the shutdown refactoring (see inline comments):

   (a) `Distributed.context()` now installs signal handlers for single-process (non-distributed) jobs. Previously gated by `instance.is_distributed()`. Intentional change (single-GPU jobs now get graceful checkpoint saves on preemption), but undocumented in the PR description and the deleted test `test_context_leaves_signals_alone_when_not_distributed` had no replacement.

   (b) `instance.shutdown()` on normal exit now runs OUTSIDE the ExitStack, after signal handlers have been restored to SIG_DFL. The old code's comment said "inside the stack: the teardown is itself a collective, so it must run while the listener still protects the process." If SIGTERM arrives during the unprotected `destroy_process_group` collective on normal exit, the process dies immediately (SIG_DFL), peers hang. The old listener-thread design protected this window.

   (c) The domain-parallel guard on restart-checkpoint callback registration is removed. Old code: `if dist.world_size == dist.total_data_parallel_ranks: add_post_abort_callback(...)`. New code: always registers. A domain-parallel setup would try to save sharded model state after the communicators are already destroyed.

3. **Differential-test** -- clean. The rewritten tests properly cover the new signal-handler design. Subprocess tests are used where the process must exit; in-process `pytest.raises(SystemExit)` is used where it survives.

4-7. **Record passes** -- N/A (ace code PR, not research records).

8. **AI-agent artifact** -- the PR description describes only the ERA5 pipeline changes (4 files, ~300 lines). It does not mention the distributed shutdown refactoring (10+ files, ~1700 lines), which is the majority of the diff by volume.

9. **Split heuristic** -- the PR bundles two independent concerns with no dependency between them:
   - ERA5 ingest pipeline: 7 new daily-mean surface channels (4 files)
   - Distributed shutdown architecture: abort-to-shutdown refactor (10+ files)

   These should be separate PRs. The shutdown refactoring is a significant architectural change (replacing listener-thread + ncclCommAbort + grace-period with main-thread handler + collective destroy_process_group + watchdog timer) that deserves its own review.

10. **Writing style** -- clean. Pipeline docstrings and the shutdown module docstring are precise and well-structured.

## ERA5 pipeline assessment

The pipeline code itself is correct:
- Processing order (regrid -> derive nonlinear channels -> average) is right for Q2m (nonlinear function of dewpoint + pressure) and wind speed (sqrt(u^2+v^2)).
- Stream wiring mirrors Stream 1 (flux): `flux_time_slice`, `chunks={"time": 6}`, offset conversion via `key.offsets["time"] // TIME_STEP`.
- Intermediate `surface_pressure` is correctly excluded from the output dataset (only the 7 named variables are added to the new Dataset).
- DESIRED_ATTRS entries match the 7 output variable names exactly.
- Template extension correctly processes one 6h chunk via `_process_surface_mean`.
- All 3 configs: all 5 dated paths bumped to 2026-08-13; 7 new channels added to `window_names` (not `snapshot_names`); alphabetically sorted; no overlap with `snapshot_names`.

## Verdict

Not ready for review. The PR should be split:

1. **Shutdown refactoring** as its own PR, with the behavior changes (single-process handler installation, unprotected normal-exit shutdown, domain-parallel guard removal) documented and reviewed on their merits.
2. **ERA5 daily-mean surface channels** as a clean, self-contained pipeline PR that is straightforward to review.

The pipeline code is ready; the bundling is what blocks this.

mcgibbon:
Claude: **Correction to the pre-review above** -- the distributed shutdown findings (passes 2b, 2c, 8, 9) are retracted. Those changes are on main, not part of this PR. The two-dot `git diff origin/main..origin/feature/...` I used initially included main's history merged into the branch; the three-dot diff (and `gh pr diff`) confirms this PR touches only the 4 pipeline/config files. Apologies for the noise.

The corrected verdict is below.

## Corrected audit passes

1. **Silent-failure** -- clean.
2. **Behavior-change on existing code paths** -- clean. The pipeline additions are purely additive (new stream, new variables, new DESIRED_ATTRS entries); no existing processing function is modified.
3. **Differential-test** -- N/A. No tests added (the "Tests added" checkbox is unchecked). The pipeline is a script, not library code, so unit tests are not the norm here. No finding.
4-7. **Record passes** -- N/A (ace code PR).
8. **AI-agent artifact** -- clean. PR description accurately describes the 4 changed files and the processing design.
9. **Split heuristic** -- clean. One concern, one PR.
10. **Writing style** -- clean.

## Pipeline assessment (unchanged)

The pipeline code is correct:
- Processing order (regrid -> derive nonlinear channels -> average) is right for Q2m and wind speed.
- Stream wiring mirrors Stream 1: `flux_time_slice`, `chunks={"time": 6}`, offset conversion.
- Intermediate `surface_pressure` excluded from output by construction.
- DESIRED_ATTRS entries match the 7 output variable names exactly.
- Template processes one 6h chunk correctly.
- All 3 configs: 5 dated paths bumped; 7 new channels in `window_names`; alphabetically sorted; no overlap with `snapshot_names`.

## Minor observation (not a blocker)

The `# no real data source, config only for computing stats` comment was dropped from the `runs:` key in all 3 configs. If the comment was still accurate (the run entry is config-only, not a real run), restoring it preserves context for future readers. If the entry will now reference a real data-producing run, dropping it is correct.

## Verdict

Ready for review.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -1239,6 +1350,28 @@ def main():
...
+            p
+            | "surface_mean_DatasetToChunks"
+            >> xbeam.DatasetToChunks(ds_surface_mean, chunks={"time": 6})
+            | beam.MapTuple(
+                process_surface_mean,
+                output_grid=args.output_grid,
+                check_data_validity=args.check_data_validity,
+            )
+            | "surface_mean_ConsolidateChunks"
+            >> xbeam.ConsolidateChunks(output_shards)
```
mcgibbon: Agent: Chore: this is the ruff-format failure making CI pre-commit red — ruff wants the manually-wrapped pipe joined onto one line (88 chars, matching the flux stream's `"mean_flux_ConsolidateChunks" >> xbeam.ConsolidateChunks(output_shards)` above).

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -634,6 +678,63 @@ def process_mean_flux(
...
+# Stream 5: Surface mean (6h mean of hourly-regridded surface fields)
+# ---------------------------------------------------------------------------
+
+
+def _process_surface_mean(ds: xr.Dataset, output_grid: str) -> xr.Dataset:
+    """Regrid surface fields hourly, derive Q2m and wind speed, then average.
+
+    Processing order: regrid each hourly stamp → derive nonlinear channels
+    (Q2m from dewpoint+pressure, wind speed from u+v) at the target grid →
+    average the 6 stamps → drop intermediates → rename to *_mean output names.
```
mcgibbon: Agent: nit: the docstring describes steps the code doesn't perform — there is no "drop intermediates" or "rename to *_mean output names" step (a fresh Dataset is built with the final names, and `surface_pressure` is simply never copied into it), and the attrs assignment happens after averaging, not a rename. Suggest: "regrid each hourly stamp → derive nonlinear channels (Q2m, wind speed) at the target grid → assign `*_mean` output names → average the 6 stamps".

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -119,6 +119,15 @@
     "10m_v_component_of_wind",
 ]
 
+FULL_37_SURFACE_MEAN_VARS = [
+    "skin_temperature",
+    "2m_temperature",
+    "2m_dewpoint_temperature",
+    "surface_pressure",
+    "10m_u_component_of_wind",
+    "10m_v_component_of_wind",
```
mcgibbon, line 128: Agent: Question (non-blocking): `surface_pressure` is already read hourly and regridded per stamp for the Q2m derivation, so emitting a `PRESsfc_mean` channel would be essentially free — one extra line in the output Dataset. Intentional to drop it? If daily-mean pressure is plausibly useful as a target later, adding it now avoids another full ingest; if it was considered and rejected, fine as is.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -218,6 +227,41 @@
         "long_name": "Surface snow amount",
         "units": "kg/m**2",
     },
+    "surface_temperature_mean": {
+        "long_name": "6-hourly mean skin temperature",
+        "units": "K",
+        "averaging_window": "[T-5h, T]",
```
mcgibbon: Agent: nit (metadata only): `time_coarsen.py`'s `coarsen(time=4).mean()` keeps attrs, so these `long_name: "6-hourly mean ..."` / `averaging_window: "[T-5h, T]"` attrs will be carried verbatim into the daily-coarsened product, where they're wrong (there the channels are 24h means over [T-23h, T]). The existing window channels dodge this by not encoding the cadence in their attrs (e.g. `LHTFLsfc` is just "Latent heat flux"). Consider dropping "6-hourly"/the window attr from the long_name, or accept the known-stale metadata — either way it's worth a conscious choice since the daily store is the one most users will read.

mcgibbon (commented):
Agent: Overall this is in good shape — I verified the load-bearing science and pipeline mechanics and they check out:

- Regrid-before-average order is correct and required (Q2m and wind speed are nonlinear, so they must be derived per hourly stamp at the target grid), and deriving Q2m from regridded dewpoint + surface pressure exactly matches Stream 4's snapshot `Q2m` convention.
- `WIND10m_mean` = mean of √(u²+v²) per stamp, as requested (not the speed of the mean wind).
- The `[T-5h, T]` window, `chunks={"time": 6}` alignment against `flux_time_slice` (6N hourly stamps → N chunks whose last stamp is the output time), and `key.offsets["time"] // TIME_STEP` all match the flux stream; the template processes the same first chunk and the merged names collide with nothing.
- Config placement is right: the `*_mean` channels belong in `window_names` (daily mean of 24 hourly snapshots after factor-4 coarsening, labeled consistently with the flux channels), none of the new names overlap `snapshot_names`, and all 7 are listed in all 3 configs (`time_coarsen.py` would KeyError on a missing name, and silently drop an unlisted one — neither applies). Stats cover all variables automatically.
- Performance is fine: ~36 field-regrids per output step vs the flux stream's ~27, and the model-level stream dominates job cost regardless.

Two things to fix before merge, both cheap:

1. **`scripts/era5/Makefile` production paths were not bumped.** The configs' `runs:` keys and all stats/coarsen paths moved to `2026-08-13-*`, but the Makefile still has `OUTPUT_PATH_ONE_DEGREE=gs://vcm-ml-intermediate/2026-03-19-era5-1deg-8layer-1940-2025.zarr` (and `2026-03-19` / `2026-06-23` for 4°/2°). On this branch, `make era5_dataflow` would write the new-channel dataset **into the existing production zarrs** that current training configs read, and the downstream coarsen/stats stages would then look for `2026-08-13-*` inputs that nothing produces. Please bump the three `OUTPUT_PATH_*` variables to `2026-08-13` to match the configs.

2. **CI pre-commit is red on this PR's file** (ruff-format; inline comment below). The `cpu`/`cpu-very-fast-no-healpix`/`build` failures are runner infrastructure (`pip._internal` ImportError during dependency install — `gpu` passed and the PR touches only `scripts/`), not this change.

Please incorporate those two and the optional comments as you see fit; otherwise LGTM — no re-review needed.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Claude: Fixed in 501beb0 — joined to one line.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Claude: Fixed in 501beb0 — trimmed to one line.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Claude: Intentional — the task plan specified this channel set (decided with Jeremy). PRESsfc_mean would be straightforward to add later without a re-ingest since surface_pressure is already read in Stream 4.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Claude: Fixed in 501beb0 — removed averaging_window and cadence prefix from long_name. Now matches the existing window channel convention.

mcgibbon:
Claude: All review findings addressed in 501beb0:
- **Makefile paths** (blocking): bumped all three OUTPUT_PATH variables to 2026-08-13 dates.
- **ruff-format** (blocking): joined ConsolidateChunks line.
- **Docstring** (nit): trimmed to one line.
- **Attrs cadence** (nit): removed averaging_window and '6-hourly' prefix from long_name to match existing window channel convention.
- **PRESsfc_mean** (question): intentionally excluded per the agreed channel set; can be added later without re-ingest.

mcgibbon:
Agent: Second-pass review — verified the 501beb0 fixes, then went deeper on dtypes, regridding, template/chunk mechanics, and the downstream coarsen/stats stages.

**Fix verification** — all four first-round items landed as claimed:
- Makefile `OUTPUT_PATH_{ONE,TWO,FOUR}_DEGREE` now match the config `runs:` keys exactly once `time_coarsen.main()` appends `.zarr` (`data_output_directory + "/" + run_name + ".zarr"` → `gs://vcm-ml-intermediate/2026-08-13-era5-{1,2,4}deg-8layer-1940-2025.zarr`).
- pre-commit CI is green (ruff-format fixed).
- Docstring trimmed; attrs now match the existing window-channel convention (no cadence encoding).

**One new finding — Chore (pre-merge, description only):** the PR description is now stale. It still says

> `DESIRED_ATTRS` extended with long_name/units/averaging_window for each new channel

but 501beb0 removed `averaging_window` (and the "6-hourly" long_name prefix) per review. The description feeds the squash-merge commit, so please drop "averaging_window" from that sentence before merge.

**Deeper checks, all clean (for the record):**
- **dtype consistency**: the new channels will be float32 like the rest of the store. The existing production zarr is float32 for every variable including regrid-then-derive products (`Q2m`), and Stream 5 uses the identical `_regrid` / `_specific_humidity_from_dewpoint` / `.mean("time")` path; `np.sqrt(f32**2 + f32**2)` stays float32. Template and chunks go through the same functions, so they can't disagree.
- **multi-time regrid**: `_regrid` already handles time-bearing inputs in production — streams 2/3 regrid `process_time_chunksize=2` chunks — and xESMF broadcasts over leading dims; the `sortby("lat")` and coord renames are per-dimension and unaffected by the extra time dim. Six stamps is the same code path.
- **template merge**: no name overlap between the `*_mean` channels and any existing variable (`surface_temperature_mean` vs `surface_temperature` are distinct arrays), so `compat="override"` is never engaged for the new names.
- **chunk alignment / write collisions**: both hourly streams chunk the identical time coordinate (same `flux_time_slice` from the same store) with `chunks={"time": 6}`, so DatasetToChunks boundaries coincide; and since the streams write disjoint variable sets, they write disjoint zarr arrays regardless. Offset math mirrors flux, and the hourly slice length (start−5h … end) is exactly 6× the output count, so no partial averaging windows.
- **daily coarsening**: `time_coarsen.coarsen()` silently drops any variable not in one of the three lists, so the `window_names` additions are load-bearing and correctly placed (mean channels → window average, labeled by the snapshot time, giving [T−23h, T] daily means consistent with the flux channels). Lists stay ASCII-sorted; snapshot/window overlap validation passes.
- **stats**: `get_stats` computes over all data_vars minus `DROP_VARIABLES`, so the new channels get centering/scaling stats automatically; stats dirs and beaker dataset names are bumped in all three configs.
- **NaN handling**: all six inputs take the strict any-NaN branch of `_check_data_validity`, which is exercised (every production Makefile target passes `--check_data_validity`), and all are globally-defined fields already strict-checked in production by streams 1/2/4. `skipna` regridding is only needed for masked fields like SST — not these.
- **`xr.set_options(keep_attrs=True)`**: global rather than a context manager, but this is the pre-existing idiom in all four existing process functions, so Stream 5 conforming changes no behavior.

**Verdict:** fix the description sentence, otherwise LGTM — no code changes needed.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Add it in this PR.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -1239,6 +1338,27 @@ def main():
             )
         )
 
+        # Stream 5: Surface mean (hourly -> regrid -> derive -> average)
```
mcgibbon, line 1346: @spencerkclark can you review this in particular?

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: Claude: Added in c07a147 — PRESsfc_mean now in the output Dataset, DESIRED_ATTRS, and all three configs' window_names.

scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
```diff
@@ -1,16 +1,16 @@
...
   start_date: "1990-01-01"
   end_date: "2019-12-31"
   data_type: ERA5
 time_coarsen:
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-1940-2025
-  stats_output_directory: gs://vcm-ml-intermediate/2026-04-17-era5-4deg-8layer-daily-stats-1990-2019
+  data_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-1940-2025
+  stats_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-stats-1990-2019
   snapshot_names:
```
mcgibbon, line 14: @spencerkclark are there any other fields in snapshot_names we need daily means of? Idk if people like having e.g. h500 daily means or daily mean soil moisture?

spencerkclark: I guess two other broader use-cases that come to mind are potentially offering a way to address what @brianhenn brought up regarding AIMIP (i.e. what constitutes a time-mean quantity currently depends on the timestep for any snapshot fields) and any potential intersection with your work emulating CMIP6 models (which if I understand correctly depends on emulating daily mean fields).  But I'm not sure whether those are things we want to address at the moment or not.

scripts/era5/pipeline/xr-beam-pipeline.py
spencerkclark: Looks good to me.

scripts/era5/pipeline/xr-beam-pipeline.py
```diff
@@ -634,6 +675,59 @@ def process_mean_flux(
...
+    output["surface_temperature_mean"] = regridded["skin_temperature"]
+    output["PRESsfc_mean"] = regridded["surface_pressure"]
+    output["TMP2m_mean"] = regridded["2m_temperature"]
+    output["DPT2m_mean"] = regridded["2m_dewpoint_temperature"]
+    output["Q2m_mean"] = q2m
+    output["UGRD10m_mean"] = regridded["10m_u_component_of_wind"]
+    output["VGRD10m_mean"] = regridded["10m_v_component_of_wind"]
+    output["WIND10m_mean"] = wind_speed
+
+    averaged = _average_hourly_to_6hourly(output)
```
spencerkclark, line 708: I suppose we could use the trapezoid rule instead of a Riemann sum as a slightly more accurate way of estimating the time-integrated value, but that might not be worth the complexity, since it would require iterating over overlapping chunks of data.

spencerkclark (approved):
Looks good—a couple comments, but I do not think they necessarily need to hold up the PR.

scripts/era5/pipeline/xr-beam-pipeline.py
mcgibbon: While in the long run I'd prefer to have a value that's actually time-integrated over the period instead of an average of snapshots, I think for now we should keep and consider these as an exact average of hourly snapshots, rather than as an approximation of integrated values. When these are time-coarsened, they will represent the average of 24 hourly snapshots spanning the diurnal cycle, with the input snapshot not counted and the output snapshot counted towards that output average.

PR #1437, opened 2026-08-13
Add include_input_step option to SecondaryDecoderConfig

mcgibbon:
Add `include_input_step: bool = False` to `SecondaryDecoderConfig`. When True, the channels provided as inputs to the primary network are concatenated with its outputs before being passed to the secondary decoder. This lets the decoder condition on both the predicted state and the original input fields (e.g. forcings). Default False preserves existing behavior.

Changes:
- `fme.core.step.secondary_decoder.SecondaryDecoderConfig`: new `include_input_step` bool field (default False)
- `fme.core.step.single_module.SingleModuleStep`: adjust secondary decoder input channel count and concatenate input tensor when `include_input_step` is True
- `fme.core.step.secondary_module.SecondaryModuleStep`: same changes for the secondary-module step variant

- [x] Tests added

### DIFF
--- a/fme/core/step/secondary_decoder.py
+++ b/fme/core/step/secondary_decoder.py
@@ -24,10 +24,14 @@ class SecondaryDecoderConfig:
             diagnosed directly from outputs without access to latent variables (i.e.,
             column-locally).
         network: Configuration for the decoder network.
+        include_input_step: If True, the channels provided as inputs to the
+            primary network are concatenated with its outputs before being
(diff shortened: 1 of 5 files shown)


mcgibbon:
Ah this PR brings me back to the good old days two months ago of simple and straightforward Opus 4.6 PRs (because I used 4.6).

AnnaKwa (approved):
LGTM!

PR #1438, opened 2026-08-13
Park at a loop boundary before the preemption abort, so the abort never meets an active NVLink transfer

mcgibbon:
#1425 aborts NCCL communicators on SIGTERM so a preempted rank can exit without faulting its peers. The [Aug 13 preemption of its own 8-GPU test run](https://beaker.org/orgs/ai2/workspaces/ace/work/01KZRC5Q06TP2J3JWERA5D91BR?taskId=01KZRC5Q0GKVJAW6632C3TAJCV&jobId=01KZWWJMVWABTEDM5TAZ6P8AYS) showed that is not enough: all 8 ranks aborted mid-training-step while kernels were actively exchanging data over NVLink, the NVSwitch raised SXid 12028, and Beaker cordoned the node (NVIDIA's recovery for that SXid is a node restart, so there is no scheduler-side relief). #1425's hardware probe had validated aborting *idle or waiting* kernels only; the unsafe case is aborting kernels that are actively transferring. So this PR defers the abort until this rank's collectives are quiet.

**Mechanism.** On a termination signal the listener records intent instead of aborting. Every batch loop calls `Distributed.park_if_terminating()` once per batch/window — a local flag read, no collective, so uneven call sites cannot deadlock. When it sees a pending termination, the main thread:

1. **drains device work** (`torch.cuda.synchronize`): reaching the boundary is a host-side fact, and DDP enqueues the gradient all-reduce asynchronously, so the last batch's collectives may still be running;
2. **parks**: blocks permanently with the device idle, freezing training state at a batch boundary — so the restart checkpoint is boundary-consistent, not merely un-torn.

The abort comes at one uniform instant across ranks — a settle period past the park deadline — even on a parked rank whose own device is idle: aborting early frees communicator buffers that peers still wedged in a later collective are polling, which the hardware trial observed as contained NVLink faults. The listener then runs the post-abort callbacks and holds the `128+signum` exit until no peer can still be aborting (see Timing).

**Fallback.** A main thread that misses the park deadline is wedged in a collective a peer will never complete — or in a drain of one, or its drain raised. Its kernels are *waiting*, not transferring: the probe-validated safe shape, so the abort then proceeds exactly as before this PR.

**Timing.** Both bounds are measured from the signal, not from any rank's own progress: every rank aborts at the uniform time (a wedged main thread at the 5 s park deadline; a frozen one a settle period later, so wedged peers' aborts land first), and no rank exits before deadline + settle + grace = 11 s, the latest instant any peer can still be aborting plus an allowance for the abort itself and signal skew. Checkpoint writes start after the abort (~6 s), leaving ~24 s of torchrun's 30 s SIGTERM→SIGKILL window (the hardware trial saw ~19 s writes, which the previous 10 s deadline could lose to SIGKILL). An overrun cannot re-fault the fabric (every abort fires long before SIGKILL) or corrupt the checkpoint (`save_checkpoint` writes tmp-then-rename); a write slower than the remaining window loses the new checkpoint and restarts from the previous one.

**Watchdog (second commit).** In the crash log, five ranks died instantly when their NCCL watchdogs rethrew the CUDA error — a `std::terminate` on the watchdog thread bypasses the settle and grace periods for any residual fault the deferral cannot prevent. `TORCH_NCCL_RETHROW_CUDA_ERRORS=0` downgrades the watchdog to logging. This is a process-global change in how a genuine mid-training NCCL fault presents: a sticky CUDA error at the main thread's next sync of that work, rather than a watchdog terminate — failures are routed through the shutdown path, not masked.

**Not covered:**
- A signal landing in a check-free region longer than the park deadline (epoch-end aggregator reductions are the main one) keeps #1425's behavior: abort into whatever kernel state exists.
- A rank that dies without running Python (segfault, OOM kill) — out of scope, as before.
- The train-evaluation loop's park point sits inside the EMA context, where the restart-checkpoint callback deliberately declines to save. A preemption in that window parks safely but loses the epoch's progress; before this PR the released main thread unwound out of the EMA context and the checkpoint was written.
- The design is argued-plus-CPU-tested until the hardware trial: the crashed test job's exact configuration rerun from this branch with deliberate preemptions during active training.

**Changes:**
- `fme.core.distributed.shutdown`: `park_if_terminating()` (new) drains device work, freezes training state, and parks the calling main thread; the listener aborts at the uniform time — `park_deadline` (5 s) after the signal for a main thread that never froze, plus `settle_period` (1 s) for one that did, so idle ranks abort after their wedged peers. The context's exit drains likewise whenever the abort has not yet fired (gating on a pending termination instead would race the listener's read of the signal). Forked children clear the armed context; DataLoader workers and non-main threads never park.
- `fme.core.distributed.Distributed`: `park_if_terminating` (public wrapper) and `_drain_local_work` (`torch.cuda.synchronize` on GPU, no-op on CPU; the context's `drain`).
- Every batch loop calls `park_if_terminating` once per batch/window/chunk: `fme.core.generics` train (both loops), validation, inference, lr-tuning, dataset comparison (ace and coupled), and `fme.downscaling` train, validation, prediction, evaluation, event evaluation.
- `fme.core.distributed.torch_distributed`, `model_torch_distributed`: set `TORCH_NCCL_RETHROW_CUDA_ERRORS=0` before process-group construction.
- Tests: the parked path (bounded on the child's own clock, callbacks against boundary-frozen state), both context-exit drain branches, the uniform abort time (a parked abort held past the deadline, the settle period on top of it), the deadline bounding a never-parking main thread, a two-rank park/wedge split where the wedged abort must land first and no rank may exit before the last abort, drain ordering, hanging and raising drains, and the fork, DataLoader-worker, non-main-thread, and no-signal guards.

- [x] Tests added

Stacked on #1435; diff shown against it. Supersedes the remaining intent of #1407 (cooperative stop) in minimal form.

### DIFF
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -11,7 +11,7 @@ from fme.core import metrics
 from .base import DistributedBackend
 from .model_torch_distributed import ModelTorchDistributed
 from .non_distributed import NonDistributed
-from .shutdown import handle_termination_signals
+from .shutdown import abort_and_exit_on_termination
 from .torch_distributed import TorchDistributed
 
 logger = logging.getLogger(__name__)
@@ -94,7 +94,7 @@ class Distributed:
         try:
             with contextlib.ExitStack() as stack:
                 if handle_signals and instance.is_distributed():
-                    stack.enter_context(handle_termination_signals(instance.abort))
+                    stack.enter_context(abort_and_exit_on_termination(instance.abort))
                 try:
                     yield
                 except BaseException:
@@ -211,13 +211,21 @@ class Distributed:
         """
         return self._distributed.total_ranks
 
+    @property
+    def has_spatial_parallelism(self) -> bool:
+        """Whether spatial co-ranks exist (world_size >
+        total_data_parallel_ranks), meaning each rank holds only a shard of
+        the model's spatial state rather than a full copy.
+        """
+        return self.world_size != self.total_data_parallel_ranks
+
     def require_no_spatial_parallelism(self, msg: str) -> None:
         """Raise if spatial parallelism is active.
 
         Use this to guard code paths that are known to be incorrect
-        when spatial co-ranks exist (world_size > total_data_parallel_ranks).
+        when spatial co-ranks exist.
         """
-        if self.world_size != self.total_data_parallel_ranks:
+        if self.has_spatial_parallelism:
             raise SpatialParallelismNotImplemented(msg)
 
     def get_sampler(
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -213,7 +213,7 @@ class Distributed:
 
     @property
     def has_spatial_parallelism(self) -> bool:
-        """Whether spatial co-ranks exist (world_size >
+        """Whether spatial co-ranks exist (world_size !=
         total_data_parallel_ranks), meaning each rank holds only a shard of
         the model's spatial state rather than a full copy.
         """
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -97,7 +97,7 @@ class Distributed:
                 if handle_signals and instance.is_distributed():
                     stack.enter_context(
                         abort_and_exit_on_termination(
-                            instance.abort, drain=instance.drain_local_work
+                            instance.abort, drain=instance._drain_local_work
                         )
                     )
                 try:
@@ -541,10 +541,10 @@ class Distributed:
         return self._distributed.abort()
 
     def park_if_terminating(self):
-        """Offer this point -- a loop boundary, past the batch's last
-        collective launch -- to a pending termination.
+        """
+        If we've received a termination signal, stop work and wait for termination.
 
-        The batch loops call this once per batch or window. It returns
+        The batch loops should call this once per batch or window. It returns
         immediately unless a termination signal has arrived; if one has, it
         drains this rank's device work, blocks, and the process exits from
         the termination listener's thread. See
@@ -552,7 +552,7 @@ class Distributed:
         """
         park_if_terminating()
 
-    def drain_local_work(self):
+    def _drain_local_work(self):
         """Block until every kernel this rank has enqueued -- collectives
         included -- has completed. On CPU there is nothing asynchronous to
         drain.
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -11,7 +11,7 @@ from fme.core import metrics
 from .base import DistributedBackend
 from .model_torch_distributed import ModelTorchDistributed
 from .non_distributed import NonDistributed
-from .shutdown import abort_and_exit_on_termination
+from .shutdown import abort_and_exit_on_termination, park_if_terminating
 from .torch_distributed import TorchDistributed
 
 logger = logging.getLogger(__name__)
@@ -534,6 +534,17 @@ class Distributed:
         to call from a non-main thread.
         """
         return self._distributed.abort()
+
+    def park_if_terminating(self):
+        """Offer this point -- a loop boundary, with no collective in flight --
+        to a pending termination.
+
+        The batch loops call this once per batch or window. It returns
+        immediately unless a termination signal has arrived; if one has, it
+        blocks, and the process exits from the termination listener's thread.
+        See `fme.core.distributed.shutdown.park_if_terminating`.
+        """
+        park_if_terminating()
 
 
 singleton: Distributed | None = None
--- a/fme/core/distributed/distributed.py
+++ b/fme/core/distributed/distributed.py
@@ -7,6 +7,7 @@ from typing import TypeVar
 import torch
 
 from fme.core import metrics
+from fme.core.device import using_gpu
 
 from .base import DistributedBackend
 from .model_torch_distributed import ModelTorchDistributed
@@ -94,7 +95,11 @@ class Distributed:
         try:
             with contextlib.ExitStack() as stack:
                 if handle_signals and instance.is_distributed():
-                    stack.enter_context(abort_and_exit_on_termination(instance.abort))
+                    stack.enter_context(
+                        abort_and_exit_on_termination(
+                            instance.abort, drain=instance.drain_local_work
+                        )
+                    )
                 try:
                     yield
                 except BaseException:
@@ -536,15 +541,24 @@ class Distributed:
         return self._distributed.abort()
 
     def park_if_terminating(self):
-        """Offer this point -- a loop boundary, with no collective in flight --
-        to a pending termination.
+        """Offer this point -- a loop boundary, past the batch's last
+        collective launch -- to a pending termination.
 
         The batch loops call this once per batch or window. It returns
         immediately unless a termination signal has arrived; if one has, it
-        blocks, and the process exits from the termination listener's thread.
-        See `fme.core.distributed.shutdown.park_if_terminating`.
+        drains this rank's device work, blocks, and the process exits from
+        the termination listener's thread. See
+        `fme.core.distributed.shutdown.park_if_terminating`.
         """
         park_if_terminating()
+
+    def drain_local_work(self):
+        """Block until every kernel this rank has enqueued -- collectives
+        included -- has completed. On CPU there is nothing asynchronous to
+        drain.
+        """
+        if using_gpu():
+            torch.cuda.synchronize()
 
 
 singleton: Distributed | None = None
--- a/fme/core/distributed/shutdown.py
+++ b/fme/core/distributed/shutdown.py
@@ -1,4 +1,4 @@
-"""Tear the distributed backend down safely when a job is preempted.
+"""Exit safely when a job is preempted, by aborting NCCL before anything else.
 
 Schedulers preempt jobs by sending SIGTERM and escalating to SIGKILL once a
 grace period expires. A rank that is killed while a peer's NCCL kernels still
@@ -13,12 +13,13 @@ termination signal it aborts its own communicators -- ``ncclCommAbort`` kills
 the local kernels and unblocks whatever host thread was waiting on them --
 then waits out a grace period so its peers' aborts finish before it exits.
 
-A Python signal handler runs only when the main thread returns to the
-interpreter, and a preempted rank's main thread is typically blocked inside a
-collective's stream sync, so a handler could never run in time. Instead the
-signal is observed on a dedicated listener thread through
-``signal.set_wakeup_fd``, which CPython's C-level handler writes regardless of
-what the main thread is doing.
+The signal is received on a *listener* -- a dedicated thread whose only job is
+to wait for the signal -- because the main thread cannot be trusted to receive
+it: a preempted rank's main thread is typically blocked inside a collective,
+where an ordinary Python signal handler would never run. How the listener
+observes the signal anyway is `SignalListener`'s concern
+(`fme.core.distributed._signal_listener`); everything that *happens* on
+termination is decided here, in `_abort_and_exit` and the steps it calls.
 
 The instants before the listener is armed remain unprotected --
 ``init_process_group`` itself, inside ``Distributed.context()`` entry, most
@@ -31,10 +32,11 @@ import signal
 import threading
 import time
 import traceback
-import types
 from collections.abc import Callable, Generator
 
 from fme.core.device import in_dataloader_worker
+
+from ._signal_listener import SignalListener
 
 TERMINATION_SIGNALS: tuple[signal.Signals, ...] = (signal.SIGTERM, signal.SIGINT)
 
@@ -45,18 +47,12 @@ TERMINATION_SIGNALS: tuple[signal.Signals, ...] = (signal.SIGTERM, signal.SIGINT
 # default timeout, torch/distributed/elastic/agent/server/local_elastic_agent.py).
 DEFAULT_GRACE_PERIOD = 5.0
 
-# How long the listener waits for the main thread to unwind into the
-# context's exit before giving up on the post-abort callbacks. One batch of
-# compute bounds the unwind in the common case; together with the callbacks
-# and grace period this must fit torchrun's 30s SIGTERM-to-SIGKILL budget.
-DEFAULT_PARK_TIMEOUT = 5.0
-
-# tells the sentinel apart from signal numbers, which are all positive
-_STOP_LISTENING = 0
-
-_armed = False  # whether the wakeup fd currently belongs to this module
-_pipe_fds: tuple[int, int] | None = None
-_at_fork_registered = False
+# How long the listener waits for the main thread to stop touching training
+# state (see _ExitCoordination.training_state_frozen) before giving up on the
+# post-abort callbacks. One batch of compute bounds the wait in the common
+# case; together with the callbacks and grace period this must fit torchrun's
+# 30s SIGTERM-to-SIGKILL budget.
+DEFAULT_STATE_FREEZE_TIMEOUT = 5.0
 
 _post_abort_callbacks: list[Callable[[], None]] = []
 
@@ -67,8 +63,8 @@ def add_post_abort_callback(callback: Callable[[], None]) -> None:
     Callbacks run only once the main thread has unwound into the context's
     exit, where it blocks until the process ends -- so their reads of training
     state cannot race it. A main thread that has not unwound within
-    ``park_timeout`` (still computing, or stuck somewhere that is not a
-    collective) forfeits the callbacks rather than risk a torn snapshot.
+    ``state_freeze_timeout`` (still computing, or stuck somewhere that is not
+    a collective) forfeits the callbacks rather than risk a torn snapshot.
     Best-effort in duration too: the scheduler's SIGKILL caps how long a
     callback may take. The communicators are gone by then, so callbacks must
     not use collectives, nor the logging module (see ``write_stderr``).
@@ -78,38 +74,6 @@ def add_post_abort_callback(callback: Callable[[], None]) -> None:
 
 def clear_post_abort_callbacks() -> None:
     _post_abort_callbacks.clear()
-
-
-def _disarm_in_child() -> None:
-    """Undo the parent's signal setup in a forked child.
-
-    A fork-started DataLoader worker inherits the ignore-handler and the wakeup
-    fd, which still feeds the parent's pipe: a signal delivered to the worker
-    would masquerade as the parent's own preemption. Restore the defaults so
-    the child dies as it would have without the inheritance. This runs in the
-    child's main thread -- fork's surviving thread is re-designated as such
-    before the at-fork hooks run -- so the signal module cooperates.
-    """
-    global _armed, _pipe_fds
-    if not _armed:
-        return
-    for sig in TERMINATION_SIGNALS:
-        signal.signal(sig, signal.SIG_DFL)
-    signal.set_wakeup_fd(-1)
-    if _pipe_fds is not None:
-        for fd in _pipe_fds:
-            os.close(fd)
-    # clear the state, or a second-generation fork (dataset code forking a
-    # worker's own subprocess) re-runs this on fd numbers the worker has
-    # since reused
-    _armed = False
-    _pipe_fds = None
-
-
-def _ignore_signal(signum: int, frame: types.FrameType | None) -> None:
-    # exists so CPython delivers the signal (writing it to the wakeup fd)
-    # instead of taking the default action; the listener thread does the work
-    pass
 
 
 def write_stderr(message: str) -> None:
@@ -127,73 +91,105 @@ def write_stderr(message: str) -> None:
         pass
 
 
-def _wait_for_termination_signal(read_fd: int) -> signal.Signals | None:
-    while True:
-        data = os.read(read_fd, 64)
-        # the wakeup fd sees every signal CPython delivers (torch's DataLoader
-        # installs a SIGCHLD handler, for one); act only on termination.
-        # Checked before the sentinel: a signal racing the context's exit must
-        # win, or the process walks on with its protection already removed.
-        for sig in TERMINATION_SIGNALS:
-            if sig in data:
-                return sig
-        if not data or _STOP_LISTENING in data:
-            return None
+class _ExitCoordination:
+    """The two facts the main thread and the listener thread tell each other.
+
+    Each starts false and flips true exactly once; the methods wrap the
+    thread-safe flag primitive (``threading.Event``) in the fact it records.
+    """
+
+    def __init__(self) -> None:
+        self._listener_owns_exit = threading.Event()
+        self._training_state_frozen = threading.Event()
+
+    def mark_listener_owns_exit(self) -> None:
+        """A termination signal arrived: from now on the listener thread ends
+        the process, and the main thread must not exit on its own.
+        """
+        self._listener_owns_exit.set()
+
+    def listener_owns_exit(self) -> bool:
+        return self._listener_owns_exit.is_set()
+
+    def mark_training_state_frozen(self) -> None:
+        """The main thread will never touch training state again, so the
+        post-abort callbacks may read it without racing a mutation.
+        """
+        self._training_state_frozen.set()
+
+    def wait_until_training_state_frozen(self, timeout: float) -> bool:
+        """Block up to ``timeout`` seconds for the freeze; False on timeout."""
+        return self._training_state_frozen.wait(timeout)
 
 
-def _listen(
-    read_fd: int,
-    abort: Callable[[], None],
-    grace_period: float,
-    park_timeout: float,
-    terminating: threading.Event,
-    main_parked: threading.Event,
-) -> None:
-    signum = _wait_for_termination_signal(read_fd)
-    if signum is None:  # the context exited normally
-        return
-    terminating.set()  # from here on this thread owns the process's exit
-    write_stderr(
-        f"Received {signal.Signals(signum).name}, aborting distributed "
-        "communicators before exiting.\n"
-    )
+def _abort_local_communicators(abort: Callable[[], None]) -> None:
+    """Stop this rank's own NCCL kernels and release its wedged host threads."""
     try:
         # not bounded: if the abort hangs, exiting anyway would guarantee the
         # peer-GPU fault, so the rank rides to the scheduler's SIGKILL instead
         abort()
     except BaseException:
         write_stderr(f"Aborting communicators failed:\n{traceback.format_exc()}")
-    if _post_abort_callbacks:
-        # the abort releases only a main thread blocked in a collective; one
-        # between collectives keeps computing until the next collective
-        # raises, and the callbacks' reads of training state would race it.
-        # Wait for it to block in the context's exit, and forfeit the
-        # callbacks rather than record a torn snapshot if it never does.
-        if main_parked.wait(park_timeout):
-            for callback in _post_abort_callbacks:
-                try:
-                    callback()
-                except BaseException:
-                    write_stderr(
-                        f"Post-abort callback failed:\n{traceback.format_exc()}"
-                    )
-        else:
-            write_stderr(
-                f"Main thread still running {park_timeout}s after the abort; "
-                "skipping post-abort callbacks.\n"
-            )
-    # peers' aborts must finish, so their kernels stop touching our memory,
-    # before we exit
+
+
+def _run_post_abort_callbacks(
+    coordination: _ExitCoordination, state_freeze_timeout: float
+) -> None:
+    """Run the registered callbacks once training state is safe to read.
+
+    The abort releases only a main thread blocked in a collective; one between
+    collectives keeps computing until the next collective raises, and the
+    callbacks' reads of training state would race it. Wait for it to block in
+    the context's exit, and forfeit the callbacks rather than record a torn
+    snapshot if it never does.
+    """
+    if not _post_abort_callbacks:
+        return
+    if not coordination.wait_until_training_state_frozen(state_freeze_timeout):
+        write_stderr(
+            f"Main thread still running {state_freeze_timeout}s after the "
+            "abort; skipping post-abort callbacks.\n"
+        )
+        return
+    for callback in _post_abort_callbacks:
+        try:
+            callback()
+        except BaseException:
+            write_stderr(f"Post-abort callback failed:\n{traceback.format_exc()}")
+
+
+def _wait_for_peer_aborts(grace_period: float) -> None:
+    """Peers' aborts must finish, so their kernels stop touching this rank's
+    memory, before this rank exits.
+    """
     time.sleep(grace_period)
+
+
+def _abort_and_exit(
+    signum: signal.Signals,
+    abort: Callable[[], None],
+    coordination: _ExitCoordination,
+    grace_period: float,
+    state_freeze_timeout: float,
+) -> None:
+    """The termination policy. Runs on the listener thread."""
+    coordination.mark_listener_owns_exit()
+    write_stderr(
+        f"Received {signal.Signals(signum).name}, aborting distributed "
+        "communicators before exiting.\n"
+    )
+    _abort_local_communicators(abort)
+    _run_post_abort_callbacks(coordination, state_freeze_timeout)
+    _wait_for_peer_aborts(grace_period)
     write_stderr(f"Exiting with code {128 + signum}.\n")
     os._exit(128 + signum)
 
 
 @contextlib.contextmanager
-def handle_termination_signals(
+def abort_and_exit_on_termination(
     abort: Callable[[], None],
     grace_period: float = DEFAULT_GRACE_PERIOD,
-    park_timeout: float = DEFAULT_PARK_TIMEOUT,
+    state_freeze_timeout: float = DEFAULT_STATE_FREEZE_TIMEOUT,
 ) -> Generator[None, None, None]:
     """Exit on SIGTERM or SIGINT, aborting the distributed backend first.
 
@@ -213,11 +209,10 @@ def handle_termination_signals(
             collective, so it must not require the main thread's cooperation.
         grace_period: Seconds to wait between aborting and exiting, so that
             peers' aborts finish while this process's memory is still mapped.
-        park_timeout: Seconds the listener waits for the main thread to block
-            in this context's exit before skipping the post-abort callbacks
-            (see ``add_post_abort_callback``).
+        state_freeze_timeout: Seconds the listener waits for the main thread
+            to block in this context's exit before skipping the post-abort
+            callbacks (see ``add_post_abort_callback``).
     """
-    global _armed, _pipe_fds, _at_fork_registered
     if threading.current_thread() is not threading.main_thread():
         # only the main thread may install handlers; a thread shares the
         # process disposition its main thread installed
@@ -230,58 +225,33 @@ def handle_termination_signals(
         yield
         return
 
-    if not _at_fork_registered:
-        os.register_at_fork(after_in_child=_disarm_in_child)
-        _at_fork_registered = True
-
-    read_fd, write_fd = os.pipe()
-    os.set_blocking(write_fd, False)
-    terminating = threading.Event()
-    main_parked = threading.Event()
-    listener = threading.Thread(
-        target=_listen,
-        args=(read_fd, abort, grace_period, park_timeout, terminating, main_parked),
-        name="termination-listener",
-        daemon=True,
+    coordination = _ExitCoordination()
+    listener = SignalListener(
+        signals=TERMINATION_SIGNALS,
+        on_signal=lambda signum: _abort_and_exit(
+            signum, abort, coordination, grace_period, state_freeze_timeout
+        ),
     )
     listener.start()
-    previous_handlers = {sig: signal.getsignal(sig) for sig in TERMINATION_SIGNALS}
-    previous_fd = signal.set_wakeup_fd(write_fd, warn_on_full_buffer=False)
-    for sig in TERMINATION_SIGNALS:
-        signal.signal(sig, _ignore_signal)
-    _pipe_fds = (read_fd, write_fd)
-    _armed = True
     try:
         yield
     finally:
-        _armed = False
-        _pipe_fds = None
-        # the main thread only joins the listener and restores dispositions
-        # from here; it will not touch training state again, so the
-        # post-abort callbacks may read it
-        main_parked.set()
-        # a signal delivered between the listener consuming this sentinel and
-        # the dispositions being restored below is dropped; the window is
-        # microseconds wide and the process is exiting anyway
-        os.write(write_fd, bytes([_STOP_LISTENING]))
-        if not terminating.is_set():
+        # the main thread only winds the listener down from here; it will not
+        # touch training state again, so the post-abort callbacks may read it
+        coordination.mark_training_state_frozen()
+        listener.request_stop()
+        if not coordination.listener_owns_exit():
             # covers the race where a signal was delivered but not yet read:
             # long enough for the listener to abort and exit
-            listener.join(timeout=grace_period + 10.0)
-        if terminating.is_set():
-            # the listener owns the exit: hold the unwinding main thread for
-            # the abort and grace period, with the ignore-handlers still
-            # installed -- restoring the previous dispositions (typically
-            # SIG_DFL) here would let a repeated signal kill the rank before
-            # the grace period ends. If the abort hangs, the scheduler's
-            # SIGKILL is the backstop.
-            listener.join()
+            listener.wait_until_finished(timeout=grace_period + 10.0)
+        if coordination.listener_owns_exit():
+            # hold the unwinding main thread for the abort and grace period,
+            # with the listener still armed -- dismantling it here would
+            # restore the previous dispositions (typically SIG_DFL) and let a
+            # repeated signal kill the rank before the grace period ends
+            listener.block_until_process_exit()
         else:
-            for sig, handler in previous_handlers.items():
-                signal.signal(sig, handler)
-            signal.set_wakeup_fd(previous_fd)
-            os.close(write_fd)
-            os.close(read_fd)
+            listener.dismantle()
             # the callbacks belonged to this context's session; a later
             # context must not fire them against torn-down state
             clear_post_abort_callbacks()
--- a/fme/core/distributed/shutdown.py
+++ b/fme/core/distributed/shutdown.py
@@ -48,7 +48,7 @@ TERMINATION_SIGNALS: tuple[signal.Signals, ...] = (signal.SIGTERM, signal.SIGINT
 DEFAULT_GRACE_PERIOD = 5.0
 
 # How long the listener waits for the main thread to stop touching training
-# state (see _ExitCoordination.training_state_frozen) before giving up on the
+# state (see _ExitCoordination.mark_training_state_frozen) before giving up on the
 # post-abort callbacks. One batch of compute bounds the wait in the common
 # case; together with the callbacks and grace period this must fit torchrun's
 # 30s SIGTERM-to-SIGKILL budget.
--- a/fme/core/distributed/shutdown.py
+++ b/fme/core/distributed/shutdown.py
@@ -13,6 +13,19 @@ termination signal it aborts its own communicators -- ``ncclCommAbort`` kills
 the local kernels and unblocks whatever host thread was waiting on them --
 then waits out a grace period so its peers' aborts finish before it exits.
 
+The abort itself is deferred until this rank's collectives are idle or merely
+waiting. Aborting communicators whose kernels are actively exchanging data
+faulted the fabric in a real preemption (8xH100, mid-training-step: contained
+NVLink peer-access errors, an unrecoverable NVSwitch SXid, node cordoned),
+while aborting idle or *waiting* kernels was validated safe on H100 and B200.
+So the training, validation, and inference loops offer every batch boundary
+as a stopping point (`park_if_terminating`), and the listener aborts only
+once the main thread has parked at one -- or has unwound into the context's
+exit -- plus a settle period for peers' kernel tails. A main thread that
+misses the deadline is wedged in a collective a peer will never complete,
+where the kernels are waiting rather than transferring and the abort is safe
+anyway.
+
 The signal is received on a *listener* -- a dedicated thread whose only job is
 to wait for the signal -- because the main thread cannot be trusted to receive
 it: a preempted rank's main thread is typically blocked inside a collective,
@@ -53,6 +66,22 @@ DEFAULT_GRACE_PERIOD = 5.0
 # case; together with the callbacks and grace period this must fit torchrun's
 # 30s SIGTERM-to-SIGKILL budget.
 DEFAULT_STATE_FREEZE_TIMEOUT = 5.0
+
+# How long the listener waits for the main thread to park at a loop boundary
+# before aborting anyway. Long enough for the tail of a training batch or an
+# inference window (sub-second to a few seconds); a main thread that misses it
+# is taken to be wedged in a collective a peer will never complete, whose
+# kernels are waiting rather than transferring, so the abort is safe without
+# the park. With the settle period, abort, post-abort callbacks, and grace
+# period behind it, the sum must fit the same 30s budget.
+DEFAULT_PARK_DEADLINE = 10.0
+
+# How long a parked rank waits before aborting. Parking means every collective
+# this rank joined has completed *for this rank*; a peer's kernel for that
+# same collective may still be retiring -- actively reading this rank's
+# memory for a little longer. Kernel tails are sub-millisecond and signal
+# skew across ranks is tens of milliseconds, so this is a generous cover.
+DEFAULT_SETTLE_PERIOD = 1.0
 
 _post_abort_callbacks: list[Callable[[], None]] = []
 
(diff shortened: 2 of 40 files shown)


mcgibbon (commented):
Claude: Independent pre-review (no authoring context), against `refactor/shutdown-readability` only.

**Verdict: not yet ready for review — one design finding I think Jeremy should rule on before the hardware trial, plus a sibling-coverage gap and two cheap test/robustness items.** The shape of the change is right and the docs are unusually good; the findings are about whether the safety property is actually *enforced* and where it isn't offered.

Verified clean: `TORCH_NCCL_RETHROW_CUDA_ERRORS` is a live env var in torch 2.7/2.8 and is read into `ProcessGroupNCCL::rethrowCUDAErrors_` at construction, so "read once, at process group construction" is accurate and both `setdefault` calls are correctly placed before the group is built. Ran `fme/core/distributed/test_shutdown.py` on CPU: 23 passed. No deadlock found between `park_if_terminating` and `_wait_for_main_thread_to_park` (the listener's wait is always bounded, and `mark_training_state_frozen()` runs at the top of the context's `finally` before `request_stop()`, so the normal-exit race can never cost the full `park_deadline`). Repeated signals are safe: the parked `threading.Event().wait()` is woken by `_route_to_listener` and resumes.

Findings are in line comments; two that have no diff line to hang on:

**Issue: two collective-dense loops offer no park point, and neither is named in the "what this does not cover" paragraph.**
- `fme/core/generics/lr_tuning.py:120` — `run_lr_tuning_trial` runs `config.num_batches` (default **200**) training batches through *two* steppers, i.e. ~400 optimizer steps with per-batch DDP allreduces, and it is called from `Trainer` whenever `params.lr_tuning` is set. That is minutes long with no park check, so a preemption there always takes the deadline fallback. Two lines to fix.
- `fme/downscaling/train.py:187` and `:267` — a separate DDP training loop (`fme/downscaling/models.py:394` calls `dist.wrap_module`) under its own `Distributed.context()` (`:558`), likewise with no park point. Follow-up PR is fine, but it should be tracked rather than silently skipped.

`fme/coupled` is covered — it builds the same `fme.core.generics.trainer.Trainer` (`fme/coupled/train/train_config.py:663`), so the three call sites carry it.

**Note: this PR has no CI.** `.github/workflows/tests.yaml` and `pre-commit.yml` trigger on `pull_request: branches: [main]`, and this PR targets `refactor/shutdown-readability` — so only GitGuardian has run. For reference, #1435's own run is green apart from a runner-image `pip` `ImportError` on `cpu` and one unrelated `fme/downscaling/test_evaluator.py` flake on `gpu`. Worth saying explicitly when presenting this, since "pre-commit passed locally" is currently the only evidence.

fme/core/distributed/shutdown.py
```diff
@@ -122,6 +154,64 @@ def wait_until_training_state_frozen(self, timeout: float) -> bool:
...
+    """
+    coordination = _active_coordination
+    if coordination is None or not coordination.listener_owns_exit():
+        return
+    if threading.current_thread() is not threading.main_thread():
+        return
+    if in_dataloader_worker():
+        return
+    write_stderr("Main thread parked at a loop boundary for termination.\n")
+    coordination.mark_training_state_frozen()
```
mcgibbon: Claude: **Issue: reaching a loop boundary on the *host* does not mean this rank's NCCL kernels have finished on the *device*, so the property the whole PR rests on isn't enforced by the code.**

`DEFAULT_SETTLE_PERIOD`'s comment says "Parking means every collective this rank joined has completed *for this rank*". CUDA launches are asynchronous, and nothing here synchronizes with the stream before `mark_training_state_frozen()`. DDP's gradient allreduce is enqueued during `backward()` and the reducer's `Work::wait()` only makes the *compute stream* wait on the event — it does not block the host unless `TORCH_NCCL_BLOCKING_WAIT` is set. So the host can be at the next `park_if_terminating()` with the previous step's allreduce still moving data over NVLink: precisely the kernel state the Aug 13 fault says must not be aborted.

What actually saves it today is incidental and config-gated:
- `Optimization._clip_gradients` → `torch.nn.utils.clip_grad_norm_(...).item()` (`fme/core/optimization.py:192`) is a host sync — but only when `max_grad_norm` is set, and it defaults to `None` (`fme/core/optimization.py:101`).
- `GradScaler.step`'s `found_inf` `.item()` is a host sync — but only when `enable_automatic_mixed_precision` is true.
- Nothing else in the batch syncs: `MetricsAggregator.record` keeps device tensors and only `get_metrics()` reduces, every `log_train_every_n_batches` batches.

With either of those on, the claim holds (no further collective is launched after the sync — EMA and metric accumulation are local). With neither, the 1 s settle would have to cover a whole queued training step, which it won't for a large model.

Suggestion: make the guarantee structural rather than incidental — synchronize the device at the park, before marking state frozen. Passing a `drain`/`synchronize` callable into `abort_and_exit_on_termination` alongside `abort` keeps `shutdown.py` backend-agnostic in the shape it already uses. It degrades safely in both directions: if the sync returns, the rank is genuinely drained; if it hangs (a peer parked and will never launch the matching collective) or raises (a peer already aborted), the freeze is never set and the listener takes the deadline path — the wedged-kernel shape the hardware probe already validated. It would need a `try/except BaseException` so a raise can't propagate into the training loop.

If instead the intent is to rely on the per-batch `.item()`, that dependency should be stated here and in the module docstring, because it is invisible from these call sites and silently disappears with `max_grad_norm: null` + AMP off.

fme/core/distributed/shutdown.py
```diff
@@ -122,6 +154,64 @@ def wait_until_training_state_frozen(self, timeout: float) -> bool:
...
+    Only the main thread of a listener-armed process parks; any other caller
+    returns immediately. Reaching a loop boundary in a DataLoader worker means
+    nothing about the parent's collectives, so workers never park.
+    """
+    coordination = _active_coordination
+    if coordination is None or not coordination.listener_owns_exit():
+        return
+    if threading.current_thread() is not threading.main_thread():
+        return
+    if in_dataloader_worker():
```
mcgibbon, line 238: Claude: Two things about this guard.

**Suggestion: `_active_coordination` has no at-fork reset, unlike `_active_listener`.** `_signal_listener._disarm_in_child` (registered via `os.register_at_fork(after_in_child=...)`) clears `_active_listener` in a forked child precisely so inherited parent state can't act there. This PR adds a second process-global with the same hazard and no equivalent: a child forked *after* the signal arrived inherits a coordination whose `_listener_owns_exit` is already set, and `threading` re-designates the surviving thread as the main thread — so `park_if_terminating` would block that child forever. `in_dataloader_worker()` is what saves the DataLoader case; nothing saves a plain `os.fork()` (the shape `test_forked_child_does_not_trigger_the_parent_abort` already exercises). Clearing `_active_coordination` from an `after_in_child` hook in this module would make the invariant hold by construction rather than by every future park call site remembering to be in a DataLoader worker.

**Issue: this branch has no test**, and it is the only thing standing between a fork-started worker and a permanent hang. The idiom is already in the file — `test_no_listener_installed_in_a_dataloader_worker` does `monkeypatch.setattr(torch.utils.data, "get_worker_info", lambda: object())`. A test that arms a context, marks the listener as owning the exit, monkeypatches `get_worker_info`, and asserts `park_if_terminating()` returns would cost a few lines. Cheap enough to do in this PR.

fme/core/generics/trainer.py
```diff
@@ -620,6 +622,7 @@ def train_one_epoch(self):
             for batch in self.train_data.subset_loader(
                 stop_batch=self.params.train_evaluation_batches
             ):
+                dist.park_if_terminating()
```
mcgibbon, line 628: Claude: **Issue (optional): this park point sits *inside* `validation_context()`, which suppresses the restart checkpoint when `validate_using_ema: true`.**

`validation_context` → `_ema_context` sets `self._in_ema_context = True` (`trainer.py:681`), and `save_restart_checkpoints_on_terminate` deliberately skips the save in that state. Parking here therefore freezes the main thread *inside* the EMA context, so the callback writes nothing.

Before this PR the same preemption released the main thread from its collective, it unwound out of `_ema_context` (restoring the non-EMA weights and clearing the flag), reached the shutdown context's exit, and the callback *did* save. So for `validate_using_ema: true` this is a behavior regression, and the loss is a full epoch of training: `_current_epoch_num_batches_seen` is still > 0 here, and `_save_restart_checkpoints()` doesn't run until after this loop (`trainer.py:632`). The window is `train_evaluation_samples // batch_size` batches at the end of every epoch.

It's config-gated (`validate_using_ema: bool = False`, `fme/ace/train/train_config.py:418`) and it is arguably the *correct* conservative behavior, so I'm not calling it blocking — but it is directly counter to the PR description's "Parking also freezes training state at a batch boundary, so the restart checkpoint the post-abort callback writes is boundary-consistent", which is true at three of the four call sites and false at this one. Either move the park to just outside the EMA context, or say so in the description.

(For contrast: the `run_validation_loop` / `run_inference` park points are *not* a regression — by then `_current_epoch_num_batches_seen` has been reset to 0, so no checkpoint was written on either path.)

fme/core/distributed/torch_distributed.py
```diff
@@ -44,6 +44,15 @@ def __init__(self):
...
             return
+        # The NCCL watchdog polls each collective's CUDA events on its own
+        # thread and by default rethrows any CUDA error it meets there; an
+        # exception on a thread with no handler is a std::terminate, killing
+        # the rank instantly and bypassing the shutdown path's settle and
+        # grace periods (five ranks died this way, mid-shutdown, in the
+        # 2026-08-13 preemption fabric fault). The same error still surfaces
+        # on the main thread at its next sync of that work, so the watchdog
+        # is downgraded to logging. Read once, at process group construction.
+        os.environ.setdefault("TORCH_NCCL_RETHROW_CUDA_ERRORS", "0")
```
mcgibbon, line 55: Claude: I verified the mechanism: `TORCH_NCCL_RETHROW_CUDA_ERRORS` is live in torch 2.7 and 2.8 (`torch/include/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp`, read into the `rethrowCUDAErrors_` member), and both call sites set it before the process group is constructed, so the comment is accurate. No objection to the change.

**Note:** worth being explicit that this is a *global* change to how every distributed ace run reports faults, not only preemptions. The PR description's "the same error still surfaces on the main thread at its next sync of that work" is plausible but not verified here, and the practical cost is diagnostic: a genuine mid-training NCCL fault will now present as a generic sticky CUDA error at whatever unrelated sync point comes next, instead of a watchdog abort carrying the NCCL context. Worth confirming on the hardware trial rather than only by argument.

**nit:** the same `setdefault` and a cross-reference comment now live in two backends. A small shared helper in the package (`_configure_nccl_watchdog()`) would stop the two from drifting.

fme/core/distributed/test_shutdown.py
```diff
@@ -525,3 +545,147 @@ def test_no_listener_installed_in_a_dataloader_worker(monkeypatch):
...
+            settle_period=0.05,
+        ):
+            signal.raise_signal(signal.SIGTERM)
+            while True:  # the training loop: a boundary between batches
+                park_if_terminating()
+        """
+    )
+
+    assert "parked at a loop boundary" in result.stderr
+    assert "has not reached a loop boundary" not in result.stderr
```
mcgibbon: Claude: nit: this assertion doesn't pin what the test name promises. "Aborted without waiting for the deadline" would also hold if `_wait_for_main_thread_to_park` slept the full `park_deadline` unconditionally — the deadline message only appears on the *timeout* branch, so its absence just says the freeze was observed. `assert result.stdout.split() == ["abort", "callback"]` would pass either way too.

`test_the_settle_period_holds_the_abort_after_the_park` is the one that genuinely differentiates the wait, so coverage isn't missing overall. An elapsed-time bound here (measure from the signal, assert `< park_deadline`) would make this test carry its own name.

fme/core/distributed/test_shutdown.py
```diff
@@ -525,3 +545,147 @@ def test_no_listener_installed_in_a_dataloader_worker(monkeypatch):
...
+        )
+
+        def side_thread():
+            park_if_terminating()
+            print("side thread returned", flush=True)
+
+        with abort_and_exit_on_termination(
+            abort=lambda: print("abort", flush=True),
+            grace_period=0.1,
+            park_deadline=0.5,
```
mcgibbon: Claude: nit (more optional than usual): the tightest timing margin among the new tests. `time.sleep(0.1)`, a `Thread.start()`, and a `print(flush=True)` all have to land inside 0.5 s, or the listener's deadline fires, `os._exit` runs, and `"side thread returned"` is lost — a loaded CI runner could plausibly miss it.

Nothing in the test depends on the deadline being *short* (it only asserts the message appears), so a larger `park_deadline` buys margin at the cost of a little wall time on an already-`medium_duration` test.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Fixed in cdd907ffb. The context now takes a `drain` callable (`Distributed.drain_local_work`, `torch.cuda.synchronize` on GPU); `park_if_terminating` drains before freezing training state, and the context's exit drains likewise when a termination is pending and the abort has not yet fired (gated on a new `abort_started` fact so the wedged path's post-abort unwind keeps its checkpoint). A drain that hangs or raises leaves the freeze unset, so the listener takes the deadline path as you suggested. Covered by three new tests: drain ordering, hanging drain, raising drain.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Fixed in cdd907ffb. The armed-context global is now cleared in forked children via its own `os.register_at_fork` hook (`_clear_armed_context_in_child`), and `test_a_child_forked_after_the_signal_does_not_park` exercises exactly the fork-after-signal shape. The DataLoader-worker guard also has a direct test now (`test_park_returns_immediately_in_a_dataloader_worker`).

fme/core/generics/trainer.py
mcgibbon: Claude: Keeping the conservative behavior (park inside the EMA context, checkpoint declined) — parking without a checkpoint still prevents the fault, and saving under swapped weights is the bug the guard exists for. Now called out explicitly in the PR description's "what this does not cover" so the epoch-loss trade is a reviewed decision rather than a surprise.

fme/core/distributed/torch_distributed.py
mcgibbon: Claude: Thanks for verifying the mechanism. On the global-behavior concern: agreed it changes how a genuine mid-training NCCL fault presents; this trade-off was reviewed with Jeremy before implementation and is now called out in the PR description so it gets weighed at review too. Leaving the two setdefault sites as-is — the second comment already defers to the first, and a shared helper for two lines seemed like more indirection than it saves.

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Fixed in cdd907ffb — the test now bounds its runtime (park_deadline=30 with an elapsed<25 assertion), so a listener that waited out the deadline fails it by construction.

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Fixed in cdd907ffb — park_deadline widened to 2.0s with the sleep kept at 0.1s, so the margin is ~20x rather than ~5x.

fme/core/distributed/shutdown.py
```diff
@@ -229,16 +391,41 @@ def abort_and_exit_on_termination(
...
+        _at_fork_registered = True
+    _armed_context = _ArmedContext(coordination, drain)
     listener.start()
     try:
         yield
     finally:
         # the main thread only winds the listener down from here; it will not
         # touch training state again, so the post-abort callbacks may read it
-        coordination.mark_training_state_frozen()
+        if coordination.listener_owns_exit() and not coordination.abort_started():
```
mcgibbon: Claude: **Issue: the drain here does not run in the case its own comment names — "an exception racing the signal" — because `listener_owns_exit()` is the flag the racing listener has not set yet.**

`mark_listener_owns_exit()` is the *first* thing `_abort_and_exit` does, but it runs on the listener thread, which has to be scheduled and return from `os.read` first. A main thread that raises immediately after the signal reaches this `finally` inside one GIL switch interval, reads `listener_owns_exit()` as **False**, takes the `else` branch, and marks training state frozen **without draining**. The listener then wakes, sees the freeze already set, skips straight past `_wait_for_main_thread_to_park`'s wait, sleeps only `settle_period`, and aborts — into a rank whose kernels were never drained. That is the F1 hazard, reintroduced on the context-exit path.

Reproduced on `cdd907ffb` (CPU, `abort`/`drain` printing to stdout):

```python
add_post_abort_callback(lambda: print("callback", flush=True))
with abort_and_exit_on_termination(
    abort=lambda: print("abort", flush=True),
    grace_period=0.1, park_deadline=10.0, settle_period=0.05,
    drain=lambda: print("drain", flush=True),
):
    signal.raise_signal(signal.SIGTERM)
    raise RuntimeError("rank raising alongside the signal")
```

stdout is `['abort', 'callback']` — no `drain`. Insert `time.sleep(0.3)` before the `raise`, so the listener wins the race and sets the flag, and the same program prints `['drain', 'abort', 'callback']`. The drain is there; only the losing side of the race skips it.

The same `else` branch also makes the freeze *sticky* on the ordinary no-signal exit: it is set before `request_stop()`, so a signal delivered between there and `dismantle()` likewise aborts after only the settle period with no drain having happened. Narrower still, but the same shape.

Neither is a regression against #1435 (which aborts immediately, with no settle at all), and both windows are milliseconds — but the PR's claim is structural ("the abort never meets an active NVLink transfer"), and on this path it rests on a race instead.

Suggestion: drop `listener_owns_exit()` from the condition, so the invariant is "every freeze is preceded by a drain unless the abort has already fired":

```python
if not coordination.abort_started():
    try:
        drain()
    except BaseException:
        write_stderr(f"Draining local device work failed:\n{traceback.format_exc()}")
    else:
        coordination.mark_training_state_frozen()
else:
    coordination.mark_training_state_frozen()
```

The cost is one `torch.cuda.synchronize()` per process at normal teardown, on a device that is idle by then. `abort_started()` remains the thing that keeps the wedged path's checkpoint, which is the only case where draining would be wrong.

**Also: both new branches of this `finally` are untested.** No test passes a non-trivial `drain` *and* unwinds into the context exit — the three drain tests all park forever and never reach here, and every test that does reach here uses the default `drain=None`, where the two branches are indistinguishable. So the `abort_started()` gate could be deleted or inverted with the suite still green. Two tests that pin them (the first is the failing reproduction above, the second passes today):

```python
@pytest.mark.medium_duration
def test_the_context_exit_drains_when_the_abort_has_not_started():
    """An exception racing the signal unwinds into the context's exit before
    the listener aborts: it must drain there as a park would."""
    result = _run_listener_program(
        """
        import signal
        from fme.core.distributed.shutdown import (
            abort_and_exit_on_termination,
            add_post_abort_callback,
        )

        add_post_abort_callback(lambda: print("callback", flush=True))
        try:
            with abort_and_exit_on_termination(
                abort=lambda: print("abort", flush=True),
                grace_period=0.1,
                park_deadline=10.0,
                settle_period=0.05,
                drain=lambda: print("drain", flush=True),
            ):
                signal.raise_signal(signal.SIGTERM)
                raise RuntimeError("rank raising alongside the signal")
        except RuntimeError:
            pass
        """
    )

    assert result.stdout.split() == ["drain", "abort", "callback"]
    assert result.returncode == 128 + signal.SIGTERM


@pytest.mark.medium_duration
def test_the_wedged_path_keeps_its_callbacks_despite_a_failing_drain():
    """A main thread released by the deadline abort unwinds into the context's
    exit with its device already torn down: draining there would raise and
    forfeit the restart checkpoint, so the exit must skip the drain once the
    abort has started."""
    result = _run_listener_program(
        """
        import signal, threading
        from fme.core.distributed.shutdown import (
            abort_and_exit_on_termination,
            add_post_abort_callback,
        )

        released = threading.Event()

        def abort():
            print("abort", flush=True)
            released.set()

        def drain():
            raise RuntimeError("device torn down by the abort")

        add_post_abort_callback(lambda: print("callback", flush=True))
        try:
            with abort_and_exit_on_termination(
                abort=abort,
                grace_period=0.1,
                park_deadline=0.2,
                state_freeze_timeout=1.0,
                drain=drain,
            ):
                signal.raise_signal(signal.SIGTERM)
                released.wait()  # wedged "in a collective" until the abort
                raise RuntimeError("rank unwinding after its collective died")
        except RuntimeError:
            pass
        """
    )

    assert "Draining local device work failed" not in result.stderr
    assert result.stdout.split() == ["abort", "callback"]
    assert result.returncode == 128 + signal.SIGTERM
```

If the condition changes as suggested, the first test's comment should stop saying "before the listener aborts" and just say the exit drains whenever the abort has not fired.

fme/downscaling/train.py
```diff
@@ -265,6 +267,7 @@ def valid_one_epoch(self) -> GenerationSummary:
                 self.validation_data, random_offset=False, shuffle=False
             )
             for batch in validation_batch_generator:
+                Distributed.get_instance().park_if_terminating()
```
mcgibbon, line 270: Claude: nit (non-blocking): two batch loops still have no park point and are not named in the PR description's "what this does not cover" -- `fme/downscaling/predict.py:221` and `fme/downscaling/evaluator.py:56`, both distributed inference entrypoints. A preemption there falls to the 10s deadline path, i.e. #1425's behavior, which is the same trade the description already accepts for check-free regions; it just isn't stated. Either add the call (one line each, same shape as the train/validation loops) or extend the "does not cover" list.

fme/core/distributed/test_shutdown.py
```diff
@@ -525,3 +546,312 @@ def test_no_listener_installed_in_a_dataloader_worker(monkeypatch):
...
+                park_if_terminating()
+        """
+    )
+    elapsed = time.monotonic() - start
+
+    assert "parked at a loop boundary" in result.stderr
+    assert "has not parked" not in result.stderr
+    assert result.stdout.split() == ["abort", "callback"]
+    assert result.returncode == 128 + signal.SIGTERM
+    assert elapsed < 25.0  # interpreter startup dominates; the deadline would add 30
```
mcgibbon: Claude: The runtime bound does what it was asked to do, but it is measured around the whole subprocess, so interpreter startup is inside the budget. On this machine `sys.executable -c "from fme.core.distributed.shutdown import ..."` costs ~11s, against a 25s bound -- ~2.2x margin, where the sibling `test_the_settle_period_holds_the_abort_after_the_park` gets its margin from measuring *inside* the child (`boundary["at"]`) and so is immune to import time.

Same trick here would make the bound both tighter and safer: print `time.monotonic()` at the park and again from `abort`, and assert the difference is well under `park_deadline`. Non-blocking -- flagging it because a loaded CI runner importing `fme` is the plausible way this one goes red for a reason unrelated to what it tests.

mcgibbon (commented):
Claude: Second-round pre-review. The four round-1 findings are all addressed and I verified each; one new blocking issue came out of the drain fix's context-exit half, with a reproduction and a suggested one-line condition change. Everything else below is non-blocking.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Verified the pieces individually and they hold: `drain` is threaded through as `Distributed.drain_local_work` (`torch.cuda.synchronize` gated on `using_gpu()`, so `cudaDeviceSynchronize` covers NCCL's internal stream too, and CPU/MPS runs no-op); `park_if_terminating` drains before `mark_training_state_frozen` and parks unconditionally afterwards, so a raising drain cannot spin the loop; and the `abort_started()` gate does keep the wedged path's checkpoint -- I confirmed that one empirically with a drain that raises and a main thread released by the deadline abort, and the callback still runs.

Two things came out of it, both on the context-exit half: the `listener_owns_exit()` half of the gate loses the race it was written for, and neither branch of that `finally` is covered by the suite. Details and a reproduction in a new comment on the `finally` block.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Verified. `_clear_armed_context_in_child` is registered before `_armed_context` is ever set and before `listener.start()`, so a child forked at any point after arming sees `None`; `test_a_child_forked_after_the_signal_does_not_park` covers the fork-after-signal shape and the parent still parks afterwards. `test_park_returns_immediately_in_a_dataloader_worker` genuinely exercises the worker guard (it runs on pytest's main thread, so the main-thread guard is not what returns), and if the guard were removed it fails on the conftest SIGALRM rather than hanging the session, since `Event().wait()` on the main thread is signal-interruptible. Nothing further from me here.

fme/core/generics/trainer.py
mcgibbon: Claude: Agreed with the decision -- parking without a checkpoint beats aborting into live kernels, and saving under swapped weights is the bug the guard exists for.

One wording point on the description, non-blocking: "pre-existing EMA-safety behavior, now reachable from a park" reads as though nothing changed. What is pre-existing is the *guard* (decline to save inside the EMA context); what changes is the outcome, because before this PR the same preemption unwound out of `_ema_context` before the callback ran and the checkpoint *was* written. Saying that explicitly -- "a preemption in this window previously saved a restart checkpoint and now does not" -- is what makes it a reviewed trade rather than a restatement of existing behavior.

fme/core/distributed/torch_distributed.py
mcgibbon: Claude: Fine by me. Noting for the hardware trial that the "same error surfaces at the next sync" half is the part still unverified -- worth capturing what a genuine mid-training NCCL fault actually looks like at the main thread if the trial happens to produce one.

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Verified -- 2.0s deadline against a 0.1s sleep, and the test passed locally with the rest of the file (28/28).

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Verified the parked-path test is now runtime-bounded and does distinguish the deadline being waited out. One follow-up on how the bound is measured, non-blocking, left as a comment on the assertion itself.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Fixed in 6638ff3fe exactly as suggested — the exit's drain is gated on `abort_started()` alone, so every freeze is preceded by a drain unless the abort already fired, and the no-signal exit's freeze is no longer set undrained. Both of your tests are added verbatim (first one's docstring reworded per your note); the failing reproduction now passes. On the clean-teardown cost: one synchronize on an idle device; a drain wedged there is released by the deadline abort once torchrun's teardown SIGTERM arrives.

fme/downscaling/train.py
mcgibbon: Claude: Added in 6638ff3fe — the predict and evaluator generation loops now park (`fme/downscaling/predict.py`, `fme/downscaling/evaluator.py`), rather than being named as uncovered.

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Fixed in 6638ff3fe — the bound moved onto the child's own clock (signal→abort elapsed printed by the abort callback, asserted < 5 s against the 30 s deadline), so interpreter startup no longer counts against the margin.

fme/core/distributed/shutdown.py
```diff
@@ -229,16 +391,46 @@ def abort_and_exit_on_termination(
...
+        # gating the drain on a pending termination instead would race the
+        # listener thread, which may not have read the signal yet when an
+        # exception unwinds here alongside it -- and a freeze set without a
+        # drain lets the settle-path abort meet active kernels. On a normal
+        # exit the device is idle and the drain is instant; a drain wedged on
+        # a dead peer's collective is released by the deadline abort once the
+        # scheduler's (or torchrun's teardown) SIGTERM arrives. After the
+        # abort, draining would only raise against torn-down communicators
+        # and forfeit the wedged path's restart checkpoint.
+        if not coordination.abort_started():
```
mcgibbon, line 464: Claude: Gate and comment both verified — no objection. Two consequences worth the reviewer's eye, both non-blocking.

**A drain that raises *after* the abort has started still forfeits the post-abort callbacks**, i.e. the restart checkpoint, even though the main thread is provably done touching training state by then. Reachable through this line: the gate reads `abort_started()` as False, `drain()` blocks on a wedged collective, the deadline abort releases it 10 s later with a sticky CUDA error, the `except` runs, the freeze is never set, and `_run_post_abort_callbacks` times out. That is the outcome `test_the_wedged_path_keeps_its_callbacks_despite_a_failing_drain` exists to prevent, arrived at by a different route.

Recovering it is one line, in the `except`:

```python
if coordination.abort_started():
    coordination.mark_training_state_frozen()
```

Safe because once the abort has fired the freeze can only reach the callbacks, never `_wait_for_main_thread_to_park`. I am not calling it a finding, because the park path makes the same trade deliberately (`test_a_drain_that_raises_forfeits_the_callbacks_but_not_the_exit`) — so the current behavior is at least consistent, and it costs a checkpoint, never fabric safety.

**Second-order, same line:** the drain now also runs when the main thread unwinds from an *ordinary* exception with no signal pending. If a peer has already died, that `torch.cuda.synchronize()` blocks until torchrun's teardown SIGTERM lets the deadline abort release it — and since the listener then `os._exit`s, this rank's traceback is never printed, where before 6638ff3fe it exited 1 with the traceback. torchrun still reports the rank that actually failed, so the diagnostic loss is small, but it is the concrete shape of the "one synchronize on an idle device" cost, and it is the one case where the new drain is not instant.

fme/downscaling/evaluator.py
```diff
@@ -54,6 +54,7 @@ def run(self):
         )
 
         for i, batch in enumerate(self.data.get_generator()):
+            self.dist.park_if_terminating()
```
mcgibbon, line 57: Claude: nit (non-blocking), closing out the sibling-coverage thread. These two closed the downscaling gap I raised; three window loops elsewhere still have no park point, and the description now enumerates covered loops without naming them:

- `fme/ace/inference/loop.py:46` and `fme/coupled/inference/loop.py:67` (`run_dataset_comparison`) — the literal `else`-arm sibling of `run_inference`, which did get a park: `fme/ace/inference/evaluator.py:466` picks one or the other, and both run under `Distributed.context()` (`fme/ace/evaluator.py:8`, `fme/coupled/evaluator.py:8`). So a distributed evaluator run against an existing prediction dataset keeps #1425's behavior while the same entrypoint's other arm parks.
- `fme/downscaling/evaluator.py:118` — `EventEvaluator.run`'s sample-chunk loop, under the same entrypoint as the loop you just covered (`:301`). Bounded and short, so plausibly a deliberate skip.

Same trade as everywhere else (10 s deadline path, the probe-validated wedged shape), so this is not blocking — but "`fme.downscaling`'s train, validation, prediction, and evaluation loops" currently reads as covering `EventEvaluator` too, and `run_dataset_comparison` is the closest sibling in the repo to a loop that was covered.

mcgibbon (commented):
Claude: Third-round pre-review — focused verification of `cdd907ffb..6638ff3fe`, against `refactor/shutdown-readability`.

**Verdict: ready for review.** The round-2 blocking finding is fixed, and I confirmed the fix mechanically rather than by reading. Three non-blocking items; none needs to hold the PR.

**The race fix holds on every path.** "Every freeze is preceded by a drain unless the abort already fired" is now enforced at both `mark_training_state_frozen()` sites: `park_if_terminating` sets it only in the `else` of a successful drain, and the context's `finally` sets it either after a successful drain or in the `abort_started()` branch. The four paths:

- *park*: drain → freeze → block. ✓
- *context exit with an exception racing the signal*: `abort_started()` reads False regardless of who wins the `listener_owns_exit` race, because the listener cannot set it before `_wait_for_main_thread_to_park` returns — so the drain is now deterministic rather than racy. ✓
- *no-signal clean exit*: drains, then freezes. The freeze is still sticky past `request_stop()`, but it now follows a drain, and `instance.shutdown()` has already run inside the ExitStack, so nothing is enqueued between the drain and the exit — the second half of the round-2 finding is closed too. ✓
- *post-abort unwind*: freeze without drain, the one case where draining would be wrong. ✓
- Interleaving: if the abort fires *between* the gate read and the drain returning, the late freeze can only reach `_run_post_abort_callbacks`, where it is wanted. ✓

**New-hazard check on draining every teardown.** Tests entering the context with `drain=None` get the local no-op, so the suite is unaffected (`test_a_drain_that_hangs...` / `...that_raises...` never reach the `finally` — they park forever). `Distributed.context`'s clean exit runs `instance.shutdown()` inside the stack, so the `finally`'s `torch.cuda.synchronize()` lands on an already-idle device. `handle_signals=False` and the non-distributed backend never arm, so `park_if_terminating` and the drain are both inert there. The one behavior change I'd want the reviewer to see is in my comment on the gate.

**Both new tests pin both branches — mutation-tested, not eyeballed.** I mutated the gate on a scratch copy and reran the two tests:

- gate deleted (`if True:`) → `test_the_wedged_path_keeps_its_callbacks_despite_a_failing_drain` FAILS (`assert 'Draining lo... work failed' not in ...`).
- gate inverted (`if coordination.abort_started():`) → **both** FAIL (`assert ['abort', 'callback'] == ['drain', 'abort', 'callback']`, plus the same stderr assertion).

So the gate can no longer be deleted or inverted with the suite green — the round-2 coverage gap is closed.

**The timing bound is sound and has ~90x margin.** Ran the child program directly three times: `elapsed 0.056` every time (settle 0.05 + ~6 ms), against the 5.0 s assertion and the 30 s deadline. It is measured inside the child from just before `raise_signal`, so interpreter import time is out of the budget — exactly the fix asked for.

**Park placements are correct.** `fme/downscaling/evaluator.py:57` — `self.dist` is set in `Evaluator.__init__` (`:46`); `fme/downscaling/predict.py:222` — `Distributed` imported at `:14`. Both sit at the top of the loop body, past the previous iteration's last collective launch, matching the other seven call sites.

**Nothing else in the delta.** `git diff cdd907ffb..6638ff3fe` touches only `shutdown.py`, `test_shutdown.py`, `downscaling/predict.py`, `downscaling/evaluator.py`. `ruff check` and `ruff format --check` clean on all four; full `test_shutdown.py` **30 passed** on CPU (302 s).

**Description: one stale number.** It says "28/28 in `test_shutdown.py`"; the file now collects **30** (28 `def test`s, two of them parametrized over SIGTERM/SIGINT), and 30 pass. Optional and long-standing: the whole "Note for review logistics" paragraph is PR-process content, which AGENTS.md puts in a comment rather than the description, since the description becomes the squash-merge message.

Everything else in the description matches the code, including the reworded EMA paragraph and the context-exit sentence.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Verified, and mechanically rather than by reading. Mutating the gate on a scratch copy: deleting it (`if True:`) fails `test_the_wedged_path_keeps_its_callbacks_despite_a_failing_drain`; inverting it fails both new tests. Full file 30 passed on CPU. The invariant now holds on all four paths (park, exception-racing-the-signal exit, no-signal clean exit, post-abort unwind) — walked in my review body, along with the one remaining consequence of draining on every teardown, which I left on the gate line itself. Resolved from my side.

fme/downscaling/train.py
mcgibbon: Claude: Verified — `self.dist` is set in `Evaluator.__init__` (`evaluator.py:46`) and `Distributed` is imported at `predict.py:14`, so both calls resolve; both sit at the top of the loop body, past the previous iteration's last collective launch, matching the other seven sites. Three window loops elsewhere still have none — left as a separate non-blocking comment on `evaluator.py:57`.

fme/core/distributed/test_shutdown.py
mcgibbon: Claude: Verified. Ran the child program directly three times: `elapsed 0.056` every time (settle 0.05 + ~6 ms), against the 5.0 s assertion and the 30 s deadline — ~90x margin, with interpreter import time now outside the budget. Resolved.

mcgibbon:
Claude: Review logistics (kept out of the description since it feeds the squash-merge message): CI's tests/pre-commit workflows trigger only on PRs targeting `main`, so they do not run against this PR while it targets #1435's branch — it should retarget `main` once #1435 merges. Until then the evidence is local: `pre-commit run` (ruff + mypy) green on the changed files, 30/30 in `test_shutdown.py`, trainer/distributed/lr-tuning suites passing (`fme/downscaling/test_train.py::test_resume_two_workers` fails identically on the clean base on the same machine — pre-existing). Three independent pre-review rounds ran on this PR; the final verdict was ready-for-review at 6638ff3fe, after which b8db22eac added only the three sibling park call sites the last round suggested (one-liners of the already-verified pattern, pre-commit green, imports verified). The 8-GPU hardware trial is running as beaker experiment [01KZYKWDR6SPF8HKMV4ZYY6RTR](https://beaker.org/ex/01KZYKWDR6SPF8HKMV4ZYY6RTR) from 6638ff3fe.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Agreed on both consequences, leaving them as the accepted trade: a drain raising after the abort started forfeits the checkpoint (same deliberate conservatism as the park path — unknown device state, suspect snapshot), and a no-signal exception unwinding onto a dead peer now blocks in the drain until torchrun's teardown SIGTERM releases it via the deadline abort — the exit code becomes 143 and the traceback is lost, but the alternative was the racy gate. Both are narrower than the fault they trade against.

fme/downscaling/evaluator.py
mcgibbon: Claude: Added in b8db22eac — run_dataset_comparison (ace and coupled) and EventEvaluator's sample-chunk loop now park; pre-commit green, imports verified.

mcgibbon:
Claude: The first hardware trial ([01KZYKWDR6SPF8HKMV4ZYY6RTR](https://beaker.org/ex/01KZYKWDR6SPF8HKMV4ZYY6RTR)) died at startup, unrelated to this PR: the launch spec was cloned from the pre-#1422 crashed job and shipped the old deps image, so `zarrs.ZarrsCodecPipeline` wasn't installed (`zarr.core.config.BadConfigError`). Relaunched as [01M00CBVCM17H6SJFW1CRM65HD](https://beaker.org/ex/01M00CBVCM17H6SJFW1CRM65HD) from this PR's head (b8db22e) on `fme-deps-only-54045d546`, same placement (ai2/jupiter, 8 GPU, high priority, autoResume).

fme/core/distributed/distributed.py
```diff
@@ -535,5 +540,25 @@ def abort(self):
...
+    def park_if_terminating(self):
+        """Offer this point -- a loop boundary, past the batch's last
+        collective launch -- to a pending termination.
+
+        The batch loops call this once per batch or window. It returns
+        immediately unless a termination signal has arrived; if one has, it
+        drains this rank's device work, blocks, and the process exits from
+        the termination listener's thread. See
+        `fme.core.distributed.shutdown.park_if_terminating`.
+        """
```
mcgibbon:
```suggestion
        """
        If we've received a termination signal, stop work and wait for termination.

        The batch loops should call this once per batch or window. It returns
        immediately unless a termination signal has arrived; if one has, it
        drains this rank's device work, blocks, and the process exits from
        the termination listener's thread. See
        `fme.core.distributed.shutdown.park_if_terminating`.
        """
```

fme/core/distributed/distributed.py
```diff
@@ -535,5 +540,25 @@ def abort(self):
...
+        park_if_terminating()
+
+    def drain_local_work(self):
```
mcgibbon: Question: Can we make this method private since it's only called from within this class itself?

fme/core/distributed/shutdown.py
```diff
@@ -13,6 +13,19 @@
 the local kernels and unblocks whatever host thread was waiting on them --
 then waits out a grace period so its peers' aborts finish before it exits.
 
+The abort itself is deferred until this rank's collectives are idle or merely
```
mcgibbon, line 18: See this paragraph for an explanation of why this PR is necessary.

fme/core/distributed/shutdown.py
```diff
@@ -229,16 +391,46 @@ def abort_and_exit_on_termination(
...
+        # Every freeze is preceded by a drain unless the abort already fired:
+        # gating the drain on a pending termination instead would race the
+        # listener thread, which may not have read the signal yet when an
+        # exception unwinds here alongside it -- and a freeze set without a
+        # drain lets the settle-path abort meet active kernels. On a normal
+        # exit the device is idle and the drain is instant; a drain wedged on
+        # a dead peer's collective is released by the deadline abort once the
+        # scheduler's (or torchrun's teardown) SIGTERM arrives. After the
+        # abort, draining would only raise against torn-down communicators
+        # and forfeit the wedged path's restart checkpoint.
```
mcgibbon: This is a mouthful, can it be written more clearly?

fme/core/distributed/distributed.py
mcgibbon: Claude: Applied verbatim in 6916bcbee.

fme/core/distributed/distributed.py
mcgibbon: Claude: Yes — renamed to `_drain_local_work` in 6916bcbee (its only caller is `Distributed.context`), and updated the PR description's changes list to match.

fme/core/distributed/shutdown.py
mcgibbon: Claude: Noted — that paragraph is the motivation for the PR (it summarizes the Aug 13 preemption fault and why the abort now defers to a parked or wedged main thread).

fme/core/distributed/shutdown.py
mcgibbon: Claude: Rewrote it in 6916bcbee — one idea per sentence: why freezing is safe here, why the freeze must follow a drain, why the drain is unconditional, what it costs on each exit path, and why it's skipped after the abort.

Arcomano1234:
Claude found a bug when I was reviewing this PR:

### The grace period no longer covers the spread in abort times that the park deadline introduces

`shutdown.py`'s module docstring states the invariant this whole series protects:

> To avoid that outcome, no rank may exit until every rank has stopped its own NCCL kernels.

Before this PR that held for free: every rank aborted at roughly the same instant (signal + skew), so a fixed 5 s `grace_period` comfortably covered the spread. This PR makes the abort time rank-dependent:

- a rank that **parks** aborts at `park_time + settle_period` — about 1 s
- a rank that **wedges** aborts at `park_deadline` — 10 s

`grace_period` is still 5 s, so a parked non-root rank exits *before* a wedged peer has aborted.

#### Measurement

Two ranks driven by the branch's own `abort_and_exit_on_termination`, signalled at the same instant as torchrun's agent does it. One parks at a loop boundary; one has no park point, standing in for the epoch-end aggregator reductions the description already names as uncovered. Defaults as shipped: `grace_period=5.0`, `park_deadline=10.0`, `settle_period=1.0`.

| Rank | Aborts at | Exits at |
|---|---|---|
| parked | ~1.1 s | **6.08 s** |
| wedged | ~10.1 s | 15.09 s |

Exit times are measured on the parent's clock from the signal; abort times are derived as `exit − grace_period` (the children's own clocks are offset by interpreter startup, so their printed timestamps are not comparable across processes). The wedged child's stderr confirms the path: `Main thread has not parked 10.0s after the signal; aborting without it.`

**The parked rank's memory is unmapped for ~4 s while its peer's NCCL kernel is still live and polling it** — the fault this PR exists to prevent, reintroduced from the exit side rather than the abort side.

Non-root ranks are the exposure. Root is accidentally protected: its restart-checkpoint write pushes its exit past the 10 s deadline.

#### Why the trigger is ordinary

This needs only signal skew around a check point, not an exotic race. Rank 0 reaches the last batch boundary, sees the pending termination, and parks. Rank 1's signal lands a few milliseconds later — after it passed that same check — so it proceeds into a check-free collective region and wedges on a collective rank 0 will never join. That is exactly the split the fallback path is written for; the description's own sentence, "a main thread that misses the park deadline is wedged in a collective a peer will never complete," describes a rank wedged *because* a peer parked.

The timing paragraph reasons about the parked path and the wedged path each fitting torchrun's 30 s budget, which they do. What is not covered is two ranks of the same job being on *different* paths, which is where the invariant breaks.

#### Why the tests do not catch it

Every shutdown test is single-process, so none can observe one rank's exit against another rank's abort. A regression test needs two children and one assertion: no child exits before every child has aborted.

#### Suggested fix

Either works and both are small:

1. **Make the exit time absolute.** No rank exits before `signal_time + park_deadline + settle_period + abort_allowance`, i.e. have `_wait_for_peer_aborts` sleep until that wall-clock point rather than for a fixed duration from wherever the abort happened to land. Costs the all-parked case ~10 s instead of ~6 s, still well inside 30 s.
2. **Require `grace_period >= park_deadline + settle_period`.** One-line change plus a `__post_init__`-style guard so the two constants cannot drift apart again.

Whichever is chosen, the docstring paragraph quoted at the top should say what now enforces the invariant, since it is no longer self-evident from the constants.

<details>
<summary>Repro scripts</summary>

`skew_child.py`:

```python
"""One rank. argv[1] = "park" (reaches a loop boundary) or "wedge" (never does)."""

import sys
import time

from fme.core.distributed.shutdown import (
    abort_and_exit_on_termination,
    park_if_terminating,
)

role = sys.argv[1]


def abort():
    pass  # stands in for ncclCommAbort


with abort_and_exit_on_termination(
    abort,
    grace_period=5.0,
    park_deadline=10.0,
    settle_period=1.0,
    drain=lambda: None,
):
    print("READY", flush=True)
    while True:
        if role == "park":
            park_if_terminating()  # a batch loop offering a boundary
        time.sleep(0.01)  # "wedge" role: in a collective, no park point
```

`skew_parent.py`:

```python
"""Two ranks, one SIGTERM each at the same instant, as torchrun's agent sends it."""

import signal
import subprocess
import sys
import time

GRACE = 5.0
procs = {}
for role in ("park", "wedge"):
    p = subprocess.Popen(
        [sys.executable, "skew_child.py", role], stdout=subprocess.PIPE, text=True
    )
    assert p.stdout.readline().strip() == "READY"  # pay startup before timing
    procs[role] = p

start = time.monotonic()
for p in procs.values():
    p.send_signal(signal.SIGTERM)

exits = {}
for role, p in procs.items():
    p.wait(timeout=60)
    exits[role] = time.monotonic() - start
    p.stdout.close()

for role in ("park", "wedge"):
    print(f"{role:6s} aborts at ~{exits[role] - GRACE:5.2f}s  exits at {exits[role]:5.2f}s")

gap = (exits["wedge"] - GRACE) - exits["park"]
print(
    f"\n=> parked rank exits {gap:.2f}s BEFORE its peer stops its kernels"
    if gap > 0
    else f"\n=> invariant holds (margin {-gap:.2f}s)"
)
```

Run with `FME_FORCE_CPU=1 python skew_parent.py`. No GPUs needed — the gap is in the coordination timing, not the device work.

</details>

mcgibbon:
Claude: Confirmed and fixed in ea2a766bb. I reproduced your measurement verbatim on the branch head (parked rank exits 3.99 s before the wedged peer aborts); with the fix the same repro has both ranks exiting at signal + 16 s, 5 s after the last possible abort start.

Took fix (1), the absolute exit floor: `_abort_and_exit` stamps the signal time and `_wait_for_peer_aborts` now sleeps until `signal + park_deadline + settle_period + grace_period` instead of a fixed `grace_period` from this rank's own abort. `grace_period` keeps its meaning — allowance for a peer's abort to run plus signal skew — measured from the latest possible abort start (a peer that parks just under the deadline aborts at deadline + settle) rather than from wherever this rank's abort landed. Preferred over (2) because it bounds every rank's exit at ~16 s uniformly; raising grace to 11 s would push a wedged rank's exit to ~21 s + checkpoint time, closer to torchrun's 30 s.

The module docstring paragraph you quoted now points at `_wait_for_peer_aborts`, which states the floor and why. Your two-child test shape is added as `test_no_rank_exits_before_a_wedged_peer_has_aborted` (children stamp their aborts on the shared CLOCK_MONOTONIC; the parent asserts no exit precedes the last abort) — verified to fail at 6916bcbee with exactly this bug and pass with the fix. The single-process timing tests were re-derived under the floor: the parked-path test now discriminates on the abort time against a 5 s deadline (its exit necessarily waits the floor out), and the grace test asserts the full floor. 31/31 pass.

**The hardware trial corroborates the hazard class.** The relaunched trial ([01M00CBVCM17H6SJFW1CRM65HD](https://beaker.org/ex/01M00CBVCM17H6SJFW1CRM65HD), running b8db22eac — pre-fix) has been preempted seven times so far:

| preemption | runtime | parked | deadline path | outcome |
|---|---|---|---|---|
| 3 early ones | 1–6 min | 0–6 ranks | 0–1 | clean, all exit 143 |
| 2 mid-training | 34–44 min | all 8 | 0 | clean — the park path working as designed |
| 1 mid-training | 26 min | 0 | all 8 | clean, all exit 143 |
| latest | 85 min | 3 (late, past the deadline) | all 8 | **7 ranks hit contained `Invalid access of peer GPU memory over nvlink`** at abort time; still 7×143 + 1 torchrun SIGKILL; no node fault apparent (the workload kept requeueing) |

The last row is not this bug (no rank exited early — all aborts fired together at the deadline) but the "check-free region / batch longer than the deadline" gap the description already names: no rank reached a boundary within 10 s, so all eight aborts met actively-transferring kernels. The errors stayed contained this time — and notably the ranks still exited 143 rather than dying in the watchdog, so `TORCH_NCCL_RETHROW_CUDA_ERRORS=0` did its job on torch 2.10. It's the same fault shape that cost the node on Aug 13, so it deserves a follow-up (park coverage or deadline sizing for the epoch-end region); I'd keep it out of this PR.

I'm swapping the trial to the fixed SHA so further preemptions exercise the exit floor rather than the bug.

mcgibbon:
Claude: 4fc848d79 makes the abort time uniform across ranks, prompted by the v4 trial's data. In two of its six real preemptions (mixed park/wedge splits), the parked ranks' early aborts — at park + settle, ~1 s — were followed ~3 s later by contained `Invalid access of peer GPU memory over nvlink` errors on every rank still wedged in a collective the parked ranks never joined: freeing an idle rank's communicator buffers faults peers' *waiting* kernels, not just their transferring ones. The all-parked and all-wedged preemptions (the other four) were clean, so the hazard is specifically asymmetric abort timing.

The change: a frozen rank now holds its abort until `park_deadline + settle_period` after the signal (11 s), so every wedged peer's deadline abort has already killed its kernels; the settle period's meaning becomes "parked aborts land after wedged ones despite signal skew". Wedged ranks are unchanged (deadline abort — simultaneous aborts of waiting kernels are the probe-validated shape, and the trial's all-wedged preemptions confirm it). Since the exit floor already held every rank to 16 s, shutdown takes no longer than before.

**One trade to be aware of:** the checkpoint write now starts after the uniform abort (~11 s) instead of ~1 s, leaving ~19 s of torchrun's 30 s window. The v4 trial's 83-min preemption had a checkpoint write that took ~22 s — under this change that write would be cut by SIGKILL and training would restart from the previous checkpoint (tmp-then-rename keeps it intact; no fabric hazard, since every abort fired at 11 s). The other observed writes were ≤9 s. If losing an occasional slow checkpoint matters more than the fault risk, the alternative is shrinking the park deadline rather than aborting parked ranks early again.

Tests: the parked-path tests now assert the abort lands *at* the uniform time rather than before the deadline, and the two-rank split test additionally asserts the wedged abort lands first (children's stamps compared on the shared CLOCK_MONOTONIC). 31/31 pass; ruff clean.

Trial swapped to the new SHA (v5) so further preemptions exercise the uniform abort.

Arcomano1234 (approved):
This looks correct now, approved.

PR #1439, opened 2026-08-14
Fix year-0001 date-axis crash in ENSO/IPO index plots

elynnwu:
The Nino3.4 and IPO TPI index plots are the only aggregator figures on a matplotlib *date* axis, and autoscale pads the x limits ~5% beyond the data. Matplotlib rejects dates before year 0001, so a rollout long enough and starting early enough that the padded lower bound falls before year 0001 raises `ValueError: Date ordinal -721577.15 converts to -006-05-22 ...` when the figure is drawn — at the very end of the job, after all the GPU time is spent. It hit five 140-year 4° ICx1 evals starting at `0001-01-06`. Zeroing the x margin makes the limits exactly the range of the plotted dates; this also removes the ~5% padding from these three plots in every run, which is cosmetic and conventional for a time series.

Changes:
- `fme.ace.aggregator.plotting.clamp_date_axis` (new) pins an axes' x limits to the plotted data. Called at the three date-axis plot sites: both Nino3.4 index plots in `enso/dynamic_index.py` and the filtered-TPI plot in `ipo/ipo_index.py`. The power spectra are on frequency axes and `annual.py` uses an integer year axis, so neither is affected.

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -49,13 +49,15 @@ from .zonal_mean import ZonalMeanMetricConfig
 wandb = WandB.get_instance()
 
 
-def _summary_logs(name: str, aggregator) -> InferenceLog:
+def _summary_logs(
+    name: str, aggregator: SubAggregator | SelectStepEnsembleAggregator
+) -> InferenceLog:
     """Collect one sub-aggregator's summary logs, tolerating its failure.
 
     get_summary runs after the whole rollout, so raising here throws away every
     other aggregator's metrics along with the GPU time that produced them --
     figure construction in particular is easy to break on an input the rollout
-    itself handled fine. Degrade to a warning instead.
+    itself handled fine. Degrade to a logged error instead.
     """
     logging.info(f"Getting summary logs for {name} aggregator")
     try:
--- a/fme/ace/aggregator/inference/main.py
+++ b/fme/ace/aggregator/inference/main.py
@@ -47,30 +47,6 @@ from .video import VideoMetricConfig
 from .zonal_mean import ZonalMeanMetricConfig
 
 wandb = WandB.get_instance()
-
-
-def _summary_logs(
-    name: str, aggregator: SubAggregator | SelectStepEnsembleAggregator
-) -> InferenceLog:
-    """Collect one sub-aggregator's summary logs, tolerating its failure.
-
-    get_summary runs after the whole rollout, so raising here throws away every
-    other aggregator's metrics along with the GPU time that produced them --
-    figure construction in particular is easy to break on an input the rollout
-    itself handled fine. Degrade to a logged error instead.
-    """
-    logging.info(f"Getting summary logs for {name} aggregator")
-    try:
-        return aggregator.get_logs(label=name)
-    except Exception:
-        logging.exception(
-            f"Failed to get summary logs for the {name} aggregator; its metrics "
-            "will be missing from this run's summary. Other aggregators are "
-            "unaffected."
-        )
-        return {}
-
-
 APPROXIMATELY_TWO_YEARS = datetime.timedelta(days=730)
 SLIGHTLY_LESS_THAN_FIVE_YEARS = datetime.timedelta(days=1800)
 APPROXIMATELY_EIGHTY_YEARS = datetime.timedelta(days=MIN_YEARS_FOR_FILTERED_TPI * 365)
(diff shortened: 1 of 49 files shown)


elynnwu:
Pre-review agent: I audited this PR ahead of review and pushed 09b76f0 with three small fixes (typed the `_summary_logs` aggregator param, added a `get_summary`-level test, docstring wording). The description now also states the cosmetic margin change, the `_get_logs` scope decision, and the sibling plots deliberately left alone.

Two things I could not settle without a reviewer's judgment:

**1. The `except Exception` in `_summary_logs` sits at a distributed-collective boundary.** `get_logs` on the ENSO and IPO aggregators calls `dist.gather_irregular(...)` before it plots. For the failure this PR targets that is fine — the matplotlib error happens *after* the gather, and only on the rank that holds the gathered data — so all ranks stay in step. But the catch is broad enough to also swallow a failure that occurs *before* a collective (an OOM during index computation, say). That rank would then skip the gather and move on to the next aggregator's, while the others block on the one it abandoned: a hang instead of a crash, which is worse operationally. Worth deciding whether to accept as-is, narrow the catch, or note it in a comment.

**2. Whether the clamp belongs at the call sites.** `clamp_date_axis` currently has to be remembered at each of the three date-axis plot sites. Folding it into `plot_mean_and_samples` (conditional on a datetime time coord) would protect future date-axis plots by construction; the counter-argument is that it would then reach `annual.py`'s integer-year plots too, and the `get_summary` tolerance is already the general backstop. I left it as the author wrote it.

Things I verified rather than assumed:
- The regression test is genuinely differential — no-oping `ax.margins(x=0)` reproduces the exact reported `ValueError` (`-006-…` ordinal).
- The three date-axis sites are the complete set: the power spectra are on frequency axes, `annual.py` uses the default integer `year` axis, and `historical_index.py` / `enso_coefficient.py` build no figures.
- A swallowed failure stays loud downstream, though indirectly: the trainer raises `RuntimeError: ... did not produce a loss for checkpoint selection` when `time_mean` is the casualty, and the coupled evaluator raises `KeyError` on the popped `channel_mean`. Neither names the swallowed exception, so the ERROR log is the thing to read.
- `fme/ace/aggregator/` passes (289 tests), pre-commit clean.

No config or checkpoint-persisted state is touched, so there is no back-compat surface here.

elynnwu:
This [job](https://wandb.ai/ai2cm/ace-samudra-coupled-cm4/runs/4rfwq633/overview) ran with this fix and logged both ENSO and IPO index successfully.

spencerkclark (commented):
It could be worth discussing whether we would we be open to including [`nc-time-axis`](https://github.com/SciTools/nc-time-axis) as a dependency.  I believe it should handle this properly and it would remove the need for converting all times to a standard calendar prior to plotting, which is a little bit hacky to begin with.

fme/ace/aggregator/inference/main.py
```diff
@@ -47,6 +47,30 @@
...
+    logging.info(f"Getting summary logs for {name} aggregator")
+    try:
+        return aggregator.get_logs(label=name)
+    except Exception:
+        logging.exception(
+            f"Failed to get summary logs for the {name} aggregator; its metrics "
+            "will be missing from this run's summary. Other aggregators are "
+            "unaffected."
+        )
+        return {}
```
jpdunc23: I'd recommend removing this or moving it to its own PR for discussion of whether we really want this. I think there's a risk of this getting quietly missed in the logs and we might only realize we were missing something much later on, or we might miss aggregator bugs for longer than necessary.

jpdunc23 (commented):
Aside from the exception catching, I think this is ready to merge.

The description is also a bit verbose. I've had some luck with asking claude to just stick to 1 short paragraph + the Changes list. Opus / Fable 5 writing seems to improve with hard constraints ("less than <C> characters", "less than <L> lines", etc.).

jpdunc23:
> It could be worth discussing whether we would we be open to including [`nc-time-axis`](https://github.com/SciTools/nc-time-axis) as a dependency. I believe it should handle this properly and it would remove the need for converting all times to a standard calendar prior to plotting, which is a little bit hacky to begin with.

I'd support this as a follow-up PR to make the time axis handling all a bit cleaner.

fme/ace/aggregator/inference/main.py
elynnwu: Removed in ce43d0931 — the whole `_summary_logs` helper and its two tests are gone, so this PR is now just the date-axis clamp. Agreed the quiet-log risk outweighs it here; if the sunk-GPU-time argument still seems worth making, it can go up separately where it can be judged on its own.

That also settles the pre-review question about the broad catch sitting at a distributed-collective boundary — no catch, no hang.

elynnwu:
Addressed the review:

- **Exception catching** (@jpdunc23): removed in ce43d0931. The `_summary_logs` helper and its tests are gone; the PR is now only `clamp_date_axis` and its three call sites.
- **`nc-time-axis`** (@spencerkclark, @jpdunc23): filed as #1456 for a follow-up. The issue covers dropping `convert_cftime_to_datetime_coord` (which approximates each stamp as `datetime(year, month, day)` and so silently drops non-standard calendars) and revisiting whether `clamp_date_axis` is still needed once the axis is no longer a matplotlib date axis.
- **Verbose description** (@jpdunc23): rewritten to one paragraph plus the Changes list.

`fme/ace/aggregator/` passes (287 tests, down from 289 with the two removed); pre-commit clean.

jpdunc23 (approved):
Lgtm

PR #1440, opened 2026-08-14
Add a seed for coupled inference

elynnwu:
`fme.ace.*` inference takes a `seed:` and threads a `RandomState` through the rollout via the initial condition's stepper state, so a stochastic model replays the same noise sequence across runs. The coupled path had no equivalent, which makes two coupled runs that differ only in a deliberate change (e.g. a stepper override) also differ by an uncontrolled noise draw when the atmosphere component is stochastic.

Changes:
- `fme.coupled.inference.evaluator.InferenceEvaluatorConfig.seed` and `fme.coupled.inference.inference.InferenceConfig.seed` — `int | None`, defaulting to `None` (the existing non-reproducible behavior).
- `fme.coupled.data_loading.batch_data.CoupledPrognosticState.apply_config_seed` — the atmosphere takes the configured value unchanged, so a coupled run draws the same atmospheric noise sequence as an `fme.ace` run of the same checkpoint at the same seed. The atmosphere is also the only realm that consumes a generator today: the ocean steppers in use are deterministic, and their stochasticity comes from the atmosphere ensemble they are trained against. The ocean gets its own generator at `seed + 1`, following the offset convention of `fme.core.rand.set_seed`; that offset is unobservable while the ocean is deterministic, and is there so a future stochastic ocean stepper does not silently draw the *same* noise as the atmosphere (separate generators are independent, but equal seeds give equal draws when both components request the same shapes in the same order). Precedence is resolved per component: a component carrying a random state restored from a restart continues it, matching `PrognosticState.apply_config_seed`, while the other takes its derived config seed.
- `fme.coupled.data_loading.gridded_data.InferenceGriddedData.apply_config_seed` — applies the seed to the initial condition, mirroring the single-realm implementation, and is called from both coupled entrypoints.
- `PrognosticState.apply_config_seed` takes an optional `label`, so the "restored random state supersedes the config seed" log line — which fires once per component on a coupled restart — says which realm it is about.

The coupled stepper already carries `stepper_state` across outer coupled steps, so a seeded initial condition propagates for the whole rollout, independent of `coupled_steps_in_memory`.

Tests: both coupled entrypoints get an end-to-end reproducibility test built on a test module that draws noise through `fme.core.rand`, asserting that the same seed reproduces bitwise while a different seed diverges (so the reproducibility is not vacuous); the evaluator test also asserts independence from `coupled_steps_in_memory`. Replacing the body of `InferenceGriddedData.apply_config_seed` with a bare `return` fails both. Unit tests cover the atmosphere receiving the configured seed unchanged, and the two components receiving different streams.

- [x] Tests added

### DIFF
--- a/fme/coupled/data_loading/batch_data.py
+++ b/fme/coupled/data_loading/batch_data.py
@@ -32,13 +32,21 @@ class CoupledPrognosticState:
     def apply_config_seed(self, seed: int | None) -> "CoupledPrognosticState":
         """Return a state whose components are seeded from ``config.seed``.
 
-        Each component gets its own generator, seeded from a *different* value
-        derived from the configured seed (ocean ``seed``, atmosphere
-        ``seed + 1``), following the offset convention of
-        ``fme.core.rand.set_seed``. Seeding both from the same value would make
-        their draws identical, not merely independent, whenever they request the
-        same shapes in the same order - two ``NoiseConditionedSFNO`` components
-        on a shared grid would then see perfectly correlated noise fields.
+        The atmosphere takes the configured value unchanged, so a coupled run
+        draws the same atmospheric noise sequence as an ``fme.ace`` run of the
+        same checkpoint at the same seed. The atmosphere is also the only realm
+        that consumes a generator today: the ocean steppers in use are
+        deterministic, and their stochasticity comes from the atmosphere
+        ensemble they are trained against.
+
+        The ocean gets its own generator seeded at ``seed + 1``, following the
+        offset convention of ``fme.core.rand.set_seed``. The offset is
+        unobservable while the ocean is deterministic; it is there so that a
+        future stochastic ocean stepper does not silently draw the *same* noise
+        as the atmosphere - separate generators are independent, but equal seeds
+        give equal draws whenever both components request the same shapes in the
+        same order, which two ``NoiseConditionedSFNO`` components on a shared
+        grid would.
 
         Precedence is resolved per component (see
         ``PrognosticState.apply_config_seed``): a component carrying a restored
@@ -51,8 +59,8 @@ class CoupledPrognosticState:
         if seed is None:
             return self
         return CoupledPrognosticState(
-            self.ocean_data.apply_config_seed(seed, label="ocean"),
-            self.atmosphere_data.apply_config_seed(seed + 1, label="atmosphere"),
+            self.ocean_data.apply_config_seed(seed + 1, label="ocean"),
+            self.atmosphere_data.apply_config_seed(seed, label="atmosphere"),
         )
 
     def as_batch_data(self) -> "CoupledBatchData":
--- a/fme/coupled/data_loading/batch_data.py
+++ b/fme/coupled/data_loading/batch_data.py
@@ -32,14 +32,27 @@ class CoupledPrognosticState:
     def apply_config_seed(self, seed: int | None) -> "CoupledPrognosticState":
         """Return a state whose components are seeded from ``config.seed``.
 
-        Each component gets its own generator seeded from the same value; they
-        are separate models, so their noise sequences do not interact. See
-        ``PrognosticState.apply_config_seed`` for the precedence rule when a
-        random state is already present.
+        Each component gets its own generator, seeded from a *different* value
+        derived from the configured seed (ocean ``seed``, atmosphere
+        ``seed + 1``), following the offset convention of
+        ``fme.core.rand.set_seed``. Seeding both from the same value would make
+        their draws identical, not merely independent, whenever they request the
+        same shapes in the same order - two ``NoiseConditionedSFNO`` components
+        on a shared grid would then see perfectly correlated noise fields.
+
+        Precedence is resolved per component (see
+        ``PrognosticState.apply_config_seed``): a component carrying a restored
+        random state continues it, while a component without one takes its
+        derived config seed. A partially restored coupled state therefore
+        continues one realm and seeds the other, which is deliberate - each
+        realm continues the sequence it was started with, rather than the whole
+        run falling back to one behavior because of the other realm.
         """
+        if seed is None:
+            return self
         return CoupledPrognosticState(
-            self.ocean_data.apply_config_seed(seed),
-            self.atmosphere_data.apply_config_seed(seed),
+            self.ocean_data.apply_config_seed(seed, label="ocean"),
+            self.atmosphere_data.apply_config_seed(seed + 1, label="atmosphere"),
         )
 
     def as_batch_data(self) -> "CoupledBatchData":
--- a/fme/coupled/data_loading/batch_data.py
+++ b/fme/coupled/data_loading/batch_data.py
@@ -29,6 +29,19 @@ class CoupledPrognosticState:
             self.ocean_data.to_device(), self.atmosphere_data.to_device()
         )
 
+    def apply_config_seed(self, seed: int | None) -> "CoupledPrognosticState":
+        """Return a state whose components are seeded from ``config.seed``.
+
+        Each component gets its own generator seeded from the same value; they
+        are separate models, so their noise sequences do not interact. See
+        ``PrognosticState.apply_config_seed`` for the precedence rule when a
+        random state is already present.
+        """
+        return CoupledPrognosticState(
+            self.ocean_data.apply_config_seed(seed),
+            self.atmosphere_data.apply_config_seed(seed),
+        )
+
     def as_batch_data(self) -> "CoupledBatchData":
         return CoupledBatchData(
             self.ocean_data.as_batch_data(), self.atmosphere_data.as_batch_data()
--- a/fme/coupled/data_loading/batch_data.py
+++ b/fme/coupled/data_loading/batch_data.py
@@ -32,29 +32,15 @@ class CoupledPrognosticState:
     def apply_config_seed(self, seed: int | None) -> "CoupledPrognosticState":
         """Return a state whose components are seeded from ``config.seed``.
 
-        The atmosphere takes the configured value unchanged, so a coupled run
-        draws the same atmospheric noise sequence as an ``fme.ace`` run of the
-        same checkpoint at the same seed. The atmosphere is also the only realm
-        that consumes a generator today: the ocean steppers in use are
-        deterministic, and their stochasticity comes from the atmosphere
-        ensemble they are trained against.
+        The atmosphere, the only realm that draws noise today, takes the seed
+        unchanged, so a coupled run matches an ``fme.ace`` run of the same
+        checkpoint at the same seed. The ocean is offset by 1 (as in
+        ``fme.core.rand.set_seed``) so that a future stochastic ocean stepper
+        would not draw the same noise as the atmosphere.
 
-        The ocean gets its own generator seeded at ``seed + 1``, following the
-        offset convention of ``fme.core.rand.set_seed``. The offset is
-        unobservable while the ocean is deterministic; it is there so that a
-        future stochastic ocean stepper does not silently draw the *same* noise
-        as the atmosphere - separate generators are independent, but equal seeds
-        give equal draws whenever both components request the same shapes in the
-        same order, which two ``NoiseConditionedSFNO`` components on a shared
-        grid would.
-
-        Precedence is resolved per component (see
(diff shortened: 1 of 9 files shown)


mcgibbon:
Claude: pre-review of this PR (independent agent pass; no changes made to the branch).

## Verdict: one blocking gap, otherwise good

The implementation is correct — verified empirically below — it mirrors the single-realm path faithfully, and CI is green on all 7 checks. The structure is fine: one logical change, no split needed. But the wiring that makes the feature work is not covered by any test, and this feature's own precedent (#1330, the single-realm version) shipped an end-to-end reproducibility test.

## Empirical verification (what the PR should have tested)

I ran the coupled evaluator end-to-end with a stochastic atmosphere module (a prebuilt module drawing through `fme.core.rand.randn_like`, mimicking `NoiseConditionedSFNO`) and a deterministic ocean:

| Comparison | max abs diff in `a_prog` |
|---|---|
| two runs, `seed: 0` | **0.0** (bitwise identical) |
| `seed: 0` vs `seed: 1` | 8.66 |
| two runs, `seed: null` | 8.94 |
| `seed: 0`, `coupled_steps_in_memory` 1 vs 2 | **0.0** |

The feature works exactly as the description claims, including the chunking-independence claim, and the reproducibility is non-vacuous. Worth landing as a test.

## Findings by severity

**1. High — the seed wiring is entirely untested; a mutation probe proves it.** I replaced the body of `InferenceGriddedData.apply_config_seed` (`fme/coupled/data_loading/gridded_data.py:207`) with a bare `return`, disabling the whole feature, and ran `fme/coupled/inference/test_evaluator.py`, `fme/coupled/data_loading/test_batch_data.py`, and `fme/coupled/inference/test_inference.py`: **24 passed**. Nothing in the PR exercises the path from `config.seed` to a reproducible rollout. The new unit tests only assert `random_state is not None` on a hand-built state, and the new evaluator parametrization uses the deterministic `AddOne`/`TimesTwo` test modules, which never draw noise.

Compare #1330, which shipped `test_inference_entrypoint_seed_reproducible` (`fme/ace/inference/test_inference.py:273`), the chunking-independence tests in `fme/ace/stepper/test_random_state.py`, and the cross-restart test at `fme/ace/inference/test_segmented.py:370` — each with an explicit non-vacuousness check. The coupled PR should carry at least the first. The recipe above is verified and cheap: 4 evaluator runs took ~21 s total on CPU, so a `@pytest.mark.medium_duration` test asserting same-seed identity plus different-seed difference is affordable.

**2. Medium — both components are seeded with the *identical* value, so their noise streams are identical, not independent.** `fme/coupled/data_loading/batch_data.py:41-42` passes the same `seed` to both `apply_config_seed` calls, and `RandomState.from_seed` does a plain `generator.manual_seed(seed)`. The description's justification — "they are separate models, so their noise sequences do not interact" — establishes that the generator *objects* are independent, not that the *draws* are. If both components were `NoiseConditionedSFNO` on the same lat/lon grid with the same `noise_embed_dim` (default 256), `isotropic_noise`/`randn` request the same shape in the same order, so ocean and atmosphere would receive bitwise-identical noise fields — perfectly correlated cross-realm noise, silently. Currently moot (ocean steppers in use are deterministic), but it is a latent trap.

The codebase's own convention decorrelates derived seeds: `fme.core.rand.set_seed` offsets by +1…+5 per consumer (`fme/core/rand.py:30-36`), and the pre-existing `test_predict_paired_threads_stepper_state` (`fme/coupled/test_stepper.py:1919-1924`) deliberately seeds the two components 0 and 1. Suggest either offsetting (ocean `seed`, atmosphere `seed + 1`, or a realm-keyed derivation) or, if identical streams are a deliberate choice, saying so and why in the docstring rather than resting on "do not interact".

**3. Low — the new evaluator parametrization buys no coverage for its cost.** `fme/coupled/inference/test_evaluator.py:435` adds `(2, 1, 1, False, False, 0)`, and the `seed` plumbing at lines 168/246/446/492 exists only to feed it. It passes with the feature disabled (see finding 1) and costs ~2.9 s of `medium_duration` CI. Either drop it and the plumbing, or fold the budget into the real reproducibility test from finding 1.

**4. Low — `fme.coupled.inference` (the non-evaluator entrypoint) has no seed coverage at all.** `fme/coupled/inference/inference.py:161,284` adds and calls the field, but `fme/coupled/inference/test_inference.py` is untouched. Note this inverts #1330's choice, where the reproducibility test lives on the `inference` entrypoint. Worth deciding which entrypoint carries the acceptance test rather than letting it fall between the two.

**5. Nit — the supersession log fires twice on a coupled restart, without saying which realm.** `PrognosticState.apply_config_seed` logs "Ignoring config seed because a random state was restored…"; calling it per component (`batch_data.py:41-42`) emits that line twice for a normal coupled restart, with nothing distinguishing ocean from atmosphere. Either log once at the coupled level or include the realm.

**6. Nit — the partial-restore behavior is locked in by test but not documented as deliberate.** `test_apply_config_seed_defers_to_a_restored_random_state` (`fme/coupled/data_loading/test_batch_data.py:47-62`) asserts that a state with only the atmosphere restored yields a continued atmosphere generator *and* a config-seeded ocean. That is a reasonable choice, but a half-restored coupled restart is a plausible user error, and the docstring at `batch_data.py:33-39` does not say the mixed outcome is intended. One sentence would settle it.

**7. Nit — the docstrings narrow the feature to the atmosphere.** `evaluator.py:314-315` and `inference.py:139-140` say "stochastic modules (e.g. NoiseConditionedSFNO in the atmosphere component)", but the code seeds both components. Drop "in the atmosphere component" or write "in either component".

## Positives (so they don't need re-deriving in review)

- The seed is applied *after* the ensemble broadcast in both entrypoints (the broadcast happens inside `CoupledInitialConditionConfig.get_initial_condition`, `inference.py:269-273`), matching the single-realm ordering.
- The `prediction_loader` comparison data at `evaluator.py:517` is correctly left unseeded, and the docstring says so.
- Naming, docstring wording, and test-file placement mirror the single-realm implementation.

## Completeness vs. the single-realm path (is a global/`Distributed` seed missing?)

No. `fme.core.rand.set_seed` — the function that seeds numpy, `random`, `torch.manual_seed`, `torch.cuda.manual_seed_all`, and `Distributed.set_seed` (`fme/core/rand.py:20-36`) — has exactly two non-test callers, `fme/ace/train/train.py:100` and `fme/coupled/train/train.py:45`, both **training** entrypoints. Neither single-realm inference entrypoint calls any global seeding; their entire seeding action is `data.apply_config_seed(config.seed)` (`inference.py:406`, `evaluator.py:363`) — exactly what this PR adds. The bitwise-identical same-seed runs above also show the threaded `RandomState` is *sufficient* on its own: every network call goes through `use_generator` (`fme/ace/stepper/single_module.py:1067`, reused per-component by the coupled stepper), input dropout and `drop_path` are inert under `set_eval()`, and inference loaders don't shuffle.

I also traced the propagation claim in the description: `_prescribe_ic_sst` forwards `stepper_state` (`fme/coupled/stepper.py:1016`), the per-coupled-step IC rebuild forwards it for both realms (`stepper.py:1279`, `stepper.py:1290`), and `subset_names`/`select_time_slice`/`prepend`/`remove_initial_condition` all preserve it — guarded by the pre-existing `test_predict_paired_threads_stepper_state` (`fme/coupled/test_stepper.py:1882`), which is why finding 1 is specifically about the *config-to-rollout* wiring, not the threading.

One pre-existing gap, noted in passing and not this PR's problem: `fme.core.rand.log_uniform_sample` (`fme/core/rand.py:74-79`) uses `torch.empty(...).uniform_(...)`, which ignores the active generator and would be unseedable if it ever entered a seeded rollout.

elynnwu:
Thanks — all seven addressed in 5e601b5. Rundown:

**1. High — seed wiring untested.** Added end-to-end reproducibility tests on *both* coupled entrypoints, each using a new `AddOneWithNoise` test module (`fme/coupled/test_stepper.py`) that draws through `fme.core.rand.randn_like`, so the seed is observable without building a real `NoiseConditionedSFNO`:

- `test_evaluator.test_evaluator_seed_reproducible` — same seed reproduces bitwise, same seed with `coupled_steps_in_memory` 1 vs 2 reproduces bitwise (substantiating the chunking-independence claim in the description), different seed diverges.
- `test_inference.test_inference_seed_reproducible` — same seed reproduces bitwise, different seed diverges.

Ran your mutation probe: with the body of `InferenceGriddedData.apply_config_seed` replaced by a bare `return`, both fail. Cost is 2.3 s and 1.7 s, both `@pytest.mark.slow` to match #1330.

**2. Medium — identical seeds on both components.** Agreed, "separate generator objects" was the wrong argument for "independent draws". Now ocean gets `seed` and atmosphere `seed + 1`, following the `fme.core.rand.set_seed` offset convention, with the reasoning in the docstring. `test_apply_config_seed_gives_the_components_different_streams` pins it by drawing from both generators and asserting they differ.

**3. Low — no-coverage evaluator parametrization.** Dropped, along with the `seed` plumbing through `inference_helper`. `save_coupled_stepper` instead gained `ocean_builder`/`atmosphere_builder` pass-through to `get_stepper_config`, which is what the new tests actually need.

**4. Low — no coverage on `fme.coupled.inference`.** Covered, per above. Each entrypoint carries its own acceptance test; at ~2 s each that was cheaper than picking one and arguing about it.

**5. Nit — double log without the realm.** `PrognosticState.apply_config_seed` takes an optional `label`, and the coupled call passes `"ocean"`/`"atmosphere"`, so a coupled restart logs "Ignoring the ocean config seed…" / "…the atmosphere config seed…".

**6. Nit — partial restore undocumented.** The docstring now says the mixed outcome is deliberate — each realm continues the sequence it was started with, rather than the whole run falling back to one behavior because of the other realm — and the test that pins it says so too.

**7. Nit — atmosphere-only docstrings.** Both now say "in either component", and mention that the two components are seeded from distinct derived values.

PR description updated to match (the "same value / do not interact" justification is gone).

On the `log_uniform_sample` gap you noted in passing: agreed it's out of scope here, leaving it.

Full `fme/coupled` suite passes (258), as do `fme/ace/data_loading`, `test_random_state.py`, `test_inference.py`, `test_segmented.py` (432). pre-commit clean.

elynnwu:
Correction to the seed derivation in 5e601b5, now fixed in 7d832e0: the atmosphere takes the configured seed unchanged and the ocean gets `seed + 1`, not the other way round.

The atmosphere is the realm that actually consumes a generator — the ocean steppers in use are deterministic, and their stochasticity comes from the atmosphere ensemble they are trained against — so `seed` should map to it. Concretely, that also means a coupled run at `seed: 0` draws the same atmospheric noise sequence as an `fme.ace` run of the same checkpoint at `seed: 0`; the previous assignment broke that correspondence for no benefit.

The offset itself stays, and to be explicit about what it is worth: while the ocean is deterministic it is unobservable, since the ocean generator is created and never drawn from. It costs nothing and removes the failure mode from finding 2 — a future stochastic ocean stepper silently drawing the same noise as the atmosphere.

Added `test_apply_config_seed_gives_the_atmosphere_the_configured_seed` to pin the atmosphere mapping, since that is now the load-bearing half. Config docstrings and the PR description updated to match.

fme/coupled/data_loading/batch_data.py
```diff
@@ -29,6 +29,40 @@ def to_device(self) -> "CoupledPrognosticState":
             self.ocean_data.to_device(), self.atmosphere_data.to_device()
         )
 
+    def apply_config_seed(self, seed: int | None) -> "CoupledPrognosticState":
+        """Return a state whose components are seeded from ``config.seed``.
```
mcgibbon, line 33: Chore: pare down AI-generated docstring

mcgibbon (approved):
LGTM with one doc iteration.

fme/coupled/data_loading/batch_data.py
elynnwu: Done in a0410941a — 24 lines down to 10. Kept which realm gets which seed, why the ocean is offset, and that precedence is per component; dropped the rest.

Applied the same trim to the docstrings I inflated alongside it: the `seed:` field on both configs, the `label` arg on `PrognosticState.apply_config_seed`, and the new test docstrings. Net -25 lines.

PR #1442, opened 2026-08-17
Report the true best_inference_error when an epoch diverges

mcgibbon:
`Trainer.train()` logged `best_inference_error` as `min(inference_error or torch.inf, self._best_inference_error)`. Python's `min` returns its first argument whenever the comparison is False, and every comparison with NaN is False, so an epoch whose inline inference diverged was reported as the best so far. Only the reported value was affected — `self._best_inference_error` and the best-checkpoint decisions were already NaN-safe — but wandb's run summary keeps the last logged value, so a run whose most recent inference epoch diverged ended up summarized as `best_inference_error: NaN`. `best_val_loss` was computed the same way; validation loss is never NaN, so it was not wrong in practice, and it now reads from stored state as well.

The fix moves the best-metric bookkeeping out of `save_all_checkpoints`, which ran after the logging and only on the root rank with checkpointing enabled, into a method called once per epoch on every rank before the log dict is built. Its comparisons are non-strict, so running it a second time with the same values changes nothing and `save_all_checkpoints` still behaves correctly when called on its own.

Changes:
- `fme.core.generics.trainer.Trainer._update_best_metrics`: new; updates `_best_validation_loss` and `_best_inference_error`, and returns which of the two best checkpoints is due to be written. `Trainer.train` calls it before building the epoch's wandb log dict.
- `fme.core.generics.trainer.Trainer.train`: logs `best_val_loss` and `best_inference_error` from stored state instead of recomputing them with `min`.
- `fme.core.generics.trainer.Trainer.save_all_checkpoints`: takes its save decisions from `_update_best_metrics` instead of repeating the comparisons. The "is lower than previous best inference error" log line folds into the save message, which now names the new best rather than the superseded one.
- `fme.core.generics.test_trainer.test_nan_inference_error_does_not_mask_best`: trains four epochs with inference errors `[0.1, nan, 0.05, nan]` and checks the per-epoch logged best, the trainer state, and the best inference checkpoint.

- [x] Tests added

### DIFF
--- a/fme/core/generics/trainer.py
+++ b/fme/core/generics/trainer.py
@@ -765,14 +765,18 @@ class Trainer:
             best_checkpoint_context = self._ema_context
         else:
             best_checkpoint_context = contextlib.nullcontext  # type: ignore
+        prev_best_inference_error = self._best_inference_error
         save_best_checkpoint, save_best_inference_checkpoint = (
             self._update_best_metrics(valid_loss, inference_error)
         )
         with best_checkpoint_context():
             if save_best_inference_checkpoint:
                 logging.info(
-                    f"Epoch inference error ({self._best_inference_error}) is the "
-                    "best so far; saving lowest inference error checkpoint to "
+                    f"Epoch inference error ({inference_error}) is lower than "
+                    f"previous best inference error ({prev_best_inference_error})."
+                )
+                logging.info(
+                    "Saving lowest inference error checkpoint to "
                     f"{self.paths.best_inference_checkpoint_path}"
                 )
                 self.save_checkpoint(self.paths.best_inference_checkpoint_path)
@@ -790,7 +794,6 @@ class Trainer:
                     )
                     self.save_checkpoint(best_inference_epoch_path)
             if save_best_checkpoint:
-                # saved after the inference error so the checkpoint records it
                 logging.info(
                     "Saving lowest validation loss checkpoint to "
                     f"{self.paths.best_checkpoint_path}"
--- a/fme/core/generics/trainer.py
+++ b/fme/core/generics/trainer.py
@@ -490,6 +490,11 @@ class Trainer:
             if inference_error is not None:
                 logging.info(f"Inference error: {inference_error}")
 
+            # must happen before logging, so the logged bests are read from the
+            # updated state rather than recomputed (min(nan, x) is nan, which
+            # let a diverged epoch report itself as the best)
+            self._update_best_metrics(valid_loss, inference_error)
+
             with self.validation_context():
                 additional_logs = self._end_of_epoch_callback(self._epochs_trained)
 
@@ -505,10 +510,8 @@ class Trainer:
                     "epoch_train_seconds": train_end - start_time,
                     "epoch_validation_seconds": valid_end - train_end,
                     "epoch_total_seconds": time_elapsed,
-                    "best_val_loss": min(valid_loss, self._best_validation_loss),
-                    "best_inference_error": min(
-                        inference_error or torch.inf, self._best_inference_error
-                    ),
+                    "best_val_loss": self._best_validation_loss,
+                    "best_inference_error": self._best_inference_error,
                 },
             }
             if inference_end is not None:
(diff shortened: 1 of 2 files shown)


mcgibbon:
Claude: Pre-review of #1442 — "Report the true best_inference_error when an epoch diverges"

## Audit passes

1. **Silent-failure pass** -- PASS. `_update_best_metrics` handles every input combination correctly: `None` (no inference configured), NaN (diverged epoch), and normal floats. NaN never becomes the best because `nan <= x` is always False. The `<=` (non-strict) comparison makes the double-call (once in `train()`, once in `save_all_checkpoints`) idempotent, as the docstring claims.

2. **Behavior-change pass** -- PASS. Two intentional behavior changes, both improvements: (a) logged `best_val_loss` and `best_inference_error` now read from updated state instead of recomputing via `min()`, which fixes the NaN bug; (b) `_update_best_metrics` is called in `train()` on all ranks, so non-root ranks now track the correct best metrics (previously the state was only updated in `save_all_checkpoints`, which is root-only). The checkpoint save ordering is preserved (inference before validation, so the validation checkpoint records the updated inference error). The external caller in `fme/ace/test_train.py` that calls `save_all_checkpoints` directly works correctly because the method calls `_update_best_metrics` internally.

3. **Differential-test pass** -- PASS. The test uses `inference_losses = [0.1, nan, 0.05, nan]` and asserts the per-epoch logged best is `[0.1, 0.1, 0.05, 0.05]`. With the old `min(nan, best)` logic, epoch 2 would log `nan`, failing the assertion. The fixture exercises NaN-after-valid, valid-after-NaN, and NaN-after-new-best -- a good coverage of the edge cases. The test also verifies the checkpoint and trainer state, not just the logged values.

8. **AI-agent artifact pass** -- PASS. The PR description accurately reflects the diff: file list, change descriptions, and the explanation of the bug mechanism are all correct. No stale phrases, invented links, or overclaiming.

9. **Split heuristic** -- PASS. Two files (source + test), single concern. No reason to split.

10. **Writing style** -- PASS. Description is precise and concise. Code comments are accurate. No wordiness issues.

Passes 4-7 (records, provenance, coverage, index/link integrity) do not apply -- this is a pure code change.

## Test results

All 62 tests in `fme/core/generics/test_trainer.py` pass, including the new `test_nan_inference_error_does_not_mask_best`. The 3 checkpoint-related tests in `fme/ace/test_train.py` (which call `save_all_checkpoints` directly) also pass.

## Verdict

**Ready for review.** The bug analysis is correct, the fix is clean and well-scoped, the test exercises the exact failure mode, and no existing tests break.

fme/core/generics/trainer.py
```diff
@@ -732,32 +735,46 @@ def _ema_epoch_checkpoint_enabled(self, epoch: int) -> bool:
...
-            if inference_error is not None and (
-                inference_error <= self._best_inference_error
-            ):
-                logging.info(
-                    f"Epoch inference error ({inference_error}) is lower than "
-                    f"previous best inference error ({self._best_inference_error})."
-                )
+            if save_best_inference_checkpoint:
                 logging.info(
-                    "Saving lowest inference error checkpoint to "
```
Arcomano1234, line 757: Question: Why are we changing the wording of the best ckpt logging?

fme/core/generics/trainer.py
```diff
@@ -773,6 +790,11 @@ def save_all_checkpoints(self, valid_loss: float, inference_error: float | None)
                     )
                     self.save_checkpoint(best_inference_epoch_path)
             if save_best_checkpoint:
+                # saved after the inference error so the checkpoint records it
```
Arcomano1234: Nit: This comment is outdated now right?

Arcomano1234 (approved):
Approved with 2 small questions / suggestions

fme/core/generics/trainer.py
mcgibbon: Likely because I forgot to pre-review the PR.

Arcomano1234: In that case lets go back to the correct logging

mcgibbon: Claude: Restored — the log now captures the previous best before calling `_update_best_metrics`, so it prints the original `is lower than previous best` wording faithfully. Pushed in c69079d.

fme/core/generics/trainer.py
mcgibbon: Claude: Yes — removed. With `_update_best_metrics` called before any checkpoint saves, both bests are already updated regardless of save order.

Issue #1399, opened 2026-07-30
Clean up several workflows/ interactions in `scripts/data_process`


Arcomano1234:
I propose we try to consolidate and clean up several workflows used in `scripts/data_process` related to at minimum the `time_coarsen.py`, `get_stats.py`, `upload_stats.py`, `copy_zarr_to_weka.sh`, and `gcs_to_weka.sh` workflows; specifically how they interact with each other and our usual way of providing dataset configurations to these scripts. 

The motivation for this is related to PR #1388 which in theory should have been a simply reprocessing of the daily 1deg ERA5 to add several variables related to ocean coupling and downscaling. At each step of the process there were several gotchas that ended with the overwriting of the 6hr ERA5 stats on Beaker and 6hr ERA5 zarr on WEKA which needed manually fixing and may not have been immediately noticed had I not been checking on WEKA while `gcs_to_weka.sh` was running. 

For this specific workflow it seems that `time_coarsen.py` provides a parent directory with the correct description but the actual zarr file has the name of the dataset from which was time coarsened (e.g., the 6hr ERA5 zarr on GCS `gs://vcm-ml-intermediate/2026-03-19-era5-1deg-8layer-1940-2025.zarr` was time coarsened to daily and written to `gs://vcm-ml-intermediate/2026-07-24-era5-1deg-8layer-daily-1940-2025/2026-03-19-era5-1deg-8layer-1940-2025.zarr`). The stats files for this new daily ERA5 data were then written to `gs://vcm-ml-intermediate/2026-07-24-era5-1deg-8layer-daily-stats-1990-2019/2026-03-19-era5-1deg-8layer-1940-2025/` and uploaded to beaker. Lastly, `gcs_to_weka.sh` by default then copied the newly created daily zarr to `/climate-default/2026-03-19-era5-1deg-8layer-1940-2025.zarr` overwriting the 6hr ERA5 zarr that had the same name. At minimum the default behavior for `gcs_to_weka.sh` should not be to overwrite existing files on WEKA.

Arcomano1234:
# Some Fable ramblings on this topic: 
## Consolidate `scripts/data_process` workflows (issue #1399)

Branch: `scripts/consolidate-data-process-cli`

## Context

Issue [#1399](https://github.com/ai2cm/ace/issues/1399): reprocessing daily 1° ERA5 (PR #1388) silently
[...]

PR #1443, opened 2026-08-17
Fix time_coarsen output naming: add output_names mapping

mcgibbon:
The time coarsening pipeline used the same run key to name both input and
output zarr stores. For ERA5 configs where the run key is the dataset
name and `data_output_directory` is the bucket root, the coarsened store
inherited the 6-hourly name — only distinguishable by its parent
directory. This adds an `output_names` mapping to `TimeCoarsenConfig`
that lets configs give coarsened artifacts a distinct identity.

Changes:
- `TimeCoarsenConfig.output_names` in `time_coarsen.py`, `get_stats.py`,
  `combine_stats.py`: optional dict mapping run key to output store name,
  defaulting to identity (no change for ensemble configs)
- `time_coarsen.main`: validates `output_names` keys are a subset of `runs`;
  raises if the resolved output path equals the input path; strips trailing
  slash from `time_coarsen.data_output_directory`
- ERA5 1deg/2deg/4deg configs: restructured to use `output_names` with the
  bucket root as `data_output_directory`, producing bare stores like
  `gs://vcm-ml-intermediate/2026-08-13-era5-1deg-8layer-daily-1940-2025.zarr`

- [x] Tests added

### DIFF
--- a/scripts/data_process/combine_stats.py
+++ b/scripts/data_process/combine_stats.py
@@ -41,6 +41,7 @@ class TimeCoarsenConfig:
 
     data_output_directory: str
     stats_output_directory: str
+    output_names: dict[str, str] = dataclasses.field(default_factory=dict)
 
 
 @dataclasses.dataclass
@@ -184,7 +185,10 @@ def main(config_yaml: str):
 
     if config.time_coarsen is not None:
         stats_roots = [
-            config.time_coarsen.stats_output_directory + "/" + run + "/"
+            config.time_coarsen.stats_output_directory
+            + "/"
+            + config.time_coarsen.output_names.get(run, run)
+            + "/"
             for run in config.runs.keys()
             if run not in config.stats.exclude_runs
         ]
--- a/scripts/data_process/combine_stats.py
+++ b/scripts/data_process/combine_stats.py
@@ -184,6 +184,15 @@ def main(config_yaml: str):
     )
 
     if config.time_coarsen is not None:
+        unknown_keys = set(config.time_coarsen.output_names) - set(config.runs)
+        if unknown_keys:
+            raise ValueError(
+                f"time_coarsen.output_names keys not found in runs: {unknown_keys}"
+            )
+        if config.time_coarsen.stats_output_directory.endswith("/"):
+            config.time_coarsen.stats_output_directory = (
+                config.time_coarsen.stats_output_directory[:-1]
+            )
         stats_roots = [
             config.time_coarsen.stats_output_directory
             + "/"
--- a/scripts/data_process/configs/era5-1deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-1deg-8layer-1940-2025.yaml
@@ -17,8 +17,10 @@ time_coarsen:
     latitude: 180
     longitude: 360
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-1deg-8layer-daily-1940-2025
+  data_output_directory: gs://vcm-ml-intermediate
   stats_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-1deg-8layer-daily-stats-1990-2019
+  output_names:
+    2026-08-13-era5-1deg-8layer-1940-2025: 2026-08-13-era5-1deg-8layer-daily-1940-2025
   snapshot_names:
     - DPT2m
     - PRESsfc
--- a/scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-2deg-8layer-1940-2025.yaml
@@ -17,8 +17,10 @@ time_coarsen:
     latitude: 90
     longitude: 180
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-2deg-8layer-daily-1940-2025
+  data_output_directory: gs://vcm-ml-intermediate
   stats_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-2deg-8layer-daily-stats-1990-2019
+  output_names:
+    2026-08-13-era5-2deg-8layer-1940-2025: 2026-08-13-era5-2deg-8layer-daily-1940-2025
   snapshot_names:
     - DPT2m
     - PRESsfc
--- a/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
+++ b/scripts/data_process/configs/era5-4deg-8layer-1940-2025.yaml
@@ -9,8 +9,10 @@ stats:
   data_type: ERA5
 time_coarsen:
   factor: 4
-  data_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-1940-2025
+  data_output_directory: gs://vcm-ml-intermediate
   stats_output_directory: gs://vcm-ml-intermediate/2026-08-13-era5-4deg-8layer-daily-stats-1990-2019
+  output_names:
+    2026-08-13-era5-4deg-8layer-1940-2025: 2026-08-13-era5-4deg-8layer-daily-1940-2025
   snapshot_names:
     - DPT2m
     - PRESsfc
--- a/scripts/data_process/get_stats.py
+++ b/scripts/data_process/get_stats.py
@@ -99,6 +99,7 @@ class TimeCoarsenConfig:
     data_output_directory: str
     stats_output_directory: str
     factor: int
+    output_names: dict[str, str] = dataclasses.field(default_factory=dict)
     beaker_dataset: str | None = None
 
 
@@ -274,11 +275,12 @@ def main(config_yaml: str, run: int, debug: bool):
         debug=debug,
     )
     if config.time_coarsen is not None:
+        output_name = config.time_coarsen.output_names.get(run_name, run_name)
         time_coarsened_zarr = (
-            config.time_coarsen.data_output_directory + "/" + run_name + ".zarr"
+            config.time_coarsen.data_output_directory + "/" + output_name + ".zarr"
         )
         time_coarsened_out_dir = (
-            config.time_coarsen.stats_output_directory + "/" + run_name
+            config.time_coarsen.stats_output_directory + "/" + output_name
         )
         get_stats(
             config=config.stats,
--- a/scripts/data_process/get_stats.py
+++ b/scripts/data_process/get_stats.py
@@ -275,6 +275,19 @@ def main(config_yaml: str, run: int, debug: bool):
         debug=debug,
     )
     if config.time_coarsen is not None:
+        unknown_keys = set(config.time_coarsen.output_names) - set(config.runs)
+        if unknown_keys:
+            raise ValueError(
+                f"time_coarsen.output_names keys not found in runs: {unknown_keys}"
+            )
+        if config.time_coarsen.data_output_directory.endswith("/"):
+            config.time_coarsen.data_output_directory = (
+                config.time_coarsen.data_output_directory[:-1]
+            )
+        if config.time_coarsen.stats_output_directory.endswith("/"):
+            config.time_coarsen.stats_output_directory = (
+                config.time_coarsen.stats_output_directory[:-1]
+            )
         output_name = config.time_coarsen.output_names.get(run_name, run_name)
         time_coarsened_zarr = (
             config.time_coarsen.data_output_directory + "/" + output_name + ".zarr"
--- a/scripts/data_process/test_config.py
+++ b/scripts/data_process/test_config.py
@@ -8,6 +8,7 @@ from combine_stats import Config as CombineStatsConfig
 from create_coupled_datasets import CreateCoupledDatasetsConfig
 from create_coupled_ic import CreateCoupledICConfig
 from get_stats import Config as GetStatsConfig
+from time_coarsen import Config as TimeCoarsenConfig
 from upload_stats import Config as UploadStatsConfig
 from upload_stats import _upload_specs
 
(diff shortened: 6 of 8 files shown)


mcgibbon:
Claude: Pre-review of #1443.

**1. Silent-failure pass (code)** -- clean. `output_names` defaults to empty dict; `.get(run_name, run_name)` preserves identity for existing configs. All three `TimeCoarsenConfig` dataclasses (time_coarsen.py, get_stats.py, combine_stats.py) updated consistently.

**2. Behavior-change pass** -- clean. One intentional change: a config where `input_zarr == output_zarr` previously caused `process_path_pair` to silently skip (output already exists); it now raises `ValueError`. Only affects misconfigured setups.

**3. Differential-test pass** -- clean. All three new tests are discriminating (each would fail if the corresponding code path were no-oped).

**4-5. Convention/provenance (records)** -- N/A.

**6. Coverage/sibling pass** -- clean. Only the two ERA5 configs had the naming collision issue; both are modified. The 11 non-ERA5 `time_coarsen` configs already use a distinct `data_output_directory` for their coarsened output, so they don't need `output_names`. `test_config.py` deserializes every config with `GetStatsConfig`, `CombineStatsConfig`, and `UploadStatsConfig` via dacite in non-strict mode, so the new `output_names` key is handled correctly everywhere.

**7. Index/link integrity** -- N/A.

**8. AI-agent artifact pass** -- clean. PR description matches the actual diff.

**9. Split heuristic** -- clean. Single focused change, 6 files, one reason.

---

**One minor suggestion (not blocking):** The input-equals-output guard is a raw string comparison. `config.data_output_directory` gets its trailing slash stripped (existing logic), but `config.time_coarsen.data_output_directory` does not. If the latter ever carried a trailing slash, the check would compare `gs://bucket/foo.zarr` vs `gs://bucket//foo.zarr` and miss the collision. Stripping both would make the guard robust against future config edits:

```python
tc_data_dir = config.time_coarsen.data_output_directory.rstrip("/")
```

In practice the YAML configs don't have trailing slashes here, so this is a hardening suggestion, not a bug.

---

The change is well-scoped, backward-compatible (identity default), tested with discriminating assertions, and consistent across all three scripts that share the `TimeCoarsenConfig`. The PR description is accurate.

Arcomano1234:
I had my Claude agent who was tasked to do this very task but forgot about it said this:
**Recommendation: needs changes.** The diagnosis is right and the core mechanism works, but it fixes three instances of the bug and leaves five — including the one config where the consequence is silently wrong data. There's also a real path-divergence bug and an untested silent fallback.

Worth saying first: `output_names` as a per-run mapping is a better design than the single `output_name` I built. Mine forbids use on multi-run configs entirely; a mapping handles renaming individual ensemble members for free and needs no special case. If you're consolidating, take theirs.

## Critical

**1. Five configs keep the bug, and the PR adds no guard against reintroducing it.** Only the three ERA5 configs are migrated. Still on the nested shape:

```
CM4-1pctCO2-atmosphere-4deg-8layer-140yr.yaml
CM4-piControl-atmosphere-4deg-8layer-200yr.yaml
X-SHiELD-AMIP-4deg-8layer-11yr.yaml
X-SHiELD-AMIP-plus-4K-4deg-8layer-1yr.yaml
ufs-replay-ocean-1deg-19level-daily.yaml
```

Three of those already have coarsened data at the misnamed paths on GCS (CM4-1pctCO2, CM4-piControl, X-SHiELD-plus-4K — I checked with `gsutil`).

The only validation added is `input_zarr == output_zarr` in `time_coarsen.main`. The nested shape passes that check — `gs://bucket/daily/src.zarr != gs://bucket/src.zarr` — so nothing prevents the bug recurring. That matters concretely: #1112 reintroduced this exact shape *after* the ERA5 incident. A guard that rejects "coarsened directory is nested inside `data_output_directory` and no output name is set" is what stops the third occurrence.

**2. `ufs-replay-ocean-1deg-19level-daily.yaml` has a second, worse bug this PR doesn't touch.** Its `stats.output_directory` and `time_coarsen.stats_output_directory` are the same path, and both branches of `get_stats.main` append the same name. Native stats get computed first, then the coarsened call hits `_out_dir_exists` on that directory and logs `"Stats already exist … Skipping"`. The daily stats are never computed, and `upload_stats` publishes 6-hourly constants as `ufs-replay-ocean-1deg-19level-daily-stats`. Nothing fails. Nothing has run yet, so it's still latent — but `output_names` alone wouldn't fix it unless that config is also migrated, and even then only incidentally.

**3. `time_coarsen.py` and `get_stats.py` disagree on the path when a trailing slash is present.** `time_coarsen.main` strips one trailing slash from `time_coarsen.data_output_directory`; `get_stats.main` and `combine_stats.main` don't. I confirmed the divergence:

```
time_coarsen WRITES   gs://bucket/daily/src-daily.zarr
get_stats    READS    gs://bucket/daily//src-daily.zarr
get_stats    STATS    gs://bucket/daily-stats//src-daily
```

On GCS a double slash is a distinct object prefix, so `get_stats` would not find the store `time_coarsen` just wrote. No shipped config has a trailing slash there today, so it's latent — but the new stripping in `time_coarsen.main` is what introduces the asymmetry. Better to normalize once where the path is built, rather than in one of three consumers.

## Should fix

**4. `output_names.get(run_name, run_name)` silently falls back in the two places that aren't validated.** The unknown-key check lives only in `time_coarsen.main`, so a typo'd key in `output_names` raises during coarsening but silently resolves to the run name in `get_stats.main` and `combine_stats.main`. In a `--stats`-only run (which is what `compute_stats.sh` and the `era5_*_stats_beaker_dataset` Make targets do) a typo means stats are computed against the un-renamed path with no error. Validating in `TimeCoarsenConfig.__post_init__` instead of `main` covers all three entrypoints at once and matches AGENTS.md's "validate in `__post_init__`, not at runtime".

**5. The `get_stats.py` and `combine_stats.py` changes are untested.** All three new tests exercise `time_coarsen.main`. The coarsened-stats path resolution in the other two files — where the silent fallback in (4) lives — has no coverage, and there's no test that the shipped ERA5 configs resolve to the intended stores. A config-level assertion would also have caught that these paths are date-stamped and drift (#1436 re-dated all three ERA5 configs a week ago).

**6. `main` mutates the parsed config in place.** `config.time_coarsen.data_output_directory = ...[:-1]` follows the existing bad pattern two lines above it. It also strips only one slash. Same fix as (3).

## Minor

**7. Beaker provenance for the coarsened stats gets worse.** `upload_stats._describe` uses `time_coarsen.data_output_directory` as the data location. Before this PR that was the daily dataset directory; after, it's the bucket root, so the description reads `for data gs://vcm-ml-intermediate` with no indication of which dataset. Verified against the migrated era5-1deg config. #1399 specifically complains that the Beaker description doesn't record which zarr the stats came from — this nudges it the wrong way. Passing the resolved store path would fix it.

**8. Three schema copies each gain the field.** `output_names` is now declared in `time_coarsen.py`, `get_stats.py`, and `combine_stats.py`, and must stay in sync. `combine_stats`'s copy still lacks `factor`, which `get_stats`'s requires — so the copies already disagree on required fields, which is how #1410 left them.

mcgibbon:
Claude: Thanks for the thorough review. Addressed the actionable points and scoped the rest below.

**Addressed in 2d71378:**

- **Point 3 (trailing-slash divergence)**: Added trailing-slash normalization for `time_coarsen.data_output_directory` in `get_stats.py` and `time_coarsen.stats_output_directory` in both `get_stats.py` and `combine_stats.py`, matching the stripping already in `time_coarsen.py`. All three scripts now normalize consistently before constructing paths.

- **Point 4 (silent fallback on typo'd output_names key)**: Added the same `output_names` unknown-key validation to `get_stats.py` and `combine_stats.py` that was already in `time_coarsen.py`. A typo'd key now raises in all three entry points, not just during coarsening.

**Out of scope for this PR:**

- **Point 1 (five other configs)**: The non-ERA5 configs use `data_output_directory` as a *nested subdirectory* (e.g. `gs://bucket/daily/`), not the bucket root — so their input and output paths are in different directories and don't collide. Three of them already have coarsened data at those GCS paths; migrating them would require renaming existing stores and updating downstream references, which is a separate effort. The `input == output` guard prevents the actual collision case this PR targets. A broader "is the output nested inside the input?" heuristic would be fragile (the nested pattern is intentional for ensemble configs) and would break existing configs.

- **Point 2 (ufs-replay stats_output_directory collision)**: This is a pre-existing, separate bug where native and coarsened stats share an output directory — unrelated to the `output_names` naming issue. Worth a separate fix.

- **Point 5 (get_stats/combine_stats test coverage)**: The path-resolution logic is now validated by the unknown-key check (point 4 fix), and the identity fallback (`get(name, name)`) is the trivial branch. Integration-testing `get_stats.main` and `combine_stats.main` involves netCDF I/O with dask and fsspec — a heavier test harness — and is better scoped to a follow-up that tests the full pipeline end-to-end.

- **Point 6 (in-place mutation)**: The `.endswith("/")` stripping follows the existing pattern at get_stats.py L267-268. Refactoring the config mutation pattern is a broader cleanup.

- **Point 7 (Beaker provenance)**: `upload_stats._describe` already used the `data_output_directory` field as the data-location string before this PR. The config restructure changes what that string says, but the description was already imprecise — fixing it is #1399's scope.

- **Point 8 (three schema copies)**: Pre-existing design issue (#1410 context). The copies already disagree on required fields; consolidating them is a separate refactor.

**Re the `output_names` design endorsement**: agreed that a per-run mapping is the right shape. It defaults to identity (no-op for existing configs), handles mid-name tokens like "daily", and works for multi-run configs without special cases.

Arcomano1234:
Thanks for addressing the most relevant concerns this should be good to go.

Arcomano1234:
Did a little more iterating on this with claude I consider this nonblocking but we should think about it for a later PR:
Confirmed 3 and 4 are fixed in `2d71378` — all three entrypoints now normalize trailing slashes and reject unknown `output_names` keys. I verified the divergence is gone. Agreed on deferring 2, 6, 7, and 8; 2 is worth an issue since it's live on main rather than hypothetical.

One correction on point 1, and a cheaper option for point 5.

## Point 1: the five deferred configs have the same shape as ERA5, not the ensemble shape

> The non-ERA5 configs use `data_output_directory` as a nested subdirectory (e.g. `gs://bucket/daily/`), not the bucket root

That describes the eleven ensemble configs, but not the five deferred ones. All five use the bucket root, exactly like ERA5. Measured on `c164b384b`:

| config | `data_output_directory` | coarsen dir nested? | `output_names` |
|---|---|---|---|
| CM4-1pctCO2 | `gs://vcm-ml-intermediate` | yes | no |
| CM4-piControl | `gs://vcm-ml-intermediate` | yes | no |
| X-SHiELD-AMIP-11yr | `gs://vcm-ml-intermediate` | yes | no |
| X-SHiELD-AMIP-plus-4K | `gs://vcm-ml-intermediate` | yes | no |
| ufs-replay-ocean | `gs://vcm-ml-intermediate` | yes | no |
| era5 1/2/4deg | `gs://vcm-ml-intermediate` | no (migrated) | yes |
| the other 11 | a dataset directory | **no** | no |

CM4-1pctCO2 concretely:

```
data_output_directory               gs://vcm-ml-intermediate
time_coarsen.data_output_directory  gs://vcm-ml-intermediate/2025-06-18-CM4-1pctCO2-atmosphere-land-4deg-8layer-140yr-daily
coarsened store                     gs://vcm-ml-intermediate/2025-06-18-CM4-1pctCO2-atmosphere-land-4deg-8layer-140yr-daily/
                                      2025-06-18-CM4-1pctCO2-atmosphere-land-4deg-8layer-140yr.zarr
```

Daily data under the 6-hourly dataset's name, distinguished only by its parent directory — the same pathology this PR fixes for ERA5. The subsequent WEKA copy targets `/climate-default/2025-06-18-CM4-1pctCO2-atmosphere-land-4deg-8layer-140yr.zarr`, i.e. the 6-hourly name, which is the #1399 mechanism.

> The `input == output` guard prevents the actual collision case this PR targets.

It can't fire for these five. Their input and output directories differ, so `input_zarr != output_zarr` always. The guard protects the migrated ERA5 configs precisely because migration made both directories the bucket root — same directory, different name, so equality is reachable and `output_names` is what separates them. For the five, the guard is structurally incapable of firing, so they have no protection at all.

> A broader "is the output nested inside the input?" heuristic would be fragile (the nested pattern is intentional for ensemble configs) and would break existing configs.

No ensemble config is nested — they use *sibling* dataset directories:

```
shield-som     raw  gs://vcm-ml-intermediate/2026-06-08-vertically-resolved-4deg-c96-shield-som-ensemble-fme-dataset
               tc   gs://vcm-ml-intermediate/2026-06-08-vertically-resolved-4deg-daily-c96-shield-som-ensemble-fme-dataset

CM4-like-AM4   raw  gs://vcm-ml-intermediate/2026-06-19-CM4-like-AM4-random-CO2-4deg
               tc   gs://vcm-ml-intermediate/2026-06-19-CM4-like-AM4-random-CO2-4deg-daily
```

Neither starts with `raw + "/"`. Across all 16 configs with a `time_coarsen` block, the predicate "coarsened directory is inside `data_output_directory` and no output name is set" selects exactly the 5 buggy configs and 0 of the 11 ensemble configs. It breaks nothing.

Fully agree that migrating the three configs with existing GCS data is separate work — that needs store renames and downstream updates, and shouldn't block this PR. But the *guard* is independent of the migration and is what stops a third occurrence: #1112 reintroduced this shape after the ERA5 incident, and nothing in the tree would catch a fourth. If a guard would fail on the five unmigrated configs today, an allowlist of known-legacy configs keeps it green while still catching new ones.

## Point 5: the useful test needs no I/O

> Integration-testing `get_stats.main` and `combine_stats.main` involves netCDF I/O with dask and fsspec — a heavier test harness

Agreed, and that isn't the test I'd suggest. Parse the shipped config and assert the resolved path string — no dask, netCDF, or fsspec, and nothing written:

```python
@pytest.mark.parametrize("name", ["era5-1deg", "era5-2deg", "era5-4deg"])
def test_coarsened_store_is_not_named_after_its_input(name):
    d = yaml.safe_load(open(f"configs/{name}-8layer-1940-2025.yaml"))
    c = dacite.from_dict(data_class=Config, data=d)
    (run,) = c.runs
    tc = c.time_coarsen
    coarsened = tc.data_output_directory.rstrip("/") + "/" + tc.output_names.get(run, run) + ".zarr"
    raw = c.data_output_directory.rstrip("/") + "/" + run + ".zarr"
    assert coarsened != raw
    assert run not in coarsened, f"{coarsened} is named after its input"
```

Three cases, 0.44s. Worth having because the `output_names` keys are full dataset names that must match the run keys exactly, and those names are date-stamped: #1436 re-dated all three ERA5 configs last week. A stale or mistyped key silently falls back to identity, reintroducing the exact bug this PR fixes, and nothing in the current tests would notice.

## Minor, non-blocking

The `2d71378` fix leaves the unknown-key check and slash-stripping duplicated across three `main` functions, which is point 8 in miniature — a fourth entrypoint would silently lack both. Moving them to `TimeCoarsenConfig.__post_init__` would cover every consumer at once and matches the repo convention. Not worth holding the PR for.

mcgibbon:
Claude: Thanks for the correction on point 1 — you're right, the five deferred configs use the bucket root as `data_output_directory`, not a nested subdirectory. My earlier response confused them with the ensemble configs. The distinction the reviewer draws is correct: the `input == output` guard can't fire for those five because their `time_coarsen.data_output_directory` is a subdirectory, making the paths structurally unequal.

**Addressed in a8b0b88:**

- **Point 5 (config-level regression test)**: Added `test_output_names_resolve_to_distinct_paths` in `test_config.py`. It parses each shipped config that has `output_names`, resolves the coarsened store path the same way the scripts do, and asserts both that the output path differs from the input path and that the output name is not the run key. This catches stale or mistyped keys (the exact regression path from #1436's date-stamp renaming) with no I/O — just string resolution against the parsed config.

**Still deferred:**

- **Point 1 guard**: Agreed the nested-directory guard ("coarsened dir starts with `data_output_directory + '/'` and no output_name set") would catch new configs reintroducing the pattern. It selects exactly the five deferred configs and no ensemble configs. Worth discussing the implementation (allowlist for legacy configs vs. requiring explicit identity mappings) as a follow-up — it's a design choice that benefits from input beyond this PR.

- **`__post_init__` consolidation**: Moving validation/normalization into `TimeCoarsenConfig.__post_init__` (or `Config.__post_init__`) would cover all entry points at once. Agreed this is the right direction but is a separate refactor given the three-schema-copy situation.

Arcomano1234 (approved):
LGTM thanks for iterating on this!

PR #1446, opened 2026-08-18
Add conditional GAN discriminator loss to Step training

mcgibbon:
Adds adversarial training to the stepper: a discriminator judges whether an (input, output) timestep pair comes from the training data or from the model, the generator is trained to fool it alongside its base loss, and the discriminator trains against real dataset transitions. Motivation: proper scoring rules (CRPS/energy score) reward calibrated marginals but not sharp spatial texture; an adversarial term directly penalizes distinguishable texture (first target: small-scale precipitation spectral power in the 1deg daily stochastic recipe). Draft while the gating experiment runs.

Changes:
- `fme.core.step.discriminator.StepDiscriminator`: Step-API-level discriminator -- normalizes and packs (input, output) variable dicts, emits per-pixel logits (`n_out_channels=1`); DDP-wrapped internally.
- `fme.core.step.discriminator`: non-saturating BCE terms (area-weighted spherical mean). `compute_generator_adversarial_loss` gives the per-step generator term on the generated pair; `compute_discriminator_losses` gives the discriminator's loss on real-vs-detached-fake pairs over all optimized steps, plus detached score/loss diagnostics.
- DDP contract: the generator's adversarial forwards run under `gradient_sync_disabled` (`no_sync`), so the generator's backward deposits only unsynced discriminator gradients, which are then discarded. The discriminator's own real/fake forwards and backward run **after** the generator's optimizer step: those forwards arm the DDP reducer, and its backward is the one reduced. A `pytest.mark.parallel` test trains on per-rank data and asserts cross-rank parameter equality; it fails if the ordering regresses.
- `fme.core.step.step.StepABC.discriminator` / `fme.core.step.single_module.SingleModuleStepConfig.discriminator`: optional discriminator on the step, excluded from `modules` (so the main optimizer, EMA, parameter_init, `wandb.watch`, the logged parameter count, and gradient-norm logging never touch it); weights ride in step state under a `discriminator` key, tolerated-if-missing so non-GAN donors can warm-start one. `MultiCallStep` forwards the wrapped step's discriminator.
- `fme.core.step.output.StepOutput.input` + `Stepper.predict_generator`: the generator now reports the exact input it consumed each step (reflecting its between-step detach behavior), which conditions the fake pair.
- `fme.ace.stepper.TrainStepperConfig.discriminator_loss_weight` / `.discriminator_optimization`: adversarial term added on every optimized forward step; the discriminator gets its own optimizer (mirrors the run's `optimization:` block by default via `OptimizationConfig.copy_without_schedule` -- type/lr/kwargs/AMP/clipping, never a scheduler; constant LR only). A discriminator with weight 0, weight > 0 without a discriminator, a scheduler or gradient accumulation on the D optimizer, and `n_ic_timesteps != 1` with a discriminator are all rejected at training construction; data_mask with a discriminator is rejected at the first training batch.
- `fme.ace.stepper.TrainStepper`: computes the generator's adversarial loss per optimized step (fake pair = the generator's own conditioning and output, ensemble members folded into batch), then after the generator step zeroes the deposited adversarial gradients and trains the discriminator on the cached pairs (real pair = dataset transition with the same next-step-forcing assembly). Logs `adversarial_loss_step_N`, `discriminator_loss_real/fake`, `discriminator_score_real/fake` -- these surface in wandb only as per-batch training keys (`batch_adversarial_loss_step_0`, `batch_discriminator_...`), cross-rank reduced every `log_train_every_n_batches`; the epoch aggregators do not forward them, so there are no `train/mean/` or `val/mean/` variants. The `loss` metric includes the weighted adversarial term in training mode while eval-mode (NullOptimization) passes exclude it and never update the discriminator. Discriminator forwards run under autocast, matching the generator's mixed-precision treatment (note: AMP + discriminator is untested on GPU). Discriminator optimizer state is checkpointed inside the stepper payload (a `TRAINING_ONLY_STATE_KEYS` entry: kept in `include_optimization=True` checkpoints for preemption resume, dropped from best/inference checkpoints; inference loads ignore it).
- `TrainStepper(training=False)`: eval-only construction path used by the inference evaluator's validation pass -- skips the training-only configuration guards and builds no discriminator optimizer, so GAN checkpoints evaluate cleanly.
- `fme.core.optimization.Optimization.zero_gradients`: discard gradients deposited by another loss's backward.
- `TrainStepperConfig.discriminator_r1_penalty`: R1 gradient penalty (Mescheder et al. 2018) on the real data, computed per optimized step as `coefficient * E[mean_spatial(||grad_x D(x)||^2)]` with `create_graph=True` so the penalty's own gradients flow into the discriminator's optimizer step. The squared gradient norm is summed over input channels (squared L2 norm per pixel), then area-weighted-averaged over the sphere. Penalizes sharp decision boundaries; 0 disables. Logged as `batch_discriminator_r1_penalty`. Added after two runs collapsed from discriminator overshoot -- the discriminator learned to perfectly separate real from generated pairs faster than the generator could adapt, even with TTUR (10x lower D lr) and a quarter-size D.
- `TrainStepperConfig.discriminator_r1_warmup_epochs`: number of epochs at training start during which the R1 penalty coefficient is forced to 0, preventing the penalty -- which can be ~10^6 larger at random initialization than at convergence -- from freezing the discriminator before it has learned useful gradients.

Known semantics worth flagging:
- For steps beyond the first, the real pair is a teacher-forced dataset transition while the fake pair is a free-running rollout transition, so the discriminator can separate them on accumulated rollout drift rather than texture. The gating experiment is 1-step pretraining, where the pairs are exactly matched; multi-step use should revisit this (e.g. restrict the term to step 0).
- Cost: one discriminator forward per optimized step in the generator's graph, plus two forwards and one backward per optimized step for the discriminator's own update. With R1 enabled, the real-side forward uses `create_graph=True` (second-order), adding one extra backward through the discriminator per optimized step.

- [x] Tests added

Design note for review: the discriminator LR intentionally has no schedule in v1 (TrainStepper never sees the epoch/valid-loss signals that drive schedulers); if this feature graduates, the plan is to move the base optimizer onto TrainStepper and plumb those signals through.

### DIFF
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -408,7 +408,9 @@ def run_evaluator_from_config(config: InferenceEvaluatorConfig):
 
             logging.info("Building validation stepper and aggregator")
             train_stepper_config = config.validation.stepper_training
-            train_stepper = TrainStepper(stepper=stepper, config=train_stepper_config)
+            train_stepper = TrainStepper(
+                stepper=stepper, config=train_stepper_config, training=False
+            )
 
             val_aggregator = config.validation.aggregator.build(
                 dataset_info=dataset_info,
--- a/fme/ace/registry/__init__.py
+++ b/fme/ace/registry/__init__.py
@@ -4,10 +4,21 @@ from fme.core.models import mlp as _mlp
 from . import land_net as _landnet
 from . import local_net as _localnet
 from . import m2lines as _m2lines
+from . import patch_discriminator as _patch_disc
 from . import prebuilt as _prebuilt
 from . import sfno as _sfno
 from . import stochastic_sfno as _sfno_crps
 from . import swin_transformer as _swin
-from .registry import ModuleSelector
+from .registry import ModuleSelector as ModuleSelector
 
-del _prebuilt, _sfno, _m2lines, _landnet, _localnet, _sfno_crps, _mlp, _swin
+del (
+    _prebuilt,
+    _sfno,
+    _m2lines,
+    _landnet,
+    _localnet,
+    _sfno_crps,
+    _mlp,
+    _swin,
+    _patch_disc,
+)
--- a/fme/ace/registry/patch_discriminator.py
+++ b/fme/ace/registry/patch_discriminator.py
@@ -1,0 +1,103 @@
+import dataclasses
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from fme.ace.registry.registry import ModuleConfig, ModuleSelector
+from fme.core.dataset_info import DatasetInfo
+
+
+def _circular_pad_lon(x: torch.Tensor, pad: int) -> torch.Tensor:
+    """Pad the longitude (last) dimension circularly; leave latitude alone."""
+    return F.pad(x, (pad, pad, 0, 0), mode="circular")
+
+
+@ModuleSelector.register("PatchDiscriminator")
+@dataclasses.dataclass
+class PatchDiscriminatorConfig(ModuleConfig):
+    """
+    Configuration for a small convolutional patch discriminator.
+
+    Applies two 3x3 convolutions with circular longitude padding and valid
+    latitude padding, followed by a 1x1 projection. All convolutions use
+    spectral normalization. Two positional channels (sin(lat), cos(lat)) are
+    concatenated to the input.
+
+    Parameters:
+        hidden_dim: First hidden layer width; the second layer doubles it.
+    """
+
+    hidden_dim: int = 64
+
+    def build(
+        self,
+        n_in_channels: int,
+        n_out_channels: int,
+        dataset_info: DatasetInfo,
+    ) -> nn.Module:
+        return PatchDiscriminator(
+            n_in_channels=n_in_channels,
+            n_out_channels=n_out_channels,
+            hidden_dim=self.hidden_dim,
+            lat=dataset_info.horizontal_coordinates.lat_1d,
+            n_lon=dataset_info.img_shape[1],
+        )
+
+
+class PatchDiscriminator(nn.Module):
+    """Two-layer 3x3 CNN with spectral normalization and positional encoding.
+
+    The forward pass prepends sin(lat) and cos(lat) as two extra input
+    channels, then runs::
+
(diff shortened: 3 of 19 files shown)


mcgibbon:
Claude: Pre-review process note (two independent review rounds ran against this PR; round 1 found three blocking defects — the DDP reducer being serviced by the generator's backward so discriminator gradients were never all-reduced, `multi_call` silently dropping a configured discriminator, and the inference evaluator's validation pass crashing on GAN checkpoints — all fixed in f884984f1 and independently verified in round 2, including a mutation test of the DDP ordering and an end-to-end evaluator run against a real GAN checkpoint).

Known limitations deliberately left for review rather than patched pre-merge; none affect the gating 1° daily ERA5 run (unlabeled, uncoupled, 1-step):

1. **Labeled batches + an unconditional discriminator crash at the first training batch** (`TypeError: Labels are not allowed for unconditional models` from `fme/core/registry/module.py`): the discriminator forwards labels unconditionally. A conditional generator with an unconditional discriminator is plausible (e.g. multi-source FM runs); options are a construction-time guard or dropping labels when the discriminator module is unconditional.
2. **The coupled path silently ignores a configured discriminator**: `CoupledComponentConfig.stepper` accepts a `discriminator:` block and builds the module, but `CoupledTrainStepper` never reads it — same silent-drop class the `multi_call` fix closed. A `NotImplementedError` in `CoupledTrainStepper.__init__` would close it.
3. **Multi-step asymmetry is documented but not guarded**: for steps beyond the first, the real pair is teacher-forced while the fake pair is free-running, so the discriminator can key on rollout drift instead of texture. If you'd rather have a warning or hard guard when a discriminator is combined with multi-step training, say the word.
4. **Validation-pass monitors aren't observable**: the GAN monitor metrics only surface as per-batch `batch_*` training keys; in validation passes the discriminator forwards run but the epoch aggregator drops the keys (small wasted compute, no `val/mean/` monitors). Plumbing them through the aggregator is a follow-up if validation-side GAN monitoring turns out to matter.

mcgibbon:
PR description updated to cover the R1 warmup commit (420dc664c). Description authored by Claude Opus 4.6 (1M context).

PR #1448, opened 2026-08-19
Time stride on coupled input datasets; 5-daily fractions for the CM4-like-AM4 random-CO2 coupled sea ice ensemble

jpdunc23:
The CM4-like-AM4 random-CO2 coupled sea ice ensemble is built from sea-surface snapshots, so it cannot use a window average; instead the sea ice and ocean fractions are subsampled to a 5-day cadence and forward-filled back onto the 6-hourly atmosphere index, staying coherent with the land fraction. This adds the one capability that was missing for that — a time stride on an input store — and rewrites the ensemble config to use it. The coupling arithmetic in `compute_coupled_sea_ice` is reused unchanged: the strided sea-surface store goes in the `ocean` input slot, where the existing ocean-sourced fraction branch already computes coherent fractions at the input's own cadence and forward-fills them.

`pr-plan.md` at the repo root is the API roadmap under review; it is removed from the branch before this PR leaves draft.

Changes:
- `scripts/data_process/create_coupled_datasets.py`: `CoupledInputDatasetConfig.time_stride` subsamples an input store in time, applied after the `first_timestamp`/`last_timestamp` slice
- `scripts/data_process/configs/CM4-like-AM4-random-CO2-ensemble-coupled.yaml`: rewritten at `version: "2026-08-18"` — sea-surface stores in the `ocean` slot at a 5-day stride, no `sea_ice` input, no `window_avg`, `include_ts: true`, GCS output directory, `2026-06-19` atmosphere inputs and stats
- `scripts/data_process/test_create_coupled_datasets.py`, `scripts/data_process/test_coupled_dataset_utils.py`: stride selection, composition with the timestamp range, invalid-stride errors, and end-to-end piecewise-constant strided fractions and `include_ts` blend

- [ ] Tests added

### DIFF
--- a/pr-plan.md
+++ b/pr-plan.md
@@ -1,0 +1,190 @@
+# Time stride on coupled input datasets; 5-daily fractions for the CM4-like-AM4 random-CO2 coupled sea ice ensemble
+
+`CoupledInputDatasetConfig` gains a `time_stride`, so an input store can be
+subsampled to a coarser cadence before any coupling arithmetic runs. Putting the
+1-degree sea-surface stores in the `ocean` slot at a 5-day stride makes
+`compute_coupled_sea_ice`'s existing ocean-sourced branch emit 6-hourly coupled
+sea ice whose sea ice and ocean fractions vary every 5 days and stay coherent
+with the land fraction. No coupling code changes.
+
+---
+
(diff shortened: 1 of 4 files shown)


pr-plan.md
jpdunc23, line 1: TODO: delete

PR #1449, opened 2026-08-19
Refactor corrector to accumulated CorrectorOutput fold with delta/diagnosis write kinds

mcgibbon:
Refactor the corrector system to use an accumulated `CorrectorOutput` fold with explicit delta/diagnosis write kinds, replacing the previous dict-update + post-hoc snapshot-delta model.

## Key changes

**Correction fold model.** Each `Correction.__call__` now receives and returns a `CorrectorOutput` instead of `(modified_dict, corrector_state)`. The `CorrectorOutput.apply_correction(diagnosed=..., deltas=...)` method distinguishes additive corrections (deltas: `corrected = original + delta`) from wholesale replacements (diagnoses: `corrected = new_value`). `CorrectorDiagnostics` gains a `pre_diagnosis_fields` mapping that records the raw network output for diagnosed fields; `modified_names` is the union of delta keys and pre-diagnosis keys.

**Delta vs. diagnosis classification.** A delta is a corrector effect that should vanish for a well-behaved network (clamps, conservation offsets, the moisture budget's multiplicative precipitation/evaporation rescaling); these are candidates for regularization toward zero. A diagnosis replaces a field wholesale with a value derived from other fields (the moisture budget's advection recompute, `zero_global_mean_moisture_advection`, the surface energy flux correction, ice reconstructed prognostics, prescribed fields).

**Prescription ordering.** Ocean SST prescription and prescribed prognostics now run *before* the corrector (previously after). This lets the corrector operate on the state with prescribed fields in place. The ocean/corrector overlap guard is removed since overlap is now handled by the fold (the corrector sees and can correct the prescribed value). `OceanConfig` gains a `prescription_as_delta` flag to control whether the SST prescription is recorded as a corrector delta or a diagnosis.

**Correction API.** `CorrectorABC.__call__` and subclasses gain an optional `seed: CorrectorOutput | None` kwarg for the step to pass a pre-seeded output (carrying prescription diagnostics) into the correction fold. The `Correction` protocol signature changes from `(input_data, gen_data, forcing_data, corrector_state) -> (TensorDict, CorrectorState | None)` to `(input_data, forcing_data, accumulated_output) -> CorrectorOutput`.

**AtmosphereData / OceanData correction verbs.** Both classes gain a `for_correction(accumulated_output)` constructor and typed `correct_*` / `diagnose_*` methods that record pending deltas/diagnoses, flushed by `result()`. Module-internal helper functions (`_force_conserve_moisture`, `_adjust_gen_dry_air_to_target`, etc.) now receive the data wrapper and mutate through these verbs instead of returning modified dicts.

**Collision check.** `AtmosphereCorrectorConfig._build` rejects `force_positive_names` entries that collide with fields diagnosed by later corrections (the moisture budget's advection recompute, `zero_global_mean_moisture_advection`) at build time, rather than silently dropping the delta at runtime. Fields the moisture budget only *rescales* (precipitation/evaporation) are deltas, not diagnoses, so they remain valid `force_positive_names` entries: the clamp and the rescaling compose as two deltas, with the rescaling operating on the clamped field as before.

**Redundancy skip.** `zero_global_mean_moisture_advection` is skipped when the `moisture_budget_correction` recomputes advection from scratch, since the recomputation supersedes the zero-mean offset.

## Behavioral notes

- **Existing configs keep building.** The ACE2-ERA5 baseline shape (`PRATEsfc` in `force_positive_names` together with `moisture_budget_correction: advection_and_precipitation`) is delta-then-delta and remains valid. The build-time `ValueError` fires only for `force_positive_names` entries on genuinely diagnosed fields (the advection tendency).
- **Correction metrics** for fields that moved from delta to diagnosis (the advection tendency) no longer appear in the `CorrectionDeltaAggregator`'s wandb metrics. Their pre-correction values are now in `pre_diagnosis_fields` and written to a separate `step_diagnostics/pre_diagnosis_fields.nc`. Precipitation/evaporation rescaling stays a delta and keeps its wandb correction metrics.
- **The corrector now sees prescribed values.** Conservation corrections (energy, moisture, dry air) account for the prescribed SST and prescribed prognostics in their budgets. Previously the prescriptions were applied after the corrector.

## Files changed (22)

Core: `corrector/output.py`, `corrector/registry.py`, `corrector/atmosphere.py`, `corrector/ocean.py`, `corrector/ice.py`, `corrector/utils.py`, `atmosphere_data.py`, `ocean_data.py`, `ocean.py`, `step/single_module.py`, `step/output.py`, `step/step_diagnostics.py`, `ace/aggregator/inference/step_diagnostics.py`

Tests: `corrector/test_output.py`, `corrector/test_atmosphere.py`, `corrector/test_ocean.py`, `corrector/test_ice.py`, `corrector/test_registry.py`, `step/test_output.py`, `step/test_step.py`, `distributed/parallel_tests/test_step.py` + checkpoint

- [x] Tests added


mcgibbon:
Claude: Pre-review applied two fixes and updated the PR title/description.

**Fix 1: Regenerate parallel test checkpoint** (commit 2f8b2e77)
The `sm_with_atmos_corr_output.pt` checkpoint was generated with the old `force_positive_names=["PRATEsfc"]` config and the old prescription-after-correction ordering. The test failed with 100% mismatched elements (max abs diff 7.94). Deleted the stale checkpoint and regenerated it against the new behavior.

**Fix 2: Include `zero_global_mean_moisture_advection` in collision check** (commit a52bc556)
The `diagnosed_prefixes` collision check in `AtmosphereCorrectorConfig._build` covered `moisture_budget_correction` and `clip_frozen_precipitation` but missed `zero_global_mean_moisture_advection`, which also diagnoses the advection tendency field. Without this, a `force_positive_names` entry for an advection tendency prefix would pass the build-time check but hit a runtime `ValueError` ("cannot diagnose fields that already carry a delta") from `CorrectorDiagnostics.apply_correction`. Hoisted the `advection_recomputed` variable so it is computed once and reused.

**Fix 3: Updated PR title and description** (via `gh api`)
The previous description listed changes as "Fix 7 spec deviations" without describing the behavioral significance. Updated to explain the prescription ordering change, backward compatibility implications (`force_positive_names` collision now raises at build time), correction metric impact (fields moving from delta to diagnosis), and the removed ocean/corrector overlap guard.

mcgibbon:
Claude: Pre-review of PR #1449

## Audit passes

1. **Silent-failure pass** -- clean. Defaults are backward-compatible (`prescription_as_delta=False`, `seed=None`). The `assert self._corrector_output is not None` pattern in correction verbs is consistent with the codebase.

2. **Behavior-change pass** -- two findings, both addressed:
   - Prescription ordering changed from after-corrector to before-corrector. Conservation corrections now budget against prescribed state. Updated PR description to document this.
   - `force_positive_names` collision with diagnosed fields now raises `ValueError` at build time (previously silent). Updated PR description to document the backward compatibility implication.

3. **Differential-test pass** -- clean. `test_correction_ordering_does_not_violate_diagnosis_after_delta` covers 4 representative config combos and would fail if correction ordering allowed diagnosis-after-delta. `test_stack_diagnostics_diagnosis_only` would fail if stacking ignored `pre_diagnosis_fields`. The `test_apply_correction_diagnosis_after_delta_raises` negative test confirms the guard.

4. **Test regression** -- one stale checkpoint found and fixed. The `sm_with_atmos_corr_output.pt` parallel test checkpoint was generated with the old config (`force_positive_names=["PRATEsfc"]`) and old ordering, causing 100% element mismatch. Regenerated.

5. **Collision check completeness** -- one gap found and fixed. `diagnosed_prefixes` in `_build` missed `zero_global_mean_moisture_advection` (which diagnoses the advection tendency field). The runtime guard in `CorrectorDiagnostics.apply_correction` would catch it, but the build-time check should be comprehensive for clear error messages. Fixed by adding the advection prefixes and hoisting `advection_recomputed`.

6. **AI-agent artifact pass** -- PR title and description were stale. Title said "Fix spec deviations" for what is a substantial refactor. Description listed changes without explaining behavioral significance (ordering change, backward compat, metric impact). Updated both.

## Open questions for reviewer

1. **Prescription-before-correction ordering**: The corrector now sees prescribed SST and prescribed prognostics. For conservation corrections this is more physically consistent (they budget against the actual output state). For `ForcePositive`, it means the corrector can clamp a prescribed value (e.g., if prescribed negative). Is this the intended contract, or should prescribed values be exempt from corrector modifications?

2. **Correction metric impact**: Fields that moved from delta to diagnosis (precipitation rate, advection tendency, frozen precipitation) no longer appear in `CorrectionDeltaAggregator` wandb metrics. Their pre-correction values are written to a separate `pre_diagnosis_fields.nc` file but are not logged as wandb scalars/maps. Is this acceptable, or should the aggregator be extended to report on diagnosed fields too?

## What I changed

- Regenerated `sm_with_atmos_corr_output.pt` (commit 2f8b2e77)
- Added `zero_global_mean_moisture_advection` to the `diagnosed_prefixes` collision check, hoisted `advection_recomputed` (commit a52bc556)
- Updated PR title and description to accurately reflect the refactor scope and behavioral changes

## Verdict

Ready for review after the fixes I applied. The refactor is well-structured: the `CorrectorOutput` fold model with explicit delta/diagnosis write kinds is cleaner than the previous dict-update + post-hoc snapshot approach, the test coverage is thorough (259 tests pass), and the behavioral changes are now documented in the PR description. The two open questions above are judgment calls for the reviewer.

fme/core/corrector/atmosphere.py
```diff
@@ -355,9 +339,49 @@ def _build(
         area_weighted_mean = gridded_operations.area_weighted_mean
         timestep_seconds = timestep.total_seconds()
         corrections: list[Correction] = []
+
+        # Fields that will be diagnosed (replaced wholesale) by later
+        # corrections.  A ForcePositive delta on such a field would be
+        # erased by the diagnosis, so a collision is rejected at build time.
+        from fme.core.atmosphere_data import ATMOSPHERE_FIELD_NAME_PREFIXES
```
mcgibbon: Issue: import belongs at top of file.

fme/core/corrector/atmosphere.py
```diff
@@ -562,80 +545,58 @@ def _force_conserve_moisture(
     evaporation_global_mean = area_weighted_mean(gen.evaporation_rate, keepdim=True)
     precipitation_global_mean = area_weighted_mean(gen.precipitation_rate, keepdim=True)
     if terms_to_modify.endswith("precipitation"):
-        # We want to achieve
```
mcgibbon, line 565: Why were these comments removed? Should they be restored? That goes for any removed comments similar to this.

mcgibbon:
OK, the PR is complete in a form that shows this can be done without large issues, I think. I would want to split into multiple PRs to be reviewed (even for me to self-review the AI code before marking it ready), but I want to get some input from @jpdunc23 and make sure there aren't any gaping holes before going through that.

The general philosophy is to propagate a CorrectorOutput object with an accumulation method through the course of all corrections. Diagnoses and deltas can be registered. A diagnosis cannot be registered after a delta, guaranteeing that if both are used, post-diagnosis-value + deltas = final. That means we only need to store the pre-diagnosis state to have all required snapshots:

- NN output for that channel before diagnosis, which could be used in diagnosis (it's a bit of a hack but one we're currently using, the NN output is not really considered to be the output it purports to be)
- The diagnosed value, if one exists
- Corrector deltas applied to the NN output or diagnosed value. The definition of a delta is a corrector effect which the user configuring the model believes should be forced to zero in a well-behaved model. In other words, when corrector regularization is enabled, this should be exactly the list of variables being regularized. We can add configurability as needed for the user to be able to specify their belief in whether an effect is a delta to be forced to zero - the ocean SST prescription in the atmosphere model has been given such an option in this PR.

No changes are made to the actual behavior, just how these changes are registered and presented. Currently in this PR the pre-corrector fields are added to data writing, but the PR strictly removes deltas fields from wandb metrics (the ones that aren't valid to think of as deltas/errors). I could add presentation of the pre-corrector values to diagnosed fields to wandb if you like (these should be presented full-field, not as deltas, in wandb, and there are some design choices to make there if we do, which is why I deferred).

This is how I broke down the deltas vs diagnostics:

Always deltas

- ForcePositive — additive clamp adjustment
- ConserveDryAir — globally-constant surface pressure offset
- TotalEnergyBudgetCorrection — uniform temperature offset
- ZeroGlobalMeanMoistureAdvection — constant offset to advection tendency
- MoistureBudgetCorrection precipitation/evaporation scaling — multiplicative adjustment (the ratio preserves the field's identity)
- Ice budget terms (source, sink, transport) — adjusted, not replaced
- OceanHeatContentCorrection thetao levels (and sst when present) — multiplicative scaling
- Frozen precipitation clip — clipped against total precip (treated as diagnosis)

Always diagnoses

- MoistureBudgetCorrection advection recompute (when advection_and_*) — the advection tendency is recomputed from scratch as the budget residual TWP_tendency - (evap - precip), completely replacing the network's value with one derived from other fields
- SurfaceEnergyFluxCorrection (both residual_prediction and prescribed methods) — hfds replaced with a value computed from atmospheric forcing fluxes and SST. I went through several rounds of discussion with Claude on this, trying to figure out if there was any sub-component of either form of the correction that could be forced towards zero without causing issues - they all bias the model in some way, particularly for partial-fraction cells.
- Ice reconstructed prognostics (siconc, simass, sisnmass) — reconstructed from budget terms, wholesale replacement
- New frozen precipitation diagnosis — multiply the frozen precipitation prediction by total precipitation, so that the NN output is treated as a fraction of total precipitation. Can be used in conjunction with the delta-treatment clipping, does not guarantee anything on its own.
- Prescribed prognostics — wholesale replacement from forcing data

User-configurable

- Ocean SST prescription — diagnosis by default; OceanConfig.prescription_as_delta makes it a delta instead (for the follow-on delta-magnitude regularizer). Made this configurable because you specifically requested being able to train the atmosphere to produce accurate open-ocean values even if they get discarded.

@jpdunc23 should I move forward with splitting this into reviewable PRs so we can review it in parts, or are there cross-cutting issues with this design you can already see from here?

fme/core/corrector/atmosphere.py
mcgibbon: Claude: Fixed in ad6cce483 — moved the import to the top-level `fme.core.atmosphere_data` import block (no circularity; the module already imports `AtmosphereData` from there).

fme/core/corrector/atmosphere.py
mcgibbon: Claude: No good reason — the refactor rewrote these functions and dropped the explanatory comments along the way, but the logic they describe is unchanged. Restored in ad6cce483, along with the other still-accurate removed comments I found sweeping the full PR diff: this derivation block and the advection-recompute rationale, the ForcePositive-before-conservation ordering rationale and the frozen-precip clip placement note in _build, the uniform-temperature-correction notes in _force_conserve_total_energy, the heat-flux priority-cascade and Kelvin-assumption comments in _force_conserve_ocean_heat_content, and the StepperState field-preservation note in step_with_adjustments. I left out only comments describing the old return-only-modified-fields Correction contract (e.g. "'changed' holds only the fields this correction modified", "only the modified field is returned"), which no longer applies under the CorrectorOutput fold.

jpdunc23:
I took a quick look. What I like: SST prescription is folded into the same accounting as corrections, which I think is a step in the right direction toward making `SingleModuleStep` truly component agnostic instead of its current atmosphere-specific bent with `OceanConfig`.

My main question: Why funnel corrections into two buckets? The two kinds aren't well-defined even on this PR's own examples:

- `force_positive` as "delta" is a spatially-conditional replacement: identity where the field is positive, overwrite where it isn't.
- SST prescription as "diagnosis" is a spatially-masked replacement: overwrite over ocean, identity over land.

Many ops here are "replace on some support, identity elsewhere," so neither "delta" nor "diagnosis" particularly fits. 

This PR encodes a judgment call ("this delta should shrink as the network improves, that one never will") as a type constraint. IMO, post-hoc modifications to network outputs are one of the key levers we can own and optimize with domain knowledge. We should not make research in this area more difficult by imposing a type system on these operations.

However, every "diagnosis" is a legitimate optimization target for some user. I believe we should respect the scientific maturity of our users by forcing them to understand the corrections and allowing them to make their own mistakes. For example, regularizing the advection-recompute delta trains network self-consistency with the recomputation. The status-quo is a network output channel that costs real compute but has zero leverage in the learning. This PR would cement that waste.

My suggestion is simpler than this PR: Rename "Corrector" to reflect its potential as a set of arbitrary user-configured post-step ops on the network outputs, prescription included, and move ACE's SST prescriptions in. What we have now with one flat per-field `delta = corrected - raw` is exact. Extending the existing snapshot delta to cover prescription (snapshot before prescription runs) keeps it exact with no overlap guard needed. Here's a draft plan for that PR: https://github.com/ai2cm/ace/pull/1452

mcgibbon:
Yes, I agree it will be nice to be able to unify post-network modifications. Keep in mind SST prescription should not be regularized as a delta in our critical path models unless and until we have positive experimental evidence supporting that model change.

> My main question: Why funnel corrections into two buckets?

Because we need a mechanism to select which operations are valid targets for pre-corrector optimization and corrector regularization. This PR puts that mechanism in the corrector itself, and constrains the choices to make it more easily user-configurable (because they don't need to take the weeks of time we've spent to understand the internal details of the corrector).

> Many ops here are "replace on some support, identity elsewhere," so neither "delta" nor "diagnosis" particularly fits.

The diagnosis/delta distinction isn't about the mechanism the correction uses to compute the change (that's an internal code implementation detail we don't expose to the user), it's explicitly about whether the effect of the corrector is considered conceptually by the configuring user to be a correction to the network's output, or a computation of a value that is conceptually distinct from the original network's output. In the first case, the user is saying they wan't this value regularized when corrector regularization is on (broadly stated - pre-corrector optimization is regularization). In the latter case, they're saying the network output is not the same conceptual value as the channel target, and shouldn't be regularized to be closer to that target.

> However, every "diagnosis" is a legitimate optimization target for some user.

Not every diagnosis is legitimate for regularization or pre-corrector optimization. There's one class where it's never valid - any diagnosis that uses the network output in the course of computing the diagnosed value. For those fields, the network output is by definition targeting a different value than the final output, and the regularization loss is in opposition to the target loss. For example, in one hfds configuration the network output's target is the hfds under sea ice, but the optimization target is the combined hfds under sea ice and open ocean - using pre-corrector optimization on this makes the network hfds channel target the wrong value. We shouldn't allow a user to configure it in ace main - a user who really wants to can modify the code to allow it.

For diagnosed values that get replaced without reading the original value to compute the diagnosed output, it's ambiguous. I would expect we should not be optimizing these (including the advection delta), you expect they should be optimized. We currently have no experimental evidence to support moving to optimizing them, and no specific issue we expect to be improved by the change. However, the PR doesn't make it more difficult to get that evidence - it defines a framework to allow the user to decide which of these paths is being used, and adds the option for SST prescription. It would be straightforward to add the option for other corrections you want to experiment with.

I am opposed to providing full configurability to users for all of these options. Full configurability of all of these variables is very confusing. It's taken me weeks to wrap my head around these! I would really like to avoid exposing a user to that full complexity, by guiding them with more carefully selected configuration options.

mcgibbon:
Closing for now. I think the pattern would be better fit by something like https://github.com/ai2cm/ace/pull/1465, which we can discuss later. The version in this 1449 was targeted at minimizing mechanical differences, but doesn't address the more significant issue of not giving an explicit label to pre-derived fields, and bites off a bit more than we should chew in one PR by updating _all_ cases at once.

PR #1450, opened 2026-08-19
Fix input dropout masking non-optimized rollout steps

yyexela:
Multi-step fine-tuning with `optimize_last_step_only` runs non-optimized rollout steps under `torch.no_grad()`. Those steps exist only to produce the trajectory fed to the optimized step, but input dropout was masking them too — perturbing that trajectory in a way inference never sees. This is a bug: the non-optimized steps should behave like inference, not like trained steps.

This gates `_draw_input_dropout_mask` on `torch.is_grad_enabled()` so non-optimized steps run unmasked unconditionally. The gate follows existing precedent: the input-NaN check in the same function is already gated on `torch.is_grad_enabled()` (cf. #1297). Both multi-step training loops run non-optimized steps under `torch.no_grad()` (`fme/ace/stepper/single_module.py` and `fme/coupled/stepper.py`), so the coupled stepper inherits the behavior without further changes.

**Training-time behavior change**: users combining `input_dropout` with `optimize_last_step_only` or `evaluate_all_steps` will see non-optimized rollout steps run unmasked where they were previously masked. No config in the repo uses `input_dropout`, and inference is unaffected.

Changes:
- `fme.core.step.SingleModuleStep._draw_input_dropout_mask`: returns `None` when gradients are disabled, unconditionally. Docstrings on the method and `SingleModuleStepConfig.input_dropout` updated to describe the behavior
- `fme/core/step/test_step.py`: step-level test that input dropout applies under `enable_grad` and is skipped under `no_grad`, using a rate-1.0 Bernoulli to make indicators deterministic
- `fme/ace/stepper/test_single_module.py`: rollout tests that a three-step `train_on_batch` with `optimize_last_step_only` masks only the final step, and that all steps are masked when every step is optimized
- `fme/coupled/test_stepper_integrations.py`: coupled rollout test that a four-step atmosphere rollout with `optimize_last_step_only` masks only the final step, pinning the claim that the coupled trainer inherits the gate from its grad contexts

- [x] Tests added


yyexela:
Claude:

Pushed `fcf228b5c` addressing a local review pass.

- **Config validation**: `SingleModuleStepConfig.__post_init__` now raises when `input_dropout_optimized_steps_only` is set without `input_dropout`, instead of silently doing nothing.
- **`Trainer._log_first_batch_metrics` now runs in eval mode.** It was the only place in the repo where a train-mode stepper ran `train_on_batch` under `no_grad`, so the grad-state gate disagreed there with "is this step optimized". The other two eval passes (`train_one_epoch`'s post-epoch evaluation, `run_validation_loop`) already call `set_eval`.
- **Docstring correction**: the flag is a no-op unless `optimize_last_step_only` is set — a training rollout is otherwise as long as the sampled loss length, so every step is optimized. Also notes that the gate reads grad state rather than an explicit per-step flag.
- **Tests**: the step-level test no longer calls the private `_draw_input_dropout_mask`; it asserts on the packed network input instead. Added coverage for the no-op case, the config validation, and the trainer eval-mode behavior.

Considered threading an explicit `optimize: bool` through `StepArgs` instead of reading grad state, but that means plumbing a per-step predicate through both `predict_generator` and the coupled stepper's per-realm generators. With `_log_first_batch_metrics` fixed, `torch.is_grad_enabled()` is an exact proxy on every path that exists today, so it did not seem worth the churn here.

yyexela:
Claude:

Pushed 14ac6c32b:

- Reverted the `_log_first_batch_metrics` eval-mode wrapper and its test. `NullOptimization.set_mode` already puts every module in `eval()` before the first forward step, so the wrapper could not affect whether input dropout applied. Its only real effect was flipping `LossSchedule` to the eval sampler, drawing the pre-training baseline's `n_loss_steps` from a different RNG stream than the training batches it is plotted against.
- Trimmed the `input_dropout_optimized_steps_only` docstring. The per-loop enumeration was inaccurate — `evaluate_all_steps` also leaves trailing steps under `no_grad` — and duplicated what the loops already express.

yyexela:
Claude:

Pushed 4e565d50a, following a self-review pass. Three changes, all on top of the flag as originally proposed:

- **Warning for the inert combination.** Setting `input_dropout_optimized_steps_only` without `optimize_last_step_only` did exactly nothing in the single-module loop, silently — and the earlier `..._noop_without_last_step_only` test locked that silence in. `TrainStepperConfig.get_train_stepper` now warns. That is the earliest point where both configs are known: `TrainConfig.__post_init__` cannot reach the step config through `CheckpointStepperConfig`. It warns rather than raises so existing training configs stay runnable.
- **`StepConfigABC.input_dropout_optimized_steps_only`.** Needed so the warning can read the flag off an arbitrary step config without an `isinstance` check. I first made it abstract; that broke 8 tests (`Can't instantiate abstract class MockStepConfig`), since every test-defined and out-of-tree step config would then have to implement it. Concrete with a `False` default instead, delegated by `StepSelector` and `MultiCallStepConfig`.
- **Coupled rollout test.** The claim that the coupled stepper inherits the behavior for free had no test. It holds only because `next(output_iterator)` is lexically inside `grad_context` in both realm arms, so a later refactor that hoists the pull or prefetches would silently re-mask non-optimized steps with nothing failing. Mutation-checked: deleting the grad gate fails the `optimized_steps_only=True` parametrization.

Also extended the config docstring to say that a coupled realm with `loss_weight: 0.0` has no optimized steps at all and so receives no input dropout for the entire run — defensible, since such a realm gets no gradients either, but the docstring renders into the public docs and was the only place a user would learn it.

Verification: `fme/core/step fme/ace/stepper fme/ace/step fme/coupled` gives 612 passed, 2 skipped. Full `fme` suite gives 3476 passed, with 8 failures that reproduce identically on a clean tree (1 graphcast regression baseline, 7 zarr tests in `test_xarray.py`). `pre-commit run --all-files` clean.

yyexela:
Ready for review!

Arcomano1234 (approved):
Thanks for getting this into a PR, LGTM

yyexela:
Claude Fable 5.1:

Pushed 06e3017d7, taking the review comment. The flag is gone. Masking non-optimized rollout steps was a bug, not a mode worth configuring: those steps only exist to produce the trajectory fed to the optimized step, and inference never masks them. Input dropout now applies only when gradients are enabled, which is how both training loops mark an optimized step.

With no config option to validate, the `StepConfigABC` property, the `StepSelector`/`MultiCallStepConfig` delegations, the `__post_init__` check, and the inert-schedule warning are all removed. The warning was also inaccurate under `evaluate_all_steps`, where trailing steps run under `no_grad` without `optimize_last_step_only`. Tests are kept but unparametrized: step-level grad gate, single-module rollout, and coupled rollout. Diff is now ~40 lines of source plus tests.

This is a training-time behavior change for anyone using `input_dropout` with `optimize_last_step_only` or `evaluate_all_steps`. No config in the repo uses `input_dropout`, and inference is unaffected.

yyexela:
Claude Fable 5.1:

PR title and description rewritten to reflect the flag removal in favor of an unconditional grad-state gate on input dropout.

yyexela:
Updated PR description: changed the precedent citation from #1262 to #1297 (the NaN check was introduced by #1297, not #1262). No other edits.

Model: Claude Opus 4.6 (1M context). No warnings from the context-gathering script.

PR #1451, opened 2026-08-19
Add cross-level attention and time-of-year conditioning to the Swin backbone

yyexela:
Adopts two design choices from ArchesWeather that the Swin backbone was missing: cross-level attention, and conditioning on the time of year.

The backbone previously stacked all vertical levels into the channel dimension, so levels could only interact through pointwise linear layers. This is the "column concatenation" scheme ArchesWeather argues against; its own design keeps levels as a tensor axis and attends along it. Separately, the data loader's times were discarded before the step was called, so a module had no way to know where in the year it was.

Both are opt-in and off by default, so existing configs are unaffected.

**Enabling `cross_level_attention` multiplies the number of latent tokens by `n_levels + 1`.** `embed_dim` should be reduced accordingly to keep compute comparable — at 8 levels, the token count grows 9x, which roughly matches halving `embed_dim`.

**The day-of-year embedding is discontinuous across the year boundary.** It reuses the sinusoidal timestep embedder ArchesWeather uses, whose geometric frequency ladder is not periodic in its input, so 31 December and 1 January are far apart in embedding space. This is deliberate — it is what ArchesWeather does, and there it outperformed coarser month-based conditioning — but a periodic encoding would be the obvious alternative if this proves to matter.

Changes:
- `fme.core.registry.module.ModuleConfig.build` and `ModuleSelector.build`: accept keyword-only `in_names` / `out_names` so a builder can tell which channels form a vertical column; passed by every step type that constructs a module
- `fme.core.stacker.ColumnLayout` and `infer_column_layout`: derive the (variable, level) grid from ordered channel names, raising when the level-resolved names do not form a rectangular, contiguous block
- `fme.core.models.swin_transformer.swin_layers.CrossLevelAttention`: multi-head attention along the vertical axis of each column, with a learned per-level positional embedding; replaces `ColumnMixer` when a layout is given
- `fme.core.models.swin_transformer.swin_transformer.SwinTransformerNet`: embeds levels per level and surface fields as one additional token in the same axis, folded into the batch dimension so window attention, patch merging and `ConditionalLayerNorm` are untouched
- `fme.ace.stepper.time_fraction.compute_time_fraction`: calendar position as a fraction in [0, 1), using the year length of the data's own calendar
- `fme.core.step.args.StepArgs`: new optional `time_fraction` field, propagated through `apply_input_process_func`; `StepABC.requires_time_fraction` lets a step declare that it needs it, so it is only derived when used
- `fme.ace.registry.stochastic_sfno.TimestepEmbedder` and `NoiseConditionedModel`: embed the calendar position into `Context.embedding_scalar`, raising if the value is missing rather than leaving the conditioning silently dead
- `fme.ace.registry.swin_transformer`: `cross_level_attention`, `column_num_heads` and `time_embed_dim` config options
- `fme.core.step.secondary_module`, `fme.core.step.radiation`: raise `NotImplementedError` when given a time-conditioned module, since they do not pass the value to their modules

- [x] Tests added

### DIFF
--- a/fme/ace/registry/hpx.py
+++ b/fme/ace/registry/hpx.py
@@ -41,6 +41,9 @@ class HEALPixUNetBuilder(ModuleConfig):
         n_in_channels: int,
         n_out_channels: int,
         dataset_info: DatasetInfo,
+        *,
+        in_names: list[str] | None = None,
+        out_names: list[str] | None = None,
     ) -> nn.Module:
         """
         Build a HEALPixUNet model.
@@ -49,6 +52,8 @@ class HEALPixUNetBuilder(ModuleConfig):
             n_in_channels: Number of input channels.
             n_out_channels: Number of output channels.
             dataset_info: Information about the dataset.
+            in_names: Ordered names of the input channels. Unused.
+            out_names: Ordered names of the output channels. Unused.
 
         Returns:
             HEALPixUNet model.
--- a/fme/ace/registry/land_net.py
+++ b/fme/ace/registry/land_net.py
@@ -22,6 +22,9 @@ class LandNetBuilder(ModuleConfig):
         n_in_channels: int,
         n_out_channels: int,
         dataset_info: DatasetInfo,
+        *,
+        in_names: list[str] | None = None,
+        out_names: list[str] | None = None,
     ):
         if len(dataset_info.all_labels) > 0:
             raise ValueError("LandNet does not support labels")
--- a/fme/ace/registry/local_net.py
+++ b/fme/ace/registry/local_net.py
@@ -75,6 +75,9 @@ class AnkurLocalNetBuilder(ModuleConfig):
         n_in_channels: int,
         n_out_channels: int,
         dataset_info: DatasetInfo,
+        *,
+        in_names: list[str] | None = None,
+        out_names: list[str] | None = None,
     ) -> nn.Module:
         params = AnkurLocalNetConfig(
             embed_dim=self.embed_dim,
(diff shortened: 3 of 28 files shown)


yyexela:
Status notes, kept out of the description since it becomes the squash-merge message.

**Throughput is estimated, not benchmarked.** Enabling `cross_level_attention` at `embed_dim: 128` (down from 256) works out to roughly 3x the per-step cost of the flat 2D backbone, from token count and `embed_dim` arithmetic alone. Nothing has been timed on GPU. Anyone adopting this for a long run should measure a few steps first.

**Verification so far** is correctness, not performance:
- unit tests for the layout inference, the attention itself (spatially pointwise, mixes levels), the calendar fraction across calendars, and `StepArgs` propagation
- builder-level tests that `cross_level_attention: true` actually produces a vertical axis and that `time_embed_dim` conditioning is live and raises when unfed
- a full stepper built from a real 4deg AIMIP config, forward pass checked, and predictions confirmed to differ between January and July

**Note on verifying the calendar conditioning by inspection:** `ConditionalLayerNorm.reset_parameters` zero-initializes the scalar `W_scale` / `W_bias`, DiT-style, so at initialization the calendar provably cannot change the output. Confirming it works requires perturbing those weights first — an untrained model will always show zero sensitivity.

**Pre-existing test failures on `main`, unrelated to this branch:** `fme/core/dataset/test_xarray.py` (7) and `fme/ace/models/graphcast/test_graphcast.py::test_graphcast_output_is_unchanged`. I confirmed these fail identically on a clean `main` checkout. `fme/ace/test_train.py::test_resume_two_workers[SphericalFourierNeuralOperatorNet]` also fails, on both this branch and another base.

PR #1426, opened 2026-08-10
Config-selected pre-corrector optimization and corrector regularization

jpdunc23:
Correctors change the network output before the loss sees it, so the loss cannot optimize the pre-corrector prediction and the correction magnitude carries no training signal. This adds a training-only `corrector_loss` on `TrainStepperConfig` with two opt-in features that consume the correction deltas already on `StepOutput`:

```yaml
stepper_training:
  corrector_loss:
    # main loss sees output − delta for these variables;
    # rollout state and returned predictions stay fully corrected
    precorrector_optimization:
      names_and_prefixes: [PRESsfc]
    # penalize delta toward zero in loss-normalized space
    regularization:
      names_and_prefixes: [specific_total_water_]
      norm: L2
      weight: 0.1
```

Selection is `NameAndPrefixMatcher` semantics and is required for each feature. Entries are validated twice: at build against the names the step loss covers, and against the corrector's actual correction-delta keys the first time the corrector loss runs on a non-empty delta. Both features may be enabled together. Defaults to `None`, so existing configs are unaffected.

Changes:
- `fme.core.loss`: `CorrectorLoss` owns the pre-corrector outputs and `CorrectorRegularizer`, the penalty; `StepOutputLoss` wraps an unchanged `StepLoss` alongside it and returns a `StepOutputLossOutput` keeping the penalty's own `LossOutput`
- `fme.core.corrector.loss_config` (new): `CorrectorLossConfig` and the two feature configs. Each feature config rejects an empty selection of its own; `CorrectorLossConfig.build` runs the build-time name check for both features and builds the pieces
- Name validation needs the corrector's delta keys, which are only observable once it has run. `require_matched_entries` is the one primitive both checks use; `CorrectorLoss.resolve_names` runs the runtime check from `StepOutputLoss.forward` on the first step whose deltas are non-empty, and fixes the channels each feature acts on. **An entry that matches a loss name but no correction delta therefore fails during training, not at build.** It fails early: the corrector is always applied in eval mode, so the end-of-epoch validation pass resolves even while `corrector_disabled_epochs` holds the corrector off the training steps. Correctors gain no new build-time requirement
- Correction deltas are no longer detached at the step seam: the corrected values they derive from are prognostic, so the subgraph is retained until backward regardless. `step_with_adjustments` is shared with validation and inference, but every non-training caller runs under `torch.no_grad()`
[...]

PR #1453, opened 2026-08-20
Apply corrector-loss features to every delta, dropping name selection

mcgibbon:
The per-variable name selection in #1426's corrector-loss features needs review iteration of its own; removing it unblocks the rest of #1426. #1454 reverts this PR to bring the selection back as a standalone follow-on. After this PR, both features act on every correction delta the corrector produces at each step, with no name accounting or storage anywhere.

Changes:
- `fme.core.corrector.loss_config`: `PreCorrectorOptimizationConfig` removed; `CorrectorLossConfig.precorrector_optimization` is now a boolean. `CorrectorRegularizationConfig` keeps only `norm` and `weight`. `CorrectorLossConfig.build` no longer takes `loss_names`, runs no name checks, and logs each enabled feature once at build, replacing the first-resolution log line
- `fme.core.loss`: `CorrectorLoss` and `CorrectorRegularizer` hold no names — `pre_corrector_outputs` swaps `prediction − delta` for every delta key, and the penalty covers each call's delta keys, rebuilding its `WeightedMappingLoss` per call (~80 µs, negligible against a training step). `resolve_names`, `require_matched_entries`, `_require_delta`, and the shrinking-delta-keys errors are removed
- `fme.core.name_and_prefix_matcher`: `NameAndPrefixSelection` removed — it was introduced by this stack solely for corrector-loss selection and is unused and untested without it
- `fme.ace.stepper.single_module`: `build_corrector_loss` no longer passes the loss names
- Tests of the deleted machinery (build-time/runtime name checks, deferred resolution, shrinking-key errors) are deleted; the remaining corrector-loss tests are updated to the selection-free config, and a test pins that the penalty's channels follow each call's delta keys

- [x] Tests added

### DIFF
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -107,7 +107,6 @@ from fme.core.corrector.ice import IceCorrectorConfig
 from fme.core.corrector.loss_config import (
     CorrectorLossConfig,
     CorrectorRegularizationConfig,
-    PreCorrectorOptimizationConfig,
 )
 from fme.core.corrector.ocean import OceanCorrectorConfig
 from fme.core.dataset.concat import ConcatDatasetConfig
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -898,8 +898,7 @@ class Stepper:
     def build_corrector_loss(
         self, corrector_loss: CorrectorLossConfig | None
     ) -> CorrectorLoss | None:
-        """Validate and build the corrector-delta half of the training loss,
-        if configured.
+        """Build the corrector-delta half of the training loss, if configured.
 
         Args:
             corrector_loss: Optional. With the null default, this method builds
@@ -911,7 +910,6 @@ class Stepper:
         if corrector_loss is None:
             return None
         return corrector_loss.build(
-            self.loss_names,
             normalizer=self._step_obj.get_loss_normalizer(),
             channel_dim=self.CHANNEL_DIM,
         )
--- a/fme/ace/stepper/test_single_module.py
+++ b/fme/ace/stepper/test_single_module.py
@@ -65,7 +65,6 @@ from fme.core.coordinates import (
 from fme.core.corrector.loss_config import (
     CorrectorLossConfig,
     CorrectorRegularizationConfig,
-    PreCorrectorOptimizationConfig,
 )
 from fme.core.corrector.output import CorrectorOutput
 from fme.core.corrector.registry import (
@@ -3254,20 +3253,17 @@ def _corrector_loss_stepper(
     disabled_epochs: int = 0,
     dataset_info: DatasetInfo | None = None,
     input_masking: StaticSpatialMaskingConfig | None = None,
-    names: list[str] | None = None,
 ) -> Stepper:
-    """Build a ``names`` -> ``names`` stepper, optionally installing a
-    correction. ``names`` defaults to ``["a"]``."""
-    names = ["a"] if names is None else names
+    """Build an ["a"] -> ["a"] stepper, optionally installing a correction."""
     config = StepperConfig(
         step=StepSelector(
             type="single_module",
             config=dataclasses.asdict(
                 SingleModuleStepConfig(
                     builder=ModuleSelector(type="prebuilt", config={"module": module}),
-                    in_names=list(names),
-                    out_names=list(names),
-                    normalization=trivial_network_and_loss_normalization(names),
+                    in_names=["a"],
+                    out_names=["a"],
+                    normalization=trivial_network_and_loss_normalization(["a"]),
                 )
             ),
         ),
@@ -3303,11 +3299,7 @@ def test_train_on_batch_pre_corrector_equivalence():
             _AddOne(), ConstantOffsetCorrection("a", offset)
         ),
         loss=StepLossConfig(type="MSE"),
-        corrector_loss=CorrectorLossConfig(
-            precorrector_optimization=PreCorrectorOptimizationConfig(
-                names_and_prefixes=["a"]
-            )
-        ),
+        corrector_loss=CorrectorLossConfig(precorrector_optimization=True),
     )
     baseline_out = baseline.train_on_batch(data, optimization=NullOptimization())
     corrected_out = corrected.train_on_batch(data, optimization=NullOptimization())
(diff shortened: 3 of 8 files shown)


mcgibbon:
Pre-review agent: independent pre-review of this PR against its base `feature/corrector-loss-config` (`9ecb124d7..4705297ea`). The removal itself is correct and complete on the symbols it names — a repo-wide grep for `PreCorrectorOptimizationConfig`, `require_matched_entries`, `resolve_names`, `_require_delta` and `names_and_prefixes`-on-corrector-configs finds no leftovers; nothing in `fme/coupled`, `fme/downscaling`, `docs/`, or any YAML config referenced the removed machinery. One thing should be fixed before review, plus some smaller cleanups.

### Should be fixed: `NameAndPrefixSelection` is left behind as dead code

`fme/core/name_and_prefix_matcher.py:42` still defines `NameAndPrefixSelection`. It was added by this stack's first commit (5fe9067b — it does not exist on `main`) purely to serve corrector-loss name selection, and after this commit **nothing in the repo references it**: no source, and no test either (`fme/core/test_name_and_prefix_matcher.py` covers only `NameAndPrefixMatcher`). Merging #1426 with this commit on top therefore lands an unused, untested public class in `fme/core`, which AGENTS.md rules out ("Commits ... should not add unused features that can't be reviewed in isolation"), and it makes the PR's "no name accounting or storage anywhere" not quite true.

Deleting the class here (along with the then-unused `dataclasses` and `Iterable` imports in that module) is consistent with the plan — #1454's revert restores it with everything else.

### Smaller cleanups

1. **`CorrectorRegularizer.forward` — `selected` is now just `dict(deltas)`** (`fme/core/loss.py:1026-1039`). With the selection gone the loop copies every key through unchanged, so only `targets` still needs building, and the name `selected` is misleading in a PR whose subject is removing selection:

   ```python
   targets = {
       name: torch.where(torch.isnan(delta), torch.full_like(delta, torch.nan), torch.zeros_like(delta))
       for name, delta in deltas.items()
   }
   return self._build_loss(sorted(deltas))(deltas, targets, data_mask)
   ```

2. **The loss object is now rebuilt on every `forward`** (`self._build_loss(names)(...)`), where before it was built once in `resolve`. `WeightedMappingLoss.__init__` builds a `Packer` and a weight tensor it moves to the device, measured at ~80 µs per build for 40 names — negligible next to a model forward (~0.1% of a step), so this is not a blocker. But nothing forces it: with `weights={}` the loss depends only on the name tuple, so memoizing on `tuple(names)` would restore the once-only build. Either way the PR description reads as a pure name-removal and doesn't mention that the per-call rebuild is new.

3. **`CorrectorLossConfig.build` lost its docstring summary line** (`fme/core/corrector/loss_config.py:99`) — it is now a bare `Args:` block. Worth a one-liner.

4. **Stale wording**: `fme/core/test_loss.py:1243`, `test_penalty_uses_the_data_mask`, still says "a data_mask hiding one sample of a selected variable".

5. **PR description**: the template asks that stacking narrative go in a comment rather than the description, since the description becomes the squash-merge commit message. The opening ("Splits the per-variable name selection out of #1426 so the rest of that PR can merge ... a follow-on PR that reverts this commit") is exactly that. The commit message on 4705297e already states the change without it and could be lifted as the opening. The template's second checkbox (deps/image) was also dropped rather than left unchecked.

### Notes for the reviewer, no action needed

- **Scope widening not spelled out in the description**: the build-time check used to require every regularization entry to match a *loss* name, so the penalty was necessarily a subset of the loss names. It is now every delta key. For `SingleModuleStepConfig` `loss_names == output_names` and the loss normalizer covers `in_names ∪ output_names`, so every delta key is normalizable and this is a distinction without a difference today — but it is a widening worth confirming as intended.
- The `len(deltas) == 0` early return dropped from `pre_corrector_outputs` is unreachable-equivalent: `StepOutputLoss.forward` short-circuits on empty deltas before calling it.
- `TrainStepperConfig.corrector_loss` is training-only config, so the config-shape break is sanctioned by AGENTS.md ("can be broken for training") — no checkpoint-inference concern.
- `fme/ace/__init__.py` computes `__all__` dynamically, so dropping the export needs no follow-up.

### Verification run

- pre-commit (ruff, ruff-format, mypy) on all 7 changed files: pass.
- `fme/core/test_loss.py` + `fme/core/corrector/test_loss_config.py`: 87 passed.
- `fme/ace/stepper/test_single_module.py`: 150 passed, 1 skipped.
- `fme/core` (`--fast`): 1476 passed, 20 skipped, 7 failed — all 7 in `fme/core/dataset/test_xarray.py` with `zarr.core.config.BadConfigError: Pipeline class 'zarrs.ZarrsCodecPipeline'`, an environment gap unrelated to this change.

mcgibbon:
Claude: process notes, kept out of the description since it feeds the squash-merge message:

- This PR splits the selection design out of #1426 so the rest of that PR can merge; #1454 — a revert of this commit, recorded with @jpdunc23's authorship — carries the selection feature as its own follow-on for review iteration.
- The head was force-pushed once (4705297 → f7385ea) to address pre-review findings: `NameAndPrefixSelection` was left behind unused, a leftover `selected` dict, a lost docstring summary, stale test wording.
- Behavior note from pre-review, worth confirming as intended: the penalty's channels are now every delta key rather than the subset also matching the loss names. For `SingleModuleStepConfig`, `loss_names == output_names` and the loss normalizer covers `in_names ∪ output_names`, so every delta key is normalizable and the widening has no effect today.

mcgibbon:
Pre-review agent: re-reviewed at f7385ea4a. Every item from my comment above is addressed — `NameAndPrefixSelection` and its now-unused imports are gone (the module keeps only `NameAndPrefixMatcher`, 36 lines), the `selected` copy in `CorrectorRegularizer.forward` is dropped, `CorrectorLossConfig.build` has a summary line, the stale test comment is fixed, and the description now discloses both the `NameAndPrefixSelection` deletion and the per-call `WeightedMappingLoss` rebuild. Keeping the rebuild unmemoized is a reasonable call given the "no accounting or storage" intent, and it is now stated rather than implicit.

Re-verified on the amended head: repo-wide grep for every removed symbol is clean; `test_loss_config.py`, `test_loss.py`, `test_name_and_prefix_matcher.py` and `test_single_module.py` give 242 passed, 1 skipped; pre-commit (ruff, ruff-format, mypy) passes on all changed files. `git diff feature/corrector-loss-config 52af0ae8b` is empty, so #1454 restores this stack's prior state exactly — the removal is fully reversible.

Ready for review.

mcgibbon:
> Behavior note from pre-review, worth confirming as intended: the penalty's channels are now every delta key rather than the subset also matching the loss names. For SingleModuleStepConfig, loss_names == output_names and the loss normalizer covers in_names ∪ output_names, so every delta key is normalizable and the widening has no effect today.

This should cause no difference in behavior in practice for any of our configurable runs, and I would argue is more correct (a corrector delta is something we could regularize even if for whatever reason the thing it's applying a delta to is not directly optimized in the loss).

PR #1455, opened 2026-08-20
Pin PR title and description authoring to claude-opus-4-6[1m]

mcgibbon:
Claude Code sessions in this repo run on whichever model the user picked, so the authoring of PR prose (titles and descriptions) varies by session. This pins PR-prose authoring to `claude-opus-4-6[1m]` via a skill whose `model:` frontmatter sets the model and whose `context: fork` scopes the pin to exactly that work (a bare `model:` pin is turn-scoped: it would leak into the rest of the turn and lapse on the next prompt). A `.claude/agents/` subagent definition could also bind the model, but a skill is directly invocable in the flow where PR prose gets written and needs no separate agent file. AGENTS.md routes both paths that produce it — PR creation and post-commit amendment — through the skill, and the pre-review skill's title/description-improvement step uses it too.

Changes:
- `.claude/skills/write-pr-prose/SKILL.md`: new skill pinned to `claude-opus-4-6[1m]` with `context: fork`; gathers the PR's context (existing prose, discussion, full diff from its base), writes or amends the title and description per the repo template, and reports what it did, which model it ran, and any context the helper script failed to retrieve as a PR comment so both silent fallbacks and retrieval gaps are visible.
- `.claude/skills/write-pr-prose/gather_pr_context.sh`: helper the skill's step 1 runs — retrieves the existing title/description, issue comments, reviews, and inline review comments via paginated GitHub API calls, and the full diff from the PR's base branch so stacked PRs are described correctly; warns when the PR base branch cannot be determined; discussion-section failures print `(warning: ...)` and continue, but a failed diff retrieval is fatal (`ERROR` + exit 1) since prose must not be written without the diff; no local fetch needed. The script requires an existing PR; with no PR it exits with an error (same fatal handling as a failed diff).
- `AGENTS.md`: the committing path (amendments) routes through the skill and references the PR's base branch rather than main; the "PR description template" section routes title/description writing through the skill in Claude Code sessions.
- `.claude/skills/pre-review/SKILL.md`: pre-review's title and description improvements go through the same skill.

- [ ] Tests added (agent-configuration markdown and a small gh/git wrapper script; no code paths to test)


mcgibbon:
Claude: bootstrap note — this PR's own description was written by the authoring session (Fable), not via the `write-pr-description` skill it introduces, since the skill doesn't exist on main yet. Descriptions of subsequent PRs would go through the pinned model.

mcgibbon:
Pre-review agent: independent pre-review (no authoring context). Structure is fine — one commit, based on current `origin/main` (267981307), 3 files, +14/-1; no split or reorder needed. `pre-commit` and `docs/build` are green.

**Verified so far, so the reviewer doesn't have to:** `claude-opus-4-6[1m]` is a real, first-class model value in the installed Claude Code (2.1.237) — the bundle carries it with the display label "Opus 4.6 (1M context)", so the pin resolves today. `model:` is also a documented SKILL.md frontmatter field accepting the same values as `/model`.

## Should fix before review

**1. The pin is turn-scoped, not skill-scoped — the description claims a guarantee the mechanism doesn't give.**
Per the skills frontmatter reference, `model:` "applies for the rest of the current turn and is not saved to settings; the session model resumes on your next prompt." Two consequences the PR doesn't account for:

- Any turn boundary during description authoring drops back to the session model, silently. The common flow — draft the description, the user asks for a change, amend it — writes the amendment on the session model, which is exactly what the PR forbids. So does a permission denial followed by a re-prompt.
- Conversely, the override leaks *forward*: everything else the session does after the skill in that same turn also runs on 4.6, not just the description.

`context: fork` (documented as setting the forked subagent's model instead of overriding the session) scopes the pin to exactly the description-writing work and is immune to turn boundaries. Was it considered? Either switch to it, or state the turn-scope caveat in the description rather than implying an unconditional pin.

**2. "the only repo-level Claude configuration mechanism that binds a specific piece of work to a specific model" is an overclaim.**
A `.claude/agents/*.md` subagent definition also carries `model:` and is equally repo-level. ace has no `.claude/agents/` today, but nothing prevents one, and per finding 1 a subagent would bind *more* tightly than a skill. Since this description becomes the squash-merge commit message, drop "the only" or say why a subagent was rejected.

**3. AGENTS.md's amend trigger doesn't route to the skill.**
The new mandate lands at line 64 under "### PR description template", but the instruction that actually *triggers* an amendment lives in "### Committing" (lines 55-56: "check whether the PR description needs to be updated", "When amending a PR description, make sure it still reflects the change _from main_"). A session following §Committing after pushing a commit has no signal the new rule exists and will amend inline. Given the PR's stated goal is "so sessions know the requirement exists", this is the one path most likely to miss it — add a pointer at line 56, or move that guidance into the skill.

## Suggestions

**4. Template checkboxes were dropped, and the skill doesn't say to keep them.**
`.github/pull_request_template.md` carries two `- [ ]` lines, and the convention in recent merged PRs is to keep and annotate them in place, not delete them (#1435: `- [x] Tests added (no new functionality; the existing suite pins behavior across the restructure)`; #1445 and #1440 keep both lines). This description replaces them with prose. Since the skill is now the sole authority on description shape and this PR is its reference example, worth restoring them here and having step 2 say the checkboxes are ticked or annotated, never removed.

**5. Step 1 should fetch first.** `git diff origin/main...HEAD` against a stale `origin/main` describes the delta to the wrong base — and ace main moves fast. Prepending `git fetch origin main` protects the exact property line 56 asks for.

**6. Silent failure has no guard.** Per the docs, when the named model is unavailable "Claude Code ignores the override ... and the skill or command runs on the session model" — no error. The string is valid today (verified above), but the day 4.6 is retired the pin degrades to the session model with nothing to notice it. One line in the skill body telling the session to state which model it is running on makes both the success and the silent fallback visible.

**7. Record why 4.6.** Neither the skill nor AGENTS.md carries the rationale or a revisit trigger, and 4.6 is the oldest of the current Opus family (the `opus` alias now resolves to Opus 5 on the Anthropic API). If the choice is deliberate, one line saying so stops a future session from either bumping it or leaving it stale forever.

**8. "never inline" is absolute, but AGENTS.md is the cross-harness guidance file** ("the single source of agent guidance for the `ai2cm/ace` repository") and only Claude Code can invoke `.claude/skills/`. A non-Claude-Code agent cannot comply with the sentence as written. Scope it, or accept it as aspirational.

## Nit

**9.** `- Improving the PR title, and the description via the `write-pr-description` skill (required for any description edit).` — the parenthetical restates AGENTS.md line 64 and the comma splice reads awkwardly. Suggest ending at "skill."

Verdict: not yet ready for review — findings 1-3 want a fix or an explicit decline first (and the PR is still a draft with `cpu`/`gpu` in flight). Findings 4-9 are the author's call. Comment-only pre-review: nothing pushed to the branch.

mcgibbon:
Claude: applied the pre-review findings in a2c9650cc and a description edit:
- (1) Switched the skill to `context: fork` so the pin is scoped to the description work rather than turn-scoped; the skill body states why.
- (2) Rewrote the description to drop the 'only mechanism' overclaim — it now names the subagent alternative and why a skill was chosen.
- (3) The §Committing amend path now routes through the skill explicitly.
- (4) Description checkboxes restored and annotated in place; the skill instructs the same.
- (5) Skill fetches origin/main before diffing.
- (6) Skill states the model it actually runs, so a silent fallback (e.g. model retirement) is visible.
- (7) Pin provenance recorded in the skill (Jeremy, 2026-08-20).
- (8) AGENTS.md routing sentence scoped to Claude Code sessions.
- (9) Pre-review bullet reworded.

Note the description edit itself was made by the authoring session (Fable) — covered by the bootstrap-note comment above.

mcgibbon:
Pre-review agent: delta re-review, 1b5d4d303 → a2c9650cc, plus the edited description. All nine findings are addressed faithfully; the reply comment matches what the diff actually does.

**Finding 1's remedy verified against the installed Claude Code (2.1.237), not just the docs.** This was the one worth checking, because the `Agent` tool's own `fork` mode ignores a `model` override and runs on the parent's model — if skill `context: fork` shared that behavior, the fix would have silently defeated the whole PR. It does not: `context: fork` is a first-class SKILL.md field ("`fork` spawns a subagent", alongside the `agent:` and `background:` companions), and a forked skill is spawned with `model: <skill manifest>.model` — the frontmatter value, not the session model. So the pin now holds for exactly the description work and survives turn boundaries. Two incidental confirmations while I was in there: there is a built-in `onModelRestricted` warning path, so a retired or unentitled pin is not entirely silent (finding 6's self-report is belt-and-braces on top of it, which is fine); and there is a fork-recursion guard that returns "Skill ... is already executing" — so the fork reading AGENTS.md's "use the write-pr-description skill" cannot loop.

Findings 2, 3, 5, 7, 8, 9 are clean and need no further comment.

## One new accuracy issue (finding 4's fix)

Both checkboxes are now ticked for work that wasn't done:

- `- [x] Tests added (agent-configuration markdown only; no code paths to test)` — asserts tests were added; none were.
- `- [x] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated (no dependency changes)` — asserts an image rebuild happened; it didn't.

The convention in the merged PRs I sampled is to tick only what was actually done: #1445 leaves both unchecked, #1440 ticks Tests (real tests) and leaves deps unchecked, #1435 ticks Tests with a justification that the existing suite covers the restructure. None of them ticks the deps box. Since the description becomes the permanent squash-merge commit message, the honest form here is unchecked-with-reason for both.

This one is partly my own fault — my finding 4 said "ticked or annotated", which invited the tick. The more useful fix is in the skill, since it now encodes the pattern for every future description: step 2's example (`- [x] Tests added (no new functionality; ...)`, borrowed from #1435 where the tick was earned) gives no guidance for the genuinely-inapplicable case. Worth one clause — tick only what was actually done; leave inapplicable items unchecked with the reason — so this doesn't repeat.

## Optional

The fork's report goes to the parent agent rather than to the user, so finding 6's "state which model you are running" lands in the parent's context and reaches Jeremy only if the parent relays it. Given the built-in `onModelRestricted` warning above, this is adequate as-is; mentioning it only so the limitation is known.

Two trims, take or leave: AGENTS.md:55's new clause is unqualified while :64 now says "In Claude Code sessions" — the "(see "PR description template")" pointer resolves it, so this is cosmetic. And the skill body's explanation of *why* fork-scoping beats turn-scoping is addressed to the fork, which does not need it to do the job; the operative sentence is the "do not bump it without his say-so". Harmless if you would rather a future maintainer find the rationale in the file.

Verdict: **ready for review** once the draft flag is cleared and CI lands green (`pre-commit` is green; `cpu`/`gpu`/`cpu-very-fast-no-healpix` were still running at review time — markdown-only, so no substantive risk). The checkbox item above is worth fixing but does not need another pre-review round-trip; Jeremy can take or leave it in review. Comment-only: nothing pushed to the branch.

mcgibbon:
Claude: applied the re-review's non-blocking note in 2f389b2dd — checkboxes now unchecked with in-place reasons, and the skill's checklist clause says to tick only work actually done. Pre-review verdict was ready-for-review; draft flag cleared.

.claude/skills/write-pr-prose/SKILL.md
```diff
@@ -0,0 +1,14 @@
+---
+name: write-pr-prose
+description: Write or amend a PR title or description. Use whenever creating a PR or updating the title or description of an existing PR, including after pushing new commits to a branch with an open PR.
+model: claude-opus-4-6[1m]
+context: fork
+---
+
+You are writing the title and/or description for a PR: the one for the current checked-out branch, or one the user names. PR titles and descriptions in this repository are authored through this skill so that the model pinned in its frontmatter writes them; `context: fork` scopes the pin to exactly this work (a bare `model:` pin is turn-scoped and would leak into the rest of the turn or lapse on the next prompt). The pinned model is a deliberate choice (Jeremy, 2026-08-20) — do not bump it without his say-so.
```
mcgibbon:
```suggestion
You are writing the title and/or description for a PR: the one for the current checked-out branch, or one the user names. PR titles and descriptions in this repository are authored through this skill so that the model pinned in its frontmatter writes them; `context: fork` scopes the pin to exactly this work (a bare `model:` pin is turn-scoped and would leak into the rest of the turn or lapse on the next prompt). The pinned model is a deliberate choice (Jeremy, 2026-08-20) — do not bump it without user permission.
```

.claude/skills/write-pr-prose/SKILL.md
```diff
@@ -0,0 +1,14 @@
...
+
+Begin your report by stating which model you are actually running, so a silent fallback to the session model (e.g. if the pinned model is ever retired) stays visible.
```
mcgibbon:
```suggestion
Once you've updated the PR prose, leave a comment on the PR which states you have done so and which model you are actually running, so a silent fallback to the session model (e.g. if the pinned model is ever retired) stays visible.
```

.claude/skills/write-pr-prose/SKILL.md
```diff
@@ -0,0 +1,14 @@
...
+
+1. Run `git fetch origin main`, then read the full diff from main (`git diff origin/main...HEAD`, or `gh pr diff <number>`), not just the recent commits, so the title and description reflect the change from main per the guidance in AGENTS.md.
```
mcgibbon: The agent should check PR comments as well, and any existing title or description, as those can be very informative as to intent. Maybe a helper script could make this more convenient?

.claude/skills/write-pr-prose/SKILL.md
mcgibbon: Claude: applied verbatim in d06d0dd2a.

.claude/skills/write-pr-prose/SKILL.md
mcgibbon: Claude: applied verbatim in d06d0dd2a — the fork now reports by commenting on the PR, naming the model it actually ran.

.claude/skills/write-pr-prose/SKILL.md
mcgibbon: Claude: added in d06d0dd2a — step 1 now runs a new helper, `.claude/skills/write-pr-prose/gather_pr_context.sh [<pr-number>]`, which fetches origin/main and prints the existing title/description, the PR discussion (issue comments, reviews, and inline review comments), and the full diff from main; the step says to read all of it because existing prose and comments carry intent the diff alone does not. Smoke-tested against this PR (all sections render, including these three inline comments) and it falls back to the local diff when no PR exists yet (the `gh pr create` path).

.claude/skills/pre-review/SKILL.md
```diff
@@ -11,7 +11,7 @@ For example, this could mean:
  - Re-structuring the PR, either by splitting it into multiple smaller PRs or by re-ordering commits, to make it easier to review.
  - Catching and fixing any issues you can find before the PR is seen by the reviewer.
  - Making comments on the PR to point out potential issues or areas of concern for the reviewer to focus on.
- - Improving the PR description and title.
+ - Improving the PR title and description (title and description edits go through the `write-pr-prose` skill).
```
frodre:
```suggestion
 - Improving the PR title and description (title and description edits should use the `write-pr-prose` skill).
```

frodre: Not sure if human style adjustments to the skill matter?

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,27 @@
...
+    gh pr view "$@"
+    echo
+    echo "=== Discussion (issue comments and reviews) ==="
+    gh pr view "$@" --comments
+    echo
+    echo "=== Inline review comments ==="
+    repo=$(gh repo view --json nameWithOwner --jq .nameWithOwner)
+    gh api "repos/${repo}/pulls/${pr_number}/comments" \
+        --jq '.[] | "--- \(.user.login) on \(.path) line \(.line // .original_line):\n\(.body)\n"'
+    echo "=== Diff from origin/main ==="
```
frodre: `pr diff` shows diff from PR base, which is functionally main for most PRs, but might not be for stacked PRs

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,27 @@
...
+    gh pr diff "$@"
+else
+    echo "(no existing PR found; showing the local diff from origin/main)"
+    echo "=== Diff from origin/main ==="
+    git diff origin/main...HEAD
```
frodre: pegged to main again, which means this can't be used for stacked PRs, which I think is fine given most use cases.  could ask for this to be configurable with a default

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,27 @@
...
+if pr_number=$(gh pr view "$@" --json number --jq .number 2>/dev/null); then
+    echo "=== Existing title and description ==="
+    gh pr view "$@"
+    echo
+    echo "=== Discussion (issue comments and reviews) ==="
+    gh pr view "$@" --comments
+    echo
+    echo "=== Inline review comments ==="
+    repo=$(gh repo view --json nameWithOwner --jq .nameWithOwner)
+    gh api "repos/${repo}/pulls/${pr_number}/comments" \
```
frodre: claude suggests you may need a --paginate to prevent truncation of comments from long threads

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+# Print the context needed to write a PR's title and description: the
+# existing title/description, the PR discussion (issue comments, reviews,
+# and inline review comments), and the full diff from main.
+# Usage: gather_pr_context.sh [<pr-number>]  (defaults to the current branch's PR)
+set -euo pipefail
+
+git fetch origin main --quiet || echo "(warning: could not fetch origin/main; diff base may be stale)"
+
+if pr_number=$(gh pr view "$@" --json number --jq .number 2>/dev/null); then
```
frodre: setting pr_number = $@ prior would be more human readable throughout

frodre: We have a failure echo if the fetch fails, do we want similar failure notifications for any of the actual context printouts for the agent?  I guess this brings up the question of what does the agent do for failures in some or all of the context retreivals?  The human would of course see that, but maybe useful to actually have it comment on information about retrieval coverage

.claude/skills/write-pr-prose/SKILL.md
```diff
@@ -0,0 +1,14 @@
+---
+name: write-pr-prose
+description: Write or amend a PR title or description. Use whenever creating a PR or updating the title or description of an existing PR, including after pushing new commits to a branch with an open PR.
+model: claude-opus-4-6[1m]
+context: fork
+---
+
+You are writing the title and/or description for a PR: the one for the current checked-out branch, or one the user names. PR titles and descriptions in this repository are authored through this skill so that the model pinned in its frontmatter writes them; `context: fork` scopes the pin to exactly this work (a bare `model:` pin is turn-scoped and would leak into the rest of the turn or lapse on the next prompt). The pinned model is a deliberate choice (Jeremy, 2026-08-20) — do not bump it without user permission.
+
+Once you've updated the PR prose, leave a comment on the PR which states you have done so and which model you are actually running, so a silent fallback to the session model (e.g. if the pinned model is ever retired) stays visible.
```
frodre: Does the model know which model it is or can we source it with a specific command/skill?

frodre (commented):
Thanks, @mcgibbon.  Here are some feedback points / questions

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Agreed, for a stacked PR, the description should still follow the diff of that PR based on the first one, not based on main.

.claude/skills/write-pr-prose/SKILL.md
mcgibbon: Good question, I'll let Claude answer it.

.claude/skills/pre-review/SKILL.md
mcgibbon: Claude: this landed via the applied suggestion (cc2e3b2c0); my identical local edit collapsed in the rebase. On whether style matters: yes, mildly — the skill text is the instruction the executing agent follows, so normative phrasing like "should use" is just as binding and reads better to humans. Style-only edits are safe.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: right — `gh pr diff` was already diffing against the PR's base, so stacked PRs got the correct diff on that line; the bugs were the hardcoded `git fetch origin main` and the section header claiming origin/main. Fixed in 27905f9e0: the script reads the PR's `baseRefName`, fetches that, and labels the diff section with it, so a stacked PR is described against its own base as @mcgibbon says.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: made configurable in 27905f9e0 — the no-PR fallback now defaults to the repo's default branch (queried from GitHub, not hardcoded main) and takes an optional second argument to override the base. When a PR exists its own base is always used, so the override only applies to the not-yet-created-PR path.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: added `--paginate` in 27905f9e0.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: done in 27905f9e0 — `pr_number` is resolved once up front (from the argument or the current branch's PR) and used throughout instead of passing "$@" around.

mcgibbon: Claude: addressed in 27905f9e0. The script no longer uses `set -e` (a mid-script failure used to kill everything after it, silently dropping the diff section); each section now falls back to a "(warning: could not retrieve ...)" line and the rest still runs, so the agent sees exactly which context is missing rather than a truncated printout. On the agent side, the skill now tells the fork to include any such warnings in the PR comment it already leaves, so retrieval coverage is visible to humans on the PR, not just in the agent's context.

.claude/skills/write-pr-prose/SKILL.md
mcgibbon: Claude: reliable, and it isn't the model guessing. In Claude Code the harness injects the exact model ID into every session's system prompt ("You are powered by the model named ...; the exact model ID is ..."), and a forked skill gets the same treatment for the model it was spawned on — so the fork reads its identity from harness-supplied text, not from weights (models' unprompted self-identification IS unreliable, which is why sourcing it from the system prompt matters). There's no shell command to query it from inside a session, but none is needed. Belt-and-braces on top: Claude Code has a built-in onModelRestricted warning path when a pinned model can't be honored, so a retired/unentitled pin isn't fully silent either.

mcgibbon:
Claude: updated the PR description via the `write-pr-prose` skill running on `claude-opus-4-6[1m]` (Opus 4.6, 1M context). Two bullets revised to reflect the current script and skill behavior:
- `gather_pr_context.sh`: now describes base-branch diffing for stacked PRs, repo-default-branch fallback with optional override, `--paginate` on comment retrieval, and per-section `(warning: ...)` fallback instead of aborting.
- `SKILL.md`: now mentions that the fork reports retrieval gaps (the script's `(warning: ...)` lines) in its PR comment alongside the model identity.

No retrieval warnings from the gather script for this run.

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+
+pr_number="${1:-}"
+base_override="${2:-}"
+
+if [ -z "$pr_number" ]; then
+    pr_number=$(gh pr view --json number --jq .number 2>/dev/null || true)
+fi
+
+if [ -n "$pr_number" ]; then
+    base=$(gh pr view "$pr_number" --json baseRefName --jq .baseRefName 2>/dev/null || true)
```
frodre: probably need a check for this, [ -z "$base" ] && echo "(warning: could not determine PR base branch)"

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+    git fetch origin "${base:-main}" --quiet 2>/dev/null \
+        || echo "(warning: could not fetch origin/${base:-main}; diff base may be stale)"
+    echo "=== Existing title and description ==="
+    gh pr view "$pr_number" || echo "(warning: could not retrieve title/description)"
+    echo
+    echo "=== Discussion (issue comments and reviews) ==="
+    gh pr view "$pr_number" --comments || echo "(warning: could not retrieve discussion)"
```
frodre: --paginate

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo
+    echo "=== Inline review comments ==="
+    repo=$(gh repo view --json nameWithOwner --jq .nameWithOwner 2>/dev/null || true)
+    gh api --paginate "repos/${repo}/pulls/${pr_number}/comments" \
+        --jq '.[] | "--- \(.user.login) on \(.path) line \(.line // .original_line):\n\(.body)\n"' \
+        || echo "(warning: could not retrieve inline review comments)"
+    echo "=== Diff from PR base (${base:-unknown}) ==="
+    gh pr diff "$pr_number" || echo "(warning: could not retrieve diff)"
```
frodre: Missing the diff probably means we cannot write the PR update.  Probably worth failing loudly in either case if this can't be retrieved.

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+else
+    base="$base_override"
+    if [ -z "$base" ]; then
+        base=$(gh repo view --json defaultBranchRef --jq .defaultBranchRef.name 2>/dev/null || true)
```
frodre: This could still fail and give an empty `base` silently defaulting to main

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+    fi
+    base="${base:-main}"
+    git fetch origin "$base" --quiet 2>/dev/null \
+        || echo "(warning: could not fetch origin/$base; diff base may be stale)"
+    echo "(no existing PR found; showing the local diff from origin/$base)"
+    echo "=== Diff from origin/$base ==="
+    git diff "origin/$base...HEAD" || echo "(warning: could not compute diff)"
```
frodre: Just for coverage. A diff failure should prevent write for this agent

AGENTS.md
```diff
@@ -52,7 +52,7 @@ not validate cross-backend correctness.
 
 When committing changes, do not include Co-Authored-By: Claude in the commit message. This is because Claude bears no accountability or responsibility for the changes, the user is the sole author accountable.
 
-When committing to a branch that has an open PR, check whether the PR description needs to be updated to reflect the changes.
+When committing to a branch that has an open PR, check whether the PR description needs to be updated to reflect the changes; update it via the `write-pr-prose` skill (see "PR description template").
 When amending a PR description, make sure it still reflects the change _from main_, not the change from a previous, ephemeral state of the PR.
```
frodre: Needs to be updated to reference main/ or requested base

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo
+    echo "=== Inline review comments ==="
+    repo=$(gh repo view --json nameWithOwner --jq .nameWithOwner 2>/dev/null || true)
+    gh api --paginate "repos/${repo}/pulls/${pr_number}/comments" \
+        --jq '.[] | "--- \(.user.login) on \(.path) line \(.line // .original_line):\n\(.body)\n"' \
+        || echo "(warning: could not retrieve inline review comments)"
+    echo "=== Diff from PR base (${base:-unknown}) ==="
+    gh pr diff "$pr_number" || echo "(warning: could not retrieve diff)"
+else
+    base="$base_override"
```
frodre: Can't override in this case because the base_override requires an argument before.

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,48 @@
...
+pr_number="${1:-}"
+base_override="${2:-}"
+
+if [ -z "$pr_number" ]; then
+    pr_number=$(gh pr view --json number --jq .number 2>/dev/null || true)
+fi
+
+if [ -n "$pr_number" ]; then
+    base=$(gh pr view "$pr_number" --json baseRefName --jq .baseRefName 2>/dev/null || true)
+    git fetch origin "${base:-main}" --quiet 2>/dev/null \
```
frodre: Everything uses `gh` to pull from remote.  Is the fetch necessary?

frodre (changes requested):
Still a few more tweaks necessary.

AGENTS.md
mcgibbon: True, I'd expect this not to be an issue for a human but agents can be somewhat pedantic at times.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: added in 8e2a957d0 — an empty `base` now prints "(warning: could not determine PR base branch)" and the diff section header shows "unknown" rather than implying a base.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: done in 8e2a957d0 — the single `gh pr view --comments` call (which can truncate) is replaced by paginated `gh api --paginate` calls for issue comments and reviews, alongside the already-paginated inline review comments.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: agreed and done in 8e2a957d0 — a failed diff retrieval now prints an ERROR to stderr and exits 1 in both paths, and the skill's step 1 says a script error means stop and report rather than write prose. Discussion sections keep the non-fatal "(warning: ...)" behavior since prose can still be written (with the gap reported) without them.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: added in 8e2a957d0 — a failed default-branch query now prints "(warning: could not determine repo default branch; assuming main)" before the main fallback.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: done in 8e2a957d0 — the local-diff fallback is fatal on failure too (ERROR + exit 1), same as the PR path.

AGENTS.md
mcgibbon: Claude: updated in 8e2a957d0 — AGENTS.md now says the description must reflect the change "from the PR's base branch (main, unless the PR is stacked on another branch)".

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: good catch — fixed in 8e2a957d0 by making it a flag: `gather_pr_context.sh [<pr-number>] [--base <branch>]`, so the override works with no PR-number argument (the only case where it applies). Verified from a detached checkout with no PR: `--base main` diffs correctly and `--base bogus-branch` fails loudly.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Claude: it wasn't — in the PR path everything (view, comments, diff) comes from the GitHub API, so the fetch was a leftover from the earlier local-diff version; removed in 8e2a957d0 with a comment noting no local fetch is needed there. The no-PR fallback keeps its fetch because `git diff origin/<base>...HEAD` needs the ref locally.

mcgibbon:
Updated the PR description via the `write-pr-prose` skill running on `claude-opus-4-6[1m]` (Opus 4.6, 1M context). Two bullets revised:
- `gather_pr_context.sh`: now describes paginated retrieval of issue comments, reviews, and inline review comments; warnings when the PR base or repo default branch cannot be determined; the fatal-on-missing-diff behavior (both paths); the `--base <branch>` flag on the no-PR fallback; and the absence of a local fetch in the PR path.
- `AGENTS.md`: now notes the amend guidance references the PR's base branch rather than main.

Title, summary paragraph, and checklist lines unchanged. No retrieval warnings from the gather script for this run.

mcgibbon:
> Updated the PR description via the `write-pr-prose` skill running on `claude-opus-4-6[1m]` (Opus 4.6, 1M context). Two bullets revised:
> 
> * `gather_pr_context.sh`: now describes paginated retrieval of issue comments, reviews, and inline review comments; warnings when the PR base or repo default branch cannot be determined; the fatal-on-missing-diff behavior (both paths); the `--base <branch>` flag on the no-PR fallback; and the absence of a local fetch in the PR path.
> * `AGENTS.md`: now notes the amend guidance references the PR's base branch rather than main.
> 
> Title, summary paragraph, and checklist lines unchanged. No retrieval warnings from the gather script for this run.

Interesting to note this PR is kind of an example of its own effect.

AGENTS.md
frodre: >  agents can be somewhat pedantic at times

agent: "Well you told me to!"

.claude/skills/write-pr-prose/gather_pr_context.sh
```diff
@@ -0,0 +1,73 @@
...
+    if [ -z "$base" ]; then
+        base=$(gh repo view --json defaultBranchRef --jq .defaultBranchRef.name 2>/dev/null || true)
+        [ -z "$base" ] && echo "(warning: could not determine repo default branch; assuming main)"
+    fi
+    base="${base:-main}"
+    git fetch origin "$base" --quiet \
+        || echo "(warning: could not fetch origin/$base; diff base may be stale)"
+    echo "(no existing PR found; showing the local diff from origin/$base)"
+    echo "=== Diff from origin/$base ==="
+    if ! git diff "origin/$base...HEAD"; then
```
frodre: Claude getting deep into the edge cases here, but technically uncommitted changes (no diff to operate on) would exit 0 and still pass here on a local feature.  This would then get a no information description written.  This is maybe only important if the agent also creates a PR for an empty branch, which I don't think would be a big risk.  We could just limit the skill to amending already existing PRs?

frodre (approved):
LGTM, I think this covers most of our basic usage and any testing from here will reveal any flaws in the agentic use of it.

.claude/skills/write-pr-prose/gather_pr_context.sh
mcgibbon: Agreed, I think we should delete the "showing the local diff from origin/$base" entirely and just say there's no PR found.

mcgibbon: Claude: done in 4f3a97ad5, exactly as agreed — the local-diff fallback (and the `--base` flag that only served it) is deleted; with no existing PR the script now prints "ERROR: no existing PR found for this branch; this script only gathers context for existing PRs." and exits 1. The skill's frontmatter and steps now scope it to existing PRs (`gh pr edit` only), and its step 1 treats any script error as stop-and-report. Verified: the PR path still prints all sections; a detached no-PR checkout gets the error and exit 1. No other behavior changed.

mcgibbon:
Updated the PR description via the `write-pr-prose` skill running on `claude-opus-4-6[1m]` (Opus 4.6, 1M context). One bullet revised:
- `gather_pr_context.sh`: replaced the clause describing the no-PR fallback and `--base <branch>` flag (removed in 4f3a97ad5) with the new behavior: the script requires an existing PR, exiting with an error otherwise (same fatal handling as a failed diff); also fixed the collateral "in both paths" and "in the PR path" phrases that implicitly referenced the removed fallback path.

Title, summary paragraph, all other bullets, and checklist lines unchanged. No retrieval warnings from the gather script for this run.

mcgibbon:
PR description updated (model: claude-opus-4-6[1m]).

Change: in the `gather_pr_context.sh` bullet, "warns when the PR base or repo default branch cannot be determined" → "warns when the PR base branch cannot be determined" (the repo-default-branch warning belonged to the fallback removed in 4f3a97ad5).

No other changes. Context warnings from gather script: ran outside the helper (direct gh call), so no warning output to report.

Up to two sentences summarizing the results.

<details>
  
  Full review goes here.
  
</details>

PR #1458, opened 2026-08-25
Allow per-channel fill values for static spatial masking

elynnwu:
Masked-input handling currently applies one fill value to every masked channel: either `0.0` everywhere or the channel mean everywhere. The right choice differs by channel, and a 1° ocean comparison over 130-year and 200-year free-running rollouts showed both directions mattering at once — the channel mean is better for every 2D surface field, while `0.0` is better for the 19-level velocity and temperature profiles by 1.2–1.4×.

There is a plain reason. Where a masked cell holds no data — temperature and salinity below the sea floor — there is no true value to restore, so the channel mean is a neutral filler and `0.0` injects a large out-of-range value near coastlines. Where a masked cell is genuinely dry, being marked out of range is usable information, and nothing else in the inputs states it per level: the land mask is a single 2D field, whereas the wet fraction falls from 0.69 at the surface to 0.07 at the deepest level, with 89% of surface-ocean columns dry by then. The mean erases that distinction; `0.0` hands it to the network per channel.

This adds the configuration needed to express both at once. It does not change the fill any existing config produces.

Changes:
- `fme.core.spatial_masking.StaticSpatialMaskingConfig.fill_value_overrides` — new optional mapping from channel name or prefix to a fill value (`"mean"` or a float), overriding `fill_value` for the channels it selects. Keys select channels through `NameAndPrefixMatcher`, the same relation `exclude_names_and_prefixes` uses, so a bare `name` selects `name` and `name_<level>`, a `name_` prefix selects `name_<level>`, and `name_<level>` selects exactly that channel. Where several keys select one channel the longest key wins, which orders `name_<level>` over `name_` over `name`; a channel no key selects keeps `fill_value`. A key selecting no channel raises at build time, so a typo fails loudly instead of silently doing nothing.
- `StaticSpatialMaskingConfig.build` — resolves the mapping eagerly over the normalizer's channel names when overrides are set, since a per-channel mapping cannot be deferred to a `defaultdict` or to `means` itself. `means` becomes required in that case, both to enumerate channels and to supply any `"mean"` fills.

`StaticSpatialMasking` already keyed its fill values by channel name, so the change is confined to the config layer; the runtime is untouched. `fme/coupled` reaches this config through the same `StepperConfig.input_masking` field, so it picks the option up with no mirror change. Output masking is unaffected: it fills NaN by construction and has no per-channel fill to configure.

Fixed along the way, both in the function this PR modifies:
- An integer `fill_value` no longer silently means the channel mean. dacite accepts an unsuffixed YAML `0` for `Literal["mean"] | float`, and `isinstance(self.fill_value, float)` then routed it to the means branch. Without this, `fill_value: 0` would have meant `0.0` with overrides set and the channel mean without them. No config in the repo passes an integer `fill_value`, so no existing run changes behavior.
- `fill_value_overrides` values are type-checked at load by dacite (`UnionMatchError`) rather than by a hand-rolled `isinstance` in `__post_init__`, which was unreachable from YAML.

With `fill_value_overrides` unset, `build()` takes the same branches it took before. Tests assert the concrete fill each channel receives on both branches, so the default path is pinned to values rather than to itself.

- [x] Tests added

### DIFF
--- a/fme/core/spatial_masking.py
+++ b/fme/core/spatial_masking.py
@@ -50,21 +50,56 @@ class StaticSpatialMaskingConfig:
         mask_value: Value of the mask variable in masked regions. Either 0 or 1.
         fill_value: A float fill value to use inside of masked regions. Can also be
             "mean", in which case the normalizer means are used as channel-specific
-            fill values.
+            fill values. Applies to every masked channel that no
+            ``fill_value_overrides`` entry selects.
         exclude_names_and_prefixes: Names (2D variables) and prefixes (3D variables)
             to exclude when applying the mask.
+        fill_value_overrides: Per-channel fill values overriding ``fill_value``,
+            keyed by name (2D variables) or prefix (3D variables), following the
+            same convention as ``exclude_names_and_prefixes``. Values take the same
+            form as ``fill_value``. A channel selected by an exact name uses that
+            entry; otherwise the longest matching prefix wins; otherwise
+            ``fill_value`` applies. Every key must select at least one channel, so
+            a typo raises at build time rather than silently doing nothing.
+
+            The motivating case is that the right fill differs by channel. A field
+            whose masked cells hold no data (temperature, salinity) has no true
+            value to restore, so the channel mean is a neutral filler. A field
+            whose masked cells are genuinely dry -- the depth-resolved channels
+            below the sea floor -- carries usable information in being marked out
+            of range, which the mean erases.
 
     """
 
     mask_value: int
     fill_value: Literal["mean"] | float = 0.0
     exclude_names_and_prefixes: list[str] | None = None
+    fill_value_overrides: dict[str, Literal["mean"] | float] | None = None
 
     def __post_init__(self):
         if self.mask_value not in [0, 1]:
             raise ValueError(
                 f"mask_value must be either 0 or 1, but got {self.mask_value}"
             )
+        for key, value in (self.fill_value_overrides or {}).items():
+            if value != "mean" and not isinstance(value, int | float):
+                raise ValueError(
+                    "fill_value_overrides values must be a float or 'mean', but "
+                    f"got {value!r} for {key!r}"
+                )
+
+    def _resolve_fill_value(self, name: str) -> Literal["mean"] | float:
+        """The fill value for one channel.
+
+        Exact name wins, then the longest matching prefix, then ``fill_value``.
+        """
+        overrides = self.fill_value_overrides or {}
+        if name in overrides:
+            return overrides[name]
+        prefixes = [k for k in overrides if name.startswith(k)]
+        if prefixes:
+            return overrides[max(prefixes, key=len)]
+        return self.fill_value
 
     def build(self, mask: HasGetSpatialMask, means: TensorMapping | None = None):
         """
@@ -72,23 +107,56 @@ class StaticSpatialMaskingConfig:
 
         """
         exclude = NameAndPrefixMatcher(self.exclude_names_and_prefixes)
-        if isinstance(self.fill_value, float):
+        if not self.fill_value_overrides:
+            # Unchanged from before per-channel overrides existed.
+            if isinstance(self.fill_value, float):
+                return StaticSpatialMasking(
+                    mask_value=self.mask_value,
+                    fill_value=collections.defaultdict(
+                        lambda: torch.as_tensor(self.fill_value)
+                    ),
+                    mask=mask,
+                    exclude=exclude,
+                )
+            if means is None:
+                raise ValueError(
+                    "fill_values mapping required by build unless configured "
+                    "fill_value is a float."
+                )
             return StaticSpatialMasking(
                 mask_value=self.mask_value,
-                fill_value=collections.defaultdict(
-                    lambda: torch.as_tensor(self.fill_value)
-                ),
+                fill_value=means,
                 mask=mask,
                 exclude=exclude,
             )
+        # With overrides the mapping is resolved per channel, so it has to be
+        # built eagerly over the known channel names rather than deferred to a
+        # defaultdict or to `means` itself.
         if means is None:
             raise ValueError(
-                "fill_values mapping required by build unless configured "
-                "fill_value is a float."
+                "fill_values mapping required by build when fill_value_overrides "
+                "is set, to enumerate the channels and to supply any 'mean' fills."
+            )
+        unmatched = [
+            key
+            for key in self.fill_value_overrides
+            if not any(name == key or name.startswith(key) for name in means)
+        ]
+        if unmatched:
+            raise ValueError(
+                "fill_value_overrides keys match no channel: "
+                f"{sorted(unmatched)}. Keys are names or prefixes, as in "
+                "exclude_names_and_prefixes."
+            )
+        fill_mapping = {}
+        for name in means:
+            resolved = self._resolve_fill_value(name)
+            fill_mapping[name] = (
+                means[name] if resolved == "mean" else torch.as_tensor(float(resolved))
             )
         return StaticSpatialMasking(
             mask_value=self.mask_value,
-            fill_value=means,
+            fill_value=fill_mapping,
             mask=mask,
             exclude=exclude,
         )
--- a/fme/core/spatial_masking.py
+++ b/fme/core/spatial_masking.py
@@ -1,5 +1,6 @@
 import collections
 import dataclasses
+from collections.abc import Iterable
 from typing import Literal, Protocol, runtime_checkable
 
 import torch
@@ -55,19 +56,19 @@ class StaticSpatialMaskingConfig:
         exclude_names_and_prefixes: Names (2D variables) and prefixes (3D variables)
             to exclude when applying the mask.
         fill_value_overrides: Per-channel fill values overriding ``fill_value``,
-            keyed by name (2D variables) or prefix (3D variables), following the
-            same convention as ``exclude_names_and_prefixes``. Values take the same
-            form as ``fill_value``. A channel selected by an exact name uses that
-            entry; otherwise the longest matching prefix wins; otherwise
-            ``fill_value`` applies. Every key must select at least one channel, so
-            a typo raises at build time rather than silently doing nothing.
+            keyed by name (2D variables) or prefix (3D variables), selecting
+            channels by the same convention as ``exclude_names_and_prefixes``.
+            Values take the same form as ``fill_value``. Where more than one key
+            selects a channel the longest key wins, so an exact ``name_<level>``
+            beats the ``name_`` prefix, which beats the bare ``name``; a channel
+            no key selects takes ``fill_value``. Every key must select at least
+            one channel, so a key that selects nothing raises at build time
+            rather than silently doing nothing.
 
-            The motivating case is that the right fill differs by channel. A field
-            whose masked cells hold no data (temperature, salinity) has no true
-            value to restore, so the channel mean is a neutral filler. A field
-            whose masked cells are genuinely dry -- the depth-resolved channels
-            below the sea floor -- carries usable information in being marked out
-            of range, which the mean erases.
+            Use this where the right fill differs by channel: the channel mean is
+            a neutral filler where masked cells hold no data, whereas ``0.0``
+            marks them as out of range, which is usable information where the
+            masked cells are genuinely dry.
 
     """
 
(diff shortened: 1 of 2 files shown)


elynnwu:
Pre-review agent: I pre-reviewed this cold (no authoring context) and applied fixes in a06a7bf. Three of them are behavior fixes with regression tests, so they are worth a look rather than a skim. All 9 audit passes are below; the two items still needing a decision are separate inline comments.

## Applied

**1. Override keys did not select channels the way the docs said.** The docstring and description both claimed keys follow the same name-or-prefix convention as `exclude_names_and_prefixes`, but resolution used a raw `name.startswith(key)`, which is a wider relation. Measured against `NameAndPrefixMatcher` on a channel set of `so_0, so_18, sos, sob, thetao_0, thetao_bias, sst, deptho`:

| key | `exclude_names_and_prefixes` selects | `fill_value_overrides` selected |
|---|---|---|
| `so` | `so_0, so_18` | `so_0, so_18, sos, sob` |
| `thetao` | `thetao_0` | `thetao_0, thetao_bias` |
| `dept` | nothing (would raise) | `deptho` |

`sos` and `sob` are real CMIP names, so `so: 0.0` intending the salinity profile would silently have filled sea-surface and sea-floor salinity too. The `dept` row also falsifies the "a typo fails loudly" guarantee: a truncation typo that happens to be a string prefix of a real channel silently fills that channel instead of raising. Keys now resolve through `NameAndPrefixMatcher`, which makes the documented claim true and drops the duplicated matching logic — the unmatched-key check had a second, separately-written copy of it.

Precedence collapses to "longest key wins" once all keys use one relation, which orders `name_<level>` over `name_` over `name` without needing the exact-name special case.

**2. `fill_value: 0` silently meant the channel mean.** dacite accepts an unsuffixed YAML `0` for `Literal["mean"] | float` (verified: `from_dict(..., {"fill_value": 0})` succeeds), and `isinstance(self.fill_value, float)` was then False, so `build` fell through to the means branch. Confirmed against `main`'s code: `fill_value=0.0` fills 0.0, `fill_value=0` fills the channel mean.

This is pre-existing, but the PR turned it into an inconsistency, which is why I fixed it here rather than leaving it: the new override path does `float(resolved)`, so the *same* `fill_value: 0` meant 0.0 with overrides set and the channel mean without them. Now both branches test `!= "mean"`. No config in the repo passes an integer `fill_value`, so no existing run changes behavior — but this is a behavior change on an existing path, so please confirm you want it in this PR rather than split out.

**3. `test_no_overrides_is_unchanged` asserted nothing.** It compared `StaticSpatialMaskingConfig(mask_value=0, fill_value=fill)` against `StaticSpatialMaskingConfig(mask_value=0, fill_value=fill, fill_value_overrides=None)` — the same dataclass, since `None` is the default (`base == explicit` is `True`). It asserted determinism, not preservation, and would have passed with the default path arbitrarily broken. Replaced with a test that asserts the concrete fill each channel receives, parametrized over both `build` branches.

**4. Removed the `fill_value_overrides` value check from `__post_init__`.** Its test comment said the guard "exists for configs built from YAML by dacite, where the static type is not enforced". That premise is false — dacite rejects `{"thetao_": "zero"}` at load with `UnionMatchError`. So the guard was unreachable from YAML, reachable only from Python where mypy already covers it, and it cost a new `isinstance` plus a `# type: ignore[dict-item]` in the test purely to reach it. A test now pins the dacite rejection instead.

**5. Minor.** Trimmed the research rationale in the `fill_value_overrides` docstring — `docs/api.rst` runs `automodule:: fme.ace` and this config is exported from `fme.ace`, so that paragraph renders into the published API docs; the full argument stays in the description and commit message. Also reworded `# Unchanged from before per-channel overrides existed.`, which described the diff rather than the code and goes stale on merge.

Each of fixes 1–3 has a test that fails without it (verified by running the new tests against 0199def: `..._key_does_not_select_similar_name[so]`, `..._unmatched_key_raises[dept]`, and `..._no_override_selects[None-0]` all fail).

## Audit passes

- ✅ **Silent-failure** — three found and fixed (items 1, 2 above); two remaining items raised as inline comments below.
- ⚠️ **Behavior change on existing paths** — `build`'s no-override branches are otherwise byte-equivalent; the one intended change is item 2, now stated in the description. The `float` fill tensor is now hoisted out of the `defaultdict` factory so all keys share one scalar tensor instead of allocating per key — read-only, no semantic change.
- ⚠️ **Differential tests** — the headline test was vacuous (item 3). Other added tests were sound: `_overrides_setup` gives each channel a distinct mean, so per-channel resolution is actually distinguishable. I added `sos`/`so_0` to the fixture so key-selection breadth is pinned, and parametrized the unmatched-key test over both a transposition and a truncation typo.
- ✅ **Sibling / mirror coverage** — `fme/coupled` reaches this config through the same `StepperConfig.input_masking` field, so it inherits the option with no mirror change (stated in the description). Considered and skipped: `SpatialMaskProvider.build_output_spatial_masker` (fills NaN by construction, no per-channel fill to configure) and the hardcoded `0` fill in `CoupledStepper._get_ocean_forcings` (`fme/coupled/stepper.py:1053-1058`), which is the ocean→atmosphere forcing handoff, not model input masking.
- ✅ **Config back-compat** — new optional field defaulting to `None`, so prior checkpoints load through `from_state`'s `dacite.Config(strict=True)` and every prior YAML still parses. No field removed or restructured, so no deprecation shim needed. Inference-time checkpoint loading unaffected.
- ✅ **Split heuristic** — 2 files, one new config field, no bundled refactor and no back-compat shim. No split warranted.
- ✅ **Abstract-or-final / inheritance depth** — `StaticSpatialMaskingConfig` and `StaticSpatialMasking` have no in-repo subclasses; nothing new inherits.
- ✅ **AI-agent artifacts** — no `Any` returns, no new `isinstance`/`type: ignore` pairs (net −1 `isinstance`, −1 `type: ignore` after item 4), no imports inside test bodies, no abbreviated identifiers. The two stale-comment instances are item 5.
- ✅ **Builder pattern** — `fill_value_overrides` is read only by methods on its own config; `_resolve_fill_values` is private; `build` resolves the mapping and hands `StaticSpatialMasking` a built collaborator rather than the config. `StaticSpatialMaskingConfig` remains a leaf (no dataclass-valued fields). No new violation.

## Verification

`pre-commit run --files fme/core/spatial_masking.py fme/core/test_spatial_masking.py` passes (ruff, ruff-format, mypy). `fme/core/test_spatial_masking.py` 32 passed; with `test_spatial_mask_provider.py` and `test_name_and_prefix_matcher.py`, 64 passed. `fme/ace/test_ocean_train.py` (the integration path that actually builds `input_masking`) passes, as do the `masking`/`normaliz` tests in `fme/ace/stepper/test_single_module.py`.

I also updated the description: it now states the selection relation precisely, records item 2 as a behavior change, and notes the coupled/output-masking sibling decisions. It no longer claims "does not change any default" unqualified, since item 2 does change what an integer `fill_value` does.

fme/core/spatial_masking.py
```diff
@@ -50,45 +51,115 @@ class StaticSpatialMaskingConfig:
...
+        """
+        overrides = self.fill_value_overrides or {}
+        matchers = {key: NameAndPrefixMatcher([key]) for key in overrides}
+        channels = list(names)
+        unselected = sorted(
+            key
+            for key, matcher in matchers.items()
+            if not any(matcher.match(channel) for channel in channels)
+        )
+        if unselected:
```
elynnwu, line 108: Pre-review agent: **Needs a decision — an override key that only selects excluded channels passes this check and then silently does nothing.**

The validation runs over `means`, which is every channel the normalizer knows, not the channels this masker will actually fill. So if a channel is listed in `exclude_names_and_prefixes`, an override key targeting it still "selects a channel" here, no error is raised, and the override has no effect anywhere. Verified:

```python
StaticSpatialMaskingConfig(
    mask_value=0,
    fill_value="mean",
    exclude_names_and_prefixes=["sst"],
    fill_value_overrides={"sst": 0.0},
).build(mask=mask, means={"sst": ...})
# no error; sst is never filled
```

That is a genuine misconfiguration (a fill value set for a channel that is never filled) and it is exactly the class of thing the loud-failure guarantee is advertised to catch, so it undercuts the guarantee at its edge.

The fix is small — resolve over `[n for n in means if not exclude.match(n)]` instead of all of `means`, which is safe because `__call__` only looks up names that pass `_masks(name)`. But it is a semantics choice (raise / warn / allow), and it slightly narrows what the returned mapping contains, so I left it for you rather than applying it.

A related, smaller point on the same check: `means` comes from `load_dict_from_netcdf(..., names=None)`, so it is every variable in the stats file, typically a superset of the model's channels. A key matching a stats-file variable that is not a model channel also passes. Worth a word in the error message ("names or prefixes of normalizer channels") even if you keep the current scope.

fme/core/spatial_masking.py
```diff
@@ -50,45 +51,115 @@ class StaticSpatialMaskingConfig:
...
         if means is None:
             raise ValueError(
-                "fill_values mapping required by build unless configured "
-                "fill_value is a float."
+                "fill_values mapping required by build when fill_value_overrides "
+                "is set, to enumerate the channels and to supply any 'mean' fills."
             )
+        fill_mapping = {
+            name: means[name] if value == "mean" else torch.as_tensor(float(value))
+            for name, value in self._resolve_fill_values(means).items()
```
elynnwu, line 158: Pre-review agent: **Worth confirming — setting any override makes `means` authoritative for the channel list, so a masked channel absent from `means` now raises where it previously took the float fill.**

Building the mapping eagerly over `means` is the right call (a per-channel mapping genuinely cannot be a `defaultdict`), but it changes the failure mode of the `fill_value`-is-a-float path. Before, that path returned a `defaultdict`, so any masked channel got the fill whether or not the normalizer knew it. Now, with overrides set, a masked channel missing from `means` hits `StaticSpatialMasking.__call__`'s `KeyError`:

```
fill_value=0.0, overrides=None        -> not_normalized filled with 0.0
fill_value=0.0, overrides={...}       -> KeyError: ... mapping is missing key 'not_normalized'
```

This is loud rather than silent, and it fires at the first forward pass rather than deep into training, so I did not change it. It is also unlikely in practice, since `means` is loaded with `names=None` and so covers every variable in the stats file. Flagging it because it is the one way enabling this option can break a config that otherwise works, and because the alternative — falling back to `fill_value` for channels `means` does not know — is a one-line `dict.get` and arguably the friendlier contract. Your call; if you keep the raise, it may be worth saying so in the `fill_value_overrides` docstring.

elynnwu:
Pre-review agent: **Question (non-blocking, no change made)** — there is a second per-channel-override pattern already in the repo, and it is worth a sentence saying why this one differs.

`VariableMaskingConfig` (`fme/core/var_masking.py:139`) solves the same shape of problem — a default behavior plus per-channel overrides — with `override_groups`, where each group lists its variables *explicitly* and `__post_init__` enforces that the groups are disjoint. No matching relation, no precedence rule, and overlap is a config error rather than something resolved silently.

`fill_value_overrides` instead uses name-or-prefix selection with a longest-key-wins tie-break. That is defensible — it matches `exclude_names_and_prefixes` in this very config, and writing out 19 levels per variable by hand is worse — but it does mean overlapping keys resolve quietly by a rule the user has to know, where the neighboring pattern rejects overlap outright. Given `exclude_names_and_prefixes` is the closer precedent I'd keep the prefix approach, but a reviewer will likely ask, so it may be worth a line in the description.

If you keep longest-key-wins, one thing that would make it self-evident rather than documented: an equal-length ambiguity is currently impossible only because `max` on distinct keys of equal length would pick one arbitrarily. Two keys of the same length cannot both select one channel under `NameAndPrefixMatcher` today, so there is no live bug — just noting it is an invariant of the matcher rather than something the code checks.

elynnwu:
Pre-review agent: CI note, for the record — the `gpu` job went red on a06a7bf and is now green on re-run. The failure was `fme/core/models/conditional_sfno/test_sfnonet.py::test_block_speed` (1 failed, 3799 passed), a GPU-only wall-clock assertion that grouped DHConv beats ungrouped over `iters=5, warmup=1`:

```
AssertionError: Expected grouped DHConv to be faster than ungrouped, but got
695.230536 ms for grouped and 417.822968 ms for ungrouped.
```

I checked it rather than assuming it was unrelated, since the previous head (0199def) had a green `tests` run. It is runner contention, not a regression:

- The deterministic half of the same test was satisfied — the log's own values show grouped `max_alloc` 4.42 GB < ungrouped 4.75 GB, so only the timing comparison inverted.
- `fme/core/models/` contains no reference to spatial masking; the module is reached only as a transitive package import, and this branch adds no import-time work beyond `from collections.abc import Iterable`. There is no mechanism by which fill-value resolution changes grouped-vs-ungrouped spectral-convolution timing.
- The masking-relevant GPU integration tests passed in the same run: `fme/ace/test_train.py::test_train_and_inference[SFNO-masking]` and `[Samudra]`.
- The re-run passed with no other change.

All checks are now green: `cpu`, `gpu`, `cpu-very-fast-no-healpix`, `build`, `pre-commit`, GitGuardian.

Non-blocking observation for a separate issue, not this PR: `test_block_speed` asserts on wall-clock with 5 iterations and 1 warmup on a shared runner, so it will keep flaking. Its memory assertion is the deterministic one. Worth either a tolerance factor on the timing half or splitting the timing assertion out.

PR #1459, opened 2026-08-25
Add per-channel patch energy score loss

mcgibbon:
Adds an optional per-channel **patch energy score** term to `EnsembleLoss`, a grid-space alternative to the spectral energy score: at each grid point and channel, the fair (2-member) energy score of the `patch_size x patch_size` patch centered on that point, with a periodic east–west boundary and a zero-padded north–south boundary. Where the spectral energy score enforces per-mode calibration globally, the patch energy score scores joint local structure, which is expected to bear on small-scale variability.

The kernel never materializes patches: since patch extraction is linear and the windows of the two fields are aligned, `||patch(u) − patch(v)||²` equals the `patch_size x patch_size` box-sum of the pointwise squared difference, so each patch-distance field costs one pad + sum-pool + sqrt, and no `patch_size**2`-times-larger patch tensor is ever materialized (as an `unfold`-based implementation would require). All three m=2 difference fields (two skill terms, one spread term) run through a single pooling call. Zero padding contributes `(0−0)² = 0` to every distance, so polar patches are effectively truncated to their valid entries. The score is divided by `patch_size` to sit on the CRPS scale; at `patch_size=1` it reduces to fair CRPS up to the numerical floor (the squared distance is floored at `1e-12` before the sqrt, so collapsed ensembles get a zero subgradient — like `abs` in CRPS — instead of NaN). Because the normalization is a fixed `1/patch_size` while polar patches are truncated by the zero padding, rows within the patch halo of a pole carry proportionally less loss weight than interior rows; this mild polar down-weighting is documented as accepted behavior. The kernel assumes a lat–lon `[..., n_lat, n_lon]` layout with periodic longitude, like `FiniteDifferenceCRPSLoss`; HEALPix layouts are out of scope.

Changes:
- `fme.core.ensemble.get_patch_energy_score` (and `_patch_l2_distance`): the kernel, mirroring `get_crps`/`get_energy_score` shape conventions; 2 ensemble members only, odd `patch_size` only, `patch_size` at most `n_lon`
- `fme.core.loss.PatchEnergyScoreLoss`: module wrapper returning a pre-weighted `(B, C, H, W)` `StandardLoss`, so per-channel reduction and logging follow the CRPS path
- `fme.core.loss.EnsembleLoss`: new `patch_energy_score_weight` (default 0.0 — off, no behavior change) and `patch_energy_score_size` (default 3) kwargs, reachable from YAML via `loss.kwargs` like the other ensemble-loss terms

- [x] Tests added

### DIFF
--- a/fme/core/ensemble.py
+++ b/fme/core/ensemble.py
@@ -139,19 +139,34 @@ def get_patch_energy_score(
         E[||X - y||] - 1/2 E[||X - X'||]
 
     The east-west boundary is periodic and the north-south boundary is
-    zero-padded; since generated and target fields receive identical padding,
-    out-of-domain entries contribute zero to every patch distance, so polar
-    patches are effectively truncated to their valid entries.
+    zero-padded. Both fields receive identical padding (implemented as
+    padding of their pointwise squared difference, which is equivalent), so
+    out-of-domain entries contribute zero to every patch distance and polar
+    patches are effectively truncated to their valid entries. Because the
+    score is divided by the fixed factor patch_size regardless of that
+    truncation, rows within halo distance of a pole contribute
+    proportionally less loss than interior rows (roughly by
+    sqrt(n_valid) / patch_size); this mild polar down-weighting is accepted
+    behavior.
 
-    The result is divided by patch_size (the square root of the patch
-    dimensionality) so its magnitude is comparable to CRPS; for
-    patch_size=1 it equals fair CRPS exactly.
+    The division by patch_size (the square root of the patch
+    dimensionality) makes the magnitude comparable to CRPS; at patch_size=1
+    it equals fair CRPS up to the numerical floor below. The squared patch
+    distance is floored at 1e-12 before the square root (a 1e-6 floor on
+    each distance), giving collapsed ensembles a zero subgradient instead
+    of the unbounded gradient of sqrt at zero.
+
+    The horizontal layout is assumed to be [..., n_lat, n_lon] with a
+    periodic last dimension (a lat-lon grid). Other layouts -- e.g. the
+    HEALPix [..., face, ny, nx] -- would be silently mis-padded, as with
+    the finite-difference CRPS loss.
 
     Args:
         gen: The generated ensemble members, of shape
             [n_batch, n_ensemble, ..., n_lat, n_lon].
         target: The target, of shape [n_batch, 1, ..., n_lat, n_lon].
-        patch_size: Odd width of the square patch in grid points.
+        patch_size: Odd width of the square patch in grid points; at most
+            n_lon.
 
     Returns:
         The patch energy score, of shape [n_batch, ..., n_lat, n_lon].
@@ -164,6 +179,11 @@ def get_patch_energy_score(
         )
     if patch_size < 1 or patch_size % 2 == 0:
         raise ValueError(f"patch_size must be a positive odd integer, got {patch_size}")
+    if patch_size > gen.shape[-1]:
+        raise ValueError(
+            f"patch_size ({patch_size}) must not exceed the longitude "
+            f"dimension ({gen.shape[-1]})"
+        )
     n_batch = gen.shape[0]
     # One pad + pool pass over all three difference fields, stacked on batch.
     diffs = torch.cat(
--- a/fme/core/ensemble.py
+++ b/fme/core/ensemble.py
@@ -1,4 +1,5 @@
 import torch
+import torch.nn.functional as F
 
 
 def get_crps(
@@ -89,3 +90,91 @@ def get_energy_score(
     target_term = torch.abs(gen - target).mean(axis=1)
     internal_term = -0.5 * torch.abs(gen[:, 0, ...] - gen[:, 1, ...])
     return target_term + internal_term
+
+
+def _patch_l2_distance(diff: torch.Tensor, patch_size: int) -> torch.Tensor:
+    """
+    Compute the per-point Euclidean distance between the patches of two
+    fields, given their pointwise difference.
+
+    ||patch(u) - patch(v)||^2 at each point equals the patch_size x patch_size
+    box-sum of (u - v)^2, because patch extraction is linear and the windows
+    of the two fields are aligned; no patch tensor is ever materialized.
+
+    Args:
+        diff: The pointwise difference of two fields, of shape
+            [..., n_lat, n_lon].
+        patch_size: Odd width of the square patch in grid points.
+
+    Returns:
+        The patch distance, of the same shape as diff.
+    """
+    halo = patch_size // 2
+    shape = diff.shape
+    squared = (diff * diff).reshape(-1, 1, shape[-2], shape[-1])
+    squared = F.pad(squared, (halo, halo, 0, 0), mode="circular")
+    squared = F.pad(squared, (0, 0, halo, halo), mode="constant", value=0.0)
+    box_sum = F.avg_pool2d(
+        squared, kernel_size=patch_size, stride=1, divisor_override=1
+    )
+    # sqrt has an unbounded gradient at zero (identical members, constant
+    # patches); clamping gives a zero subgradient there, like abs() in CRPS.
+    return torch.sqrt(box_sum.clamp_min(1e-12)).reshape(shape)
+
+
+def get_patch_energy_score(
+    gen: torch.Tensor,
+    target: torch.Tensor,
+    patch_size: int = 3,
+) -> torch.Tensor:
+    """
+    Compute a per-channel patch energy score.
+
+    Each grid point is scored with the energy score of the
+    patch_size x patch_size patch of a single channel centered on it, using
+    the Euclidean norm over patch values:
+
+    .. math::
+
+        E[||X - y||] - 1/2 E[||X - X'||]
+
+    The east-west boundary is periodic and the north-south boundary is
+    zero-padded; since generated and target fields receive identical padding,
+    out-of-domain entries contribute zero to every patch distance, so polar
+    patches are effectively truncated to their valid entries.
+
+    The result is divided by patch_size (the square root of the patch
+    dimensionality) so its magnitude is comparable to CRPS; for
+    patch_size=1 it equals fair CRPS exactly.
+
+    Args:
+        gen: The generated ensemble members, of shape
+            [n_batch, n_ensemble, ..., n_lat, n_lon].
+        target: The target, of shape [n_batch, 1, ..., n_lat, n_lon].
+        patch_size: Odd width of the square patch in grid points.
+
+    Returns:
+        The patch energy score, of shape [n_batch, ..., n_lat, n_lon].
+    """
+    if gen.shape[1] != 2:
+        raise NotImplementedError(
+            "Patch energy score is written here specifically for 2 ensemble "
+            f"members, got {gen.shape[1]} ensemble members. "
+            "Update this function (and its tests) to support more."
+        )
+    if patch_size < 1 or patch_size % 2 == 0:
+        raise ValueError(f"patch_size must be a positive odd integer, got {patch_size}")
+    n_batch = gen.shape[0]
+    # One pad + pool pass over all three difference fields, stacked on batch.
+    diffs = torch.cat(
+        [
+            gen[:, 0] - target[:, 0],
+            gen[:, 1] - target[:, 0],
+            gen[:, 0] - gen[:, 1],
+        ],
+        dim=0,
+    )
+    distance = _patch_l2_distance(diffs, patch_size)
+    target_term = 0.5 * (distance[:n_batch] + distance[n_batch : 2 * n_batch])
+    internal_term = -0.5 * distance[2 * n_batch :]
+    return (target_term + internal_term) / patch_size
--- a/fme/core/loss.py
+++ b/fme/core/loss.py
@@ -660,6 +660,9 @@ class PatchEnergyScoreLoss(torch.nn.Module):
 
     A grid-space alternative to the spectral :class:`EnergyScoreLoss` that
     scores joint local structure per channel instead of per-mode calibration.
+    Assumes a ``[..., n_lat, n_lon]`` layout with periodic longitude (a
+    lat-lon grid); see :func:`fme.core.ensemble.get_patch_energy_score` for
+    the boundary and normalization details.
 
     Returns a pre-weighted ``(B, C, H, W)`` tensor whose mean over trailing
     dims gives the per-``(B, C)`` loss, like :class:`CRPSLoss`.
--- a/fme/core/loss.py
+++ b/fme/core/loss.py
@@ -8,7 +8,7 @@ import torch.linalg
 import torch.nn.functional as F
 
 from fme.core.device import get_device
-from fme.core.ensemble import get_crps, get_energy_score
+from fme.core.ensemble import get_crps, get_energy_score, get_patch_energy_score
 from fme.core.gridded_ops import GriddedOperations
 from fme.core.normalizer import StandardNormalizer
 from fme.core.packer import Packer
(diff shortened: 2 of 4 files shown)


mcgibbon:
Claude: independent pre-review (no authoring context). Ran the diff cold against `main` at f184957be, plus ruff/ruff-format/mypy and the touched test files. Comment-only — I pushed nothing to the branch.

## Audit passes

- **Silent-failure** ⚠️ — see (1) below. Otherwise clean: the kernel raises loudly for 3+ members, even/zero/negative `patch_size`, `patch_size > n_lon`, float `patch_size`, and a `target` missing its singleton ensemble dim (stricter than `get_crps`, which silently broadcasts there). No quiet default branch.
- **Behavior change on existing paths** ✅ — the widened `crps_weight + energy_score_weight + patch_energy_score_weight == 0` check only *widens* what is accepted; no previously-valid config changes. Grepped the repo: nothing matches on the old error string. `PatchEnergyScoreLoss` is always constructed but registers no parameters or buffers — I confirmed `EnsembleLoss.state_dict()` is empty before and after, so no checkpoint-compat break. With the weight at its 0.0 default the component list is unchanged (2 components).
- **Differential tests** ❌ — see (2).
- **Math** ✅ — `0.5(d0+d1) − 0.5·d01` is the unbiased m=2 energy score, matching `get_energy_score`'s convention. The box-sum identity is right, `avg_pool2d(divisor_override=1)` is a true sum, and the output shape is restored exactly (`2·halo == patch_size−1`). Averaging a per-point proper score over points stays proper. The `clamp_min` is safe in both directions: above the clamp the chain-rule gradient `d_k/sqrt(Σd²)` is bounded by 1, below it the clamp gives exactly 0.
- **Test oracle quality** ✅ — `_patch_vectors` is genuinely independent of the implementation (roll + pad + explicit stack vs pad + pool), and the fixture is well chosen: 8×16 so a lat/lon swap fails, 3 channels, batch 2, `patch_size` up to 5 so both polar rows are exercised. The propriety test is not flaky — I re-ran it over seeds 0–4 and the min margin held at 0.0052–0.0059, ~6% of the score.
- **Design property not documented** ⚠️ — see (3).
- **Description accuracy** ⚠️ — see (4).
- **Tooling** ✅ — `pre-commit run --files` (ruff, ruff-format, mypy) passes. `fme/core/test_ensemble.py` 12 passed in 0.28s, `fme/core/test_loss.py` 75 passed in 4.11s, `test_single_module.py -k "crps or ensemble or energy"` 8 passed. Slowest new test 0.07s, so no `medium_duration`/`slow` marker is needed and it stays inside the 3s `--very-fast` budget. No docs or configs enumerate the `EnsembleLoss` kwargs, so nothing to update there.
- **Split** ✅ — one commit, four files, one coherent feature.

## What needs changing

**(1) The kernel hardcodes lat-lon boundary semantics, but the config path is grid-generic.** `LossConfig.build` will construct `EnsembleLoss` from `HEALPixOperations` just as readily as from `LatLonOperations`. On HEALPix (`[..., 12, C, ny, nx]`) the circular pad wraps within a face instead of routing to the neighbouring face, and the zero pad is not polar — the loss runs and returns a finite, wrong number. `EnergyScoreLoss` avoids this by taking `sht` from `gridded_operations`. `FiniteDifferenceCRPSLoss` already carries the same latent assumption via `torch.roll(..., dims=-1)`, so this is precedent rather than a new class of bug, but a one-line "assumes a lat-lon grid" in the docstring (or a guard) is cheap.

Related nit: `patch_size > n_lon` surfaces as `RuntimeError: Padding value causes wrapping around more than once`, which never mentions `patch_size`. One explicit check would make that config error legible.

**(2) `test_patch_energy_score_loss_in_ensemble_loss` cannot detect a broken `patch_energy_score_size` plumbing.** It passes `"patch_energy_score_size": 3`, which is the default — so if `EnsembleLoss.__init__` called `PatchEnergyScoreLoss()` without forwarding the size, the test would still pass. Pass `5` instead (I checked: on the same inputs sizes 3 and 5 give 0.6360 vs 0.6249, so the assertion becomes differential), or assert directly that two sizes give different totals.

Smaller test gaps: nothing covers `patch_energy_score_weight < 0` raising; `get_patch_energy_score` is never tested with `patch_size=0` or negative (only `PatchEnergyScoreLoss` is); and `test_patch_energy_score_finite_gradient_at_zero_spread` is the one new test that omits `torch.manual_seed`.

**(3) Dividing by `patch_size` rather than by sqrt(valid entries) down-weights the polar rows, and this isn't stated anywhere.** Measured on random fields: at `patch_size=3` the two polar rows score 0.55 against ~0.685 in the interior — **81% of interior loss weight**. At `patch_size=5` the outermost rows sit at 77% and the next rows in at 89%, so the deficit reaches `patch_size//2` rows from each pole. Almost all of it is the normalization rather than the truncation itself: `0.5596 · 3/√6 = 0.685`, essentially the interior value. This may well be the choice you want, since area weighting already de-emphasizes the poles — but it deserves a sentence in the docstring, or a per-point sqrt(valid-count) divisor.

**(4) Four description/docstring accuracy points.**

- The PR body and the `get_patch_energy_score` docstring both explain the padding as "generated and target fields receive identical padding" / "Zero padding contributes `(0−0)² = 0`". The code pads the *squared difference*, not the fields. Same result, but the stated mechanism isn't the implemented one, which costs a reader a minute.
- "`sqrt` is clamped at zero" — it is clamped at `1e-12` on the squared box-sum, so the returned value is floored at `1e-6`.
- "at `patch_size=1` it reduces exactly to fair CRPS" — exact on non-degenerate data (I measured max abs difference 0.0), but a fully collapsed point scores `5e-7` where fair CRPS gives `0.0`. Either say "exactly, up to the sqrt clamp" or drop "exactly".
- "~9x less transient memory than an `unfold`-based version at `patch_size=3`, before autograd savings" is a number with no measurement behind it, and it isn't quite right as stated: 9x is the patch-tensor ratio, but the pad and pool intermediates aren't counted in it. Either measure it or soften to "avoids the `patch_size²` memory blowup of `unfold`".

Also worth adding to the description: `patch_energy_score_size` is validated even when the weight is 0, because the module is always constructed. That is the right call — it fails loudly on a typo'd config — but it is unstated.

**Nit.** The `patch_size` validation and its error string are duplicated verbatim in `PatchEnergyScoreLoss.__init__` and `get_patch_energy_score`. ace's AGENTS.md asks for duplication to be consolidated, though a fail-at-construction guard is genuinely more useful than fail-at-first-forward, so this may be worth keeping as is.

## Stepping back

The PR does what it says: the math is correct, the kernel is a genuine efficiency win over `unfold`, and the brute-force oracle is a real independent check rather than a restatement of the implementation. The diff is one commit across four files and is easy to review. Nothing here is a correctness blocker.

**Verdict: not yet ready.** (2) is the one that must be fixed — as written the test cannot see a `patch_energy_score_size` plumbing bug. (3) and the four items in (4) are cheap and should land with it; (1) is a judgment call for the reviewer.

mcgibbon:
Claude: Fixes applied in c69abdb66, responding point by point:

- **Must-fix (test uses default size)**: the wiring test now passes `patch_energy_score_size: 5` with a comment stating it must be non-default so a dropped size-forward fails the test.
- **Polar down-weighting**: documented in the `get_patch_energy_score` docstring as accepted behavior (fixed `1/patch_size` normalization over truncated polar patches, ~sqrt(n_valid)/patch_size of interior weight), and noted in the PR description. Left as designed: the alternative (per-row sqrt(n_valid) normalization) would up-weight pole rows whose grid cells are physically smallest.
- **Docstring/description accuracy**: padding now described as applied to the squared difference (equivalent to padding both fields); the clamp documented as a 1e-12 floor on the squared distance (1e-6 on the distance); the `patch_size=1` CRPS equivalence qualified as exact up to that floor; the unmeasured ~9x memory figure replaced with the structural claim (no patch_size²-times-larger tensor materialized).
- **Lat-lon layout assumption**: documented on both `get_patch_energy_score` and `PatchEnergyScoreLoss` (HEALPix would be silently mis-padded, same caveat as `FiniteDifferenceCRPSLoss`). `patch_size > n_lon` now raises a `ValueError` naming `patch_size` instead of the circular-pad wrap error, with a test.

mcgibbon:
Claude: delta pre-review of `abfb35f05..c69abdb66` (no authoring context, own worktree off `origin/feature/patch-energy-score-loss`). Comment-only — I pushed nothing.

## Prior findings

| # | Finding | Status |
|---|---|---|
| (1) | Lat-lon layout assumption / unhelpful `patch_size > n_lon` error | **Addressed** |
| (2) | Wiring test cannot detect a dropped `patch_energy_score_size` forward | **Unaddressed** — still passes under the bug |
| (3) | Undocumented polar down-weighting | **Addressed** |
| (4) | Four docstring/description accuracy points | **Addressed** (all four) |

**(1)** Documented on both `get_patch_energy_score` and `PatchEnergyScoreLoss`. The new `ValueError` names `patch_size` and the dimension, and the added test (`patch_size=17`, `n_lon=16`) genuinely reaches the new branch — 17 is odd, so it clears the parity check first rather than short-circuiting.

**(3)** The docstring's `sqrt(n_valid) / patch_size` matches what I measured last round: at `patch_size=3` that is `sqrt(6)/3 = 0.816` against a measured 0.81. The rationale in your reply — per-row `sqrt(n_valid)` normalization would up-weight the physically smallest cells — is the right call and I withdraw the alternative.

**(4)** Checked each against the final code: the padding paragraph now matches `_patch_l2_distance` (it pads the squared difference); the clamp is stated as `1e-12` on the squared distance / `1e-6` on the distance, matching `box_sum.clamp_min(1e-12)`; "up to the numerical floor" is right — `get_crps(alpha=1.0)` reduces to `0.5(|d0| + |d1|) − 0.5|d01|`, term-for-term the `patch_size=1` expression; the memory claim is now structural rather than an unmeasured multiple.

## (2) is still open — and my prior comment is the reason

I offered two fixes and the first one does not work. Passing `5` does not make that test differential, because none of its assertions depend on the patch size: `len(components) == 2`, `torch.isfinite(total)`, and `total != crps_only_total` all hold at size 3 and size 5 alike. The only comparator is a CRPS-only loss, so the patch term merely has to be nonzero.

Verified by mutation rather than by reading. Replacing the constructor call in `EnsembleLoss.__init__`:

```python
self.patch_energy_score_loss = PatchEnergyScoreLoss()  # size forward dropped
```

leaves `fme/core/test_ensemble.py fme/core/test_loss.py` at **87 passed** — including the three `-k patch` tests. The comment now added to the test ("the test must fail if EnsembleLoss drops the size forward") asserts a property the test does not have, which is worse than the original: a future reader will trust it.

The fix is the second option — pin the expected value. This passes on `c69abdb66` (`0.5454207062721252` both sides) and fails under the mutation above:

```python
expected = _components_total(
    [
        StandardLoss(0.9 * get_crps(gen, target, alpha=1.0)),
        StandardLoss(0.1 * get_patch_energy_score(gen, target, patch_size=5)),
    ]
)
torch.testing.assert_close(total, expected)
```

It needs `get_crps`/`get_patch_energy_score` imported into `test_loss.py`. Building a second `EnsembleLoss` at size 3 and asserting the two totals differ also works and needs no new imports, but it pins less: it would not catch a wrong weight.

## Nit

The description's Changes bullet still reads "2 ensemble members only, odd `patch_size` only". `c69abdb66` added a third constraint — `patch_size` at most `n_lon` — which the prose paragraph does not mention either.

Three smaller test gaps from last round were not picked up, and I would leave all three: a negative `patch_energy_score_weight` has no test (no `EnsembleLoss` weight validation is tested at all, so this is consistent, not a gap); `get_patch_energy_score` is still untested at `patch_size` 0 or negative, only `PatchEnergyScoreLoss` is; and `test_patch_energy_score_finite_gradient_at_zero_spread` still has no `torch.manual_seed`, which is harmless since it asserts finiteness rather than a value.

## Tooling

`pre-commit run --files` over the four changed files (ruff, ruff-format, large-files, trailing-whitespace, mypy): all pass. `FME_FORCE_CPU=1 pytest fme/core/test_ensemble.py fme/core/test_loss.py`: 87 passed in 3.86s.

**Verdict: not yet ready.** One must-fix, and it is the same one — the wiring test still cannot see a dropped `patch_energy_score_size`, and now carries a comment claiming otherwise. Everything else from the prior round landed cleanly.

mcgibbon:
Claude: Fixed in 97e67d443, using the pinned-value approach from your comment: the wiring test now asserts the built loss's total equals `0.9 * get_crps(gen, target, alpha=1.0).mean() + 0.1 * get_patch_energy_score(gen, target, patch_size=5).mean()` via `torch.testing.assert_close`, with the comment corrected to describe what the test actually checks. Verified by the same mutation you ran (constructor call replaced with `PatchEnergyScoreLoss()`): 1 failed under the mutation, passes restored. The description's Changes bullet now includes the `patch_size` ≤ `n_lon` constraint.

mcgibbon:
Claude: delta pre-review of `c69abdb66..97e67d443` (own worktree off `origin/feature/patch-energy-score-loss`; comment-only, nothing pushed).

**The must-fix is genuinely closed.** Confirmed with my own mutation run, not by reading the diff.

Mutation A — drop the size forward (`self.patch_energy_score_loss = PatchEnergyScoreLoss()`): `1 failed, 86 passed`, failing exactly at `test_patch_energy_score_loss_in_ensemble_loss` with `AssertionError: Scalars are not close!`. Restored: `87 passed`. That reproduces your result.

Mutation B — mine, to check the pin is not narrowly aimed at the size. Drop the weight application (`components.append(type(c)(c.loss))`): also fails at the same test. So the pinned value now covers both `patch_energy_score_size` and `patch_energy_score_weight` plumbing, which the old `!= crps_only_total` comparator covered for neither.

The replacement is a strict improvement on what it removed: `assert_close` against a finite expected value subsumes the dropped `torch.isfinite(total)`, and the dropped CRPS-only comparator was the weaker check. It also pins `alpha=1.0`, matching `EnsembleLoss`'s `almost_fair_crps_alpha` default, so a change to that default surfaces here too. The comment now describes the mechanism that does the work rather than a property the test lacked.

Description nit resolved — the Changes bullet reads "2 ensemble members only, odd `patch_size` only, `patch_size` at most `n_lon`".

Re-verified on `97e67d443`: `pre-commit run --files` over the four changed files (ruff, ruff-format, large-files, trailing-whitespace, mypy) passes; `FME_FORCE_CPU=1 pytest fme/core/test_ensemble.py fme/core/test_loss.py` gives 87 passed in 3.66s. Everything from rounds 1 and 2 remains addressed.

elynnwu:
Claude: one ocean-specific consideration, raised as a question rather than a defect — for the atmosphere I don't think it matters at all.

We're planning to use this for a stochastic ocean emulator (Samudra), and I worked through how land interacts with the kernel. The good news is that it needs no masking code: land carries identical fill in target and prediction (or is NaN-zeroed upstream in `loss.py`), so the pointwise difference is zero there and land contributes zero to the box-sum — exactly as your north–south zero-padding does. That's a nice property and I don't think it needs changing.

The question is the normalization. Since the score is divided by a fixed `patch_size` regardless of how many patch entries are valid, a patch straddling a coastline carries proportionally less loss weight — roughly `sqrt(n_valid) / patch_size`, the same mild down-weighting you document for polar rows. For the atmosphere that only affects two thin polar bands. For the ocean it affects every coastline, which is a large fraction of the domain.

The reason I'm flagging it rather than just accepting it: we've just measured where the deterministic ocean loses variance, and it is worst exactly in that band. Interannual SST variance ratio (emulator/reference) by distance from land, over a 200-year piControl rollout:

| distance from land | ratio |
|---|---|
| 1–2 cells | 0.453 |
| 2–4 | 0.548 |
| 4–8 | 0.588 |
| 8–16 | 0.602 |
| 16–32 | 0.669 |
| >32 (open ocean) | 0.832 |

Monotone across six bins, and the western boundary currents are the most damped regions in the map (Kuroshio 0.482, Gulf Stream 0.488). So the coastal band is where we most need the loss to enforce calibration, and it's the band the fixed normalization down-weights.

Two options as I see them, and I don't have a strong view on which is right for a shared implementation:

1. Leave it as-is and treat the coastal down-weighting as accepted, consistent with the polar behavior.
2. Normalize by the valid entry count rather than `patch_size` (DLESyM-Ocean uses a Horvitz–Thompson-style reweight and drops fully-terrestrial patches). This would change atmosphere behavior at the poles too, so it probably wants to be opt-in rather than the default.

Happy to do the work on option 2 as a follow-up behind a flag if you think it's worth having — I don't want to push ocean concerns into a PR that's aimed at the atmosphere. Either way, thanks for building this: the grid-space form is what makes it usable for the ocean at all, since a spherical-harmonic energy score over a land-masked field puts spurious power at high degree exactly where the whitening reweight amplifies it.

mcgibbon:
Ah FWIW this PR isn't ready for review, I'd want some experimental validation that it helps before we merge. Granted the bar is a bit lower given how obvious the approach is, the isolation of the code, and third party results.

However I'm also not particularly worried about coastlines having less weight in an ocean context - coastlines will have less weight but also they have less valid area. The main "bug" I would maybe argue for in that context is that the loss is technically "valid" over the masked land area right next to the coastline, where the receptive field of the loss includes ocean cells. This issue already applies to the existing finite-difference-crps loss.

One could perhaps argue that for the ocean, we should be applying and summing the loss over the entire domain, and then taking the mean over the number of valid output cells, rather than taking a masked mean over valid output cells. That way deep interior cells that have no reliance on valid data will have zero loss by virtue of being exactly prescribed, while affected near-coast interior cells will contain reliable loss signal. In that case, each cell (coastline or not) would be included the same amount in the loss, once for each nearby cell that contains it in the loss's receptive field.

Another tact to take is to say we should see if this is an issue and address it if it is - we could run training with both versions and see if coastlines improve.

PR #1462, opened 2026-08-26
Let inference netCDF writers target a remote experiment_dir

elynnwu:
Inference netCDF output could only be written to a local filesystem: `RawDataWriter` and `MonthlyDataWriter` stream into an open `netCDF4.Dataset` handle, which cannot target a remote store, so both raised `ValueError` when `experiment_dir` was not local. That forced `save_prediction_files`/`save_monthly_files` runs to deposit their (often very large) output in the job's result dataset even when the intended consumer reads from GCS.

These writers now accept a remote `experiment_dir`. Each file is streamed to a local temporary directory exactly as before and copied to its destination when the writer is finalized and the handle is closed, so only complete files ever appear at the destination. Local destinations are unchanged: no staging directory is created and the file is written in place.

Two consequences are documented on `InferenceConfig.experiment_dir` and `InferenceEvaluatorConfig.experiment_dir`: the full output occupies local disk for the duration of the run (in the system temp directory, so `TMPDIR` must have room for it), and a run that dies before finalizing leaves nothing at the remote destination. The zarr writer remains the incremental, crash-durable option; the netCDF writers are the option for monthly means, which zarr does not support.

Changes:
- `fme.core.cloud.StagedFile`: new helper owning the local-staging/upload-on-close policy. `path` is where to write; `upload()` copies to the destination with the destination filesystem's fsspec `put_file` and discards the staging copy, and is a no-op for a local destination or on a repeat call. `put_file` rather than `inter_filesystem_copy` because these files are multi-GB: gcsfs's `put_file` uploads in 50 MiB chunks, against a ~5 MiB write block for a stream through `fsspec.open`, and it can verify the uploaded object's checksum, which the stream path does not. The upload passes `consistency="md5"` explicitly, because gcsfs defaults to `consistency="none"`, whose checker is a no-op — without the kwarg `put_file` would gain the chunking and no integrity check at all. md5 rather than crc32c because crc32c requires `crcmod`, which is not a dependency. The checksum request is gcsfs-specific: other fsspec backends ignore the kwarg (s3fs filters unrecognized kwargs), so a non-GCS remote destination uploads with whatever integrity check that backend does by default. `inter_filesystem_copy` itself is unchanged and keeps its small-file callers.
- `fme.ace.inference.data_writer.raw.RawDataWriter`, `fme.ace.inference.data_writer.monthly.MonthlyDataWriter`: replace the `is_local` rejection with a `StagedFile`, uploaded in `finalize`. This also lifts the restriction for `FileWriterConfig` entries using the netCDF format, including `MonthlyCoarsenConfig`.
- `fme.ace.inference.data_writer.main`: create the `step_diagnostics/` subdirectory with `fme.core.cloud.makedirs` rather than `os.makedirs`, which would have created a literal `gs:/...` local directory.
- Tests: `StagedFile` local and remote round-trips, that the upload dispatches to the destination filesystem's `put_file` with `consistency="md5"`, that `md5` is a checker gcsfs actually supports, and that a failed upload raises, keeps the staged file, and can be retried; netCDF output to a `memory://` experiment dir is byte-identical to the local output and appears only at finalize; `test_evaluator_with_non_local_experiment_dir` now enables the netCDF writers and asserts the four `.nc` files land remotely; the file-writer test that asserted rejection of a non-local dir now asserts the file is written.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/data_writer/main.py
+++ b/fme/ace/inference/data_writer/main.py
@@ -15,7 +15,7 @@ from fme.ace.data_loading.batch_data import (
     PairedData,
     PrognosticState,
 )
-from fme.core.cloud import to_netcdf_via_inter_filesystem_copy
+from fme.core.cloud import makedirs, to_netcdf_via_inter_filesystem_copy
 from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.generics.writer import WriterABC
 
@@ -175,7 +175,7 @@ class DataWriterConfig:
 
         def writer_factory(name: str) -> RawDataWriter | TimeCoarsen:
             path = os.path.join(experiment_dir, STEP_DIAGNOSTICS_DIR)
-            os.makedirs(path, exist_ok=True)
+            makedirs(path, exist_ok=True)
             writer: RawDataWriter | TimeCoarsen = RawDataWriter(
                 path=path,
                 label=name,
--- a/fme/ace/inference/data_writer/monthly.py
+++ b/fme/ace/inference/data_writer/monthly.py
@@ -1,7 +1,7 @@
 import copy
 import datetime
+import os
 from collections.abc import Iterable, Mapping, Sequence
-from pathlib import Path
 
 import cftime
 import numpy as np
@@ -17,7 +17,7 @@ from fme.ace.inference.data_writer.utils import (
     DIM_INFO_LATLON,
     get_all_names,
 )
-from fme.core.cloud import is_local
+from fme.core.cloud import StagedFile
 from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.writer import DATETIME_ENCODING_UNITS
 
@@ -105,7 +105,9 @@ class MonthlyDataWriter:
     ):
         """
         Args:
-            path: Directory to write netCDF file(s).
+            path: Directory to write netCDF file(s). May be remote (e.g.
+                ``gs://``), in which case the file is written to a local
+                staging directory and uploaded when the writer is finalized.
             label: Label to append to the filename.
             initial_condition_times: 1D array of initial condition times
                 (start time for each inference run).
@@ -116,9 +118,10 @@ class MonthlyDataWriter:
             coords: Coordinate data to be written to the file.
             dataset_metadata: Metadata for the dataset.
         """
-        if not is_local(path):
-            raise ValueError("MonthlyDataWriter only supports local file systems.")
-        filename = str(Path(path) / f"{label}.nc")
+        # netCDF4 can only stream into a local file handle, so a remote
+        # destination is staged locally and uploaded by ``finalize``.
+        self._staged_file = StagedFile(os.path.join(str(path), f"{label}.nc"))
+        filename = self._staged_file.path
         n_initial_conditions = len(initial_condition_times)
         calendar = infer_calendar(initial_condition_times)
         self._save_names = save_names
@@ -334,6 +337,7 @@ class MonthlyDataWriter:
     def finalize(self):
         self.flush()
         self.dataset.close()
+        self._staged_file.upload()
 
 
 def add_data(
--- a/fme/ace/inference/data_writer/monthly.py
+++ b/fme/ace/inference/data_writer/monthly.py
@@ -120,7 +120,7 @@ class MonthlyDataWriter:
         """
         # netCDF4 can only stream into a local file handle, so a remote
         # destination is staged locally and uploaded by ``finalize``.
-        self._staged_file = StagedFile(os.path.join(str(path), f"{label}.nc"))
+        self._staged_file = StagedFile(os.path.join(path, f"{label}.nc"))
         filename = self._staged_file.path
         n_initial_conditions = len(initial_condition_times)
         calendar = infer_calendar(initial_condition_times)
--- a/fme/ace/inference/data_writer/raw.py
+++ b/fme/ace/inference/data_writer/raw.py
@@ -124,7 +124,7 @@ class RawDataWriter:
         """
         # netCDF4 can only stream into a local file handle, so a remote
         # destination is staged locally and uploaded by ``finalize``.
-        self._staged_file = StagedFile(os.path.join(str(path), f"{label}.nc"))
+        self._staged_file = StagedFile(os.path.join(path, f"{label}.nc"))
         filename = self._staged_file.path
         calendar = infer_calendar(initial_condition_times)
         n_initial_conditions = len(initial_condition_times)
--- a/fme/ace/inference/data_writer/raw.py
+++ b/fme/ace/inference/data_writer/raw.py
@@ -1,8 +1,8 @@
 import copy
 import dataclasses
 import datetime
+import os
 from collections.abc import Iterable, Mapping, Sequence
-from pathlib import Path
 from typing import Literal
 
 import cftime
@@ -18,7 +18,7 @@ from fme.ace.inference.data_writer.utils import (
     DIM_INFO_LATLON,
     get_all_names,
 )
-from fme.core.cloud import is_local
+from fme.core.cloud import StagedFile
 from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.writer import DATETIME_ENCODING_UNITS, TIMEDELTA_ENCODING_UNITS
 
@@ -110,7 +110,9 @@ class RawDataWriter:
     ):
         """
         Args:
-            path: Directory within which to write the file.
+            path: Directory within which to write the file. May be remote
+                (e.g. ``gs://``), in which case the file is written to a local
+                staging directory and uploaded when the writer is finalized.
             label: Name of the file to write.
             initial_condition_times: 1D array of initial condition times
                 (start time for each inference run).
@@ -120,13 +122,10 @@ class RawDataWriter:
             coords: Coordinate data to be written to the file.
             dataset_metadata: Metadata for the dataset.
         """
-        if not is_local(path):
-            raise ValueError(
-                "The RawDataWriter only supports local file systems. Consider "
-                "using the ZarrWriter instead, which supports writing to a "
-                "non-local filesystem."
-            )
-        filename = str(Path(path) / f"{label}.nc")
+        # netCDF4 can only stream into a local file handle, so a remote
+        # destination is staged locally and uploaded by ``finalize``.
+        self._staged_file = StagedFile(os.path.join(str(path), f"{label}.nc"))
+        filename = self._staged_file.path
         calendar = infer_calendar(initial_condition_times)
         n_initial_conditions = len(initial_condition_times)
         self._save_names = save_names
@@ -261,6 +260,7 @@ class RawDataWriter:
     def finalize(self):
         self.flush()
         self.dataset.close()
+        self._staged_file.upload()
 
 
 def get_batch_lead_time_microseconds(
--- a/fme/ace/inference/data_writer/test_data_writer.py
+++ b/fme/ace/inference/data_writer/test_data_writer.py
@@ -1,6 +1,8 @@
 import datetime
+import os
 
 import cftime
+import fsspec
 import numpy as np
 import numpy.typing as npt
 import pytest
@@ -16,6 +18,7 @@ from fme.ace.inference.data_writer.main import DataWriterConfig
 from fme.ace.inference.data_writer.raw import get_batch_lead_time_microseconds
 from fme.ace.inference.data_writer.time_coarsen import TimeCoarsenConfig
 from fme.ace.inference.data_writer.zarr import ZarrWriterConfig
+from fme.core.cloud import open_dataset_via_inter_filesystem_copy
 from fme.core.dataset.data_typing import VariableMetadata
 from fme.core.device import get_device
 from fme.core.labels import BatchLabels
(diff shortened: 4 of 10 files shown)


elynnwu:
Pre-review agent (Claude): I ran an independent pre-review of this PR without the authoring session's context. Full audit-pass results are below; three small fixes are already applied and pushed as 5e2b04a4d.

**Applied in 5e2b04a4d**
- Removed `StagedFile.destination`. It is never read anywhere in the repo or in tests — unused public API on a brand-new class.
- `os.path.join(str(path), f"{label}.nc")` → `os.path.join(path, ...)` in `raw.py` and `monthly.py`. `path` is already annotated `str` on both writers; the `str()` was a leftover from the `Path(path) /` form being replaced.
- Strengthened `test_netcdf_writers_upload_only_on_finalize`. As written it asserted only an absence (`fs.ls(experiment_dir) == []` after `flush()`), so it would also pass if the writers were inert or never wrote anything at all. It now calls `finalize()` at the end and asserts the four `.nc` files do land, which makes the earlier assertion about the *timing* of the upload rather than about nothing happening.

**Audit passes**

✅ **Silent-failure.** The removed `ValueError` did not guard a silent path — the only behavior change is that a previously-loud rejection now succeeds. I checked the disk-exhaustion and upload-failure cases specifically: a full `TMPDIR` raises out of the netCDF4 write or `dataset.close()`, and an upload failure raises out of `inter_filesystem_copy`; neither is caught anywhere between the writer and `run_inference_from_config`, so nothing is swallowed and no partial file is presented as complete. `StagedFile.upload()` is the only thing that puts a file at the remote destination, and it runs strictly after `dataset.close()`, so the remote object is either absent or whole.

✅ **Behavior change on existing paths.** Local destinations are genuinely untouched: `StagedFile` short-circuits to `path is destination` with no temp dir and a no-op `upload()`. `os.path.join` and `Path(...) / ...` agree for every local form including a trailing slash. `cloud.makedirs` on a local path is `LocalFileSystem.makedirs`, equivalent to `os.makedirs`, so the `main.py` swap is a no-op locally and the fix it makes remotely is real (`os.makedirs` would have created a literal `gs:/...` directory).

✅ **Differential test.** I no-oped the copy inside `StagedFile.upload()` and re-ran: 5 tests fail (`test_staged_file_remote_uploads_on_upload`, `test_netcdf_writers_with_non_local_experiment_dir`, and all three `test_netcdf_file_writer_with_non_local_experiment_dir` parametrizations). The inverted `test_netcdf_file_writer_with_non_local_experiment_dir` is a correct inversion, not a weakened one — it asserts the file exists at the remote destination, and it does fail when the upload is removed. Its one limitation is that it never appends a batch, so it exercises only the empty-file path; the byte-identity check in `test_netcdf_writers_with_non_local_experiment_dir` covers the populated case, so this is a reasonable division of labor.

✅ **Sibling/mirror coverage.** I swept every remaining local-filesystem assumption on an `experiment_dir`-derived path in the inference stack. `prepare_directory` already uses `cloud.makedirs` + `fsspec.open`; the aggregator diagnostics and `PairedDataWriter.write` already use `to_netcdf_via_inter_filesystem_copy`; `step_diagnostics/` was the only remaining one and this PR fixes it. `fme/coupled` reuses `DataWriterConfig` directly, so it inherits the change and makes no contradicting claim in its own `experiment_dir` docstring. The only `os.makedirs` left on an `experiment_dir` is `fme/coupled/train/train.py`, which is training, not inference, and out of scope here. No stale "only supports local" text remains anywhere in the repo or docs. Finalize propagation is complete: `TimeCoarsen`, `FileWriter`, `PairedFileWriter`, and `StepDiagnosticsWriter` all forward `finalize()` to the wrapped writer, so the lazily-constructed step-diagnostics writers upload too.

✅ **Config back-compat.** No config field added, removed, or restructured; no checkpoint-persisted state touched. Every previously-valid YAML parses identically, and a config that previously errored now works — the permissive direction.

✅ **Split heuristic.** 10 files, one feature, no refactor bundled. Correctly a single PR.

✅ **Abstract-or-final / inheritance depth.** `StagedFile` has no in-repo subclasses, so neither rule applies.

✅ **Builder pattern.** `StagedFile` is a plain internal collaborator, not a dacite-loaded config, and no config reads another config's fields.

✅ **AI-agent artifacts.** The two found (dead `destination` property, redundant `str()`) are fixed above. Test imports are at module top, no `Any` returns, no `isinstance` + `type: ignore`, no abbreviations, no stray comments left from the refactor.

Tests: `fme/ace/inference/`, `fme/core/test_cloud.py`, and `fme/coupled/inference/` all pass (215 passed) at 5e2b04a4d, and `pre-commit` (ruff, ruff-format, mypy) is clean on every changed file.

elynnwu:
Pre-review agent (Claude): three things I deliberately did **not** change, flagged for the reviewer's attention. None of them blocks review in my judgement; the first is the one I would most want a human decision on.

**1. `inter_filesystem_copy`'s "Do not use for large files" caveat now contradicts its main new caller.** `StagedFile.upload()` delegates to `inter_filesystem_copy`, whose docstring says exactly the opposite of what this feature is for — the whole point is uploading multi-GB `save_prediction_files` / `save_monthly_files` output. I believe the caveat is stale rather than correct: `inter_filesystem_copy` uses `shutil.copyfileobj`, which streams in 64 KiB chunks, and fsspec's remote write side buffers by block and does a resumable upload, so memory stays bounded regardless of file size. But I did not want to silently rewrite a shared helper's documented contract on the strength of my own reading. Either the caveat should be narrowed to what it actually means, or `StagedFile` should say why it is exempt — otherwise the next reader hits the same contradiction.

**2. One failed upload in `finalize()` aborts the remaining writers' uploads.** `PairedDataWriter.finalize()` loops over sub-writers calling `finalize()` in turn. Before this PR, `finalize()` did local work that essentially never failed; now each one ends in a network upload, so a transient GCS error on `autoregressive_predictions.nc` means `autoregressive_target.nc` and both monthly files never upload either, even though they are complete on local disk and the temp dirs are about to be discarded. The loop structure is pre-existing, so I left it, but the failure mode is new. If it is worth hardening, the cheap version is to attempt every sub-writer's `finalize()` and re-raise afterwards.

**3. Staged temp files survive a hard kill.** `StagedFile` holds a `tempfile.TemporaryDirectory`, which cleans up on GC or interpreter exit — so a normal exception path does release the disk. A `SIGKILL` (OOM, preemption) leaves the staged copy behind. In a job container that is harmless; noting it only because the same run is now the one most likely to be near its local disk limit.

Separately, on the two consequences the description already documents (whole output resident in `TMPDIR`, nothing at the destination if the run dies before finalizing): I confirmed both are accurate, and that "nothing" rather than "a truncated file" is what a dead run leaves remotely. The `TMPDIR` sizing requirement is real and worth a reader's attention — an inference job writing raw prediction files at high frequency can want hundreds of GB of scratch, and the default container `TMPDIR` will not have it.

elynnwu:
Pre-review agent (Claude): **Verdict — ready for review** (at 5e2b04a4d, i.e. after the fixes I applied above).

The PR is well-scoped: one feature, no bundled refactor, and the staging policy is isolated in a single small class rather than smeared across the two writers. All eight applicable audit passes came back clean, the tests are differentially meaningful, and local behavior is genuinely unchanged. No structural rework or split is warranted.

Suggested reviewer focus, in order: (1) whether `inter_filesystem_copy`'s "do not use for large files" caveat is stale or real, since this feature's whole purpose is large files; (2) the accept/reject call on the two documented consequences — full output resident in `TMPDIR`, and nothing at the destination if the run dies before `finalize()` — which are inherent to netCDF4 not being streamable to object storage, not incidental to this implementation.

elynnwu:
Pre-review agent: second pass, focused on `5f88ba12f` (`inter_filesystem_copy` → `put_file`) and `a92ed2727` (`consistency="md5"`). The base change was pre-reviewed clean at `5e2b04a4d`; this pass re-checked it only for interactions with the new upload mechanism. Fixes applied in `75ad04082`.

## Audit passes

✅ **1. Silent failure.** The new-branch surface is small (`put_file` vs. nothing), but one silent path is real and is the main finding, below: `consistency=` is gcsfs-specific and is silently *dropped* by s3fs. Two related no-op-default risks check out — gcsfs's `_put_file` does `consistency = consistency or self.consistency`, so the per-call kwarg wins regardless of how the cached `url_to_fs` instance was constructed, and `get_consistency_checker` answers an unrecognized value with the no-op base checker rather than raising, which is exactly the failure the md5 test now covers (see pass 3).

⚠️ **2. Behavior change on existing paths.** `StagedFile.upload` is new in this PR, so no existing caller observes the switch. Two semantic differences from `inter_filesystem_copy` worth naming, both benign: `put_file` defaults to `mode="overwrite"` (same as the old `fsspec.open(dest, "wb")`), and for a sync backend `AbstractFileSystem.put_file` `mkdirs` the parent, which the old stream did not — hence the nested-destination case in `test_staged_file_remote_uploads_on_upload`. `inter_filesystem_copy` itself is untouched and keeps its small-file callers.

⚠️ **3. Differential tests.** Mutation-checked by hand, in a scratch worktree:
- dropping `consistency="md5"` → `test_staged_file_upload_uses_destination_put_file` fails (kwarg dict mismatch). Reverting the whole line to `inter_filesystem_copy` → that test *and* `test_staged_file_upload_error_propagates` fail. Both mutations are caught.
- `test_md5_is_a_supported_gcsfs_consistency_checker` did **not** hold up. `assert checker is not None` cannot fail: if gcsfs dropped `"md5"`, `get_consistency_checker` returns the no-op `ConsistencyChecker`, not `None`. Only the second assertion carried any weight. Rewritten to feed the checker bytes and assert it accepts the correct `md5Hash` and raises `ChecksumError` on a wrong one; verified that with `"md5"` replaced by an unrecognized value the rewritten test fails. Also lifted the import to module level — gcsfs is a hard dependency in `pyproject.toml`, so `pytest.importorskip` could only ever mask a broken install.
- `test_staged_file_upload_error_propagates` now also asserts the retry after the failed upload succeeds, which is what makes the "staged file survives" claim worth something, and cleans up its `memory://` root.

✅ **4. Sibling coverage.** `StagedFile` has exactly two callers (`RawDataWriter`, `MonthlyDataWriter`), both converted; `grep` finds no other `is_local` rejection in a writer and no `fme/coupled` mirror of these two.

✅ **5. Config back-compat.** No loaded-config or checkpoint-persisted state touched by these two commits.

✅ **6. Split.** 10 files, one feature, no config type. Do not split.

✅ **7. Abstract-or-final / inheritance depth.** `StagedFile` has no subclasses.

⚠️ **8. AI-agent artifacts.** Import inside a test body and two unused `tmp_path` fixtures; both fixed. Nothing else.

✅ **9. Builder pattern.** No config dataclass in these commits.

## Findings and what I changed

**The one thing worth a reviewer's attention (not fixed, needs your call):** `consistency="md5"` is a gcsfs kwarg. For an `s3://` destination, s3fs's `_get_s3_method_kwargs` filters kwargs down to what the boto method accepts, so `consistency` is **silently discarded** — no error, no checksum. I verified this against the installed s3fs 2025.12.0 (a `put_file` with the kwarg reaches the network and fails only on the nonexistent bucket, i.e. param validation passed). s3fs is a hard dependency and `s3://` paths are used elsewhere in the repo, so an `experiment_dir: s3://...` is reachable. I documented the gcs-specificity in the `StagedFile` docstring and the PR description rather than adding a backend branch, because (a) behavior is unchanged either way and (b) whether an s3 destination *should* get an explicit integrity check is a design call, not a pre-review call. If you want it, the shape is a `isinstance(fs, GCSFileSystem)` guard plus a warning on the else — say the word.

**Applied in `75ad04082`:**
1. `StagedFile` docstring: the checksum request is gcsfs-specific; other backends ignore it.
2. `upload` docstring: "the staging copy is kept if it fails" overstated durability. The staged file does survive the exception (so a caller that catches it can retry — a genuinely useful property, now tested), but the staging directory is a `tempfile.TemporaryDirectory` whose finalizer runs at interpreter exit, so an *uncaught* failure — which is what happens today, since `finalize` propagates out of the inference run — loses it. Reworded to say both halves.
3. `test_md5_is_a_supported_gcsfs_consistency_checker` rewritten as described in pass 3.
4. `test_staged_file_upload_error_propagates`: retry assertion + cleanup.
5. Unused `tmp_path` fixtures dropped.
6. PR description: added the gcsfs-specificity sentence, and softened "which a stream cannot" — a `GCSFile` stream also accepts `consistency`; the point is the stream path here does not use it.

**Checked and clean, for the record:** the `>= 5 MiB` resumable path in `GCSFileSystem._put_file` hashes every uploaded chunk and validates against `md5Hash` from the final chunk response via `MD5Checker.validate_json_response`; the "GCS does not provide md5 sums for composite objects" `NotImplementedError` is only in `validate_headers`, which is not on this path. The `< 5 MiB` path validates through `simple_upload` the same way. `md5` has no optional dependency (unlike `crc32c`/`crcmod`). Docstring numbers are accurate: `chunksize=50 * 2**20` and `default_block_size == 5242880`.

**Not fixed, noted:** if the checksum itself fails, gcsfs raises *after* the last chunk without deleting the object, so a `ChecksumError` can leave a corrupt object at the destination while the run aborts. That is gcsfs behavior, not this PR's, and detecting-and-aborting is strictly better than the status quo of not checking at all.

Tests: `fme/core/test_cloud.py` (17 passed), `fme/ace/inference/data_writer` (117 passed), `test_evaluator_with_non_local_experiment_dir` (passed); ruff, ruff-format, mypy hooks pass on both changed files.

## Verdict

**Ready for review.** The mechanism swap is correct and the reasoning for it is on the record in both the code and the description. One judgement call is left for you: whether a non-GCS remote destination should get an explicit integrity check rather than a documented "whatever the backend does".

PR #1463, opened 2026-08-26
Add NoiseConditionedSamudra with two noise-injection variants

elynnwu:
Makes Samudra noise-conditionable so it can be trained with an ensemble proper scoring rule (`EnsembleLoss`), reusing the generic `NoiseConditionedModel` wrapper rather than adding a second noise path.

Motivation: the deterministic `ocean_wide_ohc_clip` Samudra runs are measurably variance-damped over long piControl rollouts, worst near coasts and in western boundary currents. Stochastic training against a proper scoring rule is the candidate fix; this is the module-side prerequisite.

## Changes

**`ConvNeXtBlock` becomes context-aware.** It takes an optional `ContextConfig`; when its noise embedding is non-zero, each normalization layer is followed by a `NoiseConditioning` module that reads the `Context`'s noise field and applies `x -> x * (1 + W_scale(noise)) + W_bias(noise)`, the same construction `ConditionalLayerNorm` uses for the SFNO. Samudra's default `instance` norm is built with `affine=False`, so this is exactly a conditional instance norm. Conditioning therefore requires a norm to condition (`norm=None` is rejected).

The noise field is drawn at the model's input resolution, while a conditioned block may run coarser inside the U-Net, so the conditioning area-averages the noise onto the block's grid.

`convblock` changes from `Sequential` to `ModuleList` so the forward can interleave conditioning. Layer order and state-dict keys are unchanged, and the conditioning modules live in a separate `ModuleDict` that is empty when conditioning is off, so checkpoints trained before this change still load. The duplicated four-branch norm construction is factored into `_build_norm`.

**`Samudra`** takes `context_config` and `noise_injection` and threads the `Context` through to its ConvNeXt blocks (pooling, upsample, and the final conv still take the tensor alone).

**`NoiseConditionedSamudra` registry entry**, with both injection variants:

- `bottleneck` (default): only the block at the coarsest resolution, DLESyM-Ocean's choice with 32-dim noise. After four `AvgPool`s that sits at 1/16 resolution, so the noise perturbs large scales.
- `all_blocks`: every block, the pattern the ACE SFNO uses, which also reaches the finest scales.

Both are wanted rather than one: the measured variance deficit is worst at coastal and western-boundary-current scales, where bottleneck-only conditioning may push at the wrong scales.

**Zero-init.** The conditioning convolutions are zero-initialized, as the SFNO path does, so an untrained model is exactly deterministic and stochasticity is learned from the ensemble loss.

## Tests

- Both variants condition the expected blocks, and at init two different noise draws give bit-identical output that also matches the unconditioned model loaded with the same parameters.
- With the conditioning weights perturbed, noise changes the output, gradients reach the conditioning weights, and `checkpoint_strategy="all"` still works.
- Identical inputs stacked along the batch dimension come out as distinct members, which is what ensemble folding relies on.
- An unconditioned block's state-dict key set is unchanged.
- The ocean corrector (`force_positive`, sea-ice fraction, OHC budget) is verified to act per-member under ensemble batch folding: correcting a folded 2-member batch equals correcting each member alone. A corrector that coupled across the batch dim would silently collapse the spread the scoring rule is meant to reward. Mutation-checked — making the area-weighted mean also reduce over the batch dim fails the test.

Independent of the patch energy score (#1459); the two compose but neither blocks the other.

Sizing note for whoever trains with this: `n_ensemble=2` doubles the effective batch, so the `ocean_wide_ohc_clip` recipe (4 GPUs, 500 GiB shared memory at `n_forward_steps=8`) needs a halved batch or more memory.

### DIFF
--- a/docs/modules.rst
+++ b/docs/modules.rst
@@ -54,3 +54,9 @@ The following module builders are available:
    :undoc-members:
    :show-inheritance:
    :noindex:
+
+.. autoclass:: fme.ace.NoiseConditionedSamudraBuilder
+   :members:
+   :undoc-members:
+   :show-inheritance:
+   :noindex:
--- a/fme/ace/__init__.py
+++ b/fme/ace/__init__.py
@@ -78,7 +78,11 @@ from fme.ace.registry.hpx import (
     UNetEncoderConfig,
 )
 from fme.ace.registry.land_net import LandNetBuilder
-from fme.ace.registry.m2lines import FloeNetBuilder, SamudraBuilder
+from fme.ace.registry.m2lines import (
+    FloeNetBuilder,
+    NoiseConditionedSamudraBuilder,
+    SamudraBuilder,
+)
 from fme.ace.registry.sfno import SFNO_V0_1_0, SphericalFourierNeuralOperatorBuilder
 from fme.ace.registry.stochastic_sfno import NoiseConditionedSFNO
 from fme.ace.stepper import DerivedForcingsConfig, StepperOverrideConfig
--- a/fme/ace/models/ocean/m2lines/layers.py
+++ b/fme/ace/models/ocean/m2lines/layers.py
@@ -5,6 +5,8 @@ import torch
 import torch.nn as nn
 import torch.utils.checkpoint
 
+from fme.core.models.conditional_sfno.layers import Context, ContextConfig
+
 from .activations import CappedGELU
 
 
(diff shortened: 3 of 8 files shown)


elynnwu:
Claude: closing — the branch name did not follow CONTRIBUTING.md's `<type>/<short-description>` convention. Same commit reopened as #1464 on `feature/noise-conditioned-samudra`.

PR #1464, opened 2026-08-26
Add noise conditioning to Samudra

elynnwu:
Adds optional noise conditioning to the Samudra ocean model so it can be trained as a stochastic emulator. When `noise_embed_dim` is set above zero in `SamudraBuilder`, the network draws per-sample Gaussian noise and injects it through zero-initialized FiLM (Feature-wise Linear Modulation) layers after each normalization inside the selected ConvNeXt blocks, so an untrained model starts exactly deterministic.

The conditioning field arrives at the model's input resolution and is area-averaged onto each block's grid inside the U-Net, with a per-cell variance correction that accounts for the ragged windows `adaptive_avg_pool2d` builds at non-integer ratios, keeping every conditioned block at unit noise variance regardless of depth.

Changes:
- `MultiResolutionFiLM` (`fme.ace.models.ocean.m2lines.layers`): new FiLM layer that handles multi-resolution coarsening with optional variance preservation
- `ConvNeXtBlock` (`fme.ace.models.ocean.m2lines.layers`): accepts an optional `ContextConfig`; when given, places `MultiResolutionFiLM` modules after each norm layer in a separate `ModuleDict` so unconditioned state dicts are unchanged. Refactors duplicated norm construction into `_build_norm` and switches `convblock` from `Sequential` to `ModuleList` to allow per-layer conditioning dispatch
- `Samudra` (`fme.ace.models.ocean.m2lines.samudra`): accepts `context_config` and `conditioned_blocks` ("bottleneck" or "all_blocks"), threading a `block_context()` helper through construction to select which blocks are conditioned
- `SamudraBuilder` (`fme.ace.registry.m2lines`): adds `noise_embed_dim`, `conditioned_blocks`, and types `norm` as a `Literal`. When noise is enabled, wraps the built `Samudra` in `NoiseConditionedModel` (reused from the SFNO stochastic path). Validates that the two fields are set together
- `test_samudra.py`: tests for zero-init determinism, gradient flow, checkpoint compatibility, resampling semantics (area average with variance correction, not subsample), and config validation. Includes a released-checkpoint manifest test (`samudrace_e3smv3_ocean_manifest.json`) that rebuilds from the committed config and asserts the state dict matches the SamudrACE-E3SMv3 release
- `test_m2lines.py`: tests for the builder, registry round-trip, ensemble independence, and label rejection
- `test_ocean.py`: adds `test_ocean_corrector_is_per_member_under_ensemble_folding`, verifying that every ocean correction acts per-member when ensemble members are folded into the batch dimension

- [x] Tests added


elynnwu:
Pre-review agent: independent pre-review, run without the authoring session's context. Everything below was checked by running code against this branch and against `origin/main`, and by mutating the implementation to confirm the tests actually bite.

## Audit passes

**✅ 1. Silent-failure.** The new branches all fail loudly: `context_config` without `noise_injection` and vice versa, `embed_dim_noise <= 0`, non-noise embeddings, `norm=None` with conditioning, and a conditioned block called without a `Context`. `CONDITIONAL_BUILDERS` is correctly *not* extended — that list gates label conditioning, which `NoiseConditionedSamudraBuilder.build` rejects outright. Loading a `bottleneck` checkpoint under an `all_blocks` config (or vice versa) fails on missing/unexpected state-dict keys rather than silently.

**✅ 2. Behavior change on existing paths.** `ConvNeXtBlock.convblock` goes `Sequential` → `ModuleList`; nothing outside the class calls it as a callable (`test_m2lines.py:17` only indexes it, which still works). `forward` gains an optional second arg, so all existing positional calls are unaffected. With `checkpoint_strategy="all"` and no conditioning, the only change is an extra `None` passed into `torch.utils.checkpoint`.

**✅ 3. Differential tests — one real gap, fixed.** `test_convnext_block_conditioning_resamples_coarser_noise` only asserted the output *shape*. I replaced `adaptive_avg_pool2d` with a plain corner crop of the noise field — a completely different noise statistic — and **all 34 tests still passed**. Nothing in the suite pinned the resampling semantics, which is the single most novel piece of numerics in the PR. Pushed as `7314304`: the test now asserts that conditioning on the fine field equals conditioning on its area average, and that a stride-4 subsample of the same field gives a different answer. The crop mutation now fails it. No production code touched.

**⚠️ 4. Sibling coverage.** `NoiseConditionedSwinTransformerBuilder` — the closest sibling — does *not* subclass `SwinTransformerBuilder`; it repeats the fields and the `build` body. This PR takes the opposite approach and subclasses `SamudraBuilder`. Less duplication, but it's a deviation from the established pattern and it trips the abstract-or-final rule (see pass 7). Worth one line in the description either way.

**✅ 5. Config back-compat.** Verified directly rather than by inspection: I dumped `ConvNeXtBlock.state_dict()` key sets on this branch and on `origin/main` across the full matrix — `norm ∈ {instance, batch, layer, None}` × `upscale_factor ∈ {2, 4}` × `skip_module ∈ {identity, Conv2d}` — 16 combinations, **byte-identical key sets**. The `batch` case keeps `running_mean`/`running_var`/`num_batches_tracked`, the `None` case keeps the shifted indices (`convblock.1.cap`, `.2`, `.3.cap`, `.4`), and `in_channels != out_channels` keeps `skip_module.{weight,bias}`. The claim in the description holds well beyond the one config the test pins.

**⚠️ 6. Split heuristic.** 8 files, one feature plus a contained refactor. No split warranted.

**⚠️ 7. Abstract-or-final.** `SamudraBuilder` now has an in-repo subclass, so `build` and `__post_init__` on it should be `@abstractmethod` or `@final` per AGENTS.md — they're neither, and both are overridden. Either mark them, or follow the Swin sibling and make `NoiseConditionedSamudraBuilder` a standalone `ModuleConfig`.

**⚠️ 8. AI-agent artifacts.** Two small ones, both non-blocking:
- `_build_norm` opens with `assert self.norm_kwargs is not None` purely to satisfy mypy. Assigning `self.norm_kwargs: Mapping[str, Any] = norm_kwargs or {}` in `__init__` would type-narrow without the assert.
- The block-context threading (`iter([...])` + `next(...)` at four construction sites + an `object()` sentinel + a manual `raise AssertionError`) is the hardest part of the diff to read, and it encodes a construction-order invariant that a reader has to reconstruct from the comment. An explicit counter, or building the block list first and selecting by index, would be more obviously correct. The sentinel check is also unreachable: `n_blocks = 2 * num_steps + 1` is exactly what the four sites consume.

**✅ 9. Builder pattern.** No new cross-config field reads; the subclass reads its own inherited fields, which is within the "config passed to the impl it builds" exception.

## Verified claims

Each of these was run, not read:

- **Zero-init determinism is exact.** `max|conditioned(x, noise) − plain(x)| = 0.0` for `{bottleneck, all_blocks}` × `norm ∈ {instance, batch, layer}`, loading the plain model from the conditioned model's state dict minus the conditioning keys.
- **Checkpoint recompute does not redraw the noise.** Gradients under `checkpoint_strategy="all"` are bit-identical to the non-checkpointed run on the same `Context` (`max|Δgrad| = 0.0` over all parameters). This is sound by construction: the noise is drawn once in `NoiseConditionedModel.forward`, outside every checkpointed region, and non-reentrant checkpointing replays `forward` on the same tensor object.
- **Decoder resampling targets are right.** Instrumenting `NoiseConditioning.forward`, encoder and decoder blocks at the same U-Net level always receive the same grid, including for `img_shape=(9,18)` and `(45,90)` where the decoder pads and crops around the skips, and with `zonally_periodic_upsample=True`. Output shape always matches input.
- **The corrector test is not vacuous.** Mutating `LatLonOperations.area_weighted_mean` to reduce over `(0, *HORIZONTAL_DIMS)` makes it fail, and `fold_sized_ensemble_dim` does put members in dim 0, so it is testing the right layer. One gap: `surface_energy_flux_correction` is the one correction not in the tested config. It is pointwise so it's per-member by construction, but adding it would make the test cover the whole `OceanCorrectorConfig` surface.

## The one thing I'd want a decision on

**Area-averaging attenuates the noise, and it attenuates it most exactly where the default variant injects it.** Block-averaging unit white noise over a `k×k` window gives white noise with std `1/k`, so the resample doesn't change the *character* of the field, only its amplitude. Measured on the noise field the blocks actually see:

| block grid | noise std |
|---|---|
| 1/1 (finest) | 1.00 |
| 1/2 | 0.50 |
| 1/4 | 0.25 |
| 1/8 | 0.117 |
| 1/16 (bottleneck, 4 levels) | **0.058** |

With the default `noise_injection="bottleneck"` on the production 4-level `ch_width`, the *only* noise the model ever sees has std ≈ 0.06. `W_scale`/`W_bias` start at exactly zero, so the gradient that has to lift them off zero is scaled by that same factor — under Adam this is largely a reparametrization, but it changes the weight-decay and learning-rate interplay, and it makes `all_blocks` inject a conditioning signal that varies ~17× in amplitude between the finest and coarsest blocks, which is not what the SFNO reference pattern does (its latent never changes resolution, so `ConditionalLayerNorm` sees std-1 noise everywhere and does no resampling at all).

DLESyM-Ocean draws its bottleneck noise *at* the bottleneck resolution, i.e. std 1. Two ways to match that here: draw per-block noise at each block's grid, or keep the single input-resolution draw and rescale by `sqrt(H*W / (h*w))` after pooling. Both are one line. I deliberately did **not** change this — it changes the model, and which is wanted is a research call, not a pre-review call. But I think it should be an explicit decision recorded in the description rather than a side effect of the word "area-average", especially since `bottleneck` is the default.

## Distributed exposure (raised as the least-certain area)

The premise that `NoiseConditionedModel` slices noise to the local extent via `get_local_slices` doesn't hold for this path. For gaussian noise — Samudra's only option, since `inverse_sht=None` — the noise is drawn as `randn([x.shape[0], embed_dim, *x.shape[-2:]])`, i.e. directly at whatever spatial extent the rank holds. `get_local_slices` is used only for `pos_embed`/`label_pos_embed`, both `None` here. Only the `isotropic` path slices, and that's spectral-domain slicing for the iSHT.

So under spatial model parallelism each rank draws independent white noise on its own tile. That is distributionally identical to slicing one global draw (iid noise has no cross-tile structure to get wrong), and the per-block `adaptive_avg_pool2d` over the local tile is a local block-average of local white noise — still white, still the same std, so no silent wrongness. What *would* break first is Samudra itself: it uses plain `Conv2d` with a manual `F.pad(..., mode="circular")` along longitude and `AvgPool2d(2)`, both of which are wrong on a sharded tile, with no halo exchange anywhere. Samudra was never spatial-parallel-capable, and this PR doesn't change that. Exposure is identical to the existing SFNO gaussian path.

## Recommendation

**Ready for review, with one design question to settle first.** The mechanics are sound and I verified the three load-bearing claims (state-dict compatibility, zero-init determinism, checkpoint recompute) directly rather than trusting them. The noise-amplitude question above is the one thing I'd want an answer to before this trains, since it silently governs how hard the default variant can push. Pass-7 (`@final`) is a small cleanup; the rest are nits.

I pushed one commit (`7314304`) strengthening the resample test; no production code changed, and the touched suites (52 tests) plus `pre-commit` are green.

elynnwu:
Claude: addressed the pre-review. Pushed df5aa6a6f.

**The amplitude finding — taken, and fixed rather than documented.** The review was right that this needed an explicit decision, but I don't think it's a research call: the conditioning weights are zero-initialized and share a learning rate, so a per-depth amplitude difference silently rescales both the conditioning signal and its gradient. Nobody would choose a 20x spread across depths; it was a side effect of the word "area-average", exactly as the review said.

Measured on the actual 4° grid (45x90), area-averaging gave: 45x90 → std 1.00, 22x45 → 0.41, 11x22 → 0.20, 5x11 → 0.11, 2x5 → **0.049**. So the default `bottleneck` variant was conditioning on noise of std 0.05.

Fixed by subsampling instead of area-averaging. Subsampling iid noise leaves it iid, so every depth gets unit-variance white noise regardless of the shape ratio — which also sidesteps the non-integer-ratio problem (45 → 22 has mixed 2- and 3-row windows, so no single rescale factor would have been exact). That matches both references the PR cites: the SFNO's latent never changes resolution so `ConditionalLayerNorm` always sees std-1 noise, and DLESyM-Ocean draws its bottleneck noise at bottleneck resolution.

The review's resample test is kept and inverted — it now pins the subsample and rejects the area average — plus a new test asserting unit variance at every block resolution of the 4° grid. Thank you for that test; the shape-only assertion it replaced was the weakest part of the PR and the mutation you ran proved it.

**Smaller items, all taken:**
- `NoiseConditionedSamudraBuilder` is now a sibling of `SamudraBuilder`, not a subclass, matching `NoiseConditionedSwinTransformerBuilder`. That resolves the abstract-or-final violation at the root rather than papering over it, at the cost of repeating the architecture fields.
- Dropped the unreachable `norm is None` guard.
- Replaced the block-context iterator + `object()` sentinel + manual `AssertionError` with a `block_context()` helper called once per block.
- Typed `norm_kwargs` on assignment, dropping the mypy-only assert.

**Not changed:**
- `surface_energy_flux_correction` is still absent from the ensemble-safety test's config. Agreed it's pointwise and safe by construction; adding it would mean a full atmosphere-forcing fixture for no additional coverage.
- Distributed: agreed the premise of my flag didn't hold — the gaussian path draws at the rank's local extent and never touches `get_local_slices`, so there's nothing to be wrong. Worth recording that Samudra is not spatial-parallel-capable for an unrelated reason (circular `F.pad` and `AvgPool2d` with no halo exchange), and this PR neither helps nor hurts that.

elynnwu:
Pre-review agent: delta re-review of `7314304a4..df5aa6a6f`, again without the authoring context. **The delta is clean and the behaviour change is right.** I found one vacuous test and fixed it (`5d2d0eb53`); nothing else blocks.

## 1. Is subsampling the right choice over a variance-corrected area average?

Yes, and your stated reason is the correct one. I'll add the argument that I think settles it, and one place your docstring undersells the result.

Your objection to rescaling the area average is not just inconvenience — a scalar factor is *provably* insufficient. `adaptive_avg_pool2d` on 45 → 22 uses mixed 2- and 3-row windows, so post-average per-cell std is 1/√2 for some cells and 1/√3 for others. Any single scalar leaves per-cell variance heterogeneous across the grid (ratio 1.22 between neighbours), which is a subtler version of exactly the defect you set out to remove. Adaptive pooling doesn't expose the per-cell window counts, so there is no clean fix. Subsampling sidesteps this entirely.

On the property you gave up: I agree it doesn't matter, and more strongly than you argued. For an iid field, a correctly-normalised block average and a subsample are **identically distributed** — both are iid N(0,1) on the coarse grid. There is no distributional difference at all, so nothing about "injecting variance at coastal / WBC scales" can distinguish them. The only difference is the *joint* fine↔coarse relationship, which is visible only under `all_blocks` where one block sees the fine field and another sees the coarse one. And neither option gives a principled multiscale coupling: averaging makes a coarse cell correlate 1/k with each fine cell in its footprint, subsampling makes it correlate 1 with one cell and 0 with the rest. If genuinely scale-coherent noise were wanted, the answer is a low-pass-filtered single field — which is what the SFNO's `isotropic` mode does — not either of these. That's out of scope here.

Worth stating explicitly in the docstring, because it's a stronger claim than the one you make: **for the default `bottleneck` variant, subsampling a fine field is exactly equivalent in distribution to drawing at bottleneck resolution.** Only one block is conditioned, so there is no fine field in play for it to be jointly related to. The default now *is* DLESyM-Ocean's construction, not merely an approximation of it — which is what the docstring's DLESyM-Ocean citation wants to be saying.

## 2. Does `mode="nearest"` do what you think on non-integer ratios?

**Yes — verified, and it's provable rather than merely empirical.** You checked the aggregate std; here is the pairwise-independence check you wanted.

I fed a field whose values *are* the flat source index, so the output reveals exactly which source cell each output cell drew from. Across `45→{45,22,11,5,2}`, `90→{90,45,22,11,5}`, `180→{90,45,22,11}`, `360→{180,90,45,22}`, and the full 2-D `45x90 → {(45,90),(22,45),(11,22),(5,11),(2,5)}` chain, **every output cell drew a distinct source cell** — index maps strictly increasing, `n_distinct == n_out` in every case. No reuse anywhere, including at every non-integer ratio.

This is guaranteed, not lucky. PyTorch nearest maps `src = floor(dst * input/output)`; for any downsampling ratio `s ≥ 1`, consecutive outputs differ by `floor((i+1)s) - floor(is) ≥ 1`, so indices are strictly increasing by construction. The coarse field is therefore an exact *subset* of the iid fine cells: exactly iid N(0,1), no correlation artifact, no variance artifact. Measured over 4000 samples, lag-1 correlations are `≤ 0.004` at every resolution except the 10-cell `(2,5)` grid where `+0.012` is ~1σ of the estimator. Std is `1.000, 1.000, 1.001, 0.999, 1.002` versus the old `1.000, 0.408, 0.200, 0.110, 0.049`.

One latent caveat, not reachable today: the strictly-increasing argument needs ratio `≥ 1`. If `NoiseConditioning` were ever handed noise *coarser* than the block grid, nearest would upsample and duplicate cells — reintroducing precisely the correlation you removed. Samudra can't do this (blocks are never finer than input resolution, and noise is drawn at input resolution), so it's a note rather than a finding. A one-line guard would make the assumption explicit if you want it.

## 3. Do the two tests pin this? — one was vacuous, now fixed

Applying the same mutation discipline: I swapped `mode="nearest"` for bilinear, `nearest-exact`, `adaptive_avg_pool2d`, a variance-rescaled `adaptive_avg_pool2d`, a corner crop, `adaptive_max_pool2d`, `mode="area"`, and a fixed-integer-stride subsample. Result across all eight:

- `test_convnext_block_conditioning_resamples_coarser_noise` caught seven of eight.
- `test_conditioning_noise_keeps_unit_variance_at_every_depth` caught **zero**.

It was vacuous with respect to the production code: it computed `resampled` itself with `F.interpolate(..., mode="nearest")` and asserted on *that*, so it tested PyTorch rather than `NoiseConditioning`. The one line that touched the module — `conditioning(x, noise)` — had no assertion, so the comment "and the conditioning really consumes that resampled field" was unbacked.

That left a real gap, not just a redundant test. The surviving mutation was the **fixed integer stride** (`noise[..., ::H//h, ::W//w]`), which passed all 35 tests. It coincides with nearest at the ratios the other test uses (16→4, 32→8), but diverges on `90 → 11` and degenerates to a plain crop whenever `H//h == 1` (e.g. 45 → 30). A broken resampler was passing the suite.

Fixed in `5d2d0eb53`, test-only. The test now reads the resampled field straight off the module (zeroed scale, identity bias, zero input, so the output *is* the resampled noise) and asserts the two properties that actually matter: every output cell draws a distinct source cell, and the sample spans the field rather than cropping a corner — using the index-valued-input trick so the source map is directly observable at the non-integer 45→22 and 45→11 ratios. It now catches **all seven** semantic mutations on its own, including the integer-stride one that previously survived everything.

It deliberately still *passes* `nearest-exact`, which is an equally valid distinct subsample with identical statistical guarantees. That leaves a clean division: this test pins the property, the other pins the alignment. Tight without being over-specified.

## Smaller items — all confirmed, nothing broken

**`block_context()` selects identically.** This was the off-by-one risk, so I reimplemented the old set-based rule independently and compared block-by-block across `len(ch_width)` 1–6 × `{bottleneck, all_blocks, None}` — 18 configurations. Block counts all `2*num_steps+1`, and the conditioned mask matches exactly in every case, with the bottleneck landing at index `num_steps` throughout. No drift. Genuinely easier to read, too. One nit: `assert n_built == 2 * num_steps + 1` is a bare `assert` where the old code raised, so it vanishes under `python -O` — immaterial for an internal invariant, but the old form was stricter.

**Standalone builder copy is faithful.** All nine architecture fields present with identical names, types, defaults, *and* declaration order; only `noise_embed_dim` and `noise_injection` added. `dataclasses.asdict` defaults capture works (`noise_embed_dim: 32`, `noise_injection: "bottleneck"` both serialized), `from_state` round-trips equal, and re-selecting from the serialized config reproduces it. Matches the `NoiseConditionedSwinTransformerBuilder` pattern as intended, and the pass-7 `@final` issue is gone since `SamudraBuilder` has no subclass any more.

**`norm_kwargs` — no aliasing change.** Both old and new store the caller's object by reference when non-None (verified `is` identity) and produce a fresh `{}` per instance on None (verified not shared across instances). The mypy-only assert is correctly gone.

**Dropped `norm is None` guard** — correct, it was unreachable: `norm` is typed `str`, so dacite rejects null.

**Invariants re-verified at the new head**, since the resampler changed underneath them: zero-init determinism still exact (`max|diff| = 0.0` for both variants × `{instance, batch, layer}`), checkpoint recompute still bit-identical (`max|Δgrad| = 0.0`), and encoder/decoder block grids still mirror correctly on `(16,32)`, `(9,18)`, `(45,90)`, `(180,360)` and with `zonally_periodic_upsample=True`.

53 tests across the touched suites plus `pre-commit` (ruff, ruff-format, mypy) are green.

## Verdict

**The delta is clean — ready for review.** You were right to treat the amplitude finding as a behaviour bug rather than a research call, and the reasoning you gave for subsampling over a rescaled area average holds up under attack; the mixed-window argument is decisive and I could not find a case where nearest reuses a source cell. The refactors are faithful. The one thing your own testing missed was that the new test didn't touch the production path, which had let a genuinely broken resampler through the suite; that's fixed and mutation-verified.

elynnwu:
Claude: took both nits from the delta re-review. Pushed a8559e817.

- `NoiseConditioning` now raises when the noise field is coarser than the block grid, rather than silently upsampling. You were right that the ratio >= 1 assumption was load-bearing and unstated: subsampling only stays iid while it discards cells, so an upsample would duplicate them and reintroduce exactly the correlation-and-amplitude defect this resampling exists to remove. Unreachable from Samudra, which is why it is a raise rather than a handled case.
- The block count check raises instead of asserting, so it survives `python -O`. Worth doing precisely because the failure is silent: miscounting moves which block `bottleneck` conditions without changing any shape.

On the vacuous test — that one is a fair hit and the mutation that survived it is the instructive part. A fixed integer stride coincides with `nearest` at the ratios my other test happened to use and degenerates to a plain crop whenever `H // h == 1`, so a broken resampler was passing the suite. I have kept your replacement as-is; reading the resampled field back off the module is the right shape for that assertion and I would not have got there by tightening my own version.

Two points from your review I want to leave on the record rather than act on:

1. Your observation that for the default `bottleneck` variant there is no fine field to be jointly related to, so subsampling is *exactly* equivalent in distribution to drawing at bottleneck resolution, is a better statement of the DLESyM correspondence than the one in my docstring. I have not rewritten the docstring for it, since it is a property of the default rather than of the class, but it belongs in the record.
2. Genuinely scale-coherent noise means a low-pass-filtered field -- the SFNO's `isotropic` mode -- and neither subsampling nor averaging gives that. If the 4-degree screen shows the injected variance sitting at the wrong scales, that is the next thing to try, not a different resampling.

mcgibbon:
Hm maybe we should update the pre-review skill so it puts its posted pre-reviews in collapsible/spoiler sections in the comments.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,71 @@ def forward(self, x):
...
+
+    Applied immediately after a non-affine normalization layer, this turns that
+    norm into a conditional one, the same construction ``ConditionalLayerNorm``
+    uses for the SFNO: ``x -> x * (1 + W_scale(noise)) + W_bias(noise)``. Both
+    convolutions are zero-initialized, so an untrained model is exactly
+    deterministic and stochasticity is learned from the ensemble loss.
+
+    The noise field is drawn at the model's input resolution, while a
+    conditioned block may run at a coarser resolution inside the U-Net, so the
+    noise is subsampled onto the block's grid before projection. Subsampling
```
mcgibbon: Issue (non-blocking for this PR): Hmmmm I actually am a bit worried about subsampling for this. That means e.g. down three depths of a u-net, a 8x8 domain patch would be using exactly the upper-left (for example) corner of the patch's fine-res data for the coarse-res stochasticity. Correlating the noise in this way seems really odd! I don't think this will significantly affect your results other than maybe causing some strange artifacting within stochastic fields, but I think we should iterate on this before publishing anything.

Suggestion (optional): A better solution would be to simply multiply the noise by a scalar which accounts for the reduction in variance expected due to the area-averaging. Another possibility is to provide distinct noise arrays for each resolution/level of the convnet, though this is a bigger code change.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,71 @@ def forward(self, x):
         return self.avgpool(x)
 
 
+class NoiseConditioning(torch.nn.Module):
```
mcgibbon: Issue (non-blocking): I was originally thinking "this should specify something about the quality of the noise, is it isotropic or Gaussian?" and then realized that this object actually has nothing to do with noise - it conditions on any arbitrary tensor that just happens to be named `noise`. Its main responsibility is actually to handle the multi-resolution nature of the ConvNet.

Probably a good idea to rename appropriately, and update the documentation here to either give noise as an example, or talk in more general terms (e.g. "conditioning tensor", "context")

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,71 @@ def forward(self, x):
...
+                # would duplicate them, correlating neighbours and breaking the
+                # unit variance this resampling exists to preserve. Samudra
+                # never asks for it (the noise is drawn at input resolution and
+                # every block is coarser), so this is an invariant, not a case
+                # to handle.
+                raise ValueError(
+                    f"noise field {tuple(noise.shape[-2:])} is coarser than the "
+                    f"block grid {target}; it must be at least as fine."
+                )
+            noise = torch.nn.functional.interpolate(noise, size=target, mode="nearest")
```
mcgibbon: Issue: It's not clear to me what nearest interpolation does for a 2x2 tensor being coarsened to 1x1, aren't all input coordinates just as near as any other?

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -107,84 +178,70 @@ def __init__(
...
             torch.nn.Conv2d(
-                in_channels=int(in_channels * upscale_factor),
+                in_channels=hidden_channels,
                 out_channels=out_channels,
                 kernel_size=1,
                 padding="same",
             )
         )
-        self.convblock = torch.nn.Sequential(*convblock)
+        self.convblock = torch.nn.ModuleList(convblock)
```
mcgibbon, line 251: Question: will this load old checkpoints? Do you have a test for this that uses a previously serialized version? I think we have test infrastructure for this. Or do we not need backwards compatibility here because no released checkpoints?

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -107,84 +178,70 @@ def __init__(
...
+                    hidden_channels, embed_dim_noise
+                )
+
+    def _build_norm(self, norm: str | None, num_features: int) -> list[torch.nn.Module]:
+        if norm == "batch":
+            return [torch.nn.BatchNorm2d(num_features, **self.norm_kwargs)]
+        elif norm == "instance":
+            return [torch.nn.InstanceNorm2d(num_features, **self.norm_kwargs)]
+        elif norm == "layer":
+            return [torch.nn.LayerNorm(num_features, **self.norm_kwargs)]
```
mcgibbon, line 266: Issue (non-blocking): Because of how noise conditioning words, it's particularly valid to apply it to LayerNorm conditioned fields, and arguably invalid to apply to other forms of normalization. This is because the Linear projection happens over the hidden dimension, i.e. for each channel that projection is constant. So LayerNorm normalizes across channels and then the noise conditioning does mixing across that normalized channel dimension.

If instead we do instance norm, which normalizes each channel across the horizontal domain, the noise projection into each of those channels is _set in stone_ (learned as a parameter). That means every single sample every single time gets exactly the same noise expectation added to it relative to its signal strength, regardless of what that sample is.

By contrast, when the layer before a LayerNorm provides larger relative magnitude to a certain channel, it's basically able to say "yeah for this specific example, don't inject as much noise" by doing so - the constant-expected-magnitude noise will be smaller relative to the magnitude of the input field.

This is why I opted to create a ConditionalLayerNorm in the SFNO code rather than put a FiLM layer (https://arxiv.org/abs/1709.07871, which is what this conditioning is) after an arbitrary norm.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -107,84 +178,70 @@ def __init__(
...
+                in_channels=hidden_channels,
                 out_channels=out_channels,
                 kernel_size=1,
                 padding="same",
             )
         )
-        self.convblock = torch.nn.Sequential(*convblock)
+        self.convblock = torch.nn.ModuleList(convblock)
+
+        self.noise_conditioning = torch.nn.ModuleDict()
```
mcgibbon, line 253: Question (non-blocking): Why a ModuleDict taking indices instead of a ModuleList? You could consider using nn.Identity (or a custom version that takes two tensors) for placeholders, maybe even one that asserts if it's passed non-none conditioning - it might simplify the code.

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -178,10 +242,18 @@ def forward(self, fts):
                 fts = torch.nn.functional.pad(
                     fts, (0, 0, self.N_pad, self.N_pad), mode="constant"
                 )
+            # only the ConvNeXt blocks are conditionable; the pooling, upsample
+            # and final conv layers take the tensor alone
+            if isinstance(layer, ConvNeXtBlock):
```
mcgibbon, line 242: Suggestion (non-blocking): isinstance is always a red flag to me, but there may not be a good alternative here. One _possibility_ is if you are able to define a wrapper class `UnconditionalModule` that takes in the `context`, discards it, and calls the wrapped class, but it would need to be done in a way that doesn't mess up the state dicts. I think there's a way to make this work (possibly involving attributes or classes that aren't Module subclasses) but I don't know if it's worth your time or what the final complexity of the code would look like, so maybe leave it as-is.

fme/ace/registry/m2lines.py
```diff
@@ -57,6 +59,94 @@ def build(
...
+    n_layers: list[int] = dataclasses.field(default_factory=lambda: [1, 1, 1, 1])
+    dilation: list[int] = dataclasses.field(default_factory=lambda: [1, 2, 4, 8])
+    pad: str = "circular"
+    norm: str = "instance"
+    norm_kwargs: Mapping[str, Any] = dataclasses.field(default_factory=dict)
+    upscale_factor: int = 4
+    checkpoint_strategy: Literal["all", "simple"] | None = None
+    zonally_periodic_upsample: bool = False
+    noise_embed_dim: int = 32
+    noise_injection: NoiseInjection = "bottleneck"
```
mcgibbon: Issue: the name "noise injection" locks in (in user-facing config) the fact that noise injection needs to be controllable separately from future potential context types we might condition on. Maybe name something that uses the word "conditioning" instead of "noise"?

fme/ace/registry/m2lines.py
```diff
@@ -57,6 +59,94 @@ def build(
...
+            context_config=context_config,
+            noise_injection=self.noise_injection,
+        )
+        return NoiseConditionedModel(
+            samudra,
+            img_shape=dataset_info.img_shape,
+            embed_dim_noise=self.noise_embed_dim,
+            embed_dim_pos=0,
+            n_labels=0,
+            label_embed_dim=0,
```
mcgibbon, line 113: Comment (non-blocking): Might consider adding label support as follow-on or a quick Claude test, since 1) it should be a small change and 2) I'm a little worried there may be a subtle issue with the way this config is structured or the way the existing optional-label-conditioning code works that will make it a bigger change, and that would be good to know sooner rather than later.

fme/core/corrector/test_ocean.py
```diff
@@ -539,3 +539,94 @@ def test_ocean_corrector_empty_delta_when_nothing_modified():
...
     torch.testing.assert_close(result.corrected["so_0"], gen_data["so_0"])
+
+
+def test_ocean_corrector_is_per_member_under_ensemble_folding():
+    """Ensemble training folds the ensemble members into the batch dimension, so
+    the corrector sees several members at once. Every correction must act
+    per-member: one that coupled across the batch dim (e.g. a global mean taken
+    over samples too) would tie the members together and silently collapse the
+    ensemble spread the proper scoring rule is meant to reward.
+    """
```
mcgibbon, line 550: Question: How is this related to noise conditioning?

fme/ace/registry/m2lines.py
```diff
@@ -57,6 +59,94 @@ def build(
         )
 
 
+@ModuleSelector.register("NoiseConditionedSamudra")
+@dataclasses.dataclass
+class NoiseConditionedSamudraBuilder(ModuleConfig):
```
mcgibbon: Question: What's the reason to keep this separate instead of using `noise_embed_dim: int = 0` by default in SamudraBuilder? We'll have to keep both of these configs up to date each time something is changed, so it's worth asking, unless you think we'll stop using SamudraBuilder altogether (which is how we avoided this for the SFNO).

Suggestion (non-blocking/optional): If we are going to keep using SamudraBuilder, should we explicitly construct the SamudraBuilder configuration class and call its .build method in this class's .build method, as a way to guard against the two going out-of-date with respect to each other? Or maybe even explicitly define a `samudra: SamudraBuilder` sub-config?

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -43,6 +46,16 @@ class Samudra(torch.nn.Module):
...
         the behavior of checkpoints trained without it.
+    context_config : ContextConfig, optional
+        If given (with a non-zero noise embedding), the ConvNeXt blocks selected
+        by ``noise_injection`` take a conditional scale and bias off the noise
+        field in the ``Context`` passed to ``forward``. Only noise conditioning
+        is supported; scalar, label, and positional embeddings are not.
+    noise_injection : {"bottleneck", "all_blocks"}, optional
+        Which ConvNeXt blocks are noise-conditioned. ``"bottleneck"`` conditions
+        only the block at the coarsest resolution (DLESyM-Ocean's choice, which
+        perturbs large scales); ``"all_blocks"`` conditions every block (the
```
mcgibbon: Comment (aside): Perturbing only large scales seems like such an odd choice! Isn't most of the stochasticity in the small scale?

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -43,6 +46,16 @@ class Samudra(torch.nn.Module):
         longitude axis in the decoder, removing the lon=0 seam introduced by the
         default (non-periodic) bilinear upsampling. By default False to preserve
         the behavior of checkpoints trained without it.
+    context_config : ContextConfig, optional
+        If given (with a non-zero noise embedding), the ConvNeXt blocks selected
+        by ``noise_injection`` take a conditional scale and bias off the noise
+        field in the ``Context`` passed to ``forward``. Only noise conditioning
+        is supported; scalar, label, and positional embeddings are not.
```
mcgibbon: Issue: We can maybe be OK with this, but I'm nervous about taking the Context and saying we're going to add conditioning on specifically noise. In the SFNO code the idea is there are "conditioning points" and there's "what we're conditioning on" as two orthogonal choices, which keeps decision points down (especially since we don't allow decisions on the conditioning points). Should we be independently choosing where and if we condition on labels, or positional embedding tensors, for example?

It's kind of nice in the SFNO that to add a new conditioning type it just gets added to the Context and to the ConditionalLayerNorm that applies it, and none of the SFNO code itself needs to change/think about it.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Claude: Changed it to the area average with the variance compensation you suggested, but the compensating factor can't be a single scalar for the whole field.

Averaging n cells into one divides the standard deviation by `sqrt(n)`, so the correction is `sqrt(n)`. When the resolution ratio is an integer, every output cell averages the same n and one scalar is right. When it isn't, `adaptive_avg_pool2d` builds windows of unequal size — neighboring output cells cover different numbers of input cells — so n varies across the field and no single scalar restores unit variance everywhere. The overall area ratio gets the average amplitude about right while leaving individual cells above and below it.

So n is computed per output cell from the pooling window definition, and the average is scaled by `sqrt(n)` there. That is exact. The test now checks each output cell's variance separately rather than the field's as a whole, and fails if the correction is applied as a single scalar.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Claude: Renamed to `MultiResolutionFiLM` and rewrote the docstring around a general conditioning tensor, with the noise field as what Samudra passes today. It also now cites the FiLM paper you linked, since that's the clearest name for what it does.

The generalization did surface one thing worth an explicit knob: the variance compensation above is correct for an iid field and wrong for a smooth one, whose coarsened magnitude should be left alone. So it's `preserve_variance`, defaulting True, rather than unconditional.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Claude: It's deterministic but arbitrary — PyTorch picks by a floor index formula, so it takes the top-left cell. That was part of why it was the wrong call, and it's gone: the resampling is `adaptive_avg_pool2d` now.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Claude: Agreed on the mechanism: after instance norm the conditioning magnitude relative to the normalized signal is a learned constant, identical on every sample and every step, so the block can't say "not this one." I've documented that in `ConvNeXtBlock` and `SamudraBuilder`, naming layer as the principled choice and keeping instance permitted only because ranking them is an open question.

I did the first round of experiment using instance norm and now re-doing them with layer norm.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Claude: Left as a `ModuleDict`. The keys are the indices of the norm layers within `convblock`, so an all-Identity `ModuleList` would have to be as long as `convblock` and mostly padding, and the empty dict is what keeps an unconditioned block's state dict byte-identical to main. A custom two-argument `Identity` would also have to be a `Module` to sit in the list, which puts placeholder entries back into the state dict.

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Claude: Left as-is. The `UnconditionalModule` wrapper is the kind of thing that reads clean until it has to not perturb state-dict keys.

We can probably clean up on this later.

fme/ace/registry/m2lines.py
elynnwu: Claude: Renamed to `conditioned_blocks`, in the module and in config. It also has no default now and is required exactly when `noise_embed_dim` is non-zero, so the injection point is always an explicit choice rather than an inherited `bottleneck`.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Yes, verified against SamudrACE-E3SMv3 ocean [ckpt](https://huggingface.co/allenai/SamudrACE-E3SMv3/tree/main). 

Since we don't want to download the 300MB+ ckpt in every CI test and the property at risk is structural rather than numerical. Add a test that contains only the names and shapes the release contains and builds from the committed config and asserts the built key set and shapes match exactly.

fme/ace/registry/m2lines.py
elynnwu: Claude: Agreed it's worth finding out early, and filing it as a follow-on rather than growing this PR. One thing that's changed in its favor: with the two builders merged (below), label support is a change to one config class instead of two that would have had to stay in sync.

[Issue](https://github.com/ai2cm/ace/issues/1468) opened.

fme/core/corrector/test_ocean.py
elynnwu: Claude: Through ensemble folding. Training with `n_ensemble` > 1 folds the members into the batch dimension, so the corrector sees several members of the same sample at once. Any correction that reduced across the batch dim — a global mean taken over samples as well as space — would tie the members together and collapse the spread the proper scoring rule exists to reward, silently, with no error anywhere. This module is the thing that makes ensemble training of the ocean possible, so the test is checking a precondition it introduces.

fme/ace/registry/m2lines.py
elynnwu: Claude: No good reason, merged now. `SamudraBuilder` now carries `noise_embed_dim` and `conditioned_blocks`, build returns the bare `Samudra` or the `NoiseConditionedModel` wrapper accordingly, and the `NoiseConditionedSamudra` registry entry is gone. You were right about the drift risk; the two classes had already diverged in their `__post_init__`. It also removes the `SamudraBuilder` subclass that was making its methods need `@final`.

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Claude: Agreed, and the arithmetic is worse than the docstring made it sound. It was DLESyM-Ocean's choice and got carried over uncritically. The docstring now states what it implies, and `conditioned_blocks` has no default so nobody inherits it silently.

elynnwu: Have a new series of experiments using `all_blocks` now

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Claude: Haven't acted on yet, and I think you're right that it's the correct end state: "where we condition" and "what we condition on" as orthogonal axes, with a new conditioning type touching only `Context` and the layer that applies it.

What's here is a step toward it rather than away: `MultiResolutionFiLM` is now type-agnostic and takes a conditioning tensor, so the noise-specific part has shrunk to `ConvNeXtBlock` reading `context.noise` and `Samudra` validating that the other embeddings are zero. Making it fully orthogonal means the block consuming a general `Context` and the `FiLM` projections being per-conditioning-type, which is the same change as adding label support. I think we should do that as a follow-up PR ([issue](https://github.com/ai2cm/ace/issues/1468)).

mcgibbon:
BTW was the PR description written with Opus 4.6? If not I would suggest asking to have it blind re-written, using a fresh agent instructed very explicitly not to look at the current PR title or description, only the changelog (this might mean it can't use the script in the now-merged skill).

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,134 @@ def forward(self, x):
         return self.avgpool(x)
 
 
+class MultiResolutionFiLM(torch.nn.Module):
+    """Zero-initialized conditional scale and bias driven by a conditioning field.
```
mcgibbon, line 68: I would suggest scanning PRs for long comment blocks like this and specifically reviewing them as an author to see if they can/should be shorter or if parts should be removed.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,134 @@ def forward(self, x):
...
+    reference implementations have no such gradient in amplitude -- the SFNO's
+    latent never changes resolution, so ``ConditionalLayerNorm`` always sees
+    std-1 noise, and DLESyM-Ocean draws its bottleneck noise at bottleneck
+    resolution. ``preserve_variance`` restores that by scaling each averaged cell
+    back up by the square root of the number of cells that went into it, exact
+    for an iid field. (The count is per cell rather than a single area ratio
+    because adaptive pooling on a non-divisible ratio builds ragged windows;
+    those windows also overlap slightly, so neighbouring coarse cells are weakly
+    correlated -- inherent to coarsening 45 rows onto 22, and far milder than
+    the corner artifact a subsample would leave.)
```
mcgibbon: Issue (blocking): Still too much prose for what this is.

Suggestion: Can replace these three paragraphs with a short paragraph that says the multi-resolution handling uses area-averaging of the noise, also re-scaling that noise so that the expected variance of the noise field is constant across resolutions.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,134 @@ def forward(self, x):
...
+    ``ConditionalLayerNorm`` implements for the SFNO, and it is the one that
+    behaves: LayerNorm normalizes across channels, and the 1x1 projection is
+    constant per channel, so a preceding layer can raise one channel's relative
+    magnitude to make the conditioning smaller relative to the signal for that
+    sample. After ``instance`` norm -- which normalizes each channel over the
+    horizontal domain -- that escape hatch is gone: the conditioning magnitude
+    relative to the normalized signal is a learned constant, identical for every
+    sample on every step, so the block cannot modulate how strongly it is
+    conditioned. ``instance`` remains permitted, because ranking the two is an
+    open experimental question, but ``layer`` is the principled choice.
```
mcgibbon: Delete this paragraph. I expect this paragraph is spawned from the comment I gave in the last review, but it doesn't serve a purpose that addresses that issue - I don't expect users to read this deeply into the Module code to find these details, and a user editing this code never really needs to know these details (they need to be known when configuring modules). Instead it just adds verbosity. You could put a _short_ note in the Config/Builder to expose this to the user (optional).

fme/ace/models/ocean/m2lines/layers.py
mcgibbon: Make sure you're running a non-stochastic layer norm baseline as well, in case that degrades the model.

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -97,6 +115,44 @@ def __init__(
...
+            if conditioned_blocks is None:
+                raise ValueError("context_config requires conditioned_blocks to be set")
+        elif conditioned_blocks is not None:
+            raise ValueError("conditioned_blocks requires context_config to be set")
+        self.conditioned_blocks = conditioned_blocks
+
+        # Blocks are built in this order: the num_steps encoder blocks, the
+        # bottleneck block, then the num_steps decoder blocks. `block_context`
+        # is called once per block, in that order, and hands back the context
+        # for the blocks this injection variant conditions.
```
mcgibbon: Question: Does this comment serve a purpose the code doesn't already serve, or should it be removed?

fme/ace/registry/m2lines.py
```diff
@@ -4,16 +4,41 @@
...
+        conditioned_blocks: Which ConvNeXt blocks are conditioned, required
+            when ``noise_embed_dim`` is non-zero and rejected otherwise.
+            "bottleneck" conditions only the block at the coarsest resolution;
+            after the encoder's AvgPools that grid is 1/16 of the input, so on a
+            45x90 domain it is 2x5 cells and only the largest scales can be
+            perturbed. "all_blocks" conditions every block, the pattern the ACE
+            SFNO uses, which also reaches the finest scales.
+        norm: Note that conditioning is applied as a FiLM layer after the norm,
+            which is principled after "layer" (the SFNO's ConditionalLayerNorm
+            construction) and degenerate after "instance" -- see ConvNeXtBlock.
```
mcgibbon: Rather than referencing ConvNeXtBlock, I'd suggest putting a short message in this class (blocking). As a reader I wouldn't be likely to dig around and find that documentation.

Depending on the results of your experiments, you could also consider raising an exception if noise_embed_dim > 0 and norm is not "layer".

Also might consider typing norm with a `Literal` to make these dynamics a bit clearer (optional, pre-existing type is str).

fme/ace/registry/m2lines.py
```diff
@@ -4,16 +4,41 @@
...
     """
     Configuration for the M2Lines Samudra architecture.
+
+    Setting ``noise_embed_dim`` above zero makes the network noise-conditioned:
+    a noise field is drawn on every forward call and supplied as conditioning
+    input to the ConvNeXt blocks named by ``conditioned_blocks``, so an ensemble
+    of members can be drawn from one input and trained against a proper scoring
+    rule. The conditioning weights are zero-initialized, so an untrained
+    noise-conditioned model is exactly deterministic and identical to the
+    unconditioned one.
```
mcgibbon: (optional) Can remove unnecessary detail.

```suggestion
    Setting ``noise_embed_dim`` above zero makes the network noise-conditioned,
    zero makes it deterministic.
```

fme/ace/registry/m2lines.py
```diff
@@ -4,16 +4,41 @@
...
+
+    Parameters:
+        noise_embed_dim: Number of noise channels drawn and projected onto each
+            conditioned block's scale and bias. Zero (the default) builds the
+            deterministic network; DLESyM-Ocean uses 32.
+        conditioned_blocks: Which ConvNeXt blocks are conditioned, required
```
mcgibbon: Issue: Unclear what "None" (default) does.

fme/ace/models/ocean/m2lines/samudra.py
```diff
@@ -159,14 +218,22 @@ def __init__(
...
                 checkpoint_strategy=self.checkpoint_strategy,
+                context_config=block_context(),
             )
         )
         layers.append(torch.nn.Conv2d(b, self.output_channels, self.last_kernel_size))
 
+        if n_built != 2 * num_steps + 1:
+            # a bare assert would vanish under `python -O`, and miscounting
+            # silently moves which block "bottleneck" conditions
+            raise AssertionError(
```
mcgibbon:
```suggestion
            raise AssertionError(
```

mcgibbon (commented):
Comments are basically all on the AI docs. If you use AI to iterate on these, I'd suggest doing a manual pass after.

fme/ace/registry/m2lines.py
elynnwu: Unfortunately I don't have enough evidence supporting noise_embed_dim > 0 and layer norm is working well. My previous statement on promising stochastic Samudra was mis-configured to use instance norm, which I agree is not the right thing to do. I think we can either wait to merge this PR (more experimetns are in progress) or merge now with the knowledge that we don't have a working stochastic version and this PR is simply to support running those experiments.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Updated

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Deleted. The short note now lives on `SamudraBuilder.norm`

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Removed mostly.

fme/ace/registry/m2lines.py
elynnwu: Done

fme/ace/registry/m2lines.py
elynnwu: Updated to say None is for deterministic

fme/ace/models/ocean/m2lines/samudra.py
elynnwu: Removed

elynnwu:
Updated PR title and description via the `write-pr-prose` skill.

**Model**: Claude Opus 4.6 (1M context) (`claude-opus-4-6[1m]`)

**Context gap**: The gather script's GraphQL query for the existing title/description hit a GitHub API deprecation notice for Projects (classic) (`(warning: could not retrieve title/description)`), so the prose was written without the pre-existing description in view. The diff, PR comments, and review threads were all retrieved successfully.

fme/ace/models/ocean/m2lines/layers.py
```diff
@@ -62,11 +64,98 @@ def forward(self, x):
...
+    """Zero-initialized conditional scale and bias driven by a conditioning field.
+
+    A FiLM layer (https://arxiv.org/abs/1709.07871) applied immediately after a
+    normalization layer, turning that norm into a conditional one, the same
+    construction ``ConditionalLayerNorm`` uses for the SFNO:
+    ``x -> x * (1 + W_scale(c)) + W_bias(c)``. Both convolutions are
+    zero-initialized, so an untrained model is exactly deterministic and any
+    dependence on ``c`` is learned. Nothing here is specific to noise: ``c`` is
+    any per-sample conditioning field carried on the ``Context``, and Samudra
+    passes the noise field today.
```
mcgibbon: Suggestion (optional): Remove extra context that exists only because this object previously stated Noise and we removed it.

```suggestion
    dependence on ``c`` is learned.
```

mcgibbon (approved):
LGTM, one suggestion.

fme/ace/models/ocean/m2lines/layers.py
elynnwu: Removed

fme/ace/registry/m2lines.py
mcgibbon: Would be interested in seeing if layer norm with no noise embed also does poorly.

As long as your runs with noise and instance norm "work", I think it's fine to use that, this is just intuition around what we think might cause an issue. But if it doesn't cause issues, by all means.

Another option is to not condition on the noise at all, but instead inject additional noise input channels. That's the approach Makani took.

PR #1465, opened 2026-08-27
Derived outputs PR plan

mcgibbon:

Changes:

- [ ] Tests added


jpdunc23:
Will need to learn more about the details, but so far this sounds like a good plan.

plan.md
```diff
@@ -0,0 +1,28 @@
...
+
+Let's talk specifically about surface precipitation. In our critical path model, the corrector uses the network output for PRATEsfc in combination with the water budget defined by each other moisture field, and determines the magnitude needed for PRATEsfc to close the water budget. In other words, the network output is really the pattern of precipitation, and the corrector is responsible for deriving the final value of PRATEsfc. Afterwards, the corrector also clips PRATEsfc so it is strictly positive - this is a correction applied to the predicted PRATEsfc value.
```
Arcomano1234, line 5: Claude speak, was this with the new opus rules? If so I'm surprised

mcgibbon: Lol I literally typed this entire document myself :P

plan.md
```diff
@@ -0,0 +1,28 @@
...
+- Values computed by the Module, but which are never optimized in the loss (computed-unoptimized, in this draft).
+- Values which are derived from Module-output values, and are optimized in the loss (derived-optimized, in this draft).
+
+Certain corrector features treat these two values as one identical value, with overwriting.
+
+This first slice will target one such feature - precipitation in the context of global-mean moisture conservation - to prove out the code framework. It will introduce a field `PRATEsfc_relative_pattern` which is output by the neural network when this type of correction is enabled, has no target data, will not be optimized in the loss. It should be reported in aggregators in the way any other derived value without target data is reported. `PRATEsfc` itself will be a value derived from this value, which is directly optimized in the loss despite not being a NN output. Because of this, there will be no corrector "delta" assigned to PRATEsfc - it is derived, not corrected. In this pass, the responsibility for deriving PRATEsfc will remain in the corrector. Any renaming of the corrector or other refactoring to make this clearer is left for future work.
```
Arcomano1234, line 21: I like this idea for another reason which is I always wanted pre-corrector fields plotted in wandb, so I fully support this structuring

mcgibbon: Yes I had argued pretty strongly against presenting the "correction" pre-correction fields since what we care about is the magnitude of the change (and that's hard to get by subtracting two full fields by eye), but for these "this is an input used for derivation" fields I also would like to see them as full fields.

pr-plan.md
```diff
@@ -0,0 +1,28 @@
...
+
+Let's talk specifically about surface precipitation. In our critical path model, the corrector uses the network output for PRATEsfc in combination with the water budget defined by each other moisture field, and determines the magnitude needed for PRATEsfc to close the water budget. In other words, the network output is really the pattern of precipitation, and the corrector is responsible for deriving the final value of PRATEsfc. Afterwards, the corrector also clips PRATEsfc so it is strictly positive - this is a correction applied to the predicted PRATEsfc value.
+
+Currently, there is no way to distinguish between these two actions applied to PRATEsfc. They both appear as a "delta", and it is not possible for example to regularize the magnitude of predicted negative precipitation without also regularizing the "derivation" of PRATEsfc. Applying pre-corrector optimization on the zero clipping may (or may not) make sense, but applying it to the derivation of PRATEsfc never would, as it would remove all gradient pathway to the budget non-closure in the other moisture fields.
```
jpdunc23: > Applying pre-corrector optimization on the zero clipping may (or may not) make sense

Based on how it behaves with the sea ice clipping, my guess is that pre-corrector optimization on the zero clipping will lead to small positive biases (drizzling bias). Corrector regularization may be more effective for this.

> but applying it to the derivation of `PRATEsfc` never would, as it would remove all gradient pathway to the budget non-closure in the other moisture fields

There is a case where the network `PRATEsfc` can be optimized directly against the target using pre-corrector optimization while still retaining gradient pathways to the budget non-closure in the other mositure fields: `"advection_and_precipitation"`.

Another option is to combine pre-corrector regularization with corrector regularization of the `PRATEsfc` delta, which would preserve gradients directly through the `"precipitation"` correction while still allowing the direct pre-corrector optim. This shows how these two methods are complementary.

mcgibbon: > Based on how it behaves with the sea ice clipping, my guess is that pre-corrector optimization on the zero clipping will lead to small positive biases (drizzling bias). Corrector regularization may be more effective for this.

Agreed.

> There is a case where the network PRATEsfc can be optimized directly against the target using pre-corrector optimization while still retaining gradient pathways to the budget non-closure in the other mositure fields: "advection_and_precipitation".
> 
> Another option is to combine pre-corrector regularization with corrector regularization of the PRATEsfc delta, which would preserve gradients directly through the "precipitation" correction while still allowing the direct pre-corrector optim. This shows how these two methods are complementary.

For now the focus in this plan is on correctly representing what the code actually does in our published (and planned to publish) models.

I think we should be open to pre-correction or corrector regularization of the PRATEsfc relative pattern. Before we focus implementation on it, we will need to know the advantage we're trying to gain by using pre-corrector optimization to bring the PRATEsfc relative pattern inline in magnitude with the full PRATEsfc field, and demonstrate that we gain that advantage in an experiment. These approaches significantly complicate the conceptual model of what our code is doing (making it harder for users to understand the model configuration both when setting yaml and when reading our papers), so it's important to motivate it with experimental benefits that support that added complexity.

mcgibbon: To repeat, I do agree this is a configuration we could try, and that if it solves a worthwhile issue, we should use it. I wanted to expand a bit on my motivation for this being unlikely (of course, nothing is guaranteed in these kind of intuitive ML things, so I can be wrong).

> There is a case where the network PRATEsfc can be optimized directly against the target using pre-corrector optimization while still retaining gradient pathways to the budget non-closure in the other mositure fields: "advection_and_precipitation".

1) Say we apply pre-corrector optimization to all moisture fields.

In this case, in an ideal world the model does a good job learning the closed budget (since each term targets a value which comes from a presumably-closed budget). However PRATEsfc is a dumping ground in our corrector for global mean errors in all of the other terms. Even in single-variable corrections we know we're biasing the results by not optimizing the post-corrector state, but it's greatly compounded here because of all budget terms putting their correction into PRATEsfc.

2) Say we apply pre-corrector optimization only to PRATEsfc, not to the other moisture fields.

In this case, the same issue as in (1) happens, except that the other budget terms aren't even constrained to mostly have a closed budget, because the corrector removes the unclosed portion from consideration in the loss, and PRATEsfc also does not include them in the loss thanks to pre-corrector optimization. The PRATEsfc errors can grow unbounded, because there's no gradient pathway for these budget errors to the loss. This was the case I was talking about when I wrote the comment.

I'll update the prose (after you have a chance to post other comments) to better describe the scenario(s).

plan.md
jpdunc23, line 21: "in the way any other derived value without target data is reported", what do you have in mind here? AFAIK all of our "derived" variables are reported against the corresponding target derivations, e.g. `"ocean_heat_content"` is a derived variable which has bias / RMSE / etc. metrics, but it isn't included in the optimization.

pr-plan.md
```diff
@@ -0,0 +1,28 @@
...
+- Values computed by the Module, but which are never optimized in the loss (computed-unoptimized, in this draft).
+- Values which are derived from Module-output values, and are optimized in the loss (derived-optimized, in this draft).
+
+Certain corrector features treat these two values as one identical value, with overwriting.
+
+This first slice will target one such feature - precipitation in the context of global-mean moisture conservation - to prove out the code framework. It will introduce a field `PRATEsfc_relative_pattern` which is output by the neural network when this type of correction is enabled, has no target data, will not be optimized in the loss. It should be reported in aggregators in the way any other derived value without target data is reported. `PRATEsfc` itself will be a value derived from this value, which is directly optimized in the loss despite not being a NN output. Because of this, there will be no corrector "delta" assigned to PRATEsfc - it is derived, not corrected. In this pass, the responsibility for deriving PRATEsfc will remain in the corrector. Any renaming of the corrector or other refactoring to make this clearer is left for future work.
```
jpdunc23: Blocker: `PRATEsfc_relative_pattern` where does this name come from? Who is responsible for it and its relationship to `PRATEsfc`?

pr-plan.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+# Separation of concerns
+
+This concern will live in the Corrector and SingleModuleStep. The Corrector will own all involved field names, exposing lists of field names as needed to the Step (e.g. so it can construct its own name lists, and including the output Packer). Objects above the Step shouldn’t require any changes.
```
jpdunc23: > The Corrector will own all involved field names, exposing lists of field names as needed to the Step

I think this is a significant task that warrants its own planning. In particular, it isn't clear to me how the correctors will expose the corrected names statically without a larger refactoring of the corrections / `AtmosphereData`-like objects.

pr-plan.md
jpdunc23: > Applying pre-corrector optimization on the zero clipping may (or may not) make sense, but applying it to the derivation of PRATEsfc never would, as it would remove all gradient pathway to the budget non-closure in the other moisture fields.

- Neither "pre-corrector optimization on the zero clipping" nor on "the derivation of PRATEsfc" is well-defined. By design pre-corrector optimization doesn't target any piece of the delta.

- The statement that pre-corrector optimization would never make sense in this context is unsupported. It "may not make sense" but whether it does or not would have to be checked empirically.

I've been trying to understand the scenario you described using a concrete example focusing just on the `"precipitation"` branch without worrying for now about the `"advection"` part of the correction for now.

Given `P0` is the network output we have
```
P1 = max(P0, 0)
alpha = (<E> - <dTWP/dt>) / <P1>
P2 = alpha * P1  # final output
```
We can re-derive this in terms of deltas after each step as
```
P2 = P0 + delta_force_positive + delta_budget
```
where
```
delta_force_positive = max(-P0, 0)
delta_budget = (alpha - 1) * P1
```

Define `R = <P1> - <E> + <dTWP/dt>` which is zero for a closed budget, and `R_true` the reference target budget non-closure (which might not be 0).

We can rewrite `alpha` and `delta_budget` as
```
alpha = (<E> - <dTWP/dt>) / <P1> = 1 - R / <P1>
delta_budget = - (R / <P1>) * P1
```
Define each generated output as truth plus error `e`:
```
P1 = P_true + e_P, E = E_true + e_E, TWP = TWP_true + e_W.
```
Then
```
R = R_true + <e_P> - <e_E> + <de_W/dt>
```
Say we use the MSE loss. Then optimization of MSE toward 0 implies that each `e` term goes toward zero and `R` goes toward `R_true`. In other words, pre-corrector optimization of `PRATEsfc` is aligned with closed budgets to the extent that the target data itself is closed (`R_true = <adv_tgt>`).

The full picture is a bit trickier, as you've alluded to, since MSE will tend to favor reducing pattern errors over global mean errors, and the looser global mean errors are exactly what end up in `R`. `alpha = 1 - R/<P1>` is never penalized in training with pre-corrector optimization.

I'll also note that even with post-corrector optimization nothing stops `<E>` from becoming biased if it reduces `P2`'s MSE without hurting `E` itself too much, so that case is also not straightforward.

jpdunc23 (commented):
Meta commentary on this planning process: I think a PR / `.md` is not a great format for feedback on this sort of high-level write-up. With Google Docs it's possible to highlight a few words and add a comment, whereas GitHub PR comments are per-line, which gives poorer localization of comments and places more burden on commenters to copy-paste specific pieces of a long one-line paragraph. The `.md` diff is also not a particularly nice reading experience, and the "rich diff" doesn't allow comments.

For formal PR review, I'd recommend that we draft a procedure for these types of PR planning workflows so that we can define a set of standards for review that both authors and reviewers understand. It would be worthwhile to develop a preset template that we can all agree on for what these documents must contain in terms of sections / details / etc.

plan.md
mcgibbon: Just meaning to say we already support model outputs which might not have target data (some derived outputs don't), they only appear in "gen" type metrics and don't have error metrics reported. So the existing behavior for "new" variables this process adds that don't necessarily have targets is already well-defined.

pr-plan.md
mcgibbon: Ah yes, I'll update this. It's very possible this should be updated in its own PR, meaning this is more of a "change plan" than a "PR plan" (will rename). But I'll do the implementation in one PR first, because I really need a specific case to write the code against. For what it's worth, I don't think this requires exporting corrected names at all - I think an api like `corrector.get_network_output_names(user_requested_output_names) -> list[str]` would do the trick.

The basic idea is that there should remain one source of truth for what gets updated, that we should isolate that to a specific level, and that this level should lie within the Corrector (more likely within AtmosphereData, like you point out).

This might mean a large refactor of AtmosphereData, or it might not (I don't expect it will). I'd suggest we leave that to the implementation stage, but please let me know any restrictions/concerns you have about refactoring that code, including if you'd be concerned about it being refactored (I think that would be fine, it's very internal to the corrector). I'd like this plan to encode what we're concerned about without being overly restrictive on the code solution. The code will still go through review.

pr-plan.md
mcgibbon: I updated the prose to remove the claim, instead focusing on these being two potential changes we might want to try with very different levels of conceptual complexity, which currently cannot be made independent of each other. It might be worth drawing this line to its conclusion (I think I'll make a Google doc), but it's besides the point for this plan.

Can we agree for now that:
1.  We don't have empirical evidence this is a configuration we should use.
2. We are capable of updating the code to support that configuration after this plan if and when we get that evidence, including by re-labelling this "derivation" as a "correction" when we decide based on evidence that we should do so.
3. That the plan reflects what our current models actually do, in a way that would avoid the interpretation in Will et al.

plan.md
```diff
@@ -0,0 +1,29 @@
...
+
+Currently, the corrector zero-clips the pattern field and then rescales it. With this refactor, it will be necessary to compute the derived PRATEsfc at the same time as zero clipping it, returning the delta that corresponds only to the zero clipping, since the magnitude of the rescaling depends on whether precipitation is being zero-clipped. Another option would be to zero clip the pattern field, but we need this clipped amount to be in precipitation units, not in relative units, so that corrector regularization doesn't just encourage the pattern to have arbitrarily small magnitude. We may go this route by clipping the pre-rescaling field and then rescaling the delta to precipitation units - this still requires we perform both operations in the same scope to make sure they're correctly aligned.
```
mcgibbon, line 27: Added this paragraph after realizing the current order is `clip -> rescale` in the correctors. Shouldn't be an issue to modify (result should still be epsilon-precision identical) but wanted to give a heads up.

plan.md
```diff
@@ -0,0 +1,29 @@
...
+- Values computed by the Module, but which are never optimized in the loss (computed-unoptimized, in this draft).
+- Values which are derived from Module-output values, and are optimized in the loss (derived-optimized, in this draft).
+
+Certain corrector features treat these two values as one identical value, with overwriting.
+
+This first slice will target one such feature - precipitation in the context of global-mean moisture conservation - to prove out the code framework. It will introduce a field `PRATEsfc_relative_pattern` which is output by the neural network when this type of correction is enabled, has no target data, will not be optimized in the loss. It should be reported in aggregators in the way any other derived value without target data is reported. `PRATEsfc` itself will be a value derived from this value, which is directly optimized in the loss despite not being a NN output. Because of this, there will be no corrector "delta" assigned to PRATEsfc - it is derived, not corrected. In this pass, the responsibility for deriving PRATEsfc will remain in the corrector. Any renaming of the corrector or other refactoring to make this clearer is left for future work.
```
jpdunc23, line 21: Blocker: Tying network output names to the corrector config. 

It would be helpful if you can provide specifics of exactly how `PRATEsfc_relative_pattern` would be created, tracked, and linked to `PRATEsfc`, and how this interacts with the `ATMOSPHERE_FIELD_NAME_PREFIXES` handling for the standard name "precipitation_rate".

plan.md
```diff
@@ -0,0 +1,29 @@
...
+
+If we later decided the network output should in fact be PRATEsfc and not just its relative pattern, the code would be updated accordingly so that it continues to reflect what our model actually does.
+
+# Plan
+
+We’d like to decouple these responsibilities by formalizing two classes of output variables:
+- Values computed by the Module, but which are never optimized in the loss (computed-unoptimized, in this draft).
+- Values which are derived from Module-output values, and are optimized in the loss (derived-optimized, in this draft).
```
jpdunc23, line 17: Blocker: It isn't clear how the `computed-unoptimized` set is determined.

Is this hard-coded? Corrector config dependent? Who decides what goes in and what stays out?

plan.md
jpdunc23, line 21: > Because of this, there will be no corrector "delta" assigned to PRATEsfc - it is derived, not corrected.

Blocker: No details on how exactly `PRATEsfc` will be derived from `PRATEsfc_relative_pattern`.

Is this just the usual `"precipitation"` rescaling by the normalized non-closure? If so, there are legitimate reasons to optimize `delta` (discussed here: https://github.com/ai2cm/ace/pull/1465#discussion_r3911160521), so we should avoid putting that delta out-of-reach of the optimization.

jpdunc23 (commented):
Thank you for writing this proposal which provides a nice launching point for further discussion of the correctors. I'm starting to gain some better clarity on the correctors' role in the optimization.

I added a couple more comments that I consider "blocking" at this stage of planning.

IMO, we should hold off on code changes to the correctors and other post-module output modifications in the step until at least after the topic talk. In the meantime, I do think its important to continue these discussions and documented planning around the correctors.

mcgibbon:
> IMO, we should hold off on code changes to the correctors and other post-module output modifications in the step until at least after the topic talk.

I'm good with that. Having more and cleaner experimental comparisons will help inform some of these choices. Perhaps I can launch a 4-degree ensemble for the atmosphere cases, especially if I can reproduce the TWP bias of my current stochastic runs at that resolution and there's a chance some of these changes might improve it.

plan.md
mcgibbon: > Blocker: Tying network output names to the corrector config.

Can you describe what makes this a blocker? It would be helpful to have a specific `Issue:` or `Suggestion:` to understand what's blocking.

The corrector config is presently affecting the meaning of network outputs, whether they're labelled correctly or not. When we turn on one of the "precipitation" moisture budget corrections, the network output no longer represents PRATEsfc, only its spatial pattern.

The idea of this plan is for the variable labels to accurately reflect what the network in fact outputs, and surface it directly with that label in the step output (where currently it's hidden). This necessarily requires defining the network output names based on the corrector config.

plan.md
jpdunc23: > we already support model outputs which might not have target data (some derived outputs don't), they only appear in "gen" type metrics and don't have error metrics reported

I'm not aware of any but maybe I'm just not using configs that include them. Can you give an example of such a variable?

jpdunc23:
> I'm good with that. Having more and cleaner experimental comparisons will help inform some of these choices. Perhaps I can launch a 4-degree ensemble for the atmosphere cases, especially if I can reproduce the TWP bias of my current stochastic runs at that resolution and there's a chance some of these changes might improve it.

Sounds good! Looking at other variables involved in the precip correction would be useful as well, in particular `LHTFLsfc`, `specific_total_water`, and `PRESsfc`.

plan.md
mcgibbon: I can add some more details, but PRATEsfc will be PRATEsfc_relative_pattern multiplied by the factor it needs to be multiplied by to close the moisture budget, accounting for whether or not we know zero-clipping is going to happen.

> If so, there are legitimate reasons to optimize delta (discussed here: #1465 (comment)), so we should avoid putting that delta out-of-reach of the optimization.

How about including a module constant flag that enables the previous behaviour? Then the experiment only needs to flip that flag in the code. It's important that we validate experimentally whether that behaviour improves something before we define the configuration or behaviour based on it. I _suspect_ we should not change our models to do that, that it will not do better, but evidence would speak for itself.

I'm not saying this arbitrarily, or even because it's my inclined choice - over the past seven years I have often argued to add experimental features to our code base or our mainline models that whose code I had already written, and that I believed I had a good argument for why they would improve our model's skill and/or ability to generalize. If these significantly complicated the code or its config, and/or were hard to understand, they were pushed back on and not merged unless I could show experimentally that the model improves enough to warrant the change (sometimes even improving validation loss wasn't enough, if climate didn't improve!). This happened to me _a lot_ less over time as I started focusing on features based on specific deficits in our model, instead of having a "good idea" and then looking for something it would improve. 

Whether this should change is something we would need to discuss as a group. For example, separate mainline and experimental configuration classes could enable more experimental feature support on main. For now I hope a boolean module constant enabling this experiment is enough that this isn't a blocker.

plan.md
mcgibbon: For now, it includes PRATEsfc_relative_pattern if an advection config that derives PRATEsfc is enabled (one that has "precipitation" in the name).

It would be nice if this is owned only by the code that does the change, e.g. `_force_conserve_moisture` which currently directly calls `gen.set_precipitation_rate`, a line that needs to change. That function would need to become a class so it has this metadata. I would avoid proactively refactoring the entire corrector in this way, for this stage.

I'll add this to the plan.

plan.md
mcgibbon: > It would be helpful if you can provide specifics of exactly how PRATEsfc_relative_pattern would be created, tracked, and linked to PRATEsfc

Added a section.

> how this interacts with the ATMOSPHERE_FIELD_NAME_PREFIXES handling for the standard name "precipitation_rate".

Added a sentence with the restriction of no duplicate ownership, and how I think that will probably be achieved in the implementation.

plan.md
mcgibbon: Huh, it looks like maybe we no longer have such variables. I could have sworn we used to in earlier days of ace. Will add a specification about unmatched field behavior then (generally speaking, it should show up in "gen" type fields but not in error metrics) and check what those aggregators actually are.

pr-plan.md
mcgibbon: > In other words, pre-corrector optimization of PRATEsfc is aligned with closed budgets to the extent that the target data itself is closed (R_true = <adv_tgt>).

Following this through, I agree. I confused myself by thinking that TWP and E are also corrected by the corrector, meaning they have pre-corrector and post-corrector values - they do not.

PR #1466, opened 2026-08-28
Return frozenset from output_names to fix per-channel loss label shuffle on resume

### LINKED ISSUE #579: Unordered Lists cast to/from Set in stepper code

mcgibbon:
Per-channel loss values logged to wandb (train/mean/loss/<varname>)
were numerically wrong in every multi-GPU run and additionally
mislabeled across Beaker preemption/resume cycles.

**Multi-GPU (every run):** `PerChannelLossAggregator.get_logs`
iterates `self._weighted_sums` in insertion order, calling
`dist.reduce_mean` (an `all_reduce` collective matched by call
position) per key. Different PYTHONHASHSEED across ranks means rank
0's reduce for `air_temperature_0` paired with rank 1's reduce for
`surface_pressure` — the logged per-channel losses were means over
mixed variables. The sibling `LossAggregator.get_logs` already guards
this (gathers and sorts the key universe before reducing);
`PerChannelLossAggregator` was missing it.

**Preemption (label shuffling):** `output_names` and siblings returned
`list(set(...))`, whose iteration order depends on PYTHONHASHSEED.
Since `loss_names` derived from `output_names`, the loss Packer
received a different channel ordering in each new process, swapping
per-channel labels while leaving the total loss correct.

The fix:
1. `PerChannelLossAggregator.get_logs` now iterates
   `sorted(self._weighted_sums.items())` so every rank issues the same
   collectives.
2. `output_names`, `input_names`, `next_step_input_names`, and
   `prognostic_names` return `frozenset[str]`, making unorderedness
   explicit. `loss_names` returns `sorted(self.output_names)` for the
   Packer. Downstream signatures widened to `Collection[str]` where
   only membership/iteration is needed.

Changes:
- `PerChannelLossAggregator.get_logs` — sorted iteration over `_weighted_sums`
- `StepConfigABC.output_names`, `.input_names`, `.next_step_input_names`, `.prognostic_names` → `frozenset[str]`
- `StepConfigABC.loss_names` → `sorted(self.output_names)` (deterministic across process restarts)
- `SingleModuleStepConfig`, `SecondaryModuleStepConfig`, `FCN3StepConfig`, `SeparateRadiationStepConfig`, `MultiCallStepConfig` — same pattern
- `StepABC`, `StepSelector` passthroughs — updated return types
- `StepperConfig`, `Stepper` (ace stepper) passthroughs — updated return types
- `MultiCallConfig.validate`, `OceanConfig.build`, `add_names`, `InitialConditionRequirements`, `PrognosticStateDataRequirements`, `CoupledInitialConditionConfig.get_initial_condition` — widened to `Collection[str]`
- `step_with_adjustments` `prognostic_names` parameter — `frozenset[str]`

- [x] Tests added

Resolves #579

### DIFF
--- a/fme/ace/step/fcn3.py
+++ b/fme/ace/step/fcn3.py
@@ -230,7 +230,8 @@ class FCN3StepConfig(StepConfigABC):
             extra_residual_scaled_names = []
         return self.normalization.get_loss_normalizer(
             names=self._normalize_names + extra_names,
-            residual_scaled_names=self.prognostic_names + extra_residual_scaled_names,
+            residual_scaled_names=sorted(self.prognostic_names)
+            + extra_residual_scaled_names,
         )
 
     @classmethod
@@ -241,46 +242,45 @@ class FCN3StepConfig(StepConfigABC):
         )
 
     @property
-    def _normalize_names(self):
+    def _normalize_names(self) -> list[str]:
         """Names of variables which require normalization. I.e. inputs/outputs."""
-        return list(set(self.in_names).union(self.out_names))
+        return sorted(set(self.in_names) | set(self.out_names))
 
     @property
-    def input_names(self) -> list[str]:
+    def input_names(self) -> frozenset[str]:
         """
         Names of variables required as inputs to `step`,
         either in `input` or `next_step_input_data`.
         """
         if self.ocean is None:
-            return self.in_names
+            return frozenset(self.in_names)
         else:
-            return list(set(self.in_names).union(self.ocean.forcing_names))
+            return frozenset(self.in_names) | frozenset(self.ocean.forcing_names)
 
     def get_next_step_forcing_names(self) -> list[str]:
         """Names of input-only variables which come from the output timestep."""
         return self.next_step_forcing_names
 
     @property
-    def diagnostic_names(self) -> list[str]:
+    def diagnostic_names(self) -> frozenset[str]:
         """Names of variables which are outputs only."""
-        return []  # not currently supported
+        return frozenset()  # not currently supported
 
     @property
-    def output_names(self) -> list[str]:
-        return self.out_names
+    def output_names(self) -> frozenset[str]:
+        return frozenset(self.out_names)
 
     @property
-    def next_step_input_names(self) -> list[str]:
+    def next_step_input_names(self) -> frozenset[str]:
         """Names of variables provided in next_step_input_data."""
-        result = set(self.input_names).difference(self.output_names)
+        result = self.input_names - self.output_names
         if self.ocean is not None:
-            result = result.union(self.ocean.forcing_names)
-        result = result.union(self.prescribed_prognostic_names)
-        return list(result)
+            result = result | frozenset(self.ocean.forcing_names)
+        return result | frozenset(self.prescribed_prognostic_names)
 
     @property
     def loss_names(self) -> list[str]:
-        return self.output_names
+        return sorted(self.output_names)
 
     def replace_ocean(self, ocean: OceanConfig | None):
         """
--- a/fme/ace/step/fcn3.py
+++ b/fme/ace/step/fcn3.py
@@ -229,7 +229,7 @@ class FCN3StepConfig(StepConfigABC):
         if extra_residual_scaled_names is None:
             extra_residual_scaled_names = []
         return self.normalization.get_loss_normalizer(
-            names=self._normalize_names + extra_names,
+            names=sorted(self._normalize_names) + extra_names,
             residual_scaled_names=sorted(self.prognostic_names)
             + extra_residual_scaled_names,
         )
@@ -242,9 +242,9 @@ class FCN3StepConfig(StepConfigABC):
         )
 
     @property
-    def _normalize_names(self) -> list[str]:
+    def _normalize_names(self) -> frozenset[str]:
         """Names of variables which require normalization. I.e. inputs/outputs."""
-        return sorted(set(self.in_names) | set(self.out_names))
+        return frozenset(set(self.in_names).union(self.out_names))
 
     @property
     def input_names(self) -> frozenset[str]:
@@ -255,7 +255,7 @@ class FCN3StepConfig(StepConfigABC):
         if self.ocean is None:
             return frozenset(self.in_names)
         else:
-            return frozenset(self.in_names) | frozenset(self.ocean.forcing_names)
+            return frozenset(set(self.in_names).union(self.ocean.forcing_names))
 
     def get_next_step_forcing_names(self) -> list[str]:
         """Names of input-only variables which come from the output timestep."""
@@ -273,10 +273,11 @@ class FCN3StepConfig(StepConfigABC):
     @property
     def next_step_input_names(self) -> frozenset[str]:
         """Names of variables provided in next_step_input_data."""
-        result = self.input_names - self.output_names
+        result = set(self.input_names).difference(self.output_names)
         if self.ocean is not None:
-            result = result | frozenset(self.ocean.forcing_names)
-        return result | frozenset(self.prescribed_prognostic_names)
+            result = result.union(self.ocean.forcing_names)
+        result = result.union(self.prescribed_prognostic_names)
+        return frozenset(result)
 
     @property
     def loss_names(self) -> list[str]:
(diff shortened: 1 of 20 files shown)


mcgibbon:
This is from noticing per-channel losses jump drastically between epochs, e.g. in https://wandb.ai/ai2cm/ace/runs/r95iprjt.

fme/ace/step/fcn3.py
```diff
@@ -241,46 +242,45 @@ def from_state(cls, state) -> "FCN3StepConfig":
...
     @property
-    def _normalize_names(self):
+    def _normalize_names(self) -> list[str]:
         """Names of variables which require normalization. I.e. inputs/outputs."""
-        return list(set(self.in_names).union(self.out_names))
+        return sorted(set(self.in_names) | set(self.out_names))
 
     @property
-    def input_names(self) -> list[str]:
+    def input_names(self) -> frozenset[str]:
```
mcgibbon, line 250: Not promising sorted-ness makes it clear no particular sorting is guaranteed (that's the case because this result is a set union). Then the downstream caller can sorted() if they like, making it clear they have an alphabetical sort rather than a channel sort, for example.

fme/core/step/step.py
```diff
@@ -46,29 +46,29 @@ def n_ic_timesteps(self) -> int:
...
         Names of variables required in next_step_input_data for .step.
         """
         pass
 
     @property
     @final
-    def prognostic_names(self) -> list[str]:
-        return list(set(self.input_names).intersection(self.output_names))
+    def prognostic_names(self) -> frozenset[str]:
+        return self.input_names & self.output_names
```
mcgibbon: I know this is probably the same logic, but please minimize changes by using `set(self.input_names).intersection(self.output_names)`.

fme/coupled/stepper.py
```diff
@@ -500,14 +500,12 @@ def shared_forcing_exogenous_names(self) -> list[str]:
...
-                + self._atmosphere_forcing_exogenous_names
-            )
+        atmosphere_names = sorted(
+            self.atmosphere.stepper.output_names
+            | set(self._atmosphere_forcing_exogenous_names)
         )
-        ocean_names = list(
-            set(self.ocean.stepper.output_names + self._ocean_forcing_exogenous_names)
+        ocean_names = sorted(
+            self.ocean.stepper.output_names | set(self._ocean_forcing_exogenous_names)
```
mcgibbon: Please minimize code changes.

fme/core/step/secondary_module.py
```diff
@@ -161,57 +162,55 @@ def from_state(cls, state) -> "SecondaryModuleStepConfig":
...
             if self.secondary_decoder is not None
             else []
         )
-        return list(
-            set(self.out_names)
-            .union(secondary_decoder_names)
-            .union(self.secondary_out_names)
-            .union(self.secondary_residual_out_names)
+        return (
+            frozenset(self.out_names)
```
mcgibbon: Please keep the earlier style, here and throughout (I'm going to stop commenting this).

fme/ace/step/fcn3.py
mcgibbon: Claude: Fixed — `_normalize_names` now returns `frozenset[str]` throughout. Callers sort explicitly when they need a list (`get_loss_normalizer`, `get_network_normalizer`).

fme/core/step/step.py
mcgibbon: Claude: Fixed — reverted to `frozenset(set(self.input_names).intersection(self.output_names))`.

fme/coupled/stepper.py
mcgibbon: Claude: Fixed — switched to `set(...).union(...)` style, keeping the original code structure.

fme/core/step/secondary_module.py
mcgibbon: Claude: Fixed throughout — all set operations now use `.union()`/`.intersection()`/`.difference()` method-call style. The outer wrapper changes from `list()` to `frozenset()`, keeping the original code structure.

Arcomano1234 (approved):
This looks ready to go. However, while having claude give this a review as well. It seems to have identified a bug that at least I wasn't aware of with the per channel loss which is 

> N1 — Should fix: an empty data-parallel rank deadlocks this function. Verified on 2 ranks (FME_DISTRIBUTED_BACKEND=torch, rank 1 records nothing): rank 1 printed logs={}, passed in 0.02s and exited; rank 0 blocked in all_reduce and the job hung until I killed it ~5 minutes later. The local sort fixes call order but not key-set disagreement — the sibling PerStepLossAggregator.get_logs gathers a global key universe for exactly this reason ([loss_metrics.py:26-32](vscode-webview://049reeqlt45jfn5i1hh2t4kqcjc7evjsrlbp0smtq54v0m18uroh/fme/ace/aggregator/loss_metrics.py#L26-L32): "Ranks may record different keys") and carries test_per_step_loss_aggregator_mismatched_ranks(empty_rank). Both aggregators are fed from the same one_step/main.py path, so the reachability argument that motivated the sibling's guard applies here too.
N2 — Should fix: cross-rank means aren't count-weighted, so masking corrupts them. Verified on 2 ranks — rank 0 records loss=1.0, count=10, rank 1 records loss=3.0, count=2; logged value is 2.0 where the count-weighted answer is 1.333. reduce_mean divides by total_data_parallel_ranks regardless of counts ([model_torch_distributed.py:277-279](vscode-webview://049reeqlt45jfn5i1hh2t4kqcjc7evjsrlbp0smtq54v0m18uroh/fme/core/distributed/model_torch_distributed.py#L277-L279)), and a channel fully masked on one rank contributes a hard 0 via the else ws branch. The class docstring promises "masked-out samples never dilute the aggregated mean" — true across batches, false across ranks; the sibling's docstring states the intended contract outright ("Means must be count-weighted over the recording ranks"). Both are pre-existing, but they're the same "per-channel loss is numerically wrong in multi-GPU" family the PR is fixing, and per-sample variable masking is exactly the feature that trips it.

To avoid scope creep in this PR I dont think we should address this but I was curious if this is a bug you think we might hit in practice, I kinda doubt it.

mcgibbon:
> This looks ready to go. However, while having claude give this a review as well. It seems to have identified a bug that at least I wasn't aware of with the per channel loss which is
> 
> > N1 — Should fix: an empty data-parallel rank deadlocks this function. Verified on 2 ranks (FME_DISTRIBUTED_BACKEND=torch, rank 1 records nothing): rank 1 printed logs={}, passed in 0.02s and exited; rank 0 blocked in all_reduce and the job hung until I killed it ~5 minutes later. The local sort fixes call order but not key-set disagreement — the sibling PerStepLossAggregator.get_logs gathers a global key universe for exactly this reason (loss_metrics.py:26-32: "Ranks may record different keys") and carries test_per_step_loss_aggregator_mismatched_ranks(empty_rank). Both aggregators are fed from the same one_step/main.py path, so the reachability argument that motivated the sibling's guard applies here too.
> > N2 — Should fix: cross-rank means aren't count-weighted, so masking corrupts them. Verified on 2 ranks — rank 0 records loss=1.0, count=10, rank 1 records loss=3.0, count=2; logged value is 2.0 where the count-weighted answer is 1.333. reduce_mean divides by total_data_parallel_ranks regardless of counts (model_torch_distributed.py:277-279), and a channel fully masked on one rank contributes a hard 0 via the else ws branch. The class docstring promises "masked-out samples never dilute the aggregated mean" — true across batches, false across ranks; the sibling's docstring states the intended contract outright ("Means must be count-weighted over the recording ranks"). Both are pre-existing, but they're the same "per-channel loss is numerically wrong in multi-GPU" family the PR is fixing, and per-sample variable masking is exactly the feature that trips it.
> 
> To avoid scope creep in this PR I dont think we should address this but I was curious if this is a bug you think we might hit in practice, I kinda doubt it.

I don't think this is a bug we should fix, my understanding is the code is expected to crash when we run on a number of GPUs that doesn't evenly divide the number of batches. If we really want to put in the extra effort, we should raise an exception in Distributed.local_batch_size, e.g. if that function is ever going to return zero. But I don't think someone would complain in the case that they launch an 8-gpu job with a batch size of 4 and the code crashed with a strange error.

PR #1472, opened 2026-08-31
Support monthly mean output in zarr format

### LINKED ISSUE #1469: Monthly coarsening of time coordinates in zarr writes
Support writing monthly zarr output.

elynnwu:
Adds a zarr-backed monthly-mean writer (`MonthlyZarrWriter`) so that inference runs can produce monthly aggregates in zarr stores alongside (or instead of) the existing netCDF path. The writer reuses the core `ZarrWriter` (via a new `ZarrWriter.read_batch` method) for store construction and I/O, and `MonthlyAggregation` (extracted from the existing `MonthlyDataWriter`) for the running-mean accumulation and month indexing. The month-axis origin is fixed at construction from `IC + timestep`, so `n_months` is known before any data arrives and the pre-allocation bounds check is a real guard. Config validation rejects `separate_ensemble_members` (the separate-IC zarr writer is a different class that doesn't support monthly coarsening) and `overwrite_check` (meaningless for a writer that overwrites the months each batch touches by design). A month no sample reaches stores mean `0.0` and count `0`, matching the existing netCDF writer's behaviour for short samples on a shared lead-time axis. The legacy `save_monthly_files` path (`PairedMonthlyDataWriter`, `main.py`) stays netCDF-only.

Changes:
- `fme/ace/inference/data_writer/monthly_aggregation.py` — new module extracting `MonthlyAggregation` and `MonthIndexer` from `MonthlyDataWriter`, with `get_valid_times` folding in the former standalone `get_days_since_reference`
- `fme/ace/inference/data_writer/monthly.py` — `MonthlyDataWriter` now delegates to `MonthlyAggregation`/`MonthIndexer`; `MonthIndexer` takes the first output time at construction (no mutable "first call fixes the origin" state); `timestep` forwarded from `PairedMonthlyDataWriter`
- `fme/ace/inference/data_writer/monthly_zarr.py` — `MonthlyZarrWriter` backed by `ZarrWriter` and `MonthlyAggregation`
- `fme/core/writer.py` — `ZarrWriter.read_batch` added for read-modify-write; tested in `test_writer.py`
- `fme/ace/inference/data_writer/file_writer.py` — `FileWriterConfig.__post_init__` validates zarr + monthly; `build` wires up `MonthlyZarrWriter`
- `fme/ace/inference/data_writer/test_monthly_zarr.py` — end-to-end tests including a netCDF-comparison test and a cross-month-boundary IC test

Example config:

```yaml
data_writer:
  save_monthly_files: false        # the legacy netCDF monthly path
  files:
    - label: monthly_mean
      names: [surface_temperature, PRATEsfc]   # omit to save everything
      save_reference: false        # true in an evaluator run, for a _target store too
      time_coarsen:
        method: monthly_mean
      format:
        name: zarr
```

- [x] Tests added

Resolves #1469


elynnwu:
Claude: Pre-review agent — independent pre-review, run without the authoring session's context. Fixes noted below are applied in e069d8d.

## Audit passes

**1. Silent-failure pass — ⚠️ one finding, fixed.**
`ZarrWriterConfig.overwrite_check` was accepted and silently dropped on the monthly branch of `FileWriterConfig.build` (the raw-zarr branch forwards it). It has no meaning for a writer that overwrites the months each batch touches by design, so the default `False` was harmless but an explicit `true` was silence. `__post_init__` now raises `NotImplementedError` for that combination, with a test. No back-compat risk: zarr + monthly was rejected outright before this PR.

Other branches checked and clear: `_validate_chunks` raises rather than quietly re-chunking; the pre-allocation bound raises with the offending month range rather than truncating; `MonthIndexer.init_years/init_months` raise if read before an origin exists; `flush`/`finalize` are genuine no-ops (every `append_batch` writes through) and now say so. A month no sample reaches stores mean `0.0` / count `0` rather than NaN, because `add_data` folds into the stored value and a NaN seed would poison it — that matches what the netCDF writer already does for a short sample on a shared lead-time axis, so it is parity, not a new trap. It is now stated in the PR description.

**2. Behavior-change pass on existing code paths — ✅.**
The `monthly.py` extraction is behavior-preserving. `MonthIndexer.month_indices` reproduces `_get_month_indices` exactly, including fixing the origin from `[:, 0]` of the first batch and the `-1` sentinel; `get_valid_times` is `get_days_since_reference(...) + 14` with the same 1970-01-01 reference taken from the variable's calendar. Both moved functions are byte-identical. `MonthlyDataWriter`'s remaining call order is unchanged, so the origin is still fixed before `_extend_valid_time` reads it. No log lines, defaults, or return types moved. The one non-mechanical edit I made: the class docstring claimed it writes "monthly total data... divide the total by the counts", but `add_data` has always stored a running mean — corrected, since a reviewer reading the refactor would otherwise trip on it.

**3. Differential-test pass — ⚠️ one weak test, strengthened.**
`test_monthly_zarr.py`'s `_batch` helper built data constant over lat/lon, so `test_monthly_zarr_writer_matches_netcdf_writer` — the strongest test here — compared the two writers on values that carried no spatial signal. It now multiplies by a varying spatial pattern (1.0 at lat=0/lon=0, so the hand-computed expectations still read the same), and the accumulation test asserts an off-corner point. The `MonthIndexer` tests are properly differential already: two samples with origins in different calendar months, so the same calendar month maps to different indices per sample, and `n_months_through` is checked where the max is not attained by sample 0.

**4. Sibling / mirror coverage — ✅.**
The legacy `DataWriterConfig.save_monthly_files` path (`PairedMonthlyDataWriter`, `main.py`) stays netCDF-only; that is the right scope for #1469, and it is now stated in the description. `fme/coupled` has no data-writer mirror to update. `MonthlyZarrWriter` correctly skips `ak`/`bk` the same way `MonthlyDataWriter` does, rather than following `ZarrWriterAdapter`'s nondim-coord handling.

**5. Config back-compat — ✅.** No checkpoint-persisted state. Every previously-valid YAML still parses with the same behavior; the config surface only widens (a combination that used to raise now builds).

**6. Split heuristic — ✅.** 8 files, one refactor commit and one feature commit, in that order. Reviewing `0337c6d` first (pure extraction) then `a6359ff` is the right reading order and works well.

**7. Abstract-or-final / inheritance depth — ✅.** No inheritance introduced.

**8. AI-agent artifact pass — ✅ after one nit.** No `Any`, no `isinstance` + `type: ignore`, no in-function imports, no stale comments. The store title label was derived with a hand-rolled `rstrip`/`rsplit`; switched to `os.path.basename(path).removesuffix(".zarr")` to match `ZarrWriterAdapter`.

**9. Builder pattern — ✅.** `MonthlyZarrWriter` receives explicit values, never a config. `FileWriterConfig.build` reads `self.format.chunks` / `self.format.overwrite_check`, i.e. a config reading a sibling config's fields — but that is the pre-existing pattern of the raw-zarr branch three lines below, not something this PR introduces.

## Reviewer focus

**The lead-time pre-allocation is exact, and I verified it rather than reasoned about it.** `MonthIndexer.n_months_through` was the main correctness risk, since an undershoot would abort a long run mid-flight. I fuzzed it over 1600 random schedules — 1/2/4 initial conditions in different calendar months, timesteps from 1 h to 5 days, 1–4000 steps, across `proleptic_gregorian`, `noleap`, `360_day` and `julian`, under both the production convention (the deriver drops the IC steps, so the writer sees `IC + timestep` through `IC + n_forward_steps * timestep`) and the unit-test convention (times start at the IC) — against the month span the writer actually ends up covering. Zero mismatches in every combination. The bound is tight, not merely safe, so no spurious trailing month appears in a normal run. It also degrades safely: an undershoot would raise with the offending range, never silently truncate.

Two things worth a human eye, neither blocking:

1. **The variable set is frozen at the first batch.** `_initialize_store` creates arrays for the names in batch 1; a name appearing only in a later batch hits a bare `KeyError` from `self._store[name]`, where `MonthlyDataWriter` would create it lazily. Not reachable from the inference loop, where every window carries the same keys — flagging it as a deliberate difference rather than an oversight.
2. **No healpix coverage.** The `face`-in-coords branch of `_initialize_store` is untested — but so is the identical branch in `MonthlyDataWriter`, so this is a pre-existing gap the PR mirrors rather than widens.

## Verdict

Ready for review, after the fixes applied in e069d8d.

The PR is well structured: the extraction commit is genuinely mechanical, the feature commit is self-contained, and the netCDF-comparison test is the right way to pin the new writer to the old one's semantics. `pre-commit run` (ruff, ruff-format, mypy) is clean and `fme/ace/inference` is green at 181 passed.

fme/ace/inference/data_writer/monthly.py
brianhenn: preexisting but can you remove the stale n_months arg in the docstring here?

fme/ace/inference/data_writer/file_writer.py
```diff
@@ -228,8 +229,8 @@ class FileWriterConfig:
         format: Configuration for the output format (i.e. netCDF or zarr).
         separate_ensemble_members: Option to write ensemble members to separate files.
             In this case, time is a datetime coordinate. Only supported when using zarr
-            format. Filenames will have the suffix `_ic{member_index}` appended before
-            the file extension.
+            format without monthly coarsening. Filenames will have the suffix
```
brianhenn, line 232: (comment) It's maybe unfortunate that the separate-IC zarr writer is a separate class than the multi-IC zarr writer, because if they were the same (with an option to write either way) it would have been straightforward to make monthly coarsening available for both paths. Instead we need to add this new rule.

fme/ace/inference/data_writer/monthly.py
```diff
@@ -88,9 +92,11 @@ def finalize(self):
 
 class MonthlyDataWriter:
     """
-    Write monthly total data and sample counts to a netCDF file.
```
brianhenn, line 91: Nice to catch this.

fme/ace/inference/data_writer/monthly_aggregation.py
```diff
@@ -0,0 +1,232 @@
...
+
+def find_boundary(month_array, start_month) -> int:
+    """
+    Assuming month_array is an ordered array of months,
+    find the index of the first month that is not start_month.
+    """
+    return np.searchsorted(month_array, start_month, side="right")
+
+
+def get_days_since_reference(
```
brianhenn: (nit): While we're refactoring this function, I always found it very tightly coupled to the `get_valid_times` routine, which is the only thing that calls it. Maybe just fold it into get_valid_times?

fme/ace/inference/data_writer/monthly_zarr.py
```diff
@@ -0,0 +1,288 @@
...
+                (start time for each inference run).
+            n_timesteps: Total number of inference forward steps, used to size
+                the store's lead-time axis.
+            timestep: The time delta between each timestep.
+            save_names: Names of variables to save. If None, all predicted
+                variables will be saved.
+            variable_metadata: Metadata for each variable to be written.
+            coords: Coordinate data to be written to the store.
+            dataset_metadata: Metadata for the dataset.
+            chunks: Optional mapping of dimension name to chunk size. Omitted
```
brianhenn, line 78: (nit): Maybe worth noting that  in the time dim there's no sharding in addition to chunk size = 1 (month). These seem appropriate for the monthly output.

fme/ace/inference/data_writer/monthly_zarr.py
```diff
@@ -0,0 +1,288 @@
...
+            )
+        n_times_data = list(data.values())[0].shape[1]
+        n_times_time = batch_time.sizes[LEAD_TIME_DIM]
+        if n_times_data != n_times_time:
+            raise ValueError(
+                f"Batch time dimension mismatch, data has {n_times_data} times "
+                f"and batch_time has {n_times_time} times."
+            )
+
+        months = self._month_indexer.month_indices(batch_time)
```
brianhenn, line 171: I'm not following why the `self._n_months` calculation code has to be in `append_batch` (ie see a batch of real data). Isn't this known on init? Would be clearer to have it there if so.

fme/ace/inference/data_writer/monthly_zarr.py
```diff
@@ -0,0 +1,288 @@
...
+            time + n_timesteps * timestep for time in initial_condition_times
+        ]
+        self._chunks = _validate_chunks(chunks)
+
+        label = os.path.basename(path).removesuffix(".zarr")
+        dataset_metadata = copy.copy(dataset_metadata)
+        dataset_metadata.title = f"ACE {label.replace('_', ' ')} data file"
+        self._dataset_metadata = dataset_metadata.as_flat_str_dict()
+
+        # spatial dims and the month origin aren't known until the first batch
```
brianhenn: ```# spatial dims and the month origin aren't known until the first batch```
See comment below in append_batch, not understanding this comment

fme/ace/inference/data_writer/monthly_zarr.py
```diff
@@ -0,0 +1,288 @@
...
+        self, *data_varnames: Iterable[str]
+    ) -> Iterable[str]:
+        return get_all_names(*data_varnames, allowlist=self._save_names)
+
+    def _chunks_for(self, dims: Sequence[str], sizes: Sequence[int]) -> tuple[int, ...]:
+        return tuple(
+            self._chunks.get(dim, size) for dim, size in zip(dims, sizes, strict=True)
+        )
+
+    def _initialize_store(self, data: Mapping[str, torch.Tensor], n_months: int):
```
brianhenn: (suggestion): Is there a reason (time coordinate incompatibility?) that this function doesn't use `fme.core.writer.ZarrWriter` like the raw zarr writer does? Might save a lot of repeated code.

brianhenn (commented):
Thanks, this looks good overall. Most of my suggestions are nits, with the exception of trying to reuse the existing core zarr writer if possible.

fme/ace/inference/data_writer/monthly.py
elynnwu: Done in fbae63d.

fme/ace/inference/data_writer/file_writer.py
elynnwu: Agreed, and that asymmetry is why the new rule reads as an arbitrary exclusion rather than a consequence of anything. Unifying them, so that `separate_ensemble_members` is an option on one writer rather than a different class, would make monthly coarsening fall out for both paths for free — but it reshapes the raw zarr path that this PR otherwise leaves alone, so I'd rather not fold it in here. Happy to open a follow-up issue to track the merge if you think it's worth it.

fme/ace/inference/data_writer/monthly_aggregation.py
elynnwu: Claude: Folded in, in fbae63d. It really was a one-caller seam — the only thing `get_valid_times` added on top was the `+ 14` and the 1970-01-01 reference date.

The two direct unit tests move over to `get_valid_times`. They were written against a 2020-01-01 reference so the expectations read as days into the year (`days[0, 1] == 31`), so there's a small test-local helper that re-expresses the returned valid times as days since the first year, removing both the 14-day offset and the 1970→2020 shift. The assertions themselves are unchanged, so the month-arithmetic and `month_offset` coverage is the same as before.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: Claude: Noted in the `chunks` docstring in fbae63d: the store is unsharded, and sample and lead time are always chunked as 1.

Both are deliberate rather than incidental. Each `append_batch` reads back and rewrites exactly the (sample, month) cells the batch touches, so a chunk spanning more than one month or sample would mean read-modify-write amplification on every batch, and sharding would put unrelated months in the same storage object. At monthly resolution the array is small enough that the resulting chunk count isn't a problem.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: Claude: The month count depends on the origin of the lead-time axis, and the origin is the calendar month of the run's first **output** time, not of the initial condition. In production the deriver drops the IC step, so the first time this writer sees is `IC + timestep`, which can land in the next calendar month — an IC on Jan 31 with a 5-day timestep starts the axis in February. Nothing in the constructor args says which convention the writer is being fed (unit tests hand it times starting at the IC), so it fixes the origin from `batch_time[:, 0]` of the first batch, exactly as `MonthlyDataWriter` does. `n_months_through` needs that origin, hence the deferral.

What is known at init is the other end of the axis: `_final_times = IC + n_timesteps * timestep`. So construction pins the end and the first batch pins the start.

Rewritten the comment in fbae63d to say that, and added a one-liner at the `_n_months` computation pointing at the origin dependency.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: That comment was too brief and it also lumped in the spatial dims, which genuinely do just come from the data shape. See the reply on `append_batch` below — short version is that the axis origin is the calendar month of the first *output* time, which isn't necessarily the IC's month, so it can only be fixed once a batch arrives. Comment rewritten in fbae63d.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: Claude: I went looking for this first, and the blocker isn't the time coordinate — `init_time`/`valid_time` would fit `nondim_coords` fine. It's that `ZarrWriter` is a write-once, insert-into-a-slice writer, whereas the monthly writer is read-modify-write. Concretely, three things:

1. **No read path.** Folding a batch into a stored mean requires reading `variable[:, month_slice]` back before writing it. `ZarrWriter` exposes only `record_batch`, and `_insert_into_zarr` reopens the group per call and assigns into the slice; there's nothing to read through. I'd have to reach past it to the underlying zarr group anyway, which is exactly what `MonthlyZarrWriter` holds.
2. **`overwrite_check` encodes the opposite invariant.** It defaults to `True` and would fire on every batch after the first, since overwriting the months a batch touches is the whole design. (That's also why I made zarr + monthly + `overwrite_check: true` a config error rather than silently dropping it.)
3. **One dtype for all arrays.** `_initialize_zarr` takes a single `dtype` and applies it to the data vars *and* the dimension coordinates. The monthly store needs `f4` for the data and lat/lon but `int64` for `lead_time`, `counts`, `init_time` and `valid_time` — and `counts` isn't a coordinate at all, it's an array that gets read back and updated every batch alongside the data.

So the genuine overlap is store creation, and even that needs per-array dtypes and an `int64` non-time dimension coordinate that `ZarrWriter` doesn't support today. My read is that the honest way to share code here is to give `ZarrWriter` a read-modify-write mode plus per-array dtypes and drive both writers from it — worth doing, but a bigger change than this PR, and it touches the raw zarr path. I'd rather do it as a follow-up than reshape the core writer under a monthly-output PR. 

Let me know if you prefer this done in this PR.

fme/ace/inference/data_writer/file_writer.py
brianhenn: That's fine, I agree that it's outside this PR. Maybe will ask the downscaling folks if we still need the separate-IC writer.

fme/ace/inference/data_writer/monthly_zarr.py
brianhenn: My claude 🤖 (and I) think that this is incorrect and we can know _n_months on init, because do know both IC + timestep at init:

```
  You're right that this is knowable at init. The reply's reasoning is that the origin is the month of the first output time, which need not be the IC's month, and that the constructor can't know which convention it's fed.
  But:

  - The IC step is unconditionally stripped before any writer sees data — single_module.py:1255 (time=forcing_data.time[:, self.n_ic_timesteps:]), remove_initial_condition(...) on both the prediction and target paths, and
    test_evaluator.py:616 asserts it ("data writers do not include initial condition"). n_timesteps is n_forward_steps. So the first output time is IC + timestep, deterministically.
  - The class already relies on exactly that at init. _final_times = IC + n_timesteps * timestep (monthly_zarr.py:94-96) is only the true last output time if the IC step is dropped. So it pins one end of the axis using the
    convention it claims it can't assume for the other end. That asymmetry is the actual bug in the argument — not the choice of where to compute n_months.
  - The reply says unit tests hand it IC-first times. The tests in this PR don't: test_monthly_zarr.py:49-57 uses IC Jan 17 → first batch Jan 22, and line 128 literally writes (i_window * window + i_step + 1) * TIMESTEP. And
    no monthly test anywhere would change its result under an init-computed month_of(IC + timestep); meanwhile month_of(IC) without the + timestep would fail test_data_writer.py::test_append_batch (IC 2019-12-31 18:00,
    first batch 2020-01-01), so the tests already pin the offset.
  - Deferring costs something concrete: the bounds check at monthly_zarr.py:248-254 can never fire on a mis-fed stream, because the first batch silently defines the origin. Compute the origin at init and that check becomes a
    real guard.

  Worth noting the payoff is bigger than comment placement: with n_months known at init and spatial sizes taken from coords (as the raw zarr adapter already does), _n_months: int | None, _root: zarr.Group | None, the _store
  guard and the whole deferred-init branch all go away, and MonthIndexer loses its mutable "first call fixes the origin" state. Also: nothing currently tests the case the reply invokes — an IC in a different month from the
  first output. That test should exist regardless of who's right.
```

fme/ace/inference/data_writer/monthly_zarr.py
brianhenn: See above about know month origin on init

fme/ace/inference/data_writer/monthly_zarr.py
brianhenn: OK, I didn't catch the lack of a read method in the core writer. That seems pretty simple to add, though, and the other two issues seem trivial (set overwrite_check to False, and check whether the monthly writer actually needs int64 for the time coordinates since the raw zarr writer apparently didn't need this at all -- if the monthly writer does need this, then change the core writer to allow multiple dtypes). Overall this seems simple enough to be preferable (at least to me) to duplicating a large amount of code.

elynnwu: Ok, sounds good. Will go ahead with the refactor.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: You're both right, and the asymmetry argument is the one that settles it: `_final_times = IC + n_timesteps * timestep` is only the true last output time if the IC step is dropped, so the class was already assuming at construction exactly the convention it claimed it couldn't assume for the other end. I checked the claim rather than taking it on faith — `single_module.py:1257` passes `time=forcing_data.time[:, self.n_ic_timesteps:]` and `remove_initial_condition` runs on both paths, so the first output time is `IC + timestep` deterministically. Fixed in a0f06f0.

What changed:

- `MonthIndexer` now takes each sample's first output time explicitly and has no "first call fixes the origin" state, so `init_years`/`init_months` are plain reads and the `RuntimeError` guard for reading them too early is gone.
- `MonthlyDataWriter` takes the `timestep` it needs to compute that. `PairedMonthlyDataWriter` was already being handed `timestep` and dropping it on the floor, so this is just forwarding it.
- `MonthlyZarrWriter` knows `n_months` at construction, and the bounds check is now a real guard rather than something a mis-fed stream defines its own origin past.
- Added the test you asked for: an IC on Jan 30 with a 5-day timestep, whose first output time is in February. I checked it has teeth — dropping the `+ timestep` makes it fail with a spurious empty January month 0, which is the exact failure mode you predicted.

One thing worth recording, since it's the only part of my original reasoning that survives: the first output time is `IC[0] + timestep` only because `n_ic_timesteps == 1`. With a multi-step initial condition it would be `IC[0] + n_ic_timesteps * timestep`, and `_final_times` would be off by the same amount — so both ends would shift together and `n_months` would still be right, but the origin *label* (and hence `valid_time`) would be wrong. That's now a case the bounds check can actually catch instead of silently absorbing, which is a further argument for your version.

The existing monthly tests all still pass unchanged, including the 360-day one — its IC is 2019-01-01 with a 360-day timestep, and in that calendar `IC + timestep` is exactly 2020-01-01, the first batch time it feeds.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: Agreed — fixed in a0f06f0, see the thread on `append_batch`. The comment that prompted this is gone along with the deferral it was explaining.

fme/ace/inference/data_writer/monthly_zarr.py
elynnwu: Claude: Done in 158a56c, and you were right on all three counts — two of them needed nothing at all.

- **Read path**: added `ZarrWriter.read_batch(names, position_slices)`, mirroring `record_batch`, with a `_read_from_zarr` helper next to `_insert_into_zarr`. It raises if the store isn't initialized yet, and reads dimensions whole when they're omitted from `position_slices`. Two direct tests in `test_writer.py`.
- **`overwrite_check`**: just `False`, as you said.
- **dtypes**: you were right to push back, and the answer is that the monthly writer doesn't need multiple dtypes. `LEAD_TIME_DIM` is `"time"`, so the lead-time axis already goes through `_initialize_zarr`'s time branch and gets `int64` plus its own units (`"months"`, with `time_calendar=None`). And `init_time`, `valid_time` and `counts` all become `nondim_coords`, which are created with their own DataArray dtype — so they get `int64` for free. `counts` was the case I'd thought needed a per-array dtype, and declaring it a non-dimension coordinate is both the honest description of it and enough: its dims are a subset of the data vars' dims, so `_initialize_zarr` reproduces the `coordinates: "init_time valid_time counts"` attribute the old code set by hand. It's still read back and rewritten every batch like the data — nothing in `_insert_into_zarr` cares how an array was created.

So no new core machinery beyond the read method. `monthly_zarr.py` goes 299 → 263 lines and the ~90 lines of hand-rolled `create_array` calls are replaced by one declarative `ZarrWriter(...)`; `fme/core/writer.py` gains ~35 shared lines. The honest accounting is that the win is less about total line count and more that there's now one store-construction path instead of two.

Two consequences worth flagging, neither of which any test objected to:

1. **The store gains a `sample` coordinate** (0…n−1), because `ZarrWriter` requires a coordinate for every dimension. The old monthly store left `sample` as a bare dimension. This matches what the raw zarr writer already emits — including, unfortunately, `float32` for what is really an integer index, since `_initialize_zarr` uses the data dtype for every non-time coordinate. I left that alone rather than change the raw zarr output too, but it's the one place your "allow multiple dtypes" suggestion would still buy something.
2. **Spatial coordinates are now required at construction**, since the store is sized before any data arrives, whereas the old writer took spatial sizes from the first batch's shape and wrote a coordinate only if one was passed. `FileWriterConfig` always passes the dataset's coords, and `ZarrWriter` has always had this requirement for the raw path, so this is parity with raw zarr rather than a new constraint in practice. Noted in the docstring.

Also, `initialize_store` does the root-rank-plus-barrier dance, so the monthly writer picks up the same distributed-init behaviour as the raw path for free, where before it initialized on every rank.

elynnwu:
PR title and description updated by Claude Code (model: claude-opus-4-6, 1M context).

Note: the gather script reported `(warning: could not retrieve title/description)` due to the deprecated Projects Classic GraphQL field, so the prior draft's content was supplied externally rather than read from the PR by the script.

fme/ace/inference/data_writer/monthly.py
```diff
@@ -104,6 +106,7 @@ def __init__(
         path: str,
         label: str,
         initial_condition_times: npt.NDArray[cftime.datetime],
+        timestep: datetime.timedelta,
```
brianhenn, line 109: Is this change actually required? Seems like it could have been avoided since the monthly netCDF writer doesn't precompute the time coordinate and so doesn't need the timestep on init (other than that the MonthlyIndexer is shared and is currently requiring it, but I don't think it has to). Will leave it up to you though.

brianhenn (approved):
Thanks for the changes, they simplify the PR a lot. Have one optional suggestion.

As a follow-on now I'm wondering if we should just switch back to precomputing the time coordinate for the monthly netCDF writer, since it'd be nice if the two ways were consistent about that...

fme/ace/inference/data_writer/monthly.py
elynnwu: Agreed it's derivable, but I'd keep it. `MonthlyZarrWriter` needs it anyway to size its pre-allocated axis, and `FileWriterConfig.build` passes the same args to either writer. It also keeps the "outputs start one step after the IC" convention in one place instead of at each call. `PairedMonthlyDataWriter` already takes an unused `n_timesteps` for the same reason.

PR #1473, opened 2026-08-31
Use four data workers in example/baseline configs

oliverwm1:
Four data workers is a better default since it will use less CPU memory and lead to less CPU contention than the eight data workers previously used in some of our examples. Performance improvements in data loading performance (e.g. #1422) should ensure that training is still not bottlenecked by data loading even with only four workers.


mcgibbon (approved):
LGTM

PR #1477, opened 2026-09-03
Add CM4 ocean 1-daily data processing

jpdunc23:
The `gfdl_om4` pipeline builds the CM4 piControl and 1pctCO2 ocean and
sea-ice training stores at 5-daily cadence. The same sources carry the daily
cycle, and training on it needs the same output variable set at 1-daily
cadence. These configs produce it: every output variable of the 5-daily
stores is producible from the 1-daily sources, so the output variable set is
unchanged and only the source stores, the ice subsample stride and the time
range differ.

Changes:
- `scripts/gfdl_om4/configs/cm4-{piControl,1pctCO2}-{1deg,4deg}-1daily.yaml`:
  read `ocean_1daily_snap.zarr` and `ocean_1daily.zarr` in place of their
  5-daily subsets, stride the 6-hourly ice streams by 4 rather than 20, and
  set `start_time` to the first daily flux label so the snapshot store's
  leading initial-condition instant is dropped and every stream shares one
  time coordinate. Wetmask source, face-mask artifacts, regridding weights
  and the variable set carry over from the 5-daily configs unchanged.
- `scripts/gfdl_om4/Makefile`: a `smoke_test_*_1daily` and `dataflow_*_1daily`
  target per new config, `check_wetmask_equivalence_1daily` over the 1-daily
  1-degree pair, and a `smoke_tests_1daily` aggregate. The 5-daily
  `smoke_tests` aggregate and `check_wetmask_equivalence` are unchanged.

`make smoke_tests_1daily`'s constituent checks were run individually and are
green: each of the four configs writes 6 timesteps of 182 variables through
the DirectRunner and passes `pipeline.check_output`, and
`check_wetmask_equivalence_1daily` finds the two 1-degree wetmasks identical
at all 19 levels.

- [ ] Tests added

### DIFF
--- a/scripts/data_process/Makefile
+++ b/scripts/data_process/Makefile
@@ -346,6 +346,14 @@ cm4_piControl_ocean_stats:
 cm4_1pctCO2_ocean_stats:
 	./compute_stats.sh --config configs/CM4-1pctCO2-ocean-$(RESOLUTION)-140yr.yaml
 
+.PHONY: cm4_piControl_ocean_1daily_stats
+cm4_piControl_ocean_1daily_stats:
+	./compute_stats.sh --config configs/CM4-piControl-ocean-$(RESOLUTION)-1daily-200yr.yaml
+
+.PHONY: cm4_1pctCO2_ocean_1daily_stats
+cm4_1pctCO2_ocean_1daily_stats:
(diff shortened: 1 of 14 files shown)


Arcomano1234 (approved):
LGTM!

PR #1478, opened 2026-09-03
Add L1 loss type to StepLossConfig

elynnwu:
`LossConfig` has supported an `"L1"` branch (building `_L1Loss`) since it was introduced, but `StepLossConfig.type` never included `"L1"` in its `Literal`. Because dacite validates against the `Literal` at parse time, any stepper config with `type: L1` was rejected before the loss could be built — the single-module steppers and the coupled per-component ones alike.

This adds `"L1"` to `StepLossConfig.type` so the existing `_L1Loss` path is reachable from user-facing YAML configs. `LpLoss(p=1)` is not a workaround because it normalizes by the per-channel `‖y‖`, making its semantics different from a plain mean-absolute-error loss.

Changes:
- `fme.core.loss.StepLossConfig.type`: Literal widened to include `"L1"`
- `fme.core.test_loss.test_step_loss_config_l1_is_reachable_from_config`: parses an L1 config through dacite (covering the parse path that was broken, not just the branch), builds the loss, and asserts the MAE value (2.0) discriminates from MSE (4.0)

- [x] Tests added

### DIFF
--- a/fme/core/loss.py
+++ b/fme/core/loss.py
@@ -938,7 +938,7 @@ class StepLossConfig:
             weights to apply to their normalized losses
     """
 
-    type: Literal["LpLoss", "MSE", "AreaWeightedMSE", "EnsembleLoss"] = "MSE"
+    type: Literal["LpLoss", "L1", "MSE", "AreaWeightedMSE", "EnsembleLoss"] = "MSE"
     kwargs: Mapping[str, Any] = dataclasses.field(default_factory=lambda: {})
     global_mean_type: Literal["LpLoss"] | None = None
     global_mean_kwargs: Mapping[str, Any] = dataclasses.field(
--- a/fme/core/test_loss.py
+++ b/fme/core/test_loss.py
@@ -639,12 +639,6 @@ def test_step_loss_forwards_data_mask():
 
 
 def test_step_loss_config_l1_is_reachable_from_config():
-    """``type: L1`` must be admitted by StepLossConfig's Literal.
-
-    ``_L1Loss`` and ``LossConfig``'s ``"L1"`` branch have always worked, but
-    the value was missing from ``StepLossConfig``'s ``Literal``, so dacite
-    rejected it at config parse for every stepper.
-    """
     out_names = ["var_0"]
     normalizer = StandardNormalizer(
         means={name: torch.as_tensor(0.0) for name in out_names},
--- a/fme/core/test_loss.py
+++ b/fme/core/test_loss.py
@@ -1,3 +1,4 @@
+import dacite
 import pytest
 import torch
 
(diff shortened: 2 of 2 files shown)


elynnwu:
Claude: pre-review agent. Independent pre-review (no authoring context), run against `46b76b9` in a fresh worktree using ace's own `.claude/skills/pre-review`.

**Verdict: ready for review.** No changes applied to the branch — I found nothing that warranted a commit.

Audit passes:

- ✅ **Silent-failure.** The one new Literal member routes to the existing `LossConfig.build` `"L1"` branch (`_L1Loss`), identical in shape to the `"MSE"` branch, so it goes through the same `StandardLoss` / `_wrap_elementwise` reduction. Nothing new defaults quietly: before the change `type: L1` raised at parse, after it builds; no other value's behavior moves. `LossConfig.__post_init__`'s tuple already listed `"L1"`, so no second validation gate needed updating.
- ✅ **Behavior-change on existing paths.** Only widening. No existing config, checkpoint, or caller observes a difference.
- ✅ **Differential test.** Verified empirically: reverting the Literal to its pre-PR value makes `test_step_loss_config_l1_is_reachable_from_config` fail (dacite `WrongTypeError`) and nothing else. The numeric assertion also discriminates — `x=2.0`, `y=0` gives 2.0 for L1 vs 4.0 for MSE, so a mis-wired branch would be caught, not just the parse.
- ✅ **Sibling / mirror coverage.** Swept for parallel lists that would need the same value: `LossConfig.type` already has `"L1"` (and its `__post_init__` tuple); `CorrectorRegularizationConfig.norm` already offers `"L1"`/`"MSE"`; no docs, JSON/YAML schema, or config-reference page enumerates loss types (`docs/` has no loss-type list — Sphinx renders the dataclass). Nothing else to update. Two deliberate non-changes worth a reviewer's glance:
  - `StepLossConfig` still omits `"NaN"`, which `LossConfig` admits. That asymmetry predates this PR (`NaN` is a downscaling inference/test device) and I read it as intentional.
  - `global_mean_type` stays `Literal["LpLoss"]`. Admitting `"L1"` there is not a Literal edit: `LossConfig.build` hardcodes `LpLoss(**global_mean_kwargs)` for the global-mean term and `__post_init__` raises `NotImplementedError` on anything else, so it would need real code. Correctly out of scope here.
- ✅ **Config back-compat.** Every prior YAML still parses; no checkpoint-persisted state touched.
- ✅ **Split heuristic.** 2 files, one concern.
- ✅ **Abstract-or-final / inheritance depth.** No class changes.
- ✅ **AI-agent artifacts.** `import dacite` is at module top (this test file's first dacite use — deliberate, since the point is to exercise the parse path that was broken rather than the constructor, and the docstring says so). No stray `Any`, no `isinstance`/`type: ignore`, no unrelated deletions.
- ✅ **Builder pattern.** No new config-reads-config coupling. (`fme/coupled/stepper.py` reading `self.ocean.loss.type` is pre-existing and untouched.)

Checks run in a clean worktree: `FME_FORCE_CPU=1 pytest fme/core/test_loss.py -q` → 92 passed; `pre-commit run --files fme/core/loss.py fme/core/test_loss.py` → ruff, ruff-format, mypy all pass.

Also spot-checked the description's claim that `LpLoss(p=1)` is not a workaround — accurate: it returns `diff_norms / y_norms`, normalizing by the per-channel `‖y‖`.

One optional nit, not worth a commit: `test_step_loss_config_l1_is_reachable_from_config` repeats "config"; `test_step_loss_config_accepts_l1_type` would read better, and the docstring's history paragraph duplicates the PR description. Both are fine as-is.

elynnwu:
Title and description updated via the `write-pr-prose` skill.

Model attribution: the skill's pinned model, `claude-opus-4-6[1m]`, authored the draft in the forked skill run. The session model, `claude-opus-5[1m]`, then applied the requested edits before posting — reflowing the paragraphs onto single lines, adding the in-place reason on the unchecked deps checklist item, and dropping the "motivated by a 1° Samudra ocean finetune arm comparing L1 against the recipe's MSE" line from the description, since that is reviewer context rather than something belonging in the squash-and-merge commit message. Recording it here instead: that finetune arm is what motivated the change.

No context-gathering warnings were reported by `gather_pr_context.sh`.

Arcomano1234 (approved):
LGTM

PR #1479, opened 2026-09-04
Add a uniform-temperature ocean heat content correction

elynnwu:
The ocean heat content corrector currently has one method, `scaled_temperature`, which conserves the column heat content budget by multiplying every potential-temperature level by a single global ratio. That deposits the correction in proportion to `T_k * dz_k`, and it has a structural problem for training: the corrected field is invariant under a global rescaling of the network's temperature output, so the amplitude the network predicts is unconstrained by any loss computed after the corrector. Measured on the deployed 1° ocean runs, the correction is 0.2x to 8x the size of the field itself and differs by 2x between seeds, so the amplitude is an unpinned gauge.

This adds a second method, `uniform_temperature`, which conserves the same budget by adding a single global temperature increment to every level, depositing the correction in proportion to `dz_k` alone. The additive form has no such rescaling invariance, so a post-corrector loss does constrain the predicted amplitude. It also changes which levels absorb the correction, which is the property a follow-up training experiment tests.

`scaled_temperature` is bit-for-bit unchanged and remains what every existing config gets, including the legacy `ocean_heat_content_correction: true` migration.

The increment is `(global_input_ohc + expected_change_ohc - global_gen_ohc) / heat_capacity_per_area`, where the denominator is the area-weighted global mean of the column integral of `rho * cp` taken through the vertical coordinate's own `depth_integral`. Conservation is exact to floating-point tolerance because that denominator is the same integral that weights the heat the increment deposits: partial bottom cells and the per-level valid mask enter it exactly as they enter the heat content integral itself. A denominator built any other way, such as a nominal `dz` sum or a mean over a different set of columns, would leave the corrected heat content off by the difference between the two integrals.

Separately from conservation, the increment is applied where the vertical coordinate's `mask` marks a layer valid. This is deliberately not `dz > 0`: the mask is uniformly more permissive than the bathymetry, so a cell can be marked valid and still have zero layer thickness — on the 1° ocean store, 4.6% of all valid-marked cells and 60% of the bottom level's. Those cells hold real data, the output masker keeps them and every metric scores them, so keying the increment to thickness would split a scored level into shifted and unshifted cells along a mask/bathymetry inconsistency in the store rather than along anything physical. Shifting them costs the budget nothing, because zero thickness means they absorb no heat either way. Cells the mask excludes are left alone: they hold fill rather than data, and nothing necessarily overwrites them after the corrector runs, so an additive increment there would drift away from the fill value where a multiplicative one keeps a zero fill at zero. The increment stays on the autograd graph, since the correction runs inside the training loop and the loss differentiates through it.

Changes:
- `fme.core.corrector.ocean.OceanHeatContentBudgetConfig.method` and `OceanHeatContentCorrection.method` accept `"uniform_temperature"` alongside `"scaled_temperature"`
- `fme.core.corrector.ocean._force_conserve_ocean_heat_content` gains the additive branch, sharing the existing global-mean/flux/geothermal plumbing and `expected_change_ocean_heat_content` with the multiplicative one; its docstring states which invariant conservation rests on and which one the valid-cell mask serves
- `fme.core.ocean_data.HasOceanDepthIntegral` requires `mask`, the per-level validity flags, which `DepthCoordinate` already provided

- [x] Tests added

### DIFF
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -453,15 +453,26 @@ def _force_conserve_ocean_heat_content(
     or a mean over a different set of columns, would leave the corrected heat
     content off by the difference between the two integrals.
 
-    Restricting the increment to ``dz > 0`` is a separate concern and does not
-    affect conservation. ``vertical_coordinate.dz`` is zero wherever a layer is
-    invalid (land, or below the sea floor), so an unmasked increment would
-    deposit no heat there either. The mask is there so that invalid cells pass
-    through unchanged rather than being shifted by a physical increment: they
-    hold raw denormalized network output, and when no spatial mask provider is
-    configured nothing overwrites them after the corrector runs.
-    ``scaled_temperature`` rescales invalid cells instead of leaving them
-    alone, which is pre-existing behavior and not something the budget needs.
+    Which cells the increment lands on is a separate concern and does not
+    affect conservation. It lands on every cell ``vertical_coordinate.mask``
+    marks valid. ``dz > 0`` is not the right test for this and is not the same
+    cell set: the mask is more permissive than the bathymetry, so a cell can be
+    marked valid and still have zero thickness. On the 1 degree ocean store
+    that is 4.6% of all valid-marked cells and 60% of the bottom level's. Those
+    cells hold real data, the output masker keeps them and the metrics score
+    them, so restricting the increment to ``dz > 0`` would split a scored level
+    into shifted and unshifted cells along a mask/bathymetry inconsistency in
+    the store rather than along anything physical, leaving the unshifted
+    majority of a deep level tied to neither the budget nor its neighbors.
+    Shifting them costs the budget nothing, because zero thickness means they
+    absorb no heat either way.
+
+    Cells the mask excludes are a different matter and are left alone: they hold
+    fill rather than data, and when no spatial mask provider is configured
+    nothing overwrites them after the corrector runs, so ``+ delta_T`` there
+    would drift away from the fill value where ``* ratio`` keeps a zero fill at
+    zero. ``scaled_temperature`` rescales every cell regardless, which is
+    pre-existing behavior and not something the budget needs.
     """
     if method not in ("scaled_temperature", "uniform_temperature"):
         raise NotImplementedError(
@@ -548,22 +559,26 @@ def _force_conserve_ocean_heat_content(
         temperature_increment = (
             target_ocean_heat_content - global_gen_ocean_heat_content
         ) / heat_capacity_per_area
-        # zero thickness marks a layer the depth integral cannot see; the mask
-        # leaves those cells' raw network output untouched, and costs the budget
-        # nothing because they absorb no heat either way
-        is_valid = (vertical_coordinate.dz > 0.0).to(
+        # every cell the store marks valid is shifted, including the ones the
+        # bathymetry puts at zero thickness -- they hold data and are scored, and
+        # shifting them adds no heat. Cells outside the mask hold fill, so they
+        # are left alone. Not interchangeable with dz > 0; see the docstring.
+        # ``> 0`` rather than ``== 1`` mirrors depth_integral's own mask test.
+        is_masked_valid = (vertical_coordinate.mask > 0.0).to(
             dtype=gen_potential_temperature.dtype
         )
         for k in range(n_levels):
             name = f"thetao_{k}"
-            out[name] = gen.data[name] + temperature_increment * is_valid.select(-1, k)
+            out[name] = gen.data[name] + temperature_increment * is_masked_valid.select(
+                -1, k
+            )
         if "sst" in gen.data:
             # an increment needs no Kelvin offset, unlike the multiplicative
             # path. sst is not in the heat content integral, so shifting it by
             # the same increment as thetao_0 is a consistency choice, not a
-            # budget requirement; the surface layer's validity is the
-            # wet-column indicator depth_integral itself uses.
-            out["sst"] = gen.data["sst"] + temperature_increment * is_valid.select(
-                -1, 0
-            )
+            # budget requirement; the surface level's mask is the wet-column
+            # indicator depth_integral itself uses.
+            out["sst"] = gen.data[
+                "sst"
+            ] + temperature_increment * is_masked_valid.select(-1, 0)
     return out
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -7,6 +7,7 @@ import torch
 
 from fme.core.atmosphere_data import AtmosphereData
 from fme.core.constants import (
+    DENSITY_OF_SEA_WATER_CM4,
     FREEZING_TEMPERATURE_KELVIN,
     LATENT_HEAT_OF_VAPORIZATION,
     SPECIFIC_HEAT_OF_SEA_WATER_CM4,
@@ -95,10 +96,16 @@ class OceanHeatContentBudgetConfig:
     """Configuration for ocean heat content budget correction.
 
     Parameters:
-        method: Method to use for OHC budget correction. The available option is
-            "scaled_temperature", which enforces conservation of heat content
-            by scaling the predicted potential temperature by a vertically and
-            horizontally uniform correction factor.
+        method: Method to use for OHC budget correction. Both options enforce
+            the same column heat content budget and differ only in how the
+            correction is distributed in the vertical:
+
+            - "scaled_temperature": multiply the predicted potential
+              temperature by a vertically and horizontally uniform correction
+              factor, so the heat is deposited in proportion to ``T_k * dz_k``.
+            - "uniform_temperature": add a vertically and horizontally uniform
+              temperature increment, so the heat is deposited in proportion to
+              ``dz_k`` alone.
         constant_unaccounted_heating: Area-weighted global mean
             column-integrated heating in W/m**2 to be added to the energy flux
             into the ocean when conserving the heat content. This can be useful
@@ -107,7 +114,7 @@ class OceanHeatContentBudgetConfig:
 
     """
 
-    method: Literal["scaled_temperature"]
+    method: Literal["scaled_temperature", "uniform_temperature"]
     constant_unaccounted_heating: float = 0.0
 
 
@@ -203,7 +210,7 @@ class OceanHeatContentCorrection:
     area_weighted_mean: AreaWeightedMean
     vertical_coordinate: HasOceanDepthIntegral | None
     timestep_seconds: float
-    method: Literal["scaled_temperature"]
+    method: Literal["scaled_temperature", "uniform_temperature"]
     unaccounted_heating: float
 
     def __call__(
@@ -416,10 +423,38 @@ def _force_conserve_ocean_heat_content(
     area_weighted_mean: AreaWeightedMean,
     vertical_coordinate: HasOceanDepthIntegral,
     timestep_seconds: float,
-    method: Literal["scaled_temperature"] = "scaled_temperature",
+    method: Literal["scaled_temperature", "uniform_temperature"] = "scaled_temperature",
     unaccounted_heating: float = 0.0,
 ) -> TensorDict:
-    if method != "scaled_temperature":
+    """Adjust the generated potential temperature so the global-mean column
+    ocean heat content matches the input value plus the heat the surface and
+    geothermal fluxes deposit over one timestep.
+
+    Both methods conserve the same budget and differ only in the vertical
+    profile of the heat they deposit:
+
+    scaled_temperature
+        Multiply every level by one global ratio, depositing heat in proportion
+        to ``T_k * dz_k``.
+    uniform_temperature
+        Add one global temperature increment to every valid level, depositing
+        heat in proportion to ``dz_k`` alone. The increment is
+        ``(target - global_gen_ohc) / heat_capacity_per_area``, where the
+        denominator is the area-weighted global mean of the column integral of
+        ``rho * cp``.
+
+    The invariant the ``uniform_temperature`` masking relies on is that
+    ``vertical_coordinate.dz`` is zero exactly where a layer is invalid (land,
+    or below the sea floor), which is the same weight
+    ``vertical_coordinate.depth_integral`` applies. So restricting the
+    increment to ``dz > 0`` both keeps it out of invalid cells and leaves the
+    heat it adds equal to ``delta_T`` times the very integral that forms the
+    denominator -- making conservation exact rather than off by the masked
+    fraction. ``scaled_temperature`` needs no such mask because it scales
+    invalid cells rather than shifting them, and invalid cells carry no weight
+    in the integral either way.
+    """
+    if method not in ("scaled_temperature", "uniform_temperature"):
         raise NotImplementedError(
             f"Method {method!r} not implemented for ocean heat content conservation"
         )
@@ -470,17 +505,50 @@ def _force_conserve_ocean_heat_content(
     expected_change_ocean_heat_content = (
         energy_flux_global_mean + unaccounted_heating
     ) * timestep_seconds
-    heat_content_correction_ratio = (
+    target_ocean_heat_content = (
         global_input_ocean_heat_content + expected_change_ocean_heat_content
-    ) / global_gen_ocean_heat_content
-    # apply same temperature correction to all vertical layers
+    )
     out: TensorDict = {}
-    n_levels = gen.sea_water_potential_temperature.shape[-1]
-    for k in range(n_levels):
-        name = f"thetao_{k}"
-        out[name] = gen.data[name] * heat_content_correction_ratio
-    if "sst" in gen.data:
-        out["sst"] = (  # assuming sst in Kelvin
-            gen.data["sst"] - FREEZING_TEMPERATURE_KELVIN
-        ) * heat_content_correction_ratio + FREEZING_TEMPERATURE_KELVIN
+    gen_potential_temperature = gen.sea_water_potential_temperature
+    n_levels = gen_potential_temperature.shape[-1]
+    if method == "scaled_temperature":
+        heat_content_correction_ratio = (
+            target_ocean_heat_content / global_gen_ocean_heat_content
+        )
+        # apply same temperature correction to all vertical layers
+        for k in range(n_levels):
+            name = f"thetao_{k}"
+            out[name] = gen.data[name] * heat_content_correction_ratio
+        if "sst" in gen.data:
+            out["sst"] = (  # assuming sst in Kelvin
+                gen.data["sst"] - FREEZING_TEMPERATURE_KELVIN
+            ) * heat_content_correction_ratio + FREEZING_TEMPERATURE_KELVIN
+    else:
+        heat_capacity_per_area = area_weighted_mean(
+            vertical_coordinate.depth_integral(
+                torch.ones_like(gen_potential_temperature)
+                * SPECIFIC_HEAT_OF_SEA_WATER_CM4
+                * DENSITY_OF_SEA_WATER_CM4
+            ),
+            keepdim=True,
+            name="ocean_heat_content",
+        )
+        temperature_increment = (
+            target_ocean_heat_content - global_gen_ocean_heat_content
+        ) / heat_capacity_per_area
+        # zero thickness marks a layer the depth integral cannot see, so the
+        # increment must not land there; see the docstring.
+        is_valid = (vertical_coordinate.dz > 0.0).to(
+            dtype=gen_potential_temperature.dtype
+        )
+        for k in range(n_levels):
+            name = f"thetao_{k}"
+            out[name] = gen.data[name] + temperature_increment * is_valid.select(-1, k)
+        if "sst" in gen.data:
+            # an increment needs no Kelvin offset, unlike the multiplicative
+            # path; the surface layer's validity is the ocean-column indicator
+            # depth_integral itself uses.
+            out["sst"] = gen.data["sst"] + temperature_increment * is_valid.select(
+                -1, 0
+            )
     return out
--- a/fme/core/corrector/ocean.py
+++ b/fme/core/corrector/ocean.py
@@ -467,9 +467,20 @@ def _force_conserve_ocean_heat_content(
     Shifting them costs the budget nothing, because zero thickness means they
     absorb no heat either way.
 
+    The test is ``mask > 0`` rather than ``mask == 1`` because ``dz`` is itself
+    ``mask``-weighted, so ``dz_k`` is zero wherever ``mask_k`` is zero and
+    ``dz_k * (mask_k > 0)`` equals ``dz_k`` cell for cell whatever values the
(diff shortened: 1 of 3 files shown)


elynnwu:
Claude: PR title and description were authored by Claude Opus 5 (1M context), not by the model pinned in `.claude/skills/write-pr-prose/` (`claude-opus-4-6[1m]`). This PR was written from a research-repo agent session where that skill is not loadable, so I followed its instructions and `.github/pull_request_template.md` by hand rather than running it. Flagging so the deviation from the pin is visible.

Tests added in `fme/core/corrector/test_ocean.py`, over a masked 3x3x3 fixture with a non-trivial sea floor (one all-land column, one with only its surface layer wet, one with two of three layers, and a partial bottom cell in every wet column):

- `test_uniform_temperature_conserves_ocean_heat_content` — corrected global-mean OHC equals the target to tolerance, with and without `constant_unaccounted_heating`
- `test_uniform_temperature_deposits_heat_proportional_to_thickness` — the per-level increment is one global constant on valid cells, so heat added goes as `dz_k`; contrast assertion pins `scaled_temperature` on the same fixture to `T_k * dz_k`
- `test_uniform_temperature_leaves_invalid_cells_unchanged`
- `test_ocean_heat_content_correction_is_differentiable` — parametrized over both methods
- `test_ocean_corrector_is_per_member_under_ensemble_folding` — now parametrized over both methods
- `test_ocean_heat_content_method_round_trips_through_from_state`, `test_ocean_heat_content_unknown_method_raises`

Both the conservation and profile/masking assertions were mutation-checked: dropping the valid-cell mask fails the profile and masked-cell tests, and swapping the denominator for a nominal `dz` sum fails the conservation tests.

elynnwu:
Claude (pre-review agent): independent pre-review, run cold from a fresh session with none of the authoring context. Everything below was checked by reading the code and mutating it, not by trusting the diff description.

## What I ran

- `FME_FORCE_CPU=1 python -m pytest fme/core -q` — 1509 passed; the 7 `fme/core/dataset/test_xarray.py` failures are pre-existing on `origin/main` (verified in a separate `origin/main` worktree, same 7)
- `FME_FORCE_CPU=1 python -m pytest fme/core/corrector fme/ace/test_ocean_train.py fme/core/test_ocean_data.py fme/core/test_coordinates.py fme/core/test_dataset_info.py -q` — all green
- `pre-commit run --files fme/core/corrector/ocean.py fme/core/corrector/test_ocean.py fme/core/ocean_data.py` — ruff, ruff-format, mypy all pass

## Audit passes

**1. Silent-failure ✅.** An unknown `method` is rejected by dacite at config load and by `NotImplementedError` at the write; both covered. `vertical_coordinate is None` raises. The legacy `ocean_heat_content_correction: true` migration still resolves to `scaled_temperature`. `heat_capacity_per_area` can only be zero if every column has zero total thickness, and in that case `global_gen_ocean_heat_content` is already NaN and the multiplicative path divides by zero too, so there is no new quiet branch. The one NaN-propagating shape I looked for — a `DepthCoordinate` carrying a land mask while the `GriddedOperations` has no mask provider, which would make both the global OHC and the denominator NaN — is unreachable, since the dataset loader builds both from the same `mask_*` variables.

**2. Behavior-change on existing paths ✅.** `scaled_temperature` is bit-for-bit unchanged, verified rather than argued: same fixture (masking, partial bottom cells, `sst` present, `constant_unaccounted_heating=0.1`) run against `origin/main` and against this branch in separate worktrees, `torch.equal` true on every output tensor, max diff 0.0. The restructuring is only the hoist of `global_input + expected_change` into `target_ocean_heat_content`, and the `sst` freezing-point offset is untouched. The guard also changed from `!=` to `not in`, and still raises for anything else.

**3. Differential-test ⚠️ → addressed (see "What I changed").** The conservation tests are genuinely sensitive to the denominator: scaling it by 1.05, and separately averaging it over the dry columns as well (the "off by the masked fraction" failure), each fail both parametrizations. But two gaps:
   - Removing the `dz > 0` mask entirely leaves **both conservation tests passing**. That is correct behavior, not a test weakness — `dz` is zero in invalid cells, so an unmasked increment deposits no heat there either — but it means the docstring, the inline comments, and the conservation test's own comment described an invariant the code does not have. Fixed.
   - No test covered the 1-D-`dz` shape (dataset with no `mask_0`). Added one; it is the only test that catches building the denominator off `dz`'s shape instead of the potential temperature's.

**4. Sibling / mirror coverage ✅.** No `fme/coupled` mirror of the ocean corrector exists. The only in-repo config reference is `fme/ace/test_ocean_train.py`, which pins `scaled_temperature` deliberately.

**5. Config back-compat ✅.** A `Literal` widened, no field removed or restructured, nothing checkpoint-persisted changed, so every prior training and inference YAML still parses and every prior checkpoint still loads.

**6. Split ✅.** Three files, one reason.

**7. Abstract-or-final / inheritance ✅.** No new classes and no new subclassing.

**8. Protocol widening ✅.** `DepthCoordinate` and the test's `_MockDepth` are the only implementers, and both have `dz`. `NullVerticalCoordinate` and `HybridSigmaPressureCoordinate` have neither `depth_integral` nor `dz`, so they failed the protocol before and after — confirmed at runtime, along with `isinstance(DepthCoordinate(...), HasOceanDepthIntegral) is True`, which is what `DatasetInfo.ocean_vertical_coordinate` relies on. One consequence worth knowing: adding a property makes this a *data* protocol, so `issubclass()` against it now raises `TypeError`. Nothing in the repo uses `issubclass` here, and `isinstance` is unaffected.

**9. Gradient flow ✅.** `torch.ones_like` produces a leaf with `requires_grad=False`, so the denominator is a constant while the numerator carries `global_gen_ocean_heat_content`'s graph; the valid-cell mask is a constant multiply. `test_ocean_heat_content_correction_is_differentiable` covers both methods and asserts non-zero finite gradients on every `thetao_k` and on `hfds`.

**10. AI-agent artifact ⚠️ → addressed.** The docstring narrated a wrong invariant (pass 3). Otherwise clean: imports at module top, `cast(Any, ...)` in the unknown-method test is justified in place, no `Any` return types, no stale comments.

## What I changed

Two commits pushed to the branch, plus a description edit.

1. `ece41aa` — corrects what the `dz > 0` mask does, in the function docstring, the two inline comments, and two test comments. The claim was that masking the increment is what makes conservation exact "rather than off by the masked fraction". Mutation-checking shows otherwise: conservation survives removing the mask, because zero-thickness cells absorb no heat. What actually pins conservation is that `heat_capacity_per_area` comes from `depth_integral` over the same columns as the heat content, so the docs now say that, and say separately that the mask exists so invalid cells' raw denormalized output passes through unshifted. I also added an inline warning at `heat_capacity_per_area` so a future reader does not "optimize" it into a hand-rolled `dz` sum. No behavior change.

2. `6304187` — adds `test_uniform_temperature_conserves_with_an_unmasked_depth_coordinate`, covering the dataset-without-`mask_0` shape where `DepthCoordinate`'s mask is 1-D, `dz` has no horizontal dimensions, and the mask provider is Null. Conservation is exact there (relative error 0.0, versus 1.7e-7 for `scaled_temperature` on the same fixture) and output shapes are preserved.

3. Description: rewrote the conservation paragraph to match the corrected reasoning, and **removed the reference to the private research-repo investigation slug** — `ai2cm/ace` is public and the description becomes the permanent squash commit message. If you want that pointer kept, a PR comment is the place for it. As with the original prose, this was not written through `.claude/skills/write-pr-prose/` (not loadable from this session), so the same model-pin deviation the author flagged applies; I kept the edits surgical rather than rewriting.

## Left for you

None of these are blockers; they are judgment calls I did not want to make unilaterally on code that gates a training run.

- **`dz > 0` is strictly narrower than `mask == 1`.** `dz_from_idepth` clamps to `deptho`, so a layer the dataset marks valid but whose `deptho` falls at or above its top gets `dz == 0`. Such a cell takes no increment and contributes no heat, which is self-consistent, but the corrector's notion of "wet" is now `dz > 0` rather than the dataset mask. Worth confirming the 1° ocean dataset has no such cells, or that you are happy for them to stay unwarmed.
- **`sst` gets the same additive increment as `thetao_0`.** `sst` is not in the heat content integral, so this is a modelling choice rather than a budget requirement (the multiplicative path scales it, consistently within its own form). Since `sst` is what the atmosphere sees, it seems worth a deliberate nod.
- **fp32 cancellation is not a problem, for the record.** The increment comes from the difference of two nearly-equal global-mean OHC values. I measured the cost against float64 at production-like depth (5500 m, 30 levels): 0.03% relative error on the increment, and the same order for the multiplicative path's ratio. No action needed; noting it so nobody re-derives the worry.
- **`heat_capacity_per_area` is rebuilt every step** from a full `ones_like(potential_temperature)`, three temporaries at that tensor's size, for a quantity that is constant in time. Caching it would take that out of the step, but the current form is deliberately shaped off `gen` rather than off `dz` (that is what the new test pins), so I left the arithmetic alone rather than perturb the numerics of a run-gating correction.

## Overall

The change does what it says: an additive branch that shares all the budget plumbing with the multiplicative one, leaves the multiplicative path bit-identical, and conserves exactly for the reason it should. The diff is three files and easy to follow, and the algebra checks out by hand and under mutation. The one real defect was documentation stating the wrong invariant, which on a PR whose whole point is a conservation property would have cost you real review time; that and one missing shape case are now fixed.

**Verdict: ready for review** (after the two fixes above).

elynnwu:
Claude: reworked the masking of the additive increment, and amended the description's masking paragraph and `HasOceanDepthIntegral` bullet to match (the rest of the description, including the provenance sentence, is untouched). As before, the prose was authored by Claude Opus 5 (1M context) rather than the model pinned in `.claude/skills/write-pr-prose/` — this session runs outside the ace clone and cannot load that skill, so I followed its SKILL.md and `.github/pull_request_template.md` by hand. `gh pr edit --body-file` silently no-ops here (it aborts on the Projects-classic GraphQL deprecation error), so the amendment went through `gh api --method PATCH /repos/ai2cm/ace/pulls/1479`.

What changed in the code: the increment now lands where `vertical_coordinate.mask > 0` rather than where `dz > 0`, and the protocol requires `mask` instead of `dz`. `dz` is no longer read by the implementation, so requiring it would have been an unused protocol member; that is the one deviation from "widen with `mask` alongside `dz`".

Tests changed:

- `_make_sea_floor_fixture` column (2, 0) now has `deptho = 22` with all three levels mask-valid, so its bottom cell is `mask == 1, dz == 0` — the case that separates the two cell sets. The fixture asserts that cell exists.
- New `test_uniform_temperature_shifts_valid_zero_thickness_cells`: that cell **is** shifted by `delta_T`, and the budget still closes exactly because it absorbs no heat.
- `test_uniform_temperature_leaves_invalid_cells_unchanged` renamed to `..._leaves_cells_outside_the_mask_unchanged` and now pins `mask == 0`, with an assertion that `mask == 0` and `dz == 0` are genuinely different sets on the fixture.
- `test_uniform_temperature_deposits_heat_proportional_to_thickness`: the shifted-cell set is now the mask; the heat-goes-as-`dz` and `scaled_temperature`-goes-as-`T_k * dz_k` assertions are unchanged.
- `test_uniform_temperature_conserves_with_an_unmasked_depth_coordinate` still passes and now also asserts the 1-D `mask` shape.

Mutation-checked: `mask > 0` → all-ones fails `..._leaves_cells_outside_the_mask_unchanged`; `mask > 0` → `dz > 0` fails `..._shifts_valid_zero_thickness_cells`; a nominal-`dz` denominator fails both conservation tests. Each mutation is caught only by the test meant to catch it.

elynnwu:
Pre-review agent: **ready for review** (re-pre-review of `685b7413e`, after the mask fix; two documentation-accuracy fixes and one test applied in `79fc87fa9`).

This is a second pass — the first passed `f40eb9149`, which keyed the increment to `dz > 0`. That verdict no longer applied once `685b7413e` moved it onto `mask > 0`, so everything below was re-derived from scratch in a clean worktree.

**No correctness problem found. The launched training run does not need to be killed.**

### What I ran

- `python -m pytest fme/core -q` → `1511 passed, 7 failed`. All 7 failures are in `fme/core/dataset/test_xarray.py` (zarr fixtures) and fail identically on `origin/main` in the same environment — pre-existing, unrelated.
- `python -m pytest fme/core/corrector fme/coupled -q` → 370 passed.
- `python -m pytest fme/core/corrector fme/core/test_coordinates.py fme/core/test_ocean_data.py fme/core/test_dataset_info.py fme/ace/test_ocean_train.py -q` → 195 passed.
- `pre-commit run --files fme/core/corrector/ocean.py fme/core/corrector/test_ocean.py fme/core/ocean_data.py` → ruff, ruff-format, mypy all pass.

All runs under `PYTHONPATH` pinned to the worktree; the editable `fme` install resolves elsewhere, so the interpreter's `fme.__file__` was printed and checked on every comparison run.

### 1. Conservation is exact, and for the stated reason

The identity is `dz_k * 1[mask_k > 0] == dz_k` at every cell, because `dz_from_idepth` returns `dz.nan_to_num() * mask` — `dz_k` is already zero wherever `mask_k` is zero. So

```
mean(depth_integral(cp*rho*is_valid)) == mean(depth_integral(cp*rho*1)) == heat_capacity_per_area
```

exactly, whatever the numerator's cell set does beyond `dz > 0`. Both the numerator terms and the denominator go through the *same* `area_weighted_mean(..., name="ocean_heat_content")` call, so the column selection and NaN handling are identical by construction.

Verified numerically on a synthetic 3x3x3 grid with land columns, partial bottom cells, and a valid-marked zero-thickness cell: `mean int(ones) == mean int(mask>0) == 37.235294342041016`, conservation rel err `0.000e+00`.

Denominator mutations all still fail the conservation tests:

| mutation | result |
|---|---|
| denominator from a nominal `idepth.diff().sum()` instead of `depth_integral` | 3 tests fail |
| denominator meaned over all columns (`name` dropped, NaNs zeroed) | 3 tests fail |
| denominator from a `dz > 0` count instead of true partial-cell `dz` | 4 tests fail |

### 2. `mask > 0` is right, and for a stronger reason than the comment gave — fixed in `79fc87fa9`

The comment justified `> 0` as mirroring `depth_integral`'s own mask test. That is only half true: `depth_integral` uses `mask_0 > 0` as a *column* validity gate, but the per-level mask enters `dz` **multiplicatively**. The real reason `> 0` is correct is the identity in §1, which holds for *any* mask values in [0, 1].

`mask == 1` would be the unsafe choice. Measured on the same synthetic grid with one full-thickness cell given `mask = 0.5`:

```
mean int(ones)                = 35.4706
mean int(mask>0)  [as coded]  = 35.4706   <- conservation holds
mean int(mask==1) [alt]       = 33.7059   <- budget would miss by 5%
```

`DepthCoordinate` documents a 0/1 mask but does not enforce one — it is read straight out of the store's `mask_<k>` variables in `_get_vertical_coordinate` with no thresholding — so this is a latent, not purely hypothetical, distinction. `mask == 1` was the one mutation that survived the whole suite, so I added `test_uniform_temperature_conserves_under_a_fractional_mask` to pin it, and rewrote the docstring/comment to state the actual invariant.

### 3. The protocol change is a pure addition, not a narrowing

Net from `main`, `HasOceanDepthIntegral` gains `mask` and loses nothing: `main`'s protocol declared only `depth_integral`. `dz` existed only on an intermediate commit within this PR, so there is nothing to revert to.

`HasOceanDepthIntegral` is `@runtime_checkable` and `DatasetInfo.ocean_vertical_coordinate` dispatches on `isinstance`, so widening it *is* a live behavior change — checked it empirically on both trees:

| coordinate | `isinstance` on `main` | on the PR |
|---|---|---|
| `DepthCoordinate` | True | True |
| `HybridSigmaPressureCoordinate` | False | False |
| `NullVerticalCoordinate` | False | False |

and `DatasetInfo.ocean_vertical_coordinate` returns / raises / returns `None` identically. `DepthCoordinate` and the one test fake (`_MockDepth`) are the only `depth_integral` implementers in the repo; both provide `mask`. The implementation reads no `dz`, so `mask` alone is the correct minimal protocol.

### 4. `scaled_temperature` is bit-identical to `origin/main`

Two worktrees (`origin/main` at `d17bde28e` vs this branch), one seeded fixture with land columns and partial bottom cells, three configs (OHC only; OHC with `constant_unaccounted_heating`; OHC + `force_positive` + sea-ice correction), 36 output tensors compared with `torch.equal`: **all identical**. `fme.__file__` was printed on each run and confirmed to point at the respective worktree.

The refactor that could have drifted it — hoisting `target_ocean_heat_content` out of the ratio — reassociates nothing: `main` computed `(a + b) / c`, the PR computes `t = (a + b)` then `t / c`, the same two float ops in the same order.

### 5. Every new test bites

Eight independent mutations, each killed:

| mutation | tests that fail |
|---|---|
| `mask > 0` → all ones | `..._deposits_heat_proportional_to_thickness`, `..._leaves_cells_outside_the_mask_unchanged` |
| `mask > 0` → `dz > 0` | `..._deposits_heat_proportional_to_thickness`, `..._shifts_valid_zero_thickness_cells` |
| `mask > 0` → `mask == 1` | `..._conserves_under_a_fractional_mask` (added here) |
| drop the per-level `.select(-1, k)` | 2 tests |
| use level 0's mask at every level | 2 tests |
| `sst` shifted without the mask | `..._leaves_cells_outside_the_mask_unchanged` |
| 3 denominator mutations | see §1 |

The fixture's `(2, 0)` column does carry the intended `mask == 1, dz == 0` cell — asserted in `_make_sea_floor_fixture` itself and re-asserted independently in both load-bearing tests. `test_uniform_temperature_conserves_with_an_unmasked_depth_coordinate` passes and asserts both `mask.shape == (nz,)` and `dz.shape == (nz,)`; it is killed by the `dz > 0`-count denominator mutation, so it is not just a smoke test.

### 6. Docstring claims — two corrected in `79fc87fa9`

The three claims are now stated separately and each is correct. Two were not:

- **`> 0` vs `== 1`** — see §2. The stated reason was a resemblance to `depth_integral`'s column gate; the real one is the `dz`-weighting identity.
- **Out-of-mask cells** — the docstring said the increment must skip them because "when no spatial mask provider is configured nothing overwrites them". That specific combination cannot arise from the dataset path: `mask_0` present gives *both* a horizontal `DepthCoordinate` mask *and* a populated `SpatialMaskProvider`; `mask_0` absent gives an all-ones 1-D mask with no excluded cells at all. The real exposure is per-variable: `StaticSpatialMasking.__call__` does `if mask is None: continue`, and `get_mask_tensor_for("sst")` falls through `mask_sst` → (no level match) → `mask_2d`, so on a store with per-level masks but no `mask_sst`/`mask_2d`, `sst` is left exactly as the corrector wrote it. Restated to that.

### Remaining audit passes

**Silent-failure** — clear. `dacite` rejects an out-of-vocabulary `method` (tested), and a value that got past config validation raises `NotImplementedError` at the write (tested). Worth noting in the PR's favour: the new path divides by `heat_capacity_per_area`, a fixed positive geometric constant, where `scaled_temperature` divides by `global_gen_ocean_heat_content`, a network-dependent quantity that can in principle pass through zero. Both divisions are unguarded; the new one is the more robust of the two.

**Behavior-change** — clear. The only modified existing paths are the `method` validation (`!=` → `not in`, same accept/reject for every previously-valid value), the ratio hoist (bit-identical, §4), and the protocol widening (dispatch unchanged, §3). No log lines touched, no fields dropped.

**Sibling/mirror** — clear. No `fme/coupled` mirror of the ocean corrector; it is shared. `configs/experiments/2025-05-13-ocean-heat-corrector/ace-train-config.yaml` still uses `method: scaled_temperature` and still parses. Nothing in `docs/` names the method.

**Config back-compat** — clear. `Literal` widened, never narrowed, so every prior checkpoint and inference YAML still loads. The legacy `ocean_heat_content_correction: true` bool migration in `from_state` still maps to `scaled_temperature` (tested).

**Split heuristic** — clear. 3 files, one feature, no refactor or back-compat shim bundled in. Don't split.

**Abstract-or-final / inheritance depth** — clear. `OceanHeatContentCorrection` has no in-repo subclass; nothing new inherits.

**AI-agent artifact** — clear. Imports at module top, no `Any` returns in src, no abbreviated identifiers, no `isinstance` + `type: ignore`. The one `cast(Any, "not_a_method")` is in a test and is the sanctioned way to feed an invalid `Literal`.

**Builder pattern** — no *new* violation. `OceanCorrectorConfig._build` reading `self.ocean_heat_content_correction.method` is a parent-reads-child-config violation, but those lines are unchanged from `main`. `OceanHeatContentCorrection` itself receives built collaborators plus scalars, never a sub-config.

### For the reviewer

- The `_force_conserve_ocean_heat_content` docstring is now ~60 lines. That is a lot, but it is answering the two questions a reviewer will actually have (why the denominator must come from `depth_integral`; why the cell test is the mask and not `dz`), and both were subtle enough to need a second pass to get right. Worth a judgement call on whether to trim.
- I could not independently reproduce the end-to-end 1°-grid numbers (703,524 cells shifted, 32,056 zero-thickness, 0.06021 K increment) — no checkpoint on this machine. My verification is analytic plus a synthetic fixture reproducing the same mask/bathymetry pathology, and it agrees with those numbers: exactness follows from the identity in §1 regardless of grid, and the zero-thickness cells are shifted by construction.

### On the launched run

`685b7413e` is an ancestor of the experiment branch, so the run carries the fixed code. `79fc87fa9` is docstring, comment, and one new test only — no behavior change, and `scaled_temperature` and `uniform_temperature` outputs are unaffected. **The run does not need to be killed or relaunched.**

fme/core/corrector/ocean.py
```diff
@@ -107,7 +114,7 @@ class OceanHeatContentBudgetConfig:
 
     """
 
-    method: Literal["scaled_temperature"]
+    method: Literal["scaled_temperature", "uniform_temperature"]
```
jpdunc23, line 117: Nit: `"uniform_temperature"` is good, but `"constant_temperature"` is consistent the atmosphere energy corrector's naming.

elynnwu: Let's keep it as `uniform_temperature` since it's a better description.

elynnwu:
PR title and description rewritten for conciseness by Claude Opus 4.6 (1M context) (`claude-opus-4-6[1m]`), which matches the model pinned in `.claude/skills/write-pr-prose/`.

Note: the gather script's GraphQL query for title/description hit a Projects-classic deprecation warning (`(warning: could not retrieve title/description)`), but the title and body were retrieved successfully via `gh pr view --json`.

elynnwu:
PR description amended: replaced the sentence "`scaled_temperature` is bit-for-bit unchanged and remains the default." with "`scaled_temperature` is bit-for-bit unchanged and remains what every existing config gets, including the legacy `ocean_heat_content_correction: true` migration." — `OceanHeatContentBudgetConfig.method` is a required field with no default, so "remains the default" was inaccurate.

Model: Claude Opus 4.6 (1M context) (`claude-opus-4-6[1m]`), matching the pin in `.claude/skills/write-pr-prose/`.

Context gap: `gather_pr_context.sh` could not retrieve the title/description via GraphQL due to the Projects Classic deprecation warning; the existing prose was fetched via the REST API instead.

PR #1483, opened 2026-09-08
Add example for uploading a dataset to a HF bucket via the `hf` CLI

spencerkclark:
This PR provides an example lightweight Beaker image and gantry submission script for uploading a zarr store on WEKA to a Hugging Face bucket.  This approach uses the `hf` command line interface, which is documented [here](https://huggingface.co/docs/huggingface_hub/en/guides/cli).  The scratch bucket this job writes to can be viewed [here](https://huggingface.co/buckets/allenai/ai2cm-scratch) and  [this is a link](https://beaker.org/orgs/ai2/workspaces/ace/work/01M215MYFH3RTKPZ73DDWVD8VM?taskId=01M215MYFQFZ707RN9141TGW6R&jobId=01M215MYK7VQWY0NSFMHER3TT1) to the successful Beaker job.

Changes:
- Adds a `scripts/upload_to_hugging_face` directory with a `Makefile`, `Dockerfile`, and `example.sh` script.

### DIFF
--- a/scripts/upload_to_hugging_face/Dockerfile
+++ b/scripts/upload_to_hugging_face/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.12-slim
+
+# Install git and curl since they are required for running with gantry.
+RUN apt-get update && apt-get install -y --no-install-recommends git curl
+
+# Install the Hugging Face CLI and ensure it is in the search path.
+RUN curl -LsSf https://hf.co/cli/install.sh | bash
+ENV PATH="/root/.local/bin:$PATH"
+
+# Set environment variables for performant uploads per Paul Laskowski.
+ENV HF_XET_HIGH_PERFORMANCE=1
+ENV HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB
+
+ENTRYPOINT ["/bin/bash"]
--- a/scripts/upload_to_hugging_face/Dockerfile
+++ b/scripts/upload_to_hugging_face/Dockerfile
@@ -12,6 +12,9 @@ ENV PATH="/root/.local/bin:$PATH"
 # uploads; HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB is larger than the default 16GB,
 # which also facilitates performance. See documentation here for more details:
 # https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables
+
+# Running without both results in roughly 10x slower peak upload throughput for
+# the same store on ai2/phobos.
 ENV HF_XET_HIGH_PERFORMANCE=1
 ENV HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB
 
--- a/scripts/upload_to_hugging_face/Dockerfile
+++ b/scripts/upload_to_hugging_face/Dockerfile
@@ -7,7 +7,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends git curl
 RUN curl -LsSf https://hf.co/cli/install.sh | bash
 ENV PATH="/root/.local/bin:$PATH"
 
-# Set environment variables for performant uploads per Paul Laskowski.
+# Set environment variables for performant uploads based on recommendation from
+# Paul Laskowski. HF_XET_HIGH_PERFORMANCE=1 enables optimal parallelization of
+# uploads; HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB is larger than the default 16GB,
+# which also facilitates performance. See documentation here for more details:
+# https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables
 ENV HF_XET_HIGH_PERFORMANCE=1
 ENV HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB
 
--- a/scripts/upload_to_hugging_face/example.sh
+++ b/scripts/upload_to_hugging_face/example.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -e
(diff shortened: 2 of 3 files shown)


scripts/upload_to_hugging_face/example.sh
```diff
@@ -0,0 +1,30 @@
...
+    --priority high \
+    --cluster ai2/phobos \
+    --env-secret HF_TOKEN=hugging-face-token \
+    --gpus "${N_GPUS}" \
+    --shared-memory 64GiB \
+    --min-runtime 8h \
+    --no-python \
+    --allow-dirty \
+    --weka climate-default:/climate-default \
+    -- hf sync "${SOURCE}" "${DESTINATION}"
```
brianhenn, line 29: Nice to see how simple this is in comparison with ~1 year ago when it seemed that managing files with git/git-lfs was needed.

brianhenn: Ah, realizing that this destination is a "bucket" not a "dataset", so that's a difference. But it seems like it's possible to publish a snapshot of a bucket to a dataset ie to our ACE collection (and also seems like the HF CLI has improved in terms of both destinations).

scripts/upload_to_hugging_face/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# Install git and curl since they are required for running with gantry.
+RUN apt-get update && apt-get install -y --no-install-recommends git curl
+
+# Install the Hugging Face CLI and ensure it is in the search path.
+RUN curl -LsSf https://hf.co/cli/install.sh | bash
+ENV PATH="/root/.local/bin:$PATH"
+
+# Set environment variables for performant uploads per Paul Laskowski.
+ENV HF_XET_HIGH_PERFORMANCE=1
+ENV HF_XET_SHARD_CACHE_SIZE_LIMIT=40GiB
```
frodre, line 19: If you have any insights into why these are better, would be nice to include in the comment.  No worries if it was just passed along.

scripts/upload_to_hugging_face/example.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd $REPO_ROOT
+
+N_GPUS=0
```
frodre: Since we don't need GPUs I would hardcode this below so it doesn't seem configurable to others/agents

scripts/upload_to_hugging_face/example.sh
```diff
@@ -0,0 +1,30 @@
...
+    --name "${JOB_NAME}" \
+    --description 'Sync dataset on WEKA with a Hugging Face bucket' \
+    --beaker-image "${BEAKER_IMAGE}" \
+    --workspace ai2/ace \
+    --priority high \
+    --cluster ai2/phobos \
+    --env-secret HF_TOKEN=hugging-face-token \
+    --gpus "${N_GPUS}" \
+    --shared-memory 64GiB \
+    --min-runtime 8h \
```
frodre, line 25: I assume these generally have no issue being scheduled without a GPU request.

frodre (approved):
LGTM!  Thanks for sharing this example.

scripts/upload_to_hugging_face/example.sh
spencerkclark: That's right, yeah, I did not have any trouble scheduling these when doing the full upload on Friday.

scripts/upload_to_hugging_face/example.sh
spencerkclark: Good thought.

scripts/upload_to_hugging_face/Dockerfile
spencerkclark: I added a couple comments.  Running without both results in a roughly 10x slower peak upload throughput for the same store on `ai2/phobos` (compate logs in [this job without them](https://beaker.org/orgs/ai2/workspaces/ace/work/01M235AXH55NYNEP1SP3DNDPE3?taskId=01M235AXHKQX7XWNTCG8QYFWW9&jobId=01M235AXNBGSXP2G0GRC7VMX2G) to [this job with them](https://beaker.org/orgs/ai2/workspaces/ace/work/01M23AH5NN6BRQKQ1VX7BHH8Q7?taskId=01M23AH5NW99JCDAKR68DGWWY8&jobId=01M23AH5WTZYAV4F6JM873BKS6)).  If I had to guess, `HF_XET_HIGH_PERFORMANCE=1` is the most important, but I have not done an ablation of each.

scripts/upload_to_hugging_face/example.sh
spencerkclark: There are some limitations with Hugging Face datasets that prevent working well with zarr (e.g. number of files per directory for traditional stores or file size limits for zip stores), which is why we went with the bucket approach for the time being. See comments in https://github.com/huggingface/datasets/issues/4096 and https://github.com/google-research/weatherbench2/issues/226.

brianhenn: Thanks good to know

spencerkclark: [Here's a more authoritative reference](https://huggingface.co/docs/hub/en/storage-limits) from the Hugging Face documentation for these limits.  The file size limit is 500 GB, contrary to the 50 GB noted in https://github.com/google-research/weatherbench2/issues/226#issuecomment-2606777772, but that's still prohibitive for us in many cases if we were to try to use a zip store.

PR #1485, opened 2026-09-09
ACE data writer benchmarking

brianhenn:
Inference output write cost was a single aggregate `data_writer` timer. This adds a `data_writer_io` key that wraps only the storage-library calls in the zarr and netCDF writers (raw and monthly), always recorded inside `data_writer`, plus a standalone write benchmark to measure zarr write throughput of the production writer stack against weka versus GCS for the appending and monthly-mean writers; the netCDF writers are timed for a consistent `data_writer` breakdown, not as a benchmark target. The remaining `data_writer` time covers variable creation, the monthly fold, and zarr store initialization. Initial-condition and restart writes in `PairedDataWriter.write` and the core `fme.core.writer.ZarrWriter` stay outside `data_writer_io`, so downscaling output is not timed. The netCDF writers now convert every variable in a window to numpy before the timed block, so peak host memory rises by about one window of data.

The benchmark (`python -m fme.ace.inference.data_writer.benchmark`) is YAML-configured with a `DataWriterConfig` and writes prediction-side payload only through the production writer stack. Each run writes into its own timestamped subdirectory of `experiment_dir`. It reports the final summary to both the text log and W&B, and raises if no `data_writer_io` time was recorded.

Changes:
- `fme.ace.inference.data_writer.raw` and `monthly`: netCDF storage calls gathered into contiguous `data_writer_io` timer blocks
- `fme.ace.inference.data_writer.zarr` and `monthly_zarr`: `record_batch` and `read_batch` wrapped in `data_writer_io`; `initialize_store` moved out of the timed path
- `fme.ace.inference.data_writer.benchmark`: new `BenchmarkConfig` and `benchmark()` entrypoint; `__post_init__` rejects `save_prediction_files`, `save_monthly_files`, `save_step_diagnostics`, and per-file `save_reference`
- `configs/experiments/2026-09-04-ace-data-writing-benchmark/`: weka and GCS configs for the appending zarr writer and the monthly-mean zarr writer, plus gantry `run.sh` launching all four arms

- [x] Tests added

### DIFF
--- a/fme/ace/inference/data_writer/benchmark.py
+++ b/fme/ace/inference/data_writer/benchmark.py
@@ -1,0 +1,213 @@
+"""This module provides an entrypoint for benchmarking data writing performance.
+It writes real data loaded from a configured dataset, timing the writers
+separately from the loader that feeds them.
+"""
+
+import argparse
+import dataclasses
+import logging
+import os
+import shutil
+import time
+import uuid
+
+import dacite
+import torch
+import yaml
+
+from fme.ace.data_loading.batch_data import PairedData
+from fme.ace.data_loading.getters import get_inference_data
+from fme.ace.data_loading.gridded_data import InferenceGriddedData
+from fme.ace.data_loading.inference import InferenceDataLoaderConfig
+from fme.ace.inference.data_writer.dataset_metadata import DatasetMetadata
+from fme.ace.inference.data_writer.main import DataWriterConfig, PairedDataWriter
+from fme.ace.requirements import DataRequirements, PrognosticStateDataRequirements
+from fme.core import logging_utils
+from fme.core.cloud import makedirs
+from fme.core.dicts import to_flat_dict
+from fme.core.distributed.distributed import Distributed
+from fme.core.timing import GlobalTimer
+from fme.core.wandb import WandB
+
+TMPDIR = f"/tmp/ace_write_benchmark_{uuid.uuid4()}"
+
+
+@dataclasses.dataclass
+class BenchmarkConfig:
+    """
+    Configuration for benchmarking data writing.
+
+    Parameters:
+        experiment_dir: Directory to write output to. May be local or a remote
+            path recognized by fsspec, such as ``gs://bucket/results``.
+        loader: Parameters for the inference data loader supplying the windows
+            that are written. The inference loader is required because its
+            windows tile the time axis, whereas training loader windows start at
+            consecutive indices and overlap, leaving no coherent time slot for
+            the writers to fill.
+        data_writer: Configuration for the data writers under test.
+        logging: Configuration for logging.
+        names: Names of the variables to load and write.
+        n_forward_steps: Total number of timesteps to write.
+        forward_steps_in_memory: Number of timesteps written per window.
+    """
+
+    experiment_dir: str
+    loader: InferenceDataLoaderConfig
+    data_writer: DataWriterConfig
+    logging: logging_utils.LoggingConfig
+    names: list[str]
+    n_forward_steps: int
+    forward_steps_in_memory: int
+
+    def __post_init__(self):
+        self.data_writer.validate_time_coarsen(
+            self.forward_steps_in_memory, self.n_forward_steps
+        )
+        self._validate_writers()
+
+    def _validate_writers(self):
+        if (
+            self.data_writer.save_prediction_files
+            or self.data_writer.save_monthly_files
+        ):
+            raise ValueError(
+                "save_prediction_files and save_monthly_files must be False. The "
+                "benchmark provides no reference data, so writers of it would "
+                "produce empty output. Configure writers through 'files' instead."
+            )
+        if self.data_writer.save_step_diagnostics:
+            raise ValueError(
+                "save_step_diagnostics must be False. The benchmark runs no "
+                "stepper, so the diagnostics writer would produce empty output."
+            )
+        for file_config in self.data_writer.files or []:
+            if file_config.save_reference:
+                raise ValueError(
+                    f"File writer {file_config.label!r} sets save_reference, but the "
+                    "benchmark provides no reference data. Set it to False."
+                )
+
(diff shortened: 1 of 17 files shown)


brianhenn:
PR description updated by the `write-pr-prose` skill, running on model **claude-opus-4-6[1m]**. No warnings from the context-gathering script; full diff, comments, and reviews were available.

brianhenn:
PR description updated by the `write-pr-prose` skill, running on model **claude-opus-4-6[1m]**. No warnings from the context-gathering script; full diff, comments, and reviews were available. This update was written from the local (rewritten) branch history, not the GitHub-side diff.

fme/ace/inference/data_writer/benchmark.py
```diff
@@ -0,0 +1,205 @@
...
+            the writers to fill.
+        data_writer: Configuration for the data writers under test.
+        logging: Configuration for logging.
+        names: Names of the variables to load and write.
+        n_forward_steps: Total number of timesteps to write.
+        forward_steps_in_memory: Number of timesteps written per window.
+    """
+
+    experiment_dir: str
+    loader: InferenceDataLoaderConfig
```
oliverwm1, line 56: What's the rationale to load real data instead of just generating random data on-the-fly? I guess cost of compression step will be more realistic.

brianhenn: Could do either, but loading real data didn't seem to have big downsides (slightly slows the test, but the timers correctly capture the write portion).

fme/downscaling/inference/inference.py
```diff
@@ -255,4 +256,7 @@ def main(config_path: str):
     logging.info("Starting downscaling generation...")
     downscaler = generation_config.build()
     logging.info(f"Number of parameters: {count_parameters(downscaler.model.modules)}")
-    downscaler.run_all()
+    with GlobalTimer():
```
brianhenn: Side effect of the core ZarrWriter including timing now is that downscaling entrypoints need to introduce the GlobalTimer context to avoid a "no global timer" warning; now they get timing of just the data writes. If this is annoying there are probably other ways around this.

fme/ace/inference/data_writer/monthly.py
```diff
@@ -247,13 +249,15 @@ def append_batch(
         old_size = self.dataset.variables[LEAD_TIME_DIM].size
         new_size = month_min + month_range
 
-        self._extend_lead_time(old_size, new_size)
-        self._extend_valid_time(old_size, new_size)
-        self._extend_variable(COUNTS, old_size, new_size, initial_value=0)
+        with timer.context("storage_write"):
```
brianhenn: We're not actively timing the netCDF writer in the benchmarks so could delete this and in the raw writer, but seems more consistent to include it.

fme/ace/inference/data_writer/monthly.py
```diff
@@ -278,26 +282,31 @@ def append_batch(
...
             add_data(
                 target=month_data,
                 target_start_counts=count_data,
                 source=array,
                 months_elapsed=months - month_min,
             )
-            self.dataset.variables[variable_name][
-                :, month_min : month_min + month_range
-            ] = month_data
+            with timer.context("storage_write"):
```
oliverwm1, line 299: I think this is a bug. The write really happens when `self.dataset.sync()` is called below.

My suggestion is to remove these `storage_write` and `storage_read` timers since it is a bit unobvious exactly which lines of code they should wrap, and there seem to be some differences between the netCDF and zarr writers.

oliverwm1 (commented):
Did you find the `storage_write` and `storage_read` timers useful? My recommendation is to remove these since it is not that obvious exactly what code lines they should include or not include.

fme/ace/inference/data_writer/monthly.py
oliverwm1: Nevermind, I missed the new indentation, so looks like the sync call is included. I still find it a bit arbitrary what gets included here—what do you think?

fme/ace/inference/data_writer/zarr.py
```diff
@@ -338,22 +346,29 @@ def append_batch(
...
     ) -> None:
+        vars = self.data_vars or list(data.keys())
+        numpy_data = {k: v.cpu().numpy() for k, v in data.items() if k in vars}
         # Zarr store initialization needs the full time coordinate information,
         # which is not available until the first batch is seen.
         if self._writers is None:
             self._initialize_writers(batch_time)
-        vars = self.data_vars or list(data.keys())
+            for writer in self.writers:
+                writer.initialize_store(
```
brianhenn, line 356: This is called here instead of previously on the first `writer.record_ batch` call for cleaner timing.

fme/ace/inference/data_writer/monthly.py
```diff
@@ -271,33 +266,49 @@ def append_batch(
...
                     [INIT_TIME, VALID_TIME, COUNTS]
                 )
+            arrays[variable_name] = data[variable_name].detach().cpu().numpy()
 
-            array = data[variable_name].detach().cpu().numpy()
-
-            # Add the data to the variable totals
+        timer = GlobalTimer.get_instance()
+        month_data = {}
+        with timer.context("data_writer_io"):
```
brianhenn, line 273: Again just reorganization to get to fewer timer calls, one for reading the existing monthly values and counts, and one for writing the new ones (separate by the add, which is excluded). The one difference is that we move all variables off GPU at once, instead of one at a time, so the CPU memory burden is slightly higher. We could revert that with an extra set of timing calls, I'm fine either way.

fme/ace/inference/data_writer/monthly.py
brianhenn: With the reorganization hopefully it is clearer what is now included in the "data_writer_io" timer for both netCDF and zarr.

oliverwm1 (approved):
LGTM

PR #1486, opened 2026-09-09
Concurrent zarr variable writes

brianhenn:
Write all zarr variables in a batch concurrently instead of one at a time, using `asyncio.gather` through zarr's existing `zarr.core.sync.sync` bridge. This turns latency-bound object-storage writes (GCS) into throughput-bound ones, yielding a 4x speedup on GCS.

Changes:
- `fme.core.writer._insert_into_zarr` and `_read_from_zarr`: keep their signatures but now resolve per-variable `AsyncArray` handles with one `asyncio.gather`, then write (or read) every variable's slice concurrently. `_check_for_overwrite` became async (`await arr.getitem`, `arr.metadata.fill_value`)
- Behavior change: all validation (overwrite checks and data-size checks) now completes for every variable before any write starts, so a conflict or shape mismatch leaves the store unchanged. The previous sequential loop validated-then-wrote per variable, so a failure on a later variable left earlier ones already written
- Safe by construction: gathering is across distinct arrays that never share a chunk; ACE's layout writes one complete shard per variable per window

Performance measured with #1485's benchmark after its timer rework, 27.5 GB output, 53 variables x 2 initial conditions x 20 windows, on phobos. Control (benchmark branch head af8c84477) and treatment (this branch with the benchmark branch merged, bff850e14) launched 40 minutes apart on 2026-09-17. Throughput is payload per window divided by the median per-window time, robust to single-window stalls.

| Backend | Control (MB/s) | Concurrent (MB/s) | Speedup |
|---------|---------------:|-------------------:|--------:|
| GCS     |           41.6 |              151.2 |    3.6x |
| weka    |          215.1 |              279.6 |    1.3x |

pr2-write-throughput-v2

The whole-run data_writer_io aggregate gives GCS 16.8 -> 147.2 and weka 220.6 -> 297.7 MB/s (the black ticks in the figure). The GCS control aggregate is dragged down by one window whose uploads stalled for 16 minutes (990 s vs a 33 s median), exactly the exposure a sequential writer has and what the median-window figure removes. These reproduce the September 9 pair (38 -> 159 GCS, 240 -> 301 weka). Weka sits within its run-to-run spread. GCS is the reliable signal because object storage is latency-bound, which is what the gather addresses.

- [x] Tests added

### DIFF
--- a/fme/core/writer.py
+++ b/fme/core/writer.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from collections.abc import Mapping, Sequence
 from typing import Literal
@@ -7,6 +8,7 @@ import fsspec
 import numpy as np
 import xarray as xr
 import zarr
+from zarr.core.sync import sync
 
 from fme.core.distributed import Distributed
 from fme.core.timing import GlobalTimer
@@ -21,11 +23,11 @@ def _encode_cftime_times(times, calendar="julian"):
     return cftime.date2num(times, units=DATETIME_ENCODING_UNITS, calendar=calendar)
 
 
-def _check_for_overwrite(arr, insert_slices_tuple):
+async def _check_for_overwrite(arr, insert_slices_tuple):
     # Loads the slice of zarr array
-    existing = arr[insert_slices_tuple]
+    existing = await arr.getitem(insert_slices_tuple)
 
-    fill_value = arr.fill_value
+    fill_value = arr.metadata.fill_value
 
     if fill_value is None:
         # If no fill_value, assume all entries are valid → forbid any overwrite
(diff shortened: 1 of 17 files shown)


brianhenn:
PR description updated by the `write-pr-prose` skill, running on model **claude-opus-4-6[1m]**. No warnings from the context-gathering script; full diff, comments, and reviews were available.

brianhenn:
**Intended merge order: #1485 first, then this PR.**

The dependency is textual rather than semantic — this branch changes `_insert_into_zarr` / `_read_from_zarr` in `fme/core/writer.py`, while #1485 touches `record_batch` / `read_batch` and separate files. Either could merge first; taking #1485 first just avoids a rebase here.

Because this PR is based on `main`, its diff currently also shows #1485's commits. Once #1485 squash-merges, the diff here collapses to `fme/core/writer.py` and its tests.

The throughput figures in the description were measured with the benchmark entrypoint added in #1485, so they are not reproducible from this branch alone until that merges.

brianhenn:
Pre-review agent: notes for the reviewer. Nothing blocking.

**`asyncio.gather` does not cancel siblings on the first exception.** If one `setitem` fails, the remaining writes keep running on zarr's event-loop thread after `sync()` re-raises, so writes can land after the error surfaces. The sequential version also left the store partial on failure, but deterministically — everything before the failing variable. The partial state is now nondeterministic. Likely fine since a write failure aborts the run, but it is a real change in the error path.

**Equivalences I verified so you do not have to:** `arr.metadata.fill_value` is exactly what `Array.fill_value` returns in zarr 3.3.0, and `AsyncArray.name` / `.shape` match the sync versions, so the `RuntimeError` message strings are unchanged. `_check_for_overwrite` has one in-module caller, so making it async is contained.

**On `zarr.core.sync.sync` being private:** the pre-PR `zarr_array[slices] = data` already called it. Zarr 3's sync `Array` is a thin wrapper over an async one — `__setitem__` -> `set_basic_selection` -> `sync(self.async_array._set_selection(...))` — on the same module-global loop and `zarr_io` thread. So the runtime coupling is unchanged by this PR; only the import is new, moving the `sync` boundary from once per variable to once per batch. A relocation would break zarr's own public sync API too, and would surface as an ImportError at startup. `asyncio.run` would be worse, not more public: it would run a second loop alongside the one `_initialize_zarr` and `consolidate_metadata` still use, and skip the executor, fork and atexit handling `sync` sets up.

**Test fix applied during pre-review.** The overwrite-conflict test put the conflict on `var0`, first in dict order, so the old check-then-write loop also raised before writing anything — the test passed against the pre-PR implementation. It now places the conflict last, which is what actually distinguishes check-all-then-write from check-and-write-per-variable.

**Smaller notes.** Peak *host* RAM during a write rises by roughly one extra copy of the batch (measured +618 MB on a 687 MB, 53-variable batch), since all encode buffers are live at once rather than one variable at a time; transient, GPU untouched, and it does not compound across ICs. The per-IC loop at `data_writer/zarr.py:353` is still sequential — the next throughput lever, needing a bigger refactor.

brianhenn:
PR description updated by the `write-pr-prose` skill, running on model **claude-opus-4-6[1m]**. Change: reworded the "Behavior change" bullet to cover all validation (overwrite checks and data-size checks), not just overwrite checks. No warnings from the context-gathering script.

brianhenn:
Merge #1485 into main before this one. This PR's base is `feature/ace-data-writing-benchmark`, so merging it now would fold this change into #1485 rather than into main.

fme/core/writer.py
```diff
@@ -60,42 +62,80 @@ def _check_data_size_fits_slice(data: np.ndarray, insert_slices: Mapping[int, sl
...
+    insert_slices: Mapping[int, slice],
+    overwrite_check: bool,
+):
+    """Write every variable's slice concurrently.
+
+    Overwrite checks all complete before any write starts, so a conflict
+    leaves the store untouched.
+    """
+    names = list(data)
+    group = await zarr.api.asynchronous.open_group(store=path, mode="r+")
```
frodre: I would consider moving an import `from zarr.api.asynchronus import open_group` just so it fails loudly if anything changes with future zarr versions.  Claude was chirping about unstable api.

fme/core/writer.py
```diff
@@ -60,42 +62,80 @@ def _check_data_size_fits_slice(data: np.ndarray, insert_slices: Mapping[int, sl
...
+        )
+    if overwrite_check:
+        await asyncio.gather(
+            *(
+                _check_for_overwrite(arr, slices_tuple)
+                for arr, slices_tuple in zip(arrays, slices_tuples)
+            )
+        )
+    await asyncio.gather(
+        *(
```
frodre, line 98: More from chatting w/ Claude.  This is relevant only if we ever implement restarts w/ inference writers, but  leaving a comment so it surfaces in the discussion w/ bots.  I don't think it's worth anything more than a comment, if that.  

Our writes are not atomic, so one variable could potentially fail while the others complete.  The overwrite check would then fail on restart because there's no cancel sent to other async variables and they're allowed to complete.  If we want to have preemption, we would probably have to also make sure these writes are all or nothing.

frodre (approved):
LGTM, thanks @brianhenn

fme/core/writer.py
brianhenn: Yeah, I think we can leave this as is for now. Not actually different from our current for-loop over variables in terms of preemption state.

PR #1487, opened 2026-09-09
Add per-variable residual prediction and residual-normalized stepping configuration

Arcomano1234:
This PR allows configurable control of which prognostic variables are residual or full field, previously we had full field predictions by default and residual_prediction bool that would switch all prognostic variables to a residual prediction.  There is no breaking changes for previous full field models or models trained with residual_prediction: true (e.g. all variables residual). Adds a `SingleModuleStepConfig` option that lets a stepper predict some prognostics as tendencies and the rest full-field, and lets those tendencies be expressed in tendency-normalized units.

Changes:
- `fme.core.step.single_module.ResidualPredictionConfig`: new dataclass with `names: list[str] | None` (which prognostics to step as residuals; `None` means all) and `normalized: bool` (treat residual outputs as tendencies in `normalization.residual` units so a unit network output is one tendency standard deviation). The transform is scale-only by design: the stats convention pairs full-field centering with tendency stds, so residual means are never applied and any nonzero mean tendency is learned through the network output. The `names` docstring now states that naming a subset also narrows tendency-std loss scaling to that subset: the remaining prognostics are scored in full-field-std units. `ResidualPredictionConfig.__post_init__` rejects `names: []`, which would read as the default but silently disable residual stepping.
- `fme.core.step.single_module.SingleModuleStepConfig`: replace the flat `residual_prediction: bool` with `residual_prediction: ResidualPredictionConfig | None = None`. Typed against `None` rather than kept as a union with `bool` deliberately: dacite swallows a union dataclass member's `__post_init__` `ValueError` and re-raises `UnionMatchError`, so `ResidualPredictionConfig | bool` would have destroyed every validation message this config raises.
- `SingleModuleStepConfig.remove_deprecated_keys`: migrate the legacy `bool` to `ResidualPredictionConfig() | None`. Both checkpoints and user YAML reach this hook via `from_state` on the `StepConfigABC` base class, so existing configs load unchanged.
- `SingleModuleStepConfig.__post_init__`: reject a `residual_prediction.names` entry that is not prognostic (it has no input to add its residual to), `residual_prediction.normalized` without a `normalization.residual` block, and `residual_prediction.normalized` with an explicit `normalization.loss` block. The last check raises one clear error naming the option, instead of the previous two-step dead end (a missing-residual-block error, then a loss/residual-conflict error, neither of which names the option that forced the conflict). The deprecated `bool` is also coerced on direct Python construction (the config is public API, exported from `fme.ace`), not only via the serialized-state hook.
- `SingleModuleStepConfig.residual_names`: property that resolves the config to a concrete name set. `get_loss_normalizer` uses it so that residual-stepped names are scored in tendency-std units and full-field names in full-field-std units.
- `fme.core.step.step.StepConfigABC.residual_names`: new property (default: every prognostic) so that `MultiCallStepConfig.get_loss_normalizer` can score a multi-called variant by its base variable's residual convention. `SingleModuleStepConfig` overrides it to respect `ResidualPredictionConfig.names`. Previously, variants of a full-field prognostic in a hybrid step were scored in tendency-std units while the base variable used field-std units.
- `fme.core.step.single_module.step_with_adjustments`: the residual feature is now a single `residual_names: Collection | None` parameter (`None` disables it) plus an optional `residual_transform` scale. The bool and prognostic-names fallback are gone; other callers (`radiation`, `secondary_module`, `fcn3`) pass their prognostics when their own `residual_prediction` bool is set. The transform indexes strictly on both the network output and the scale dict: a missing name is an invariant violation, not a silent full-field fallback.
- Tests cover the residual-normalized step, the hybrid subset, the two combined, the per-variable loss normalizer (including multi-call variant scaling), config validation (including the `normalization.loss` conflict and empty `names`), loading a legacy `bool` config (both serialized state and direct construction), and the `residual_names` restriction to prognostics.

- [x] Tests added


elynnwu:
> One downstream contract worth knowing: weights carry no output-convention semantics, so a config that loads a checkpoint's weights but declares its own step config must declare the same residual_prediction_names the checkpoint was trained with.

This feels like something we should address to avoid mis-configuration. We can simply compare the checkpoint's `residual_prediction`/`residual_prediction_names`/`residual_normalized_prediction` against the declared config and raise (or at minimum log a warning) on mismatch. I lean towards dis-allowing it.

elynnwu:
Can you look into this? This case is not documented.

> Claude: `NetworkAndLossNormalizationConfig` forbids loss and residual together, you cannot use `residual_normalized_prediction` with an explicit loss normalization. Opting into the prediction transform forces you into the residual loss convention.

Arcomano1234:
PR description updated to reflect commit 9d039d6a7: the `__post_init__` bullet now documents the explicit check for `residual_normalized_prediction` combined with `normalization.loss`, which raises one clear error naming the option instead of the previous two-step dead end, plus the corresponding test.

Model: claude-opus-4-6[1m] (Opus 4.6, 1M context). No warnings from the context-gathering script.

jpdunc23:
Responding to @elynnwu's [comment](https://github.com/ai2cm/ace/pull/1487#issuecomment-5625640832).

I agree there is a risk of unintentional misconfiguration of fine-tuning jobs, but this is also a risk for other parts of the stepper config, e.g. the `in_names` and `out_names`. We have at least two mechanisms to minimize these risks: `job_runners`'s `create_finetune_config.sh` and `CheckpointStepperConfig`.

I think the sort of validation you're proposing requires some design discussion which would weigh down this PR too much. 

IMO, we don't need that validation. Some intentional changes to a fine-tuning run's step config are clearly desirable, e.g. this is the only way to fine-tune a CM4 checkpoint on the E3SMv3 data with differing variable naming conventions. Maybe fine-tuning a set of non-residual model weights with a residual config would actually work? 

Nor do I think we need the "One downstream contract worth knowing" sentence in the PR description, since this is automatically true for just about anything we could add to `SingleModuleStepConfig` or the like.

jpdunc23 (commented):
High-level, as requested: Code comments and docs are pretty LLM verbose.

The configuration is a bit gnarly with 3 coupled params in `residual_prediction`, `residual_prediction_names` and `residual_normalized_prediction`. I wonder if we could instead get something like `residual_prediction: ResidualPredictionConfig | None = None` and handle backwards compat for checkpoints with `residual_prediction: bool`. Another option is `residual_prediction: ResidualPredictionConfig | bool = False`.

Arcomano1234:
> High-level, as requested: Code comments and docs are pretty LLM verbose.
The configuration is a bit gnarly with 3 coupled params in residual_prediction, residual_prediction_names and residual_normalized_prediction. I wonder if we could instead get something like residual_prediction: ResidualPredictionConfig | None = None and handle backwards compat for checkpoints with residual_prediction: bool. Another option is residual_prediction: ResidualPredictionConfig | bool = False.

Strongly agree with this, thank you for suggesting this. I'll iterate on this idea.

Arcomano1234:
PR description updated to reflect commit 845818772: the three flat fields are now a single `residual_prediction: ResidualPredictionConfig | None = None`, the legacy bool migration is documented, and the "One downstream contract worth knowing" sentence is removed per reviewer request.

Model: claude-opus-4-6[1m] (Opus 4.6, 1M context). No warnings from the context-gathering script.

elynnwu:
> I think the sort of validation you're proposing requires some design discussion which would weigh down this PR too much.

Yes, Troy and I have decided to punt on this and not expand the scope of this PR.

fme/core/step/single_module.py
```diff
@@ -95,13 +120,38 @@ class SingleModuleStepConfig(StepConfigABC):
     )
     next_step_forcing_names: list[str] = dataclasses.field(default_factory=list)
     prescribed_prognostic_names: list[str] = dataclasses.field(default_factory=list)
-    residual_prediction: bool = False
+    residual_prediction: ResidualPredictionConfig | None = None
```
elynnwu, line 142: I think we need to handle backwards compatibility here. Either `ResidualPredictionConfig | bool = False` or handle `residual_prediction: bool` specifically as James suggested.

elynnwu (approved):

Arcomano1234:
PR description updated to reflect commits 12fee59d5, 9fac79924, f70de05b5: added bullet for `StepConfigABC.residual_names` and multi-call loss scaling, updated `step_with_adjustments` bullet for the `Collection | None` signature and strict indexing, added `__post_init__` bool coercion and empty-names rejection, and noted the `names` docstring's loss-scaling narrowing.

Model: claude-opus-4-6[1m] (Opus 4.6, 1M context). No warnings from the context-gathering script.

fme/core/step/single_module.py
Arcomano1234: marking this as resolved everything should be handled fine with `residual_prediction: ResidualPredictionConfig | None = None`

PR #1490, opened 2026-09-09
Make enso_coefficient cross-process reduction unconditional

Arcomano1234:
The ENSO coefficient aggregator conditionally entered cross-process reductions only when a process had samples overlapping the reference index. Whether a process has overlapping data is data-dependent, so when some processes had data and others did not, the processes with data would block in the collective waiting for processes that never entered it, hanging until the collective timeout killed the job.

Every process now takes part in the same collectives unconditionally, weighted by its number of contributing samples, so a process whose samples do not overlap the reference index can neither deadlock nor bias the result.

Changes:
- `EnsoCoefficientEvaluatorAggregator`: track per-variable spatial shapes (`_target_spatial_shapes`, `_gen_spatial_shapes`) on every process so that processes without data know what to reduce
- `EnsoCoefficientEvaluatorAggregator._get_coefficients`: replace conditional reduction with unconditional calls to the new `reduce_sample_coefficients`
- `reduce_sample_coefficients` (replaces `reduce_data`): every process participates in `reduce_sum` over both the coefficient values and the per-variable sample counts, then divides on root to get a properly weighted mean; the `spatial_shapes` parameter tells every process the shape of each variable
- Tests: `test_collectives_do_not_depend_on_index_overlap` verifies overlapping and non-overlapping processes issue the same collective calls; `test_reduce_sample_coefficients_weights_by_sample_count` and `test_reduce_sample_coefficients_none_when_no_process_has_samples` cover the new reduction logic

- [x] Tests added

### DIFF
--- a/fme/ace/aggregator/inference/enso/enso_coefficient.py
+++ b/fme/ace/aggregator/inference/enso/enso_coefficient.py
@@ -106,6 +106,11 @@ class EnsoCoefficientEvaluatorAggregator:
         n_samples = len(self._sample_index_series)
         self._target_covariances: list[TensorDict] = [{} for _ in range(n_samples)]
         self._gen_covariances: list[TensorDict] = [{} for _ in range(n_samples)]
+        # spatial shape of each variable, tracked on every process regardless of
+        # whether any of its samples overlap the reference index, so that a
+        # process without data can still take part in the cross-process reduction
+        self._target_shapes: dict[str, torch.Size] = {}
+        self._gen_shapes: dict[str, torch.Size] = {}
         self._index_variance: list[torch.Tensor] = [
             torch.tensor(0.0, dtype=torch.float32, device=get_device())
             for _ in range(n_samples)
@@ -133,6 +138,12 @@ class EnsoCoefficientEvaluatorAggregator:
         assert time.sizes["sample"] == len(
             self._sample_index_series
         ), "number of index series must match number of samples"
+        for shapes, batch_data in (
+            (self._target_shapes, target_data),
+            (self._gen_shapes, gen_data),
+        ):
+            for name, tensor in batch_data.items():
+                shapes[name] = tensor.shape[2:]
         for i_sample, sample_index_series in enumerate(self._sample_index_series):
             if sample_index_series is not None:
                 sample_index_series_window = sample_index_series.sel(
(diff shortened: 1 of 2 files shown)


Arcomano1234:
PR title and description updated by Claude Code.

**Model**: claude-opus-4-6 (1M context)
**Context gaps**: None — the gather script retrieved the diff, existing prose, and discussion without warnings.

Arcomano1234:
Pre-review agent: audit passes are clean apart from two points worth a reviewer's eye, plus context that may save time.

**1. Accumulator dtype is now pinned to float32 (behavior change).** `reduce_sample_coefficients` builds its accumulator as `torch.zeros(..., dtype=torch.float32)`, so `summed[i] += coefficients[name]` silently downcasts if the coefficients are float64. Previously the reduced tensor inherited the coefficients' dtype. This is not incidental: an inert process has no coefficients to infer a dtype from, and every process must agree on the dtype or the collective itself breaks — so *some* dtype has to be fixed a priori. float32 matches everything inference currently produces. If you'd rather not rely on that, the alternatives are tracking dtype alongside shape in `record_batch`, or asserting float32 on entry. Flagging rather than picking, since it's your call.

**2. The `if not names: return None` early return skips the collectives entirely.** That is only safe because `record_batch` is called the same number of times on every process (so `_target_shapes` / `_gen_shapes` are non-empty everywhere or empty everywhere). That invariant is what the whole fix rests on, and it is currently only a comment. Worth confirming you agree it holds for every inference driver.

**Context that may shorten review:**
- The sum + counts shape here deliberately mirrors the existing `_reduce_datasets` in `fme/ace/aggregator/inference/seasonal.py`, which already reduces a stacked tensor alongside a `counts` tensor. Note `_reduce_datasets` additionally raises a `ValueError` on mismatched variable shapes; this function relies on `torch.stack` to fail instead, matching what the old `reduce_data` did.
- Sibling check on `dynamic_index.py` in the same package: `get_indices()` skips the `gather_irregular` collective via `if sst_name not in self._raw_indices: continue`. That skip keys off the *prediction variable set*, which is identical across processes, so it is schema-dependent rather than data-dependent and does not have this bug. No change needed, but it is the obvious next place a reviewer would look.
- The two behaviors this PR adds were mutation-tested: reintroducing conditional participation fails `test_collectives_do_not_depend_on_index_overlap`, and replacing the per-sample weight with a per-process one fails `test_reduce_sample_coefficients_weights_by_sample_count` (and an existing value test).
- The weighting change alters numbers for existing multi-process runs where processes hold differing numbers of overlapping samples — the old code took an unweighted mean over processes. Intended, and described in the PR body.

fme/ace/aggregator/inference/enso/enso_coefficient.py
```diff
@@ -188,59 +199,16 @@ def _compute_coefficients(
...
-        gen_coefficients = self._compute_coefficients("gen")
-        # average coefficients across samples
-        target_coefficients_all, gen_coefficients_all = {}, {}
-        target_names = set(
-            [
-                name
-                for target_coefficient in target_coefficients
-                for name in target_coefficient.keys()
-            ]
+        # Every process must make the same collective calls in the same order,
```
brianhenn: (nit) Probably don't need the verbose claude comments here as well as in reduce_sample_coefficients, a docstring on the latter seems like the better place to describe this

fme/ace/aggregator/inference/enso/enso_coefficient.py
```diff
@@ -133,6 +138,12 @@ def record_batch(
         assert time.sizes["sample"] == len(
             self._sample_index_series
         ), "number of index series must match number of samples"
+        for shapes, batch_data in (
+            (self._target_shapes, target_data),
+            (self._gen_shapes, gen_data),
+        ):
+            for name, tensor in batch_data.items():
+                shapes[name] = tensor.shape[2:]
```
brianhenn: (nit): For clarity a name like `spatial_shapes` might be better.

fme/ace/aggregator/inference/enso/enso_coefficient.py
```diff
@@ -437,28 +405,63 @@ def data_index_covariance(
     return (data * index_values_broadcast).sum(dim=index_dim)
 
 
-def reduce_data(dist: Distributed, rank_tensor_dict: TensorDict) -> TensorDict | None:
-    """Reduce tensor dicts across distributed processes by taking the mean.
+def reduce_sample_coefficients(
+    dist: Distributed,
+    sample_coefficients: list[TensorDict],
+    shapes: Mapping[str, torch.Size],
```
brianhenn: (nit): Same comment re: `spatial_shapes`

brianhenn (approved):
Thanks for fixing @Arcomano1234 . Some stylistic comments but looks good, though didn't review the test coverage (assumed it added a test that caught this).

fme/ace/aggregator/inference/enso/enso_coefficient.py
Arcomano1234: Done in 82aa13c9c — dropped the comment block, the `reduce_sample_coefficients` docstring is now the only place that explains it.

fme/ace/aggregator/inference/enso/enso_coefficient.py
Arcomano1234: Done in 82aa13c9c — `_target_spatial_shapes` / `_gen_spatial_shapes`, and the loop variable is `spatial_shapes`.

fme/ace/aggregator/inference/enso/enso_coefficient.py
Arcomano1234: Done in 82aa13c9c — the parameter and its docstring entry are `spatial_shapes` too.

Arcomano1234:
Pre-review agent: on the test-coverage point — `test_collectives_do_not_depend_on_index_overlap` is the one that catches this. It builds two aggregators with identical variables, one whose samples fully overlap the reference index and one shifted entirely off the end, records the collectives each would issue through a fake distributed singleton, and asserts the two call sequences are identical (while the non-overlapping one still logs nothing). Reintroducing the conditional participation makes it fail.

Both new behaviors were mutation-tested: making participation conditional again fails that test, and changing the per-sample weight to a per-process one fails `test_reduce_sample_coefficients_weights_by_sample_count` plus an existing value test.

The two items I flagged above (float32 accumulator, and the `if not names` early return resting on `record_batch` being called uniformly across processes) are unchanged — flagging that they stand as-is rather than assuming the approval covered them. Happy to change either if you'd rather.

Arcomano1234:
PR title and description updated by Claude Code.

**Model**: claude-opus-4-6 (1M context)
**Context gaps**: None — the gather script retrieved the diff, existing prose, and discussion without warnings.
**What changed**: Updated attribute names in the description to match the final code: `_target_shapes` / `_gen_shapes` → `_target_spatial_shapes` / `_gen_spatial_shapes`, and noted the `spatial_shapes` parameter name on `reduce_sample_coefficients`.

PR #1491, opened 2026-09-11
Allow parameter_init to load a checkpoint into a submodule of the destination

elynnwu:
Initializing a noise-conditioned module from a deterministic checkpoint of the same network was not possible: `SamudraBuilder` returns a bare `Samudra` at `noise_embed_dim: 0` and `NoiseConditionedModel(samudra, ...)` above zero, so enabling conditioning both adds the zero-initialized FiLM tensors and re-parents every existing parameter under `conditional_model`. `overwrite_weights` requires the source names to be a subset of the destination's, so the added parameters were fine but the renamed ones made every source name report as missing. That blocks "pretrain deterministic, then finetune stochastic", which is how the coupled Samudra work initializes its stochastic ocean.

`ParameterInitializationConfig.weights_submodule` names the submodule of the module being initialized that the checkpoint's module corresponds to; loaded parameter names are prefixed with it before matching. Because the conditioning layers are zero-initialized, the wrapped model at initialization then reproduces the checkpoint's predictions exactly, which is what makes the second stage a continuation rather than a restart.

The name is inserted after any leading `module.`, since that names the `DistributedDataParallel`/`DummyWrapper` layer rather than a submodule of the network. The renaming is applied on both paths that consume the loaded weights — `apply_weights` and the `alpha`/`beta` L2-SP regularizer, which pairs each parameter with the base value it was initialized from — so `parameters` exclude and frozen patterns are written against destination names throughout.

The precondition for `weights_submodule` is that weights were loaded, not that `weights_path` was set: the coupled stepper loads component weights from `CoupledParameterInitConfig.checkpoint_path`, which requires the component's `weights_path` to be `None`. Only the initializer knows whether anything was loaded, so `apply_weights` raises there rather than `__post_init__` rejecting a valid coupled configuration.

The failure this fixes was previously reported as a list of every missing parameter name with no indication of the cause, so `overwrite_weights` now reports counts and, when prefixing by a submodule would make the names match, names the candidates and the config field to set. It lists every candidate rather than singling out one: the subset test can be satisfied by more than one submodule, and satisfying it is not proof that the architectures correspond, so the message suggests and does not decide. This is also why `weights_submodule` is explicit rather than auto-detected.

Behavior change on an existing path: in `get_l2_sp_tuning_regularizer` the `.to(device)` applied to the base weights was rebound to a local that was then discarded, so the regularizer closure differenced against the originally-loaded tensors. The moved tensors are now the ones the closure uses.

Changes:
- `fme.ace.stepper.parameter_init.ParameterInitializationConfig`: new optional `weights_submodule` field, validated to reject empty dotted segments; applied via `apply_submodule_prefix` in `ParameterInitializer.apply_weights` and `get_l2_sp_tuning_regularizer`
- `fme.core.weight_ops.prefix_submodule`: renames a state dict onto a submodule of its destination, inserting after the wrapper prefix that `strip_leading_module` describes
- `fme.core.weight_ops.require_subset`: the source-is-subset-of-destination check, shared by `overwrite_weights` and the regularizer so both report the same diagnosis; its failure message reports counts and candidate submodules

- [x] Tests added

### DIFF
--- a/fme/ace/stepper/parameter_init.py
+++ b/fme/ace/stepper/parameter_init.py
@@ -9,7 +9,7 @@ from torch import nn
 
 from fme.core.device import get_device
 from fme.core.training_history import TrainingHistory
-from fme.core.weight_ops import overwrite_weights
+from fme.core.weight_ops import overwrite_weights, prefix_submodule
 from fme.core.wildcard import apply_by_exclude, apply_by_include, wildcard_match
 
 Weights = list[Mapping[str, Any]]
@@ -106,6 +106,19 @@ class ParameterInitializationConfig:
     Parameters:
         weights_path: path to a Stepper checkpoint
             containing weights to load
+        weights_submodule: dotted name of the submodule of the module being
+            initialized that the checkpoint's module corresponds to. Use it
+            when the destination wraps the checkpoint's architecture, so the
+            checkpoint's parameters live one level down: initializing a
+            noise-conditioned module from a deterministic checkpoint of the
+            same network needs ``conditional_model``, since
+            ``NoiseConditionedModel`` holds the wrapped network under that
+            name. Every loaded parameter name is prefixed with it before being
+            matched against the destination, so ``parameters`` exclude and
+            frozen patterns are written against destination names either way.
+            Parameters the destination has and the checkpoint does not (the
+            zero-initialized conditioning layers, for instance) keep their
+            initialized values.
         parameters: list of ParameterClassification objects, each specifying
             whether parameters are excluded from initialization or frozen.
             By default modules are unfrozen and all parameters are included.
@@ -120,6 +133,7 @@ class ParameterInitializationConfig:
     """
 
     weights_path: str | None = None
+    weights_submodule: str | None = None
     parameters: list[ParameterClassification] = dataclasses.field(default_factory=list)
     alpha: float = 0.0
     beta: float = 0.0
@@ -127,6 +141,17 @@ class ParameterInitializationConfig:
     frozen_parameters: FrozenParameterConfig | None = None
 
     def __post_init__(self):
+        if self.weights_submodule is not None:
+            if self.weights_path is None:
+                raise ValueError(
+                    "weights_submodule has no effect without weights_path; "
+                    "it renames the parameters loaded from that checkpoint."
+                )
+            if not self.weights_submodule or self.weights_submodule.endswith("."):
+                raise ValueError(
+                    "weights_submodule must be a dotted submodule name without a "
+                    f"trailing '.', got {self.weights_submodule!r}"
+                )
         if self.exclude_parameters is not None or self.frozen_parameters is not None:
             if len(self.parameters) > 0:
                 raise ValueError(
@@ -154,6 +179,15 @@ class ParameterInitializationConfig:
         return ParameterInitializer(
             config=self, load_weights_and_history=load_weights_and_history
         )
+
+
+def _maybe_prefixed(
+    state_dict: Mapping[str, Any], submodule: str | None
+) -> Mapping[str, Any]:
+    """``prefix_submodule`` when a submodule is configured, else a passthrough."""
+    if submodule is None:
+        return state_dict
+    return prefix_submodule(state_dict, submodule)
 
 
 def null_weights_and_history(*_) -> StepperWeightsAndHistory:
@@ -211,7 +245,7 @@ class ParameterInitializer:
                 modules, self.base_weights, filled_parameters
             ):
                 overwrite_weights(
-                    state_dict,
+                    _maybe_prefixed(state_dict, self.config.weights_submodule),
                     module,
                     exclude_parameters=classification.exclude,
                 )
--- a/fme/ace/stepper/parameter_init.py
+++ b/fme/ace/stepper/parameter_init.py
@@ -9,7 +9,7 @@ from torch import nn
 
 from fme.core.device import get_device
 from fme.core.training_history import TrainingHistory
-from fme.core.weight_ops import overwrite_weights, prefix_submodule
+from fme.core.weight_ops import overwrite_weights, prefix_submodule, require_subset
 from fme.core.wildcard import apply_by_exclude, apply_by_include, wildcard_match
 
 Weights = list[Mapping[str, Any]]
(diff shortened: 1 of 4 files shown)


elynnwu:
Claude: the title and description above were **not** authored through the `write-pr-prose` skill, and this is disclosed here per that skill's own instruction that a fallback stay visible.

The skill pins `claude-opus-4-6[1m]` with `context: fork`. This session runs Claude Opus 5 from the research repo, which cannot invoke another repository's skill or select that pinned model, so I wrote the prose directly following AGENTS.md's "PR description template" section and `.github/pull_request_template.md`. If the pin matters here, the prose should be re-authored from an ace session via `.claude/skills/write-pr-prose/`.

Process notes, kept out of the description:
- Pre-existing failures on `main`, unrelated to this change and unaffected by it: 7 tests in `fme/core/dataset/test_xarray.py` (zarr). Verified identical on unmodified `main`.
- `fme/ace/stepper` and `fme/core` otherwise pass (1721 passed), as do ruff, ruff-format and mypy on the changed files.
- Origin: this is the missing capability behind the failure of coupled run [01M26T4AN7KVHN718YTDBVCZ4Z](https://beaker.org/ex/01M26T4AN7KVHN718YTDBVCZ4Z).

elynnwu:
Pre-review agent: independent pre-review, run without the authoring session's context. Two bugs found and fixed in e2a6b9d, plus smaller cleanups; the PR description was updated to describe the final state. Details below, with the two items I am leaving to the reviewer at the end.

## Fixed: `weights_submodule` was ignored by the L2-SP regularizer

`get_l2_sp_tuning_regularizer` builds its own copy of the subset check against the *unrenamed* base weights, so any non-zero `alpha` or `beta` raised

```
ValueError: Dest module is missing parameters {'module.layers.2.convblock.3.bias', ...}
```

at stepper construction — the exact failure `weights_submodule` exists to fix, and reported with the old unimproved message immediately after setting the field that was supposed to fix it. Reproduced before the fix with a Samudra `noise_embed_dim=6` destination, `weights_submodule="conditional_model"`, `alpha=0.1`.

This combination is not exotic: L2-SP is the natural regularizer for "pretrain deterministic, finetune stochastic", which is the motivating use case. Had it not been caught, the feature would have worked in the no-regularization case and failed the moment `alpha` was turned on.

It also made the docstring's claim about exclude patterns false on that path — the closure matched `classification.exclude` against unrenamed source names while `apply_weights` matched renamed destination names, so one pattern list could not be correct for both. The regularizer now applies the same renaming, which makes the documented rule ("exclude patterns are written against destination names") true throughout. Regression test: `test_weights_submodule_applies_to_the_l2_sp_regularizer`, verified to fail when the renaming is reverted.

## Fixed: the `weights_path` precondition rejected the coupled configuration

`__post_init__` required `weights_path` whenever `weights_submodule` was set. But `CoupledTrainStepperConfig.__post_init__` requires the component `weights_path` to be `None` when `CoupledParameterInitConfig.checkpoint_path` is used, while the component initializer is still built with a loader that returns that checkpoint's weights. So "finetune a coupled checkpoint, turning the ocean stochastic" was rejected with a message asserting the field would have no effect, when it would.

The real precondition is that weights were loaded — which the config cannot see and the initializer can — so the check moved to `apply_weights`. `test_weights_submodule_without_loaded_weights_raises_at_apply` covers it. Worth a coupled-level integration test too; I did not add one, since the component config is shared and the unit-level check pins the contract.

## Fixed: latent device bug on the existing alpha/beta path

Pre-existing, not introduced here, but adjacent enough to fix while in the code:

```python
for module, state_dict in zip(modules, base_weights):
    state_dict = {name: value.to(device) for ...}  # rebinds a local, then discarded
...
def regularizer():
    for module, state_dict, ... in zip(modules, base_weights, ...):  # the ORIGINAL list
        ... param - state_dict[name]
```

The moved tensors were thrown away and the closure differenced against whatever `load_weights_and_history` returned. Benign today because `load_stepper` already places weights on the device, but it silently depends on that. The moved dict is now the one the closure uses. Called out in the PR description as a behavior change on an existing path.

## Smaller changes

- `require_subset` in `weight_ops` is now shared by `overwrite_weights` and the regularizer. The duplicate check had already diverged — one arm got the improved message, the other kept the old one — which is what let the first bug hide.
- The error message listed every candidate submodule but then recommended `fixes[0]` specifically, presenting an alphabetical tie-break as the answer. It now lists all candidates and says explicitly that a matching prefix does not establish that the architectures correspond. `test_missing_parameters_message_lists_every_candidate_submodule` pins the two-candidate case.
- `dict.fromkeys(from_names)` was building throwaway dicts to reuse `strip_leading_module` on what are really name sets; replaced by a name-level `_strip_wrapper_prefix`, used by both.
- The missing-name list is capped at 10 with an "and N more" suffix. The full dump is what made the original failure unreadable.
- Validation now rejects any empty dotted segment, so leading dots (`.conditional_model`) and doubled dots (`a..b`) are caught, not just trailing ones.
- `_maybe_prefixed` became `ParameterInitializationConfig.apply_submodule_prefix`, keeping the field private to its own config per the builder-pattern convention.
- `prefix_submodule` had no coverage for a dotted submodule name, which the validation permits; added.
- The forward-equivalence test now runs both modules in `eval()` and additionally asserts `loaded(x) == loaded(x)`. The wrapper draws fresh noise every call, so that second assertion is what actually demonstrates the conditioning is inert at initialization rather than the FiLM path happening to be unexercised.

## Audit passes

✅ Silent-failure — the new field's only quiet path (set but nothing loaded) now raises. ✅ Behavior-change — one found (the device rebind), intended, documented. ✅ Differential tests — the key forward-equivalence test fails without the renaming (the subset check raises); the new L2-SP test verified failing on a reverted fix. ✅ Sibling/mirror — `fme/coupled` shares this config rather than mirroring it, so it is covered; that sharing is what surfaced the second bug. ✅ Config back-compat — additive optional field, prior checkpoints and YAML unaffected. ✅ Split — 4 files, one feature, no split warranted. ✅ Abstract-or-final and inheritance depth — no new classes outside tests. ✅ AI-agent artifacts — none remaining after the `dict.fromkeys` and free-function cleanups. ✅ Builder pattern — no new violations; the new field is read only through methods on its own config.

Tests: `fme/ace/stepper`, `fme/core`, `fme/coupled` — 1995 passed, 7 failed, all 7 the pre-existing `test_xarray.py` zarr failures the author verified identical on `main`. ruff, ruff-format and mypy clean on the changed files.

## Two things I did not change, for the reviewer

1. **The field name reads in the wrong direction.** `weights_submodule` parses as "the submodule *of the weights* to load", when the meaning is the opposite: the submodule *of the destination* the weights are loaded into. `destination_submodule` or `load_into_submodule` would be unambiguous. No configs use the field yet, so renaming is still free — but it is a naming call, so I left it.

2. **`prefix_submodule` branches on the source name, not the destination.** It inserts after a leading `module.` if the *source* has one, which assumes source and destination are wrapped the same way. True for everything reaching it through `apply_weights` (both come from a stepper's `.modules`), and documented, but it is an implicit coupling in a function that otherwise looks general.

Also noting, not a finding for this PR: `overwrite_weights` swallows non-parameter state via `except AttributeError: pass`, so buffers are never transferred. Irrelevant with Samudra's default `norm: instance` (no buffers), but `norm: batch` would carry running statistics that silently stay at their initial values — the forward-equivalence property this PR relies on would then not hold in `eval()` for a trained checkpoint. Pre-existing and out of scope.

PR #1492, opened 2026-09-11
Add ensemble step for weighted multi-checkpoint inference

mcgibbon:
Allow inference to mix multiple trained checkpoints, e.g. residual and non-residual models, by producing the weighted sum of their per-step outputs.

Changes:
- `fme.core.step.EnsembleStepConfig` / `EnsembleStep`: a new step type whose output is the weighted sum of its member steps, in member 0's variable order. Weights are used as given (not normalized). Members must agree on input, output, next-step-input, and next-step-forcing names, `n_ic_timesteps`, and ocean config. Quantities a step exposes once (normalizer, SST prescription, carried stepper state) come from member 0; every member is stepped on member 0's carried state and the other members' state updates are discarded. Corrector diagnostics are dropped; `allow_missing_variables` is `all()` over members. Stepper overrides (`StepperOverrideConfig`) apply to every member.
- `fme.ace.stepper.load_stepper_ensemble` / `load_stepper_ensemble_config`: parse each checkpoint independently, require mutually compatible dataset infos, build the ensemble step and stepper. The non-step stepper config (input masking, derived forcings), dataset info, and training history come from member 0, with a logged warning naming any field on which another member differs.
- `fme.ace.stepper.single_module._ParsedStepperState`: refactored out of `Stepper.from_state` so that both legacy and modern checkpoints parse once into a config + dataset-info + step-state triple, with an `ensemble` classmethod that composes members into a single parsed state. `Stepper.from_state` now delegates to it.
- `InferenceEvaluatorConfig.checkpoint_path` widened to `str | list[WeightedCheckpointPath]`, where the new `fme.ace.WeightedCheckpointPath` config holds a checkpoint `path` and its `weight`. The three `isinstance(checkpoint_path, str)` checks in `__post_init__`, `load_stepper`, and `load_stepper_config` are the resolution points for this dacite union field.

- [x] Tests added

### DIFF
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -167,6 +167,21 @@ class ValidationConfig:
 
 
 @dataclasses.dataclass
+class WeightedCheckpointPath:
+    """
+    A stepper checkpoint and its weight in an ensemble of checkpoints.
+
+    Parameters:
+        path: Path to the stepper checkpoint.
+        weight: Weight applied to this checkpoint's step output in the
+            weighted sum, used as given (not normalized).
+    """
+
+    path: str
+    weight: float
+
+
+@dataclasses.dataclass
 class InferenceEvaluatorConfig:
     """
     Configuration for running inference including comparison to reference data.
@@ -196,14 +211,11 @@ class InferenceEvaluatorConfig:
                 written to a local ``experiment_dir``.
 
         n_forward_steps: Number of steps to run the model forward for.
-        checkpoint_path: Path to stepper checkpoint to load, or a list of paths.
-            Given a list, the loaded stepper predicts each step as the weighted
-            sum of the checkpoints' steps (see
-            :class:`fme.core.step.EnsembleStepConfig`), so e.g. residual and
-            non-residual models can be mixed.
-        checkpoint_weights: One weight per checkpoint in ``checkpoint_path``,
-            used as given (not normalized). Defaults to equal weights summing
-            to one. Only valid when ``checkpoint_path`` is a list.
+        checkpoint_path: Path to stepper checkpoint to load, or a list of
+            weighted checkpoint paths. Given a list, the loaded stepper
+            predicts each step as the weighted sum of the checkpoints' steps
+            (see :class:`fme.core.step.EnsembleStepConfig`), so e.g. residual
+            and non-residual models can be mixed.
         logging: configuration for logging.
         loader: Configuration for data to be used as initial conditions, forcing, and
             target in inference.
@@ -237,11 +249,10 @@ class InferenceEvaluatorConfig:
 
     experiment_dir: str
     n_forward_steps: int
-    checkpoint_path: str | list[str]
+    checkpoint_path: str | list[WeightedCheckpointPath]
     logging: LoggingConfig
     loader: InferenceDataLoaderConfig
     forward_steps_in_memory: int
-    checkpoint_weights: list[float] | None = None
     prediction_loader: InferenceDataLoaderConfig | None = None
     data_writer: DataWriterConfig = dataclasses.field(
         default_factory=lambda: DataWriterConfig()
@@ -263,22 +274,8 @@ class InferenceEvaluatorConfig:
         )
         # checkpoint_path is a dacite union, resolved by isinstance here and in
         # load_stepper / load_stepper_config below.
-        if isinstance(self.checkpoint_path, str):
-            if self.checkpoint_weights is not None:
-                raise ValueError(
-                    "checkpoint_weights requires checkpoint_path to be a list"
-                )
-        else:
-            n_checkpoints = len(self.checkpoint_path)
-            if n_checkpoints == 0:
-                raise ValueError("checkpoint_path list must not be empty")
-            if self.checkpoint_weights is None:
-                self.checkpoint_weights = [1.0 / n_checkpoints] * n_checkpoints
-            elif len(self.checkpoint_weights) != n_checkpoints:
-                raise ValueError(
-                    f"got {n_checkpoints} checkpoint paths but "
-                    f"{len(self.checkpoint_weights)} checkpoint_weights"
-                )
+        if not isinstance(self.checkpoint_path, str) and not self.checkpoint_path:
+            raise ValueError("checkpoint_path list must not be empty")
 
     def configure_logging(self, log_filename: str):
         config = dataclasses.asdict(self)
@@ -290,9 +287,10 @@ class InferenceEvaluatorConfig:
         logging.info(f"Loading trained model checkpoint from {self.checkpoint_path}")
         if isinstance(self.checkpoint_path, str):
             return load_stepper(self.checkpoint_path, self.stepper_override)
-        assert self.checkpoint_weights is not None  # set in __post_init__
         return load_stepper_ensemble(
-            self.checkpoint_path, self.checkpoint_weights, self.stepper_override
+            [checkpoint.path for checkpoint in self.checkpoint_path],
+            [checkpoint.weight for checkpoint in self.checkpoint_path],
+            self.stepper_override,
         )
 
     def load_stepper_config(self) -> StepperConfig:
@@ -301,9 +299,10 @@ class InferenceEvaluatorConfig:
             return load_stepper_config_with_override(
                 self.checkpoint_path, self.stepper_override
             )
-        assert self.checkpoint_weights is not None  # set in __post_init__
         return load_stepper_ensemble_config(
-            self.checkpoint_path, self.checkpoint_weights, self.stepper_override
+            [checkpoint.path for checkpoint in self.checkpoint_path],
+            [checkpoint.weight for checkpoint in self.checkpoint_path],
+            self.stepper_override,
         )
 
     def get_data_writer(
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -32,6 +32,8 @@ from fme.ace.stepper import (
     StepperOverrideConfig,
     load_stepper,
     load_stepper_config_with_override,
+    load_stepper_ensemble,
+    load_stepper_ensemble_config,
 )
 from fme.ace.stepper.single_module import (
     StepperConfig,
@@ -194,7 +196,14 @@ class InferenceEvaluatorConfig:
                 written to a local ``experiment_dir``.
 
         n_forward_steps: Number of steps to run the model forward for.
-        checkpoint_path: Path to stepper checkpoint to load.
+        checkpoint_path: Path to stepper checkpoint to load, or a list of paths.
+            Given a list, the loaded stepper predicts each step as the weighted
+            sum of the checkpoints' steps (see
+            :class:`fme.core.step.EnsembleStepConfig`), so e.g. residual and
+            non-residual models can be mixed.
+        checkpoint_weights: One weight per checkpoint in ``checkpoint_path``,
+            used as given (not normalized). Defaults to equal weights summing
+            to one. Only valid when ``checkpoint_path`` is a list.
         logging: configuration for logging.
         loader: Configuration for data to be used as initial conditions, forcing, and
             target in inference.
@@ -228,10 +237,11 @@ class InferenceEvaluatorConfig:
 
     experiment_dir: str
     n_forward_steps: int
-    checkpoint_path: str
+    checkpoint_path: str | list[str]
     logging: LoggingConfig
     loader: InferenceDataLoaderConfig
     forward_steps_in_memory: int
+    checkpoint_weights: list[float] | None = None
     prediction_loader: InferenceDataLoaderConfig | None = None
     data_writer: DataWriterConfig = dataclasses.field(
         default_factory=lambda: DataWriterConfig()
@@ -251,6 +261,23 @@ class InferenceEvaluatorConfig:
             self.forward_steps_in_memory,
             self.n_forward_steps,
         )
+        # checkpoint_path is a dacite union; this is the one place it is resolved.
+        if isinstance(self.checkpoint_path, str):
+            if self.checkpoint_weights is not None:
+                raise ValueError(
+                    "checkpoint_weights requires checkpoint_path to be a list"
+                )
+        else:
+            n_checkpoints = len(self.checkpoint_path)
+            if n_checkpoints == 0:
+                raise ValueError("checkpoint_path list must not be empty")
+            if self.checkpoint_weights is None:
+                self.checkpoint_weights = [1.0 / n_checkpoints] * n_checkpoints
+            elif len(self.checkpoint_weights) != n_checkpoints:
+                raise ValueError(
+                    f"got {n_checkpoints} checkpoint paths but "
+                    f"{len(self.checkpoint_weights)} checkpoint_weights"
+                )
 
     def configure_logging(self, log_filename: str):
         config = dataclasses.asdict(self)
@@ -260,12 +287,22 @@ class InferenceEvaluatorConfig:
 
     def load_stepper(self) -> Stepper:
         logging.info(f"Loading trained model checkpoint from {self.checkpoint_path}")
-        return load_stepper(self.checkpoint_path, self.stepper_override)
+        if isinstance(self.checkpoint_path, str):
+            return load_stepper(self.checkpoint_path, self.stepper_override)
+        assert self.checkpoint_weights is not None  # set in __post_init__
+        return load_stepper_ensemble(
+            self.checkpoint_path, self.checkpoint_weights, self.stepper_override
+        )
 
     def load_stepper_config(self) -> StepperConfig:
         logging.info(f"Loading trained model checkpoint from {self.checkpoint_path}")
-        return load_stepper_config_with_override(
-            self.checkpoint_path, self.stepper_override
+        if isinstance(self.checkpoint_path, str):
+            return load_stepper_config_with_override(
+                self.checkpoint_path, self.stepper_override
+            )
+        assert self.checkpoint_weights is not None  # set in __post_init__
+        return load_stepper_ensemble_config(
+            self.checkpoint_path, self.checkpoint_weights, self.stepper_override
         )
 
     def get_data_writer(
--- a/fme/ace/inference/evaluator.py
+++ b/fme/ace/inference/evaluator.py
@@ -261,7 +261,8 @@ class InferenceEvaluatorConfig:
             self.forward_steps_in_memory,
             self.n_forward_steps,
         )
-        # checkpoint_path is a dacite union; this is the one place it is resolved.
+        # checkpoint_path is a dacite union, resolved by isinstance here and in
+        # load_stepper / load_stepper_config below.
         if isinstance(self.checkpoint_path, str):
             if self.checkpoint_weights is not None:
                 raise ValueError(
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -1946,7 +1946,8 @@ class _ParsedStepperState:
 
         The members' dataset infos must be mutually compatible; the ensemble
         takes the first member's dataset info, training history, and stepper
-        configuration other than the step (input masking, derived forcings).
+        configuration other than the step (input masking, derived forcings),
+        logging a warning for any member whose values differ.
         The ensemble step config validates the members' step compatibility.
         """
         first = members[0]
@@ -1958,6 +1959,16 @@ class _ParsedStepperState:
                     f"ensemble member {i} was trained on a dataset incompatible "
                     f"with member 0's: {err}"
                 ) from err
+            for field in dataclasses.fields(first.config):
+                if field.name == "step":
+                    continue
+                if getattr(first.config, field.name) != getattr(
+                    member.config, field.name
+                ):
+                    logging.warning(
+                        f"ensemble member {i} has a different {field.name} than "
+                        f"member 0; using member 0's for the whole ensemble"
+                    )
         step = StepSelector(
             type="ensemble",
             config={
--- a/fme/ace/stepper/single_module.py
+++ b/fme/ace/stepper/single_module.py
@@ -3,7 +3,7 @@ import dataclasses
 import datetime
 import logging
 import pathlib
-from collections.abc import Callable, Generator, Mapping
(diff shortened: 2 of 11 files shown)


mcgibbon:
PR title and description written by claude-opus-4-6[1m] (Claude Opus 4.6, 1M context) via the write-pr-prose skill. The code itself was authored by Claude Fable 5.1 (claude-fable-5-1) at Jeremy's request.

The gather_pr_context.sh helper failed (ran against the wrong repo), so context was retrieved directly via `gh pr view` / `gh pr diff` against ai2cm/ace. No review comments or discussion existed at the time of writing.

fme/core/step/ensemble.py
```diff
@@ -0,0 +1,244 @@
...
+    checks: list[tuple[str, Any, Any]] = [
+        ("input_names", first.input_names, other.input_names),
+        ("output_names", first.output_names, other.output_names),
+        (
+            "next_step_input_names",
+            first.next_step_input_names,
+            other.next_step_input_names,
+        ),
+        ("n_ic_timesteps", first.n_ic_timesteps, other.n_ic_timesteps),
+        ("ocean", first.get_ocean(), other.get_ocean()),
```
mcgibbon, line 144: Claude: **the compatibility check is missing `get_next_step_forcing_names()`, and the gap fails silently in the wrong direction.**

`SingleModuleStepConfig.next_step_input_names` is `input_names - output_names | ocean.forcing_names | prescribed_prognostic_names` (`fme/core/step/single_module.py:197-202`) and does *not* include `next_step_forcing_names`. Two members with identical in/out names can therefore disagree about which forcings come from the output timestep and still pass every check in this list.

The consequence is not a config that looks wrong, it is wrong physics with no error. `Stepper.predict_generator` builds one `input_data` for the whole ensemble from member 0's list:

```python
input_forcing = {
    k: (forcing_dict[k][:, step]
        if k not in self._step_obj.next_step_forcing_names
        else forcing_dict[k][:, step + 1])
    for k in self._input_only_names
}
```

(`fme/ace/stepper/single_module.py:1168-1173`; `next_step_forcing_names` resolves through `EnsembleStepConfig.get_next_step_forcing_names` to `self.members[0]`.) Every member is then stepped on that one input. A member trained with insolation at the output timestep, ensembled behind a member 0 that was not, runs half an interval off for the entire rollout and nothing reports it.

Mixing models from different training recipes is the whole point of this PR, so this is the disagreement most likely to actually occur. Adding `("next_step_forcing_names", first.get_next_step_forcing_names(), other.get_next_step_forcing_names())` to `checks` fixes it in one line.

mcgibbon (commented):
Claude: pre-review agent (Opus 5, no authoring context). Comment-only per the Fable-authored fix policy, so nothing has been pushed to the branch.

I checked the branch out in an isolated worktree and ran `fme/core/step/`, `fme/ace/stepper/`, `fme/ace/inference/test_evaluator.py` and `fme/coupled/test_stepper.py` (451 passed, 2 skipped) plus `pre-commit` on the changed files (ruff, ruff-format, mypy all pass).

I verified the `_ParsedStepperState` refactor is behavior-preserving. `Stepper.load_state` reads only `state["step"]`, so replacing `stepper.load_state(state)` with `stepper.load_state({"step": step_state})` is equivalent on both the legacy and the modern branch, and dropping the old in-place `state["step"] = ...` mutation has no consumer: the callers of `Stepper.from_state` are `load_stepper` and `fme/coupled/stepper.py:1424-1425`, and neither re-reads the dict afterwards. The `get_state`/`from_state` round trip through the new ensemble step is covered end-to-end by `test_load_stepper_ensemble_mixes_residual_and_non_residual`.

One finding I would fix before merge (the `next_step_forcing_names` gap) and six non-blocking; details inline.

fme/ace/stepper/single_module.py
```diff
@@ -1914,6 +1858,130 @@ class StepperOverrideConfig:
...
+        first = members[0]
+        for i, member in enumerate(members[1:], start=1):
+            try:
+                first.dataset_info.assert_compatible_with(member.dataset_info)
+            except IncompatibleDatasetInfo as err:
+                raise IncompatibleDatasetInfo(
+                    f"ensemble member {i} was trained on a dataset incompatible "
+                    f"with member 0's: {err}"
+                ) from err
+        step = StepSelector(
```
mcgibbon, line 1972: Claude: the same class of silent mismatch as the `next_step_forcing_names` comment on `ensemble.py`, one level up. `dataclasses.replace(first.config, step=step)` keeps member 0's `input_masking` and `derived_forcings` and discards the other members', with nothing checking that they agree. `derived_forcings` decides what forcing is computed and handed to *all* members, so a member trained under different derived forcings is fed inputs it never saw in training.

"Member 0 wins" is the agreed design, but it is only safe when the others match. Either assert equality of `input_masking`/`derived_forcings` across members here, or log a warning naming the fields that differ. Non-blocking if you would rather take it as a follow-up, but it should be a deliberate call rather than an omission.

fme/core/step/ensemble.py
```diff
@@ -0,0 +1,244 @@
...
+    Configuration for a step whose output is a weighted sum of member steps.
+
+    Every member is called on the same inputs, and each output variable is the
+    weighted sum of the members' predictions for it. The weights are used as
+    given (not normalized), so a weighted mean needs weights summing to one.
+    Members must agree on their input, output, and next-step input names,
+    number of initial-condition timesteps, and ocean configuration; anything
+    else that a step exposes once (normalizer, surface temperature name, SST
+    prescription, carried stepper state) is taken from the first member.
+    Corrector diagnostics from the members are dropped.
```
mcgibbon: Claude: this docstring says what the ensemble takes *from* member 0 but not what it feeds *into* the other members, and the input side is the surprising half.

`EnsembleStep.step` passes the same `args` (including `args.stepper_state`) to every member, and that state is member 0's from the previous step, since only `outputs[0].stepper_state` is carried forward. So members 1..N read member 0's `corrector_state` (the corrector references seeded from the IC, `fme/core/step/single_module.py:693-696`) and member 0's `random_state`, and their own state updates are dropped. For stochastic members that also means they are perfectly correlated rather than independently sampled.

This follows from the agreed "carried state comes from member 0" decision, but a reader of this docstring would not predict it. One sentence, something like "every member is stepped on member 0's carried state; the other members' state updates are discarded", saves the next person the trace.

fme/core/step/ensemble.py
```diff
@@ -0,0 +1,244 @@
...
+        for member in self._members:
+            member.set_epoch(epoch)
+
+    def step(
+        self,
+        args: StepArgs,
+        wrapper: Callable[[nn.Module], nn.Module] = lambda x: x,
+    ) -> StepOutput:
+        outputs = [member.step(args=args, wrapper=wrapper) for member in self._members]
+        names = set(outputs[0].output)
```
mcgibbon: Claude: `names = set(outputs[0].output)` makes the assembled `output` dict's key order depend on set iteration order, which varies with `PYTHONHASHSEED`. Downstream writers iterate the output mapping, so variable order in written files can shuffle run to run for no reason.

`names` is only used for the membership comparison and for this loop, so iterating `outputs[0].output` directly in the accumulation loop preserves member 0's ordering and costs nothing.

fme/ace/inference/evaluator.py
```diff
@@ -251,6 +261,23 @@ def __post_init__(self):
             self.forward_steps_in_memory,
             self.n_forward_steps,
         )
+        # checkpoint_path is a dacite union; this is the one place it is resolved.
```
mcgibbon: Claude: the justification comment is not accurate as written. "this is the one place it is resolved" is followed by two more `isinstance(self.checkpoint_path, str)` checks, in `load_stepper` (line 290) and `load_stepper_config` (line 299). The PR description repeats the slip, calling them "the two `isinstance` checks" when there are three.

ace's AGENTS.md asks for isinstance to be justified, so this comment is doing real work and should say something true. Either reword it ("`checkpoint_path` is a dacite union; these three sites are where it is resolved"), or resolve the union once in `__post_init__` into a normalized private field and have the two loader methods branch on that, which makes the comment true as written and removes two of the three checks.

fme/ace/stepper/single_module.py
```diff
@@ -1970,6 +2038,67 @@ def load_stepper(
...
+        A stepper with an ``ensemble`` step over the checkpoints' steps. See
+        :class:`fme.core.step.EnsembleStepConfig` for how the members are
+        combined and which member's values are used where only one applies.
+    """
+    stepper = _parse_stepper_ensemble(checkpoint_paths, weights).build_stepper()
+    apply_stepper_override(stepper, override_config)
+    return stepper
+
+
+def load_stepper_ensemble_config(
```
mcgibbon, line 2090: Claude: minor naming asymmetry. This is the analogue of `load_stepper_config_with_override`, not of `load_stepper_config`: it takes an override and has the same shape (no override, parse only; override, build the stepper and take `_config`). The name makes the evaluator's two branches read as if they call different kinds of thing:

```python
return load_stepper_config_with_override(self.checkpoint_path, self.stepper_override)
...
return load_stepper_ensemble_config(self.checkpoint_path, self.checkpoint_weights, self.stepper_override)
```

The behavior is right and consistent with the single-checkpoint pair; it is only the name that hides the parallel. `load_stepper_ensemble_config_with_override` is a mouthful, so take it or leave it.

fme/core/step/test_ensemble.py
```diff
@@ -0,0 +1,250 @@
...
+            weights=[1.0],
+        )
+
+
+def test_ensemble_config_rejects_no_members():
+    with pytest.raises(ValueError, match="at least one member"):
+        EnsembleStepConfig(members=[], weights=[])
+
+
+def test_ensemble_config_rejects_disagreeing_members():
```
mcgibbon, line 195: Claude: three coverage notes, all non-blocking.

1. This test exercises one of the five entries in `_assert_members_agree`'s `checks` list (`output_names`). `input_names`, `next_step_input_names`, `n_ic_timesteps` and `ocean` are unexercised, so a wrong operand or wrong label in any of them would go unnoticed. `checks` is already a uniform list of triples, so a parametrize over it is nearly free.

2. `wrapper` pass-through to the members, `set_epoch` forwarding, and `get_regularizer_loss` summation are implemented in `EnsembleStep` but asserted nowhere. `wrapper` is the one worth adding: it is the checkpointing/DDP hook, and silently dropping it would be invisible at inference and wrong under training.

3. The legacy branch of `_ParsedStepperState.from_state` has no test. The committed `fme/ace/inference/stepper_test_data` that `test_inference_backwards_compatibility` loads is in the *modern* format (top-level keys `config`/`dataset_info`/`step`), so it takes the `except dacite.exceptions.DaciteError` path. I satisfied myself that the legacy branch is behavior-preserving by reading it (see the review summary), but nothing in CI would catch a regression there.

mcgibbon:
Claude: pre-review verdict — **one blocking finding, then ready for review.**

Blocking:

1. [`_assert_members_agree` is missing `get_next_step_forcing_names()`](https://github.com/ai2cm/ace/pull/1492#discussion_r3992076417). Members that disagree on which forcings come from the output timestep pass every compatibility check, and the ensemble then steps all of them on member 0's choice — half an interval off, for the whole rollout, with no error. It is a one-line addition to the `checks` list plus a case in the parametrized rejection test.

Everything else I found is a comment, a docstring sentence, or a test to add, and none of it needs to gate the review. Summary of what I checked and did not find a problem with:

- **`_ParsedStepperState` refactor**: behavior-preserving on both branches. `Stepper.load_state` reads only `state["step"]`, and the dropped in-place `state["step"] = ...` mutation had no consumer.
- **`StepABC` contract**: `EnsembleStep` implements every abstract member; `train`/`eval`, `modules`-as-a-fresh-`ModuleList`, and `set_epoch` forwarding all follow the established `MultiCallStep`/`SeparateRadiationStep` patterns. `get_state`/`load_state` round-trip through the real `Stepper.get_state()` in `test_load_stepper_ensemble_mixes_residual_and_non_residual`.
- **Dropped corrector diagnostics are safe downstream**: consistently-empty deltas make `StepOutput.stack_diagnostics` return `None` rather than raising on inconsistent keys.
- **`replace_ocean` / `replace_prescribed_prognostic_names` / `replace_multi_call` on an ensemble** all work: the config-level replacements propagate to every member selector, and `Stepper.replace_*` rebuilds through `get_state()`/`load_state()`, which the ensemble's `{"members": [...]}` state round-trips.
- **`load_stepper_ensemble_config` without an override** matches `load_stepper_config_with_override`'s structure exactly.
- **Docs**: `docs/available_steps.rst` is generated in `docs/conf.py` from the registry, so registering `ensemble` is enough; the `steps.rst` autoclass entry is the only hand-edit needed and it is there.
- **No other consumer** of `InferenceEvaluatorConfig.checkpoint_path` exists that the widened type could break.
- **Tests and lint**: `fme/core/step/`, `fme/ace/stepper/`, `fme/ace/inference/test_evaluator.py`, `fme/coupled/test_stepper.py` — 451 passed, 2 skipped. `pre-commit` (ruff, ruff-format, mypy) passes on all changed files.

The PR description matches the change from main, with one correction: it says "the two `isinstance(checkpoint_path, str)` checks" where there are three (`__post_init__`, `load_stepper`, `load_stepper_config`).

fme/core/step/ensemble.py
mcgibbon: Claude: Agreed, and fixed in ea58fa416: `_assert_members_agree` now compares `frozenset(get_next_step_forcing_names())` as well, and the parametrized test covers it by name.

fme/ace/stepper/single_module.py
mcgibbon: Claude: Took the warning option in ea58fa416. `_ParsedStepperState.ensemble` now compares every non-step `StepperConfig` field across members and logs a warning naming each field on which a member differs from member 0. Not a hard error because Jeremy's instruction was to take member 0's value for anything duplicated; a warning makes the choice visible without rejecting mixes he may intend. Covered by `test_parsed_stepper_state_ensemble_warns_on_differing_stepper_config`.

fme/core/step/ensemble.py
mcgibbon: Claude: Added to the `EnsembleStepConfig` docstring in ea58fa416: every member is stepped on member 0's carried state, other members' state updates are discarded, and stochastic members therefore share member 0's random state rather than sampling independently.

fme/core/step/ensemble.py
mcgibbon: Claude: Fixed in ea58fa416: the accumulation loop iterates `outputs[0].output` directly, so the ensemble output keeps member 0's variable order.

fme/ace/inference/evaluator.py
mcgibbon: Claude: Reworded the comment in ea58fa416 to name all three sites, and corrected the PR description to say three. I kept the three checks rather than normalizing into a private field: a second field that duplicates `checkpoint_path` would have to be kept consistent with it in `configure_logging`'s `asdict` output and anywhere else the config is serialized.

fme/ace/stepper/single_module.py
mcgibbon: Claude: Leaving the name as is. The single-checkpoint pair exists because `load_stepper_config` predates overrides; the ensemble loader has only ever had the optional override, so there is no non-override sibling for the suffix to distinguish it from. Happy to rename if Jeremy prefers the parallel.

fme/core/step/test_ensemble.py
mcgibbon: Claude: (1) and (2) done in ea58fa416: `test_ensemble_config_names_the_disagreeing_property` is parametrized over input, output, next-step-input, next-step-forcing, and ocean disagreements using real single-module configs (`n_ic_timesteps` cannot differ between single-module steps, so it stays covered only by the mock-based test's plumbing), and `test_ensemble_step_forwards_wrapper_epoch_and_regularizer_loss` asserts the wrapper reaches every member, `set_epoch` is forwarded, and regularizer losses are summed. (3) I wrote a legacy-format test and it fails on main too: building the default atmosphere corrector requires `horizontal_coordinates`, which legacy checkpoints never stored, so `Stepper.from_state` raises `MissingDatasetInfo` on any legacy state (verified against 480aba1a3 with the same synthetic legacy dict). The legacy branch is unreachable in practice on main, so I did not add a test for it; that is a pre-existing condition outside this PR.

mcgibbon:
Claude: **Verdict at `ea58fa416`: ready for review.**

Re-reviewed the head that addresses my round-1 comments. All five source fixes are correct and complete, the three new tests are sound, and I found no new issues. One factual correction to an in-thread reply is below; it does not block.

### Fixes verified

1. `_assert_members_agree` now compares `frozenset(get_next_step_forcing_names())` — this was the one blocking finding, and the `frozenset` wrapper is right (the underlying accessor returns a list, so order differences between otherwise-identical members would have produced spurious failures).
2. `_ParsedStepperState.ensemble` iterates `dataclasses.fields(first.config)`, skips `step`, and warns naming each field on which a member differs from member 0. Warn-not-raise matches Jeremy's member-0-wins decision, and skipping `step` is correct since the step configs are expected to differ (that's the point of the ensemble).
3. `EnsembleStep` docstring now states that every member is stepped on member 0's carried state and that stochastic members share member 0's random state — the semantics a user needs before mixing stochastic members.
4. `EnsembleStep.step` iterates `outputs[0].output` directly (`# keep member 0's variable order`), so the output dict's insertion order is member 0's rather than a set's arbitrary order.
5. The `isinstance` justification comment names all three union-resolution sites, and the PR description says three. Adequate justification under AGENTS.md: the field is a dacite `str | list[str]` union and `isinstance` is the resolution point, not a type switch on behavior.

### New tests

- `test_ensemble_config_names_the_disagreeing_property` — I traced each parametrize case (input / output / next-step-input / next-step-forcing / ocean) and confirmed the named property is genuinely the *first* `checks` entry that differs for that case, so each case tests what its id claims.
- `test_ensemble_step_forwards_wrapper_epoch_and_regularizer_loss` — covers wrapper reaching every member, `set_epoch` forwarding, and regularizer-loss summation.
- `test_parsed_stepper_state_ensemble_warns_on_differing_stepper_config` — uses `caplog` with two real steppers differing on `derived_forcings`; asserts the field name appears in the warning.

Residual gaps, all non-blocking: `n_ic_timesteps` is the one `checks` entry still exercised only through the mock-based test; `set(outputs[0].output)` is recomputed inside the mismatch loop rather than hoisted (irrelevant, the loop exits on first mismatch); `get_single_module_selector` uses shared mutable list defaults (harmless — `SingleModuleStepConfig.__post_init__` does not mutate `in_names`/`out_names` in place, which I checked).

### Correction: the legacy branch *is* reachable

The reply on my legacy-checkpoint comment says `Stepper.from_state` on a legacy-format state raises `MissingDatasetInfo("horizontal_coordinates")` from the default atmosphere corrector build on `main` too, so the legacy branch is unreachable and out of scope. That is not right. I synthesized a legacy-format state (top-level `config`/`normalizer`/`loss_normalizer`/`sigma_coordinates`/`area`/`img_shape`/`module` with `"module."`-prefixed keys/`encoded_timestep`/`training_history`) and `Stepper.from_state` built a working `MultiCallStep` and stepped successfully on both `480aba1a3` and `ea58fa416`. (A legacy state whose `img_shape`/`area` are present is enough; the earlier probe likely hit the corrector because of a different synthetic state, not because the branch is dead.)

That said, the finding stays non-blocking, because I verified the refactor empirically rather than relying on a test: running the same legacy state through `main` and this head gives identical `step` type, identical output (`OUT_SUM -81.50000000`, sum-of-squares `231.54687408`), identical module weights, identical normalizer means/stds, identical serialized-config hash (`efdfef065f1916ef`), and identical `dataset_info` keys. The only observable difference is that `main` mutates the caller's state dict by inserting a `"step"` key and this head does not — and `Stepper.load_state` reads only `state["step"]`, so nothing consumes that mutation. So `_ParsedStepperState.from_state` + `build_stepper` is behavior-preserving on the legacy path, and a regression test for it would be a nice-to-have rather than a gate. Worth noting in the PR (or a follow-up) that the branch is live, since the reasoning recorded here would otherwise suggest it can be deleted.

### Checks

- `FME_FORCE_CPU=1 python -m pytest fme/core/step/ fme/ace/stepper/ fme/ace/inference/test_evaluator.py fme/coupled/test_stepper.py -q -p no:warnings` → 500 passed, 2 skipped.
- `pre-commit run --files <5 changed files>` → ruff, ruff-format, mypy all pass.
- PR description matches the final change from `main`: it covers next-step-forcing-name agreement, member 0's variable order, the carried-state/random-state semantics, the logged warning on differing stepper config, and the three `isinstance` sites.

fme/core/step/test_ensemble.py
mcgibbon: Claude: Correction to my point (3) above: the reviewer is right and my claim was wrong. My synthetic legacy state carried a serialized DatasetInfo whose gridded_operations and img_shape were None, so the legacy branch (which reads area + img_shape, never horizontal_coordinates) dropped the grid and the failure was in my fixture, not on main. Real legacy checkpoints store area + img_shape and load on both main and this branch. Added `test_stepper_from_state_loads_legacy_checkpoint_format` (8c8bab504) built that way: it asserts the multi_call step type, the network and loss normalizers, and the stepped output through the legacy path.

fme/ace/inference/evaluator.py
```diff
@@ -228,10 +237,11 @@ class InferenceEvaluatorConfig:
 
     experiment_dir: str
     n_forward_steps: int
-    checkpoint_path: str
+    checkpoint_path: str | list[str]
```
mcgibbon: Make this a list[WeightedCheckpointPath], where WeightedCheckpointPath is a new config exposing weight: and path: attributes.

mcgibbon: Claude: Done in 9aa97721f. `checkpoint_path` is now `str | list[WeightedCheckpointPath]`, with `WeightedCheckpointPath(path: str, weight: float)` defined next to `InferenceEvaluatorConfig`, exported from `fme.ace`, and listed under the evaluator config docs. The separate `checkpoint_weights` field is gone, so each entry names its own weight and there is no default weight; an empty list is still rejected. Tests updated accordingly.

PR #1493, opened 2026-09-11
Make gradient checkpointing differentiable in the conditional SFNO

elynnwu:
The four `torch.utils.checkpoint` call sites in the conditional SFNO omitted `use_reentrant=False`, so they used the reentrant implementation. That variant returns an output with no `grad_fn` when none of its *inputs* require grad, which is always the case here because the checkpointed input is data rather than a parameter. The effect was silent: with `checkpointing >= 1` the encoder's parameters (`conditional_model.encoder.{0.weight,0.bias,2.weight}`) were left at `grad=None`, so training proceeded with a frozen, randomly-initialised input encoder and raised nothing louder than a `UserWarning`. Any single-process run that set `checkpointing >= 1` is suspect; under DDP the default `find_unused_parameters=False` should have raised at the first backward instead.

Passing `use_reentrant=False` makes the checkpointed path differentiable. Checkpointing is exact recompute, so gradients are bit-identical to the uncheckpointed path at every level. `use_reentrant=False` is also the variant PyTorch documents as compatible with DDP, so this is a robustness improvement as well as a correctness fix.

This matters beyond correctness: on A100-80GB, `checkpointing: 3` takes the E3SMv3 historical atmosphere from 61.8 GB/GPU to 28.4 GB/GPU at the same batch size, which is what lets the coupled configuration fit with headroom.

`fme/ace/models/makani*` and `fme/ace/models/makani_fcn3/*` already pass `use_reentrant=False`; `fme/core/models/conditional_sfno/` did not. `fme/ace/models/modulus/{sfnonet,layers}.py` (4 sites — the SFNO registered as `SphericalFourierNeuralOperatorNet` in `fme/ace/registry/sfno.py`) and `fme/downscaling/modules/swinir.py` (1 site) carry the same defect; they are deliberately out of scope so that the fix here stays identical to the upstream one, and are tracked as separate follow-up work.

Changes:
- `fme.core.models.conditional_sfno.sfnonet.SphericalFourierNeuralOperatorNet`: pass `use_reentrant=False` at the three checkpointed call sites (blocks at `checkpointing >= 3`, encoder and decoder at `checkpointing >= 1`)
- `fme.core.models.conditional_sfno.layers.MLP.checkpoint_forward`: same, for the `checkpointing >= 2` path
- `fme.core.models.conditional_sfno.test_sfnonet`: add `test_checkpointing_gradients_match_uncheckpointed`, parametrised over `checkpointing` 1/2/3, asserting that every parameter's gradient matches the uncheckpointed reference. It fails without the fix — a dropped gradient is `None` against a grad-bearing reference, which `torch.testing.assert_close` rejects — so it covers both the missing gradients and any future divergence in the recomputed values.

The fix and its original test are authored by @mahf708 in E3SM-Project/ace@c5d39a0f, cherry-picked here unchanged; a follow-up commit simplifies the test setup and drops a second test that duplicated this one's coverage.

- [x] Tests added

### DIFF
--- a/fme/core/models/conditional_sfno/layers.py
+++ b/fme/core/models/conditional_sfno/layers.py
@@ -409,7 +409,7 @@ class MLP(nn.Module):
     @torch.jit.ignore
     def checkpoint_forward(self, x):  # pragma: no cover
         """Forward method with support for gradient checkpointing"""
-        return checkpoint(self.fwd, x)
+        return checkpoint(self.fwd, x, use_reentrant=False)
 
     def forward(self, x):  # pragma: no cover
         if self.checkpointing >= 2:
--- a/fme/core/models/conditional_sfno/sfnonet.py
+++ b/fme/core/models/conditional_sfno/sfnonet.py
@@ -761,7 +761,7 @@ class SphericalFourierNeuralOperatorNet(torch.nn.Module):
     def _forward_features(self, x: torch.Tensor, context: Context):
         for blk in self.blocks:
             if self.checkpointing >= 3:
-                x = checkpoint(blk, x, context)
+                x = checkpoint(blk, x, context, use_reentrant=False)
             else:
                 x = blk(x, context)
 
@@ -774,7 +774,7 @@ class SphericalFourierNeuralOperatorNet(torch.nn.Module):
             residual = self.norm_big_skip(residual, context=context)
 
         if self.checkpointing >= 1:
-            x = checkpoint(self.encoder, x)
+            x = checkpoint(self.encoder, x, use_reentrant=False)
         else:
             x = self.encoder(x)
 
@@ -815,7 +815,7 @@ class SphericalFourierNeuralOperatorNet(torch.nn.Module):
             x = torch.cat((x, residual), dim=1)
 
         if self.checkpointing >= 1:
-            x = checkpoint(self.decoder, x)
+            x = checkpoint(self.decoder, x, use_reentrant=False)
         else:
             x = self.decoder(x)
 
--- a/fme/core/models/conditional_sfno/test_sfnonet.py
+++ b/fme/core/models/conditional_sfno/test_sfnonet.py
@@ -754,3 +754,65 @@ def test_clip_latent_global_means_envelope_synchronized_across_ranks():
     # different envelopes and this reduction would change the values.
     torch.testing.assert_close(dist.reduce_min(model._gm_min.clone()), model._gm_min)
     torch.testing.assert_close(dist.reduce_max(model._gm_max.clone()), model._gm_max)
+
+
+def _make_checkpointing_model(checkpointing: int, img_shape=(9, 18)):
+    device = get_device()
+    params = SFNONetConfig(
+        embed_dim=16,
+        num_layers=2,
+        filter_type="linear",
+        checkpointing=checkpointing,
+    )
+    model = get_lat_lon_sfnonet(
+        params=params, img_shape=img_shape, in_chans=2, out_chans=2
+    ).to(device)
+    context = Context(
+        embedding_scalar=torch.zeros(4, 0, device=device),
+        labels=torch.zeros(4, 0, device=device),
+        noise=None,
+        embedding_pos=None,
+    )
+    return model, context, device
+
+
+def _grads_after_backward(checkpointing: int):
+    torch.manual_seed(0)
+    model, context, device = _make_checkpointing_model(checkpointing)
+    model.train()
+    torch.manual_seed(1234)
+    x = torch.randn(2, 2, 9, 18, device=device)
+    model(x, context).square().mean().backward()
+    return {
+        name: param.grad.detach().clone() if param.grad is not None else None
+        for name, param in model.named_parameters()
+    }
+
+
+@pytest.mark.parametrize("checkpointing", [1, 2, 3])
+def test_checkpointing_does_not_drop_gradients(checkpointing):
+    """Every parameter must receive a gradient at every checkpointing level.
+
+    torch.utils.checkpoint defaults to the reentrant implementation, which
+    produces an output with no grad_fn when no *input* tensor requires grad --
+    which is the case here, since the network input is data. That silently
+    leaves the encoder's parameters at grad=None, i.e. training a frozen,
+    randomly-initialized encoder with no error raised. Passing
+    use_reentrant=False is what makes the checkpointed path differentiable.
+    """
+    missing = [
+        name
+        for name, grad in _grads_after_backward(checkpointing).items()
+        if grad is None
+    ]
+    assert missing == [], f"no gradient for {missing}"
+
+
+@pytest.mark.parametrize("checkpointing", [1, 2, 3])
+def test_checkpointing_gradients_match_uncheckpointed(checkpointing):
+    """Checkpointing recomputes activations, so it must not change gradients."""
+    reference = _grads_after_backward(0)
+    actual = _grads_after_backward(checkpointing)
+    assert set(actual) == set(reference)
+    for name, grad in reference.items():
+        torch.testing.assert_close(actual[name], grad, msg=f"gradient for {name}")
--- a/fme/core/models/conditional_sfno/test_sfnonet.py
+++ b/fme/core/models/conditional_sfno/test_sfnonet.py
@@ -756,8 +756,12 @@ def test_clip_latent_global_means_envelope_synchronized_across_ranks():
     torch.testing.assert_close(dist.reduce_max(model._gm_max.clone()), model._gm_max)
 
 
-def _make_checkpointing_model(checkpointing: int, img_shape=(9, 18)):
+def _grads_after_backward(checkpointing: int):
+    """Gradient of every parameter after one backward at a checkpointing level."""
+    img_shape = (9, 18)
+    n_samples, n_channels = 2, 2
     device = get_device()
+    torch.manual_seed(0)
     params = SFNONetConfig(
         embed_dim=16,
         num_layers=2,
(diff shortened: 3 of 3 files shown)


elynnwu:
Claude: Pre-review agent — **verdict: ready for review**, with two description edits I was not permitted to apply (proposed text at the end) and two notes for the reviewer. No code changes required.

Reviewed cold in an isolated worktree of `origin/fix/conditional-sfno-gradient-checkpointing` (04b5c05, up to date with `main` @ 480aba1a3), tests run on CPU with the `fme` env.

## Verification

- **Cherry-pick fidelity.** Compared the PR's added/removed lines against the upstream patch for E3SM-Project/ace@c5d39a0 via the GitHub API: byte-for-byte identical, same three files and same +1/-1, +3/-3, +62/-0 line counts. Author metadata is `mahf708 <naser.mahfouz@pnnl.gov>`, authored 2026-08-24, committed 2026-09-11 — authorship preserved.
- **Applied with correct semantics against our `main`,** not just textually: `SFNONetConfig` has all four fields the test sets (`embed_dim`, `num_layers`, `filter_type`, `checkpointing`); `get_lat_lon_sfnonet(params, in_chans, out_chans, img_shape, ...)` matches the keyword call; `Context` is the 4-field dataclass the test constructs. All 33 tests in `test_sfnonet.py` and 119 passed / 3 skipped across `fme/core/models/conditional_sfno/` — exactly the numbers the description claims.
- **Differential test pass.** Reverting both source files to `origin/main` and keeping the new tests: all 6 parametrisations fail (`no gradient for [...encoder...]`), with the `UserWarning: None of the inputs have requires_grad=True` fired. Restoring them: 6 pass. Three repeat runs were identical — both helpers re-seed (`manual_seed(0)` before construction, `manual_seed(1234)` before the input) and `drop_rate`/`drop_path_rate` default to 0, so there is no seed fragility and no RNG dependence in the recompute.
- **Lint/types.** `pre-commit run --files <the three files>`: ruff, ruff-format, mypy all pass.

## Notes for the reviewer

**1. The `layers.py` hunk is not independently exercised by the tests.** Reverting *only* `layers.py` (keeping the `sfnonet.py` fix) leaves all 6 tests passing. That is not a test weakness — it is a property of the level ordering: `MLP.forward` only checkpoints at `checkpointing >= 2`, which implies `>= 1`, so once the encoder is differentiable the MLP's input already requires grad and the reentrant variant works there. The `layers.py` change is correct and worth keeping (it silences the deprecation warning and is future-proof against torch flipping the default), but it was never the load-bearing part of the bug. Worth knowing if you are deciding how much scrutiny the hunk needs.

**2. The blast radius is probably narrower than "any run with `checkpointing >= 1`".** Both `Distributed.wrap_module` implementations construct `DistributedDataParallel` with the default `find_unused_parameters=False`, so encoder parameters that never receive a gradient should make DDP raise at the first backward rather than train silently. If that holds, the silent-corruption window is single-process training, and multi-GPU runs would have failed loudly. I did not verify this empirically (no multi-GPU here) — flagging it because it changes which past runs need re-auditing. Secondary point in the same direction: `use_reentrant=False` is the variant PyTorch documents as compatible with DDP, so this is also a robustness improvement, not only a correctness fix.

## Side-effect audit of `use_reentrant=False` in our tree (no findings)

- **torch.jit:** the only `@torch.jit.script` uses in `conditional_sfno/` are on free functions (`contractions.py`, `s2convolutions.py`, `layers.py:323`); both modified call sites sit in Python-level methods, and `MLP.checkpoint_forward` is `@torch.jit.ignore`, so the new kwarg is never seen by the scripter.
- **Callable-twice requirement:** the `clip_latent_global_means` envelope update — the one stateful, collective-issuing block in `forward` (`dist.reduce_min`/`reduce_max`, `_gm_min`/`_gm_max` `copy_`) — sits *outside* every checkpointed region, so recompute cannot double-count it or desynchronise ranks. The blocks checkpointed at level 3 contain no collectives (only `dist.get_local_slices`, a metadata call). No in-place mutation of checkpointed inputs, no batchnorm running stats, no dropout at default config.
- **Sibling pass:** I audited every `torch.utils.checkpoint` call site in `fme/`. `fme/ace/models/makani/{layers,sfnonet}.py`, `fme/ace/models/makani_fcn3/*`, `fme/ace/models/ocean/m2lines/{samudra,layers}.py`, `fme/downscaling/modules/physicsnemo_unets_v{1,2}/unets.py` and `fme/core/optimization.py` already pass `use_reentrant=False`. The only remaining defective sites are the five the description names: `fme/ace/models/modulus/sfnonet.py` (3) + `modulus/layers.py` (1) and `fme/downscaling/modules/swinir.py` (1). The description's sibling claim is therefore complete and accurate.
- Silent-failure, behaviour-change, config back-compat, inheritance, and builder-pattern passes: no findings. This PR adds no config field, no branch, and no persisted state; it changes only the kwarg on four existing calls, and the only observable behaviour change is the intended one (gradients now flow). Prior checkpoints load unchanged.

## Two description edits I recommend (I lacked permission to apply them)

`AGENTS.md` routes PR-process detail to comments, and the description feeds the squash message.

1. Drop the parenthetical "(confirmed on this branch: 6 failed before, 6 passed after; full `fme/core/models/conditional_sfno/` suite 119 passed, 3 skipped)" — that is this-branch process evidence, now recorded above.
2. "left out of this PR to keep the cherry-pick clean, tracked separately" — nothing actually tracks the `modulus`/`swinir` sites yet. Since `fme/ace/models/modulus/sfnonet.py` is the SFNO registered as `SphericalFourierNeuralOperatorNet` in `fme/ace/registry/sfno.py` (a live, user-settable `checkpointing` config), this is worth an issue before the claim is made. Suggested replacement:

> `fme/ace/models/makani*` and `fme/ace/models/makani_fcn3/*` already pass `use_reentrant=False`; `fme/core/models/conditional_sfno/` did not. `fme/ace/models/modulus/{sfnonet,layers}.py` (4 sites — the SFNO registered as `SphericalFourierNeuralOperatorNet` in `fme/ace/registry/sfno.py`) and `fme/downscaling/modules/swinir.py` (1 site) carry the same defect; they are deliberately out of scope so the cherry-pick stays identical to upstream.

Also a genuine nit I deliberately did **not** fix, to preserve the byte-identical cherry-pick: the test's `Context` is built with batch 4 (`torch.zeros(4, 0, ...)`) while the input is batch 2, and `img_shape` is a helper default that `_grads_after_backward` re-hardcodes as `torch.randn(2, 2, 9, 18)`. Both are harmless — all context embedding dims are 0, so those tensors are never consumed — but they read as inconsistent. Keeping the cherry-pick identical to upstream is worth more than the cosmetics; noting it so it is not mistaken for an oversight.

The PR is still a draft; `gpu` and `cpu` CI were pending at review time (`build`, `pre-commit`, `cpu-very-fast-no-healpix`, GitGuardian all green). The `gpu` job is the only thing left that could surprise — `assert_close` at default float32 tolerances assumes recompute reproduces activations exactly, which holds when conv algorithm selection is cached per shape but is the one claim I could not test on this machine.

elynnwu:
Claude: PR title and description rewritten via the `write-pr-prose` skill (ran `gather_pr_context.sh 1493` first), applying the pre-review's two recommended edits — the this-branch test-evidence parenthetical is dropped (it lives in the pre-review comment above), and the `modulus`/`swinir` sites are now genuinely tracked as follow-up work rather than only claimed to be.

Two disclosures the skill asks for:

1. **Model.** The skill's frontmatter pins PR prose to `claude-opus-4-6[1m]`. This session is running **Opus 5** (`claude-opus-5[1m]`) from a different repo, where that pin cannot be applied to a subagent, so the prose above was written by Opus 5, not the pinned model. Flagging rather than silently substituting; happy to have it rewritten under the pinned model if that matters here.
2. **Context gaps.** `gather_pr_context.sh` emitted `(warning: could not retrieve title/description)` — `gh pr view`/`gh pr edit` currently fail on this repo with `GraphQL: Projects (classic) is being deprecated ... (repository.pullRequest.projectCards)`, and `gh pr edit` fails *silently* (it leaves the description unchanged while looking like it succeeded). So the prose was written without the script's copy of the prior title/description, though I had them from this session, and was applied via `gh api -X PATCH repos/ai2cm/ace/pulls/1493` instead of `gh pr edit`. Worth knowing if others hit the same wall.

Follow-up work for the sibling sites (`fme/ace/models/modulus/{sfnonet,layers}.py`, `fme/downscaling/modules/swinir.py`) is tracked on my side; say the word if you would rather have a GitHub issue on this repo for it.

elynnwu:
Claude: pushed 13c6afe, a follow-up commit simplifying the test setup, and re-ran the `write-pr-prose` skill to amend the description for it. Naser's commit (04b5c05) is untouched underneath, so the fix and its original test remain byte-identical to upstream.

What the follow-up changes, all in the new tests:

- Folds the single-use `_make_checkpointing_model` into `_grads_after_backward` (AGENTS.md puts the helper threshold at 3+ call sites; the split forced a three-value return whose `device` element existed only to build the input).
- Removes two latent inconsistencies that fold exposed: the image shape was written twice — an `img_shape` argument nothing overrode, plus a re-hardcoded `randn(2, 2, 9, 18)` that would not have followed if it changed — and the `Context` was built with batch 4 against a batch-2 input. The latter is harmless, since every context embedding dim is 0 so those tensors are never consumed, but it reads as a defect.
- Drops `test_checkpointing_does_not_drop_gradients`. It added no detection: `torch.testing.assert_close` raises on a `None` against a tensor, so a dropped gradient already fails `test_checkpointing_gradients_match_uncheckpointed` against its grad-bearing level-0 reference. The surviving test's docstring now says so explicitly. What is lost is only the failure message naming *which* parameters lost their gradients.
- Drops a no-op `model.train()` (`nn.Module` defaults to training) and a redundant `detach()` on a leaf's `.grad`.

Verified: with the fix, `fme/core/models/conditional_sfno/` is 116 passed / 3 skipped; reverting both source files fails all 3 parametrisations of the surviving test. `pre-commit` (ruff, ruff-format, mypy) clean on the touched file. Full CI on the prior commit was green including `gpu` (17m4s), which is what confirms the bit-identical-gradient assertion holds on device at default float32 tolerances — worth a re-check on this head.

Same two disclosures as my earlier comment still apply: the prose was written by **Opus 5**, not the `claude-opus-4-6[1m]` the skill pins (this session runs from another repo and cannot apply that pin), and `gather_pr_context.sh` again emitted `(warning: could not retrieve title/description)` because `gh pr view`/`gh pr edit` fail on this repo with the Projects-classic GraphQL error — `gh pr edit --body` fails *silently*, so the description was applied via `gh api -X PATCH` again.

mahf708 (approved):

jpdunc23 (approved):
LGTM!

PR #1496, opened 2026-09-16
Stop copying data no coupled validation metric reads

elynnwu:
In the cm4 `fto` coupled recipe the atmosphere is frozen — `loss_weight: 0.0`, `n_steps: 0`, so no atmosphere step is ever optimized — but its data window is `n_coupled_steps * 20 + 1` timesteps against the ocean's `n_coupled_steps + 1`, roughly 95% of the data a batch holds. Peak memory is set by how many copies of that window one batch is worth, which caps usable `n_coupled_steps` at 8 on 4xB200 at batch 8 (measured 85.2 GiB at cap 4, 148.2 GiB at cap 8, i.e. 0.79 GiB per atmosphere timestep on a 21.4 GiB base). The long-tail arms need 12-20 and OOM'd at 20 in the first validation.

Three places copied a whole window that nothing then read. Removing them halves the memory a validation batch needs per atmosphere timestep, with bit-identical output.

- `fme.ace.stepper.TrainOutput.ensemble_derive_func` broadcast the entire forcing mapping across ensemble members before calling the derive function. Folding an ensemble dimension into the sample dimension is a copy rather than a view — `expand` is a view, but the `reshape` that merges the two dimensions cannot be a stride — and derive functions ignore forcing entries whose name is already in the data, so every generated variable was copied once per ensemble member only to be discarded. It now broadcasts only the names the data lacks, which for the coupled atmosphere is a handful of exogenous forcings rather than the whole dataset. This was the allocation immediately preceding the observed OOM.
- `fme.coupled.CoupledTrainStepper.train_on_batch` kept the per-step generated tensors alive after `_process_ensemble_output_list` had stacked them, and kept the stacked tensors alive after `prepend_initial_condition` had concatenated them into fresh ones. Both are released once superseded.
- `fme.ace.aggregator.one_step.OneStepAggregator` folded and normalized the whole window, though every one-step metric is pinned to forward step 1 (`MeanAggregator` and the spectrum and ensemble aggregators default to `target_time=1`; the snapshot and map aggregators hard-code it). It now slices the batch to `N_TIMESTEPS_READ` leading timesteps first, via a new `TrainOutput.select_first_timesteps`; the slices are views, so the fold and the normalization no longer touch the unread remainder. This was the largest of the three.

The comments on `fme.core.tensors.repeat_interleave_batch_dim` and `fold_sized_ensemble_dim` claimed these folds were view-only. They are not, and that is why `BatchData.broadcast_ensemble` is expensive on a long window; the comments now say so.

## Measured

`scripts/coupled/coupled_memory_scaling.py`, added here, runs one frozen-atmosphere validation batch (`train_on_batch` with derived variables, then the validation `OneStepAggregator`) under `torch.no_grad()` and reports peak live tensor bytes. At 24x48, 2 samples per device, `n_ensemble` 2, 20 atmosphere steps per ocean step:

| `n_coupled_steps` | before | after |
|---|---|---|
| 4 | 0.571 GiB | 0.276 GiB |
| 12 | 1.694 GiB | 0.820 GiB |
| 16 | 2.255 GiB | 1.092 GiB |
| 20 | 2.816 GiB | 1.363 GiB |

That is 7.185 -> 3.480 MiB per atmosphere timestep, a 52% cut. Tensor bytes are linear in grid area and in samples per device, so on the 1 degree production grid at 2 samples per device this is 0.395 -> 0.191 GiB per atmosphere timestep of live tensor bytes.

The production figures above are device memory, which also carries the caching allocator's overhead; the reduction in device memory is therefore somewhere between the absolute 0.20 GiB per atmosphere timestep this removes (if that overhead is a fixed addition) and the full 52% (if it scales with what is allocated). On 4xB200 at batch 8 that brackets the affordable `n_coupled_steps` between 12 and 19, and on 8xH100 at batch 8 between 9 and 15. The same model reproduces the currently observed ceilings (cap 8 and cap 6-7 respectively), so cap 12 is safe to launch and 16 and 20 are worth probing.

Numerics are unchanged, not merely close: for a frozen-atmosphere coupled validation batch, all 1588 aggregator log values — 664 floats, 252 arrays and 672 images, including the atmosphere channel-mean RMSE the coupled read-outs quote — and a sha256 fingerprint of every gen and target tensor in both realms are bit-identical to this branch's base. Atmosphere validation metrics stay comparable against the already-run crps4 and crps8 arms.

## Not done

- `broadcast_ensemble` in `_accumulate_loss` still copies the whole atmosphere window once per ensemble member, including the target variables the rollout generator never reads. Subsetting it first would cut about two window-copies per timestep, but the transient peak inside `_accumulate_loss` (0.445 GiB at cap 8 in the harness) already sits below the post-change overall peak (0.548 GiB), so it would not lower the validation peak. It is the binding term for the training phase, and becomes worth doing if validation drops further.
- `prepend_initial_condition` rebuilds a target window that is element-for-element the batch already in memory, worth about a sixth of what remains. Reusing the loaded tensors depends on an invariant — initial condition plus forward window equals the loaded batch — that a change to `get_forward_data` could break without any test noticing.
- Derived variables are still computed for the frozen realm. Skipping them would have dropped 11 atmosphere derived variables from the validation metrics, which the coupled read-outs use and which must stay comparable to crps4 and crps8.

Changes:
- `fme.ace.stepper.single_module.TrainOutput`: `ensemble_derive_func` broadcasts only forcing names absent from the data; new `select_first_timesteps`
- `fme.ace.aggregator.one_step.main`: `N_TIMESTEPS_READ`, and `OneStepAggregator.record_batch` slices to it before folding and normalizing
- `fme.coupled.stepper.CoupledTrainStepper.train_on_batch`: releases the per-step and pre-prepend generated-data references
- `fme.core.coordinates.DeriveFnABC.__call__`: documents that a name in both `data` and `forcing_data` is taken from `data`, the contract the derive change relies on
- `fme.core.tensors.repeat_interleave_batch_dim`, `fold_sized_ensemble_dim`: corrected comments about which folds copy
- `fme.core.testing.PeakTensorMemory`: peak live tensor-storage bytes on any device, so memory can be asserted in CPU tests (`fme.core.benchmark.memory` needs CUDA)
- `scripts/coupled/coupled_memory_scaling.py`: the measurement above
- Tests: the one-step aggregator's logs are unchanged by the slice; `ensemble_derive_func` broadcasts only unread forcing and matches per-member derivation; the coupled validation peak stays within a bounded number of window copies per atmosphere timestep

- [x] Tests added

## What `n_coupled_steps` this makes reachable

The saving is on the per-atmosphere-timestep term only; the base term is
unchanged and is **arm-specific**, so size a run against an anchor with the same
ocean init on the same hardware rather than against a shared bracket.

On 4xB200 at `batch_size: 8` (2 samples/GPU) two production runs anchor it, and
both imply the same base:

| anchor | init | `n_coupled_steps` | measured peak | implied base |
|---|---|---|---|---|
| crps4 `cghk183p` | `owc` | 4 | 85.2 GiB | 21.2 GiB |
| crps8 `o0nc2dp3` | `owc8` | 8 | 148.2 GiB | 21.0 GiB |

At the post-change slope that projects, against a 179 GiB card:

| `n_coupled_steps` | projected peak | |
|---|---|---|
| 12 | 112 GiB | 63% |
| 16 | 143 GiB | 80% |
| 18 | 158 GiB | 88% |
| 20 | 173 GiB | 97% — do not |

Cap 16 sits at the headroom crps8 already demonstrated (82.7%).

**The same projection does not transfer to 8xH100 unexamined.** Two cap-4 runs
on that hardware imply very different bases — `mae2n` 18.0 GiB, `owc8v4`
40.2 GiB — and the ~22 GiB difference is not explained by their
`stepper_training` blocks, which are byte-identical. Anchor an H100 run on its
own arm's observed peak.


elynnwu:
Pre-review agent: the cap bracket's arithmetic reproduces, and the slope is corroborated two independent ways — but the **base** term is arm-specific in a way the bracket does not capture, and for an `owc8v4`-shaped run on 8xH100 that moves the answer by about four caps. Worth settling before the next launch.

### What checks out

Both brackets reproduce from the stated inputs. Writing `peak(N) = base + slope * (20N + 1)`:

- The production two-point fit (`crps4` cap 4 = 85.2 GiB, `crps8` cap 8 = 148.2 GiB, 4xB200 @ 2 samples/GPU) gives slope 0.7875 and base 21.4 GiB — the description's numbers exactly.
- The harness's allocated-tensor slope scales to the same place: 7.185 MiB/atm-timestep at 24x48 x (180x360)/(24x48) = 0.395 GiB, and 0.395 x 2 = 0.79. **The two methods agree on the slope independently**, which is the part of the model I'd have most expected to be wrong. That is good corroboration and it deserves to be said out loud in the description, because it is what makes the 52% figure trustworthy rather than merely internally consistent.
- The two bounds are the right pair to draw, and the brackets follow: on 4xB200 the pessimistic (fixed-overhead) slope 0.586 puts cap 12 at 162.6 GiB and the optimistic (proportional) slope 0.382 puts cap 19 at 166.9 GiB. On 8xH100 @ batch 8 the slope halves with samples-per-GPU while the base does not, giving 0.293 / 0.191 against an unchanged 21.4 GiB base — cap 9 and cap 15. That the base is *not* halved is correct and is what makes the model reproduce the observed cap 6–7 ceiling there (cap 7 = 77.1 GiB on an 80 GiB card).

One caveat on the top end: cap 19 at 166.9 GiB is 93% of a 180 GiB B200, and the highest peak actually demonstrated on that hardware is `crps8` at 148.2 GiB, or 82%. The bracket's upper half extrapolates into headroom this configuration has never been shown to reach. That is an argument for the description's own recommendation — launch the pessimistic end — rather than against the bracket.

### Where it breaks: the base is not a constant

The 22 GiB gap between the two cap-4 8xH100 runs is a **base** difference, not a slope difference, and it is large enough to move the cap materially. Backing each run's base out at the pre-change slope of 0.395 GiB/atm-timestep (1 sample/GPU), with 81 atmosphere timesteps at cap 4:

| run | observed peak | implied base |
|---|---|---|
| `mae2n` (`8chmcf8l`) | 50.0 GiB | 18.0 GiB |
| `owc8v4` (`jc8l3r1d`) | 72.2 GiB | 40.2 GiB |

18.0 GiB is consistent with the 21.4 GiB derived on 4xB200 at 2 samples/GPU, as it should be for a term that is mostly weights and optimizer state. `owc8v4` carries about 22 GiB more. I could not find a config explanation: `owc8v4` and `mae2n` have byte-identical `stepper_training` blocks — same `n_coupled_steps: 4`, `n_ensemble: 2`, `batch_size: 8`, same `EnsembleLoss` with `crps_weight: 0.9` on both realms, same frozen atmosphere. The difference is on the module side, which is exactly where a base-term difference would live. **This PR's analysis does not explain the gap, and does not need to** — but the bracket inherits it.

Carrying each base forward against a ~75 GiB practical ceiling (the figure the earlier cap-20 OOM reported):

| run | pessimistic (0.293) | optimistic (0.191) |
|---|---|---|
| `mae2n` | cap 9 | cap 14 |
| `owc8v4` | cap 5 | cap 9 |

So the description's 8xH100 bracket of 9–15 describes an `mae2n`-shaped run well, and is roughly four caps too optimistic for `owc8v4`. Note too that `owc8v4` at cap 4 is already at 90.7% — it is at its own ceiling *now*, so the "observed ceiling cap 6–7" the description cites cannot be describing that arm either.

### Suggestion

Anchor the cap on the target arm's own most recent observed peak rather than on a shared base:

```
peak_new(N) = peak_observed - 0.395 * (20 * N_observed + 1) + slope_new * (20 * N + 1)
```

with `slope_new` between 0.191 and 0.293 on 8xH100 @ 1 sample/GPU, and between 0.382 and 0.586 on 4xB200 @ 2 samples/GPU. The slope reduction this PR delivers is the solid part and applies to every arm; only the intercept needs to come from the arm being launched.

Two consequences worth stating in the description, since a wrong call here costs a launch cycle:

1. The 4xB200 recommendation ("cap 12 is safe, 16 and 20 worth probing") rests on `crps4` and `crps8` sharing a base — they were the two points the slope was fit through, so the fit cannot detect a base difference between them. The harness agreeing on the slope is real evidence they do share one, which is reassuring; it is worth citing as the reason rather than leaving the two-point fit to stand alone.
2. On 8xH100, the bracket should be stated per-arm, or the description should say which arm's base it assumes. As written a reader launching `owc8v4` at cap 9 would OOM.

I have not changed the description — under AGENTS.md that is a `write-pr-prose` pass, and this is a judgement call about what to claim rather than a mechanical fix.

elynnwu:
Pre-review agent: ready for review, with one fix applied to the branch (d5a0782). The three changes claimed bit-identical are bit-identical — I reproduced that independently against the base for a frozen-atmosphere coupled batch across both realms — and the aggregator slice is safe against every sub-aggregator the `OneStepMetricConfig` union is capable of building. Separate comment above covers the cap bracket, which is where I'd want a second opinion before launching.

<details>

### Independent check of the bit-identity claim

Rather than take the description's numbers, I ran a fingerprint harness on this branch and on `origin/exper/2026-07-16-jamesd`, in separate worktrees and separate processes, over one frozen-atmosphere coupled validation batch (`n_ensemble` 2, both realms, derived variables on):

- 184 sha256 tensor fingerprints — every `gen_data`, `target_data` and tensor-valued metric in **both** realms
- all 1588 aggregator log values: 664 scalars by value, 252 arrays and 672 images by sha256 of their contents

`diff` was empty, and the 664/252/672 split reproduced exactly. So the claim holds, and it holds specifically for the coupled frozen-atmosphere path rather than only for uncoupled ACE.

I also confirmed the change lands where the OOM was: `evaluate_before_training` in `fme/core/generics/trainer.py` calls `validation_callback`, which is the `validate_stepper` closure in `fme/coupled/train/train_config.py` that builds the coupled `OneStepAggregator` and runs `run_validation_loop`. That is exactly the code path these three changes touch.

### ✅ 1. Silent-failure pass

"Every one-step metric is pinned to forward step 1" is a claim about all current and future sub-aggregators, so I checked whether it is enforced structurally or merely true today. It is structural:

- `OneStepBuildContext` carries no `n_forward_steps` and no `n_ic_steps`. `StepMeanMetricConfig` and `EnsembleMetricConfig` — the two configs in those same modules that *do* take a configurable `step` — build against `MetricBuildContext`, the inference-side context, and are not members of the `OneStepMetricConfig` union. None of the five union members has a `step` or `target_time` field at all. `OneStepMeanMetricConfig` and `OneStepSpectrumMetricConfig` take `MeanAggregator`/`SpectrumAggregator`'s `target_time=1` default, `OneStepEnsembleMetricConfig` passes `target_time=1` literally, and map and snapshot hard-code index 1 (snapshot also reads index 0). No YAML can push a one-step metric past window index 1.
- `log_zonal_mean_images` and `log_histograms` in the cm4 coupled configs sit on the **inference evaluator** aggregator, a different class this PR does not touch. `InlineValidationConfig.aggregator` is typed `OneStepAggregatorConfig` and nothing else, so the production coupled validation path has no knob that reaches past step 1. A test run under defaults and a production run exercise the same metric set here.
- A window shorter than `N_TIMESTEPS_READ` is returned whole, so the existing `time_len > target_time` and `i_time_start` range guards behave exactly as before — the slice introduces no new quiet skip.

Worth knowing, though not a problem today: `one_step/spectrum.py` also defines a `PairedSphericalPowerSpectrumAggregator` that records the **whole** window at `i_time_start=0`. Nothing builds it — no metric config references it — so it cannot reach the slice, but it is the one class in that module that would break silently if a future metric wired it in.

### ⚠️ 2. Behavior-change pass — one gap, fixed in d5a0782

`test_logs_ignore_timesteps_past_the_first_forward_step` is the tripwire for the entire slice, and it filtered the sliced-vs-unsliced comparison to `float | int`. That left the array and image metrics unguarded, which is backwards: those are produced by map and snapshot, the two aggregators whose target step is a hard-coded index rather than a configured one, so a drift there is the *least* likely to announce itself.

Demonstrated rather than asserted: with `one_step/map.py` changed to `.mean(dim=time_dim)` instead of selecting index 1 — the realistic silent drift — all ten scalars still matched and the test passed, while every `mean_map` image had changed.

d5a0782 compares all three kinds by value (scalars by `repr`, arrays and images by sha256 of their contents), raises on a kind the comparison does not handle so a new one cannot be skipped, and asserts all three kinds are present so the comparison cannot narrow back to scalars if a metric default changes. Verified it fails on the drifted map and passes on this branch.

The other modified paths are clean:

- `del output_list` — last read is the `_process_ensemble_output_list` call immediately above; `torch.stack` copies, so nothing observable changes.
- `del ocean_stepped, atmos_stepped, ocean_gen_data, atmos_gen_data` — `stepped` still holds those objects until `stepped = stepped.prepend_initial_condition(ic)` rebinds, so the `del` is what lets the rebind actually drop them. Load-bearing, not incidental cleanup.
- `ensemble_derive_func` — the new filter predicate (`k not in flattened_data`) is the exact complement of the derive functions' `if key not in data`, so the names excluded are precisely the ones that were being materialized and discarded. The `n_ensemble == 1` branch is untouched and still passes the whole mapping, which costs nothing there since no broadcast happens.

### ✅ 3. Differential-test pass

- `test_coupled_validation_memory_scales_with_window` — I ran it against base with the branch's test file: **16.4** copies/timestep against the 10.0 threshold, versus 8.1 on this branch. Real differential with margin, and the 16.4 → 8.1 ratio independently corroborates the claimed 52% cut from a completely different measurement than the harness table.
- `test_ensemble_derive_func_broadcasts_only_the_forcing_it_will_read` and the rewritten `test_train_output_compute_derived_quantities` assert the forcing key set exactly (`{"forcing_only"}` and `{"f"}`), which under base would be `{"shared", "forcing_only"}` and `{"a", "b", "f"}` — both fail if the change is no-oped.
- The rewrite of `test_train_output_compute_derived_quantities` is a fair characterization and a genuine strengthening. The old body asserted `forcing_data["a"].shape == (6, 4, 8)` — that a name already present in `data` was broadcast anyway — inside a `try/except AssertionError` that fell through to a second branch, so a real failure in the gen-call assertions could be masked by the target-call ones passing. The new version captures both calls, drops the `except`, and adds a forcing-only name so there is something the derive function can only get from forcing. Not a test weakened to fit the change.

### ✅ `PeakTensorMemory` is a trustworthy instrument

The whole memory argument rests on this, so I validated it rather than reading it. Storage `PyObject` preservation holds (`t.untyped_storage() is t.untyped_storage()`), so the `id()` key is stable and 51 repeated `track` calls count once; peak tracks an interleaved alloc/free pattern exactly (24.0 MiB expected, 24.0 measured); views are free and a `clone` is not; the weakref callback fires on free with no accounting leak across 20 sequential allocations; and 200 wrapper create/destroy cycles do not trigger an early free. It depends on `torch.utils._python_dispatch` and `torch.utils._pytree`, both private — acceptable for test-only infrastructure, but that is what would break on a torch upgrade, and it is worth a line in the module docstring saying so.

### ✅ 4. Sibling / mirror coverage

The `fme/ace` change and its coupled consumer travel together, as they should. `fme/coupled/aggregator.py`'s `OneStepAggregator` delegates to `OneStepAggregatorConfig.build()` for both realms, so the slice reaches ocean and atmosphere by construction rather than by a second edit that could drift.

### ✅ 5. Config back-compat

No config schema, dacite union, or checkpoint-persisted state is touched. `N_TIMESTEPS_READ` is a module constant, not a config field. Prior checkpoints load for inference unaffected.

### ✅ 6. Split heuristic

12 files, one coherent change plus the measurement that justifies it. No new typed config, no back-compat shim, no refactor bundled with a feature. Not a split candidate.

### ✅ 7. Abstract-or-final / inheritance depth

`PeakTensorMemory` is the only `TorchDispatchMode` subclass in the repo and has no in-repo subclass. `TrainOutput` has no in-repo subclass, so `select_first_timesteps` needs no `@final`. Depth 1 throughout.

### ⚠️ 8. AI-agent artifact pass

Clean on the usual patterns — no `Any` returns, no `isinstance` + `type: ignore`, no imports inside test bodies, no abbreviated identifiers, no unrelated docstring or comment removals. The corrected comments on `repeat_interleave_batch_dim` and `fold_sized_ensemble_dim` are a real fix: the old `# view-only, no copy` was wrong and is the reason this cost was invisible.

One typing observation, not worth changing in this PR: the contract the derive change relies on is documented on `DeriveFnABC.__call__`, but `TrainOutput.derive_func` is annotated `Callable[[TensorMapping, TensorMapping], TensorDict]`, so nothing at the type level binds a caller to it. In practice every production path assigns `stepper.derive_func`, which `build_derive_function` guarantees is a `DeriveFnABC`; I checked all three concrete implementations (`NullDeriveFn`, `AtmosphericDeriveFn`, `OceanDeriveFn`) and both `_compute_derived_variable` and `_compute_ocean_derived_variable` honor `if key not in data`. Only tests pass bare lambdas. Tightening the annotation would be a bigger change than this PR should carry, but it is the gap between "documented contract" and "enforced contract".

### ✅ 9. Builder pattern

No config field read from outside its own class; no new builder/impl pair introduced.

### Test runs

`pre-commit` (ruff, ruff-format, mypy) passes over the full PR range including d5a0782.

`fme/ace fme/coupled fme/core` run on this branch and on `origin/exper/2026-07-16-jamesd` with the same interpreter and seed settings:

| | passed | failed | skipped |
|---|---|---|---|
| this branch | 3085 | 9 | 35 |
| base | 3081 | 9 | 35 |

The failure sets are byte-identical between the two (`diff` empty) — 7 in `fme/core/dataset/test_xarray.py` (zarr), plus `test_ice_train.py::test_train_and_inference` (an `ImportError` on a missing optional dep) and `test_train.py::test_resume_two_workers`. All pre-existing and unrelated to this PR. The +4 on this branch are the tests it adds.

</details>

PR #1497, opened 2026-09-16
Make CorrectorConfigABC.remove_deprecated_keys abstract

mcgibbon:
`CorrectorConfigABC.from_state` calls `cls.remove_deprecated_keys(state)` before strict dacite loading, but the hook had a no-op default implementation. This makes it an abstract classmethod so every registered corrector config must explicitly state its deprecations, aligning with the same contract used by sibling ABCs.

Changes:
- `fme.core.corrector.registry.CorrectorConfigABC.remove_deprecated_keys`: made `@classmethod @abc.abstractmethod`; updated docstring to document the contract (must return new dict, never mutate input, implement as `return dict(state)` when nothing to remove)
- `fme.core.corrector.atmosphere.AtmosphereCorrectorConfig`: added no-op `remove_deprecated_keys`
- `fme.core.corrector.ice.IceCorrectorConfig`: added no-op `remove_deprecated_keys`
- `fme.core.registry.corrector.CorrectorSelector`: added no-op `remove_deprecated_keys`
- `fme.core.corrector.ocean.OceanCorrectorConfig`: already had a real implementation, unchanged

- [x] Tests added


mcgibbon:
Claude: Pre-review of #1497

**1. Silent-failure pass** -- clean. The change moves from a silent no-op default to an explicit abstract requirement, which is the right direction. Every subclass now must state its deprecation handling.

**2. Behavior-change pass** -- clean. All four in-tree subclasses (`AtmosphereCorrectorConfig`, `IceCorrectorConfig`, `OceanCorrectorConfig`, `CorrectorSelector`) have `remove_deprecated_keys` implemented. Out-of-tree subclasses (via `CorrectorSelector.register`) would need to add the method, but this is a correct and expected consequence of the abstract contract.

**3. Differential-test pass** -- clean. `test_corrector_selector_remove_deprecated_keys_builds` would fail if `remove_deprecated_keys` were no-oped (dacite strict mode would reject the unstripped `old_key`). `test_corrector_selector_unknown_key_raises` verifies the complement (unhandled keys still raise).

**8. AI-agent artifact pass** -- clean. PR description matches the diff; no stale claims or invented references.

**9. Split heuristic** -- single concern across 5 files; no split needed.

Pre-commit hooks pass on all changed files. All 14 tests in `test_registry.py` pass (including the 2 new ones).

jpdunc23 (approved):
LGTM!

PR #1498, opened 2026-09-16
Add remove_deprecated_keys hook to ModuleConfig builders

mcgibbon:
ModuleConfig (the base of every module builder selected by ModuleSelector) had no hook for rewriting a checkpoint's config dict before dacite loads it, so deleting config fields would break old inference checkpoints. CorrectorConfigABC already solved this with a remove_deprecated_keys classmethod called from a @final from_state. This PR brings the same pattern to ModuleConfig.

Changes:
- `fme.core.registry.module.ModuleConfig`: `from_state` is now `@final` and calls `cls.remove_deprecated_keys(state)` before dacite strict loading. A new `@classmethod @abc.abstractmethod remove_deprecated_keys` requires every subclass to implement the hook (no-op when there is nothing to remove).
- All registered builders (`SphericalFourierNeuralOperatorBuilder`, `SFNO_V0_1_0`, `LandNetBuilder`, `NoiseConditionedSFNOBuilder`, `HEALPixUNetBuilder`, `SwinTransformerBuilder`, `NoiseConditionedSwinTransformerBuilder`, `AnkurLocalNetBuilder`, `LocalNetBuilder`, `PreBuiltBuilder`, `SamudraBuilder`, `FloeNetBuilder`, `MLPConfig`): no-op `remove_deprecated_keys` (`return dict(state)`).
- `fme.core.registry.test_module_registry`: `MockModuleBuilder.from_state` override removed (now uses the base @final version); all mock builders implement the hook. New `MockModuleBuilderWithDeprecation` and three tests verify: deprecated key dropping and renaming, selector.config normalization, and input immutability. Existing `test_module_selector_raises_with_bad_config` continues to verify that genuinely unknown keys raise `dacite.UnexpectedDataError`.

- [x] Tests added


mcgibbon:
Claude: Pre-review of #1498

## Audit passes

1. **Silent-failure pass** -- `remove_deprecated_keys` is `@abc.abstractmethod`, so any new `ModuleConfig` subclass that omits the hook gets a `TypeError` at class creation. `from_state` is `@final`, preventing silent bypass. No silent-failure paths.
2. **Behavior-change pass** -- `from_state` gains `@final` and now routes through `remove_deprecated_keys` before dacite loading. No real `ModuleConfig` subclass previously overrode `from_state`; the only override was `MockModuleBuilder` in tests, which is correctly removed. The no-op hooks (`return dict(state)`) are equivalent to the old behavior. Return type refined from `"ModuleConfig"` to `Self` (backward compatible).
3. **Differential-test pass** -- All three new tests fail when the code path is no-oped: `test_remove_deprecated_keys_drops_and_renames` (dacite strict rejects unknown keys), `test_remove_deprecated_keys_preserves_selector_config` (normalized config would contain deprecated keys), `test_remove_deprecated_keys_does_not_mutate_input` (without the `dict(state)` copy, mutation would occur). Existing `test_module_selector_raises_with_bad_config` still catches genuinely unknown keys via `dacite.UnexpectedDataError`.
4. **Convention-conformance** -- N/A (code-only PR).
5. **Provenance** -- N/A (no run/commit references).
6. **Coverage / sibling pass** -- All 13 registered `ModuleConfig` subclasses updated, plus all 3 test mock builders. Verified via `grep -rn 'class.*ModuleConfig\b'` -- no stragglers. `downscaling/` has its own `ModuleConfig` Protocol (not inheriting from `fme.core.registry.module.ModuleConfig`), correctly unaffected.
7. **Index/link integrity** -- pre-commit (ruff, ruff-format, mypy) passes on all changed files. All 11 registry tests pass.
8. **AI-agent artifact pass** -- PR description accurately lists all 13 builders and 3 tests. No invented links, stale phrases, or drifted claims.
9. **Split heuristic** -- Single concern (adding the deprecation hook) across 11 files. No split needed.
10. **Writing style** -- PR description is clear and factual. The "mirroring" claim about CorrectorConfigABC is accurate at the pattern level; the deliberate difference (abstract vs default implementation) is explicitly called out.

## Assessment

The PR cleanly extends `ModuleConfig` with the `remove_deprecated_keys` hook, correctly mirroring the established `CorrectorConfigABC` pattern (with a stricter, abstract variant). The `@final` on `from_state` prevents subclasses from accidentally skipping the hook. All subclasses are covered, tests are well-designed, and pre-commit passes clean.

No fixes applied; no findings.

brianhenn (approved):
LGTM.

Only thing that came to mind is whether the FCN3 module in ace/fme/ace/step/fcn3.py (not a registered module, apparently in its own one-off registry?) should be covered by this? Probably not, though separately raises questions about maintaining that registry/module.

mcgibbon:
> LGTM.
> 
> Only thing that came to mind is whether the FCN3 module in ace/fme/ace/step/fcn3.py (not a registered module, apparently in its own one-off registry?) should be covered by this? Probably not, though separately raises questions about maintaining that registry/module.

I agree on "no". The step is registered, but it uses its own registry rather than ModuleConfig, because the FCN3 fundamentally needs to be able to distinguish between surface (which are treated like context) and 3D variables, and I think also needs to receive those 3D fields on an aligned vertical grid. It also doesn't allow diagnostic outputs, I think, and each level is independently encoded.

Generally speaking I think if FCN3 poses a maintenance burden we should remove it, if we really wanted to proactively remove it I wouldn't block that but it's also not something I'd ask for.

PR #1499, opened 2026-09-16
Add remove_deprecated_keys hook to StepConfigABC

mcgibbon:
`StepConfigABC` subclasses each re-implemented `from_state` with the same dacite-strict-loading boilerplate, and only two of them hooked a private `_remove_deprecated_keys` before loading. This consolidates the pattern established by `CorrectorConfigABC`: a concrete `@final from_state` on the ABC that calls `remove_deprecated_keys` before dacite strict loading. Unlike `CorrectorConfigABC` (where the hook has a default no-op), `StepConfigABC` makes the hook abstract so every subclass explicitly addresses deprecated keys. Future config field deprecations only need to edit the hook on the affected subclass.

Changes:
- `fme.core.step.step.StepConfigABC`: `from_state` is now concrete `@final`; new abstract `remove_deprecated_keys` classmethod
- `fme.core.step.step.StepSelector`: deleted `from_state`, added no-op `remove_deprecated_keys`
- `fme.core.step.single_module.SingleModuleStepConfig`: deleted `from_state` and `_remove_deprecated_keys`; added public `remove_deprecated_keys` preserving `crps_training` removal
- `fme.ace.step.fcn3.FCN3StepConfig`: same pattern (no-op hook, `_remove_deprecated_keys` was already a plain copy)
- `fme.core.step.radiation.SeparateRadiationStepConfig`, `fme.core.step.secondary_module.SecondaryModuleStepConfig`, `fme.core.step.multi_call.MultiCallStepConfig`: deleted `from_state`, added no-op hook
- `fme.core.step.test_step_registry`: added `DeprecatingMockStepConfig` with a non-trivial hook (drops a key, renames another) and tests for the hook mechanism, strict-mode preservation, input non-mutation, and `crps_training` removal

- [x] Tests added


mcgibbon:
Claude: Pre-review of #1499

**1. Silent-failure pass** -- clean. The abstract hook forces explicit handling; dacite strict catches unknown keys after the hook runs.

**2. Behavior-change pass** -- clean. Every deleted `from_state` is faithfully replicated by the ABC's consolidated version. Verified all six subclasses plus `StepSelector` against their `origin/main` implementations.

**3. Differential-test pass** -- clean. All four new tests would fail if their target behavior were removed. The `test_remove_deprecated_keys_drops_and_renames` test exercises the full pipeline (StepSelector -> registry -> from_state -> hook -> dacite), not just the method in isolation.

**8. AI-agent artifact pass** -- one fix applied (see below).

**9. Split heuristic** -- clean. Single concern, 7 files.

### Fixes applied

- **Moved inline `import dacite` to module level** (test_step_registry.py line 317): the inline import inside `test_remove_deprecated_keys_unknown_key_raises` was left over from replacing the old top-level dacite import with pytest. Moved it back to the top.
- **Updated PR description**: the original description said CorrectorConfigABC uses "an abstract `remove_deprecated_keys` classmethod", but CorrectorConfigABC's hook is non-abstract with a default no-op. Corrected to accurately describe the divergence.

### Open question for the reviewer

The hook is `@abc.abstractmethod` on `StepConfigABC`, requiring every subclass to implement it (5 of 7 are pure `return dict(state)` no-ops). `CorrectorConfigABC` took the other approach: a non-abstract default no-op that subclasses override only when needed. Both are defensible -- abstract is more explicit, non-abstract saves boilerplate. Worth aligning the two ABCs on one convention?

### Verdict

**Ready for review** after the fixes applied above.

fme/core/step/test_step_registry.py
```diff
@@ -1,9 +1,11 @@
 import dataclasses
 import datetime
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
+from typing import Any
 from unittest.mock import MagicMock
 
 import dacite
```
mcgibbon, line 7: Claude: Moved the inline `import dacite` to the module-level imports (commit f3fd5a6ac). The inline import was left over from replacing the old top-level dacite import with pytest.

mcgibbon:
GPU test failure appears to be unrelated:
```
FAILED fme/downscaling/test_predict.py::test_predictor_runs - struct.error: unpack requires a buffer of 8912 bytes
```

jpdunc23 (approved):
LGTM!

PR #1500, opened 2026-09-16
Drop deprecated fields from NoiseConditionedSFNOBuilder

mcgibbon:
Eight fields on `NoiseConditionedSFNOBuilder` were kept solely for backwards compatibility and never affected the built module. This PR removes them from the dataclass and handles legacy configs through the `remove_deprecated_keys` hook added in the parent PR.

The three fields that previously raised `ValueError` on non-default values (`factorization`, `separable`, `operator_type`) still raise from the hook, preserving the error messages. The other five (`spectral_transform`, `rank`, `complex_network`, `complex_activation`, `spectral_layers`) are silently dropped.

Changes:
- `fme.ace.registry.stochastic_sfno.NoiseConditionedSFNOBuilder`: remove the eight deprecated field declarations and their `__post_init__` checks; implement `remove_deprecated_keys` to pop them (with validation for the three that must match their old defaults)
- `fme/core/registry/testdata/NoiseConditionedSFNO_module_config.yaml`: remove deprecated keys from the frozen fixture
- `configs/`: strip deprecated keys from all six `NoiseConditionedSFNO` config blocks
- `fme/core/registry/test_module_registry.py`: remove deprecated keys from the "latest" builder config; add tests for legacy-key loading and the three error paths

- [x] Tests added


mcgibbon:
Claude: Pre-review of #1500.

**Audit passes:**

1. Silent-failure pass: ✅ — The three validated keys (`factorization`, `separable`, `operator_type`) default to their safe values when absent; the five silently-dropped keys use `pop(key, None)` appropriately.
2. Behavior-change pass: ✅ — The three `ValueError` checks moved from `__post_init__` to `remove_deprecated_keys` with identical messages, and the hook fires at the same point in the loading flow (`ModuleSelector.__post_init__` → `Registry.get` → `from_state`). Serialized `ModuleSelector.config` (e.g. wandb logging) no longer carries the 8 fields, which is the intended consequence.
3. Differential-test pass: ✅ — All four new tests are discriminating: `test_ncsfno_legacy_keys_load_successfully` would fail under a no-op hook (dacite strict rejects unknown keys); the three error-path tests fail under a no-op hook because dacite raises `UnexpectedDataError`, not the expected `ValueError`.
4–7. (Records passes: N/A — code-only PR.)
8. AI-agent artifact pass: ✅ — Description accurately lists 8 fields, names the correct 3 that raise, counts 6 config files (matching the branch; the 4 additional NoiseConditionedSFNO configs on main post-date the branch point and are handled by the hook at runtime).
9. Split heuristic: ✅ — Single purpose, 9 files.
10. Writing style: ✅ — Description is concise and accurate.

**Coverage/sibling check:** The remaining `operator_type`/`spectral_transform`/`spectral_layers` references in `configs/` are all under `SphericalFourierNeuralOperatorNet` (the non-noise-conditioned builder in `sfno.py`), where those fields are still active and passed to the model. No stragglers.

**Overall:** The PR does what it says, the hook preserves backward compatibility for old configs and checkpoints, and the tests cover both the happy path and all three error paths. Ready for review.

mcgibbon:
Note no training configuration examples need to be updated, none use NoiseConditionedSFNO with these keys.

PR #1501, opened 2026-09-17
Add `"prescribed_open_ocean"` hfds correction

jpdunc23:
Adds the `method="prescribed_open_ocean"` option to `SurfaceEnergyFluxCorrectionConfig`. `hfds` is prescribed from forcings only on cells that are entirely ice-free ocean, and the network prediction passes through unweighted everywhere else.

Changes:
- Add `"prescribed_open_ocean"` to `SurfaceEnergyFluxCorrectionConfig.method`.

- [x] Tests added


elynnwu (approved):
Potentially controversial, should we remove/deprecate "prescribed" option? I know we have a lot of checkpoints with that so it'll be annoying, but feels like that's something we think it's wrong and shouldn't support.

jpdunc23:
> Potentially controversial, should we remove/deprecate "prescribed" option? I know we have a lot of checkpoints with that so it'll be annoying, but feels like that's something we think it's wrong and shouldn't support.

I think it makes sense to deprecate this at some point in the future. Let's leave as is for now in case we need to run additional experiments with existing checkpoints.

PR #1505, opened 2026-09-18
Switch inference IC sharding from strided to contiguous blocks

mcgibbon:
Switches inference IC sharding from strided (IC `i` assigned to rank `i % W`) to contiguous blocks (rank `r` gets ICs `[r·n/W, (r+1)·n/W)`). This is a prerequisite for multi-GPU data writers: contiguous blocks make zarr region-writes a single slice and gather-to-root concatenation produce output in natural sample order.

No behavioral change under single-rank inference. The inline inference path during training already uses this same `InferenceDataset`, so it also moves to contiguous blocks, which is scientifically neutral since every aggregator reduces across ranks.

Changes:
- `fme.ace.data_loading.inference.local_ic_range` — new helper returning the `[start, end)` IC range for a rank, with a divisibility check that replaces the existing bare assertion
- `fme.ace.data_loading.inference.InferenceDataset._get_batch_data` — uses `local_ic_range` instead of the strided `i % W` filter

- [x] Tests added


elynnwu (commented):
Can you make this switch for coupled inference too (use the same `local_ic_range` helper)? We should do this before the multi-GPU writer to avoid future issues.

mcgibbon:
> Can you make this switch for coupled inference too (use the same `local_ic_range` helper)? We should do this before the multi-GPU writer to avoid future issues.

Will do. I'd really like to make a change that could unify coupled and ace inference under a single entrypoint, so we can stop having to make these duplicate updates, but I'll punt on it for now.

mcgibbon:
Claude: Done — applied the same contiguous-block switch to coupled inference (`fme/coupled/data_loading/inference.py`). The coupled `_get_batch_data` now uses `local_ic_range` instead of the strided `i_member % world_size` pattern. All 30 coupled data loading tests pass.

elynnwu:
> I'd really like to make a change that could unify coupled and ace inference under a single entrypoint

Agree, we should just do that refactor so we don't have to keep track of this duplication.

elynnwu (approved):

PR #1510, opened 2026-09-19
Finish an epoch-boundary resume instead of re-entering the trained epoch

elynnwu:
`_save_restart_checkpoints` runs *before* `_epochs_trained` is incremented, deliberately, so that a resume re-runs the validation and inference an interrupted job never reached. The side effect is that such a checkpoint carries a full epoch of `current_epoch_num_batches_seen`, and resuming from it re-enters an epoch whose batch subset is empty: `train_one_epoch` logs `Subsetted train loader created, has 0 batches`, trains nothing, then spends the whole post-epoch train-evaluation pass building a train summary for an epoch it did not train.

On an 8xH100 coupled run that last step died, twice, on two different nodes, in `PerStepLossAggregator.get_logs`:

```
TrainAggregator.get_summary -> PerStepLossAggregator.get_logs -> dist.gather_object
RuntimeError: NCCL Error 2: unhandled system error   (ranks 4, 5)
RuntimeError: NCCL Error 1: unhandled cuda error     (rank 0)
```

Ranks 4/5/6 failed `ncclProxyCallBlockingUDS` / `ncclProxyClientGetFd` to rank 0 over `/tmp/nccl-socket-0-*`, with 205-234 GB of host RAM still free (exit 1, not a `-9` OOM kill). A fresh epoch and a resume with 509 batches left both passed the same collective, so the correlation is with the zero-batch epoch; the mechanism linking a step-free epoch to the proxy fd exchange is not established here. What is established is that the state is reachable, deterministic, and costs ~95 min per retry before it fails.

Detecting the empty subset removes the state: the epoch is already trained, so finish it and fall through to the validation the checkpoint was written to get.

Changes:
- `fme.core.generics.Trainer.train_one_epoch`: when a resumed epoch's batch subset is empty, increment `_epochs_trained`, reset `_current_epoch_num_batches_seen`, and return an empty `AggregatorSummary` rather than running the train-evaluation pass and reducing a summary.

- [x] Tests added


elynnwu:
Pre-review agent (Claude): the fix is correct and narrowly scoped — I traced every invariant the early return skips and none is observably broken, and I confirmed the new test is discriminating by reverting the hunk (it fails on `validation_batches_seen == []`). I pushed one commit (`ca4a186`) softening the new log line and recording the dropped train-evaluation pass in the comment; one description gap is noted below for the author.

<details>

## Applied in `ca4a186`

The log line read "Resumed from a checkpoint written at an epoch boundary" — a stated cause, but the branch condition is only "the resumed subset is empty". If the training dataset shrank between runs, the same branch fires and the message misdirects. It now reports what was observed. The code comment also now records that the early return drops the epoch's train-evaluation pass.

## Audit passes

**1. Silent-failure pass — ⚠️ (addressed)**
The only new branch is `len(epoch_data) == 0`, and it is nested under `if self._current_epoch_num_batches_seen > 0`, so a fresh epoch over a genuinely empty loader is untouched. The one quiet case is a resume whose training dataset shrank below `_current_epoch_num_batches_seen`: the epoch is then declared finished rather than cut short. Per AGENTS.md, resuming across config changes is not guaranteed, and the pre-existing line above already logs the skip count and `has 0 batches`, so this is a documentation matter, not a behavior one — handled by the log change.

**2. Behavior-change pass — ✅ (one delta, intended)**
I diffed the old and new paths through `train_one_epoch` for `len(epoch_data) == 0`. Everything skipped is either re-done downstream or has no observable effect:

- `_started_training = True` — only read by `save_restart_checkpoints_on_terminate`, which is gated on `_current_epoch_num_batches_seen > 0`. That is 0 after the epoch on *both* paths, so the handler was already unreachable here.
- `_last_saved_num_batches_seen` — only read at the `num_batches_seen > _last_saved_num_batches_seen` check later in the same call, which the early return precedes. No read-before-write.
- `stepper.set_eval()` / `seed_eval(0)` — the stepper is left in train mode, but `fme/core/generics/validation.py:53` and `inference_one_epoch` each call `set_eval()` and `seed_eval(0)` themselves, so nothing leaks into validation or inference.
- `train_data.alternate_shuffle()` — the next epoch's `set_epoch` re-seeds the sampler, so skipping it is a no-op.
- `_epochs_trained += 1` / `_current_epoch_num_batches_seen = 0` — identical to the old path, which also advanced the counter after burning the train-evaluation pass.

The real delta: that epoch's `train/*` wandb metrics and its `epoch_NNNN` train diagnostics are no longer written. No in-repo consumer globs epoch diagnostics directories, so nothing breaks; the visible effect is a one-epoch gap in the train panels on a boundary resume, and `epoch_train_seconds` near zero for it.

`loss=None` is safe through `train()`: it lands in `train_loss`, which only feeds a log line. `_update_best_metrics` and `step_scheduler` both take `valid_loss`, not it. `logs={}` spreads into `all_logs` harmlessly.

**3. Differential-test pass — ✅**
I removed the fix hunk and reran: `test_resume_at_epoch_boundary_does_not_re_enter_the_trained_epoch` fails on `assert stepper.validation_batches_seen == []` with `[4, 6]` (the train-evaluation pass), and the other 62 tests in the file still pass. `assert stepper.train_batches_seen == []` is non-discriminating on its own — `[]` either way — but the test's comment already says so and it documents intent.

**4. Sibling / mirror pass — ✅**
`Trainer` has no in-repo subclass. `fme/downscaling`'s separate `Trainer` classes carry no `current_epoch_num_batches_seen` and no mid-epoch resume, so there is no mirror to fix. No `fme/coupled` trainer.

**5. Config back-compat pass — ✅**
No config or checkpoint schema change; the checkpoint keys read and written are unchanged.

**6. Split heuristic — ✅** Two files, one hunk.

**7. Abstract-or-final / inheritance pass — ✅** No class or signature change.

**8. AI-agent artifact pass — ✅** No `Any`, no `isinstance`+`type: ignore`, no in-function imports, no stale log lines or abbreviations.

**9. Builder pattern pass — ✅** No config touched.

## Cross-rank safety (called out because a divergence would deadlock)

`len(epoch_data) == 0` cannot differ across ranks. `_current_epoch_num_batches_seen` is restored on every rank from the same `latest_checkpoint_path`, and `subset` derives the length deterministically from it and the `DistributedSampler`'s per-rank index count, which is padded equal across ranks. Pre-fix, all ranks equally saw 0 batches and all entered the train-evaluation pass together; post-fix all take the early return together. No new collective asymmetry.

## Also verified

- Epoch accounting matches an uninterrupted run exactly. The pre-existing `test_resume_after_interrupted_training[_validation_callback]` already reaches this state and its `set_epoch` call sequence is unchanged.
- Idempotent under repeated preemption: if the job dies again during the boundary epoch's validation, the abort handler does not save (batches is 0), so the next resume re-enters the same branch and re-runs validation. Converges once validation completes and `save_all_checkpoints` writes `epoch=N, batches=0`.
- `evaluate_before_training` stays skipped on resume (guarded on `_current_epoch_num_batches_seen == 0`), and `_maybe_tune_lr` still short-circuits for the boundary epoch, as before.
- `fme/core/generics/` + `fme/ace/test_train.py`: 124 passed. `pre-commit` (ruff, ruff-format, mypy) clean on both files.

## For the author

The description does not mention that the boundary-resume epoch now contributes no `train/*` metrics or train diagnostics. That is the change's only user-visible behavior delta, and "spends the whole post-epoch train-evaluation pass building a train summary for an epoch it did not train" reads as pure waste — the pass does produce a valid evaluation of the model at that epoch. Worth a sentence via the `write-pr-prose` skill (I did not edit the prose myself, since that skill pins its own model).

## Pre-existing, not for this PR

`_last_saved_num_batches_seen` has no initializer in `__init__`; it exists only because `train_one_epoch` assigns it before its only read. Safe today, including on the new path, but an `AttributeError` waiting for a fourth path through that method.

</details>

PR #1511, opened 2026-09-21
Add configurable parameters for hf upload

frodre:
The Hugging Face upload script hardcodes `HF_XET_HIGH_PERFORMANCE=1` in the Docker image with no way to override it at runtime. This aggressive parallelism can cause upload timeouts on stores containing multiple large (>4 GB) files. Additionally, when an upload fails partway through there is no way to resume without re-uploading files that already succeeded.

This PR makes both behaviors configurable in `example.sh` so operators can tune them per job without rebuilding the image.

Changes:
- `scripts/upload_to_hugging_face/Dockerfile`: document that `HF_XET_HIGH_PERFORMANCE` can be overridden at runtime and note the >4 GB timeout risk
- `scripts/upload_to_hugging_face/example.sh`: add `HF_XET_HIGH_PERFORMANCE` env var (default 1, set to 0 to disable) forwarded to the container via `gantry --env`; add `IGNORE_EXISTING` env var (default 0, set to 1 to pass `--ignore-existing` to `hf sync`) for resuming failed uploads without re-uploading completed files

- [ ] Tests added (script-only change, no testable code)
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated (N/A)


scripts/upload_to_hugging_face/example.sh
```diff
@@ -12,6 +12,21 @@ STORE=abrupt4xCO2-ic_0001.zarr
...
+
+# Set to 1 to pass --ignore-existing, which skips files already present at the
+# destination rather than re-uploading them. Useful for resuming a failed sync so
+# it doesn't redo completed work. Left off by default so a fresh run performs a
+# full sync (uploading new files and re-uploading any that changed at the source).
+IGNORE_EXISTING=${IGNORE_EXISTING:-0}
+SYNC_ARGS=()
+if [ "${IGNORE_EXISTING}" = "1" ]; then
+    SYNC_ARGS+=(--ignore-existing)
+fi
```
spencerkclark, line 28: In practice I found `hf sync` to be quite fast in determining which files did not require re-syncing, since I think it only checks the file sizes and modification times rather than computing and comparing full checksums ([documentation](https://huggingface.co/docs/huggingface_hub/main/en/guides/buckets#comparison-modes)).  Was that not the case for you?

scripts/upload_to_hugging_face/Dockerfile
```diff
@@ -15,6 +15,10 @@ ENV PATH="/root/.local/bin:$PATH"
 
 # Running without both results in roughly 10x slower peak upload throughput for
 # the same store on ai2/phobos.
+#
+# HF_XET_HIGH_PERFORMANCE is a default that can be overridden at runtime (e.g.
+# via gantry `--env HF_XET_HIGH_PERFORMANCE=0`); its aggressive parallelism can
+# cause upload timeouts on stores with multiple large (>4GB) files.
```
spencerkclark, line 21: Thanks—in hindsight I probably should not have baked these into the image and just set them from gantry.

scripts/upload_to_hugging_face/example.sh
frodre: I wasn't sure and it was hard to tell from the UI whether it was skipping files or not.  This allowed me to be sure that it was skipping successfully uploaded files.

scripts/upload_to_hugging_face/Dockerfile
frodre: In the end, I'm not sure if I actually needed this.  I got about 2K of the 3K files uploaded and I would continually run into timeout.  For whatever reason, the session I had open in using image would timeout on even a small batch of files (batching by around 20-objects even failed at one point).  I restarted the session and then it went fine for small batch uploads.  Maybe there was some intermediate state of the local caching operation that wasn't playing well with the actual HF commit.

spencerkclark: Interesting, yeah, it seems like their interface can still be a bit flaky at times.

scripts/upload_to_hugging_face/example.sh
spencerkclark: Sounds good.  It skipped files that were already in the bucket automatically for me, but no harm in keeping this option around.  

I looked into reproducing this and I agree, their UI can confusingly make it seem like some files were 100% uploaded, but for some reason do not show up in the bucket if the encompassing upload command was interrupted.  At least in my experiment, since those files did not end up in the bucket, they appeared to need to be re-uploaded upon resumption, whether `--ignore-existing` was provided or not.

spencerkclark (approved):
I'm good to merge this, though feel free to tweak more if you have more updates based on your experience.

PR #1514, opened 2026-09-21
Extend torch.compile to every step type and module builder

yyexela:
Add `compile: bool` to every step type so that `torch.compile` can be turned on from YAML config. Each step compiles every module it owns after distributed wrapping, so the distributed wrapper is inside the compiled graph and state-dict keys are unchanged (checkpoint compatibility is preserved).

### Semantics

- **single_module**: compiles the main module and the secondary decoder.
- **secondary_module**: compiles the main module, the secondary module, and the secondary decoder.
- **separate_radiation**: compiles the main module and the radiation module.
- **FCN3**: compiles its single module via a dual reference (`_forward_module`), since the FCN3 network is not a registry `Module`.
- **multi_call**: owns no modules; inherits compilation from its `wrapped_step`.
- **prebuilt**: compile support depends on the module the user supplies.

`Module.compile()` sets the process-global `torch._dynamo.config.fail_on_recompile_limit_hit` so a static-shape recompile-limit hit raises instead of silently falling back to eager. Varying the batch size uses dynamic shapes and does not trip it.

Builders can declare `torch.compile` unsupported via `ClassVar compile_unsupported_reason` on their `ModuleConfig`. `Module.compile()` then raises `NotImplementedError` with the reason instead of letting dynamo trace for minutes before failing. **FloeNet** uses this: it rebuilds mesh graphs from numpy and uses boolean-mask indexing, producing data-dependent shapes.

### Builder fixes (bit-identical numerics)

All changes below preserve numerical identity with the previous implementations, verified by reference-comparison tests:

- **`irfft` `.imag` setattr** (`fme/fft.py`, `fme/core/disco/_fft.py`): rewritten through `torch.view_as_real` because dynamo cannot trace `Tensor.imag` setattr.
- **`NullTimer.child`** (`fme/core/benchmark/timer.py`): converted to a `@contextlib.contextmanager` generator; dynamo graph-breaks on arbitrary objects used as context managers but traces through generator-based ones.
- **`CappedGELU` `.item()`** (healpix and m2lines copies): replaced `.item()` + scalar clamp with `torch.clamp(x, max=self.cap)` against the 0-d buffer directly.
- **Samudra** (`fme/ace/models/ocean/m2lines/samudra.py`): `isinstance(x, A | B)` changed to tuple form (dynamo cannot evaluate a class union); numpy shape arithmetic replaced with plain int arithmetic (dynamo traces numpy arrays as tensors, then data-dependent reads break the graph).

### Remaining graph breaks

| Builder | Graphs | Breaks | Reason |
|---|---|---|---|
| MLP | 1 | 0 | - |
| prebuilt | 1 | 0 | - |
| LandNet | 1 | 0 | - |
| SphericalFourierNeuralOperatorNet | 1 | 0 | - |
| SFNO-v0.1.0 | 1 | 0 | - |
| Samudra | 1 | 0 | - |
| AnkurLocalNet | 1 | 0 | - |
| LocalNet | 1 | 0 | - |
| NoiseConditionedSFNO | 1 | 0 | - |
| SwinTransformer | 1 | 0 | - |
| NoiseConditionedSwinTransformer | 1 | 0 | - |
| HEALPixUNet | 1 | 0 | - |
| NoiseConditionedSFNO (clip_latent_global_means) | 3 | 2 | Data-dependent branching around the tracked-envelope check |
| FCN3 (step module) | 16 | 15 | Sparse tensors in torch_harmonics DISCO (not fixable locally) |
| FloeNet | - | - | Excluded: declares compile unsupported |

*(torch 2.7.1, CPU, eval mode, `torch._dynamo.explain`)*

The FCN3 sparse-tensor breaks and the NoiseConditionedSFNO data-dependent branch are documented on their respective config fields.

### Follow-ups

- Consolidate the two `CappedGELU` copies and the two `irfft` copies.
- `SecondaryDecoder.wrap_module` mutates in place; with a `Checkpoint` wrapper it would replace `_module` (pre-existing, latent).
- `prebuilt` compile support depends on the module the user supplies.

Changes:
- `fme/core/registry/module.py` (`Module`, `compile_torch_module`, `ModuleConfig.compile_unsupported_reason`): compile infrastructure on the core `Module` type
- `fme/core/step/single_module.py` (`SingleModuleStepConfig.compile`): compile flag for single-module steps
- `fme/core/step/secondary_module.py` (`SecondaryModuleStepConfig.compile`): compile flag for secondary-module steps
- `fme/core/step/radiation.py` (`SeparateRadiationStepConfig.compile`): compile flag for separate-radiation steps
- `fme/ace/step/fcn3.py` (`FCN3StepConfig.compile`): compile flag for FCN3 steps via dual module reference
- `fme/core/step/multi_call.py`: docstring noting compile is inherited from wrapped step
- `fme/core/step/secondary_decoder.py` (`SecondaryDecoder.compile`): compile passthrough for the secondary decoder
- `fme/fft.py`, `fme/core/disco/_fft.py` (`irfft`): `.imag` setattr replaced with `view_as_real` for traceability
- `fme/core/benchmark/timer.py` (`NullTimer.child`): generator-based context manager for dynamo traceability
- `fme/ace/models/healpix/healpix_activations.py`, `fme/ace/models/ocean/m2lines/activations.py` (`CappedGELU`): `.item()` removed
- `fme/ace/models/ocean/m2lines/samudra.py` (`Samudra.forward`): isinstance tuple form, int arithmetic
- `fme/ace/registry/m2lines.py` (`FloeNetBuilder.compile_unsupported_reason`): declares compile unsupported
- `fme/ace/registry/stochastic_sfno.py` (`NoiseConditionedSFNOBuilder.clip_latent_global_means`): documents graph breaks
- `fme/ace/registry/test_module_compile.py`: parametrized builder compile coverage
- `fme/core/testing/compile.py` (`compile_backend`, `dynamo_hygiene`): test helpers

- [x] Tests added
- [ ] If dependencies changed, "deps only" image rebuilt and "latest_deps_only_image.txt" file updated — dependencies unchanged


yyexela:
Pre-review agent: this PR is now stacked. Base is #1520 (SecondaryDecoder.wrap_module copy fix), which is stacked on #1519 (torch.compile traceability fixes: irfft, NullTimer.child, CappedGELU, Samudra). Those two carry the bit-identical refactors and the bug fix that were previously bundled here; this PR now holds only the compile infrastructure, the step flags and the builder coverage tests. Review #1519 and #1520 first; the zero-graph-break rows in the table below depend on #1519.

Two points worth a reviewer's attention, neither changed by the pre-review:

1. **Process-global dynamo flag.** `compile_torch_module` sets `torch._dynamo.config.fail_on_recompile_limit_hit = True`. This also governs the pre-existing `@torch.compile`-decorated kernels in `fme/ace/models/makani_fcn3/mpu/layer_norm.py`, which this PR does not edit: with any step's `compile=True`, a recompile-limit hit there now raises instead of falling back to eager. Default `recompile_limit` is 8; train/eval and grad-mode guards consume a few of those per compiled region.

2. **`Module.wrap_module` on a compiled `Module` invokes the wrapper twice**, once on the parameter-owning module and once on the compiled callable. Only the second result is used on the forward pass; the first becomes a transient `_module` (a plain function under `Checkpoint`). It works and training-through-checkpointing is tested, but an alternative is to wrap only the forward callable when compiled and leave `_module` untouched. Design call for the reviewer.

